PR #285, opened 2019-10-14
(fix) show amount of outstanding emburse transfers when processing

tmb:
image

one thing to watch out for before merge, there are a few old load requests with `transfer in progress` that have been sitting for months. they may have been tests or something, idk, but we should get rid of them before this goes in

closes #264


thesephist (approved):
dope

PR #297, opened 2019-10-29
(fix) g suite domain should never be an email address

tmb:
closes #294


app/models/g_suite_application.rb
```diff
@@ -34,6 +34,10 @@ def domain_without_protocol
     errors.add(:domain, 'shouldn’t include http(s):// or ending /') if bad
   end
 
+  def domain_not_email
+    errors.add(:domain, 'shouldn’t be an email address') if domain.include? "@"
```
lachlanjc:
```suggestion
    errors.add(:domain, 'shouldn’t be an email address') if domain.include? '@'
```

tmb:
going to merge because no activity & small change

PR #310, opened 2019-11-02
Add a link from auth page back to the Bank landing page

thesephist:
Screen Shot 2019-11-02 at 16 17 18

I found that some people were still going to bank.hackclub.com and
getting confused, because they didn't go to the landing page. This isn't
super helpful, but it's a step in the right direction in that if they
want to get set up with bank, it'll redirect them to the HCB landing
page.

I also want to start using the 'HCB' acronym more and more publicly, in
the case that we want to transition to HCB as our canonical name instead
of Hack Club Bank / "Bank" / "Hack Bank", so I'm calling it that here.

Open to better phrasings. I tried "Sign up for HCB" and "Apply to HCB"
but they're all awkward.

Fix #305


tmb:
adding to zone/theo

PR #320, opened 2019-11-06
(feature) show previous lcrs to users

tmb:
image


thesephist:
Can we make "show all/collapse" more like "show pending / show all"?

I also wonder if it makes more sense to put this in the LCR creation page or in the cards page itself

thesephist:
Also, can we have an empty state?

Screen Shot 2019-11-06 at 15 47 03

thesephist:
Otherwise, LGTM to merge. when those things are addressed. I think it's a little wonky that there are two tables next to each other that are different widths, but we can iterate on it in prod.

thesephist:
Also: "created by" would sound better to users as "requested by"

app/views/events/card_overview.html.erb
```diff
@@ -118,6 +118,34 @@ West Hollywood, CA 90069</pre>
...
+        <tr>
+          <th>Requested at</th>
+          <th>Requested by</th>
+          <th>Status</th>
+          <th>Amount</th>
+        </tr>
+      </thead>
+      <tbody>
+        <% @event.load_card_requests.each do |lcr| %>
+          <tr <%= "under-review" if lcr.under_review? %> id="lcr_row">
```
lachlanjc: Did you mean to make this a class?

PR #336, opened 2019-11-23
Fix a couple of auth-related UI bugs

thesephist:
- Redirect GET requests to login code POST routes back to /users/auth
- Fix issue where incorrect login codes will show a login code form with no


config/routes.rb
```diff
@@ -17,6 +17,10 @@
       post 'login_code', to: 'users#login_code'
       post 'exchange_login_code', to: 'users#exchange_login_code'
       delete 'logout', to: 'users#logout'
+
+      # sometimes users refresh the login code page and get 404'd
```
lachlanjc: THANK YOU!

PR #340, opened 2019-11-25
Allow admins to reject check requests

thesephist:
Rejected checks are never synced with Lob.

Closes #332


thesephist:
Would be nice if the admin could leave a message as the rejection reason, but this is fine for now as we normally end up contacting the team about the reason we reject checks anyways.

thesephist:
**NOTE to whoever merges this PR**:

This PR contains a migration, but the migration should be run automatically and gracefully by our Heroku postdeploy script. You should watch to make sure that this is the case, and file an Issue against this repo if that is not the case.

tmb (approved):
if this is tested lgtm

PR #341, opened 2019-11-25
Add comments to checks, ACH transfers

thesephist:
Useful for leaving accounting documentation / papertrail attached to direct deposits and ACH transfers.


lachlanjc:
Was talking to Matthew yesterday & he had no idea this feature existed in Bank. That isn’t a crisis, but I think the majority of our users don’t know about the notes feature. Though I know about them I’ve never used it.

thesephist:
Yeah, that's an interesting point. We do a terrible job in the UI in general of surfacing features we don't cover in the onboarding call. I also don't really think there are great use cases for Notes in particular outside of recording papertrail. As an admin managing 100+ events I've only used it like 3 times ever.

On the broad problem of surfacing feature utility, I think that's something that requires further high level design maybe also related to rethinking the onboarding flow.

PR #343, opened 2019-11-25
When removing organizer, also remove any pending invites they sent

thesephist:
Close #327


lachlanjc:
Not positive, but couldn’t this just be a cascade: :destroy in the model? Or because of the cancel method can we not?

thesephist:
I prefer not to destroy models in general unless necessary, because soft-deleting allows for fixing mess ups easily and cascading-destroy isn't always what you want.

In this particular case, we can't cascade-destroy because we want to keep invites around if they've been accepted, so we only want to remove pending ones.

Also, yeah, I wanted to use the cancel method since it's there, rather than just blowing away the model.

lachlanjc:
Got it, didn't realize you couldn't override the destroy with the cancel, but completely agree

PR #346, opened 2019-11-26
Donation pages

tmb:


tmb:
will be finishing this up after thanksgiving

app/views/donations/_donation_form.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  <div class="field">
+    <%= form.label :email %>
+    <%= form.text_field :email %>
+  </div>
+
+
+  <%= form.label :amount %>
+  <div class="field">
+    <div class="flex items-center">
+      <span class="bold muted" style="width: 1rem;">$</span>
```
lachlanjc: This would be an excellent candidate for a helper/CSS class/abstraction of some sort

https://github.com/hackclub/bank/search?l=HTML%2BERB&q=form+amount

thesephist: Yep. Probably for a different pr tho

thesephist:
I'm telling GitHub this closes #81

tmb:
last step here is testing + doing the front-end thing that invoices have. will be doing within next 2 days

tmb:
After @thesephist does a frontend pass, this is ready for beta. 

**TODO before merge:**
* Setup webhook on stripe

**TODO before 🚢:**
* Make the fee reimbursement look nice in the list like an invoice
* Setup receipt sending functionality

image

image

thesephist:
@tmb Can you merge master into this branch and resolve conflicts before I take a last pass? to resolve schema conflicts, you'll have to import the db fresh and re-migrate.

db/schema.rb
```diff
@@ -27,6 +26,7 @@
     t.datetime "approved_at"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.text "recipient_phone_number"
```
thesephist: This change is (1) unrelated to anything else happening here, and (2) not needed since I fixed this in an earlier PR. Can you remove the migration that causes this?

app/controllers/donations_controller.rb
```diff
@@ -0,0 +1,87 @@
+class DonationsController < ApplicationController
+  before_action :set_donation, only: [:show, :edit, :update, :destroy]
+  skip_after_action :verify_authorized, only: [:start_donation, :make_donation, :finish_donation, :accept_donation_hook]
+  skip_before_action :signed_in_user, only: [:start_donation, :make_donation, :finish_donation, :accept_donation_hook]
+
+  # GET /donations
+  def index
+    @donations = Donation.all
```
thesephist: Should this (`index`) and `new` below both `authorize @donation(s)`?

app/controllers/donations_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    authorize @donation
+  end
+
+  # GET /donations/new
+  def new
+    @donation = Donation.new
+  end
+
+  # GET /donations/1/edit
+  def edit
```
thesephist: This should probably `authorize @donation` and/or throw since you probably shouldn't be able to edit a donation after creation.

app/controllers/donations_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    donation = Donation.find_by_stripe_payment_intent_id(request.params['data']['object']['id'])
+
+    pi = StripeService::PaymentIntent.retrieve(id: donation.stripe_payment_intent_id, expand: ['charges.data.balance_transaction'])
+    donation.set_fields_from_stripe_payment_intent(pi)
+    donation.save
+
+    # create donation payout
+    donation.queue_payout!
+
+    # something something email
```
thesephist: Please finish, this seems unfinished. If you can write a bare-bones email I can polish the wording and write actually good copy.

app/controllers/fee_reimbursements_controller.rb
```diff
@@ -9,7 +9,12 @@ def index
 
   # GET /fee_reimbursements/1
   def show
-    @event = @fee_reimbursement.invoice.event
+
```
thesephist: Can we build this logic into `FeeReimbursement` so that `fee_reimbursement.event` is always the right thing, and fee reimbursements controller doesn't have to know about invoices and/or donations?

app/jobs/sync_transactions_job.rb
```diff
@@ -116,6 +118,24 @@ def check_for_check(transaction)
     end
   end
 
+  def check_for_donation(transaction)
+    DonationPayout.all.each do |payout|
```
thesephist: Can we use the `DonationPayout.lacking_transaction` scope that's a single SQL query and only loop through pending DPs? This (doing `.all`) won't scale well if we keep looping over everything.

app/models/donation.rb
```diff
@@ -0,0 +1,78 @@
...
+    self.payout_creation_balance_stripe_fee = b_tnx.fee
+    self.payout_creation_balance_available_at = funds_available_at
+
+    self.save!
+  end
+
+  def create_payout!
+    pi = StripeService::PaymentIntent.retrieve(id: stripe_payment_intent_id, expand: ['charges.data.balance_transaction'])
+
+    # raise StandardError, 'Funds not yet available' unless Time.current.to_i > pi.charges.data.first.balance_transaction.available_on
```
thesephist: Why is this commented out? It seems like this is the case where it should either (1) never happen, or (2) if it's too early, it should re-queue the job for two days later or something. right?

app/models/donation.rb
```diff
@@ -0,0 +1,78 @@
...
+      donation: self
+    )
+
+    self.fee_reimbursement.save
+
+    # if a transfer takes longer than 5 days something is probably wrong. so send an email
+    fee_reimbursement_job = SendUnmatchedFeeReimbursementEmailJob.set(wait_until: DateTime.now + 5.days).perform_later(self.fee_reimbursement)
+    self.fee_reimbursement.mailer_queued_job_id = fee_reimbursement_job.provider_job_id
+
+    self.fee_reimbursement.save
```
thesephist: Is there a reason we're saving this twice, given that we already call `.save` on this reimbursement in line 51?

app/models/donation_payout.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  # find donation payouts that don't yet have an associated transaction
+  scope :lacking_transaction, -> { includes(:t_transaction).where(transactions: { donation_payout_id: nil }) }
+
+  # although it normally doesn't make sense for a paynot not to be linked to an donation,
+  # Stripe's schema makes this possible, and when that happens, requiring donation<>payout breaks bank
+  has_one :donation, inverse_of: :payout, foreign_key: :payout_id
+  has_one :t_transaction, class_name: 'Transaction'
+
+  validates_length_of :statement_descriptor, maximum: 22
```
thesephist: Do you know why this is 22? That needs to be documented.

app/models/donation_payout.rb
```diff
@@ -0,0 +1,76 @@
...
+  def dropdown_description
+    "##{self.id} | #{self.stripe_created_at.to_date} - #{render_money self.amount} - #{self.donation&.name}"
+  end
+
+  private
+
+  def default_values
+    return unless donation
+
+    self.statement_descriptor ||= "DONATION #{SecureRandom.hex(6)}"[0...22] # limit to 22 characters, the stripe limit
```
thesephist: Can we make this 22 some global constant like `StripeService::StatementDescriptorCharLimit` or something so we don't have to keep hard-coding this number into random places in the codebase? I don't like seeing magic numbers in old code :/

app/models/fee_reimbursement.rb
```diff
@@ -60,9 +65,9 @@ def admin_dropdown_description
   # this needs to exist for the case where amount of reimbursement is less than $1 and we need to do fee weirdness
   def calculate_fee_amount
     if amount < 100
-      return (amount * self.invoice.event.sponsorship_fee) + (100 - amount)
+      return invoice.present? ? (amount * self.invoice.event.sponsorship_fee) + (100 - amount) : (amount * self.donation.event.sponsorship_fee) + (100 - amount)
     else
-      amount * self.invoice.event.sponsorship_fee
+      invoice.present? ? amount * self.invoice.event.sponsorship_fee : amount * self.donation.event.sponsorship_fee
```
thesephist: Woah, these lines are getting really unwieldy. can we break this ternary down into more if-else's to make these more readable? I would not trust my future self to be able to parse out what any of this code does at a glance. I think it should be easy to just look at this `calculate_fee_amount` and understand exactly what each line of code is doing.

I think a simple refactor here would be to just create a `FeeReimbursement.event` getter (I alluded to this in an earlier comment as well) and use that here to get the sponsorship fee instead of doing this weird ternary thing every single time.

app/models/transaction.rb
```diff
@@ -23,6 +23,8 @@ class Transaction < ApplicationRecord
   belongs_to :check, inverse_of: :t_transactions, required: false
   belongs_to :ach_transfer, inverse_of: :t_transaction, required: false
 
+  belongs_to :donation_payout, inverse_of: :t_transaction, required: false
```
thesephist: This is minor, but let's put this line right next to the `:invoice_payout` line just to keep ourselves more organized.

app/views/donations/_form.html.erb
```diff
@@ -0,0 +1,47 @@
...
+    <div id="error_explanation">
+      <h2><%= pluralize(donation.errors.count, "error") %> prohibited this donation from being saved:</h2>
+
+      <ul>
+      <% donation.errors.full_messages.each do |message| %>
+        <li><%= message %></li>
+      <% end %>
+      </ul>
+    </div>
+  <% end %>
```
thesephist: I'm being sort of repetitive here these days but let's use our `form_errors` helper here. Replacing this whole block with `<%= form_errors donation %>` should work, I think?

app/views/donations/_form.html.erb
```diff
@@ -0,0 +1,47 @@
...
+  </div>
+
+  <div class="field">
+    <%= form.label :status %>
+    <%= form.text_field :status %>
+  </div>
+
+  <div class="field">
+    <%= form.label :stripe_client_secret %>
+    <%= form.text_field :stripe_client_secret %>
```
thesephist: The client secret should never ever _ever_ **ever** leave the server. If there's any way we can avoid putting it in the form, we should absolutely totally do it. Otherwise this is like exposing our Stripe password to the rest of the world. Correct me if I'm wrong?

app/views/donations/index.html.erb
```diff
@@ -0,0 +1,37 @@
...
+
+<table>
+  <thead>
+    <tr>
+      <th>Email</th>
+      <th>Name</th>
+      <th>Amount</th>
+      <th>Amount received</th>
+      <th>Status</th>
+      <th>Stripe client secret</th>
```
thesephist: Can you explain to me what the stripe client secret is? Is it not just one secret for all of Hack Club Bank? Why does each donation need a separate `stripe_client_secret`?

app/views/events/_form.html.erb
```diff
@@ -10,6 +10,10 @@
     <%= form.text_area :address %>
   </div>
 
+  <% if event.beta_features_enabled %>
+    <%= link_to 'Give this link to other people if they want to donate to you', start_donation_donations_path(event.slug) %>
+  <% end %>
```
thesephist: Note mostly to self: before we officially push ship on this, we should make a new per-event page where we tell people how Donations works, and give them this link. Putting it in the events form is jank. But this PR can be merged w/o it.

app/views/transactions/_form.html.erb
```diff
@@ -63,6 +63,17 @@
...
+        <div class="field">
+          <%= form.label :donation_payout, 'Donation payout' %>
+          <%= form.collection_select(
+            :donation_payout_id,
+            DonationPayout.lacking_transaction.reorder(:amount) + [transaction.donation_payout].compact, # .compact to remove nil values if transaction.donation_payout is nil
+            :id,
+            :dropdown_description,
+            include_blank: 'Select one…'
+          ) %>
+        </div>
```
thesephist: It seems weird to have to go through two drop-down menus to find a matching Payout. Is there a way we can make the Donation payouts immediately distinguishable from TX memo separate from Invoice payouts? Maybe instead of `HACK CLUB EVENT XXX` we use `DONATION EVENT XXX` or something.

config/routes.rb
```diff
@@ -25,6 +25,9 @@
     post 'delete_profile_picture', to: 'users#delete_profile_picture'
   end
 
+  # webhooks
+  post 'webhooks/donations', to: 'donations#accept_donation_hook'
```
thesephist: What does Stripe do if this hook is down for some reason?

thesephist (changes requested):
Left some comments, mostly questions about failure cases and how things work. Specifically I'm curious about this `stripe_client_secret` and why each Donation needs a separate one -- feel like that should just work like an API key.

I would have merged this, but having an extra unneeded migration in here is a deal breaker. Let's take that out and get to the bottom of the Stripe secret issue, and I'm find with merging other stuff as-is and resolving them on master, or fixing them here if it's quick.

db/schema.rb
tmb: think my local copy was out of sync

app/controllers/donations_controller.rb
tmb: this is left over from scaffolding, need to remove

app/models/donation.rb
tmb: oops- artifact from testing

app/models/donation_payout.rb
tmb: Yup- That’s a Stripe-imposed limit. Will add comment

app/views/donations/index.html.erb
tmb: `stripe_client_secret` is the unique token for each PaymentIntent that Stripe.js uses to process the payment

app/models/donation_payout.rb
thesephist: Actually this should probably be a constant somewhere in `StripeService` -- see comment below.

app/views/donations/index.html.erb
thesephist: Ahhhhh ok interesting. So it's meant to be exposed on the client-side but only for like the one donation-maker. That makes sense. If that's the case disregard my other complaints about the secret.

app/controllers/donations_controller.rb
tmb: fixed by removing unused methods

app/controllers/fee_reimbursements_controller.rb
tmb: yes but that's for a future PR. @event needs to exist because of the way our views are setup.

app/jobs/sync_transactions_job.rb
tmb: done!

app/models/donation.rb
tmb: yeah, we're adding in that `mailer_queued_job_id` value

app/views/donations/_form.html.erb
tmb: sorry autogenerated stuff again :(

config/routes.rb
tmb: nothing. afaik it will just retry—all this affects is payout generation + receipt sending, so if the hook is down, that doesn't matter anyway (because bank is down)

app/views/transactions/_form.html.erb
tmb: Took a screenshot and adding to checklist for 🚢/future

tmb:
Ready for merge after your pass! @thesephist

app/controllers/fee_reimbursements_controller.rb
thesephist: Made https://github.com/hackclub/bank/issues/394 to track this!

app/views/donations/_form.html.erb
thesephist: I was wrong here. nvm

thesephist (approved):

PR #348, opened 2019-11-28
Show nicknames in the badge under the header in index page

thesephist:
yeety yeet

Screen Shot 2019-11-27 at 22 33 29


app/views/static_pages/index.html.erb
```diff
@@ -6,9 +6,7 @@
   <h1 class="heading flex-column items-center justify-center mt2 w-100">
     <span class="h3 caps slate block">Welcome to</span>
     <span class="primary">Hack Club Bank</span>
-    <% if Rails.env.development? %>
-    <span class="badge h5 caps">Development Mode</span>
-    <% end %>
+    <span class="badge h5 caps"><%= random_nickname() %></span>
```
lachlanjc: This isn’t JS, you can omit the parens :wink:

thesephist: I kind of like it there actually because it tells me it's a function and not a variable that I'm just dropping in, so it' has notational value when I'm reading the code -- it was placed there intentionally.

If it breaks some sacred Rails convention, I guess I'll remove it, but I think it serves a purpose to the reader.

lachlanjc: Yeah, that’s not really a differentiation made in Ruby/Rails—I’ve never seen the parens used anywhere

PR #362, opened 2019-12-07
Allow fully voided checks to be refunded, not just pending_void ones

thesephist:
- Also includes some code cleanup
- Fixes #357


thesephist:
@tmb you know the Check state machine better than I do, so can you sanity-check to make sure that in particular my change in `app/models/checks.rb` makes sense and is correct?

app/models/check.rb
```diff
@@ -126,7 +122,7 @@ def in_transit?
 
   # if a void was put in and the sum of transaction is neutral (no money lost or gained)
   def voided?
-    approved? && voided_at.present? && t_transactions.size == 4 && t_transactions.sum(&:amount) == 0 || !approved_at && voided_at
+    approved? && voided_at.present? && t_transactions.size == 4 && t_transactions.sum(&:amount) == 0 || !approved_at? && voided_at?
```
thesephist: Should double-check that adding question marks don't break things. A better approach would be to do `approved_at.present?` and `voided_at.present?`

PR #374, opened 2019-12-12
Fix check/transaction auto matching error

tmb:
based on Bugsnag. Checks have multiple transactions but transactions only have one check.


thesephist (approved):
🧳

PR #377, opened 2019-12-14
parallax cards

maxwofford:
~Still a WIP, shouldn't be merged~

Got bored & added this:
https://www.youtube.com/watch?v=tNqP1ucQYwI


thesephist:
Woaw nice.

I think we want to reduce motion for when the user CSS preference is set to reduce motion, and maybe make the tilt a little more subtle (I like the amount that our hackathons site has). But this is something I also wanted to add for a while.

I'll take a stab at the above points when I get time, but feel free to finish them if you want 🤷‍♂️

thesephist (changes requested):
See my above comments 4 minor things :)

maxwofford:
I just toned down the tiling effect (now uses the old scaling factor, & shimmer is subtle). https://youtu.be/2hKbomgWCJc

Working on CSS prefers reduced motion next

app/assets/javascripts/ui.coffee
```diff
@@ -158,10 +158,15 @@ $(document).on 'turbolinks:load', ->
...
     maxTilt: 15
     speed: 400
     perspective: 1500
     glare: true
     maxGlare: .25
     scale: 1.0625
+
+  window.matchMedia('(prefers-reduced-motion: reduce)').addEventListener 'change', -> 
+    tilt.methods.destroy.call(tilt)
+    console.log(mediaQuery.media, mediaQuery.matches);
```
thesephist: This should probably be removed?

maxwofford: Couldn't get this running locally so I'm using Heroku deploys as my development env. will let you know when I'm ready for review

maxwofford:
BAM https://youtu.be/6viKa8_Kqlw

maxwofford:
Ok, laptop's at 2% and they just called my boarding group for the flight. Will leave any other tweaks / changes up to your discretion.

maxwofford:
(Think I addressed everything in your comments)

thesephist (approved):
☕️

PR #378, opened 2019-12-14
Fix retrofees modal in dark mode

lachlanjc:
Before:
Screen Shot 2019-12-14 at 2 00 44 PM
After:
Screen Shot 2019-12-14 at 2 01 16 PM


thesephist (commented):
💰💰💰

PR #379, opened 2019-12-14
Add super-subtle transition to filterbar items

lachlanjc:
The instant flashing of color change on the filters was slightly jarring, so toned it down.


thesephist:
When you make Ui changes, it'd be nice to attach screenshots or videos so we don't have to spin up  / log into PR deployments to test it. Just a tip 👍

thesephist (approved):
🌊

PR #404, opened 2019-12-21
Fixes upper-casing

seanvictory:
Fixes upper-casing in "Fee Refunds" to make it consistent with the casing on the rest of the platform.


thesephist (approved):
never have such monumental and groundbreaking changes been merged into hackclub/bank. Amazing

PR #406, opened 2019-12-23
Stop notifying admins about new transactions via email

tmb:
Closes #403


thesephist (dismissed):
dope as hell m8

thesephist:
Can we also remove `app/mailers/transaction_mailer.rb`? @tmb

PR #407, opened 2019-12-23
Add the ability to hide events

tmb:
image

image


app/controllers/events_controller.rb
```diff
@@ -22,6 +22,11 @@ def create
     fixed_event_params[:club_airtable_id] = nil if event_params[:club_airtable_id].empty?
     fixed_event_params[:partner_logo_url] = nil if event_params[:partner_logo_url].empty?
 
+    fixed_event_params[:hidden_at] = DateTime.now if fixed_event_params.key?(:hidden) && fixed_event_params[:hidden] == "1" && !@event.hidden_at.present?
+    fixed_event_params[:hidden_at] = nil if fixed_event_params.key?(:hidden) && fixed_event_params[:hidden] == "0" && @event.hidden_at.present?
```
thesephist: Why do we set it twice? Can't we just determine a value once and set it?

also, this boolean expression is really hard to read. Can we split it into if-else's that makes it possible to tell what's going on at a glance?

app/controllers/events_controller.rb
```diff
@@ -78,11 +83,24 @@ def update
...
     fixed_event_params = event_params
+    fixed_user_event_params = user_event_params
 
     fixed_event_params[:club_airtable_id] = nil if event_params.key?(:club_airtable_id) && event_params[:club_airtable_id].empty?
     fixed_event_params[:partner_logo_url] = nil if event_params.key?(:partner_logo_url) && event_params[:partner_logo_url].empty?
 
-    if @event.update(current_user.admin? ? fixed_event_params : user_event_params)
+    # processing hidden for admins
+    fixed_event_params[:hidden_at] = DateTime.now if fixed_event_params.key?(:hidden) && fixed_event_params[:hidden] == "1" && !@event.hidden_at.present?
+    fixed_event_params[:hidden_at] = nil if fixed_event_params.key?(:hidden) && fixed_event_params[:hidden] == "0" && @event.hidden_at.present?
```
thesephist: Since this seems to be getting repeated, can we maybe move it out into a separate method? Same with `club_airtable_id` and `partner_logo_url` above :/

app/controllers/events_controller.rb
```diff
@@ -78,11 +83,24 @@ def update
...
+    
+    fixed_event_params.delete(:hidden)
+
+    # processing hidden for users
+    fixed_user_event_params[:hidden_at] = DateTime.now if fixed_user_event_params.key?(:hidden) && fixed_user_event_params[:hidden] == "1" && !@event.hidden_at.present?
+    fixed_user_event_params[:hidden_at] = nil if fixed_user_event_params.key?(:hidden) && fixed_user_event_params[:hidden] == "0" && @event.hidden_at.present?
```
thesephist: Can you help me understand why there is a hidden for admins but also a hidden for users? Isn't it just one?

app/controllers/events_controller.rb
```diff
@@ -78,11 +83,24 @@ def update
...
+    byebug
```
thesephist: is this supposed to be left in here?

app/models/event.rb
```diff
@@ -1,7 +1,9 @@
 class Event < ApplicationRecord
   extend FriendlyId
 
-  default_scope { order(id: :asc) }
+  default_scope { order(id: :asc)}
```
thesephist: pls fix formatting

app/models/event.rb
```diff
@@ -1,7 +1,9 @@
...
+  scope :hidden, -> { where.not(hidden_at: nil) }
+  scope :not_hidden, -> { where(hidden_at: nil)}
```
thesephist: pls fix formatting

app/views/events/_form.html.erb
```diff
@@ -14,6 +14,11 @@
     <%= link_to 'Give this link to other people if they want to donate to you', start_donation_donations_path(event.slug) %>
   <% end %>
 
+  <div class="field">
+    <%= form.label "Hide this project from your home?" %>
+    <%= form.check_box :hidden, checked: event.hidden_at.present? %>
+  </div>
```
thesephist: Can we make this only visible when the user has more than one event under their account? Otherwise it would get super confusing.
Also, what do you think about "Hide this project from Bank home page" instead? (and no question, since none of our other prompts are phrased as questions)

thesephist (changes requested):
yeetsicles

app/controllers/events_controller.rb
tmb: the two lines are for the box being checked/unchecked. will split into if/else

app/controllers/events_controller.rb
tmb: we submit a different set of acceptable params (see bottom of file) depending on whether it is a user or admin logged in

app/controllers/events_controller.rb
tmb: no lol will fix

app/views/events/_form.html.erb
tmb: can do. that wording makes more sense.. will change it

app/controllers/events_controller.rb
tmb: I don't think this makes sense to methodize because it's only being used twice (once in `create`, once in `update`)

thesephist: Sounds good, I'm not too worried here.

PR #408, opened 2019-12-24
Give check delivery window instead of exact date

tmb:
image

Made it between Lob provided date & a date 7 business days away (which I think is a solid margin)


thesephist (approved):
💵

thesephist:
I'm both amazed and a little worried that `business_days` is a builtin lmao looks great thanks theo.

tmb:
it’s not a built in lol! we have a library

PR #409, opened 2019-12-24
Update payout memos

tmb:
Small changes so that invoice & donation payouts are easier to identify. the other change was made in stripe so that the prefix is now "HC" vs. "Hack Club Event" when there's a custom descriptor.

This should close #395


thesephist:
@tmb is there any way to test this without going to prod?

thesephist:
I also think auto-matching code in `ui.coffee` might need to be updated. I can take a look at that.

thesephist (approved):
LGTM, pending any bugs we encounter when this hits production (don't think we can really test how everything integrates until prod, which is kind of bad but it's rare)

thesephist:
wow bold move merging on Christmas day @tmb you're on call

PR #410, opened 2019-12-24
Make FeeReimbursement.event & refractor

tmb:


app/models/fee_reimbursement.rb
```diff
@@ -40,6 +40,14 @@ def status_color
     'error'
   end
 
+  def event
+    if donation
+      return donation.event
+    else
+      invoice.event
```
thesephist: Tiny nit, but can we be consistent here and stick to either using `return` keyword or implicit return for both cases?

app/views/fee_reimbursements/index.html.erb
```diff
@@ -23,10 +23,10 @@
       <tr class="<%= 'shade-yellow' if reimbursement.unprocessed? %>">
         <td><%= "##{reimbursement.id}" %></td>
         <% if reimbursement.invoice %>
-          <td><%= link_to "#{reimbursement.invoice.event.name}", reimbursement.invoice.event, class: 'underline' %></td>
+          <td><%= link_to "#{reimbursement.event.name}", reimbursement.event, class: 'underline' %></td>
           <td><%= link_to "##{reimbursement.invoice.id}", reimbursement.invoice, class: 'underline' %></td>
         <% else %>
-          <td><%= link_to "#{reimbursement.donation.event.name}", reimbursement.donation.event, class: 'underline' %></td>
+          <td><%= link_to "#{reimbursement.event.name}", reimbursement.event, class: 'underline' %></td>
```
thesephist: Can we pull this line out of the if-else case now that it's the same across both cases?

thesephist (changes requested):
Almost there ♨️

app/models/fee_reimbursement.rb
tmb: yes

app/views/fee_reimbursements/index.html.erb
tmb: yes

thesephist (approved):
✌️

PR #418, opened 2019-12-29
Auto-pair all possible transactions

thesephist:
Close #364

This PR introduces `Transaction.try_pair_automatically!`, which tries to
use some smarts to automatically pair all transactions that occur in the normal
operations of Hack Club Bank (i.e. not paper checks received, not Bill.com
invoice payments, but everything else).

It works as tested locally, and should not break existing functionality in prod.

If it breaks in prod, a SyncTransactions job should roll back the transaction, so
there's minimized risk, and we'll hear about it from Bugsnag.

I think in the future, we ideally want to split up the `potential_????` and
`try_pair_?????` methods to their respective classes and register it from there
instead of throwing it all in the `Transaction` model, but this is an MVP.

@tmb I think this is all good but would appreciate a sanity check from you. The only
untested algorithms are for pairing invoices and donations, because our Tx memos
changed recently, but I'm fairly confident they both work.


app/models/transaction.rb
```diff
@@ -173,6 +146,226 @@ def ensure_paired_correctly
...
+  # and meant to aid human work.
+  #
+  # They are used in try_pair_automatically! and rendering Transaction#edit
+  #
+  # TODO: eventually, we want to move these concerns out to their
+  # corresponding models. i.e. potential_donation_payout?
+  # should probably be a part of the Donation class, not Transaction.
+
+  def potential_invoice_payout?
+    self.name.start_with?('HACK CLUB EVENT Payout ') && amount > 0
```
tmb: these need to be updated with the prefix `HC` vs `HACK CLUB EVENT` (change was made by me for donations)

app/models/transaction.rb
```diff
@@ -173,6 +146,226 @@ def ensure_paired_correctly
...
+  end
+
+  def potential_donation_payout?
+    self.name.start_with?('HACK CLUB EVENT DONATE ') && amount > 0
```
tmb: need to be updated with the prefix `HC` vs `HACK CLUB EVENT

app/models/transaction.rb
```diff
@@ -173,6 +146,226 @@ def ensure_paired_correctly
...
+      try_pair_github
+    elsif potential_check?
+      try_pair_check
+    end
+    # NOTE: we cannot curently auto-pair Expensify txs
+  end
+
+  def try_pair_invoice
+    InvoicePayout.lacking_transaction.each do |payout|
+      if self.name.include? payout.statement_descriptor
```
tmb: I don't think this will work because sometimes the statement descriptor with the sponsors name gets cut off. not totally sure

app/models/transaction.rb
```diff
@@ -173,6 +146,226 @@ def ensure_paired_correctly
...
+
+    self.fee_relationship = FeeRelationship.new(
+      event_id: event.id,
+      is_fee_payment: true,
+    )
+
+    self.save
+  end
+
+  def try_pair_emburse
```
tmb: general comment but i feel like each of these should be running the `potential_xxx` methods before trying anything else... again not sure

app/models/transaction.rb
thesephist: From checking data on prod, we truncate the payout string before saving it as statement_descriptor so I think it should work. We'll try it, since it shouldn't break even if it doesn't work.

PR #421, opened 2020-01-01
Pretty donations in transaction view

tmb:


tmb:
Second commit is because `calculate_fee_amount` wasn't being run when fee reimbursements were manually matched, leading to some really weird math being done while trying to calculate the fee.

app/views/donations/_donation.html.erb
```diff
@@ -0,0 +1,75 @@
+<% transaction = donation.payout.t_transaction %>
```
thesephist: The indentation on this template is kind of fucked can we fix it

app/controllers/transactions_controller.rb
```diff
@@ -98,6 +99,14 @@ def update
         should_delete_fee_relationship = true if fee_relationship&.persisted?
       end
 
+      if current_fee_reimbursement.nil? && @transaction.fee_reimbursement.present?
+        @transaction.fee_relationship = FeeRelationship.new(
+          event_id: @transaction.fee_reimbursement.event.id,
+          fee_applies: true,
+          fee_amount: @transaction.fee_reimbursement.calculate_fee_amount
+        )
+      end
```
thesephist: It seems weird that this was missing before but it worked for invoices -- if it's being handled here now for all transactions, should we remove the code that was only handling it for invoices before, so we only have one place in the code that concerns this?

thesephist (changes requested):
need a lil fix 📈

app/controllers/transactions_controller.rb
tmb: Invoices don't have the weird under $1 fee stuff—that's what thsi fixes

PR #423, opened 2020-01-02
Deactivate Emburse cards from within Bank

tmb:
Close #380


app/models/card.rb
```diff
@@ -42,6 +44,16 @@ def department_id
     emburse_obj&.department&.id
   end
 
+  def deactivate!
+    self.deactivated_at = DateTime.now
```
lachlanjc:
```suggestion
    update(deactivated_at: DateTime.now)
```
Then you can remove the next line :) same for next method
Also see: update_attribute

tmb: it's more LOC but frankly I think the way it is right now is better. easier to read.

tmb:
@thesephist it's finished now except for that notice that organizer should contact us for lost/stolen.. can you add that?

app/models/card.rb
thesephist: I'm ambivalent here, one might be more rails-y but I honestly think both are fine to read.

thesephist:
@tmb a reminder that if there's any UI changes please share screenshots in the PR

tmb:
oops seems like I forgot to push last night before I commented lol

lachlanjc:
“Deactivated” is one word btw

app/models/card.rb
```diff
@@ -55,6 +88,21 @@ def emburse_id_format
     end
   end
 
+  def sync_with_emburse
+    if self.deactivated_at_changed?
```
thesephist: Pro tip: when I write out object properties that denote some binary state (like "has the "deactivated_at" of this card changed?), to remove ambiguity, I tend to write these properties in the form of `is/was[something something]` or `has[something something]`. 

In this case, `self.was_deactivated_at_changed?` is clearer and leaves less room for confusion for someone else reading the code than simply `self.deactivated_at_changed?` -- the latter could also mean the changed-at time of the property, or the changed value of deactivated_at to someone unfamiliar with the codebase.

app/models/card.rb
```diff
@@ -55,6 +88,21 @@ def emburse_id_format
...
+      if emburse_obj[:state] != 'active' && emburse_obj[:state] != 'suspended'
+        errors.add(:card, 'cannot be suspended until it is activated')
+        return
+      end
+
+      if !self.deactivated_at.nil?
```
thesephist: When I have an if-else case, I usually flip it so that the positive case (in this case `self.deactivated_at.nil?`) comes first, which I find a little easier to read than if the negative case comes first in an if-else.

It's the difference between: "if the deactivated_at is not null, do X, else do Y" and "if the deactivated_at is null, do X, else do Y"

thesephist:
Screen Shot 2020-01-14 at 22 37 49
added this

db/schema.rb
```diff
@@ -29,6 +29,7 @@
     t.datetime "updated_at", null: false
     t.string "recipient_tel"
     t.datetime "rejected_at"
+    t.datetime "scheduled_arrival_date"
```
thesephist: Let's try to avoid pushing PRs when your schemas are dirty from other migrations.

thesephist (approved):
LGTM. Left one pointer that might be helpful perspective.

PR #424, opened 2020-01-04
Button to export CSV of pending CRs

tmb:
Closes #376 

cc @seanvictory


thesephist (approved):
🚀 LGTM if works in testing!

PR #425, opened 2020-01-06
Add "payment for" to ACH transfers & checks

tmb:


thesephist:
screenshot pls <3 @tmb 


also side note if a PR is work in progress feel free to add WIP in the title so I don't prematurely review things

thesephist:
Wait actually, I was stupid. I feel like for checks, the memo field already solves this problem and for 99% of cases the check memo is going to be the same thing as payment-for. 

So I actually think we only needed this for transfers, sorry. Thought @tmb ?

thesephist:
Two things need to happen:

1) re-migrate from a production backup so the schema is correct.
2) reorder migration so it runs properly when deployed.

app/controllers/ach_transfers_controller.rb
```diff
@@ -90,7 +90,7 @@ def set_event
   end
 
   def ach_transfer_params
-    params.require(:ach_transfer).permit(:routing_number, :account_number, :bank_name, :recipient_name, :recipient_tel, :amount)
+    params.require(:ach_transfer).permit(:routing_number, :account_number, :bank_name, :recipient_name, :recipient_tel, :amount, :payment_for)
```
thesephist: We should probably spread this out vertically at this point, but NBD.

app/views/checks/_edit_form.html.erb
```diff
@@ -54,6 +54,11 @@
     </div>
   </div>
 
+  <div class="field">
+    <%= form.label :payment_for %>
+    <%= form.text_field :payment_for, placeholder: 'Meal from Subway', required: true %>
+  </div>
+
```
thesephist: Do you know if the checks edit form is _ever_ used? If not, we should just remove it from the codebase, right?

AFAIK checks are never modified after they are created, and the `_new.html.erb` template has its own fields and doens't render the `_edit_form.html.erb` form.

thesephist (approved):
Yeet

PR #432, opened 2020-01-12
Donations must be greater than or equal to $1.00

tmb:


app/helpers/application_helper.rb
```diff
@@ -40,7 +40,7 @@ def pop_icon_to(icon, url, options = {})
     end
   end
 
-  def form_errors(model, name = nil)
+  def form_errors(model, name = nil, prefix = "We couldn't save this")
```
thesephist: This abstraction / generalization isn't really necessary imo and probably will add unnecessary edge cases, but I don't feel too strongly about it yet.

thesephist (approved):
🏦

PR #433, opened 2020-01-12
Select state from list in card request

tmb:
I feel like there's supposed to be a better way to do this but idk


app/views/card_requests/_form.html.erb
```diff
@@ -58,7 +58,7 @@
   <div class="flex">
     <div class="field">
       <%= form.label :shipping_address_state, 'State / Region' %>
-      <%= form.text_field :shipping_address_state, required: true, placeholder: "DC" %>
+      <%= form.select :shipping_address_state, options_for_select(['AL','AK','AZ','AR','CA','CO','CT','DE','DC','FL','GA','HI','ID','IL','IN','IA','KS','KY','LA','ME','MD','MA','MI','MN','MS','MO','MT','NE','NV','NH','NJ','NM','NY','NC','ND','OH','OK','OR','PA','RI','SC','SD','TN','TX','UT','VT','VA','WA','WV','WI','WY']), prompt: "Select state", required: true %>
```
lachlanjc:
```suggestion
      <%= form.select :shipping_address_state,
        options_for_select(['AL','AK','AZ','AR','CA','CO','CT','DE','DC','FL','GA','HI','ID','IL','IN','IA','KS','KY','LA','ME','MD','MA','MI','MN','MS','MO','MT','NE','NV','NH','NJ','NM','NY','NC','ND','OH','OK','OR','PA','RI','SC','SD','TN','TX','UT','VT','VA','WA','WV','WI','WY']),
        prompt: 'Select state',
        required: true %>
```

lachlanjc:
> I feel like there's supposed to be a better way to do this but idk

Unfortunately in Rails there's not, since Rails is region-agnostic. The only choices besides this are sticking it in a helper or downloading a gem that does essentially the same thing. Should definitely be in a helper if we use the dropdown on any other form, but other than that, this is it.

app/views/card_requests/_form.html.erb
thesephist: I don't see this quote thing as something to nitpick over, tbh. I'm ok with either. We don't even have a linter running, so don't think it matters. If we wanted to enforce it it should be software doing the linting, not humans wasting code review time.

It _would_ be nice to spread out the list of states vertically instead of having one big huge line that stretches to the right forever and ever.

thesephist (dismissed):
LGTM except for the comment about spreading the list out vertically thing I mentioned earlier, if possible. Maybe it's not a thing that works in irb templates easily, idk.

app/views/card_requests/_form.html.erb
lachlanjc: The vertical list might overwhelm the file, even though obviously the scrolling isn't great either. I just changed it to a maybe-better in-between, so the attributes after the list are accessible even if the list is long.

(With the quotes, I know it's mixed across the repository, but even without a linter, it just seems messy to use different quote styles for the same purpose on the same line.)

PR #436, opened 2020-01-14
No longer throw error in webhook when payment intent isn't invoice

tmb:


thesephist (approved):
LGTM. We should test this in production with Stripe dev mode to make sure the data persists through the Stripe flow.

tmb:
I tested in dev mode locally—going to merge & we'll see if all is well in prod

PR #445, opened 2020-01-16
Copy over TX details when pending TX go from pending to complete

thesephist:
Close #443


thesephist:
Hey @tmb, would appreciate a second pair of eyes on this 🙏

tmb:
I don't see any immediate problems with it-- looks good to merge.

PR #453, opened 2020-01-18
Fix heading centering & small design improvements

lachlanjc:
- Fix centering of event name, closes #449 
  - Removing the ID badge from the header for simplicity—does anyone ever use that? (Still available on /events if unexpectedly needed)
- Super minor aesthetic improvements to dock + menu
- Fixes gap in POC text on event page

Before:
Screen Shot 2020-01-17 at 11 49 55 PM

After:
Screen Shot 2020-01-17 at 11 57 01 PM


thesephist:
We still use the admin-only event ID number -- it's useful for admin tasks where we need to reference event ID without having to do additional or clicks.

I agree with moving team + event settings to be always there in the three-dot dropdown, that's a good move 👍

thesephist (approved):
🌈

tmb:
Merging bc approved

PR #462, opened 2020-01-19
Small Bugs

tmb:
Closes #457, #451, #454


app/models/transaction.rb
```diff
@@ -41,11 +41,12 @@ class Transaction < ApplicationRecord
   after_initialize :default_values
 
   def self.total_volume
-    self.sum('@amount').to_i
+    self.includes(:fee_relationship).where.not(fee_relationships: { event_id: 183 }).sum('@amount').to_i
```
thesephist: Let's add a comment here to explain why we're using this number 183. Otherwise it's just a magic number that someone else reading the code in the future will get super confused by.

ORRrrrr

Another way to write code here that documents itself / the intent here correctly would be to do something like this:

```ruby
hq_event_id = 183
transactions_not_for_hq = self.includes(:fee_relationship).where.not(fee_relationships: { event_id: hq_event_id })
transactions_not_for_hq.sum('@amount').to_i
```

This way, we express exactly what we intend to do here without any comments, and there's no risk of future readers getting confused.

app/models/transaction.rb
```diff
@@ -41,11 +41,12 @@ class Transaction < ApplicationRecord
...
   end
 
   def self.during(start_time, end_time)
-    self.where(["transactions.created_at > ? and transactions.created_at < ?", start_time, end_time])
+    self.includes(:fee_relationship).where.not(fee_relationships: { event_id: 183 })
+      .where(["transactions.created_at > ? and transactions.created_at < ?", start_time, end_time])
```
thesephist: Same point applies as above. I think we can just define a variable here or a global constant that's `HQ_EVENT_ID` and reference that instead of a hard-coded in number that can be cryptic later.

app/views/card_requests/_form.html.erb
```diff
@@ -59,8 +59,9 @@
     <div class="field w-50">
       <%= form.label :shipping_address_state, 'State / Region' %>
       <%= form.select :shipping_address_state, options_for_select(['AL','AK','AZ','AR','CA','CO','CT','DE','DC','FL','GA','HI','ID','IL','IN','IA','KS','KY','LA','ME','MD','MA','MI','MN','MS','MO','MT','NE','NV','NH','NJ','NM','NY','NC','ND','OH','OK','OR','PA','RI','SC','SD','TN','TX','UT','VT','VA','WA','WV','WI','WY']),
-        prompt: "Select state", 
-        required: true %>
+        { prompt: "Select state" },
+        { required: true }
```
thesephist: Can you shed some light on why this works this way? I'm a little confused by why this now works but didn't before.

thesephist (changes requested):
Couple minor requests, but otherwise lit.

app/views/card_requests/_form.html.erb
lachlanjc: Check the HTML, but I don’t think this will actually solve the issue. You need “Select state” to have `disabled`, or it’s a valid option to submit the form with, so even if `required` is rendered it is circumvented by the default option.

tmb: It seems to work fine as is.. prompts the user to select a state if "Select state" is selected.

PR #474, opened 2020-01-21
Small GSuite Fixes

tmb:
No UI changes

Closes #469


app/views/g_suite_accounts/_panel.html.erb
```diff
@@ -21,10 +21,12 @@
...
           </p>
-          <%= link_to 'Reset password', g_suite_account_reset_password_path(account.id), method: :put,
-              data: { confirm: "Reset the account password for #{account.address}? We'll email a new password to the backup email." },
-              class: 'h5 muted underline'
-            %>
+          <% if account.accepted? %>
+            <%= link_to 'Reset password', g_suite_account_reset_password_path(account.id), method: :put,
+                data: { confirm: "Reset the account password for #{account.address}? We'll email a new password to the backup email." },
+                class: 'h5 muted underline'
+              %>
```
lachlanjc: In Rails, typically would put the if statement directly after the link_to, before the closing %>, to avoid wrapping

thesephist: I think that's normally more readable, but this tag is super long so having the if tacked on at the end might make it easy to miss it, no?

Having it be a separate tag thing above I think makes it easier to spot in this case.

thesephist (approved):
lgtm

PR #483, opened 2020-01-28
Allow users to suspend GSuite accounts

tmb:


app/views/g_suite_accounts/_panel.html.erb
```diff
@@ -25,6 +28,9 @@
             <%= link_to 'Reset password', g_suite_account_reset_password_path(account.id), method: :put,
                 data: { confirm: "Reset the account password for #{account.address}? We'll email a new password to the backup email." },
                 class: 'h5 muted underline'
+              %> | <%= link_to "#{account.suspended? ? 'Un-suspend' : 'Suspend'}", g_suite_account_toggle_suspension_path(account.id), method: :put,
```
lachlanjc: No need for interpolation if that's the only string :)
```suggestion
              %>
            <%= link_to (account.suspended? ? 'Un-suspend' : 'Suspend'),
                g_suite_account_toggle_suspension_path(account.id),
                method: :put,
```

lachlanjc:
Quick design thoughts:

- "Suspend" is a short word & an easy tap. Even though there's a confirmation dialog, the design should indicate this is a destructive action with potentially terrible consequences for your event if you do it mistakenly.
- We don't use "|" as a divider anywhere else in Bank, so would avoid (fragmentation of UI patterns)
- Thoughts on adding a … popup here, like on the events nav? Would have Reset password, Suspend user, Delete (admin-only). Would introduce an extra step before any of these destructive actions, perhaps usefully? & emphasize they're not the standard actions you'll need.

app/controllers/g_suite_accounts_controller.rb
```diff
@@ -100,11 +100,28 @@ def reset_password
...
+    end
   end
 
+  def toggle_suspension
+    authorize @g_suite_account
+
+    @event = @g_suite_account.g_suite.event
+
+    if @g_suite_account.toggle_suspension!
+      flash[:success] = "#{@g_suite_account.address} has been successfully #{@g_suite_account.suspended? ? 'suspended' : 'un-suspended'}."
```
thesephist: Can we use "re-activated" instead of "un-suspended"? double negatives are confusing.

app/views/g_suite_accounts/_panel.html.erb
```diff
@@ -26,6 +29,12 @@
                 data: { confirm: "Reset the account password for #{account.address}? We'll email a new password to the backup email." },
                 class: 'h5 muted underline'
               %>
+            <%= link_to (account.suspended? ? 'Un-suspend' : 'Suspend'),
+                g_suite_account_toggle_suspension_path(account.id),
+                method: :put,
+                data: { confirm: "Do you really want to #{account.suspended? ? 'un-suspend' : 'suspend'} #{account.address}?"},
```
thesephist: Likewise, let's use "re-activate" instead of suspend.

thesephist (dismissed):
wording changes, but otherwise LGTM.

Re: @lachlanjc 's suggestion of having a drop-down menu, I like it, but (1) people won't discover it as easily, and (2) it takes time to build, and I think the current alternative is fine, not beautiful, but better than delaying other improvements imo.

lachlanjc:
Yeah that totally makes sense. Making that control less discoverable is exactly the point btw—I don’t think such a destructive action should be so exposed.

PR #489, opened 2020-02-01
Add donations to incoming deposits

tmb:
image

Donations aren't clickable in this dialog while invoices are.. I don't think this is super important to fix but definitely something to think about


thesephist (approved):
🎉✅

PR #492, opened 2020-02-02
Donations summary page for events, add donation message, first donation mailer

tmb:
image

Same as before, not totally sure where you want the `donation.message` field to go so I left it out


lachlanjc:
If you’re expecting them to be somewhat long sometimes you could put them in a little popover or modal

app/controllers/events_controller.rb
```diff
@@ -144,6 +144,12 @@ def g_suite_overview
     @g_suite_status = @event.g_suite_status
   end
 
+  def donation_overview
+    @event = Event.find(params[:event_id])
+    authorize @event
+    @donations = @event.donations.where(status: 'succeeded').sort_by {|d| d.created_at }
```
thesephist: I think it makes more sense for this list to be reverse-chronological, so it matches the transactions list.

app/mailers/donation_mailer.rb
```diff
@@ -4,4 +4,11 @@ def donor_receipt
 
     mail to: @donation.email, subject: "Receipt for your donation to #{@donation.event.name}"
   end
+
+  def first_donation_notification
+    @donation = params[:donation]
+    @emails = @donation.event.users.map { |u| u.email }
+
+    mail to: @emails, subject: "Congrats! 🎉 You've received a donation from #{@donation.name}"
```
thesephist: This seems a bit long and might not show on phone screens in one line.. how about

>Congrats on receiving your first donation! 🎉

thesephist: Specifying "first" may also clear up confusion, so people are less likely to get the impression that they'll get notified for every donation.

app/models/donation.rb
```diff
@@ -38,6 +40,24 @@ def queue_payout!
...
   end
 
+  def status_text
+    if status == 'succeeded' && self&.payout&.t_transaction.present?
+      return 'Deposited'
+    elsif status == 'succeeded' && payout_id == nil
+      return 'In transit'
+    end
+
+    'Unknown'
```
thesephist: Might we want to raise an exception here so we (devs) are notified? This should never happen in practice, right? So it's better to fail loudly here since it probably means we messed up some code logic somewhere and need to fix it.

app/views/donation_mailer/first_donation_notification.html.erb
```diff
@@ -0,0 +1,14 @@
+<p><%= @emails.length > 1 ? 'Hi all,' : 'Hi there,' %></p>
+
+<p>
+  Congrats! We're emailing to let you know that you've received your first ever donation.
+</p>
+
+<p>
+  You can <%= link_to 'click here', event_donation_overview_path(@donation.event.slug) %> to see the full details, including when you'll receive the payment to be able to spend it.
```
thesephist: Might be nice to include the first donation's donor and amount, so it's not super generic. Maybe like:

>Congrats! We're emailing to let you know that you've received your first-ever donation of $XX from $XX.

app/views/events/donation_overview.html.erb
```diff
@@ -0,0 +1,51 @@
...
+      aria-label="Filter invoices"
+      class="flex-auto md-mr2"
+      style="width: auto; max-width: none;">
+    <div data-behavior="filterbar" role="tablist" class="filterbar__filters flex mt1 lg-mt0">
+      <%= filterbar_item 'All', :exists, true %>
+      <%= filterbar_item 'Deposited', :deposited %>
+      <%= filterbar_item 'In transit', :in_transit %>
+    </div>
+  </div>
+	<% @donations.each do |d| %>
```
thesephist: I think this UI needs work but we can handle that once we merge to master.

app/views/events/_nav.html.erb
```diff
@@ -33,6 +33,7 @@
       <%= link_to 'Promotions & perks', event_promotions_path(event_id: @event.slug) %>
       <%= link_to 'Make reimbursements', event_reimbursements_path(event_id: @event.slug) %>
       <%= link_to 'Documents', event_documents_path(event_id: @event.slug) %>
+      <%= link_to 'Donations', event_donation_overview_path(event_id: @event.slug) %>
```
thesephist: Can we put this behind the `donations_page_enabled` flag?

db/schema.rb
```diff
@@ -274,6 +275,8 @@
     t.text "club_airtable_id"
     t.boolean "beta_features_enabled"
     t.datetime "hidden_at"
+    t.boolean "donation_page_enabled"
+    t.text "donation_page_message"
```
thesephist: Is it just me or does this PR not include a migration to add these two columns?

thesephist (changes requested):
TLDR almost ready to merge, we can polish UI stuff once it's in master.

app/models/donation.rb
tmb: If we raise an exception the whole view will break (bad)—I'm leaving Unknown there with the hope that people will hit us up.

db/schema.rb
tmb: Another one of the donations one does that we already merged.

PR #496, opened 2020-02-09
Add new static messages

seanvictory:
Fun, new static messages to be displayed while in development mode.

Includes slogans such as "Open on weekends" and "Open on holidays" to contrast how most banks do not work during those times, but Hack Club Bank does!


thesephist (approved):
🎉

PR #500, opened 2020-02-18
Add search for users based on email name or number

tmb:
image


.nova/Configuration.json
```diff
@@ -0,0 +1,4 @@
+{
+  "workspace.preview_type" : "custom",
+  "workspace.preview_url" : "localhost:3000"
```
thesephist: Can you remove this from the PR? I know you already ignored it in gitignore, but that may have come too late.

app/views/static_pages/index.html.erb
```diff
@@ -90,6 +90,7 @@
       <%= card_to 'G Suites', g_suites_path %>
       <%= card_to 'G Suite Accounts', g_suite_accounts_path, { badge: @active[:g_suite_accounts] } %>
       <%= card_to 'Negative Events', negative_events_path, { badge: @active[:negative_events] } %>
+      <%= card_to 'Search', admin_search_path %>
```
thesephist: Can we use some more descriptive name like "search user"

app/models/user.rb
```diff
@@ -86,4 +87,8 @@ def profile_picture_format
     profile_picture.purge_later
     errors.add(:profile_picture, 'needs to be an image')
   end
+
+  def format_number
+    self.phone_number = Phonelib.parse(self.phone_number).sanitized
```
thesephist: Does this ever raise an exception? What happens if that happens when the user's trying to first set up their account on Bank?

app/views/static_pages/search.html.erb
```diff
@@ -0,0 +1,25 @@
...
+<%= form_with(url: admin_search_path, local: true) do %>
+  <%= label_tag(:q, "Search for:") %>
+  <%= text_field_tag(:q) %>
+  <%= submit_tag("Search") %>
+<% end %>
+<p class="mt0">you can use name, phone number, or email</p>
+
+<% if @result.present? %>
+	<% @result.each do |r| %>
+		<% next if r == nil %>
```
lachlanjc: Filtering array (I believe it’s .compact? Not positive) might be more elegant

app/views/static_pages/search.html.erb
```diff
@@ -0,0 +1,25 @@
...
+		<p><%= r.name %></p>
+		<% r.events.each do |e| %>
+			<a href="<%= e.slug %>"><%= e.name %></a>
+		<% end %>
+		<p><%= r.email %></p>
+		<p><%= r.phone_number %></p>
+		<p>------</p>
```
lachlanjc: `<hr>` or a .list :)

lachlanjc (commented):
👍

app/views/static_pages/search.html.erb
thesephist: Yeah an `<hr>` is probably more useful here.

app/models/user.rb
tmb: afaik this runs after the validation that confirms that it indeed is a phone number

PR #505, opened 2020-02-19
Version lock postgres docker image

maxwofford:
This fixes an issue where freshly installed repos (on a new system w/o existing docker images) pulls the latest image of postgres, which is incompatible with Bank


thesephist (approved):
🔥

thesephist:
Glad the version fix was the root cause and you fixed the issue!

PR #531, opened 2020-03-03
Many small fixes

tmb:
Closes #526 
Closes #523 
Closes #522 
Closes #519 

image


app/views/events/team.html.erb
```diff
@@ -14,7 +14,7 @@
       <p class="flex-auto">
         <strong class="h3 block"><%= position.user.full_name || 'Unnamed' %></strong>
         <%= position.user.email %>
-        <span class="muted h5 block">(added <span title="<%= position.created_at %>"><%= time_ago_in_words position.created_at %> ago</span>)</span>
+        <span class="muted h5 block">(added <span title="<%= position.created_at %>"><%= time_ago_in_words position.created_at %> ago</span> by <%= position.organizer_position_invite.sender.full_name %>)</span>
```
lachlanjc: Any reason not to use a user mention?

app/views/g_suites/index.html.erb
```diff
@@ -21,9 +21,9 @@
         <td><%= link_to g_suite.event.name, g_suite.event %></td>
         <td><%= g_suite.verification_key %></td>
         <td><%= g_suite.deleted_at %></td>
-        <td><%= link_to 'Show', g_suite_path(g_suite, event_id: g_suite.event.id) %></td>
-        <td><%= link_to 'Edit', edit_event_g_suite_path(g_suite, event_id: g_suite.event.id) %></td>
-        <td><%= link_to 'Destroy', g_suite_path(g_suite, event_id: g_suite.event.id), method: :delete, data: { confirm: 'Are you sure?' } %></td>
+        <td><%= link_to 'Show', g_suite_path(g_suite, event_id: g_suite.event.slug) %></td>
```
lachlanjc: Can’t you do `link_to ‘Show’, g_suite` here? I might be missing something but you shouldn’t need to spell everything out if you’re using regular actions

app/views/events/team.html.erb
thesephist: If I'm thinking of the right component, I think user mention would look huge and out of place here -- this is supposed to be pretty subtle, and already next to the avatar of the invited person -- having the inviter also show as an avatar might be confusing.

app/views/g_suites/index.html.erb
thesephist: +1, ditto for the rows below. I'm also not 100% sure that this is the case.

thesephist (approved):
Couple of questions in comments, but overall looks great.

My only qualm here is that now we're forcing every money input to be a `text_field` that can't be parsed as numerical input, and we lose all the nice numerical input things like having a number keyboard on mobile. We also have to be careful to parse those inputs with `gsub(',', '')` in the backend. All this ceremony for... not something that is critical. I think it sort of makes the codebase dirty. But I don't feel too strongly about it since it's very explicit, and doesn't affect inputs that aren't explicitly opted in.

I think long term we need some abstraction over money inputs that addresses these problems.

app/views/events/team.html.erb
lachlanjc: Yeah, had similar reservations. Just as a general thing if we’re breaking UI conventions (team members are always shown as mentions), we should have an explicit reason.

thesephist: I don't think we're consistent enough that it matters -- we use just the name without avatar or lightning as user reference in other places imo.

app/views/g_suites/index.html.erb
tmb: It makes you provide the ID of the event because the path is /event/g_suite afaik

PR #537, opened 2020-03-06
Things for Clubs: virtual cards, grant-only accounts, and Disbursements

thesephist:
## Things that are done (MVP-stage, requires some UI polish but works)

- Grant-only accounts
- Virtual card support via Emburse virtual cards
- Disbursements, for scheduling giving money to events

## Things that are not done

- APIs to automatically create events and disbursements - @tmb 

## Shots of screens and such:

What a virtual card looks like
Screen Shot 2020-03-06 at 21 45 16

Virtual card more things
Screen Shot 2020-03-06 at 21 46 29

Spend only accounts see this:
Screen Shot 2020-03-06 at 21 46 33

Full fiscally sponsored accounts see this:
Screen Shot 2020-03-06 at 21 46 38

Spend-only accounts see this
Screen Shot 2020-03-06 at 21 46 46

When you order a card you currently check this box, but before it #ships officially the UI will be improved so it isn't this ugly
Screen Shot 2020-03-06 at 21 47 09

Disbursements:

Screen Shot 2020-03-07 at 01 43 44
Screen Shot 2020-03-07 at 01 43 56
Screen Shot 2020-03-07 at 01 44 03
Screen Shot 2020-03-07 at 01 44 09

\* Disbursements are currently not viewable by non-admin users. We can probably change that, but there are UX related unknowns here, and this is an MVP.


thesephist:
Hey @tmb  - as mentioned in imessage this covers everything Chris asked for in our call except the actual API itself, which is specced out in the `README.md` of this branch. I'd prefer to merge this as soon as Sean's flow can handle virtual cards, and then push the API out when you're done with that part.

This branch is ready to be merged from my POV. But I'd appreciate a look-over from you to make sure I'm not gonna burn the house down.

app/controllers/api_controller.rb
```diff
@@ -0,0 +1,145 @@
...
+  before_action :check_token
+
+  # find an event by slug
+  def event_find
+    # pull slug out from JSON
+    slug = params[:slug]
+
+    e = Event.find(slug)
+
+    # event not found
```
thesephist: Doesn't `Event.find` raise an exception when it doesn't exist? I don't think it returns `nil`.

I believe the way to do this is to use `Event.where(slug: slug)` and then check if the `result.length > 0`

app/controllers/api_controller.rb
```diff
@@ -0,0 +1,145 @@
...
+        address: 'N/A',
+        sponsorship_fee: 0.07,
+        expected_budget: 100.0,
+        has_fiscal_sponsorship_document: false,
+        point_of_contact_id: point_of_contact_id,
+        is_spend_only: true,
+    )
+
+    if !e.save
+      render json: { error: "Event couldn't be created!" }, status: 500
```
thesephist: Could we stringify the value of `e.errors` (I think that's the right variable) (what error occurred) and include it in the response? Would make debugging easier.

app/controllers/api_controller.rb
```diff
@@ -0,0 +1,145 @@
...
+      return
+    end
+
+    render json: {
```
thesephist: This method doesn't currently invite people from `organizer_emails`. Did I miss something?

thesephist (commented):
couple places where I think things break / don't work, but by and large this looks pretty good 👍

couple small fixes and we're gucci to merge this bad boi

thesephist:
Also @tmb heroku found an issue :/

Screen Shot 2020-03-08 at 11 29 08

PR #540, opened 2020-03-10
Make comments editable

thesephist:
Fix #536

Attachments currently cannot be edited once they're attached. This can be fixed later with an update.

Screen Shot 2020-03-10 at 00 34 40

Screen Shot 2020-03-10 at 00 34 35


app/views/comments/_item.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="mt2 mb2">
+  <%= creator_bar comment %>
+  <div class="card">
+    <%== MarkdownService.instance.renderer.render comment.content %>
```
tmb: why is there an extra `=` here?

app/controllers/comments_controller.rb
```diff
@@ -1,5 +1,6 @@
 class CommentsController < ApplicationController
-  before_action :set_commentable
+  before_action :set_comment, only: [:edit, :update]
+  before_action :set_commentable, except: [:edit, :update]
```
tmb: shouldn't `set_commentable` still apply for new?

tmb (approved):
looks good but two things i am wondering

app/controllers/comments_controller.rb
thesephist: Yeah isn't that what's happening here? It's only exempted for edit and update. Right?

app/views/comments/_item.html.erb
thesephist: Not sure. This was present in the original template so I just kept it? It seemed to work, however, as expected. I don't know what this notation is and am having trouble googling for it but maybe it's a typo.

app/controllers/comments_controller.rb
thesephist: It turns out the `<%==` means "render this string as literal HTML". So whatever markdown service spits out, it goes in the HTML verbatim, which is what we want. Otherwise the content gets escaped.

app/views/comments/_item.html.erb
lachlanjc: It uses the `raw` helper: https://stackoverflow.com/questions/13142326/what-does-do-in-rails-erb/13142764

PR #543, opened 2020-03-10
Added a slogan

seanvictory:
Added a slogan for my younger brother to show him I can make a change to a system.


thesephist:
waow our kpis are thru the roof with this change

PR #555, opened 2020-03-21
Cease storing card number + CVV for virtual cards

tmb:
Only visual change is that we now show a link to the cardholder instead of the number


thesephist:
LGTM on a first pass. I'm gonna double check this later tonight before I merge, thx.

thesephist (approved):
🦠 wash your hands baby 🦠

PR #566, opened 2020-04-06
API token can be used for multiple events

maxwofford:


maxwofford:
For posterity, this is just a hack the Frank.ly team can use to test out multiple auth keys while I build out a token generator

PR #570, opened 2020-04-08
Prefill AirTable Bug-Fix

seanvictory:
Fixes bug where Promotions page would not load due to an error in the AirTable form URL.


maxwofford:
Just tested this on the preview deployment & it worked.

maxwofford (approved):

PR #559, opened 2020-03-27
Virtual cards are visually distinct

maxwofford:
Reverts hackclub/bank#554

PR #574, opened 2020-04-16
Add et.during method

maxwofford:
This was deleted in #559 because I didn't realize it was used in Emburse Transaction syncing


maxwofford:
merging with approval from @tmb over slack

PR #581, opened 2020-04-20
Change destroy button to red background

seanvictory:
Closes #580 

Updates the destroy button on the common documents section to have a red background with gray text. 

image


lachlanjc (approved):
This looks totally fine, honestly it’s been awhile since I was focusing on the Bank UI a ton though. Let’s go ahead & we can fix anything later if this causes any other issues, but we shouldn’t be using this same class anywhere that’s not a delete/cancel/remove button.

PR #587, opened 2020-04-29
Create emburse department ID on create new event

seanvictory:
Create the Emburse Department ID for an event when creating a new event on Hack Club Bank.


db/schema.rb
```diff
@@ -276,6 +276,7 @@
 
   create_table "events", force: :cascade do |t|
     t.text "name"
+    t.bigint "event_id"
```
lachlanjc: Hmm, I don’t think this should be here?

maxwofford:
Closing this for the time being– might reopen if I work on this feature in the future

PR #595, opened 2020-05-07
Ops streaks

maxwofford:


app/views/static_pages/index.html.erb
```diff
@@ -9,6 +9,10 @@
     <span class="badge h5 caps"><%= random_nickname() %></span>
   </h1>
 
+  <% admin_tools('w-100 mt3') do %>
+    <%= OpsCheckin.streak %> days since last incident
```
lachlanjc: It’d be rad to make the numbers look like a flipboard like this https://culturedcode.com/status/

maxwofford: oooooooohhhhh, yes!

PR #615, opened 2020-05-15
Initial implementation of universal ID

maxwofford:


lachlanjc:
What is this/why are we implementing it?

maxwofford:
@lachlanjc I was playing around with this back when we did transparency mode. A bunch of our routes depend on private info (ex. how certain transactions have the bank account number in the URL). This would implement a universal ID that we could use for everything by default

motdotla:
@maxwofford and @lachlanjc you might find Rails' Global IDs interesting for this use case as well. They are sort of a hidden secret built into Rails.

https://mikerogers.io/2019/08/03/using-signed-global-id-in-rails.html

lachlanjc:
Are we still moving forward on this?

motdotla:
I'd personally like to table this for now and close the PR. I think we can better address after the transaction engine.

Thoughts @maxwofford  ?

maxwofford:
Yep! Closing

ezgif com-gif-maker

PR #632, opened 2020-05-20
Remove check auto-voiding & language around it

tmb:
Changes:
- Removed the job that autovoids checks
- Removed the call to schedule the autovoiding job from the Check model
- Removed the language telling the user about autovoiding from the send page.


maxwofford:
@tmb wish we had talked about the specific implementation for this in a proposal first. The flow that this PR allows that I'd like to prevent is:

- user issues a check
- months pass & user forgets to void
- bad actor deposits the check

- user issues a check
- months pass & user forgets
- check money is left floating in our positive pay account, making auditing/transitioning to a new positive pay account more complex

I'd like to have a solution that covers those.

tmb:
@maxwofford what if we proceeded forward with what I currently have & we send an email to the user giving them a heads up after 3 months "hey! your check to x for $y hasn't been deposited yet—please make sure to void it or ask the recipient to cash it"

maxwofford:
Can we send the email + auto-void after 6 months or something?

tmb:
@maxwofford how about we do one at 3, auto-void plus another at 6?

maxwofford:
@tmb love it. let's do that

app/jobs/refund_sent_check_job.rb
```diff
@@ -1,11 +1,11 @@
...
+    return if check.deposited? || check.pending_void? || check.voided?	
 
-    CheckMailer.undeposited(check: check)
-    CheckMailer.undeposited_organizers(check: check)
-    check.void!
-  end
-end
+    CheckMailer.undeposited(check: check).deliver_later
+    CheckMailer.undeposited_organizers(check: check).deliver_later
+    check.void!
```
maxwofford: whitespace

maxwofford (changes requested):

maxwofford (dismissed):

app/models/check.rb
```diff
@@ -188,10 +188,13 @@ def approve!
       errors.add(:check, 'has already been approved!')
       return false
     end
-    #if update(approved_at: DateTime.now)
-      #RefundSentCheckJob.set(wait_until: send_date + 1.month).perform_later(self)
-      #return true
-    #end
+    
+    if update(approved_at: DateTime.now)
+      RemindSentCheckJob.set(wait_until: send_date + 3.month).perform_later(self)
```
maxwofford: Do we want to kill this line if we're preventing auto-voids? Assume that would also kill the `undeposited_reminder.html.erb` file.

lachlanjc: I think it's still useful functionality to know a check hasn't been deposited after months—it feels like Bank is looking out for you more

PR #643, opened 2020-05-21
Hide banners on embedded donations, improve frontend

lachlanjc:
- Hides Transparency Mode banner across the app when embedded
- Hides footer on embedded donation form when embedded
- Hides Transparency Mode banner on donation forms for all users
- Typo fix & a few small frontend refinements


app/views/donations/show.html.erb
```diff
@@ -1,5 +1,3 @@
-<p id="notice"><%= notice %></p>
```
maxwofford: I'm confused about this change. If there's a failure processing someone's donation where will that be shown to the donor?

app/views/layouts/application.html.erb
```diff
@@ -21,8 +21,8 @@
     <%= yield(:head) if content_for?(:head) %>
   </head>
   <body class="bg-snow <%= content_for :page_class %>">
-    <% if @event&.is_public && !organizer_signed_in? %>
-      <div class="container container--sm">
+    <% if @event&.is_public && !organizer_signed_in? && !params[:action].end_with?('_donation') %>
```
maxwofford: This seems kinda jank for view logic. What if you had a `before_action` in the `donations_controller` that runs on certain actions? You could set a variable there that you reference in the view

app/views/donations/show.html.erb
lachlanjc: The show page I don’t think is ever shown to users? If so, desperately needs some design love.

I couldn’t find anywhere that generates the `notice` this would display, so I figured it was accidentally left in from generating Rails scaffolds. We use `flash` for all user-facing notifications

maxwofford: > We use flash for all user-facing notifications

Totally forgot, thanks for catching this!

app/views/layouts/application.html.erb
lachlanjc: You're totally right, thanks for flagging. Switched out to already-in-use method.

maxwofford (approved):

PR #646, opened 2020-05-21
HC-7 Unify transaction lists

lachlanjc:
Screen Shot 2020-05-21 at 12 02 26 PM

This is fully functioning! but *not at all* ready for production & needs extensive performance tuning & improvements.


app/controllers/events_controller.rb
```diff
@@ -41,10 +43,14 @@ def create
...
   def show
     authorize @event
+
+    pagination_size = 100
+
     @organizers = @event.organizer_positions.includes(:user)
-    transactions = @event.transactions.includes(:fee_relationship, :comments)
-    emburse_transactions = @event.emburse_transactions.where.not(transaction_time: nil)
-    @transactions = (transactions + emburse_transactions).sort_by(&:date).reverse
+    transactions = @event.transactions.includes(:fee_relationship).limit(pagination_size)
```
lachlanjc: You've undone the N+1 optimizations for comments, did you mean to?
```suggestion
    transactions = @event.transactions.includes(:fee_relationship, :comments).limit(pagination_size)
```

app/controllers/events_controller.rb
```diff
@@ -41,22 +43,28 @@ def create
...
+    pagination_size = 100
+
     @organizers = @event.organizer_positions.includes(:user)
-    @transactions = @event.transactions.includes(:fee_relationship, :comments)
+    transactions = @event.transactions.includes(:fee_relationship, :comments).limit(pagination_size)
+    emburse_transactions = @event.emburse_transactions.where.not(transaction_time: nil).limit(pagination_size)
+    transaction_list = (transactions + emburse_transactions).sort_by(&:date).reverse
+    @transactions = paginate transaction_list, per_page: pagination_size
 
     @invoices_being_deposited = (@event.invoices.where(payout_id: nil, status: 'paid')
```
lachlanjc: Moving this to a separate action, loaded in via JS, could probably reduce database queries here significantly

maxwofford:
@lachlanjc passing this back off to you– think I've got a setup on the backend here that will work whatever you do with the frontend. 

re: endpoint– Go to `/hq/unified_transactions` for the API endpoint. Pagination [config docs here](https://github.com/davidcelis/api-pagination#configuration-optional).

lachlanjc:
Amazing, thanks! Do we want to roll in #666? I would expect the export buttons to export the same list as below it.

maxwofford:
@lachlanjc yes, but there are going to be implementation details to figure out before I add that. One example is: do we export the real transaction list or do we show the exact same numbers that we see on the unified list?

PR #655, opened 2020-05-21
Set User Agent on requests to API

lachlanjc:
This is for #480. It doesn't seem feasible in a simple way to forward the request headers, since the Faraday requests are made via a service via a model via a helper, so we'd have to pass the request through all those & gracefully degrade because some requests are made not-for-the-user.

In the meantime, setting user agent to `Hack Club Bank (commit_hash)`.


maxwofford:
@lachlanjc there are a couple edge cases where that function won't get the commit hash. is that going to break anything?

maxwofford:
Specifically, we're using an experimental Heroku feature to get `ENV['HEROKU_SLUG_COMMIT']` in production & staging environments. They've changed the way it works before without warning & I want to make sure hcb can still make requests if that method returns nil

lachlanjc:
Good point, fixed!

maxwofford (approved):

PR #660, opened 2020-05-23
Fix send only account permissions

maxwofford:
Screenshot 2020-05-22 21 33 01


maxwofford:
In production I'm also running the following to fix the affected spend only accounts that have fiscal sponsorship agreements:

```rb
events = Event.all.select{|e| e.has_fiscal_sponsorship_document == e.is_spend_only}
events.each {|e| e.update(has_fiscal_sponsorship_document: false)}
```

maxwofford:
For posterity:
```rb
irb(main):008:0> events.pluck :id
=> [392, 393, 396, 407, 447, 454, 455, 456, 457, 458, 461, 462, 463, 464, 465, 466, 467, 469, 470, 471]
```

PR #672, opened 2020-06-05
remove routing number from transfers page

pixelyunicorn:
removed routing numbers on transfers page.

it was public in transparency mode, but someone at monzo recommended hiding it. it also doesn't make sense to show it on the transfers page since it doesn't add relevant context.


maxwofford:
Thanks for catching this @pixelyunicorn! Just updated the PR to hide this in public mode but still display for organizers. I think there's utility in showing to organizers, but this should absolutely not be part of the public info.

PR #692, opened 2020-06-29
Add missing receipt count & redesign card page

maxwofford:
@lachlanjc can I get your help making this info helpful and actionable to the user?


lachlanjc:
Sure, I can take a pass designing these. I think it’s time for a new receipt icon we can use across the UI…will try my hand at that.

What’s the outlook on receipts with Stripe Issuing? Are we going to continue sending folks over to another dashboard, or are we bringing receipts in-house? Going to be an interesting design challenge for sure.

maxwofford:
Receipts will be in-house for sure– stripe issuing doesn't have tools for collecting receipts afaik

lachlanjc:
Cool. Are we going to be able to import all the receipts from Emburse into our database/storage? I’m thinking about a kind of checklist/inbox UI for transactions missing receipts…especially for events that are already passed, we need to migrate somehow. (Maybe this could tie into the inbox/notifications view we were already thinking about?)

lachlanjc (approved):

PR #724, opened 2020-07-14
Omit from count events

maxwofford:


maxwofford:
The public stats are updated to pull from any event without "omitted from stats" checked

Screenshot 2020-07-13 22 47 48

maxwofford:
& events have an admin setting on their edit page that removes them from the public stats

Screenshot 2020-07-13 22 49 48

PR #732, opened 2020-07-23
Hide transaction names on non-renamed transactions

maxwofford:


maxwofford:
Screen Shot 2020-07-22 at 9 51 40 PM

PR #751, opened 2020-08-01
[Impersonate] Add power user ability to impersonate a user

motdotla:
Useful for debugging issues around gsuite and for other uses.

How to use:

While signed in as an admin only, you can type in the following url:

```
http://bankurl/users/impersonate?user_id=1234
```

You will then be logged in as that user. The feature is a little janky in that then you will have to log back out to log back in as an admin. That can be solved later but requires much more complexity via multiple cookies typically.

Here's a gif demonstrating its usage and verifying it is only permitted as an admin.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -3,6 +3,14 @@ class UsersController < ApplicationController
   skip_after_action :verify_authorized, except: [:edit, :update]
   before_action :hide_footer
 
+  def impersonate
+    authorize current_user
+
+    sign_in(User.find(params[:user_id]))
+
+    redirect_to root_path
+  end
+
   # view to log in
   def auth
   end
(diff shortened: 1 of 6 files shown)


iamcpdev (dismissed):
lgtm

maxwofford:
Love this feature @motdotla! Can we also add a link for this on the admin search page under each user's account? (https://bank-hackclu-impersonat-ibnqlu.herokuapp.com/admin_search)

iamcpdev:
I can add this super quick. Give me a sec

lachlanjc (dismissed):

iamcpdev:
Lol I fell asleep 😂. I'll add this later today.

iamcpdev:
Hey @motdotla I though I was going to be able to play around with this but I can't figure out rails path helper.

I'll leave this for you whenever you get to it.

iamcpdev (commented):
Looks great

iamcpdev:
I had to revert this @motdotla because it gives this after signing into bank.hackclub.com

image

Slack me if you have any other questions

PR #757, opened 2020-08-10
[753] Reduce home page resources

motdotla:
https://github.com/hackclub/bank/issues/753


app/models/event.rb
```diff
@@ -4,6 +4,38 @@ class Event < ApplicationRecord
...
+        select fr.event_id, sum(t.amount) from fee_relationships fr
+        inner join transactions t on t.fee_relationship_id = fr.id
+        where fr.is_fee_payment is true and t.deleted_at is null
+        group by fr.event_id
+        ) q2
+
+        on q1.event_id = q2.event_id
+      ) q3
+      where fee_balance > 100
+    SQL
```
motdotla, line 31: this was the major performance improvement. it replaced the slow ruby select statement `self.pending_fees`

app/models/event.rb
```diff
@@ -124,29 +151,8 @@ def balance_available
...
-    # fee payments are withdrawals, so negate value
-    -self.fee_payments.sum(:amount)
-  end
-
   def fee_balance
-    total_fees = self.billed_transactions.sum('fee_relationships.fee_amount')
-    total_payments = self.fee_paid
-
-    total_fees - total_payments
+    @fee_balance ||= total_fees - total_fee_payments
```
motdotla, line 155: eventually I'd like to make this a sql query for improved performance but that should wait till I have a better understanding of the transaction system

app/services/static_page_service/index.rb
```diff
@@ -0,0 +1,71 @@
+module StaticPageService
+  class Index
```
motdotla, line 2: this service acted as a convenience wrapper to populate the majority of the data for the / admin page

app/views/static_pages/_card.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= link_to task_path, target: "_blank" do %>
+  <li class="card card--item card--hover relative overflow-visible line-height-3">
+    <strong><%= human_name %></strong>
+    <% if defined? task_name %>
+      <span class="badge <%= count > 0 ? "bg-accent" : "bg-muted" %>"><%= count %></span>
+    <% end %>
+  </li>
+<% end %>
```
motdotla, line 8: this card replaced Card.js where possible. eventually I'd like to do the same for the http remote 3rd party calls but that requires more effort and is not a priority for now

app/views/static_pages/deprecated.html.erb
```diff
@@ -0,0 +1,128 @@
+<% page_sm %>
```
motdotla, line 1: keeping the deprecated page around if we really want access to it. we can remove this later on as the new / admin is proven working.

app/views/static_pages/index.html.erb
```diff
@@ -76,46 +76,48 @@
...
-      <%= react_component "admin_task/Card", human_name: "Donations", task_path: all_donations_path %>
-      <%= react_component "admin_task/Card", human_name: "Disbursements", task_path: disbursements_path %>
-      <%= react_component "admin_task/Card", human_name: "Cards", task_path: emburse_cards_path %>
-      <%= react_component "admin_task/Card", human_name: "G Suites", task_path: g_suites_path%>
-      <%= react_component "admin_task/Card", human_name: "G Suite Accounts", task_name: :g_suite_accounts, task_path: g_suite_accounts_path%>
-      <%= react_component "admin_task/Card", human_name: "G Suite Applications", task_name: :g_suite_applications, task_path: g_suite_applications_path%>
-      <%= react_component "admin_task/Card", human_name: "Negative Events", task_name: :negative_events, task_path: negative_events_path %>
-      <%= react_component "admin_task/Card", human_name: "Search by User", task_path: admin_search_path %>
-      <%= react_component "admin_task/Card", human_name: "Knowledgebase", task_path: 'https://hack.af/hcb-knowledgebase' %>
+      <%= render partial: "static_pages/card", locals: { task_path: events_path, human_name: "Events" } %>
```
motdotla, line 102: these ongoing ajax calls were replaced with a simple static page. I think it wisest we keep the admin simple. less ajax will lead to easier to maintain code which should lead to us being able to provide useful tools to support more quickly. support can be trained to just refresh the page now and again or eventually we can build in a simple page refresh on the whole page.

spec/services/static_page_service/index_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  describe "private" do
+    describe "#admin?" do
+      it "returns true" do
+        result = service.send(:admin?)
+
+        expect(result).to eql(true)
+      end
+    end
+  end
```
motdotla, line 74: quite a few more specs added to service this ticket. this will increase as more tickets are completed, and the app will slowly get into top notch shape without requiring a risky rewrite.

app/models/event.rb
```diff
@@ -77,11 +109,6 @@ def self.create_send_only(event_name, user_emails)
     end
   end
 
-  def self.pending_fees
-    # minimum that you can move with SVB is $1
-    select { |event| event.fee_balance > 100 }
```
motdotla, line 82: this was extremely expensive and will grow more expensive as events and transactions grow

iamcpdev (approved):

PR #768, opened 2020-08-12
762 auto migrate

motdotla:
This will automatically run migrations on deploy - for prod only. The review apps will still require migrations to be run by hand by design.

https://github.com/hackclub/bank/issues/762


iamcpdev (approved):
gtg

PR #779, opened 2020-08-20
[778] GSuite bad domain

motdotla:
Solves for this problem.

What was actually happening here is that the domain has not yet been verified, but the messaging does not indicate that to the user.

Added a remote call to Google to check status of domain. The result is now this:

This will help support team and our users more quickly identify the issue so it can more quickly be solved - an expired verification code that needs to be updated in this case.

Additionally I added paper trail to start tracking changes to this verification code - since I'm surprised to see it different than what Google is showing. This will help us in the future track down other bugs related to this. Ideally Google would provide a way for us to generate and fetch this ourselves but they do not provide an API to do so.

### DIFF
--- a/app/controllers/g_suite_accounts_controller.rb
+++ b/app/controllers/g_suite_accounts_controller.rb
@@ -22,6 +22,8 @@ class GSuiteAccountsController < ApplicationController
     if @g_suite_account.save
       flash[:success] = 'G Suite account application submitted.'
     else
+      Airbrake.notify(@g_suite_account.errors.full_messages.to_sentence)
+
       if @g_suite_account.errors.messages[:domain].present?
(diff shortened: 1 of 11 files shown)


app/controllers/g_suite_accounts_controller.rb
```diff
@@ -22,6 +22,8 @@ def create
     if @g_suite_account.save
       flash[:success] = 'G Suite account application submitted.'
     else
+      Airbrake.notify(@g_suite_account.errors.full_messages.to_sentence)
```
motdotla, line 25: In the future, this will notify us if a user had an issue. Could be removed later on, but for now can help us get out ahead of these issues.

iamcpdev (approved):
lgtm

PR #787, opened 2020-08-21
Rescue when domain does not yet exist and return false for verified

motdotla:
this patches https://heroku-app97991095.airbrake.io/projects/288439/groups/2816241784821871935?tab=overview

Screen Shot 2020-08-21 at 10 12 31 AM


lachlanjc (approved):
Thanks!

PR #811, opened 2020-08-31
Added warning to check notes area

iamcpdev:
Communication is key notes are often left in the notes area with context for the reasoning behind a transfer or reasoning for the delay of one.


app/views/ach_transfers/start_approval.html.erb
```diff
@@ -5,6 +5,7 @@
 <p class="bold">Steps to process ACH:</p>
 <ol>
   <li>Check that all values below look correct.</li>
+  <li class="bold">Check notes area for any messgaes from any HCB staff before processing.</li>
```
lachlanjc:
```suggestion
  <li class="bold">Check the notes area for any messages from Bank staff before processing.</li>
```

lachlanjc (approved):

PR #812, opened 2020-08-31
Add notes to donations

lachlanjc:
Everything should be discuss-able in Bank. This was the most obvious place that needed comments.

(Haven’t fully tested this yet)


lachlanjc:
Have tested this fully, ready to go

PR #813, opened 2020-09-02
Axing the POC heading

iamcpdev:
This is temporary but a POC system doesn't make sense for HCB given the current load/volume of requests we field.


lachlanjc:
Got it. Is this still a useful thing to store/display at all?

PR #818, opened 2020-09-03
Updated check disabling

iamcpdev:
Made the changes @lachlanjc suggested.


app/views/checks/start_void.html.erb
```diff
@@ -4,6 +4,28 @@
 <h1>Void a check to <%= @check.lob_address.name %></h1>
 
 <%= render 'checks/suspended' %>
+#<div class="card border b--warning pb0 mt3 mb3">
```
lachlanjc: This will show up on the page with all the hash marks :) 
```suggestion
<%# put this back when check sending is online %>
<% if false %>
<div class="card border b--warning pb0 mt3 mb3">
```

app/views/checks/start_void.html.erb
```diff
@@ -4,6 +4,28 @@
...
+    #<ul class="mt1 mb2">
+      #<li>This will stop the check from being sent.</li>
+      #<li>No money will leave your account.</li>
+    #</ul>
+  #<% end %>
+  
+  #<div class="border-top card__banner card__darker">
+    #<strong class="warning">Questions?</strong> Talk to your point of contact.
+  #</div>
+  #</div>
```
lachlanjc:
```suggestion
  #</div>
<% end %>
```

lachlanjc (changes requested):
Thanks! One quick thing is the comment will show up on the page

iamcpdev (commented):
Done!

lachlanjc (approved):

lachlanjc:
Thanks for your work here!!

PR #851, opened 2020-09-12
Removed link to disbursement

iamcpdev:
Update for ops it's hard to copy the memo line because it's a link


lachlanjc (approved):
¯\_(ツ)_/¯ go for it!

PR #855, opened 2020-09-14
HC-158 Receipt uploading

maxwofford:


maxwofford:
People are making card transactions at the time of writing, so for the sake of time with @lachlanjc's irl permission I'm merging this. Will revert if this takes down any pages.

PR #863, opened 2020-09-15
Transaction Engine > Raw -> Hashed -> Canonical Transactions

motdotla:
This PR creates the foundation for the first pillar of the new modern transaction engine: taking raw data and putting it into our own canonical transactions.

The flow at a high level is:

`raw` -> `hashed` -> `canonical`

In addition to there, there is a bit of early foundational code for the EventMapping Engine (the 2nd pillar of the transaction engine) but later PRs will start to work out that code better.

You can test this PR locally by entering the console and running:

```
TransactionEngineJob::Nightly.new.perform
```


Gemfile
```diff
@@ -149,3 +149,4 @@ gem 'namae' # multi-cultural human name parser
 gem 'paper_trail' # track changes on models
 gem 'sidekiq-cron', '~> 1.1' # run sidekiq scheduled tasks
 gem 'strong_migrations' # protects against risky migrations that could cause application harm on deploy
+gem 'xxhash' # fast hashing
```
motdotla, line 153: very fast hashing algorithm across multiple languages - so should be quite future proof while serving performance needs. I've used this in production before and it has performed very well.

app/services/hashed_transaction_service/plaid_transaction/import.rb
```diff
@@ -0,0 +1,28 @@
+module HashedTransactionService
+  module PlaidTransaction
+    class Import
+      def run
+        ::PlaidTransaction.find_each do |pt|
+          attrs = {
+            primary_hash: primary_hash(pt),
+            plaid_transaction_id: pt.id
```
motdotla: For now dependent on plaid_transaction_id still. I will start to unravel this next with importing Stripe and Emburse Transactions.

app/services/hashed_transaction_service/primary_hash.rb
```diff
@@ -0,0 +1,57 @@
+module HashedTransactionService
+  class PrimaryHash
```
motdotla: this service does the bulk of the work for this PR - maintaining a consistent hashing algorithm. Changing this will change things downstream. This might still evolve - at which time I'll just delete_all on the hashed_transactions table.

app/services/plaid_transaction_service/likely/ach.rb
```diff
@@ -0,0 +1,9 @@
+module PlaidTransactionService
+  module Likely
```
motdotla: The multiple `::Likely` module classes below all duplicate the logic of `transaction.potential_*` in transaction.rb.

lachlanjc (dismissed):
If it works for you, works for me!

app/services/hashed_transaction_service/primary_hash.rb
maxwofford: Wouldn't delete_all remove the canonical association? How does the linking between raw & canonical through hashed work?

app/models/raw_emburse_transaction.rb
```diff
@@ -0,0 +1,33 @@
+class RawEmburseTransaction < ApplicationRecord
```
maxwofford, line 1: Is there a reason to not have a `has_one :hashed_transaction` for the raw tx classes?

app/models/hashed_transaction.rb
```diff
@@ -0,0 +1,19 @@
+class HashedTransaction < ApplicationRecord
+  belongs_to :raw_plaid_transaction, optional: true
+  belongs_to :raw_emburse_transaction, optional: true
```
maxwofford, line 3: What's the reasoning for this vs polymorphic hashed_transactions?

app/services/event_mapping_engine/map/github.rb
```diff
@@ -0,0 +1,30 @@
...
+            event_id: guessed_event_id
+          }
+          ::CanonicalEventMapping.create!(attrs)
+        end
+      end
+
+      private
+
+      def likely_githubs
+        ::CanonicalTransaction.exclude(excluded_ids).likely_github
```
maxwofford, line 21: I read this as 'check canonical transactions, excluding ones with an event mapping, with memo like "github grant"', but GitHub grants are always associated with an event.

app/services/transaction_engine/canonical_transaction_service/import.rb
```diff
@@ -0,0 +1,40 @@
...
+            ::CanonicalHashedMapping.create!(attrs)
+          end
+
+        end
+      end
+
+      private
+
+      def hashed_transactions_ready_for_processing
+        ::HashedTransaction.where('id not in (?)', duplicate_hashed_transaction_ids + previously_processed_hashed_transaction_ids)
```
maxwofford, line 28: Doesn't this mean duplicates picked up in the same nightly job won't create a canonical TX? Do we want to hold them at that stage, or create a canonical TX for just one?

app/services/transaction_engine/hashed_transaction_service/primary_hash.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      def input_array
+        [
+          @date,
+          @memo,
+          @amount_cents
+        ]
+      end
+
+      def seed
```
maxwofford, line 34: What is this used by?

app/models/raw_emburse_transaction.rb
```diff
@@ -0,0 +1,33 @@
...
+  monetize :amount_cents
+
+  def memo
+    return "#{merchant_name}, Card: #{card_description}, Member: #{member_full_name}" if amount_cents < 0
+
+    "Transfer from #{bank_account_description}"
```
maxwofford, line 7: This could also be a refund, not just a transfer from a bank account.

maxwofford: Also, it feels funny to have memo logic in the RawEmburseTransaction. Not proposing a better way, just commenting that this doesn't match the abstraction in my head of what this model is for.

app/services/transaction_engine/canonical_transaction_service/import.rb
maxwofford: Scratch my comment. After pausing to think longer it totally makes sense why we'd want to hold it at this stage & wait for human intervention

config/schedule.yml
```diff
@@ -3,7 +3,7 @@ g_suite_mark_verifieds_job:
...
   queue: "default"
   
-plaid_transaction_imports_job:
-  cron: "0 * * * *" # run every 1 hour
-  class: "PlaidTransactionJob::Imports"
-  queue: "default"
\ No newline at end of file
+transaction_engine_nightly_job:
+  cron: "0 6 * * *" # run at 6am UTC each morning
+  class: "TransactionEngineJob::Nightly"
```
maxwofford, line 8: How do you envision this handling TXs coming in throughout the day? Plaid only syncs daily or so, but Emburse & Stripe update realtime.

maxwofford (dismissed):
Left a few comments I'd like to get some clarity on, but none that require blocking this from merge– feel free to merge when you're comfortable with it @motdotla

app/services/hashed_transaction_service/primary_hash.rb
motdotla: indeed. that was not ideal. has been removed.

app/models/raw_emburse_transaction.rb
motdotla: no reason. i just tend not to add has_* until needed.

app/models/raw_emburse_transaction.rb
motdotla: it's mainly a convenience method, that the canonical_transaction can use to build it's memo.

making a ticket to handle refund (i think this code is not handling any of those scenarios yet (especially around adjusting for negative values)). thought there is to make small adjustments as we start rolling out comparisons of this data with the current live data

app/services/transaction_engine/hashed_transaction_service/primary_hash.rb
motdotla: it's used to seed the generated hash from XXhash. 0 is default. before we launch this we should settle on our own secret seed that we protect.

config/schedule.yml
motdotla: We can run more often, but currently I anticipate pending handling the near real-time transactions. These are the settled transactions that I'm anticipating don't really settle until end of day.

app/models/hashed_transaction.rb
motdotla: considered that. but I'm thinking that some day as we phase out old systems (emburse first), we might just drop the raw tables completely. that has other implications but I could anticipate that.

maxwofford (approved):

PR #864, opened 2020-09-15
Unified transaction lists: Bank, Emburse, Stripe

lachlanjc:
3FE13AC0-DAFF-4A5F-A478-0EF10AD4FFC2

- Removes the row expansion/showing fee + donation/invoice as multiple rows
- Adds Emburse + Stripe Card transactions to main transactions table
- Adds icons/tooltips to each transaction showing the origin/type
- Paginates transactions for smaller pages & faster load times


app/controllers/events_controller.rb
```diff
@@ -44,7 +45,11 @@ def show
     authorize @event
     @organizers = @event.organizer_positions.includes(:user)
 
-    @transactions = (@event.transactions.includes(:fee_relationship, :comments) + @event.stripe_authorizations.approved).sort_by(&:created_at).reverse
+    @transactions = paginate((
```
lachlanjc: N+1 up the wazoo!!! I could totally use some backend help here on how to make this go mildly somewhat efficiently. The real N+1s are in the views, where we’re referencing the associated fees/invoices/donations/their metadata without preloading them.

app/views/donations/_donation.html.erb
```diff
@@ -1,73 +1,27 @@
 <% transaction = donation.payout.t_transaction %>
```
lachlanjc: Example of what’s generating the N+1s. The potential (& boy do I mean potential) upside of generating N+1s in partials here is that since we’re paginating, we’ll potentially load less of this N+1 data because it won’t get called for every transaction? But we should fix with something more efficient regardless.

app/views/invoices/_invoice.html.erb
```diff
@@ -21,55 +23,7 @@
   <td>
     <%= render_money transaction.amount + refund_t&.amount.to_i %>
   </td>
-  <td>
```
lachlanjc: Removing the inner/expandable rows on these since they’re confusing & require a lot of code

app/views/kaminari/_first_page.html.erb
```diff
@@ -0,0 +1,3 @@
+<li>
```
lachlanjc: Generated out of our pagination library so we can customize them

app/views/stripe_authorizations/_stripe_authorization.html.erb
```diff
@@ -10,7 +13,7 @@
   </td>
   <td></td>
   <td class="nowrap">
-    <%= render_money -stripe_authorization.amount %>
+    <%= render_money stripe_authorization.amount %>
```
lachlanjc: All the Stripe Card transactions are currently showing the opposite of the correct amount & color

lachlanjc:
Two issues:

- This will be initially confusing to users who still have a remaining Emburse balance, since the numbers won’t add up. My proposal: add a temporary/fake pending transaction to the top of the list for the transfer with a link to an Airtable form with an explanation.
- The Export button won’t export all these transactions, only the Bank ones. Could add a note specifying that, unless we can get Export working quickly enough?

lachlanjc:
@maxwofford approved irl!

PR #882, opened 2020-09-19
Changed email auth are sent to

iamcpdev:
Right now these were flooding bank@hackclub.com.


lachlanjc:
Is that address a thing? Otherwise will be lost in the void

iamcpdev:
Yep it’s an email I made specifically for stuff like this.

PR #888, opened 2020-09-21
Update cards page for migration week

lachlanjc:


app/views/comments/_comments.html.erb
```diff
@@ -1,2 +1,2 @@
 <%= render 'comments/list' %>
-<%= render 'comments/form' %>
+<%= render 'comments/form' if current_user %>
```
lachlanjc: This is a semi-fix for seeing the note creation UI if you’re not logged in. This fixes it if you’re not logged in, but not if you are but you’re viewing an event in transparency mode. Ideally we would use `organizer_signed_in?` but this partial is also used on admin pages where I think that won’t work, so just leaving it at this for now.

app/views/events/_nav.html.erb
```diff
@@ -16,19 +16,15 @@
...
+    'card', 'purple', event_cards_overview_path(event_id: @event.slug) %>
   <%= dock_card 'Send',
     'Send money from Hack Club Bank.',
     'transactions', 'success',
     event_transfers_path(event_id: @event.slug) %>
   <%= dock_card 'Donations',
     'Receive donations online.',
     'support', 'warning',
-    event_donation_overview_path(event_id: @event.slug) if @event.donation_page_enabled %>
+    event_donation_overview_path(event_id: @event.slug) if @event.donation_page_enabled && !@event.is_spend_only %>
```
lachlanjc: I didn’t check with anyone but if you can only spend money, you shouldn’t have invoices or donations would be my assumption

PR #906, opened 2020-09-25
Polymorphic receipts

maxwofford:


app/views/emburse_transactions/_emburse_transaction.html.erb
```diff
@@ -21,7 +21,7 @@
     <% end %>
   </td>
   <td>
-    <%= list_badge_for et.comments.size, 'note', 'post', optional: true %>
+    <%= list_badge_for et.receipts.size, 'receipt', 'payment-docs', required: true  %>
```
lachlanjc: This should be added, not replacing, the comments indicator I think

PR #917, opened 2020-10-01
HC-203 Allow opting into session reporting

maxwofford:
New users will be confronted with an opt-in checkbox on signup:
Screen Shot 2020-10-01 at 2 49 09 PM

Existing users will not be opted-in automatically, but can choose to on their settings page:
Screen Shot 2020-10-01 at 2 50 02 PM


lachlanjc:
Heads up: I’m planning to take a design pass on this, as well as implement the tracking itself :), so don’t merge until both of those are done.

@maxwofford—did you want to add back FullStory for this? Whatever tool is most useful to us, I think. Also, should we opt users in or out by default? Previously the opt-in was required, so many already consented, but I suppose newer users have not. But I think people pretty rarely visit the settings page, so unless we ask folks in #bank, I feel like no one will do it since they don’t even know the option exists?

maxwofford:
Let's do fullstory as a default, but I'm happy with an alternative if we find something that gives the same experience.

The db migration explains the new defaults, but to;dr: old accounts have it turned off by default, and new accounts have it on by default. New accounts see the setting on their account creation page.

maxwofford:
I'm totally ok with old users not having this turned on, but I might post in #bank or DM a few users i personally know to turn it on.

If you have thoughts about how to make old users comfortable turning it on I'm all ears

lachlanjc:
Design

lachlanjc (approved):

PR #928, opened 2020-10-06
Switch G Suite to Google Workspace

lachlanjc:
Updated copy for G Suite’s rebranding. I’m not migrating any of our route names, tables etc for now, since I don’t want to break stuff & I don’t think we really care. I also did a quick design pass on the admin pages to make them clearer & more useful (like hiding “rejected at” & shading rejected rows instead).

New UI (ofc these aren’t shown at the same time):

897C3BA6-8CE6-4B2C-9464-6624C8EB8980

Admin show page:
257FA085-3DD1-455C-B6FC-013664F6B7BF

I also switched the admin homepage tools to a 2-column grid with a cleaner layout—I find it really difficult to find the tool I’m looking for with such a chaotic layout, so this is a little less dense but I think easier to use. Lmk if you want it reversed.


lachlanjc:
Admin tools homepage before:

46CC98C1-4DA3-4B04-BBA7-23D7D1231F03

After:

2450A214-8308-43AE-A869-A0FD21F8AE5D

PR #929, opened 2020-10-07
Removed link to card

iamcpdev:


app/views/stripe_cards/show.html.erb
```diff
@@ -4,7 +4,6 @@
 <article class="check--form flex align-start justify-center mt2 mb3">
   <section class="center mt0 mx-auto" style="width: 400px; max-width: 100%;">
     <%= render @card %>
-    <%= link_to @card, class: 'btn bg-info mt2 md-mt3' do %>
```
lachlanjc: This will introduce a 500 error, as there’s the `end` line on line 9 that will be un-matched. Also, this link is for a button below the card, but we’re currently not using it/it’s not rendered (see the `end if false` part). Sorry, this is really unclear code/shouldn’t really be here, but yeah, this won’t fix the issue & will take down the page :)

This isn’t the easiest fix since that card is being rendered by a partial (line 6 renders `app/views/stripe_cards/_stripe_card.html.erb` via Rails magic), which we need to keep the link in. So we could either not use the partial (probably not a good idea since the markup is complex & we don’t want to repeat it), or add an option (like `link: false`) to the partial to conditionally not wrap in a link. It’s on my list to fix, alongside not showing team members each others’ cards, but I can work on that tomorrow.

PR #938, opened 2020-10-08
Zero out check task size

maxwofford:


maxwofford:
@motdotla when going through this I found some duplicate code between the service & controller (ex https://github.com/hackclub/bank/blob/master/app/services/static_page_service/index.rb#L22 & https://github.com/hackclub/bank/blob/master/app/controllers/static_pages_controller.rb#L210)

Cool if I just put all the counts in the service & have the controller pass through to that? or do you have a different vision for this

PR #954, opened 2020-10-11
Match likely historical transactions

maxwofford:
I just threw this together so I could get a better feel for the stuff in #863. All a WIP that I'd love to go over with you (@motdotla) on our next call.


maxwofford:
Whoops! Compared the wrong branch. Closing in favor of #955

PR #955, opened 2020-10-11
Initial historical assignment (#952)

maxwofford:
I just threw this together so I could get a better feel for the stuff in #863. All a WIP that I'd love to go over with you (@motdotla) on our next call.


app/services/event_mapping_engine/map/historical.rb
```diff
@@ -1,11 +1,36 @@
...
+        likely_historicals.find_each do |ct|
+          historical_transactions = Transaction.where(
+            amount: ct.amount_cents,
+            name: ct.memo,
+            date: ct.date
+          )
+
+          return unless historical_transactions.size == 1
+
+          guessed_event_id = historical_transactions.first&.event&.id
```
motdotla: nice. yeah, this is the idea

app/services/event_mapping_engine/map/historical.rb
```diff
@@ -1,11 +1,36 @@
 module EventMappingEngine
   module Map
     class Historical
-      def initialize
+      def run
+        likely_historicals.find_each do |ct|
+          historical_transactions = Transaction.where(
+            amount: ct.amount_cents,
+            name: ct.memo,
```
motdotla: I'm not positive we will get complete matches here because the memo for the canonical transactions are upcased.

motdotla: we might need to instead match on plaid ids.

we can add some convenience methods to the `CanonicalTransaction` model to fetch the raw source plaid id, and then we can match 1 for 1 with the plaid_id present in the `Transaction`

motdotla (changes requested):
can you do some initial digging and see if this current logic gets us more accurate matches or if matching on plaid_id, emburse_id, etc would be more accurate historically ?

app/services/event_mapping_engine/map/historical.rb
maxwofford: Good point– I'll play with something like this:
```rb
          raw_plaid_ids = ct.hashed_transactions.pluck(:raw_plaid).compact

          historical_transactions = raw_plaid_ids.map do |rpid|
            Transaction.find_by(plaid_id: rpid)
          end
```

app/services/event_mapping_engine/map/historical.rb
```diff
@@ -1,11 +1,36 @@
...
+          ::CanonicalEventMapping.create!(attrs)
+        end
       end
 
-      def run
-        # IMPLEMENT
+      private
+
+      def likely_historicals
+        ::CanonicalTransaction.where('created_at <= ?', last_historical_transaction_date)
```
motdotla: rather than filter based on time, let's return everything that has not yet been mapped. the following gives an example of that:

https://github.com/hackclub/bank/pull/863/files#diff-7c96101cfa8411f3f9d1f8fda186f4d69db3c1984f6239887c8716fe56cf8d35R24


then later on as we dependency inject start_date and end_date throughout these services, we will filter down additionally on time.

motdotla (changes requested):
just one more request, otherwise looks solid

app/services/event_mapping_engine/map/historical.rb
maxwofford: I'm down to remove the date constraint, but this will still look through CanonicalTransactions that don't have an event related on purpose (ie. Stripe Topup TXs). Maybe we could have a scope on canonical TXs for anything that needs to be matched, then update it in the future when we decide how to denote TXs that shouldn't be associated?

maxwofford: https://github.com/hackclub/bank/pull/955/commits/6e9a6014e109a0e1ba51600b5f661a8233b970c1

It seems to me like GitHub & Historical mappings need the same thing & I'd imagine future mappings will also need it.

Now there's an `awaiting_match` scope for any canonical TX that needs to get mapped. I'm not attached to the name, but this is just a proof of concept.

motdotla: baller

motdotla: >  through CanonicalTransactions that don't have an event related on purpose (ie. Stripe Topup TXs

yeah, that is ok for now. we'll be following up with a mechanism to solve that soon enough.

motdotla (approved):

PR #977, opened 2020-10-15
Initial no/lost receipt logic

maxwofford:


lachlanjc:
Really appreciate you getting this going. Thoughts on making it `lost_or_no_receipt_at`, so we can make it a timeline event later with when it was marked that way?

lachlanjc:
Great! Will polish design/copy this week

maxwofford:
@lachlanjc we don't want users abusing this once we start adding in validations for receipts (ie. email the user if they have over $75 in missing receipts), so we might want a notice or something saying we'll review when someone marks a TX as missing receipt

lachlanjc:
Great. I might add a self.receipt_required? method on those models so this won't show up on bank account transactions. I'm not quite sure how to say this in the copy—it's not that we'll get notified, but we audit them? What exactly is the implication? Is there a hard limit we want to enforce? Does it depend on transaction amount?

maxwofford:
@lachlanjc if there isn't a good way to put this up now we can totally wait until there are measurable requirements about receipt uploading in place

lachlanjc:
Sorry for the delay here, just letting you know this is on my list for this week 👍

lachlanjc (approved):

lachlanjc:
I’m ready to merge but don’t want to break anything since this has DB changes. Can we ship today/Friday? @maxwofford

PR #985, opened 2020-10-15
Migrate initial admin pages to AdminController

lachlanjc:
- Moved all the admin pages out of StaticPagesController.
- Called it AdminController instead of AdminPagesController since a bunch are API endpoints/exports so they aren’t pages.
- Also made the user search *way* faster & a bit denser/prettier since I’d never looked at that page before.
- If anyone’s got a better place for `airtable_info`, please suggest!


motdotla (approved):
LGTM

PR #986, opened 2020-10-16
Custom maintenance page

maxwofford:
Closes https://github.com/hackclub/bank/issues/967

Once this is merged, we need to serve it with GitHub Pages with these settings:
Screen Shot 2020-10-16 at 9 53 11 AM

Then run configure Heroku to point to the GH Page URL:
```
heroku config:set \
  ERROR_PAGE_URL=//hackclub.github.io/bank/application-error.html \
  MAINTENANCE_PAGE_URL=//hackclub.github.io/bank/maintenance-mode.html
```


maxwofford:
@lachlanjc this is just the code for the existing heroku error pages. Can you turn these static pages into something that looks nice? Anything in the `/docs` folder will be served from GitHub Pages, so you can link external resources in there if they're relative URLs

maxwofford:
The pages are now live on github pages (keep in mind they are error pages) https://hackclub.github.io/bank/application-error.html.

The configuration variables are set on heroku and will only show up on the next deploy.

PR #1009, opened 2020-10-22
Sort stripe authorizations by date

maxwofford:


lachlanjc (approved):

lachlanjc:
Absolutely. Doesn’t that put the latest at the bottom though? In every list we have in the app, I think we want latest at the top

maxwofford:
Latest at the top-- created_at: :desc means sort largest first, and more recent dates are larger

lachlanjc:
Totally not paying attention/blanked. Great

PR #1016, opened 2020-10-24
Stripe authorizations have comments

maxwofford:
I need to add some notes for the TXs going on my cards (specifically the cards I'm lending out to people).


lachlanjc:
I can’t test these on the deploy since it’s a Stripe page but:

- Let’s remove the Notes heading to be consistent/bridge towards timeline
- If a non-organizer is viewing that page, something is wrong with our policy—those should be private
- The routes were messed up when I tested this previously & it didn’t work—did posting & editing work for you?

maxwofford:
Ok, just fixed up the routing issues. Same front-end as before

PR #1038, opened 2020-11-09
Stripe refund not tied to authorizations

maxwofford:
Puts the 'fun' in 'refund'


maxwofford:
Once this is merged this command should be run in production: `StripeAuthorization.all.each {|sa|sa.save!}`

PR #1039, opened 2020-11-09
Removed Emburse language and added Stripe

iamcpdev:
With Emburse no longer being our provider we don’t need this language here anymore and it can be substituted.

### DIFF
--- a/app/views/organizer_position_deletion_requests/_form.html.erb
+++ b/app/views/organizer_position_deletion_requests/_form.html.erb
@@ -20,7 +20,7 @@
   </div>
 
   <p>
-    Does <%= user_mention @target %> have transactions without receipts on Emburse?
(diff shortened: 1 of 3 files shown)


app/views/organizer_position_deletion_requests/_form.html.erb
```diff
@@ -20,7 +20,7 @@
   </div>
 
   <p>
-    Does <%= user_mention @target %> have transactions without receipts on Emburse?
+    Does <%= user_mention @target %> have transactions without receipts on Stripe?
```
maxwofford: /Stripe/their cards/s

iamcpdev: ?

maxwofford: Most users don't know what Stripe is. Let's replace "Stripe" with "cards"

PR #1052, opened 2020-11-13
Don't flash users with bright light when loading dark mode

maxwofford:


iamcpdev:
This looks good to go @maxwofford did you want to deploy?

PR #1102, opened 2020-12-20
Remove mentions of point of contact & update FAQ

garyhtou:
Added a FAQ about international projects. Let me know your thoughts on the answer below.

> **Are international projects, events, and clubs (outside of the US) eligible to use Hack Club Bank?**
> We are currently in the process of building out our support for international projects. As a result, we do not have a definitive answer. If you are interested, we would love for you to apply so we can hear what you're working on and whether we can possibly help.

Also, in `application_helper.rb` I added `help_email` and `help_phone`, not sure if those need `.html_safe` 🤷‍♂️. I'm new to ruby on rails :)

Haven't been able to test these changes since I'm having issues getting Bank to run locally on my machine.


app/helpers/application_helper.rb
```diff
@@ -160,8 +160,16 @@ def filterbar_item(label, name, selected = false)
...
-  end
+    content_tag :span, "Contact the Bank team at #{help_email} or #{help_phone}.".html_safe
+	end
+	
+	def help_email
+		mail_to 'bank@hackclub.com'
+	end
+
+	def help_phone
+		link_to '+1-360-502-2935', 'tel:+1-360-502-2935'
```
motdotla: can you clean up the extra spacing here - and the methods above.

motdotla (changes requested):
great stuff. thanks, Gary. just one request to fix up spacing.

garyhtou:
let me know if that fixed it!

I ditched vscode and used rubymine to reformat the file (is there a preferred formatter like rubocop? or if rubymine's default formatter fine?)

motdotla (approved):
that's good. thanks. just update with branch master and go ahead and merge.

PR #1103, opened 2020-12-22
Add backdrop blur for tooltips

maxwofford:


maxwofford:
Previously tooltips just darkened their background:
Screen Shot 2020-12-22 at 11 34 32 AM

Now (on supported browsers like safari & google), they also blur their background:
Screen Shot 2020-12-22 at 12 27 03 PM

PR #1142, opened 2021-01-21
Making ACH forms more efficient for admins

garyhtou:
- Change the input format of ACH arrival dates from `YYYY-MM-DD` to `MM/DD/YYYY` because the latter is the format that SVB displays it. This will make it possible to copy and paste the arrival date into our form, rather than manually converting it into YYYY-MM-DD and typing it in. (fast and less error-prone).
- On admin ACH approval page, move the displayed phone number above the nickname since this is the order which it appears on the SVB ACH form
- Force the phone number input on the user's new ACH form to be exactly 10 digits. Some people have been putting in country codes (which SVB does not accept), slightly slowing down admin work.


garyhtou:
For bullet point number one: Looking back at the commit history, the date format was original `MM/DD/YYYY` but was later changed to the current `YYYY-MM-DD` in _[4b67ff4f8ccd225f4dab63e740890ac05f2f446f "Fix incorrect date entry format"]_.

I test the changes for the new date format, however, it seems to not work. I'm guessing the default date format is YYYY-MM-DD and it needs to be explicitly converted or set somewhere that the incoming data is MM/DD/YYYY format. Help/pointers would be appreciated here (but, this is fairly low priority compared to everything else).

motdotla:
@garyhtou yeah, probably a conversion issue. chronic gem generally handles these situations the best with least amount of effort.

motdotla:
moved your changes here: https://github.com/hackclub/bank/pull/1289

PR #1231, opened 2021-02-19
Add to donation receipt that no goods or services were provided

garyhtou:
Would there ever be a case where goods or services **ARE** provided?

https://secure.helpscout.net/conversation/1423234767/132?folderId=4386696


maxwofford:
That's a good questions. I guess the assumption we made was that users would invoice if goods/services were provided

EDIT: to clarify, donation receipts are sent so the donor can record it as a donation to the IRS

maxwofford (approved):

maxwofford:
LGTM

I'm holding off on merging b/c it's a draft pr

garyhtou:
Good point. This should be good to merge then

PR #1255, opened 2021-02-27
Remove No Card Support Message

sampoder:
I'm fairly sure that it now exports card transactions, at least when I tried it out it did :)

best, 

sam

### DIFF
--- a/app/views/events/show.html.erb
+++ b/app/views/events/show.html.erb
@@ -91,8 +91,7 @@
   <button
     type="button"
     <%= organizer_signed_in? ? 'data-behavior=menu_toggle aria-expanded=false tabindex=0' : 'disabled=disabled' %>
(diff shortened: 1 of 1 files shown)


app/views/events/show.html.erb
```diff
@@ -91,8 +91,7 @@
   <button
     type="button"
     <%= organizer_signed_in? ? 'data-behavior=menu_toggle aria-expanded=false tabindex=0' : 'disabled=disabled' %>
-    class="btn bg-info menu__toggle overflow-visible tooltipped tooltipped--n"
-    aria-label="Export currently only processes bank transactions—let the Bank team know if you want card transactions.">
+    class="btn bg-info menu__toggle overflow-visible tooltipped tooltipped--n">
```
sampoder, line 94: I think this change needs to be made, but I'm not sure

```suggestion
    class="btn bg-info menu__toggle overflow-visible">
```

PR #1286, opened 2021-03-06
Admin Process ACH: Show amount in dollars, not cents

garyhtou:


motdotla:
good catch here @garyhtou !

PR #1307, opened 2021-03-09
Default has_fiscal_sponsorship_document to false on event creation

garyhtou:
I believe by default, an event shouldn't have a fiscal sponsorship document. `has_fiscal_sponsorship_document` should only be true when they have signed the contract and are moved off the spend-only plan.

Can anyone confirm?


motdotla:
good catch :+1:

PR #1356, opened 2021-03-22
Custom sort for event listing

garyhtou:
#1156

I'm kinda suck

I changed `Event.reorder(:name)` to `Event.reorder(CUSTOM_SORT)` in the new disbursement view to test, but it seems to not work.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -148,7 +148,7 @@ class Event < ApplicationRecord
                       WHEN id = 999 THEN '2'
                       WHEN id = 689 THEN '3'
                       WHEN id = 636 THEN '4'
-                      ELSE name END ASC"
+                      ELSE 'z' || name END ASC"
 
   # Used by the api's '/event' POST route
   def self.create_send_only(event_name, user_emails)
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -144,6 +144,12 @@ class Event < ApplicationRecord
 
   before_create :default_values
 
+  CUSTOM_SORT = "CASE WHEN id = 183 THEN '1'
+                      WHEN id = 999 THEN '2'
+                      WHEN id = 689 THEN '3'
+                      WHEN id = 636 THEN '4'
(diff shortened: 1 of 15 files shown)


app/models/event.rb
```diff
@@ -144,6 +144,12 @@ class Event < ApplicationRecord
 
   before_create :default_values
 
+  CUSTOM_SORT = "CASE WHEN id = 183 THEN '1'
+                      WHEN id = 999 THEN '2'
+                      WHEN id = 689 THEN '3'
+                      WHEN id = 636 THEN '4'
+                      ELSE name END ASC"
```
motdotla: looks almost there from testing the raw sql. but needs a way to still put event names that start with a number at the end:


```
select * from events
order by
CASE WHEN id = 183 THEN '1'
                  WHEN id = 999 THEN '2'
                  WHEN id = 689 THEN '3'
                  WHEN id = 636 THEN '4'
                  ELSE 'z' || name END ASC
```

motdotla (changes requested):

app/views/admin/disbursement_new.html.erb
```diff
@@ -8,13 +8,13 @@
     <p>
       <%= form.label :source_event_id, "Source event" %>
       <br/>
-      <%= form.collection_select(:source_event_id, Event.reorder(:name), :id, :admin_dropdown_description, include_blank: "Select one…") %>
+      <%= form.collection_select(:source_event_id, Event.reorder(CUSTOM_SORT), :id, :admin_dropdown_description, include_blank: "Select one…") %>
```
motdotla: `Event.reorder(Event::CUSTOM_SORT)`

motdotla (changes requested):
couple minor changes and then this looks good

garyhtou:
Yay! I think I got it. Thanks for the help 😄
lmk if any edits should be made

motdotla (approved):
nice work here

PR #1390, opened 2021-03-28
Autofill: "Yes, I have used Hack Club Bank before" or "No, first time!" based on a user's project amount

sampoder:
This removes the need to fill in this field


sampoder:
Alright, I feel like this is ready for review... not sure why deployment is failing :(

motdotla:
I believe deployment was failing because of the mime-magic issue (since patched). I will merge shortly. Thanks Sam!

PR #1424, opened 2021-04-06
Redirect user to page before auth

maxwofford:
Reverts hackclub/bank#1399


maxwofford:
I've had the chance to test this further in development & I'm more confident in it.

PR #1442, opened 2021-04-08
Add buildpacks that are used in heroku prod

garyhtou:
They will be used in review dynos (such as for generating pdfs with `wkhtmltopdf`)


garyhtou:
it works

motdotla (approved):

PR #1454, opened 2021-04-11
Negative balance for Athul's Personal Trip

garyhtou:
We have been withholding a portion of Athul's salary in [event 946](https://bank.hackclub.com/athulwithholding).

He needs access to the funds ASAP, however, we must disburse the funds to a separate event ([event 1073](https://bank.hackclub.com/athul-personal-bmw-trip)) for him to spend in. The disbursement will take a few days to process. So this will temporarily allow his event to go into negative spending.


garyhtou:
No longer necessary, we can just use the existing withholding account

PR #1484, opened 2021-04-17
Job to set all point of contacts to Mel (#1426)

garyhtou:
Tested local and it seems to work.

Run in console after merging:

```ruby
TempJob::EventPointOfContactMel.new.perform
```

### DIFF
--- a/app/jobs/temp_job/event_point_of_contact_mel.rb
+++ b/app/jobs/temp_job/event_point_of_contact_mel.rb
@@ -1,0 +1,11 @@
+module TempJob
+  class EventPointOfContactMel
+    def perform
+			mel_user_id = 2046
+			Event.where.not(point_of_contact_id: mel_user_id).find_each(batch_size: 100) do |e|
+				e.point_of_contact_id = mel_user_id
+				e.save!
+			end
+    end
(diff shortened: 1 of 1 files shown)


garyhtou:
#1426

app/jobs/temp_job/event_point_of_contact_mel.rb
```diff
@@ -0,0 +1,11 @@
+module TempJob
+  class EventPointOfContactMel
+    def perform
+			mel_user_id = 2046
+			Event.where.not(point_of_contact_id: mel_user_id).find_each(batch_size: 100) do |e|
+				e.point_of_contact_id = mel_user_id
```
motdotla: let's switch this to `.update_column(:point_of_contact_id, mel_user_id)` just as a precaution to avoid any possible callbacks being triggered from .save

motdotla (changes requested):

app/jobs/temp_job/event_point_of_contact_mel.rb
garyhtou: thanks for the feedback 😄

PR #1615, opened 2021-05-23
Removal of has_fiscal_sponsorship_document boolean

garyhtou:
#1539


app/models/event.rb
```diff
@@ -9,7 +9,7 @@ class Event < ApplicationRecord
   monetize :total_fees_v2_cents
 
   default_scope { order(id: :asc) }
-  scope :pending, -> { where(has_fiscal_sponsorship_document: false) }
```
motdotla: at this point in time everything with has_fiscal_sponsorship_document set to false is basically 'unapproved', right? Since otherwise, Mel has been creating approved events by hand now for the past 2-3 weeks?

If, so, I think we can simply rename locations that use `.pending` to instead use `.unapproved`

garyhtou: yeah, sounds reasonable

PR #1582, opened 2021-05-19
Use new transaction engine for stats endpoint

garyhtou:
Updated the endpoint to be compatible with the original schema and hackclub.com/bank:

- adding `revenue` (fee revenue)
- return `events` as an array of `{ created_at: 1530053412}`

Also updated the admin dashboard's "#____ in transaction & counting..." to use the new engine.

### Tested locally:
[hackclub.com/bank](https://hackclub.com/bank) | bank admin
-|-
image | image

PR #1616, opened 2021-05-23
New transaction engine for stats endpoint

garyhtou:
fixed error in #1582

also moved calculations into a service and added more useful data to the endpoint

image


motdotla (approved):
nice improvements

PR #1620, opened 2021-05-24
Don't mark a donation as deposited if it has been refunded

garyhtou:
might fix #1561 (untested)

it's eating up our airbreak quota so I decided to take a look at it


motdotla:
thanks for getting this moving. that would solve the problem, but we'll use this opportunity to further remove some old cruft (the old succeeded status boolean). i'll wrap this one up.

PR #1627, opened 2021-05-26
Filter events list by funded

garyhtou:
Mel double-checked and there were no issues with the `omitted` and `not hidden` queries


app/models/event.rb
```diff
@@ -18,6 +18,8 @@ class Event < ApplicationRecord
   scope :v2, -> { where.not(transaction_engine_v2_at: nil) }
   scope :hidden, -> { where.not(hidden_at: nil) }
   scope :not_hidden, -> { where(hidden_at: nil) }
+  scope :funded, -> { select { |e| e.balance_v2_cents > 0 } }
+  scope :not_funded, -> { select { |e| e.balance_v2_cents <= 0 } }
```
garyhtou: Im having trouble adding a `funded` and `not_funded` scope to Event because `select` returns an array rather than an active record relation. this means it can't be chained

image

any ideas on how to fix?

motdotla:
I'm going to start closing drafts that are on the older side. We can always re-open if deemed necessary.

PR #1637, opened 2021-05-27
Use new transaction system for admin transaction volume

maxwofford:
I'm creating this to surface a bug, but this is not the full fix.

From an up-to-date backup this computes $12mil in transactions:
Screen Shot 2021-05-26 at 20 10 21

One thing I found that contributes (but not the full story) is an event for transactions without an event:
Screen Shot 2021-05-26 at 20 11 43

That's only 0.3mil accounted for of $9mil off, so I'm not sure what the full fix is.


motdotla:
What are the NoEvent transactions? They were transactions that no one owns - not even HQ?

maxwofford:
There's an event (id 999) called NoEvent. I assume this is being used now instead of having TXs that aren't assigned to events (ie. stripe topups). This is a new thing so I don't have context on it.

garyhtou:
> There's an event (id 999) called NoEvent. I assume this is being used now instead of having TXs that aren't assigned to events (ie. stripe topups). This is a new thing so I don't have context on it.

yeah, this system replaced the `is event-related` checkbox on the old engine

motdotla:
This looks ready now, right Max? I believe someone marked event 999 to now omit_stats. So this should return the value expected.

PR #1639, opened 2021-05-28
UI for Bank Connect continue

garyhtou:
bank logo should be red on production (similar to login page)

image

image


motdotla:
great work Gary!

PR #1678, opened 2021-06-07
Add external boolean to Partners table

garyhtou:
closes #1676

run after merging
```ruby
p = Partner.where(slug: 'bank').first
p.external = false
p.save!
```

### DIFF
--- a/db/migrate/20210607030039_add_external_to_partners.rb
+++ b/db/migrate/20210607030039_add_external_to_partners.rb
@@ -1,0 +1,5 @@
+class AddExternalToPartners < ActiveRecord::Migration[6.0]
+  def change
+    add_column :partners, :external, :boolean, null: false, default: true
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20210607030039_add_external_to_partners.rb
```diff
@@ -0,0 +1,5 @@
+class AddExternalToPartners < ActiveRecord::Migration[6.0]
+  def change
+    add_column :partners, :external, :boolean, null: false, default: true
```
motdotla, line 3: well done here with the default.

can you add a spec to the partner_spec.rb that shows the external defaults to true?

motdotla (changes requested):
one small addition. add a spec to partner_spec.rb showing that external defaults to true

db/migrate/20210607030039_add_external_to_partners.rb
garyhtou: done!

image

motdotla:
merging. i will update Bank partner to false once deployed.

PR #1691, opened 2021-06-13
Admin reorganize dashboard + a few tweaks

garyhtou:
- Reorganize links on home page
- Move old v1 links to their own page
- On admin ledger, by default, show txs that require mapping
- Tooltips for event's status emoji on index page

closes #1682


garyhtou:
Imma ask for Mel's feedback before we merge

PR #1692, opened 2021-06-13
Automate setting google workspace verification key

garyhtou:
## Current ops workflow
1. Install [Bank Ops Plugin](https://github.com/garyhtou/HCB-Ops-Plugin)
2. Set it up with an API key
3. Go to Google Workspace process page
4. Copy verification key from plugin's injected element and paste it into field (the plugin gets the key from G-Verify)
image
5. Hit the button ("Update verification key")

## This PR

**Simpler ops workflow**
1. Just click the approve button lol
image

**How it works**
1. When the approval button is clicked, it calls an aysnc job (`GSuiteJob::SetVerificationKey`)
2. The async job GETs the verification key from [G-Verify](https://github.com/garyhtou/G-Verify) (same server/service the plugin uses) in `GSuiteService::GetVerificationKey`
3. Uses the existing `GSuiteService::Update` service to update the key and notify organizers

## Future goal

Remove dependency on G-Verify and do all the oauth google webmaster and key fetching within this bank repo


motdotla (approved):
cool feature! coordinate with @MelSmith104 on how to use it.

PR #1696, opened 2021-06-16
Event partners donation page

garyhtou:
Slug: `/event_name/partner_donations`

image

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -175,18 +175,35 @@ class EventsController < ApplicationController
     authorize @event
 
     relation = @event.donations.not_pending
+
+    @stats = {
+      deposited: relation.deposited.sum(:amount),
+      in_transit: relation.in_transit.sum(:amount),
+      refunded: relation.refunded.sum(:amount)
+    }
+
     relation = relation.in_transit if params[:filter] == "in_transit"
     relation = relation.deposited if params[:filter] == "deposited"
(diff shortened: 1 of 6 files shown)


app/controllers/events_controller.rb
```diff
@@ -175,18 +175,35 @@ def donation_overview
     authorize @event
 
     relation = @event.donations.not_pending
+
+    @stats = {
+      deposited: relation.deposited.sum(:amount),
+      in_transit: relation.in_transit.sum(:amount),
+      refunded: relation.refunded.sum(:amount)
+    }
+
```
garyhtou, line 184: **Patch:**
I moved the overview stats calculation before the filters since these probably shouldn't be affected by the filters

app/views/events/_nav.html.erb
```diff
@@ -29,6 +29,7 @@
     <%= dock_item 'Donations',
       'Receive donations online',
       'support',
+      @event.partner.external? ? event_partner_donation_overview_path(event_id: @event.slug) :
       event_donation_overview_path(event_id: @event.slug) unless @event.unapproved? %>
```
garyhtou: Should all of these `unless @event.unapproved?` in the event nav be `unless !@event.approved?` since pending events shouldn't have access, right?

motdotla: yes, they should be now. go ahead and switch that.

motdotla (changes requested):

motdotla (approved):

PR #1703, opened 2021-06-16
Calculate ACH & Checks stats before filters

garyhtou:


motdotla:
@garyhtou is this ready for review?

PR #1716, opened 2021-07-04
Create User Service

garyhtou:
dependent on hackclub/api#447 (this PR will require some more testing after merging hackclub/api#447)

---

Here's how this works:
1. Using an Admin account, we call Hack Club's API to create a user. This returns their access_token.
2. With their access_token, we get their user information from Hack Club's API.
3. We create a User for them in Bank and store some information provided by the api (access_token, admin_at, and api_id).
4. We skip creating an OrganizerPositionInvite, instead, we directly create an OrganizerPosition. (OrganizerPositions are usually created when a user accepts an invite)


motdotla:
nice work here, Gary. I ended up re-creating and borrowing from it, but I also re-implemented it (mainly so I could start getting back into the swing of things). Closing this as a result.

PR #1717, opened 2021-07-04
Create a user for the partner org during connect_finish

garyhtou:
dependent on #1716


motdotla:
this is another that I borrowed code from and ended up re-implementing

PR #1718, opened 2021-07-04
Generating Login Tokens and authenticating with Login Token URLs

garyhtou:
dependent on #1716 for further testing.

TODO: finish specs for `login_token_url`

---

When I added `LoginToken` and migrated, it also made some un-related changes to `schema.rb`. should those be merged?


app/models/login_token.rb
```diff
@@ -0,0 +1,26 @@
...
+    loop do
+      self.token ||= SecureRandom.alphanumeric(8) #2.2 tredecillion combinations
+      # if we happen to need more, we can recycle previously expired codes
+
+      # repeat until token is unique
+      break unless LoginToken.find_by(token: token)
+
+      # below is what it would look like to recycle previously expired codes
+      # break unless LoginToken.active.find_by(token: token)
+    end
```
motdotla: nice code here

PR #1720, opened 2021-07-07
GSuite domain uniqueness only on not_deleted

maxwofford:
This doesn't implement the `deleted` state mentioned in #1715, but I believe it will fix the error users are running into when trying to create a GSuite for `example.com` after (soft) deleting a previous GSuite for `example.com`.

I'm not sure how to safely test this locally w/o creating stuff on our Google account, so this is untested.


maxwofford:
Forgot to mention, I was debugging some stuff locally & wrote this. Implementing the `deleted` state in #1715 would still need the validation to be updated (this PR), & involve updating the `not_deleted` scope to use AASM instead of the currently used `deleted_at` field.

garyhtou:
FYI, this PR only fixes the domain uniqueness bug. it doesn't implement the deleted state into aasm

PR #1756, opened 2021-07-29
Add Lockbox for Partner `stripe_api_key`

garyhtou:
need some help here :smile:
not sure what the problem is. it seems like lockbox is installed correctly
image

### TODO:
https://github.com/ankane/lockbox#migrating-existing-data


motdotla:
I think you just need to restart your server. try that.

motdotla (approved):

PR #1757, opened 2021-07-29
Configure Event deliver webhook retry

garyhtou:
A total of 16 attempts over a span of 3 days with an exponential back off

#1675


motdotla:
great work here

PR #1818, opened 2021-09-29
GitHub Codespace configuration

kunalbotla:


garyhtou:
For restoring the db (use this one instead)
```
pg_restore --verbose --clean --no-acl --no-owner -h db -U postgres -d bank_development latest.dump
```

<br/>

For the master key, create the file `config/master.key` with the master key in it (what you can do is: if the file doesn't already exist, prompt the user for the key, then pipe it into the file)

<br/>

For attaching to the docker's bash you can use `docker exec -it bank_web_1 /bin/bash`, but idk how useful it will be in this script bc you can probably chain all the commands together


also, if possible, make it so that we can run the script multiple time. This is useful for when docker just decides to screw up and you need to redo ur installation. you would probably need to add a couple of things such as removing the old dump file before downloading the new one (`rm latest.dump`)

maxwofford:
@kunalbotla I just rebased this so I can merge. great job on this

maxwofford:
For posterity, I tested this in a codespace over a call with Kunal & it worked

PR #1825, opened 2021-10-01
Bank can check own version hash

maxwofford:


maxwofford:
@eilla1 @kunalbotla @garyhtou if bank can know details like what version of deploy it's at & when the current version was deployed at, is there anything cool we can do with it?

kunalbotla:
i mean it'd be cool. it used to have it before this redesign iirc

exu3:
is the `Build 3d3b438 from about 6 hours ago.` different from what you're describing? or are you referring to the SemVer version numbers?
Screen Shot 2021-10-01 at 6 45 28 PM 

Either way I think it'd be cool to have in the footer on certain pages and maybe some flavor text too

maxwofford:
Doesn't seem like this is needed then!

PR #1828, opened 2021-10-03
Update README.md with Codespaces

kunalbotla:


kunalbotla:
@MelSmith104 & @maxwofford can you take a look at these to make sure its on brand

maxwofford:
@kunalbotla this is an internal doc so i wouldn't worry about keeping it on brand. as long as @garyhtou & @eilla1 can spin it up we're good. Bonus points if the next person who starts working on bank can also read through this unguided & setup a dev server.

kunalbotla:
awesome; wasn't sure if it seen by any partners. they both are all set and it looks to be good with that change so i shall merge & sleep

PR #1848, opened 2021-10-09
Add flag icon to 'Dispute this transaction' link

garyhtou:
image

the spacing and size of the icon seem kinda off. i'm not sure how to make it better. we might be better off without the icon


maxwofford:
@eilla1 thoughts on how this looks?

exu3:
I think the flag looks alright here but it also looks fine without the icon

maxwofford:
Merging b/c i think it helps

PR #1849, opened 2021-10-09
Fix Airbridge url for Disputes

exu3:
The table name was changed in Airtable and Airbridge, but wasn't updated here so the admin task card is showing `9999`. This PR updates the Airbridge url for the task card.


garyhtou:
Good catch! thanks :)

garyhtou (approved):

PR #1858, opened 2021-10-13
Admin Events filter by Funded vs Not Funded

garyhtou:
#1625. related to old PR #1627


app/controllers/admin_controller.rb
```diff
@@ -110,26 +110,28 @@ def events
...
-    states = []
-    states << "pending" if @pending
-    states << "unapproved" if @unapproved
-    states << "approved" if @approved
+    relation = relation.transparent if @transparent
+    relation = relation.omitted if @omitted
+    relation = relation.hidden if @hidden == 'hidden'
+    relation = relation.not_hidden if @hidden == 'not_hidden'
+    relation = relation.where(id: relation.funded_array.pluck(:id)) if @funded == 'funded'
+    relation = relation.where(id: relation.not_funded_array.pluck(:id)) if @funded == 'not_funded'
```
garyhtou: @maxwofford I FIGURED IT OUT 😄!!!! it was a shower thought where I realized that I can pluck the ids and use them in a where statement hahaha

MelSmith104:
@garyhtou this is EXACTLY what i was looking for, thank you! 

2 quick notes:
- omitted filter always goes back to "both" even though i select "not omitted" - not sure what's going on here
- the page feels slower to load than before

Awesome job! this is crazy helpful for us :)

maxwofford:
@garyhtou #1865 should speed this up

maxwofford:
Commenting here, brought up in bank sync call, this is fine to merge

maxwofford (approved):

PR #1860, opened 2021-10-14
Enforce restrictions for disputing transactions

garyhtou:
#1859

Stripe has some conditions on disputing card transactions. One of these conditions is that the transaction must not be captured more than 110 days ago — otherwise, it can no longer be disputed.

Here are the things I've enforced:
- Stripe transactions older than **90** days can't be disputed. This gives Ops a 20-day buffer (from Stripe's 110-day restriction) to work with the disputer to submit a dispute to Stripe.
- Bank account transactions older than **90** days (this number is arbitrary) can't be disputed.
  - Bank account transactions that are positive can always be disputed. This helps in the case that Ops misassigns some money (such as an incoming check)
- Donations (and Partner Donations) can be disputed at any time — disputing a donation would be refunding it to the donor.

The normal dispute flow remains unchanged for the user — besides an unnoticeable architectural change: The "Dispute this transaction" link no longer links directly to AirTable. instead, it's linked to `/hcb/:id/dispute` which then redirects to the AirTable form if it can be disputed.

## Disputing a Stripe transaction older than 90 days
Screen Shot 2021-10-13 at 8 06 38 PM

## Disputing a negative Bank account transaction older than 90 days
image

## Disputing an unsupported transaction (such as ACH)
Any user-initiated transaction is unsupported (such as ACH, Checks, etc. — PayPal and Wires are disputable since Bank recognizes those as bank account transactions)
This is a fail-safe since "Dispute this transaction" links are not shown on unsupported transactions.
image


maxwofford:
This is great! @MelSmith104-- let's sit down sometime and go over this

MelSmith104:
@garyhtou this lgtm - literally perfect - awesome job!!!!!!!

maxwofford:
Go ahead and merge when ready @garyhtou

PR #1872, opened 2021-10-20
Initial logic to make my_cards page work with v2 TX system

maxwofford:
This is still a WIP– it fixes the list of shown TXs, but it hasn't been styled yet


maxwofford:
This is a little janky in the view logic, but it fixes the following:

- /my/inbox wasn't showing a correct list of TXs (didn't include pending stripe TXs)
- /my/cards wasn't showing a correct list of TXs without receipts (still used v1 TX engine)
- event show page didn't indicate TXs that required receipts

Closes https://github.com/hackclub/bank/issues/1835

PR #1877, opened 2021-10-26
Fix codespace config migration overwrite

maxwofford:


kunalbotla:
i think this looks good; if it causes any issues for anyone we can revert.

PR #1901, opened 2021-10-27
Add PublicIdentifiable concern

maxwofford:


maxwofford:
@garyhtou any clue why these tests are failing?

app/models/event.rb
```diff
@@ -429,9 +432,10 @@ class Event < ApplicationRecord
     ZM: 248,
   }
 
-
   validate :point_of_contact_is_admin
 
+  validates :public_id, presence: true, uniqueness: true
+
```
garyhtou: tests are failing because Event models being created aren't valid. i think `validates` is looking for a `public_id` column, but it's actually a method

maxwofford: Whoops! Thanks for catching that– forgot to unstage that line

maxwofford:
Merging this for now w/ the givebutter branch b/c there aren't any db changes in it, but if you'd implement this a different way @garyhtou, lmk!

PR #1902, opened 2021-10-27
Spec out initial partnered signup model

maxwofford:


maxwofford:
@garyhtou are there any other fields you need on this model?

PR #1903, opened 2021-10-28
⚠️ WIP: Givebutter 2021 10 27 sprint

maxwofford:
In addition to #1880, also added:

- `PartnerSignup` api index and get specific object endpoints

before merge:
- fix 'webhook_url' migration error


maxwofford:
RSpec is failing, but I'm ok with that for now to get this in prod so the demo-app. @garyhtou can you make an issue for these failing tests– we should fix or remove them sometime today. Prioritize making the demo app work over the tests

PR #1910, opened 2021-10-29
Make my/cards missing receipts match the cards details page

tanatorn:
Changes the filter `missing_receipt` to `without_receipt`

Basically there's some inconsistency here where on the `/my/cards` page we show the transactions we identify as `missing_receipt` which means that we don't have the receipt and haven't marked it as lost.

They both appear with the red receipt missing icon so I just made it consistent. Another fix is to disable the red receipt missing icon when we mark transactions as  `no or lost receipt`

image

image


maxwofford:
Great catch on this!

PR #1916, opened 2021-10-30
Revert 'Fix codespace config migration overwrite'

exu3:
Both Kunal & I were having issues with the codespaces start/setup script, but reverting #1877 seems to fix it after trying locally

```
ActiveRecord::NoDatabaseError (FATAL:  database "bank_development" does not exist
):
```


kunalbotla:
I just tested in a fresh Codespace with this script and its working normally again!

image

PR #1918, opened 2021-11-01
Hacky implementation of admin accept/reject partnered signups

maxwofford:


app/policies/partnered_signup_policy.rb
```diff
@@ -9,11 +9,11 @@ def update?
...
 
-  def admin_accept?
+  def partnered_signups_accept?
     user&.admin?
   end
 
-  def admin_reject?
+  def partnered_signups_reject?
     user&.admin?
   end
```
garyhtou: prob don't need these anymore since these methods no longer exist in partnered_signup_controller

maxwofford: Pundit policies are by model (weird, i know). If these are removed validations will fail

PR #1919, opened 2021-11-01
Webhook types and signature

garyhtou:
This is what our outgoing webhooks look like:

- `hcb-signature` is a [Stripe-styled webhook verification signature](https://stripe.com/docs/webhooks/signatures)
- webhook type is under `meta`. this prevents conflict if there is a model with a "type" field

image

This pr restructures existing webhook code (to allow for types) and creates a model specific serializer (we're moving away from endpoint specific serializers)


garyhtou:
https://github.com/hackclub/bank/issues/1892
https://github.com/hackclub/bank/issues/1883

maxwofford:
This is great Gary! Go ahead and bring it into the givebutter branch whenever you're ready

PR #1926, opened 2021-11-01
Only expire session cookies after 30 days

maxwofford:


garyhtou (commented):
30 days would be nice! but at the same time, these accounts manage money. I'm not sure where that balance of security and convenience should be. my banks literally log me out after like 30 minutes without use haha.

i'm cool with 30 days, but any thoughts/inputs from everyone else?

maxwofford:
I'm going to make the executive decision to bring this in for the time being. We should get a security review on hcb soon (esp now that gb partnership is going live).

maxwofford:
Also, for posterity, CI is testing b/c of #1903. I'm merging w/ failing tests b/c it's the same tests that failed in #1903

PR #1927, opened 2021-11-01
Branding: change logo from svg to png

exu3:
The SVG for the logos don't seem to be working in chrome. The badges are SVGs and they work fine, but I couldn't tell what was wrong with the logo SVGs. This PR changes the preview image for logos to PNG so it works in all browsers, but lmk if there's a better way fix the SVG not rendering issue 
Screen Shot 2021-11-01 at 8 38 54 AM


garyhtou:
amazing work @eilla1!!

PR #1945, opened 2021-11-02
Check Bank Account Copy Edits

kunalbotla:
#1932


kunalbotla:
@MelSmith104 + @hackliv can you make sure that the updated copy here for step 4 of checks is now correct?

hackliv:
it looks the same on my end, but maybe it will change once a new check
comes in?

kunalbotla:
Checked with Liv over Slack and confirmed accounts mentioned look correct

garyhtou:
Is it `checking`? cause i think both fs main and fs operating are checking accounts. (i find it slightly confusing because we don't use that term anywhere else — usually just call it fs main/operating)

and yeah, checks are kinda is a broken state right now. one of the 4ish ('ish' bc i don't really remember haha) transactions it generates needs to be manually mapped

hackliv:
@Gary it is checking. That step is confusing bc when you do a transfer in
bill pay classic, we are using the checking accounts and there is not an
option that says main/operating.

garyhtou:
cool, gotcha. thanks!

garyhtou (approved):

PR #1947, opened 2021-11-02
Fix bug where it was using the wrong id when referencing stripe cards

clarencejychan:
Bug fix for #1935 

Transparency mode
image

Logged In w/ No Cards
image

Logged In w/ Cards
image


clarencejychan:
The `XXXX` seems to be something to do with formatted card numbers but this isn't in the scope of this change. @tanatorn  is this something you need to address?

clarencejychan:
Looks like its coming from `def stripe_obj` under `sripe_card`. Seems like it's not related to this change though.

exu3 (commented):
Screen Shot 2021-11-02 at 11 18 59 AM
For some reason I'm getting the heroku error on the cards page in staging (maybe it's my school's internet filters?idk) but it seems to be working fine in development. Thanks for fixing it!
Screen Shot 2021-11-02 at 11 25 02 AM

maxwofford (approved):
Thanks for fixing this!

PR #1950, opened 2021-11-02
Simplify generate login url for now

garyhtou:
**Untested**

also moved swagger params to body (although they work anywhere)


garyhtou:
#1958

PR #1958, opened 2021-11-02
Login serializer just goes to login page for time being #1951

maxwofford:


maxwofford:
I was able to send an invite to another user by hitting the login url with their email.

Screen Shot 2021-11-02 at 20 37 55

PR #1836, opened 2021-10-05
Admin Side Country Codes

kunalbotla:

PR #1960, opened 2021-11-03
Cleans up country selection for Admin

tanatorn:
Country selector is actually usable now.
Closes #1923

image


kunalbotla:
hey Tan, one concern I have on this is that I remember @maxwofford and I had some reason for doing it this way in #1836. I don't remember the exact reason so running this by Max to make sure that it doesn't conflict with something else might something to keep in mind.

maxwofford:
@tanatorn thanks for this! I'd love to get the country code list out of the Events model enum in the future, and this mapping will help us get there.

maxwofford (approved):
Go ahead and rebase+merge!

PR #1982, opened 2021-11-07
Fix the Wrapping on Mobile

clarencejychan:
image
image

Fixes #1981


clarencejychan:
Didn't see @eilla1 's comments on the issue. Going to checkout FF and Safari (although not sure how to do it with Safari) maybe browserstack?

maxwofford:
We have a browserstack account if you need it– DM me on Slack for access

clarencejychan:
I think we're actually good for now since @eilla1  checked out the Safari version already but I will in the future. Thanks!

maxwofford (approved):
LGTM!

PR #1983, opened 2021-11-07
Fix flavor text pills in safari

exu3:
Removing the `width: min-intrinsic` css property seems to fix the narrow-pill problem in Safari

https://user-images.githubusercontent.com/72365100/140636232-06d5a622-7476-43df-9776-9c336cd4153f.mov


clarencejychan (approved):
Thanks for helping out with the fix on Safari.

Could we just add a SS to see what it looks like on mobile?

exu3:
(in development mode, but i just changed the text to test out the pill width)
Screen Shot 2021-11-07 at 12 32 02 AM

mobile, in production:
Screen Shot 2021-11-07 at 12 34 11 AM

maxwofford:
This looks great @eilla1! Thanks for the fix

PR #1968, opened 2021-11-04
[BETA] Add Free Domains Perk

exu3:
This PR adds the Free Domains promo card in the Perks tab as well as the admin dashboard. 

I'm not sure how the background image on the promo card fits in with the rest of the page, but here's the [commit](https://github.com/hackclub/bank/pull/1968/commits/e984dcf319e5bc5bc36bf1138e4728ac68fb2646) to revert if it's better without the bg.

I also moved the inline `style`s (on the PVSA card) to a scss file for the promos page which also includes styles for the new Free Domains card.

Closes #1931 

Screen Shot 2021-11-03 at 9 45 27 PM
Screen Shot 2021-11-03 at 9 46 06 PM

PR #1984, opened 2021-11-07
Free Domains 🚀 

exu3:
This is the PR that addresses all the comments / revisions from #1931 and beta testing with users.


exu3:
(No idea why the commit message is formatted weirdly for 1d097a79cc8afb4f1f1552aef816bffa0c820495) It's meant to say "Add `id` attributes to perks cards so that we can link to them directly"

kunalbotla:
marking as ready for review so it builds properly; do not merge yet

MelSmith104:
This looks really awesome! couple copy changes --> remove the TLD section from the card, this is addressed in the form and please change the copy around Google Workspace. 
Other than that - this is awesome!! amazing job :) i love it and am really excited to see it ship

app/views/events/promotions.html.erb
```diff
@@ -6,7 +6,7 @@
   Promotions &amp; perks
 </h1>
 
-<div class="card pb0 mb3">
+<div class="card pb0 mb3" id="pvsa">
```
maxwofford: Are these IDs used anywhere? I saw them also on domains, replit, sendy, & github

app/assets/stylesheets/components/_promos.scss
```diff
@@ -30,28 +30,6 @@
  }
 
  html[data-dark="true"] {
-   .promo-pvsa {
```
maxwofford: Was this intentionally deleted?

maxwofford (approved):
This is awesome. I left a couple comments on stuff that seemed like it wasn't for this PR, but I'm comfortable with this getting merged.

I made a couple changes to wording (typos) in here as well as added restrictions to when we show the card (ie. hide the perk for users that already have a gsuite setup).

I made a pass on the airtable form copy to simplify questions & bring the prefilled questions to the bottom of the form.

You can test the flow by going to an event with US set for the country code & no workspace, and viewing the google workspace overview + perks pages. Then view the same 2 pages on an org that already has a workspace or doesn't have their country field set.

Go ahead and merge when ready.

app/assets/stylesheets/components/_promos.scss
kunalbotla: I believe so. @eilla1 can you confirm?

exu3: Screen Shot 2021-11-10 at 11 11 30 AM
Yeah that added with #1968 but I realized that the dark mode styles made it look disabled. Thoughts on how it looks?

maxwofford: Totally fine! Just wanted to make sure it was intentional. You're totally right it looks disabled, so I like that we're removing it. It's a little bright in dark mode now though, so it'd be cool to see it replaced with something else. Don't block on that for this PR

app/views/events/promotions.html.erb
exu3: On the google_workspace page I was going to use that to directly link directly to the Domains card but I didn't end up doing that. I could remove these. Would it be better to remove these IDs since they're not used?

PR #1986, opened 2021-11-08
Fixes donation amount

tanatorn:
New donations now don't transition to `Deposited` until the fees are reimbursed. This should solve some confusion about the donation page not matching up with the balance.

To test, I went and two transaction states for John and Josh locally to `:in_transit`  then ran 
```
irb(main):044:0> DonationService::Nightly.new.run
```
in irb to make sure that we don't update it to deposited until everything's settled.

I also added some convenience functions in canonical_transaction_grouped and the made the `linked_object` work with Donations created on the TransactionEngineV2
 
image

### DIFF
--- a/app/models/canonical_transaction.rb
+++ b/app/models/canonical_transaction.rb
@@ -178,9 +178,9 @@ class CanonicalTransaction < ApplicationRecord
   end
 
   def fee_reimbursement
-    return linked_object.fee_reimbursement if linked_object.is_a?(Donation)
+    return linked_object.fee_reimbursement if linked_object.is_a?(Donation) || linked_object.is_a?(Invoice)
 
-    nil # TODO
+    nil
   end
 
   def check
--- a/app/models/canonical_transaction.rb
+++ b/app/models/canonical_transaction.rb
@@ -178,6 +178,8 @@ class CanonicalTransaction < ApplicationRecord
   end
 
   def fee_reimbursement
+    return linked_object.fee_reimbursement if linked_object.is_a?(Donation)
+
     nil # TODO
   end
 
--- a/app/models/donation.rb
+++ b/app/models/donation.rb
@@ -194,7 +194,7 @@ class Donation < ApplicationRecord
   end
 
   def fee_reimbursed?
-    amount_settled == amount
+    fee_reimbursement.canonical_transaction.present?
   end
 
   private
--- a/app/models/donation.rb
+++ b/app/models/donation.rb
@@ -189,6 +189,14 @@ class Donation < ApplicationRecord
     remote_donation[:charges][:data][0][:refunded]
   end
 
+  def amount_settled
+    canonical_transactions.sum(:amount_cents)
+  end
+
+  def fee_reimbursed?
+    amount_settled == amount
+  end
+
   private
 
   def raw_pending_donation_transaction
--- a/app/services/donation_service/nightly.rb
+++ b/app/services/donation_service/nightly.rb
@@ -8,6 +8,7 @@ module DonationService
 
         next unless cpt
         next unless cpt.settled?
+        next unless donation.fee_reimbursed?
 
         raise ArgumentError, "anomaly detected when attempting to mark deposited donation #{donation.id}" if anomaly_detected?(donation: donation)
 
--- a/app/services/transaction_engine/syntax_sugar_service/linked_object.rb
+++ b/app/services/transaction_engine/syntax_sugar_service/linked_object.rb
@@ -7,6 +7,7 @@ module TransactionEngine
 
       def initialize(canonical_transaction:)
         @canonical_transaction = canonical_transaction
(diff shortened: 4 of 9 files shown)


maxwofford (commented):
This is a good start, but the total event balance shown to an event is off. If i receive a donation for $1, my transaction list shows a $1 deposit, but my balance will only show $0.68 until the fee refund comes in. I'd like the balance to only go up to $1 once the fee refund and the payout have arrived.

maxwofford:
[Here](https://hackclub.slack.com/archives/GBXBFEED9/p1636389051038900) is an example of something that could confuse the user mentioned in https://github.com/hackclub/bank/pull/1986#pullrequestreview-800190994. It's an admin in the Slack comment, but users don't have access to that transaction-specific breakdown.

tanatorn:
@maxwofford I think the bug here is actually the fee reimbursement engine hasn't been running for awhile so there's been no reimbursement. Since there's no reimbursement, they _are_ displaying the correct value. @garyhtou said he'll get that reimbursement portion working - I'll stop working on this in the mean time to see if the issue self resolves.

tanatorn:
For context: https://hackclub.slack.com/archives/C026RKHLPNJ/p1636518674031100?thread_ts=1636517056.026700&cid=C026RKHLPNJ

maxwofford:
@tanatorn you're totally right fee refunds are delayed, but the underlying issue is still there– our UI expects either (no payout nor fee refund) or (payout and refund). It doesn't handle (payout paid but fee refund pending).

Right now because fee refunds are delayed the issue is worse b/c the UI is confusing for a week, but regularly people will try using bank by donating a $1 to themselves, and when they see the funds come into their account their total balance shows $0.32 right above a TX history that just shows a $1 donation successfully paid. That confusing state can stay there for a day or two.

tanatorn:
PTAL @maxwofford

maxwofford:
Requesting @garyhtou's review before merge for his experience with this part of hcb codes & transactions v2

app/models/canonical_transaction.rb
```diff
@@ -178,6 +178,8 @@ def invoice_payout
   end
 
   def fee_reimbursement
+    return linked_object.fee_reimbursement if linked_object.is_a?(Donation)
+
     nil # TODO
   end
```
garyhtou: I believe `Invoice` also has fee reimbursements.

also, extra `nil # TODO` on line 183

app/models/donation.rb
```diff
@@ -189,6 +189,14 @@ def remote_refunded?
     remote_donation[:charges][:data][0][:refunded]
   end
 
+  def amount_settled
+    canonical_transactions.sum(:amount_cents)
+  end
+
+  def fee_reimbursed?
+    amount_settled == amount
+  end
```
garyhtou: Edge case: when a donation is refunded, the `amount_settled` would equal 0. In that case, `fee_reimbursed?` would return `false`. Should that be the case?

I think `true` might make more sense bc it had a Fee Reimbursement. open to thoughts on this tho

---

**preferred method:**

a more robust way to do it may be to look at the donation's `fee_reimbursement` relation, and check if `fee_reimbursement.canonical_transaction` is nil.

there is a `completed?` method in FeeReimbursement model, but it's using the v2 engine. would be best to update that to use v2

app/services/donation_service/nightly.rb
```diff
@@ -8,6 +8,7 @@ def run
 
         next unless cpt
         next unless cpt.settled?
+        next unless donation.fee_reimbursed?
```
garyhtou, line 11: how about something like `donation.fee_reimbursement.settled?`

related to comment on Donation model

app/services/transaction_engine/syntax_sugar_service/linked_object.rb
```diff
@@ -28,8 +28,13 @@ def run
...
           return likely_bank_fee if outgoing_bank_fee?
 
-          nil
         end
+        hcb_code = HcbCodeService::FindOrCreate.new(hcb_code: @canonical_transaction.hcb_code).run
+        if hcb_code.donation?
+          return hcb_code.donation
+        end
+
+        nil
```
garyhtou: combine with `likely_donation`?

garyhtou (changes requested):
good work so far!! couple organizational and design changes

app/services/donation_service/nightly.rb
tanatorn: i made `donation.fee_reimbursed?` a convenience method for accessing fee_reimbursment.ct

tanatorn:
PTAL @garyhtou

app/models/donation.rb
garyhtou: whoops, typo: but it's using the v1 engine*

garyhtou (approved):
good to go! thanks for making those edits

PR #2000, opened 2021-11-09
Fix FAQ anchor links

cjdenio:
This PR fixes a tiny bug where FAQ items aren't highlighted when they're clicked.


maxwofford (approved):
Thanks for adding this!

PR #2003, opened 2021-11-09
Disable remote form submission

cjdenio:
Disables remote form submission, which breaks a whole bunch of forms (Turbo replaces this behavior, see https://github.com/hotwired/turbo-rails/blob/main/UPGRADING.md) (thanks @eilla1 for catching this)


exu3 (approved):
Going to merge this cause someone just emailed reporting issues with the donation form.
Thanks for the quick fix Caleb!

PR #2004, opened 2021-11-09
Revert "Fix Turbo(links) integration"

maxwofford:
Reverts hackclub/bank#1985

This is to test the effects on the donation pages that recently broke


maxwofford:
Got what I needed out of this & implemented a fix in https://github.com/hackclub/bank/pull/2005

PR #2006, opened 2021-11-09
Reduce size of 'honk club' flavor text

maxwofford:
Previously:

ezgif com-gif-maker 1

Now it's the same size as the text:

Screen Shot 2021-11-09 at 10 36 22


exu3 (approved):
honk LGTM!

PR #2009, opened 2021-11-09
Fix: change `country_name` to `country`

exu3:
Previously, the country field wouldn't update when the "Update" button was clicked because it used `country_name` instead of `country`. https://hackclub.slack.com/archives/GBXBFEED9/p1636488971070100

now:

https://user-images.githubusercontent.com/72365100/141008594-3ccaae63-07ab-48e2-b259-37170b517827.mov


tanatorn (approved):
Looks like i forgot to change that, thanks for the quick catch!

PR #2010, opened 2021-11-10
Opt-in certain links to Turbo

cjdenio:
This PR brings back some Turbo-powered goodness (hopefully less, er, colorful than last time).

Notes:
- Forms still submit normally, for now
- Only links that lead to "ordinary" pages (without special JavaScript behavior) have been opted-in


maxwofford:
Seems like the double loading messes with light/dark mode & will flash the user. Also double-loads the flavor text.

https://user-images.githubusercontent.com/5891442/141154678-079eab3e-96c8-49c7-8770-e07cbf2e25ad.mov

maxwofford (changes requested):
https://github.com/hackclub/bank/pull/2010#issuecomment-965530292

cjdenio:
Hmm, I can't reproduce the light/dark mode issue 🤔 Did you click some magic series of buttons to get it to break?

As for flavor text, it seems to be a (less desirable in this case) side-effect of Turbo's cache:
> If possible, Turbo Drive will render a preview of the page from cache immediately after the visit starts. This improves the perceived speed of frequent navigation between the same pages.

maxwofford:
Looks like the flash is caused by a local extension on my machine

PR #2016, opened 2021-11-10
Add event name and id to the check mailing address

garyhtou:
This will make it much easier to map incoming checks


exu3 (commented):
I think the proper way to use c/o would be 
```
Project name
c/o The Hack Foundation
14234 Address Rd #38423
everything else
```
^according to this https://www.wikihow.com/Address-an-Envelope-in-Care-of-Someone-Else

The mailing address is also on the `receive_check` documentation page so it might make sense to change it there as well.

garyhtou:
oof, yeah i think ur right. please feel free to fix

garyhtou:
thanks for updating that!

MelSmith104:
this looks good to go for me but the concern i have is that we've run into issues before with printed checks that ran out of room trying to get the entire c/o + the Hack Foundation on the line or orgs just not adding that part because they see their name + our address. we dropped using the c/o and switched to requesting orgs reach out to us back in april / late march and ive yet had a check i couldnt map. i think this is something we can leave as is for now and keep requesting orgs reach out to us / put their name in the memo

PR #2018, opened 2021-11-11
Admin events country filter

exu3:
Adds a country filter on the `/admin/events` page.


app/controllers/admin_controller.rb
```diff
@@ -177,6 +177,7 @@ def events
     @omitted = params[:omitted].present? ? params[:omitted] : "both" # both by default
     @funded = params[:funded].present? ? params[:funded] : "both" # both by default
     @hidden = params[:hidden].present? ? params[:hidden] : "both" # both by default
+    # @country = params[:country].present? ? params[:country] : "All" # all by default
```
garyhtou: use lower case `all`. just to keep it consistent with the others

app/controllers/admin_controller.rb
```diff
@@ -189,6 +190,7 @@ def events
     relation = relation.not_hidden if @hidden == "not_hidden"
     relation = relation.funded if @funded == "funded"
     relation = relation.not_funded if @funded == "not_funded"
+    # relation = relation.country(@country) if @country == "country"
```
garyhtou: I'm assuming `@country` will be the abbreviation such as `US`. If so, I think it would need to get converted to an integer using the country enum in Event. Then, you can do something like:

```ruby
@country_id = somehow_convert_it_lol(@country)
relation = relation.where(country: @country_id) if @country == "country"
```

---

preferably, have the form select values be the country integers. this way @country is an integer and you can just do:
```ruby
relation = relation.where(country: @country) if @country == "country"
```
it makes the code a tad bit cleaner

app/models/event.rb
```diff
@@ -21,6 +21,7 @@ class Event < ApplicationRecord
   scope :omitted, -> { where(omit_stats: true) }
   scope :not_omitted, -> { where(omit_stats: false) }
   scope :hidden, -> { where("hidden_at is not null") }
+  # scope :country, -> { where(country: :first) }
```
garyhtou: not needed!

app/views/admin/events.html.erb
```diff
@@ -51,6 +51,11 @@
       Funded
       <%= form.select :funded, [['Both', 'both'], ["Not Funded", 'not_funded'], ['Funded', 'funded']], :selected => @funded %>
     <% end %>
+    |
+    <%= form.label :country do %>
+      Country
+      <%= form.collection_select :country, Event.countries.map{|c| [c.first]}, :first, :last, { include_blank: 'All'}, :selected => @country %>
```
garyhtou: `collection_select` might not work cause it takes in a Collection of models

garyhtou (commented):
good start!

exu3:
how do i make a preview deploy? the previous one is inactive because of the 5 day limit, but i can't seem to make a new deploy for this PR

garyhtou:
just sent you an invite for the heroku bank pipeline!

exu3:
hmm where do i find the invite?

garyhtou:
ella@hackclub.com

lmk if you can't find it

exu3:
Right now, the filter doesn't filter. It gets reset when the search button is clicked.

https://user-images.githubusercontent.com/72365100/144678154-bfdfde38-682d-474f-8600-d531c2d158c2.mov

garyhtou:
the issue is most likely that the selected country isn't being passed all the way through. I'll take a look at this later today!

app/views/admin/events.html.erb
```diff
@@ -51,6 +51,11 @@
       Funded
       <%= form.select :funded, [['Both', 'both'], ["Not Funded", 'not_funded'], ['Funded', 'funded']], :selected => @funded %>
     <% end %>
+    |
+    <%= form.label :country do %>
+      Country
+      <%= form.select :country, Event.countries, { include_blank: 'all'}, :selected => @country %>
```
garyhtou: Try moving the `selected` into the hash!

garyhtou:
btw, these aren't alphabetical
image

app/controllers/admin_controller.rb
```diff
@@ -189,6 +190,7 @@ def events
     relation = relation.not_hidden if @hidden == "not_hidden"
     relation = relation.funded if @funded == "funded"
     relation = relation.not_funded if @funded == "not_funded"
+    relation = relation.where(country: @country) if @country == "country"
```
garyhtou: we only want to filter by country if `country` is not `all` (If `country` != `all`, then we can filter by `@country`). lmk if this helps!

exu3: That helps a lot; thank you so much Gary!

garyhtou (approved):
Yayyy good work! This works. This is cool to merge, but it would be nice if the countries were listed in alphabetical order.

It seems like it's getting it's ordering from the emum
https://github.com/hackclub/bank/blob/d8e55c3a37c4353eca6b158c0046996210a71b66/app/models/event.rb#L186-L436

I'm not sure why the enum's not in order (where is this getting its ordering/numbering from?)

exu3:
If I manually alphabetize the enum, would that break anything? Some of the countries would just have a different corresponding # value.

garyhtou:
> If I manually alphabetize the enum, would that break anything? Some of the countries would just have a different corresponding # value.

See if we can change the order without reassigning values. instead, only reorder the keys. (this might work if the ordering is based on alphabetizing the keys. But, if it's based on the value, then it probably won't work)

^ if that doesn't work, then we attempt sorting when creating the form select (somewhat similar to `CUSTOM_SORT` in Event)

exu3:
Reordering the keys seems to do it! The dropdown is in alphabetical order now, with US, CA, and IN at the top.

exu3:
Screen Shot 2021-12-08 at 12 39 03 PM

PR #2026, opened 2021-11-13
Prefill phone number on application

exu3:
Prefills the phone number field on the Bank application hack.af/bank-apply


garyhtou (approved):
Good idea! LGTM

PR #2027, opened 2021-11-14
Fixes circular dependency on syntax sugar

tanatorn:
Fixes this error: 
https://heroku-app97991095.airbrake.io/projects/288439/groups/3142086725441650381?tab=overview

I tested it by running:
`::TransactionGroupingEngine::Nightly.new(start_date: '2021-10-01').run` before and after my change locally to make sure that it does the exception is gone.


garyhtou (approved):
cool!

PR #2036, opened 2021-11-16
Fix z-index on export button toggle

exu3:
Fixes #2034

I tested it out on the other pages that use the same `menu__toggle` class and it doesn't break anything :)

Screen Shot 2021-11-15 at 5 45 01 PM


garyhtou (approved):
Thank you!!

PR #2043, opened 2021-11-18
SMS Login

tanatorn:
Depends on https://github.com/hackclub/api/pull/454

This is the second portion of the SMS login. There's probably a few UX things that we can fix/improve, but this PR was getting pretty big so we can do that in a follow up PR.

## Enrolling

Note: the video shows a bug where it would say "you've successfully enabled blah blah" earlier, it's already fixed.


maxwofford:
I just approved hackclub/api#454.

Can users log in with email code if SMS is enabled on their account? My phone has died before & it sucks being locked out of everything that requires SMS code. Haven't talked to users about this, but I wonder what their preferences are?

tanatorn:
@maxwofford currently, that's not possible. I'm gonna add a gate behind this feature for admins only for now and then we can add more a "preference" option as well before we make it widely available

tanatorn:
PTAL @maxwofford  I ended up gating the features to admin only for now and i'll work on the other stuff in a following PR

maxwofford (approved):
Great! Go for merge, I'm excited to try it out!

PR #2045, opened 2021-11-19
Add title attribute to  org `status` emojis

exu3:
This PR adds a `title` attribute so that the definition of the emoji will show up on hover. 

The other day Dev asked me what the emojis meant so I had to look it up in the codebase, but I think this would make it more convenient if any admin ever wondered or forgot what the emojis meant. 

Screen Shot 2021-11-19 at 1 27 44 AM


garyhtou (approved):
good idea!

PR #2049, opened 2021-11-21
Multiple User Sessions

clarencejychan:
1. Create user sessions table
2. Add user session into the table on sign in
3. Destroy a particular user session on sign out depending on the encrypted cookie
4. On sign in check if a session is valid and `nil` otherwise


garyhtou (changes requested):
Love this! I do have a couple of things.

**old implementations of user session tokens:**
- https://github.com/hackclub/bank/blob/52f885a815f02889d043287c1f728c0dfc227a72/lib/admin_constraint.rb
This is used to authenticate admins for Sidekiq
- https://github.com/hackclub/bank/blob/52f885a815f02889d043287c1f728c0dfc227a72/app/controllers/api/application_controller.rb#L20-L28 it's used by Bank Connect (Givebutter?) API to automatically sign-in users via a magic link


**cool additional features to think about:**

- Let users invalidate all other sessions
- Fingerprint device and add that info to user sessions
  - email on a new device (would probably require soft deleting sessions or a new table for devices)

clarencejychan:
Thanks @garyhtou  

What's the best way to test out those two? I tested out the prior changes with the actual app.

Also I noticed that the other `sign in` method has a `30 day limit` whilst the other was `1 day`. What's the preferred length of time for auth in general?

cc: @maxwofford

clarencejychan:
Invalidating sessions is a good one, created a ticket for it under #2050
For fingerprinting could probably just add an extra column to the `sessions` table.

app/helpers/sessions_helper.rb
```diff
@@ -8,9 +8,8 @@ def impersonate_user(user)
   # DEPRECATED - begin to start deprecating and ultimately replace with sign_in_and_set_cookie
   def sign_in(user, impersonate = false)
     session_token = SecureRandom.urlsafe_base64
```
tanatorn: should we update the `sign_in_and_set_cookie` function as well?

tanatorn: ah looks like gary beat me to it

garyhtou:
> Also I noticed that the other `sign in` method has a `30 day limit` whilst the other was `1 day`. What's the preferred length of time for auth in general?

Let's go with 30 days because of the comments in https://github.com/hackclub/bank/pull/1926

garyhtou:
> What's the best way to test out those two? I tested out the prior changes with the actual app.

Don't worry about testing out `sign_in_and_set_cookie!` because that portion of the code is still WIP.

For Sidekiq, verify that only admins can access https://bank.hackclub.com/sidekiq

garyhtou (approved):

maxwofford (approved):
This is awesome, go for merge! Love that we have #2050. Can you also add fingerprinting to it or open another ticket? We'll want that for showing users extra info about their other sessions, and internally for access logs.

PR #2053, opened 2021-11-25
Invalidate All Sessions

clarencejychan:
Addresses #2050


clarencejychan:
Need to add screenshots but added a `Sign out of all sessions` to exist on the `Settings` page. I was thinking about doing a `dropdown` for the sign-out button but people needing to sign out of all sessions probably don't need to do it so often that it warrants having an extra button in the header. I'll need to add a tooltip to guide them towards the `Settings` page to sign out of all sessions.

maxwofford (commented):
Couple comments:

- How do I know if I need to use it? Can I see a list of active sessions?
- The "sign out of all sessions" button blends in next to the "save profile" button. Let's give it a different color, section, or give a brief description of what it does– right now it's a destructive action that looks like and is physically close to a save button.
- The tooltip description feels too long
- How do I know if it worked? When I click, does it just sign me out like the regular signout button? Maybe we can give a specific flash w/ details?

clarencejychan:
1. The list of active sessions will come in another PR (to keep them small, I'm planning on having a table contain it)
2. I can do that and just change the colour
3. I can't shorten the description there.
4. It does sign you out like a regular sign-out button. I could add a flash if it's helpful to say all your sessions are signed out but I think the button already describe what's supposed to happen and getting moved out should signify that it worked.

clarencejychan:
Screen Shot 2021-12-03 at 1 52 52 AM
Screen Shot 2021-12-03 at 1 58 27 AM

clarencejychan:
As noted in the previous comment the table with the session info will be in a follow-up pr this weekend. Will be using the 
`Session Information` sub heading in order to contain that info

app/helpers/sessions_helper.rb
```diff
@@ -82,4 +82,11 @@ def sign_out
     cookies.delete(:session_token)
     self.current_user = nil
   end
+
+  def sign_out_of_all_sessions
+    # Destroy all the sessions
+    current_user(false).user_sessions.destroy_all if current_user(false)
```
maxwofford: I'm not familiar with the `false` in the argument here. What's that used for?

maxwofford (approved):
This is awesome. I haven't tested the latest in a preview, but I'm cool with how it looks in screenshots. Go for merge if you're ready!

app/helpers/sessions_helper.rb
garyhtou: `false` used to prevent checking whether the user's Hack Club API session is still valid. it's kinda a quirk of using Hack Club API for auth
https://github.com/hackclub/bank/blob/c3c23e19557699db53a798eea2163242ba1bd3d9/app/helpers/sessions_helper.rb#L44-L45

However, now that I look into git history, it seems like that was removed a while back in https://github.com/hackclub/bank/commit/05ac25d61899d15ce85e5871beb8cb5a90850426

PR #2055, opened 2021-11-27
Fixes force use email bug

tanatorn:
This fixes a bug where if we force use email when the default is SMS login and submit the wrong login code, we'd be taken back to the SMS login page.


maxwofford (approved):
I just tested using the reproduction steps in your comment and I didn't run into the issue anymore– LGTM!

PR #2056, opened 2021-11-27
Fixes enroll SMS bug

tanatorn:
Fixes https://hackclub.slack.com/archives/C026RKHLPNJ/p1637747191096800

We were not catching the Unauthorized error which forced a logout by default. Now we'll catch the error in our `complete_verification` method and convert that to invalid login code which we can handle later in code.


exu3 (approved):
it works! thanks for the fix!

PR #2057, opened 2021-11-27
Raises error on login code != 6 digits

tanatorn:
Doing some validation that the login code so we don't have to make an extra API call


maxwofford (approved):
Great addition! Feel free to merge.

Should we have this on the front-end too?

garyhtou:
Unrelated, but cool feature to have: When a 6 digit code is pasted in, automatically submit the login form (prevents the user from having to hit enter)

PR #2058, opened 2021-11-28
Session Fingerprinting

clarencejychan:
Might need to rework the submit on auth instead

Was testing the fingerprinting on just the nav bar but this should happen as soon as the login exchange is done. Need to probably have more rich interaction on the current submit button in order to make it work

1. Wait for login code exchange to finish and we have no errors -> run the fingerprinting code


clarencejychan:
Or i guess just someway to attach the callback to the form (maybe just using `jQuery`) is fine too

clarencejychan:
Saves fingerprinting info on sign in now via the hidden fields. Will add display of the multiple sessions in a table in the `settings` page in a follow-up PR

maxwofford (commented):
This is going to be super helpful building access logs, thanks for adding it! 

The code looks like it's still a WIP, but here are a couple comments on the high level:

- Could you also drop in IP if we're not currently storing that?
- Also, is there an advantage to doing these fields on the frontend? If we can get any of these on the backend there's a higher chance people don't try messing with them.
- This PR makes session records important to keep around. Could you also clean up any instances of code that delete session data? Let's make sure https://github.com/hackclub/bank/pull/2053 doesn't destroy access data from this. Maybe we could add an `invalidated_at` to sessions to soft-delete?

clarencejychan:
1. Ack 
2. We can get some of the information via like just the base `request` object in Rails I believe (`request.user_agent` etc) but when I was looking I didn't see something in particular to do the actual fingerprinting (which I believe is unique to the browser). What we could probably do is just pass the fingerprint data via the library then grab the rest from the request as well.
3. Yeah we can do a soft delete - I'll add that tomorrow.

garyhtou:
irrc, a lot of the data required for fingerprinting can only be accessed on the client via javascript

clarencejychan:
So did some investigation and there's one Ruby Gem that could potentially help decipher user agent strings but honestly the JS tool I found does it a little cleaner and provides more relevant info so unless there's large qualms about getting it from the FE I'd like to keep it for now.

I added soft delete (`deleted` column) (maybe we should also only keep session records for X number of days?)

Added IP info

Table coming in upcoming PR.

maxwofford:
> maybe we should also only keep session records for X number of days

I'd like to keep these indefinitely for the time being-- if we need to remove data in specific instances we can do that manually.

EDIT: For context, access details on api.hackclub.com have been extremely good in cases of spam/fraud/moderation & we sometimes have to go surprisingly far back in history.

db/migrate/20211205051705_add_ip_and_soft_delete_to_sessions.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddIpAndSoftDeleteToSessions < ActiveRecord::Migration[6.0]
+  def change
+    add_column :user_sessions, :ip, :string
+    add_column :user_sessions, :deleted, :boolean, default: false
```
maxwofford: Let's have a deleted_at so we can keep track of when users are signed out.

maxwofford: We've used `acts_as_paranoid` for this in the past, see https://github.com/hackclub/bank/blob/33268c7b7e575124ef3dd01376e76df56b2765e1/app/models/emburse_transaction.rb#L9 for an example.

config/routes.rb
```diff
@@ -25,6 +25,7 @@
   get "project_stats", to: "static_pages#project_stats"
   get "bookkeeping", to: "admin#bookkeeping"
   get "stripe_charge_lookup", to: "static_pages#stripe_charge_lookup"
+  post "create_fingerprint_info", to: "users#create_fingerprint_info"
```
maxwofford: Where is this used?

maxwofford (commented):
Looks awesome! I added a couple comments on the PR

maxwofford (approved):
Marking approved so you can merge, but FYI test are failing b/c of the lines in schema.db -- recommend using `git add -p` or selecting specific lines in a GUI instead of adding whole files.

PR #2060, opened 2021-11-29
Fix donation page og:image

exu3:
Previously:
Screen Shot 2021-11-28 at 11 13 56 PM

Now:
Screen Shot 2021-11-28 at 11 32 41 PM


garyhtou:
whoops, I think it's related to this https://github.com/hackclub/bank/pull/1863

Is there an alternative to `URI`? (because it's depcreated)

exu3:
Oh oops

I think it maybe works without the encoding (idk maybe?) I remember trying it in hackclub/v3 without encoding the spaces and it seemed to work but idk if that's good practice 

I'll look into it tomorrow

garyhtou:
not encoding spaces should be fine, but special characters will probably break it

exu3:
Oh right

I can't seem to find the Ruby equivalent of JS's `encodeURIComponent()` which is what vercel/og-image uses.

Maybe the fix is to make a PR in vercel/og-image to support this encoding :P but idk what the JS equivalent to `CGI.escape` is.

garyhtou:
I think [vercel/og-image](https://github.com/vercel/og-image) doesn't support it because the text is located in path of the URL. Any spaces in the path should be encoded with `%20` and any spaces in the query is encoded with `+`. The rails methods are assuming we're encoding for queries, not path.

try these solutions:

- https://www.rubydoc.info/stdlib/erb/1.8.7/ERB/Util:url_encode
- replace (`gsub`) any `+` to `%20` after `CGI.escape`

maxwofford:
Does `url_encode` work here? https://rdoc.info/stdlib/erb/1.8.7/ERB/Util:url_encode

exu3:
It does work. https://github.com/hackclub/bank/pull/2060/commits/02614d385341652a327a52af498001329cc59f90 commit message is supposed to say `url_encode` but Codespaces does this weird thing sometimes where it just cuts off part of my commit message

I think this PR is good to merge
Screen Shot 2021-11-30 at 1 29 30 PM

maxwofford:
I'm cool with merge, but my previous message was because we're [defining it here](https://github.com/hackclub/bank/commit/02614d385341652a327a52af498001329cc59f90#diff-74910f4fc98892347120743cf759f0db22e4063b637c419bf4d89fcf5ec60266R16-R19) when it seems like Rails has it built in.

exu3:
Ah got it, just fixed that

PR #2061, opened 2021-11-29
Partner Sign-up form spice ✨

exu3:
🌶️ 
<https://bank--pr-2061.herokuapp.com/partnered_signups/sup_508tB6>


devenjadhav:
the spice must flow ✨

maxwofford:
Couple things I forgot to mention:

- Currently the connect button submits, but the final version will take the user to an embedded docusign
- We can change all the stuff on the frontend. All those boring field names like "Owner name" can be changed
- This'll be the first hack club page users arrive at, so if you want to do weird stuff to it (ie. loading animation or something), we totally can. First impressions are lasting impressions!

exu3:
I think there could be some improvements made on the animation timing/duration part. Can't quite put a word to it, but it feels a bit awkward, maybe slow, but maybe because I've been watching it over and over for the last hour or so.

Would appreciate any feedback on this!

Also, is there a world where the address form fields could be separate? Like separate fields for Street address, second address line, city/state, zip, country. I think it would look a little nicer, but I'm not sure if this would be a pain to implement in the db given that there are already several records with addresses formatted this way.

exu3:
Closing and re-opening to trigger a Heroku review app deployment.

garyhtou:
> Closing and re-opening to trigger a Heroku review app deployment.

https://dashboard.heroku.com/pipelines/e61db6c3-496d-4233-ad3c-4a29752dcc12

i think u should have access to redeploy the review app from the heroku pipeline. lmk if u don't have access

garyhtou:
woahhh cool stuff!!! i love the animation, and yeah, we 100% should make the address separate fields (and db columns). Don't worry about the existing data, it's all test data.

for the animation, i think simpler might be better here. would it be possible to make the loading animation slightly shorter? and some type of `ease in` (https://easings.net/). i think having the animation end quicker than it starts might make bank feel a tad bit more snappy.

and once it loads, only animate the gb and hcb logos?

exu3:
> > Closing and re-opening to trigger a Heroku review app deployment.
> 
> https://dashboard.heroku.com/pipelines/e61db6c3-496d-4233-ad3c-4a29752dcc12
> 
> i think u should have access to redeploy the review app from the heroku pipeline. lmk if u don't have access
Screen Shot 2021-12-30 at 10 37 14 AM

My level of access doesn't allow me to redeploy "inactive" review apps.

garyhtou:
> My level of access doesn't allow me to redeploy "inactive" review apps.

whoops! just invited you as a member to the heroku team. once u accept i think i should be able to give u manage access to the pipeline

exu3:
It does, thanks!

exu3:
Pushed some fixes for the animation and created the form fields for the address input.

Now for migrations 😶‍🌫️ which I did on a branch of this branch because I'm not sure if I did it right #2150

db/schema.rb
```diff
@@ -874,6 +874,12 @@
     t.datetime "submitted_at"
     t.string "docusign_envelope_id"
     t.boolean "signed_contract"
+    t.string "owner_address_line1"
+    t.string "owner_address_line2"
+    t.string "owner_address_city"
+    t.string "owner_address_state"
+    t.text "owner_address_postal_code"
+    t.string "owner_address_country"
```
garyhtou: fyi, `t.string "owner_address_country"` is a `string` here. That probably needs to be changed over to `integer`

exu3: Ah thanks for catching that!

maxwofford:
Is it cool if we merge this as is? The only current users are the Givebutter team while testing, so it still needs to work, but I'd rather bias for merging quickly than waiting for large changes.

exu3:
@maxwofford Sure I think it's cool to merge now. Just rebased.

maxwofford:
Here's the slick loading animation & form changes this PR has. Thanks so much @eilla1!

PR #2063, opened 2021-11-29
copy edit: Clarify tax deduction eligibility for non US events

garyhtou:
#2062

United States | International
-|-
image | image
https://bank--pr-2063.herokuapp.com/donations/start/hq | https://bank--pr-2063.herokuapp.com/donations/start/turner-fenton-hack-club


app/views/donations/_donation_form.html.erb
```diff
@@ -28,7 +28,10 @@
   </div>
 
   <p class="muted h5 mt2 mb1">
-    Your contribution is tax-deductible.
+    Your contribution is tax-deductible
+    <% if @event.country != "US" %>
+      in the US.
+    <% end %>
```
garyhtou: thanks! Your code looks good

Can add this logic in the `Event` model? I think it'll be helpful in other places as well.
Maybe something like:
```
# somewhere before the private separator in the Event model
def country_us?
  country == "US"
end
```
and usage would look something like this:
```
<% if !@event.country_us? %>
  in the US.
<% end %>
```

app/views/donations/_donation_form.html.erb
```diff
@@ -28,7 +28,10 @@
   </div>
 
   <p class="muted h5 mt2 mb1">
-    Your contribution is tax-deductible.
+    Your contribution is tax-deductible
+    <% if !@event.country_us? %>
+      in the US.
+    <% end %>
```
garyhtou:
```suggestion
    Your contribution is tax-deductible
    <% if !@event.country_us? %>
      in the US
    <% end %>
    .
```

maxwofford:
I haven't seen activity on this or #2062, so I'm going to go ahead with merging b/c it doesn't break anything.

PR #2075, opened 2021-12-01
Improvements to admin tx mapping

garyhtou:
Yellow row highlights only show on transactions that actually need manual mapping

gives you a little warning when mapping transactions you probably shouldn't be mapping
image


garyhtou:
In the first picture, the first transaction ("UBER TRIP") is unmapped, but the row is not highlighted because the mapping for that transaction will be taken care of by the transaction engine

PR #2077, opened 2021-12-01
Return on discrepancy

garyhtou:


garyhtou:
nvm, it's taken care of by `next if current_canonical_event_mapping`

PR #2078, opened 2021-12-01
Make many codespace-config steps only run if they need to

maxwofford:
This improves the script to only run things that need to run (ie. if there's already a `latest.dump`, don't download it again)


codespace-config.sh
```diff
@@ -32,13 +34,16 @@ fi
...
 echo "
 $(tput setaf 9)Hack Club Bank:$(tput sgr0) Step 4/7: Get Heroku Backups"
-heroku pg:backups:capture
-heroku pg:backups:download
+if ! test -f "./latest.dump"; then
+  heroku pg:backups:capture
+  heroku pg:backups:download
+fi
 echo "$(tput setaf 9)Hack Club Bank:$(tput sgr0) $(tput setaf 10)Done$(tput sgr0)"
+exit
```
garyhtou: was this `exit` intentional?

maxwofford: Nice catch!

kunalbotla (approved):
this is great! I don't see any issues when I test it but if something comes up we can check back here

PR #2089, opened 2021-12-03
Fix session to 30 days

clarencejychan:
This might be the offender which is kicking out login's after 1 day.


tanatorn (approved):
LGTM

maxwofford (approved):
Thanks for catching this! Go ahead with merging!

PR #2091, opened 2021-12-04
Adds better SMS error message

tanatorn:
The error message that used to be displayed was basically a wall of text nobody read.

image

This PR adds more interactive toast to it when it is clicked for all errors. Example:

image


tanatorn:
@garyhtou  i'm still experimenting, I'm trying to see how to show Rails flash messages via react

tanatorn:
@garyhtou I looked around the internet for a bit and didn't find a good answer since our API calls are async here. For this toast, I think a way around this is to render the HTML elements manually, thoughts? cc: @maxwofford

garyhtou:
Yeah, it seems like there isn't an easy way. I guess we have two options here:

- Manually render the HTML elements for these specific flash messages using Javascript/React
  - Similar to this https://github.com/hackclub/bank/blob/d8e55c3a37c4353eca6b158c0046996210a71b66/app/views/donations/finish_donation.html.erb#L8-L20 Although this doesn't seem to render them the same way as Rails flash messages.
- Rewrite our Rails flash messages system to render all of them via React. Instead of generating HTML for them on the server, send the flash messages through the header and use react to render them. It would make it easier to "inject" flash messages from the front end. Example: https://easysavecode.com/LZAYmx21

---

Flash message partial: https://github.com/hackclub/bank/blob/d8e55c3a37c4353eca6b158c0046996210a71b66/app/views/application/_flash.html.erb

maxwofford:
Is there a way to make the updates in a background job or one-time service? This feels like a weird thing to see as a user if I just have to type in information I already gave bank.

app/views/users/edit.html.erb
```diff
@@ -90,61 +90,54 @@
...
-                phoneNumber: @user.phone_number
-              },
-              { prerender: true }
-            %>
-          </div>
-        <% end %>
-
-      </fieldset>
-  <% end %>
+  </fieldset>
```
garyhtou: just wondering, were there any new changes here? it seems to be reverting https://github.com/hackclub/bank/pull/2123 and https://github.com/hackclub/bank/pull/2144

garyhtou:
i'll finish looking this over on tuesday!

clarencejychan:
Fixed the merge here

maxwofford:
With the changes in https://github.com/hackclub/bank/pull/2324, I'm removing the request for @garyhtou's review

maxwofford:
Merging this (which has been tested locally and in staging) to launch the feature behind a feature-flag. Will create a PR to move to general availability in the future after getting feedback.

PR #2092, opened 2021-12-06
Session table with individual sign out

clarencejychan:
Screen Shot 2021-12-05 at 10 03 59 PM


app/controllers/users_controller.rb
```diff
@@ -72,6 +72,17 @@ def logout_all
     redirect_to root_path
   end
 
+  def logout_session
+    begin
+      session = UserSession.find(params[:id])
```
clarencejychan: Can't do this until https://github.com/hackclub/bank/pull/2058/ is merged in as it includes soft deleting

garyhtou: try validating that the session belongs to the current user

clarencejychan: oh i just meant that i want to wait since I'll need to adjust the value for the session (but the soft delete isn't in main yet)

maxwofford (approved):
Haven't tried the preview deploy, but in screenshots this looks functional and fine to merge. @eilla1 can do a design pass once merged.

app/controllers/users_controller.rb
garyhtou: oops, I'm not sure why my comment was threaded under here. Yeah, it made sense to wait for #2058.

I was meaning that we should check that the `session` from `params[:id]` belongs to the current user. Otherwise, any signed-in user can log anyone else out.

clarencejychan: Done

PR #2095, opened 2021-12-08
Attempt to fix sql escape in #2094

maxwofford:


maxwofford:
This gets the page loading without error, so I'm merging it. There's still an issue that may or may not be fixed by a nightly job/service. The transaction is showing up with an incorrect HCB code & isn't linked to the ACH record.

PR #2096, opened 2021-12-09
Initial Docusign Integration

tanatorn:
Created a base class and the beginning of the docusign <> give butter pdf integration

Note that this is a WIP, but I figured splitting the PR out is probably better than doing it all at once

The flow currently works as follows.

Follow up work:
* Make all redirects/webhooks work


garyhtou:
occupied with finals right now! i'll take a dive into this mid-next week

app/controllers/partnered_signups_controller.rb
```diff
@@ -38,6 +40,29 @@ def update
...
+    if @partner.id == 1
+      service = DocumentService::Docusign::PartneredSignupContract.new(@partnered_signup)
+      unless @partnered_signup.docusign_envelope_id
+        data = service.create
+        @partnered_signup.docusign_envelope_id = data[:envelope].envelope_id
+        if @partnered_signup.save
+          redirect_to data[:signing_url]
+          return
+        else
+          flash[:error] = "Something went wrong, please contact support@hackclub.com for help"
```
garyhtou:
```suggestion
          flash[:error] = "Something went wrong, please contact bank@hackclub.com for help"
```

app/services/document_service/docusign/partnered_signup_contract.rb
```diff
@@ -0,0 +1,58 @@
...
+module DocumentService
+  module Docusign
+    class PartneredSignupContract
+      # ID of the template sent to the user, this isn't private so
+      # we're good to store it here
+      TEMPLATE_ID = if Rails.env.production?
+                      nil
+                    else
+                      "9e79448a-43f3-4116-a65c-71a5f173b7e6"
+                    end
```
garyhtou: Can we make this `Partner` specific? (store in a column in `Partner` model).

it can later be access via something like `partnered_signup.partner.docusign_template_id`

app/controllers/partnered_signups_controller.rb
```diff
@@ -22,11 +22,13 @@ def edit
   # PATCH /partnered_signups/:public_id
   def update
     @partnered_signup.update_attributes(partnered_signup_params)
+    # TODO: move this to the webhook
     @partnered_signup.submitted_at = Time.now
```
garyhtou: Maybe we should have multiple states and timestamps for this. one state for when the form is submitted and another state for when the contract is signed.

i've been thinking about moving PartneredContract's statement management over to `aasm` (similar to the Donations model)

garyhtou (commented):
looks good!

heads up, I'm not sure if this is a design requirement or not: there could be the possibility that for a single application to Bank (PartneredSignup), multiple people will want to sign the contract (@maxwofford might have more info on whether this needs to be supported or not)

app/services/document_service/docusign/partnered_signup_contract.rb
tanatorn: 👍

app/controllers/partnered_signups_controller.rb
tanatorn: I moved this to the webhook for now, I do think we should move this to the aasm, but we should probably do it in another PR

tanatorn:
@garyhtou  I ended up finishing the flow, since the PR is getting big, I think we can move the state management + other things to the next one. PTAL :)

app/services/document_service/docusign/api.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+module DocumentService
+  module Docusign
+    # Api is a singleton service that abstracts away some of the complexities in creating
+    # a Docusign request
+    class Api
```
garyhtou: It might be best to move this class into `app/services/partners` since that's where other 3rd party API services are held

app/services/document_service/docusign/signing_completion_webhook.rb
```diff
@@ -0,0 +1,46 @@
...
+      WEBHOOK_URL_BASE = "https://bank.hackclub.com"
+
+      if Rails.env.production?
+        ENVIRONMENT_KEY = :production
+      else
+        ENVIRONMENT_KEY = :development
+      end
+      HMAC_KEY = Rails.application.credentials[:docusign][ENVIRONMENT_KEY][:hmac_key]
+
+      # Creates a redirect URL that is protected by a HMAC so it isn't forge-able
```
garyhtou: good idea!

app/controllers/docusign_webhook_controller.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class DocusignWebhookController < ApplicationController
```
garyhtou: Is this a webhook or a redirect from Docusign?

Seems to be a redirect. If so, can we...
- rename this controller to `DocusignController`
- Update the service and method names to be "Redirect" rather than "Webhook"
- Change the route of the method below to be `docusign/signing_complete` (rather than `docusign_webhook`)

app/controllers/docusign_webhook_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  skip_after_action :verify_authorized # do not force pundit
+  skip_before_action :signed_in_user, only: [:signing_complete] # do not require logged in user
+
+  def signing_complete
+    hmac = params.require(:hmac)
+    timestamp = params.require(:timestamp)
+    partnered_signup_id = params.require(:partnered_signup_id)
+    service = DocumentService::Docusign::SigningCompletionWebhook.new
+    unless service.valid_webhook?(partnered_signup_id, timestamp, hmac)
+      render json: { error: "invalid signature" }, status: :bad_request
```
garyhtou: In the future (separate PR), we should make this an HTML response since this is a user endpoint (not api)

app/controllers/docusign_webhook_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+      return
+    end
+    partnered_signup = PartneredSignup.find(partnered_signup_id)
+    partnered_signup.submitted_at = Time.now
+    partnered_signup.signed_contract = true
+    partnered_signup.save!
```
garyhtou: this looks good for this pr, but something to think about:

as long as the redirect url to this endpoint can't be intercepted by the user (such as through Docusign's website), then this should be secure. If they can get their hands on the valid redirect url before they sign and are redirected, then we would probably need to ping DocuSign here to make sure the envelope's actually signed.

app/services/document_service/docusign/signing_completion_webhook.rb
```diff
@@ -0,0 +1,46 @@
...
+        partnered_signup_id = partnered_signup.id.to_s
+        hmac = compute_hmac(partnered_signup_id, timestamp)
+        params = ActionController::Parameters.new(
+          {
+            timestamp: timestamp,
+            partnered_signup_id: partnered_signup_id,
+            hmac: hmac
+          }
+        ).permit(:timestamp, :partnered_signup_id, :hmac)
+        WEBHOOK_URL_BASE + "/docusign_webhook?" + params.to_query
```
garyhtou:
```suggestion
        Rails.application.routes.url_helpers.docusign_webhook_url(params)
```
and remove `WEBHOOK_URL_BASE`

image

app/controllers/docusign_webhook_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+    hmac = params.require(:hmac)
+    timestamp = params.require(:timestamp)
+    partnered_signup_id = params.require(:partnered_signup_id)
+    service = DocumentService::Docusign::SigningCompletionWebhook.new
+    unless service.valid_webhook?(partnered_signup_id, timestamp, hmac)
+      render json: { error: "invalid signature" }, status: :bad_request
+      return
+    end
+    partnered_signup = PartneredSignup.find(partnered_signup_id)
+    partnered_signup.submitted_at = Time.now
```
garyhtou: Update values only if `signed_contract` is `false` because if a user is redirected multiple times, it will reset `submitted_at`.

I think this can happen if:

- They click the back button, then click the forward button (landing back on this url)
- If they land on Docusign, but the envelope has already been signed. I'd assume that DocuSign will redirect to this url again

app/controllers/partnered_signups_controller.rb
```diff
@@ -38,6 +37,36 @@ def update
 
   private
 
+  def signed_contract?
+    # Don't sign contract unless we have a docusign template id
+    unless @partner.docusign_template_id
+      @partnered_signup.submitted_at = Time.now
+      @partnered_signup.signed_contract = true
+      return true
+    end
```
garyhtou: All Partners that use PartneredSignup must have a `docusign_template_id`.

So, let's error and flash something to the user ("Something went wrong, please contact bank@hackclub.com for help")

As well as:
```ruby
Airbrake.notify("Partner ##{@partner.id} is missing a 'docusign_template_id'. Error creating docusign contract for SUP ##{@partnered_signup.id}")
```

garyhtou (approved):
cool! we're in a good place. It all works. Good job here!

I just a couple of edge cases and organizational things

app/controllers/partnered_signups_controller.rb
garyhtou: yup sounds good!

tanatorn:
I updated ~everything from the comments, gonna merge this @garyhtou

garyhtou:
Thanks for making those changes @tanatorn!

PR #2098, opened 2021-12-12
Capitalise `SMS` and place it in admin only view

sampoder:
It seems that it should be in admin only view as only admin's have anything in this section. Sorry if I'm wrong


exu3 (approved):
Thanks for catching this Sam!

PR #2100, opened 2021-12-13
Disable replit applications

garyhtou:
image


exu3:
Thoughts on moving it to the bottom since it's unavailable and possibly may not be available again in the near future? ([context](https://hackclub.slack.com/archives/GBXBFEED9/p1639430044412300?thread_ts=1639429427.410000&cid=GBXBFEED9))

garyhtou:
good idea! feel free to go ahead and make that change

PR #2110, opened 2021-12-17
Small Transparency Mode Redesign

sampoder:
For a while I've wished that I could share transparency mode pages that looked a bit like this. This PR was my attempt at coding it. I'm not sure, what do other people think?

Screenshot 2021-12-17 at 11 42 38 AM
Screenshot 2021-12-17 at 11 43 20 AM


garyhtou:
I like how streamlined this looks, however, iirc Zach prefers for transparency mode to be a literal read-only mode of Bank (same layout, just no edit permission). If you get the chance, this might be something you could chat with him about!

maxwofford:
I love how the message is showing up at the top though-- I feel like it gives better context then being in the aside where you have to scroll to it. @sampoder maybe we can drop the hidden nav changes, but keep the description where you moved it to for now? I'd be down to merge that and see how people like the change.

PR #2115, opened 2021-12-18
Fancier error pages

cjdenio:
👋 this PR replaces the somewhat Rails-default-looking error pages with some snazzy Hack Club ones.

(quick note: this doesn't affect the way errors are rendered in development)

<table>
<tr>
	<td> Before
	<td> After
<tr>
	<td> Screen Shot 2021-12-18 at 5 05 23 PM
	<td>
		Screen Shot 2021-12-18 at 6 44 16 PM
		Screen Shot 2021-12-18 at 6 21 28 PM
<tr>
	<td> Screen Shot 2021-12-18 at 5 05 32 PM
	<td> Screen Shot 2021-12-18 at 6 20 54 PM
</table>


app/controllers/events_controller.rb
```diff
@@ -308,8 +308,12 @@ def toggle_hidden
...
     @event = Event.friendly.find(params[:event_id] || params[:id])
   rescue ActiveRecord::RecordNotFound
-    flash[:error] = "We couldn’t find that event!"
-    redirect_to root_path
+    unless signed_in?
+      return redirect_to auth_users_path
+    end
+
+    @is_event_error = true
+    render "errors/not_found", status: :not_found
```
garyhtou: Can we flash a message and return to root instead? If someone goes to an incorrect event url, they'll probably want go to the home page and click on the event

garyhtou (changes requested):
I love this @cjdenio!! thank you so much for this PR

app/controllers/events_controller.rb
cjdenio: Yeah, absolutely!

garyhtou (approved):
Awesome! Good to merge ☃️

PR #2121, opened 2021-12-21
Fix Invoice helper payout date logic

garyhtou:
https://heroku-app97991095.airbrake.io/projects/288439/groups/3150975699857931084/notices/3155782036597162447?tab=notice-detail


config/routes.rb
```diff
@@ -390,7 +390,6 @@
     get "donations", to: "events#donation_overview", as: :donation_overview
     get "partner_donations", to: "events#partner_donation_overview", as: :partner_donation_overview
     get "bank_fees", to: "events#bank_fees", as: :bank_fees
-    # suspend this while check processing is on hold
```
garyhtou: Old comment from b9523e52d161c3ee615e8408f75e6f10b37dd850 and should have been removed in fdc9cad6

PR #2123, opened 2021-12-21
User sessions table to grid

exu3:
Did a little design pass on the session table as per [this comment](https://github.com/hackclub/bank/pull/2092#pullrequestreview-824289850).

This PR makes the session information section into a grid (formerly a table). The table is kind of difficult to use on mobile because of the horizontal scroll whereas this grid allows for responsiveness across different screen sizes.

There's also a tooltip on the sign out button that says "Sign out of this session" which I forgot to include in the screenshot.


app/views/users/edit.html.erb
```diff
@@ -92,46 +92,38 @@
...
-            <td>
-              <%= link_to logout_session_users_path(user: @user, id: session.id), method: :delete,
-                class: 'btn bg-error' do %>
-                  Sign out
-              <% end %>
-            </td>
-          </tr>
+    <div class="session-header pb2">
+      <legend class="heading flex flex-row justify-start items-center h2 pt2">
+        Session Information
```
```suggestion
        Sessions
```

app/views/users/edit.html.erb
```diff
@@ -92,46 +92,38 @@
 
 <h2>Authentication</h2>
```
```suggestion
<h2 class="mt4">Authentication</h2>
```

app/views/users/edit.html.erb
```diff
@@ -92,46 +92,38 @@
...
+            <% if !session.device_info %>
+              <p class="center">No data available for this session.</p>
+            <% else %>
+              <span class="flex items-center justify-between">
+                <span class="bold"><%= session.device_info %> on <%= session.os_info %></span>
+                <%= link_to logout_session_users_path(user: @user, id: session.id), method: :delete, class: 'muted tooltipped tooltipped--w z5', 'aria-label': 'Sign out of this session' do %>
+                  <%= inline_icon 'door-leave', size: home_action_size %>
+                <% end %>
+              </span>
+              <span>IP: <code><%= session.ip %></code> • <%= session.timezone %></span>
```
garyhtou: Maybe try icons here! possibly a globe (don't think our icon set has that) or crosshairs for IP. and clock for timezone? up to you!

app/views/users/edit.html.erb
```diff
@@ -92,46 +92,38 @@
...
-                  Sign out
-              <% end %>
-            </td>
-          </tr>
+    <div class="session-header pb2">
+      <legend class="heading flex flex-row justify-start items-center h2 pt2">
+        Session Information
+        <%= badge_for @user.user_sessions.count, class: 'bg-muted' %>
+      </legend>
+       <%= link_to logout_all_users_path, method: :delete,
```
garyhtou, line 102: would it be possible to make this a tad bit smaller?

garyhtou (commented):
Ayy love this! especially on mobile! Great work here, i've just got a couple minor suggestions and one feature request: indicating which session is the current one. you can rebase can use the new method from #2129

app/views/users/edit.html.erb
exu3: Good idea!

exu3:
Screen Shot 2021-12-24 at 1 38 32 AM
Thoughts on how this looks?

garyhtou (commented):
Love it! this is much better. just a minor thing:
image

good to merge after that!

---

however, if you want to improve on this further, here are a couple of ideas: (you 100% don't need to do these)
- reverse the order of the cards so that more recent sessions show up at the top. you can add `.order(created_at: :desc)` to `@user.user_sessions`.
   ```ruby
   @user.user_sessions.order(created_at: :desc).each
   ```
- allow the card to take up full width if there's only 1 session (i think this might look a tad bit better, but not completely sure)

exu3:
Added a non-breaking space for the green badge ordered the sessions by time created!

garyhtou (approved):
wahoo!

PR #2137, opened 2021-12-24
Delete server.pid on startup

exu3:
Removing the `bundle exec rails s -p 3000 -b "0.0.0.0"` part from the end of the docker-compose command in codespace-start.sh seemed to successfully delete `/usr/src/app/tmp/pids/server.pid` if it exists. Screen recording below demonstrates this:

https://user-images.githubusercontent.com/72365100/147321096-344e643c-9455-4965-b907-14e6dc163c9b.mov


exu3:
Here’s a walk through of the screen recording:

1. `docker ps`
    1. Server didn’t shut down cleanly so there are 2 processes running
2. `codespace-start.sh`
    1. Error: “A server is already running”
3. Edit `codespace-start.sh`
4. Run codespace-start.sh again
    1. Server is successfully started on port 3000 which means that `server.pid` was deleted.

PR #2131, opened 2021-12-22
Update icon set (@hackclub/icons)

garyhtou:
Our icon set is now up to date with https://github.com/hackclub/icons and still includes some one-off icons unique to Bank.

https://github.com/hackclub/icons-for-bank
^ that's the repo i used to do some sketchy icon set comparsion
Deployed version: https://icons-for-bank.hackclub.dev/

---

Summary of process:

1. Import & overwrite (except `admin-badge` and `view-forward`
2. Remove duplicates and update their references
3. remove outdated, unused, and broken icons
4. fix svg errors

PR #2139, opened 2021-12-24
Add width/height to view-forward icon

exu3:
I'm not sure if this solution is the best way to go about fixing this issue, but this PR changes the view-forward icon to the old one because the new one introduced in #2131 seems larger which makes some of the icons overlap with the text. This is what it looks like:
Screen Shot 2021-12-24 at 3 06 57 AM

This icon seems to only be used on the admin tasks section, so it shouldn't be a pressing issue.


exu3:
Ah there seems to be some styling in the SVG code which I think we wanted to get rid of? This is probably not a great solution lol. Will think of something else.

garyhtou:
yup! let's see if we can get rid of that styling. that would most likely mean that we need to apply the styling somewhere here https://github.com/hackclub/bank/blob/1a1afe082c69861b8f7ec04406362e7ccf6746d4/app/assets/stylesheets/components/_lists.scss#L1-L1
^ that's the only file where the icon is used

exu3:
The icon is being used as a `background` (which has a lot of styling limitations in the CSS) which is why I think the icon was originally implemented this way (fill and styles in the SVG file). [I don't think it's possible to style an SVG in a CSS stylesheet](https://stackoverflow.com/questions/18434094/how-to-style-svg-with-external-css); the most painless way to do it is via `style` tag in the SVG file.

As of https://github.com/hackclub/bank/pull/2139/commits/e22f17db5d6041c9634610306a72280799128216, this PR adds back the width, height, and fill color to the SVG file, which fixes the large icon issue. If this is alright, I think this PR is ready to go. 

However, if it's important to get rid of the width, height, and fill color styles in the SVG, we could refactor the Card component & make the arrow an icon/SVG rather than a background. But I don't know if that's necessary because this icon is only used in the admin Card component anyways.

garyhtou:
> However, if it's important to get rid of the width, height, and fill color styles in the SVG, we could refactor the Card component & make the arrow an icon/SVG rather than a background. But I don't know if that's necessary because this icon is only used in the admin Card component anyways.

I'd prefer this instead so that we can use the icon else where. however, this is good to merge until we get that sorted out!

garyhtou (approved):

PR #2144, opened 2021-12-25
Hide session management during onboarding

cjdenio:
Just a tiny li'l change

Screen Shot 2021-12-25 at 6 48 38 PM


exu3 (approved):
Good call, thanks Caleb!

PR #2145, opened 2021-12-26
Show holiday feature field only when transparency enabled

exu3:
Wahoo! It's 4:30 and I had a thought! 

This PR adds an event listener on the transparency mode checkbox; so when the box is checked, the holiday-features-in-transparency-mode checkbox will appear, but when the box is not checked, the holiday-features-in-transparency-mode checkbox will disappear.

https://user-images.githubusercontent.com/72365100/147407969-e810ae96-6093-4d9c-8ac3-d652612f0712.mov


app/assets/javascripts/ui.js
```diff
@@ -249,6 +249,17 @@ $(document).on('turbo:load', function () {
...
 
+  if (BK.thereIs('holiday_features_transparency')) {
+    const holidayFeaturesTransparency = BK.s('holiday_features_transparency')
+    const transparencyToggle = $('#event_is_public')
+    $(transparencyToggle).on('change', e => {
+      if (e.target.checked) holidayFeaturesTransparency.slideDown()
+    })
+    $(transparencyToggle).on('change', e => {
+      if (!e.target.checked) holidayFeaturesTransparency.hide()
+    })
```
garyhtou:
```suggestion
    $(transparencyToggle).on('change', e => {
      if (e.target.checked) {
        holidayFeaturesTransparency.slideDown()
      } else {
        holidayFeaturesTransparency.hide()
      }
    })
```

garyhtou (changes requested):
Heck yeah! cool stuff!

Can you handle the case where transparency is initially on? (at this moment, when the page loads with transparency on, the holiday features checkbox is auto-hidden).

There are a couple of ways to solve this:

- Remove `autohide` css class and use JS to hide holiday features (on initial load) if transparency is on. this could cause some layout shift (similar to [CLS](https://web.dev/cls/)) on the initial load if the element's being hidden. but it might be far enough down the page that it doesn't matter (not in viewport on initial load)
- use ERB to add `autohide` css class to the element `unless @event.is_public`. easier, but it's kinda tacky because now we have the "same" logic in multiple places

exu3:
I added it in the ERB!

app/views/events/_form.html.erb
```diff
@@ -43,7 +43,7 @@
       <%= form.check_box :is_public %>
       <%= form.label :is_public, "Enable Transparency Mode" %>
     </div>
-    <div class="field field--checkbox">
+      <div class="field field--checkbox" data-behavior="holiday_features_transparency <% if !event.is_public? %>autohide<%end%>">
```
garyhtou: good! just small formatting things

```suggestion
    <div class="field field--checkbox" data-behavior="holiday_features_transparency <% if !event.is_public? %>autohide<% end %>">
```

app/assets/javascripts/ui.js
```diff
@@ -249,6 +249,18 @@ $(document).on('turbo:load', function () {
...
   }
 
+  if (BK.thereIs('holiday_features_transparency')) {
+    const holidayFeaturesTransparency = BK.s('holiday_features_transparency')
+    const transparencyToggle = $('#event_is_public')
+    $(transparencyToggle).on('change', e => {
+      if (e.target.checked) {
+        holidayFeaturesTransparency.slideDown()
+      } else {
+        holidayFeaturesTransparency.hide()
```
garyhtou:
```suggestion
        holidayFeaturesTransparency.slideUp()
```
https://api.jquery.com/slideup/

garyhtou (approved):
just these two small tweaks, then good to merge! ty for this!

exu3:
Thanks for the fixes!

PR #2146, opened 2021-12-27
Fix: echo statements in codespace-config.sh

exu3:
Got a little error when I ran the configuration script. I think it's because it's trying to execute the line as a command instead of printing it. This PR attempts to fix that.

I also updated the docker-compose command at the end to match the one in the codespace-start script.

This is the error:
```
./codespace-config.sh: line 30: $'\E[91mHack': command not found
```


exu3:
This doesn't touch Bank code, so I don't think it will break anything. Can revert later if it doesn't work :)

PR #2147, opened 2021-12-29
Fix flavor text

exu3:
The lack of an ending </a> tag breaks the UI a bit. This PR fixes that.

Screen Shot 2021-12-28 at 7 00 58 PM


app/helpers/static_pages_helper.rb
```diff
@@ -362,7 +362,7 @@ def holiday_flavor_text
       "<a href='https://santatracker.google.com/' target='_blank' style='color: inherit'>Santa's on the way!</a>",
       "dashing through the snow",
       "defrosting...",
-      "send snow photos to <a href='mailto:bank@hackclub.com' target='_blank' style='color: inherit'>",
+      "send snow photos to <a href='mailto:bank@hackclub.com' target='_blank' style='color: inherit'>bank@hackclub.com</a>",
```
garyhtou: ty!

PR #2141, opened 2021-12-25
LET THERE BE SNOW ☃️🎅🎄❄️

garyhtou:
image

Co-authored-by: Ella <eilla1@users.noreply.github.com>

PR #2150, opened 2021-12-31
Migrations: add address lines to partnered signups table

exu3:
I wonder if I did this right 🤔 

This is for #2061


garyhtou (changes requested):
ayyy good job with your first migration! Just a couple more steps for it to be complete:
1. run the migration
    - you can use `bundle exec rails db:migrate` or click the "run pending migrations" button if you see one after loading Bank.
	image
2. `bundle exec rubocop -a` to add the magic comment
3. commit the changes (`schema.rb` should have changed)

Also, just a thought, for `country`, maybe we should use [country numeric codes](https://www.iban.com/country-codes) (data type `:integer`) to make it easier to sort (similar to what Kunal and Tan had [set up for Events](https://github.com/hackclub/bank/blob/60354e6046b18f17a669f2224227f302e259cfc2/db/migrate/20211005200012_add_country_to_events.rb))

exu3:
For `owener_address_country`, I used string because I didn't think we'd ever have to look up a partnered organization owner's country. I see that the `country` column in partnered_signups uses :integer though, so maybe it'd be nice to keep it consistent.

Would I have to undo this migration to change that? Or can I just edit the file

garyhtou:
Since this migration hasn't been deployed to `main`, it is possible to edit the migration file then reset your database with `bundle exec rails db:reset`.

In a scenario where it has already deployed, then we should create another migration that changes the :string to an :integer.

either one works here!

db/schema.rb
```diff
@@ -526,6 +526,7 @@
     t.date "owner_birthdate"
     t.string "webhook_url"
     t.integer "country"
+    t.boolean "holiday_features", default: true, null: false
```
garyhtou: whoops, looks like I forgot to push as a part of #2141. i'm updating it in #2155 so that the git history is a tad bit more clear.

once u update this branch with main, this line change should disappear from this PR

PR #2151, opened 2021-12-31
Rename search query to `q`

garyhtou:
There are a couple of other spots that still should be switched over

image


exu3:
Screen Shot 2022-01-01 at 3 10 59 AM

garyhtou:
Close! Good job with updating the forms. There's one more part that we need to update: the controllers.

For example, the `donation_overview` method in the Events controller is still using the `:search` param. We'll need to change this over to `:q`.
https://github.com/hackclub/bank/blob/5f2d76b4414772b40b9e2d1e35b2e987138d985c/app/controllers/events_controller.rb#L220-L220

Here's an example of what that might look like:
```ruby
relation = relation.search_name(params[:q]) if params[:q].present? 
```
<br/>
However, now that we have changed it to `:q`, any existing links that used the deprecated `:search` param will no longer work. For example, this link would break: https://bank.hackclub.com/hq/donations?search=matthew (there probably won't be very many of these links saved/bookmarked by people, but it's something we should keep in mind and support if possible)

So, to keep backwards compatibility, we can add this at the top of each method that we switched `:search` to `:q`:
```ruby
params[:q] ||= params[:search]
```
This will set `params[:q] = params[:search]`, only if `params[:q]` is falsey (such as nil). In other words, it's just a fancy way of writting:
```ruby
if !params[:q]
  params[:q] = params[:search]
end
```

Here's an example of what that looks like for the event show method:
https://github.com/hackclub/bank/blob/8e07598293660d101bfc424bbc743356c16c80fa/app/controllers/events_controller.rb#L16-L22

garyhtou:
wahoo 🐟 !! looks good

exu3:
🐟

PR #2157, opened 2022-01-02
Limit GSuite domain uniqueness to those not deleted

garyhtou:
another attempt at https://github.com/hackclub/bank/pull/1720 https://github.com/hackclub/bank/pull/2156


db/schema.rb
```diff
@@ -614,7 +614,6 @@
     t.text "remote_org_unit_id"
     t.text "remote_org_unit_path"
     t.index ["created_by_id"], name: "index_g_suites_on_created_by_id"
-    t.index ["domain"], name: "index_g_suites_on_domain", unique: true
```
garyhtou: we need to drop uniqueness db validation because it does not allow for conditionals (such as limiting the scope of the uniqueness constraint to only those which are not deleted). Instead, uniqueness validation is handled solely in the model

https://github.com/hackclub/bank/blob/efa831150859b2083644308bb6eeb88b28703233/app/models/g_suite.rb#L45

PR #2158, opened 2022-01-02
EXPERIMENT: Command Palette (Kbar)

garyhtou:
hey you! you're on GitHub right now, hit `cmd`+`k` (or `ctrl`+`k`).

Yeah, that's cool, isn't it? and yeah, I think we should jump on the bandwagon!

image

https://user-images.githubusercontent.com/20099646/147875117-7efe6429-f401-4feb-b9fe-f144d30342bf.mov

---

This PR is currently WIP and very very sketchy. I need to find a better way to render ERB/HTML inside a React component. This is required since `Kbar` uses a React context provider. Due to limitations with [`react-rails`](https://github.com/reactjs/react-rails) (it can't take in children... why), i'm using `dangerouslySetInnerHTML` to render the ENTIRE application 😬 

tbh this is more of proof of concept

check it out: https://bank--pr-2158.herokuapp.com/ (hit `cmd`+`k` — works without signing in right now)


garyhtou:
another challenge will be loading different actions for each page. `Kbar` uses React hooks for this... but that doesn't really work with rails. So the best way might be to set up some kind of instance variable array that each view can append to. that variable will then be passed into the React `kbar` component as a prop

garyhtou:
useful links:

- https://github.com/reactjs/react-rails/pull/166
- https://github.com/reactjs/react-rails/issues/223
- https://github.com/timc1/kbar/blob/main/example/src/App.tsx
- https://blog.superhuman.com/how-to-build-a-remarkable-command-palette/
- https://github.com/stefanjudis/awesome-command-palette

PR #2162, opened 2022-01-03
Add Hack Club Theme to Admin pages

garyhtou:
https://hackclub.slack.com/archives/C026RKHLPNJ/p1641202229248200

@eilla1 feel free to continue this on. At the moment, I've commented out the existing admin CSS, however, some of it might still be useful.

And, some inline styles need to be updated to support dark mode (ideally, they'd move into the stylesheet as classes).

https://github.com/hackclub/bank/blob/b0763e94b4f13b23cf47452c4ceb2e225939ba4d/app/views/admin/check.html.erb#L39-L39


garyhtou:
tbh, why don't the admin pages use the same stylesheet as the rest of the application... I think the admin stylesheet should extend the application stylesheet. (or at least use the basecss and spaces sheets)

Here's what it looks like if admin pages use the Application stylesheet:
image
https://bank--pr-2163.herokuapp.com/ (https://github.com/hackclub/bank/pull/2163) definitely needs some tweaking

exu3:
All the pages that are linked in the nav bar at the top have been styled. I think there are a few inner pages like Google Workspace that I haven't looked at yet.

exu3:
It works now, and it looks alright, but the biggest issue here is that it doesn't work on mobile at all. Thoughts on extending the the application stylesheet vs using a separate one for admin? I think either/both are cool, but this one is less work as it just extends the css.hackclub.com stylesheet. The former would make the admin pages feel more integrated into Bank though.

garyhtou:
> I think either/both are cool, but this one is less work as it just extends the css.hackclub.com stylesheet. The former would make the admin pages feel more integrated into Bank though.

I agree! Tbh, let's go with keeping them separate. Cause now that I take a closer look, the application stylesheet mainly loops in the components (which will rarely be used on admin pages). So let's not extend. However, I think we should still import these into the admin stylesheet:
```scss
@import 'variables';
@import 'basscss';
@import 'space';
@import 'utilities';
@import 'admin_dark'; // duplicate `app/assets/stylesheets/_dark.scss` and remove unnecessary stuff
```
so that we can use dark mode, color variables, and all the handy css class libraries

imma close #2163

exu3:
Wahoo, I think this is almost ready.

Here's a little screen recording: 
(The theme switch was me toggling my system theme; there's no theme toggle on these pages.)


https://user-images.githubusercontent.com/72365100/148672496-6ce620ea-7f30-4f26-ae33-9a761b044f33.mov

exu3:
@garyhtou It seems that I am unable to request your review as the author of this PR but if you could take a look at this whenever you get a chance, that would be awesome.

garyhtou:
Looks much better! Thanks for working on this! good to merge

PR #2166, opened 2022-01-03
Invert dino color in dark mode

exu3:
This PR makes the dinos on the error pages SVGs so that the color of their outline is based on the page theme.

https://user-images.githubusercontent.com/72365100/147985834-d76be340-620c-4ea5-a6a2-2d6d011521c9.mov


cjdenio (approved):
🦕

PR #2174, opened 2022-01-04
Allow users to delete Google Workspace accounts

garyhtou:
@maxwofford do you know why this was kept to admins during the initial design?


app/views/g_suite_accounts/_panel.html.erb
```diff
@@ -30,7 +30,6 @@
         <action type="button" class="pop menu__toggle" data-behavior="menu_toggle" aria-expanded="false" tabindex="0">
           <%= inline_icon 'more' %>
           <div class="menu__content" data-behavior="menu_content">
-            <% if account.accepted? && organizer_signed_in? %>
```
garyhtou: not necessary (it's a tautology) because accounts are accepted by default and the page is only visible to organizers

maxwofford: This was previously visible in transparency mode. Is there a reason it was removed?

garyhtou: 🤷‍♂️  005675f31b52df95104df2dbb6941b9f7ca5f378. should we bring it back?

garyhtou: on a second thought, making that page public would reveal all of the org's email addresses. is that maybe a bit much?

exu3 (commented):
I think it's alright to allow users to delete Google Workspace accounts!

For this PR I'm getting a 404 on google_workspace pages of orgs that have not set up Google Workspace.

https://bank--pr-2174.herokuapp.com/bank/google_workspace
https://bank--pr-2174.herokuapp.com/hq/google_workspace

Also when I try to delete a Google Workspace account (from a regular user account), I get returned to root and flash "You are not authorized to perform this action"

garyhtou:
> For this PR I'm getting a 404 on google_workspace pages of orgs that have not set up Google Workspace.


whoop. it's an issue with an admin tool #2177

garyhtou:
> Also when I try to delete a Google Workspace account (from a regular user account), I get returned to root and flash "You are not authorized to perform this action"

thx for catching. 15bb1f79f304a4465e882358fbb0d6552b75df97

exu3 (approved):
Wahoo! Looks good to me.

app/views/g_suite_accounts/_panel.html.erb
maxwofford: Let's leave that for another day then-- i'd love to eventually have it show up in transparency (maybe the secondary and gmail addresses redacted), just so people looking through transparency mode can peruse an event, but that is out of scope of this PR

kunalbotla:
just seeing this but it is probably a bad idea to fully leave for the user, people already suspend accounts without knowing what it does and individual accounts don't cause a billing burden for us and very few people actually delete accounts. 

maybe something where they request would be better so there's that intermediary?

maxwofford:
https://github.com/hackclub/bank/pull/2174#issue-1093812904 We used to have a design philosophy of "any potentially destructive actions that happen in our code should be triggered by an admin". for example,

- you can't remove a co-lead, you ask an admin to do it
- ordering a card actually just put a request in, an admin would be the one to order it
- gsuite setup intentionally not automated so an admin would review it

That philosophy was back when we were a bit more defensive with our codebase. We have a much stronger, much more experienced team working on bank now so I think it's fine to let regular users access this feature as long as we're confident in the code.

app/views/g_suite_accounts/_panel.html.erb
```diff
@@ -41,12 +40,10 @@
                 method: :put,
                 'data-confirm': "Do you really want to #{account.suspended? ? 're-activate' : 'suspend'} #{account.address}?"
             %>
-            <% end %>
             <%= link_to 'Delete',
               account,
               method: :delete,
-              'data-confirm': "Deleting #{account.address} is *permanent, irreversible, and does not back up any data*. Think carefully before you click ok.",
-              class: 'admin-tools' if admin_signed_in?
+              'data-confirm': "Deleting #{account.address} is *permanent, irreversible, and does not back up any data*. Think carefully before you click ok."
```
maxwofford: Does this give everyone in the org access to delete anyone else's gmail account?

kunalbotla:
I agree that users should have an in-dashboard way of starting the action but Google Workspace is distinct from the others since it can't be simply reversed because of the data associated. i.e., if a user is removed, they can be re-added

users unintentionally deleting Google Workspace accounts would be more destructive and users regularly don't read warnings or text

app/views/g_suite_accounts/_panel.html.erb
garyhtou: Yes, it does

maxwofford:
https://github.com/hackclub/bank/pull/2174#issuecomment-1005757836 That's a good point. We intentionally don't let users delete cards, they can only freeze and unfreeze them. Is there an advantage to deleting email accounts over freezing them?

exu3:
Some organizations (like this [one](https://bank.hackclub.com/reach-for-the-stars/google_workspace)) have a huge list of suspended accounts which tends to clog up the page a bit. In that case, it might be better option to delete them if they've been suspended for a while/the account owner is no longer affiliated with the organization.

I've seen very few requests for deleting Google Workspace accounts though (HS-1232 is the only one I could find from a quick search), so maybe this feature won't be useful to most orgs?

maxwofford:
> Some organizations (like this one) have a huge list of suspended accounts which tends to clog up the page a bit.

How about a hide/archive feature for suspended gsuite accounts? I can totally see how all the outdated accounts get in the way.

garyhtou:
what if someone wants to completely reset an account (that would involve delete/recreate). i feel like we don't see as many requests because it's a lot of friction to reach out to bank support to see if account deletion is even a feature.

i think we should give users as much autonomy as possible and allow deletion (~~soft deletion (https://github.com/hackclub/bank/pull/2174#issuecomment-1008768347)~~), but have ux that save guards it. examples:
image
image
image

garyhtou:
on a second thought, we shouldn't implement soft deletion (do hard deletion instead) for google workspace accounts — there's some complexity with recreating those accounts especially since the unique identifier is their email address

maxwofford:
Maybe we can limit this to only users deleting their own account until we have per-user permissions in place? I think I'd feel weird knowing everyone suddenly has access to delete all my emails.

garyhtou:
the best we can do right now is `backup_email == current_user.email`.

we can also do `[current_user.email, User.find(creator_id).email].includes?(backup_email)`, however this feels less secure bc it's common for one person creates all/most of the accounts.

average # accounts per account creator, per event: 10 accounts
image

image
https://bank.hackclub.com/blazer/queries/224-g-suite-average-accounts-per-creator-per-event/

---

what if we did:

```ruby
if backup_email == current_user.email
	# allow deletion from dashboard via a button
else
	# send mailer to backup_email for confirmation
	#   - would require a GSuiteAccountDeletionRequest
    #   - GSuiteAccountDeletionRequest expires after `x` days
    #   - mailer contains a link to the GSuiteAccountDeletionRequest. delete on visit
end

```

garyhtou:
Just chatting with Max about this.

Here were some ideas that we had:
- implementing a trash system for G Suite Accounts (and possibly other models — such as invoices)
- Accounts in the trash can be auto deleted after 30 days (hard delete on google)
- or, we can "soft delete" on google by
    1. suspending the account
    2. change the account's email address to something unique and non-conflicting (such as `deleted_timestamp_username@domain.com `)
    3. remove the allias created from changing the email address
- whether data is actually deleted probably doesn't matter that much to the user. being visually hidden (and the terminology around it) is more important. example, trash/deleted feels more complete than archiving


in the meantime, i'm going to close out this PR

PR #2175, opened 2022-01-04
Hide settings link in transparency mode

cjdenio:
Not sure if this is desired, but right now the settings link is the only one that's completely non-functional in transparency mode.


exu3 (commented):
Google Workspace and Settings are completely non-functional in transparency mode. But I think it'd be alright to still display it just to show what features the Bank dashboard has for prospective users.

PR #2183, opened 2022-01-05
Update creds

tanatorn:
Change integration keys to @maxwofford (Developer Account max@hackclub.com)


maxwofford (approved):
This is good! Thanks for switching us out of base64. It's fine with me as long as we have that comment in the credentials file saying this is the `max@hackclub.com` account

PR #2185, opened 2022-01-06
Make JSON and table scroll horizontally

exu3:
I'm not sure if this is desired, but this PR makes the admin_viewer records (from #2171) look a tad bit cleaner by applying the "overflow-x-auto" class to a parent div element so that the overflow is horizontally scrollable.


maxwofford:
This is admin-facing so I'm going to merge for the sake of time. It looks great in the screen recording. Thanks for the fix!

PR #2186, opened 2022-01-06
Export donations as CSV/JSON

exu3:
This PR adds the ability to export donations as CSV or JSON similar to how transactions are exported. Code of questionable quality.

The button placement: I wasn't sure if it would be better on the donations page or main event page in the same dropdown as the current export button, so I've put it on both for now (which can be seen in the staging deployment).
Tangentially related idea—what if the export button opened a modal, and the modal had a description of what kind of data gets exported, etc., and at the bottom of the modal, a field for selecting what to export (All transactions as JSON, all txs as CSV, donations json, donations csv).

I did see #2033 but I realize it likely refers to Partner Donations because it's on the Givebutter task board. 

Here's what the export looks like (CSV).

This is what it's like currently on the donations page. One flaw with this is that the user might not understand it's for exporting data unless it says "export."

What it looks like if it's added to the current menu:


app/policies/donation_policy.rb
```diff
@@ -9,6 +9,10 @@ def index?
     user&.admin?
   end
 
+  def export?
+    user
```
maxwofford:
```suggestion
    record.event.users.include?(user) || user&.admin?
```

This method is run to check if the route can be reached. If we use just `user`, any user could start an export of any event's donations.

app/views/events/show.html.erb
```diff
@@ -78,8 +78,10 @@
         Export
         <% if organizer_signed_in? %>
           <div class="menu__content h5" data-behavior="menu_content">
-            <%= link_to 'CSV', export_transactions_path(event: @event, format: :csv), target: :_blank %>
-            <%= link_to 'JSON', export_transactions_path(event: @event, format: :json), target: :_blank %>
+            <%= link_to 'Transactions (CSV)', export_transactions_path(event: @event, format: :csv), target: :_blank %>
+            <%= link_to 'Transactions (JSON)', export_transactions_path(event: @event, format: :json), target: :_blank %>
+            <%= link_to 'Donations (CSV)', export_donations_path(event: @event, format: :csv), target: :_blank %>
+            <%= link_to 'Donations (CSV)', export_donations_path(event: @event, format: :json), target: :_blank %>
```
maxwofford: Let's remove the donations from this dropdown and add them to the donations page... the transaction export will also show donations, and we want to reduce confusion

app/policies/donation_policy.rb
exu3: Ah I see, thanks.

app/views/events/show.html.erb
exu3: https://github.com/hackclub/bank/pull/2186/commits/007a205b21da102d7ba7dde3f51c4d213837cf65

maxwofford:
As per #2033, this feature should probably also export PartnerDonations. Up to you if you want to do that in a different PR. Events will probably only ever have Donations **_or_** PartnerDonations, never both

exu3:
https://github.com/hackclub/bank/pull/2186#issuecomment-1006887819 I can work on that in a separate PR!

app/services/donation_service/export/csv.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      private
+
+      def event
+        @event ||= Event.find(@event_id)
+      end
+
+      def header
+        ::CSV::Row.new(headers, ["date", "donation_from", "amount_cents"], true)
```
maxwofford: Could we also have a link to the hcb code show page in here? It'd certainly help our accountant to have that info & I bet users would want it too.

app/services/donation_service/export/csv.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require "csv"
+
+module DonationService
+  module Export
+    class Csv
+      BATCH_SIZE = 1000
```
maxwofford: What does this do?

maxwofford: I'm cool with it in here, but want to sanity check that it doesn't prevent the user from downloading all their donations if they have over 1000 donations

app/views/events/show.html.erb
```diff
@@ -78,8 +78,8 @@
         Export
         <% if organizer_signed_in? %>
           <div class="menu__content h5" data-behavior="menu_content">
-            <%= link_to 'CSV', export_transactions_path(event: @event, format: :csv), target: :_blank %>
-            <%= link_to 'JSON', export_transactions_path(event: @event, format: :json), target: :_blank %>
+            <%= link_to 'All transactions (CSV)', export_transactions_path(event: @event, format: :csv), target: :_blank %>
+            <%= link_to 'ALl transactions (JSON)', export_transactions_path(event: @event, format: :json), target: :_blank %>
```
maxwofford:
```suggestion
            <%= link_to 'All transactions (JSON)', export_transactions_path(event: @event, format: :json), target: :_blank %>
```

(typo)

app/views/events/donation_overview.html.erb
```diff
@@ -29,6 +29,16 @@
         </div>
       </action>
     </div>
+
+    <div class="md-pl1">
```
maxwofford: Screen Shot 2022-01-06 at 15 22 24

The icons look pretty similar and they're right next to each other. Could we tooltip this or something?

app/services/donation_service/export/json.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      private
+
+      def event
+        @event ||= Event.find(@event_id)
+      end
+
+      def row(ct)
+        {
+          created_at: ct.created_at,
```
maxwofford: Mentioned already, but it'd be nice to have a link to the hcb show page in here

app/controllers/donations_controller.rb
```diff
@@ -94,6 +96,53 @@ def refund
...
+
+    authorize @event.donations.first
+
+    respond_to do |format|
+      format.csv { stream_donations_csv }
+      format.json { stream_donations_json }
+    end
+  end
+
+  def stream_donations_csv
```
maxwofford: Seems like https://github.com/hackclub/bank/pull/2186/files#diff-ffc09b43f46260b97fd886063b4d5bdd45981eb81ce208c1b0af8e11f536950aR110-R145 should be `private`

maxwofford (changes requested):
Nice work! This is pretty close to done. Left a couple comments. Additionally, this shows refunded donations in the export:
Screen Shot 2022-01-06 at 15 33 31
Screen Shot 2022-01-06 at 15 34 41

app/services/donation_service/export/csv.rb
exu3: `batch_size` supposedly reads the csv file in batches; it still downloads all donations (tested with HQ's 21194 records). Performance-wise, it also feels alright without this line though even with 21194 donation records.

https://stackoverflow.com/questions/2962134/ruby-read-file-in-batches/41069373

app/services/donation_service/export/csv.rb
exu3: Good idea! https://github.com/hackclub/bank/pull/2186/commits/28ecf504a511984330f62e54155307f8ebda88b1

app/views/events/donation_overview.html.erb
exu3: Agree, I also thought it might be better if the design of the export button was more consistent with the existing export button, so I moved it a little further down. Do you think that makes it more clear?

Screen Shot 2022-01-06 at 7 17 55 PM

exu3:
> Additionally, this shows refunded donations in the export:

Would this be useful information to know? For bookkeeping perhaps? I've added a column to show the `aasm_state` (pending, deposited, refunded) so once exported, the user can easily create a filter to filter out refunded donations if need be.

app/views/events/donation_overview.html.erb
maxwofford: That looks nice! It feels a little off center and isn't consistent with the way we show price amounts on other pages, but it feels like it's getting closer.

maxwofford:
> I've added a column to show the aasm_state 

@eilla1 Perfect, love it

app/services/donation_service/export/csv.rb
maxwofford: Cool! Let's leave the `BATCH_SIZE = 1000` then. I think we'll want it not to speed up downloads, but to prevent the process from using up too much memory– if 21194 donations are loaded into RAM when I click "export" that's potentially RAM away from someone else trying to load a page or download their own export.

maxwofford (approved):
This looks great. I think we can still improve the "export" button placement, but this is totally functional and I'm cool with it getting merged.

garyhtou:
amazing!!

PR #2187, opened 2022-01-06
Add blankslate text if no donations/invoices/transfers exist

exu3:
Fixes #1153. 

This PR adds a little message if no donations, invoices, or transfers exist on the respective pages.

This PR also changes the tooltip text on "Send" to say "**Transfer money**" instead of the current message, "Send checks and wires", because people don't have the ability to send wires through the Bank dashboard (only checks, ACHs, and PayPal at the moment).

Prev:
Screen Shot 2022-01-06 at 12 33 30 AM

This PR:
Screen Shot 2022-01-06 at 12 33 08 AM


app/views/events/donation_overview.html.erb
```diff
@@ -81,6 +81,7 @@
   </div>
 <% end %>
 
+<% if @donations.count != 0 %>
```
maxwofford:
```suggestion
<% if @donations.size > 0 %>
```

Count is a little ambiguous here b/c I'm not sure if it's running ruby's `count`, which is good, or active record's `count`, which can be slow in this context. http://web.archive.org/web/20100210204319/http://blog.hasmanythrough.com/2008/2/27/count-length-size

exu3: Ah I see, thanks for the link!

maxwofford (changes requested):
This looks great! It's good to go once you remove the other instances of `.count` here. If you wanted to get super fancy you can also ruby it up with something like `.any?` or `.none?`
```rb
if @donations.any?
  do_something
else
  do_something_else
end
```

maxwofford (approved):
Great! Thanks for fixing

PR #2188, opened 2022-01-06
Current user's comments are highlighted

maxwofford:
Screen Shot 2022-01-06 at 13 30 09
Screen Shot 2022-01-06 at 13 30 05

Comments from the current user now have a faint blue background


maxwofford:
This also gives a faint heighlight to the current user's avatar across every use of the `avatar_for` method:

Screen Shot 2022-01-06 at 13 43 13
Screen Shot 2022-01-06 at 13 42 56

And adds a tooltip for the same avatars to clearly show it's the currently logged in user:

Screen Shot 2022-01-06 at 13 51 39
Screen Shot 2022-01-06 at 13 51 31
Screen Shot 2022-01-06 at 13 51 04

maxwofford:
@eilla1 I'm not attached to these changes, but this was the original reasoning:

Screen Shot 2022-01-06 at 13 56 49

I like how GitHub makes it so clear which comments/actions came from you vs other users.

exu3 (approved):
Cool! I like these changes. Very klassy.  The avatar outline is a bit subtle though; I don't think I'd be able to tell unless it was pointed out to me but I don't think that's necessarily a bad thing.

Screen Shot 2022-01-06 at 11 08 03 AM

maxwofford:
@eilla1 Do you have thoughts on how to make it more apparent which avatar is your own user? I totally agree & am all ears.

exu3:
https://github.com/hackclub/bank/pull/2188#issuecomment-1006854219 Maybe a badge? I can't think of a good icon off the top of my head though.

I think it's also fine to not clearly indicate the current user's avatar because I'd assume that everyone has different avatars so it wouldn't be too hard to distinguish yourself from your teammates.

maxwofford:
Merging this for the improvement to the comment designs. The avatar stuff is on the subtle side, but I don't think it degrades the experience for anything.

PR #2190, opened 2022-01-06
Add DNS Check to Admin Google Workspace Details

kunalbotla:
users have removed records following verification causing issues in a few cases. this PR adds the `DNS Check` url to the always visible Google Workspace Details table so admins can check DNS records even after verification.


kunalbotla:
image

adds a clickable link to nslookup.io the same way the `DNS Check` link on the verification screen functions
image

PR #2191, opened 2022-01-06
Prevent document previews from overflowing cards

maxwofford:
Before legal doc previews would overflow their cards:

Screen Shot 2022-01-06 at 14 45 21

Now, they're limited to the width of the CSS grid they're a part of:

Screen Shot 2022-01-06 at 14 44 55

(Images don't show because this is development, but you can see the outline of the image box being forced into the card)


exu3 (commented):
Neat! Thoughts on making the card heights the same height? Similar to the recent change on the event cards on the homepage.
Screen Shot 2022-01-06 at 11 59 22 AM

maxwofford:
That's a good idea! Let's hold off on it until we get all the preview issues fixed up though & a placeholder added-- it looks kinda broken on some events:
Screen Shot 2022-01-06 at 15 18 27

PR #2193, opened 2022-01-07
Add fallback for document preview

exu3:
This PR fixes #2189. Uses `placeholder-document.svg` if a preview image doesn't exist. Haven't been able to get the word doc preview to work.

Screen Shot 2022-01-06 at 10 02 59 PM


maxwofford:
```rb
Event.last.documents.last.file.filename.extension
=> "pdf"
```

Maybe we can also include the extension in/near the icon? Some documents don't show their type in the filename

exu3:
Hmm it's not working, but I think it's because I don't really know what I'm doing. Is the file extension method supposed to go in the controller or model?

> ```ruby
> Event.last.documents.last.file.filename.extension
> => "pdf"
> ```
> Maybe we can also include the extension in/near the icon? Some documents don't show their type in the filename

app/views/documents/_document.html.erb
```diff
@@ -1,8 +1,15 @@
...
-
-    <strong class="brand-sans h3 block line-height-2"><%= document.name %></strong>
+  <li class="card card--hover h-100">
+  <div class="overflow-hidden mb-" style="max-height: 250px;">
+    <% if document.preview_url %>
+      <%= image_tag(document.preview_url, width: 330, style: 'max-width: 100%;') %>
+    <% else %>
+      <%= image_tag("placeholder-document.svg", width: 330, style: 'max-width: 100%;') %>
+      <%# Show the document file extension (wip) -@eilla1 %>
+      <span><%= document.extension %></span>
```
garyhtou: try this!

```suggestion
      <span><%= document.file.filename.extension %></span>
```

exu3:
Screen Shot 2022-01-10 at 9 15 19 AM
🌞 
Screen Shot 2022-01-10 at 9 18 09 AM

app/controllers/documents_controller.rb
```diff
@@ -24,6 +24,7 @@ def new
   def create
     @document = Document.new(document_params)
     @document.user = current_user
+    @document.extension = Event.last.documents.last.file.filename.extension
```
maxwofford: Is this line supposed to be here?

maxwofford (commented):
Looks great, but it looks like https://github.com/hackclub/bank/pull/2193/files#diff-106fa7a2d2cf35ff0f0ff6cf0fd8bb91bd21371ded68df439ef8992f2ee9c2b6R27 was left over. Go ahead and remove it and it's good to merge!

PR #2194, opened 2022-01-07
Users can export CSV|JSON of their partner donations

maxwofford:
Closes #2033


maxwofford:
Going to merge for the time being b/c this only affects partnered orgs (which aren't a public feature yet). Thanks for the code this was based on @eilla1!

PR #2197, opened 2022-01-08
Partners editable in Admin

tanatorn:
This PR adds the ability for admins to edit partner details. Currently, we're only allow them to edit the `docusign_template_id`, but I'd imagine it might be useful for other things as well.

The partners page now has an "edit" button:
image

The new edit page looks like this:
image


maxwofford (approved):
Thanks for adding this!

PR #2198, opened 2022-01-08
Admin Contract Signing

tanatorn:
* Creates an endpoint - `admin#partnered_signup_sign_document` - to handle redirect to docusign from admin for signing of the contract
* Adds a role to the docusign redirect endpoint to actually support both the signer role and the administrator role
Here's the full flow:

There's some state management left to do, e.g. we need to move the date of `submitted_at`, `accepted_at` and `rejected_at` to a state machine model, but I figured we can do that in another PR


app/services/partners/docusign/admin_contract_signing.rb
```diff
@@ -0,0 +1,20 @@
...
+      def initialize(partnered_signup)
+        @partnered_signup = partnered_signup
+        @docusign_api = Partners::Docusign::Api.instance
+      end
+
+      def get_admin_signing_link
+        @docusign_api.create_recipient_view(
+          @partnered_signup.docusign_envelope_id,
+          "zach@hackclub.com",
+          "Max Wofford",
```
tanatorn: @maxwofford  I need the email/name you put for the `admin` role for the docusign template.I'm currently getting this error:
```
{"errorCode":"UNKNOWN_ENVELOPE_RECIPIENT","message":"The recipient you have identified is not a valid recipient of the specified envelope."}
```

app/controllers/partnered_signups_controller.rb
```diff
@@ -81,13 +81,19 @@ def set_partner
...
       :owner_phone,
       :owner_email,
       :owner_address,
       :owner_birthdate,
+      :owner_address_line1,
+      :owner_address_line2,
+      :owner_address_city,
+      :owner_address_state,
+      :owner_address_postal_code,
+      :owner_address_country
```
tanatorn: this seems to be a bug with the partnered signups cc: @eilla1

exu3: The partnered signup address line (previously just `owner_address`) was separated into 6 fields in #2061 ([migration](https://github.com/hackclub/bank/blob/main/db/migrate/20211231074013_add_address_lines_to_partnered_signups.rb)). Did that cause an issue? I can't seem to tell 😅 It's possible I could have done something wrong or missed a step.

tanatorn: I wasn't able to create a new partnered signup as it was looking for `owner_address` that wasn't there, we needed to add this change and the one in the partnered_signups model

exu3: Ahhh sorry about that. I made a PR #2199 to fix it so that the form works again for anyone who needs to work with it.

app/services/partners/docusign/admin_contract_signing.rb
maxwofford: The template doesn't have anything in the name or email field. Maybe this is in test mode and you can only send to the developer account? The email would be max@hackclub.com in that case.

Screen Shot 2022-01-09 at 21 51 33

tanatorn:
PTAL @maxwofford @garyhtou

maxwofford:
This looks great! If this interferes with the partnered signup flow (sending "connect" to docusign, then to an error flash) we can't merge it just yet while givebutter is still testing the flow on their end. Could we have an exception for partner id#2 before merge?

maxwofford (approved):
https://github.com/hackclub/bank/pull/2198#issuecomment-1012177334 Go ahead and merge once this is hidden for partner id#2

tanatorn:
@maxwofford this shouldn't block partner 2 since we have this conditional logic here - https://github.com/hackclub/bank/blob/main/app/controllers/partnered_signups_controller.rb#L26

PR #2199, opened 2022-01-09
Fix partnered signup form address fields

exu3:
This PR removes the old owner_address field from the partnered signup controller and model so the form works again.
Addresses https://github.com/hackclub/bank/pull/2198#discussion_r780712517.

Currently errors on submit:
Screen Shot 2022-01-08 at 8 43 33 PM


tanatorn (approved):
LGTM! Thanks for the quick turnaround

PR #2205, opened 2022-01-11
Add owner_address fields to API

exu3:
This PR adds the 6 separate the owner_address fields to all places in the code where I found a `owner_address`.


app/contracts/api/v2/partnered_signups_new_contract.rb
```diff
@@ -10,7 +10,13 @@ class PartneredSignupsNewContract < Api::ApplicationContract
         required(:owner_name).filled(:string)
 
         optional(:owner_phone).filled(:string)
+        # (@eilla1) :owner_address was deprecated in #2061 in favor of separating it into 6 fields
```
exu3: I wasn't sure if the `owner_address` should be removed completely yet, so the only place where I completely removed it is the documentation. I added a comment in the other places. Although if it's cool to just delete the line entirely, I could make that change.

garyhtou:
Go ahead and leave `:owner_address` in the db, but unused in the code.

once everything's working, we can remove it from the db using migrations in a separate pr

garyhtou:
remove this too
https://github.com/hackclub/bank/blob/e6c85abeb5f0509533ae4022059c0ea57972210a/app/controllers/partnered_signups_controller.rb#L89-L89

garyhtou:
@eilla1 is this good to merge?

exu3:
@garyhtou It should be good to go.

PR #2207, opened 2022-01-11
SUP address line 2 is not required

maxwofford:
Screen Shot 2022-01-11 at 13 47 45


exu3 (approved):
Thanks for fixing it!

PR #2209, opened 2022-01-11
Remove mention of application from GSuiteAccount creation

garyhtou:


kunalbotla:
can we make it "request created". it sounds a little like the creation has already happened otherwise

garyhtou:
whoops, mean to make it "Google Workspace created."

PR #2210, opened 2022-01-12
Dont show impersonated sessions

clarencejychan:
#2102


db/schema.rb
```diff
@@ -845,6 +846,7 @@
   create_table "partner_donations", force: :cascade do |t|
     t.bigint "event_id", null: false
     t.string "hcb_code"
+    t.string "donation_identifier", null: false
```
clarencejychan: idk if this is still relevant? it got picked up in the migration

maxwofford: Please amend it out of this PR. We can add it back in another commit, but I'd like that to be separate from this being merged in.

db/schema.rb
```diff
@@ -384,6 +384,7 @@
     t.text "message"
     t.text "hcb_code"
     t.string "aasm_state"
+    t.string "donation_identifier"
```
maxwofford: https://github.com/hackclub/bank/pull/2210/files#r784084603

db/migrate/20220112041139_add_impersonated_by_to_user_sessions.rb
```diff
@@ -0,0 +1,5 @@
+class AddImpersonatedByToUserSessions < ActiveRecord::Migration[6.0]
+  def change
+    add_column :user_sessions, :impersonated_by, :bigint
```
garyhtou: Should this be a reference? rather than just a bigint?

db/schema.rb
clarencejychan: The `create table` in `schema.rb` is just a representation of the db state right. I'm assuming on `main` there's currently a migration that's actually adding this atm.

maxwofford: I'd just like the `git blame` to reflect where stuff came from. If you need to make a separate branch from main to run migrations & put it in a separate PR that's fine.

db/migrate/20220112041139_add_impersonated_by_to_user_sessions.rb
clarencejychan: I'm having some trouble specifically with some `postgres` issues. I think the fact that there'll be 2 references to `users` is kinda messing up the migrations.

So `user_id` references `users` which is fine, but it's complaining that when using a method that assigns `impersonated_by` to ref `user_id` it's complaining that `impersonated_by` doesn't exist. I did some reading and there's some concerns about how postgres locks the tables when adding foreign keys and the strong migration isn't allowing for just a normal add ref. Following the instructions found here https://github.com/ankane/strong_migrations#adding-an-index-non-concurrently isn't really helping though unfortunately.

Any suggestions are welcome

clarencejychan: I could keep looking into it but tbh because `impersonations` are supposed to be short lived anyways the ref doesn't help that much in protecting data integrity unless for whatever reason the `impersonated` user deletes themselves during the course of an impersonation which is kinda rare

db/schema.rb
clarencejychan: done

maxwofford:
Moving this over to https://github.com/hackclub/bank/issues/2102 so I can amend without dealing with the merge commits in this.

PR #2212, opened 2022-01-14
Fix Stripe Partner Charges List

garyhtou:
BEFORE:
image

AFTER:
image


garyhtou:
This should fix the duplicate payout mentioned in https://github.com/hackclub/bank/issues/2164.

The payouts are created via an async job, but it's interesting how these guard clauses and aasm's `in_transit` state transition didn't prevent it from creating duplicate payouts in the first place
https://github.com/hackclub/bank/blob/cbe0a60fd90b15663b066dde57c3adc9e1d47107/app/services/partner_donation_service/create_remote_payout.rb#L11-L19
https://github.com/hackclub/bank/blob/cbe0a60fd90b15663b066dde57c3adc9e1d47107/app/models/partner_donation.rb#L47-L49

maxwofford (approved):
Nice catch. thanks for the fix!

maxwofford:
> didn't prevent it from creating duplicate payouts in the first place

Should we make a ticket for this?

garyhtou:
> > didn't prevent it from creating duplicate payouts in the first place
> 
> Should we make a ticket for this?

https://github.com/hackclub/bank/issues/2214

PR #2221, opened 2022-01-17
Use forbidden icon instead of 'error' icon

exu3:
An 'error' icon seems not to have been in existence anytime in the past few months. This PR changes the error icon in the partner_donation model to the 'forbidden' icon. Fixes #2220


garyhtou:
Thank you!

PR #2233, opened 2022-01-20
Login form: prevent double submission

garyhtou:


maxwofford (approved):
Just tested this locally & every edgecase i tried worked! Thanks for adding this.

PR #2236, opened 2022-01-20
Add heroku metric dependencies

garyhtou:
https://devcenter.heroku.com/articles/language-runtime-metrics-ruby#getting-started


maxwofford (approved):
Thanks for adding this!

maxwofford:
Could we also add any documentation we need for the configuration of bank this requires outside of code? (ie. [buildpack](https://github.com/hackclub/bank#heroku-tasks))

garyhtou:
> Could we also add any documentation we need for the configuration of bank this requires outside of code? (ie. [buildpack](https://github.com/hackclub/bank#heroku-tasks))

#2245

PR #2247, opened 2022-01-21
Fancy mailers!

garyhtou:
three things for mailers!

1. fancy layout
2. inline styles
3. CSS utility classes

## fancy layout
All emails now have a default header, footer, and styling. @eilla1 helped make this look nice!

it comes with four different layouts options:
- `default` (header and footer)
- `no_header`
- `no_footer`
- `body_only` (removes header and footer, but retains basic styling and max-width)

you can override the default header and footer!
```ruby
<% content_for? :footer do %>
  <p>I replaced the default footer!!!!</p>
<% end %>

<% content_for? :header do %>
  <p>I replaced the default header!!!!</p>
<% end %>
```

you can add blocks before and after the header/footer :)
```ruby
<% content_for? :before_header do %>
  <p>yo I'm before the header!</p>
<% end %>

<% content_for? :after_header do %>
  <p>yo I'm after the header!</p>
<% end %>
```

## Inline styles (`premailer-rails`)
Writing fancy html emails suck for various reason. the biggest culprit is CSS. it all has to be [inline](https://www.litmus.com/blog/do-email-marketers-and-designers-still-need-to-inline-css/), and sometimes it just doesn't work. yuck! this PR adds `premailer-rails` to help with inline styles.

`premailer-rails` also generates the plain text format for HTML emails. yay!

## Import utility CSS stylesheets
Mailer templates now have access to our CSS stylesheets. these are tree shaken by `premailer-rails` since styles are inlined.

---

## How emails look

image

image

Light mode screenshot from Gmail. Dark mode screenshot from Office Outlook 365.

Dark mode styles/colors depend on the email client. We aren't explicitly defining dark mode styles, but we did choose colors that are compatible with darker backgrounds.

<details>
<summary>first version (outdated)</summary>

image
image
image
image
image
image
image
image
image
image

</details>


garyhtou:
@eilla1 thoughts on moving the bank logo to the left of "Hack Club Bank"? and then adding a `<hr/>` or bottom border to the header

garyhtou:
@eilla1 can you also double check that these colors (especially the links) work well in dark mode?

https://github.com/hackclub/bank/blob/fancy-mailer/app/assets/stylesheets/mailer.scss#L29-L31

exu3:
> @eilla1 can you also double check that these colors (especially the links) work well in dark mode?
> 
> https://github.com/hackclub/bank/blob/fancy-mailer/app/assets/stylesheets/mailer.scss#L29-L31

How do I view it in dark mode? All my emails seem to show up in light mode despite setting my theme to dark.

garyhtou:
> How do I view it in dark mode? All my emails seem to show up in light mode despite setting my theme to dark.

ah yeah, some email clients (like gmail) will still show the email in light mode. i think outlook shows dark mode. i think the easiest way to test this is to just manually inject some dark mode CSS into the letter opener page from ur browser

---

it seems like some clients also support prefers-color-mode. we could use this to switch the background and body text color
https://sidemail.io/articles/dark-mode-in-html-email/

PR #2250, opened 2022-01-21
Fixes admin redirect

tanatorn:
Scaffolding for doing the state transitions after admin signing.
The cron basically looks at all pending applications with a document, checks docusign and sets the application state as completed _after_ all signatures have been entered.

The second part here is the uploading file, I haven't put this in the flow yet, but it should be part of the "accept" flow. Where after the Admin clicks `accept`, we'll create the hack club user, associate the user with a certain event and then upload the docusign contract associated with their account.


app/services/partners/docusign/admin_contract_signing.rb
```diff
@@ -11,7 +11,6 @@ def initialize(partnered_signup)
       def admin_signing_link
         @docusign_api.create_sender_view(
           @partnered_signup.docusign_envelope_id,
-          Partners::Docusign::SigningCompletionRedirect.new.create(@partnered_signup, role: :admin),
```
garyhtou: just wondering, is docusign not redirecting back? or is there an issue with getting the state from the redirect?

garyhtou (approved):

tanatorn:
Docusign actually doesn't do the redirect back for whatever reason.

PR #2256, opened 2022-01-24
Fix `<select>` alignment in Firefox

cjdenio:
Fixes the alignment of `<select>` components, specifically in Firefox.

before:
Screen Shot 2022-01-24 at 11 36 40 AM
after:
Screen Shot 2022-01-24 at 11 37 06 AM


exu3 (approved):
LGTM, hooray thanks! 🦑

PR #2267, opened 2022-01-27
Introduce Partner Representatives and move invite to service

garyhtou:
What are representatives:
https://github.com/hackclub/bank/blob/8810a5343c394cf746016d334c1519d5f9220dcd/app/models/partner.rb#L59-L67


maxwofford (approved):
This is an interesting approach. I definitely see the want for something like this-- let's merge it!

garyhtou:
resolves https://heroku-app97991095.airbrake.io/projects/288439/groups/3186688579995399988?filters=%7B%22order%22%3A%22last_notice%22%2C%22resolved%22%3A%22false%22%2C%22search%22%3A%22user%22%7D&tab=notice-detail

PR #2268, opened 2022-01-28
Magic login links 🔮🪄✨ + "peaceful" session

garyhtou:
### Magic Login Links
- Allows partners to login users automatically (no need to sign in). Read more in the [Bank Connect API swagger docs](https://bank.hackclub.com/docs/api#/Organizations/v2OrganizationsGenerateLoginUrl)
- This PR refactors and re-enables existing code
- And adds extra security:
  - link must be used in 15 seconds of generation (should be generated on demand)
  - must not be a new account. this forces them to sign in via email (allows us to verify that they own that email address)
  - if the user had clicked "sign out of all sessions", then they must sign in with email/sms first before they're able to use magic links again
  - must be using an IP from an active or expired peaceful session
- Login Tokens track which Partner requested them.
  - Allows for adding more information to Sessions table in user settings. For example, the session was signed in with a magic link from Givebutter (`UserSession`→`LoginToken`→`Partner`)

### About "peaceful" sessions
https://github.com/hackclub/bank/blob/85878dac10d99d963f98d9b75761949ab79ccd07/app/models/user_session.rb#L11-L34


app/services/auth_service/token.rb
```diff
@@ -2,15 +2,72 @@
...
+    end
+
+    # @return [User]
+    def user
+      @user ||= login_token.user
+    end
+
+    def trusted_ips
+      # Sessions that were peacefully expired are considered "trusted"
+      user.user_sessions.with_deleted.peacefully_expired
```
maxwofford: Do we want event deleted sessions to be trusted? If I'm on an untrusted network or VPN someone could use that to gain a trusted IP.

garyhtou: that's a good point!

If we were to only use active sessions, that would only allow a user to use a magic link when they are already signed in with that IP. the only use case I can think of is multiple browsers on the same device. is that a bit too restrictive?

if someone did obtain an IP you previously used and wanted to use it to gain access, then:
- the IP would need to belong to a session created within 6 months ago
- get a login link
  - This means they are signed into Givebutter and have already gone through all of Givebutter's security measures (as long as givebutter doesn't "leak" links)
- use the login link within 15 seconds of it being generated
- if the user had clicked "sign out of all sessions", then they must sign in with email/sms first before they're able to use magic links again
  - this could be a solution, but it's not obvious to users
- the account can't be new (not relevant in this problem)
  - lets us verify that they own that email address

i think including deleted peaceful sessions in the trusted ips is okay, but a permanent fix is something like a "remember me" option for the session (but i personally hate the UX of those)

app/services/api_service/v2/generate_login_token.rb
```diff
@@ -0,0 +1,41 @@
...
+      def initialize(partner:, user_email:, organization_public_id:)
+        @partner = partner
+        @user_email = user_email
+        @organization_public_id = organization_public_id
+      end
+
+      def run
+        # Validate that the Partner has permission to create a login URL for
+        # this user and organization
+        if organization.partner == @partner && user.events.include?(organization)
```
maxwofford: I'm a little confused by this conditional, should this be an `unless`?

maxwofford (approved):
This looks great @garyhtou! I left a comment, but go for merge once you're ready.

Also ran into an issue while testing, but don't think it's related to this PR specifically... when I signed in for the first time with a newly created account I ran into this.

Screen Shot 2022-02-25 at 07 41 48

app/services/api_service/v2/generate_login_token.rb
garyhtou: whoops! yes, it should be an unless. good catch

maxwofford:
@garyhtou Is this ready for merge now? If so, let's do it!

garyhtou:
> Also ran into an issue while testing, but don't think it's related to this PR specifically... when I signed in for the first time with a newly created account I ran into this.

Yeah, it seems to be unrelated — but it's something we do need to handle

garyhtou:
I'm working on implementing a cluster analysis algorithm to better compare IP addresses. Right now we're directly comparing them (binary comparison) — but IP addresses can change often. That'll come in another PR

PR #2269, opened 2022-01-28
[1792] Convet HEIC to Jpeg

clarencejychan:
image

So this is for the routes under `/hcb` I know that there's transactions that use it as well but I need to find out how those one's get attached then I'll submit a PR there


garyhtou:
I'm getting this error in staging https://github.com/hackclub/bank/issues/2270

Screen Shot 2022-01-27 at 11 23 16 PM

clarencejychan:
I'm guessing it might have to do with how files are stored on staging vs dev cause dev i think it's in a `tmp` folder that can just be accessed but I'm guessing the `tempfile.path` isn't necessarily accurate with opening files since it's probably in some staging s3 bucket somewhere. I'll look into this,

maxwofford:
Seems like we can now do this via activestorage: https://guides.rubyonrails.org/v6.1/active_storage_overview.html#transforming-images, and current edge rails will convert the preview automatically https://edgeguides.rubyonrails.org/active_storage_overview.html#transforming-images.

Going to close this PR for now, but leave https://github.com/hackclub/bank/issues/1792 open b/c we haven't made the underlying variant call changes we need.

PR #2275, opened 2022-01-28
Add robots.txt rules & metadata to auth page

exu3:
Fixes #2274

This PR adds rules in robots.txt so that pages can be indexed by search engines except pages like cards, invoices, transfers, etc where it might be weird to have that data show up in a Google search.

Also adds some basic metadata to the auth page.


maxwofford:
Can we wildcard disallow and have individually allowed pages? I think the current setup also allows every org in transparency mode, which feels kinda weird.

exu3:
Good point. 

How would `allow`ing bank.hackclub.com/ work if `Allow: /` is a wildcard?

maxwofford:
@eilla1: I haven't done it before, but I bet there's a way in the robots.txt spec to just allow the root path and disallow all else.

maxwofford:
:eyes: https://stackoverflow.com/a/21794569

garyhtou:
omg what if we custom generated a robot.txt based on transparency ahhhhhhhhh that would be cool

maxwofford:
> custom generated a robot.txt

That'd be sick! Could we also add transparent events to our sitemap.xml and OG metatags so they look good on google/twitter/bookface/vrchat?

If so, let's also update the transparent event toggle on the settings page to mention this makes them indexable.

maxwofford (approved):
Looks great! Could we also add an og:image?

exu3:
Thoughts on this as the og:image?

woof

maxwofford:
@eilla1 That looks great! I'm a little worried though that someone will see that for a different event (ie. a link to MAHacks shows Hack Pennsylvania), so let's just use that OG image on the root.

For event OG tags:
- it's a link someone is using to share their org, so let's not make it an ad for bank-- if I tweet my hackathon's finances, I want it to say something like "ORG_NAME has open-sourced their finances on Hack Club Bank. See how their money is used."
- i'd love to in the future to do dynamic OG images, but I don't want to hold up this PR. for the time being an HCB logo with a line that says something like "Transparent finances on Hack Club Bank" is fine.

exu3:
🐴

exu3:
> I'm a little worried though that someone will see that for a different event (ie. a link to MAHacks shows Hack Pennsylvania), so let's just use that OG image on the root.

I've changed the og image to `summary` (instead of `summary_large_image`) and it now uses the HCB logo instead of the marketing-esque image. I think it feels better that way.

> it's a link someone is using to share their org, so let's not make it an ad for bank-- if I tweet my hackathon's finances, I want it to say something like "ORG_NAME has open-sourced their finances on Hack Club Bank. See how their money is used."

Totally agree. This PR adds event og images using the workshop-cards which also shows the event name (for transparent projects only). 

### Other:

Since the FAQ/Branding static pages are now search indexable, I've also added some basic metadata for those as well in this PR.

---
<details>
<summary>Screenshots</summary>


Screen Shot 2022-01-28 at 4 38 29 PM
Screen Shot 2022-01-28 at 4 38 13 PM

</details>

PR #2286, opened 2022-02-02
Fix double HTML escaping in page titles

cjdenio:
👋 Hello! This fixes a bug where certain page titles were escaped twice - see [this page](https://bank.hackclub.com/melody/documentation) for an example.

Screen Shot 2022-02-02 at 5 08 55 PM

### Quick explanation:

`yield()` emits a string marked with `html_safe`, causing the Rails view system to assume the value has already been escaped. Interpolation negates that, however (specifically the interpolation [here](https://github.com/hackclub/bank/blob/68cb5634b04d3ee5cc52d1087d265772c63087b9/app/views/layouts/application.html.erb#L5)), so Rails tries to escape the title a second time.

```ruby
irb(main):001:0> title = "Caleb's Event".html_safe
=> "Caleb's Event"
irb(main):002:0> title.html_safe?
=> true
irb(main):003:0> "#{title}".html_safe?
=> false
```

(I also removed the `.html_safe` call inside event-related page titles; I think whoever added that thought it did something else)


exu3 (commented):
Interesting, TIL  I think this makes sense; thanks for the PR!

maxwofford:
Nice catch for this! After a quick git blame, looks like this was the original issue: https://github.com/hackclub/bank/issues/226. I have your branch checked out though and it's rendering just fine, so the underlying issue was probably fixed sometime over the last couple years.
Screen Shot 2022-02-10 at 14 58 15

cjdenio:
The underlying issue was [this line over yonder](https://github.com/hackclub/bank/blob/2787984342de1966e9f1d96e3c8e46e746b60c5a/app/views/layouts/application.html.erb#L5), which this PR fixes! (https://github.com/hackclub/bank/pull/2286/files#diff-f43fe075643e681b2c01c2f853bb0c4299d135b47fcbd4da96890d521c49e3eb)

maxwofford:
Covered this on a call w/ @garyhtou, so I'm removing the requested review. Thanks for the fix @cjdenio!

PR #2300, opened 2022-02-11
Add country_enumable concern

maxwofford:


exu3 (approved):
Nice, thanks Max!

app/models/partnered_signup.rb
```diff
@@ -6,6 +6,9 @@ class PartneredSignup < ApplicationRecord
   include PublicIdentifiable
   set_public_id_prefix :sup
 
+  include CountryEnumable
+  has_country_enum :country
```
garyhtou: Should this be:
```suggestion
  has_country_enum :owner_address_country
```

maxwofford:
Thanks for catching that @garyhtou! The previous code technically works, but we should probably update it from being a string to an int.

app/models/partnered_signup.rb
exu3: Ah thanks!

garyhtou:
yeah np! i think it might already be an int
image

maxwofford:
In that case, is this ok to merge @garyhtou @exu3?

garyhtou:
go for it!

PR #2305, opened 2022-02-11
Start collecting paper trail data

garyhtou:
Applied to all models that have meaningful changes.

This data is useful in case:
-  we want to add a feature later down the road that needs this data later
- use change logs to auditing


garyhtou:
@maxwofford are there any other models we should add/remove?

image

maxwofford:
I think this is good for now!

PR #2311, opened 2022-02-15
Add organized_by_hack_clubbers column to events

exu3:
This PR adds a column `organized_by_hack_clubbers` to the Events table.

Screen Shot 2022-02-14 at 20 01 12
Screen Shot 2022-02-14 at 20 00 38


db/schema.rb
```diff
@@ -884,6 +885,8 @@
     t.datetime "applicant_signed_at"
     t.datetime "completed_at"
     t.boolean "legal_acknowledgement"
+    t.text "organization_url"
+    t.text "organization_description"
```
exu3: Errr, I'm not sure why these lines changed but lmk if I should revert.

garyhtou: yeah, please see if you can revert those two lines. otherwise, this looks good!

garyhtou (approved):

MelSmith104:
this is perfect @exu3! good to merge :)

MelSmith104 (approved):

PR #2313, opened 2022-02-16
Fix textarea sizing bug

cjdenio:
All the textareas with `data-behavior="autosize"` are varying degrees of broken.

https://user-images.githubusercontent.com/34525547/154366922-9bbe6a36-e3ab-4704-bc7f-666a72d9ba94.mov


exu3 (approved):
Wack Flug Jank 

Thanks for catching this!

PR #2315, opened 2022-02-17
Support multiple Stripe signing secrets

cjdenio:
New credentials structure is as follows:

```diff
stripe:
  live:
    publishable_key: wahoo
    secret_key: yay
+   webhook_signing_secrets:
+     primary: signing secret for /stripe/webhook
+     donations: signing secret for /webhooks/donations

```


garyhtou:
@cjdenio Fyi, storing stripe secrets in rails creds is being deprecated. we're now storing them in the Partner model to support multiple stripe accounts (for Givebutter). would you mind setting up an encrypted column on the Partner model using lockbox?

See `Partner.stripe_api_key` as an example

https://github.com/ankane/lockbox

maxwofford (approved):
I'm cool with this change if it's working locally with stripe's test webhook events. When this merges please be ready to purchase something on your card after it deploys to ensure it still works properly. Also, let's add a comment about why we're skipping the signing secret if it's blank (a "TODO" is fine).

cjdenio:
I can confirm that cards + donations still work!

PR #2316, opened 2022-02-17
Name cannot be a blank space

exu3:
Currently, Bank lets you set your name to `" "` which seems to cause some issues and render the account in question unusable.

This PR adds a `pattern` attribute to the input which prevents names from starting with `" "`

https://hackclub.slack.com/archives/C02AN8HRS0H/p1645112884658499?thread_ts=1644962966.252649&cid=C02AN8HRS0H


garyhtou (commented):
Maybe also try adding
```ruby
validates :full_name, presence: true
```
above this: https://github.com/hackclub/bank/blob/25c47cac38620be575ad049041ad95fc5a4fd83e/app/models/user.rb#L51-L51

the validates presence will cause the form save to fail if `user.full_name.blank?` evaluates to true

the table at the bottom of https://blog.appsignal.com/2018/09/11/differences-between-nil-empty-blank-and-present.html is helpful for understanding what qualifies as `blank`/`present`.

---

also, I like the HTML field validation! it's nice to know it's invalid without having to save and reload the page.

garyhtou:
oh actually, don't validate presence yet. looks like there are 245 users who have a `blank` name
image

some of these are accounts generated via API (givebutter), others are normal user. hm

---

```ruby
User.all.map {|u| u.full_name.blank? ? {id: u.id, email: u.email} : nil}.compact!
```

garyhtou:
if name is blank, we should probably fill it with their email address. is there something better?

exu3:
> Maybe also try adding
> 
> ```ruby
> validates :full_name, presence: true
> ```

When I try it locally the form save fails & I get a `You were signed out. Please re-login.` I think it's because newly invited users have a blank name.

> oh actually, don't validate presence yet. looks like there are 245 users who have a blank name
> some of these are accounts generated via API (givebutter), others are normal user. hm

I think most of them are invited users who haven't signed in yet.

> if name is blank, we should probably fill it with their email address. is there something better?

I think it's alright to have it blank because it makes more sense for a new user who has yet to fill out their profile.

exu3:
https://github.com/hackclub/bank/pull/2316/commits/984b5635cd2b79713c39f28d83a25a6be8cc4f2d - https://hackclub.slack.com/archives/C02AN8HRS0H/p1644517093507069

maxwofford:
Should this be a validation on the `Users#update` method? A user's name _can_ be blank, but a user can _never_ update their name to be blank. We also direct users to the `Users#edit` method on signin if their name is blank (search for `@onboarding` in the User controller for examples).

garyhtou (approved):
Good work!

PR #2319, opened 2022-02-23
Add Link to Legacy Page

kunalbotla:
The Legacy "G Suite" page is now hidden on Bank but is still a needed page for deletion functionality. This PR adds a direct link to the legacy page for each Google Workspace from their current page.


app/models/g_suite.rb
```diff
@@ -66,6 +66,10 @@ def dns_check_url
     "https://nslookup.io/dns-records/#{domain}"
   end
 
+  def legacy_page_url
+    "https://bank.hackclub.com/g_suites/#{id}"
```
maxwofford: This should probably be a rails url helper to still work correctly in development/preview

kunalbotla:
oh I like this, but I think we should also link to the old page so we can view it if we need to.

exu3:
👍, Is it cool if I commit those changes from the screen recording here?

kunalbotla:
for sure! i'll make a few phrasing edits before final review but it should be mostly set from there

maxwofford (approved):
LGTM! I'm cool with this getting merged whenever you're ready

exu3:
@kunalbotla is this good to merge?

kunalbotla:
yep! merging

PR #2227, opened 2022-01-19
VS Code language support (via Solargraph)

garyhtou:
it's janky and not well supported by solargraph.

this adds onto #2182 by eliminating the need to attach to the docker container
Screen Shot 2022-01-19 at 1 33 22 AM

```
docker-compose -f docker-compose.yml -f docker-compose.solargraph.yml build;
docker-compose -f docker-compose.yml -f docker-compose.solargraph.yml up -d solargraph;
```

PR #2321, opened 2022-02-23
Documentation + Codespace setup scripts for Solargraph

exu3:
Solargraph seems to work quite well for me in Docker/Codespaces, so this PR adds the stuff from #2227 to the Codespace setup scripts. I tested it out a couple times, and everything seems to work.
Screen Shot 2022-02-22 at 20 47 25


README.md
```diff
@@ -115,6 +115,13 @@ env $(cat .env.docker) docker-compose run --service-ports web bundle exec rails
 env $(cat .env.docker) docker-compose run --service-ports web bundle exec rails s -b 0.0.0.0 -p 3000
 ```
 
+Run Solargraph in Docker: (You may also need to install the [Ruby Solargraph](https://marketplace.visualstudio.com/items?itemName=castwide.solargraph) VS Code extension)
```
garyhtou: Can we note that this is optional? And explain what it does?

codespace-config.sh
```diff
@@ -53,12 +53,17 @@ echo "
...
 echo "
 $(tput setaf 9)Hack Club Bank:$(tput sgr0) Step 7/7: Docker Database Setup"
 env $(cat .env.docker) docker-compose run --service-ports web bundle exec rails db:create db:migrate
 env $(cat .env.docker) docker-compose run --service-ports web pg_restore --verbose --clean --no-acl --no-owner -h db -U postgres -d bank_development latest.dump
 echo "$(tput setaf 9)Hack Club Bank:$(tput sgr0) $(tput setaf 10)Done$(tput sgr0)"
 
+echo "$(tput setaf 9)Hack Club Bank:$(tput sgr0) Step 8/7: Solargraph"
+env $(cat .env.docker) docker-compose -f docker-compose.yml -f docker-compose.solargraph.yml build
+echo "$(tput setaf 9)Hack Club Bank:$(tput sgr0) $(tput setaf 10)Done$(tput sgr0)"
+
```
garyhtou: Can we use a command line flag to activate this feature? I don't think we should force/automatically run solargraph. it should be an explicit choice since it takes some set up and might not work for everyone

https://stackoverflow.com/questions/2875424/correct-way-to-check-for-a-command-line-flag-in-bash

garyhtou: Same thing with `codespace-start.sh`

garyhtou (changes requested):
Yuhhh! i'm glad this is working for you!

codespace-config.sh
exu3: Good point!

Running the command with the flag `--with-solargraph` should activate this feature.

exu3:
I tested it out, and it seems to work:

```sh
$ codespace-config.sh --with-solargraph # sets up the Codespace with Solargraph
$ codespace-start.sh --with-solargraph # starts the dev server along with Solargraph
```

garyhtou (approved):
great! ty!

PR #2322, opened 2022-02-23
Add organized_by_hack_clubbers filter on admin events page

exu3:
This PR adds a `organized_by_hack_clubbers` filter on the admin events page, requested by @MelSmith104.


app/models/event.rb
```diff
@@ -38,6 +38,8 @@ class Event < ApplicationRecord
       .references(:canonical_transaction)
   }
   scope :not_funded, -> { where.not(id: funded) }
+  scope :organized_by_hack_clubbers, -> { where(organized_by_hack_clubbers: true) }
+  scope :not_organized_by_hack_clubbers, -> { where(organized_by_hack_clubbers: nil) }
```
```suggestion
  scope :not_organized_by_hack_clubbers, -> { where.not(organized_by_hack_clubbers: true) }
```

allows for `organized_by_hack_clubbers` to be `false` or `nil` — otherwise, we might miss some events that have `organized_by_hack_clubbers = false`

garyhtou (approved):
Amazing job! just a small change then good to merge

PR #2324, opened 2022-02-24
SMS signin is no longer gated

maxwofford:
- fixes for the onboarding flow breaking with an invalid phone number
- anyone can enable SMS auth (not just admins)
- the SMS verification process is mostly ERB
- the SMS verification form is a react component just used for submitting codes

Known issues to fix:
- After verifying successfully and clicking out of the verification modal, the settings page has an outdated section that still says "codes are sent to your email". This is fixed after a refresh, but the user doesn't know to refresh. Will work on a fix for that

@exu3 for the time being, could I get your review on the experience of it with a new account? Please also test out with invalid inputs (incorrect verification, incorrectly formatted phone number, etc). I want to know if there is anything unclear or that needs to change other then the aforementioned stale section on the edit page.

EDIT: forgot to mention, but once this is merged I want to make a ship post in the Slack about this, so it should be polished enough for general public use.

EDIT: @exu3 to retest verification with only 1 number, you can change a digit in your phone number, save, and change the digit back– we don't cache verified phone numbers.


maxwofford:
@garyhtou I forgot what else we needed to do to validate/update all user phone numbers, but I want to do that before this is merged. What was the context on that?

app/controllers/users_controller.rb
```diff
@@ -140,21 +150,26 @@ def complete_sms_auth_verification
...
-    authorize user
+    authorize current_user
     svc = UserService::EnrollSmsAuth.new(current_user)
-    if user.use_sms_auth
+    if current_user.use_sms_auth
       svc.disable_sms_auth
+      flash[:success] = "SMS signin turned off"
     else
       svc.enroll_sms_auth
+      flash[:success] = "SMS signin turned on"
```
exu3:
```suggestion
      flash[:success] = "SMS sign-in turned off"
    else
      svc.enroll_sms_auth
      flash[:success] = "SMS sign-in turned on"
```

app/views/users/edit.html.erb
```diff
@@ -79,14 +89,45 @@
...
+          <legend class="heading flex flex-row justify-start items-center">
+            <span class="h2">Login Codes</span>
+            <%= badge_for @user.use_sms_auth ? "SMS" : "Email", class: 'bg-muted' %>
+          </legend>
+          <% if @user.phone_number_verified %>
+            <% if @user.use_sms_auth %>
+              You are currently recieving sign-in codes via SMS. Email codes can still be sent as a backup.
+              <%= link_to "Turn off SMS sign-in codes.", toggle_sms_auth_users_path, method: :post %>
+            <% else %>
+              You are currently recieving sign-in codes via email.
```
exu3:
```suggestion
              You are currently receiving sign-in codes via SMS. Email codes can still be sent as a backup.
              <%= link_to "Turn off SMS sign-in codes.", toggle_sms_auth_users_path, method: :post %>
            <% else %>
              You are currently receiving sign-in codes via email.
```

exu3 (commented):
I'm getting a 500 error when I try to sign in on the staging deployment. #2335 ([Airbrake](https://heroku-app97991095.airbrake.io/projects/288439/groups/3196423098138852495?tab=overview))

The flow looks + feels great for the most part! Just a couple things:

- When I enter a phone number with an extra digit (not invalid enough to trigger the form validation), the button gets stuck on cursor-wait when I try to get a verification code.
  - https://cloud-l33gur28h-hack-club-bot.vercel.app/0155837355-eb6524b1-a92d-484a-baef-beaa171f1319.mov
- Is it possible to format the phone number better? Currently, it says "...sent a code to +11231234423." which is not very readable. (Maybe just the last 4 digits would be alright?)
- Could we add some styles to the error message in the modal? eg. `list-reset bg-error`
  - Screen Shot 2022-02-26 at 01 18 25
- Small thing, but "login code" and "sign-in code" are used interchangeably in the UI for this feature; I think it might make more sense to make it consistent and just use "login code".

maxwofford:
@exu3 Thanks for the review! I updated the error message styling & the login/signin/sign-in discrepancy. Cool if we merge this?

exu3:
I think it's cool to merge. Is this feature being gated to users on orgs that are `organized_by_hackclubbers` for beta testing?
Screen Shot 2022-03-01 at 10 40 29

maxwofford:
Based on convo in slack (screenshotted) I think it makes sense to enable for users with beta features flag then merge

maxwofford:
Merging this into https://github.com/hackclub/bank/tree/tan-add-better-sms-error-message to deal with merge conflicts, will do some triage/fixes in there before it goes into main

PR #2325, opened 2022-02-24
Invoice create form `min` amount (HTML field validation)

garyhtou:


exu3:
For some reason, I can still submit the form with `0` in the `item_amount` field.

I've attempted to fix that behavior here https://github.com/hackclub/bank/pull/2342/commits/a28d81cc99867fff3eb3d778f8cda31b8f4eb276 (don't think it fully works though)

PR #2327, opened 2022-02-24
Fix user unable to accept invite bug

garyhtou:
closes #910 (woah a very old bug)

The issue seems to be that Organizer Position Invites are only linked to a user when they sign in. This means that if
1. a user is already signed in
2. receives an invite
3. and tries to attempt it

then they will receive an error since `user` is `nil` on the invite.

This PR overrides the default `user` getter and dynamically associates the `user` to the invite object


garyhtou:
Local test of method:

Before:
image

Dynamically linking:
image

After:
image

PR #2332, opened 2022-02-25
Include sender full name in mailer

exu3:
Closes #2329


garyhtou:
Can you test this with one of the Partner's representative? (representatives are just users)

here's a list of them:
image

---

more info on reps if you're interested
https://github.com/hackclub/bank/blob/8810a5343c394cf746016d334c1519d5f9220dcd/app/models/partner.rb#L59-L67

exu3:
I tried the following in a Rails console, but it doesn't seem quite right. I'm not sure where to find what sort of value(s) go with `invite` or `@invite` or if I'm doing this right at all

```ruby
::OrganizerPositionInvitesMailer.notify(invite: 123).deliver_now
```

Pointers would be greatly appreciated

garyhtou:
It's definitely not easy to tell which type it takes, but if you look at https://github.com/hackclub/bank/blob/2afdbd66d6e63c5b3144dfe90dc734803549ac3c/app/mailers/organizer_position_invites_mailer.rb#L3-L10

you can see that it's calling `.email` and `.event` on `@invite`. That means it's gotta be an object. In this case, it's a `OrganizerPositionInvite`.

Try this instead:
```ruby
::OrganizerPositionInvitesMailer.notify(invite: OrganizerPositionInvite.find(123)).deliver_now
```

exu3:
Ah thanks for the link; any idea what I might be missing here?

```ruby
irb(main):013:0> ::OrganizerPositionInvitesMailer.notify(invite: OrganizerPositionInvite.find(1234)).deliver_
now
  OrganizerPositionInvite Load (0.5ms)  SELECT "organizer_position_invites".* FROM "organizer_position_invites" WHERE "organizer_position_invites"."id" = $1 LIMIT $2  [["id", 1234], ["LIMIT", 1]]
OrganizerPositionInvitesMailer#notify: processed outbound mail in 0.3ms
Traceback (most recent call last):
        3: from (irb):12
        2: from (irb):13:in `rescue in irb_binding'
        1: from app/mailers/organizer_position_invites_mailer.rb:4:in `notify'
ArgumentError (unknown keyword: :invite)
```

garyhtou:
ah so since the code is getting `invite` from `params` instead of as a method parameter, we need to pass the invite using `with`.

```ruby
OrganizerPositionInvitesMailer.with(invite: OrganizerPositionInvite.find(123)).notify.deliver_later
```

`param` (instead of method parameters) is sometimes used in mailers to get arguments because it makes them easier to call from controllers. In this case, we aren't calling from the controller, so the mailer doesn't have the context/arguments it needs. we can use `with` to provide it those arguments

exu3:
Ooh got it, thanks for the explanation.

User:
Screen Shot 2022-03-04 at 22 42 41

GB:
Screen Shot 2022-03-04 at 22 46 28

Dev:
Screen Shot 2022-03-04 at 22 48 11

app/views/organizer_position_invites_mailer/notify.html.erb
```diff
@@ -1,7 +1,7 @@
 <p>Hi there! 👋</p>
 
 <p>
-  You've been invited to help manage <strong><%= @invite.event.name %></strong>, which is running on Hack Club Bank, an online dashboard to store finances, raise money, receive grants, manage team emails, and so much more.
+  You've been invited by <strong><%= @invite.sender.full_name %></strong> to help manage <strong><%= @invite.event.name %></strong>, which is running on Hack Club Bank, an online dashboard to store finances, raise money, receive grants, manage team emails, and so much more.
```
garyhtou: "you've been invited by Givebutter to _help_ manage `<event name>`" sounds kinda weird since Givebutter users will be receiving this email when their org is approved or when they log into bank for the first time.

Thoughts on removing the word "help" or updating the copy so it works in the case it's a rep and a normal user doing the inviting?

garyhtou (changes requested):

app/views/organizer_position_invites_mailer/notify.html.erb
garyhtou: If you'd like, you can also do have a conditional here (if it's a rep, then show a different copy)!

exu3: > Thoughts on removing the word "help" or updating the copy so it works in the case it's a rep and a normal user doing the inviting?

Yeah, I think removing "help" => "You've been invited by Givebutter :sparkles: to manage Butter Foundation..." makes sense for both cases. 

When a Bank admin creates a new org & sends the initial invite, it would also say something like "You've been invited by Liv Cook to _help_ manage DevHacks...", so I think removing "help" might make more sense.

garyhtou (approved):
Looks good, thanks Ella!

PR #2342, opened 2022-03-02
Add HTML validation to forms

exu3:
Resolves https://heroku-app97991095.airbrake.io/projects/288439/groups/3220060474006270564?tab=overview and https://heroku-app97991095.airbrake.io/projects/288439/groups/3210960171084510984?tab=notice-detail

This PR adds HTML validation to the New Invoice form and Partnered Sup form.

I think this makes for a better experience because users won't be able to submit the form without filling out the required fields. Currently, the form can be submitted with empty fields, but if they only miss one field, the entire form is reset.

---

to do:
- [x] Add HTML validation to Partnered Signup form - this also might help with the issue where the page reloads if the values don't satisfy the validation.
- [x] Invoices form
- [x] Checks form


exu3:
The `min` attribute added in #2325 doesn't seem to work as I can still submit the form with the dollar amount, 0. https://github.com/hackclub/bank/pull/2342/commits/a28d81cc99867fff3eb3d778f8cda31b8f4eb276 is my attempt at fixing that behavior; though it doesn't fully work.

app/assets/javascripts/ui.js
```diff
@@ -151,6 +151,19 @@ $(document).on('turbo:load', function () {
...
+  $('input[data-behavior~=enforce_min_max]').on({
+    change: function (e) {
+      if(e.value != "") {
+        if(parseFloat(e.value) < parseFloat(e.min)){
+          e.value = e.min;
+        }
+        if(parseFloat(e.value) > parseFloat(e.max)){
+          e.value = e.max;
+        }
+      }
```
garyhtou: just a small suggestion here: try using a guard clause

https://betterprogramming.pub/refactoring-guard-clauses-2ceeaa1a9da

exu3: Turns out the validation wasn't working because it was a `text_field`; I changed it to a `number_field` so the `min` attribute works as intended.

app/views/checks/_new_form.html.erb
```diff
@@ -20,10 +20,13 @@
       <%= form.number_field :amount,
         placeholder: '500.00',
         value: (@check.amount.nil? ? nil : (@check.amount.to_f / 100)),
+        disabled: @event.balance == 0,
```
garyhtou: `<=0` bc there are some negative balance events

garyhtou: also add a message about insufficient funds if form is disabled

exu3: https://github.com/hackclub/bank/pull/2342/commits/cdd2bbc553065dbbf242e6a6a3164dfa1702290d adds a message about insufficient funds

garyhtou (approved):
thx!

PR #2343, opened 2022-03-02
Initial work on autolinking hcb codes in comments

maxwofford:
Comments now automatically link HCB-code urls to their corresponding hcb page, and shorten their description to just the hcb code:

My comment:
Screen Shot 2022-03-02 at 01 12 02

Previously the whole link would show:

Screen Shot 2022-03-02 at 01 08 45

Now:

Screen Shot 2022-03-02 at 01 11 49

For future exploration, we can also pull additional info about the HCB code in question, for example the type of HCB:

Screen Shot 2022-03-02 at 01 26 36

EDIT: poorly implemented example of tooltip with extra details

Screen Shot 2022-03-02 at 01 43 19


maxwofford:
Screen Shot 2022-03-02 at 14 00 58

app/services/markdown_service.rb
```diff
@@ -4,22 +4,84 @@ class MarkdownService
...
-      "<img src=\"#{link}\" alt=\"#{alt_text}\" title=\"#{title}\" width=\"100%\" />"
+      image_tag link, title: title || alt_text,
+                      alt:   alt_text,
+                      style: 'width: 100%;'
+    end
+
+    private
+
+    def try_event_autolink(link)
+      found_event = link.match(/bank.hackclub.com\/([\w|-]*)$/)
```
garyhtou: Get host from Rails

app/services/markdown_service.rb
```diff
@@ -4,22 +4,84 @@ class MarkdownService
...
+      if found_event
+        event_id = found_event[1]
+        event = Event.find(event_id)
```
garyhtou: `find` can throw an exception if the Event isn't found. Try wrapping this in a `rescue` or using a `find_by` method (which doesn't throw exceptions)

app/services/markdown_service.rb
```diff
@@ -4,22 +4,84 @@ class MarkdownService
...
+      else
+        return nil
+      end
     end
 
+    def try_card_autolink(link)
+      found_card = link.match(/stripe_cards\/(\d*)/)
+      if found_card
+        card_id = found_card[1]
+        card = StripeCard.find(card_id)
```
garyhtou: Same comment as `Event.find`

app/services/markdown_service.rb
```diff
@@ -4,22 +4,84 @@ class MarkdownService
...
+      end
+    end
+
+    def try_hcb_autolink(link)
+      # TODO: A design flaw of this is that any link will match:
+      # ie. example.com/hcb/asd123
+      # ie. example.com/hcb/asd123#comment-1234
+      # ^ (non-hackclub domains)
+      found_link = link.match(/\/hcb\/(\w{5,6})\#comment-(\w{5,6})/)
+      found_link ||= link.match(/\/hcb\/(\w{5,6})/)
```
garyhtou:
```ruby
\/hcb\/(\w{5,6})(?:\#comment-(\w{5,6}))?
```
you can also prefix the regex with the Rails host to only match URLs from Bank

app/services/markdown_service.rb
```diff
@@ -4,22 +4,84 @@ class MarkdownService
...
+
+      if found_link
+        hcb_code = found_link[1]
+        comment = found_link[2]
+        hcb = HcbCode.find(hcb_code)
+        link_to "HCB-#{hcb.short_code} (#{'comment on ' unless comment.blank?}#{hcb.type.to_s.humanize})",
```
garyhtou: I think this might result in "comment on Unknown".

maybe create a method called `humanized_type` and handle the case there `type == :unknown` and have it return "Bank Account Transaction" (that's what they're called in the UI https://github.com/hackclub/bank/blob/master/app/views/canonical_transactions/_canonical_transaction.html.erb#L39-L39)

garyhtou: also same comment as `Event.find`

PR #2345, opened 2022-03-02
Change `bank-alerts@hackclub.com` to `bank@`

exu3:
Follow up question: what's bank-alert@hackclub.com? Is it checked/used?

Screen Shot 2022-03-02 at 11 00 04


maxwofford:
Screen Shot 2022-03-02 at 14 03 24

Screen Shot 2022-03-02 at 14 03 29

Go for merge if this is expected behavior

exu3:
Ah I think bank@hackclub.com might be a better place? bank-alert@ doesn't seem to reach any humans based on the way it's currently set up (it's a Google group with no members)

We could set up an email filter in Helpscout for these.

Screen Shot 2022-03-02 at 11 07 10

maxwofford (approved):
Awesome, thanks for finding this!

PR #2347, opened 2022-03-03
Remove unused notify-slack buildpack

maxwofford:
It looks like this buildpack has been failing on every build for the past couple months, even though we're getting slack notifications from new builds:

Screen Shot 2022-03-02 at 21 05 58

Seems like this buildpack isn't needed


garyhtou (approved):
we're currently getting deploy notifs through the slack github app

maxwofford:
This code change seems to be overwritten by Heroku's web interface, so I'm removing the buildpack there

Screen Shot 2022-05-10 at 12 17 46

PR #2348, opened 2022-03-03
Interface Instances of `Docusign` → `DocuSign`

kunalbotla:
there are instances of "DocuSign" being [incorrectly capitalized](https://www.docusign.com/IP/trademark-brand-guide) as "Docusign" in user & admin-facing pages. this PR changes those instances. this doesn't impact any controllers


kunalbotla:
I'm getting an error on Zeitwerk's check. I'm not sure if this is an error with the code itself or the check since it should be expecting `DocuSignController` now

image

cjdenio:
It's a Rails autoloading thing; `docusign_controller.rb` needs to define `DocusignController` (with that exact capitalization): https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#project-structure

kunalbotla:
oh I see, that clarifies it. thanks, @cjdenio!

I'm going to have this PR only make visual changes to the user and admin side of the dashboard

garyhtou (approved):
Thanks!

PR #2350, opened 2022-03-03
Disable TX notifications email notifs to Bank admins

exu3:
This PR disables email notifications to the Bank team for Stripe authorizations and CPTs; I don't think this notification is necessary because the same information can be accessed from the show page of the respective TX. 

(HS-2659 is an example of the sort of email being sent)


garyhtou:
i agree with this! make sure to also remove the calls to those mailer methods (the places that request for an email to be sent)

example:
https://github.com/hackclub/bank/blob/2afdbd66d6e63c5b3144dfe90dc734803549ac3c/app/services/stripe_authorization_service/create_from_webhook.rb#L28-L28

exu3:
Should be good! 
Screen Shot 2022-03-03 at 15 57 10
Screen Shot 2022-03-03 at 15 56 56

PR #2356, opened 2022-03-04
Remove validation for full_name

exu3:
https://hackclub.slack.com/archives/C02AN8HRS0H/p1646386456780789

(have not tested this yet)


garyhtou:
If this is blocking for you, then go ahead and merge!

I do wanna see if we can try using custom contexts: https://guides.rubyonrails.org/active_record_validations.html#on

exu3:
Hmm I think this was an edge case; I can't seem to reproduce this on other accounts with a `blank` name.
> I do wanna see if we can try using custom contexts: [guides.rubyonrails.org/active_record_validations.html#on](https://guides.rubyonrails.org/active_record_validations.html#on)

This validation should work as intended with `on: :update` right? or is there something else that might be better?

PR #2359, opened 2022-03-04
Initial work on user-created disbursements

maxwofford:
This makes a couple changes to disbursements:
- [x] Feature is gated to users with beta-features enabled
- [x] Event transfer page filters also filter disbursements
- [x] Stats on event transfer page include disbursements
- [x] Renamed "disbursements" -> "transfers"/"HCB transfers" on user interface
- [x] Users can now make disbursements between orgs they're a part of
- [x] Add validations to disbursements (can't go over budget, etc.)
- [x] Disbursement form only shows related events (admins can still see all events)
- [x] Disbursements have HCB codes (wasn't working before)
- [x] Redesign HCB code "show page" for disbursements & make sure they work in transparency mode
- [x] Disbursements have "requestor" & "approver" (associated users)
   - [x] Admin-created disbursements shouldn't require manual approval
- [x] Admin tools for reviewing user-created disbursements
   - [x] Disbursement automation only runs on approved disbursements
 
Screen Shot 2022-03-08 at 10 34 46

Screen Shot 2022-03-08 at 10 35 00

Screen Shot 2022-03-08 at 10 35 13

Screen Shot 2022-03-08 at 10 35 30


maxwofford:
Ok, I think this is ready to add in. The disbursement changes are gated to beta-enabled users, but there are changes to
- disbursement scopes (which might affect the nightly job) cc @garyhtou 
- event transfer overview page ("Send" button on the navbar) cc @exu3

app/views/disbursements/_form.html.erb
```diff
@@ -1,34 +1,48 @@
...
         placeholder: '500.00',
         step: 0.01,
+        min: 0.01,
         required: true %>
     </div>
   </div>
 
+  <div class="field">
+    <%= form.label :name, "What is the transfer for?" %>
+    <%= form.text_field :name, placeholder: 'Donating extra funds to another organization', required: true %>
```
exu3:
```suggestion
    <%= form.text_field :name, placeholder: 'Donating extra funds to another organization', maxlength: 60, required: true %>
```

app/views/disbursements/_form.html.erb
```diff
@@ -1,34 +1,48 @@
...
+                                required: true
+                                ) %>
   </div>
 
   <%= form.label :amount, 'Amount' %>
   <div class="field">
     <div class="flex items-center">
       <span class="bold muted" style="width: 1rem;">$</span>
-      <%= form.number_field :amount, value: (@disbursement.nil? ? nil : (@disbursement.amount.to_f / 100)),
+      <%= form.number_field :amount, value: (@disbursement&.amount.nil? ? 10 : (@disbursement.amount.to_f / 100)),
```
exu3: Just curious, why is the default value $10?

app/views/disbursements/_form.html.erb
```diff
@@ -1,34 +1,48 @@
...
     </div>
   </div>
 
+  <div class="field">
+    <%= form.label :name, "What is the transfer for?" %>
+    <%= form.text_field :name, placeholder: 'Donating extra funds to another organization', required: true %>
+    <span class="muted">This is to help Hack Club Bank keep record of our transactions. It will be used as the memo on your transaction list.</span>
+  </div>
+
   <div class="actions tooltipped tooltipped--n inline-block" aria-label="The Hack Club Bank staff will make the disbursement">
```
exu3:
```suggestion
  <div class="actions tooltipped tooltipped--n inline-block" aria-label="The Bank team will make the transfer">
```
disbursement -> transfer
Hack Club Bank staff -> Bank team

app/views/disbursements/new.html.erb
```diff
@@ -1,6 +1,25 @@
 <% title 'Schedule a disbursement' %>
 <% page_sm %>
+<%= render 'events/nav' if @event %>
 
-<h1>Schedule a disbursement</h1>
+<h1>Transfer funds</h1>
+
+<p class="h5 muted mt0">You can transfer funds between any HCB organization you are a part of.</p>
+<p class="h5 muted mt0">Transfers take 2-3 business days– they don't send over weekends.</p>
```
exu3:
```suggestion
```

This is redundant with the copy in the warning message below.

app/views/disbursements/new.html.erb
```diff
@@ -1,6 +1,25 @@
 <% title 'Schedule a disbursement' %>
 <% page_sm %>
+<%= render 'events/nav' if @event %>
 
-<h1>Schedule a disbursement</h1>
+<h1>Transfer funds</h1>
+
+<p class="h5 muted mt0">You can transfer funds between any HCB organization you are a part of.</p>
```
exu3:
```suggestion
<p class="h5 muted mt0">You can transfer funds between any organizations you are a part of.</p>
```

app/views/disbursements/new.html.erb
```diff
@@ -1,6 +1,25 @@
...
+<p class="h5 muted mt0">Transfers take 2-3 business days– they don't send over weekends.</p>
+
+<div class="card border b--warning pb0 mt3 mb3">
+  <div class="flex items-center warning">
+    <%= inline_icon 'info', size: 32, class: 'mr1' %>
+    <p class="bold mt0 mb0">Important info about transfers:</p>
+  </div>
+  <ul class="mt1 mb2">
+    <li><strong>Transfers take 2-3 business days</strong>—they don't send over weekends.</li>
+    <li>Once a transfer is sent, it will leave your account balance, but won't immediately be added to the recipient organization's balance.</li>
```
exu3:
```suggestion
    <li>Once a transfer is sent, it will leave your account balance but won't immediately be added to the recipient organization's balance.</li>
```

app/views/events/transfers.html.erb
```diff
@@ -19,10 +22,11 @@
           data: { behavior: "modal_trigger", modal: "paypal_transfer" } do %>
           PayPal
         <% end %>
+        <%= link_to "Hack Club Bank organization", new_event_disbursement_path %>
```
exu3: Screen Shot 2022-03-10 at 12 32 57

I think it's a bit confusing to have these 2 buttons that both do the same thing (create a disbursement). Looks like the paths are different though, so I'm not sure if this was intended. I think the `new_event_disbursement_path` makes more sense as it's consistent with the other transfer options. 

Would a better name for this menu option be something like "HCB Transfer" or "HCB Organization" so it's similar in length to the other menu options?

Also, we should add the beta feature flag to this menu option right?

app/views/hcb_codes/_disbursement.html.erb
```diff
@@ -1,12 +1,18 @@
...
+<% @source = @hcb_code.disbursement.source_event %>
+
 <article class="card pb0 mt3 mb3">
   <h2 class="h2 mt0 border-none">
+    <%= @source.name %>
     <span class="regular muted">
-      <%= @hcb_code.disbursement.fulfilled? ? 'Disbursed' : (@hcb_code.disbursement.pending? ||
-        @hcb_code.disbursement.processed?) ? 'Disbursing' : 'Disbursement of' %>
+      <%= @hcb_code.disbursement.fulfilled? ? 'transfered' : (@hcb_code.disbursement.pending? ||
+        @hcb_code.disbursement.processed?) ? 'transferring' : 'transfer of' %>
```
exu3:
```suggestion
      <%= @hcb_code.disbursement.fulfilled? ? 'transferred' : (@hcb_code.disbursement.pending? ||
        @hcb_code.disbursement.processed?) ? 'transferring' : 'transfer of' %>
```

app/views/admin/disbursements.html.erb
```diff
@@ -40,19 +40,23 @@
   </thead>
   <tbody>
     <% @disbursements.each do |disbursement| %>
-      <tr class="<%= "admin-bg-pending" if disbursement.pending? %>">
+      <tr class="<%= "admin-bg-transit" if disbursement.pending? %><%= "admin-bg-pending" if disbursement.reviewing?%>">
```
exu3:
```suggestion
      <tr class="<%= "admin-bg-transit" if disbursement.pending? %><%= "admin-bg-pending" if disbursement.reviewing? %>">
```

very small thing :P

app/controllers/disbursements_controller.rb
```diff
@@ -16,26 +16,45 @@ def show
...
+      source_event_id: disbursement_params[:source_event_id],
+      amount: disbursement_params[:amount],
+      requested_by_id: current_user.id
+    }
+    disbursement = DisbursementService::Create.new(attrs).run
+
+    flash[:success] = "Transfer successfully requested."
+
+    if current_user.admin?
       redirect_to disbursements_path
```
exu3:
```suggestion
      redirect_to disbursements_admin_index_path
```

https://bank.hackclub.com/admin/disbursements instead of https://bank.hackclub.com/disbursements

exu3 (commented):
This is awesome, nice work! Sorry for the long review :P I also have a little bit more I'll add in a few

app/views/events/transfers.html.erb
```diff
@@ -85,17 +89,23 @@
...
-            <%= d.name if organizer_signed_in? %>
-            <%= "(redacted)" unless organizer_signed_in? %>
+            <% if organizer_signed_in? %>
+              <% if d.is_a?(Disbursement) and organizer_signed_in?(d.destination_event) %>
+                <%= link_to d.destination_event.slug, d.destination_event %>
+              <% end %>
+              <%= d.name %>
+            <% else %>
+              <span class="muted">(Sign in to view)</span>
+            <% end %>
```
exu3:
```suggestion
            <% if organizer_signed_in? %>
              <% if d.is_a?(Disbursement) and organizer_signed_in?(d.destination_event) %>
                <%= d.destination_event.name %>:
              <% end %>
              <%= d.name %>
            <% else %>
              <span class="muted">(Sign in to view)</span>
            <% end %>
```

Case: If Wahoo Hacks is repaying a front to Hack Club Bank & an admin initiates the disbursement. The link would go to bank.hackclub.com/bank which is not a transparent. 

(Alternatively, only link to destination event, if event is_public?)

app/views/admin/disbursement_process.html.erb
```diff
@@ -0,0 +1,55 @@
...
+
+<%= form_with(model: nil, local: true, url: disbursement_approve_admin_path(@disbursement), method: :post) do |form| %>
+  <%= form.submit 'Approve disbursement' %>
+<% end %>
+
+<%= form_with(model: nil, local: true, url: disbursement_reject_admin_path(@disbursement), method: :post) do |form| %>
+  <%= form.submit 'Reject disbursement',
+    class: 'admin-bg-orange',
+    data: { confirm: "Are you sure you want to reject this disbursement?" } %>
+<% end %>
```
exu3:
```suggestion
<h3>Instructions</h3>

<% if @disbursement.rejected? || @disbursement.pending? %>
  <p>Nothing more to do here. ✨</p>
<% else %>

	  <p>This is a human review check.</p>
	  
	  <%= form_with(model: nil, local: true, url: disbursement_approve_admin_path(@disbursement), method: :post) do |form| %>
	    <%= form.submit 'Approve disbursement' %>
	  <% end %>
	  
	  <%= form_with(model: nil, local: true, url: disbursement_reject_admin_path(@disbursement), method: :post) do |form| %>
	    <%= form.submit 'Reject disbursement',
	      class: 'admin-bg-orange',
	      data: { confirm: "Are you sure you want to reject this disbursement?" } %>
	  <% end %>
<% end %>
```

Don't show process buttons if disbursement is pending or rejected.

app/views/disbursements/_form.html.erb
maxwofford: Uhh, arbitrary value from debugging. Would you want this sort of thing to start blank? or another value?

exu3:
- Maybe don't show `hidden` events in the dropdown; they're usually zeroed out before being `hidden` anyways so I don't think there will be a need to disburse funds in/out of `hidden` events
- Currently it seems like the source event is the default value for the "To..." field
	- If I initiate a transfer from [Wahoo Hacks](https://bank--pr-2359.herokuapp.com/wahoo-hacks/disbursements/new) (source event) to DevHacks (recipient), I think I would expect the "From..." field to be the source event by default.

app/views/disbursements/_form.html.erb
exu3: I think blank would make sense

app/views/events/transfers.html.erb
garyhtou: ^ what yeah agree with this. another name idea: "Internal Transfer"

app/helpers/sessions_helper.rb
```diff
@@ -42,7 +42,7 @@ def current_user=(user)
     @current_user = user
   end
 
-  def organizer_signed_in?
+  def organizer_signed_in?(event = @event)
     @organizer_signed_in ||= ((signed_in? && @event&.users&.include?(current_user)) || admin_signed_in?)
```
garyhtou: Should `event` be the one from the method parameter?

```suggestion
  def organizer_signed_in?(event = @event)
    @organizer_signed_in ||= ((signed_in? && event&.users&.include?(current_user)) || admin_signed_in?)
```

app/models/disbursement.rb
```diff
@@ -4,8 +4,16 @@ class Disbursement < ApplicationRecord
...
+  include Commentable
+
+  has_paper_trail
+
+  belongs_to :fulfilled_by, class_name: "User", optional: true
+  belongs_to :requested_by, class_name: "User", optional: true
+
+  belongs_to :destination_event, foreign_key: 'event_id', class_name: "Event", inverse_of: 'incoming_disbursements'
+  belongs_to :source_event, class_name: "Event", inverse_of: 'outgoing_disbursements'
   belongs_to :event
```
garyhtou: I see what you're using `event_id` for `destination_event`. Should we remove this and change all usage of `.event` to `.destination_event`?

image

app/views/disbursements/_form.html.erb
```diff
@@ -1,34 +1,48 @@
...
-    <%= form.label :name %>
-    <%= form.text_field :name, placeholder: 'ex. GitHub Grant', required: true %>
+    <%= form.label :event_id, 'To...' %>
+    <%= form.collection_select(:event_id,
+                                @allowed_events,
+                                :id,
+                                current_user.admin? ? :admin_dropdown_description : :disbursement_dropdown_description,
+                                {prompt: 'Select one…'},
+                                required: true
+                                ) %>
```
garyhtou: I think the collection select for `:event_id` has an unintended effect for https://bank--pr-2359.herokuapp.com/hq/disbursements/new

It defaults the destination event to the current event because it's id is in the path
image

app/views/disbursements/_form.html.erb
```diff
@@ -1,34 +1,48 @@
...
-    <%= form.label :source_event_id, 'Source event' %>
-    <%= form.collection_select(:source_event_id, Event.reorder(Event::CUSTOM_SORT), :id, :admin_dropdown_description, include_blank: 'Select one…') %>
+    <%= form.label :source_event_id, 'From...' %>
+    <%= form.collection_select(:source_event_id,
+                                @allowed_events,
+                                :id,
+                                current_user.admin? ? :admin_dropdown_description : :disbursement_dropdown_description,
+                                {prompt: 'Select one…'},
+                                required: true
+                                ) %>
```
garyhtou: The UX feels a bit off. You're currently on HQ. and click the button to make a transfer, but then you're asked which event it should come out of.

Should we lock in the source event to the current event?

app/views/events/transfers.html.erb
garyhtou: To add to this, I think we should avoid using HCB bc of name collisions, we already have many things branded/named HCB

app/views/events/transfers.html.erb
```diff
@@ -85,17 +89,23 @@
       <% @transfers.each do |d| %>
         <tr>
           <td>
-            <%= d.is_a?(Check) ? "CHECK" : "ACH" %>
+            <%= d.is_a?(Check) ? "CHECK" : d.is_a?(AchTransfer) ? "ACH" : "HCB" %>
```
garyhtou:
```suggestion
            <%= d.is_a?(Check) ? "CHECK" : d.is_a?(AchTransfer) ? "ACH" : "TRANSFER" %>
```
there's probably a shorter name we could use. maybe `XFER` with a tooltip?

app/views/events/transfers.html.erb
```diff
@@ -85,17 +89,23 @@
...
           </td>
           <td>
-            <%= d.name if organizer_signed_in? %>
-            <%= "(redacted)" unless organizer_signed_in? %>
+            <% if organizer_signed_in? %>
+              <% if d.is_a?(Disbursement) and organizer_signed_in?(d.destination_event) %>
+                <%= link_to d.destination_event.slug, d.destination_event %>
+              <% end %>
+              <%= d.name %>
+            <% else %>
```
garyhtou: untested

```suggestion
            <% if organizer_signed_in? %>
              <% if d.is_a?(Disbursement) %>
                <strong>
                  <% if d.destination_event.is_public? || organizer_signed_in?(d.destination_event) %>
                    <%= link_to d.destination_event.name, d.destination_event %>
                  <% else %>
                    <%= d.destination_event.name %>
                  <% end %>
                </strong>
                for
              <% end %>
              <%= d.name %>
            <% else %>
```

app/views/events/transfers.html.erb
garyhtou: https://github.com/hackclub/bank/pull/2359#discussion_r824353500

app/views/events/transfers.html.erb
maxwofford: Thanks for catching that! I just renamed to "transfer" to have it fit nicely with the others. Button feels redundant, but I want the new feature to be in a somewhat visible area for beta, so maybe we can keep both for now?

app/views/disbursements/_form.html.erb
maxwofford: That's a great idea– just locked source event and now destinations are only non-hidden transparent orgs that you're a part of

app/models/disbursement.rb
maxwofford: Hmm.... yeah, there were a couple refactoring things like this that came to mind while working on this–

- disbursements have `event` & `source_event` (`event -> `destination_event`)
- event#transfer is a namespace conflict with disbursements renamed to "transfers"-- maybe we move to event#send?
- `HCB.find`, `HCB.find_by_id`, `HCB.find_by_short_code` etc. when I actually needed `HCB.find_by_hashid`

Is there a good example of moving column names I can reference? I tried doing it how I'm used to doing it, but strong migrations yelled at me.

I think I might go forward with merging this w/o blocking on the refactor, but I def want to fix this.

maxwofford:
Quick update: got through the comments (TYSM @exu3 @garyhtou!). I gated the feature to only events with beta features enabled, and destination events owned by the current user which are in transparency mode & not hidden. This isn't just to make it exclusive– I expect there are a couple instances of borked link permissions (you make a disbursement, but then can't view it).

app/controllers/disbursements_controller.rb
```diff
@@ -16,26 +16,51 @@ def show
...
+
+    @allowed_source_events = if current_user.admin?
+                               Event.all
+                             else
+                               [@source_event]
+                             end
+    @allowed_destination_events = if current_user.admin?
+                                    Event.all
+                                  else
+                                    current_user.events.not_hidden.transparent.where.not(id: @source_event.id)
```
garyhtou: this is non-blocking for merge.
im confused why we're filtering by transparent. should that be beta_features_enabled?

garyhtou (approved):
this looks good to merge in!

app/controllers/disbursements_controller.rb
maxwofford: Yeah! The thinking on beta_features_enabled is that from the user's perspective this is a feature for your org to send money to other orgs. So I figured it'd make sense if only the source org needs to be in the beta.

For transparent, that's kinda just an impulse decision rn. I'd like to eventually switch this to maybe all transparent orgs + orgs the current user is in, so things like execute big or project boom can collect donations from other orgs.

garyhtou: > I'd like to eventually switch this to maybe all transparent orgs + orgs the current user is in, so things like execute big or project boom can collect donations from other orgs.

Just a small thought about that. This would "expose" a list of all transparent events running on Bank. I personally think that's something we should avoid. Although yes these orgs are transparent, I think there's a difference between allowing transparency vs "prompting" their transparency

A workaround could be updating donation pages to include a "transfer" option if the user is signed in and has orgs with this feature enabled. This would create a disbursement rather than a stripe donation

PR #2360, opened 2022-03-05
Link event url in invoice_payouts notification email

exu3:
Closes #2339
Screen Shot 2022-03-05 at 00 11 40


exu3:
Screen Shot 2022-03-05 at 00 21 01 🤔

PR #2361, opened 2022-03-05
Update Google Workspace notification mailer to Bank Ops

exu3:
Screen Shot 2022-03-10 at 00 03 36


exu3:
This is only admin-facing & I tested it locally, so I'm going to go ahead and merge.

PR #2362, opened 2022-03-05
Log out of all sessions except the current session

exu3:
https://hackclub.slack.com/archives/CN523HLKW/p1646405305155659

Screen Shot 2022-03-05 at 02 25 30

---

This PR:


app/controllers/users_controller.rb
```diff
@@ -78,7 +78,7 @@ def logout
 
   def logout_all
     sign_out_of_all_sessions
-    redirect_to root_path
+    redirect_to edit_user_path(current_user), flash: { success: "Success" }
```
garyhtou: love this!

garyhtou: This isn't really related to this pr, but could you add a flash to `logout_session`'s redirect too?

garyhtou (approved):
Good work here!

app/controllers/users_controller.rb
exu3: Screen Shot 2022-03-07 at 13 37 52

I think there's already a flash

garyhtou: whoops! you're right! I see it on line 94 now. thanks Ella!

PR #2364, opened 2022-03-09
Validation for ACH & check sending

exu3:
https://hackclub.slack.com/archives/GBXBFEED9/p1646751093527939?thread_ts=1646751022.179469&cid=GBXBFEED9

This PR adds validation to the `amount` field for ACH transfers and checks, and a better error flash.

Screen Shot 2022-03-09 at 01 26 30


app/views/ach_transfers/_form.html.erb
```diff
@@ -36,7 +36,9 @@
   <div class="field">
     <div class="flex items-center">
       <span class="bold muted" style="width: 1rem;">$</span>
-      <%= form.number_field :amount, value: (@ach_transfer.nil? ? nil : (@ach_transfer.amount.to_f / 100)), placeholder: '500.00', step: 0.01, required: true %>
+      <%= form.number_field :amount, 
+      value: (ach_transfer.amount.nil? ? nil : (ach_transfer.amount.to_f / 100)),
```
garyhtou: just wondering, should `ach_transfer` be `@ach_transfer`? (same with the `ach_transfer` in the `form_with` and `form_errors`)

if everything works properly, then it should be alright

app/views/ach_transfers/_form.html.erb
```diff
@@ -36,7 +36,9 @@
...
+      placeholder: '500.00', step: 0.01, min: 1, required: true %>
```
garyhtou: looks like we're validating `amount >= 1` here, but `amount > 0` in the model

app/views/ach_transfers/_form.html.erb
exu3: Good catch, yeah, I don't think `form_errors` was working. Fixed in https://github.com/hackclub/bank/pull/2364/commits/825b889cc906559a10c6f302c3fe5b78af18049f

garyhtou (approved):

PR #2370, opened 2022-03-12
Fix Transfers search bar

exu3:
https://heroku-app97991095.airbrake.io/projects/288439/groups/3227510214805387636?tab=notice-detail

This PR fixes the search bar on the Transfers page. Transfers (ACH, Check, Disbursements) are searchable. Currently in production, the search bar throws a 500 error 

Try https://bank--pr-2370.herokuapp.com/hq/transfers?q=bank

https://user-images.githubusercontent.com/72365100/158008142-78fcb181-596e-46e2-accf-53a85aec64a6.mov


app/models/disbursement.rb
```diff
@@ -2,7 +2,7 @@
 
 class Disbursement < ApplicationRecord
   include PgSearch::Model
-  pg_search_scope :search_name, against: [:name]
+  pg_search_scope :search_name, against: [:name, :source_event_id, :event_id]
```
garyhtou: i'm curious as to why we're searching by ids here

garyhtou (approved):
looks good to merge 👍

app/models/disbursement.rb
exu3: Oops yeah I don't think that's needed. Thanks for the review :p

PR #2371, opened 2022-03-12
UI Improvements on Transfers page 

exu3:
https://bank--pr-2371.herokuapp.com/hq/transfers

- Memo is truncated so all fields are visible without horizontal scroll
- TYPE column is an icon instead of text (tooltipped with the type of transfer)
  - ACH and Check use `bank-account` & Disbursements use the same icon as the disbursements icon for the transaction partial `door-leave`
- Only show filter/search tools when signed in
  - Recipients aren't visible in transparency mode, so there's not much to search for.

Screen Shot 2022-03-12 at 00 11 52

<details>
<summary>Previously:</summary>

Screen Shot 2022-03-12 at 00 12 26

</details>


garyhtou:
I can review this next week. Or, this looks okay to merge if it's blocking on anything

---

one thing is that it might be nice to abstract out the icon logic (e.g, `d.is_a?(Check)`) since I think we're repeating this in multiple places

additional thought: if we want to tighten up on security, then in the controller before we filter the ach/checks/transfers, etc. we should check if the organizer is signed in. otherwise, anyone can still search by manually adding query params to the URL

app/views/events/transfers.html.erb
```diff
@@ -88,16 +90,28 @@
     <tbody>
       <% @transfers.each do |d| %>
         <tr>
-          <td>
-            <%= d.is_a?(Check) ? "CHECK" : d.is_a?(AchTransfer) ? "ACH" : "TRANSFER" %>
+          <td class="transaction__icon">
+            <% if d.is_a?(Check) %>
+              <span class="tooltipped tooltipped--e" aria-label="Check">
+                <%= inline_icon 'bank-account' %>
```
maxwofford: Maybe `email` or `email-fill` for this (on icons.hackclub.com) & "Mailed check" for the tooltip?

maxwofford (approved):
This is looking great! merge when ready!

PR #2372, opened 2022-03-14
Don't show shipping if card status is `shipped`

exu3:
`shipped` is the final status of cards shipped without tracking. ([Relevant section from Stripe docs](https://stripe.com/docs/api/issuing/cards/object#issuing_card_object-shipping-status)) Currently, the cards page shows the shipping info unless the card's status is `delivered` (which isn't possible given the way that physical cards are shipped currrently.)

With this PR, the shipping info won't be visible if the card's status is `shipped`.

Fixes #2022


exu3:
Hmm I think a another way to fix this might be to only show the shipping info for the card if the card `eta` < `Time.now`. Not sure which would be better.

app/models/stripe_card.rb
```diff
@@ -9,7 +9,7 @@ class StripeCard < ApplicationRecord
   scope :canceled, -> { where(stripe_status: "canceled") }
   scope :frozen, -> { where(stripe_status: "inactive") }
   scope :active, -> { where(stripe_status: "active") }
-  scope :physical_shipping, -> { physical.includes(:user, :event).reject { |c| c.stripe_obj[:shipping][:status] == "delivered" } }
+  scope :physical_shipping, -> { physical.includes(:user, :event).reject { |c| c.stripe_obj[:shipping][:status] == "delivered" || c.stripe_obj[:shipping][:status] == "shipped" } }
```
garyhtou: I think we might be able to use the `shipping_has_tracking?` method

See if something like this would work:
```ruby
scope :physical_shipping, -> { physical.includes(:user, :event).accept { |c| c.shipping_has_tracking? }.reject { |c| c.stripe_obj[:shipping][:status] == "delivered" } }
```
I'm putting the `accept` first because the `reject` makes an API call to Stripe for every object (`stripe_obj` calls `::Partners::Stripe::Issuing::Cards::Show`). This lets us prevent unnecessary API calls

garyhtou (changes requested):

app/models/stripe_card.rb
garyhtou: Oh actually, scrape this.

> Hmm I think a another way to fix this might be to only show the shipping info for the card if the card eta < Time.now. Not sure which would be better.

I like this better

garyhtou:
> Hmm I think a another way to fix this might be to only show the shipping info for the card if the card eta < Time.now. Not sure which would be better.

Let's go with this. But do something like `eta > 1.weeks.ago`. This way, we show that card for a week after the ETA, in case shipping is slow.

Can you update the count so that it's more accurate?
https://github.com/hackclub/bank/blob/master/app/views/stripe_cards/shipping.html.erb#L4-L4
Instead of getting the count of `@stripe_cards`, we need to change the way we calculate this.

- If the card has tracking, and isn't delivered yet. Then include it in the count
- If the card doesn't have tracking, then include it in the count unless we're past the ETA

This calculation can be performed in the controller, then the value can be assigned to a new instance variable such as `@stripe_cards_arriving_count`

exu3:
https://github.com/hackclub/bank/pull/2372/commits/378a1206f8939b0f549944ef35825a85c075e6a1 attempts to address the comment above. However, I haven't tested it because Development Bank doesn't seem to have the full Card object. Is there a way to test it? I'm not very confident that my code works :p

app/controllers/stripe_cards_controller.rb
```diff
@@ -13,7 +13,9 @@ def shipping
       authorize @event
       @stripe_cards = @event.stripe_cards.physical_shipping
     else # my cards page
-      @stripe_cards = current_user.stripe_cards.physical_shipping
+      eta = @card.stripe_obj[:shipping][:eta]
+      # Only show shipping for phyiscal cards if the eta is in the future (or 1 week after)
+      @stripe_cards = current_user.stripe_cards.physical_shipping.where("eta > 1.week.ago.to_i")
```
garyhtou: I think we might be missing a couple of things here

We would want to get the eta for every card. try something like this:
```ruby
@stripe_cards = current_user.stripe_cards.physical_shipping.reject { |sc| Time.at(sc.stripe_obj[:shipping][:eta]) < 1.week.ago }
```

exu3:
I think this is ready for another review whenever you get a chance.  Ty!

garyhtou (approved):

PR #2374, opened 2022-03-14
Fix transfers

exu3:
It seems like I broke disbursements in #2373 

Cases to test:

- [x] Making a disbursement -- sorta, the new_event_disbursement_path is still broken
- [x] Making a disbursement of an amount higher than the source event balance
- [x] Making an ACH
- [x] ACH of amount higher than source event balance
- [ ] Making a check - errrrr I think this is broken, not sure why -- seems not to be related to this PR


maxwofford:
Thanks for pushing this! Testing now

app/services/disbursement_service/create.rb
```diff
@@ -19,7 +19,7 @@ def initialize(source_event_id:, destination_event_id:,
     def run
       raise ArgumentError, "amount is required" unless @amount
       raise ArgumentError, "amount_cents must be greater than 0" unless amount_cents > 0
-      raise ArgumentError, "You don't have enough money to make this disbursement." unless ample_balance?(amount_cents, @source_event)
+      raise ArgumentError, "You don't have enough money to make this disbursement." unless ample_balance?(amount_cents, source_event)
```
maxwofford:
```suggestion
      raise ArgumentError, "You don't have enough money to make this disbursement." unless ample_balance?(amount_cents, @source_event)
```

app/controllers/disbursements_controller.rb
```diff
@@ -60,7 +60,7 @@ def create
 
   rescue ArgumentError, ActiveRecord::RecordInvalid => e
     flash[:error] = e.message
-    redirect_to new_event_disbursement_path(event_id: @source_event&.slug)
+    redirect_to new_disbursement_path(event_id: @source_event.id)
```
maxwofford:
```suggestion
    redirect_to new_event_disbursement_path(event_id: disbursement_params[:source_event_id])
```

I think this is the intent of this change?

exu3:
These changes were approved by Max over Slack

PR #2379, opened 2022-03-17
More accurate event count in stats endpoint

exu3:
This PR:
```
irb(main):007:0> Event.not_omitted.not_hidden.approved.size
=> 527
```

On the other hand, <https://bank.hackclub.com/stats> currently returns 886 for `events_count`. I think it'd be nice to use the `events_count` for the number on hackclub.com/bank rather than having to manually update every time the number increases by 50.


app/controllers/static_pages_controller.rb
```diff
@@ -110,7 +110,7 @@ def stats
 
     render json: {
       date: now,
-      events_count: Event.not_omitted.where("created_at <= ?", now).size,
+      events_count: Event.not_omitted.not_hidden.approved.size,
```
maxwofford:
```suggestion
      events_count: Event.not_omitted.not_hidden.approved.where("created_at <= ?", now).size,
```

The stats page takes a date argument, so let's include it

exu3: Good call, thanks!

maxwofford (approved):
Thanks for catching this! Go for merge– I can't wait to see the hackclub.com/bank page have this live update!

PR #2388, opened 2022-03-24
Don't include archived & unpaid invoices in stats

exu3:
Fixes #2386

Screen Shot 2022-03-24 at 12 37 51


exu3:
I moved the $ calculations to the Invoices controller as discussed yesterday.

app/views/invoices/index.html.erb
```diff
@@ -41,7 +41,7 @@
 <% end %>
 
 <div class="statset mt3 mb2">
-  <div class="stat">
+  <div class="stat tooltipped tooltipped--s" aria-label="This does not include archived invoices">
```
garyhtou: Nitpick suggestion: Would "Excludes archived invoices" work here? it's just a tad bit shorter

garyhtou (approved):
Code looks good!

A thought about the UX:

Should we include a little note during the archival process so that the users are aware that it will hide the invoice from the stats. and also explain that paid archived invoices will still show in stats

https://github.com/hackclub/bank/blob/98f71b9ff6aaa599fde66a0dd4cd74435f48b669/app/views/hcb_codes/_invoice.html.erb#L15-L15

app/views/invoices/index.html.erb
garyhtou: Also, right margin might help here

Screen Shot 2022-03-28 at 11 39 58 AM

exu3: Screen Shot 2022-03-28 at 20 19 59

Good point, ty!

PR #2392, opened 2022-03-27
Clarify that the donation form is in USD

sampoder:
It would be cool if I could change this as a user, to avoid situations such as https://bank.hackclub.com/donations/start/the-innovation-circuit


exu3 (commented):
Good call, thanks for the PR!

Thoughts on adding this clarification just on organizations that are not US-based? I think the currency can just be assumed for US-based organizations.

You can use the `<% if !@event.country_us? %>` condition (example on line 33).

https://github.com/hackclub/bank/blob/d6f6448ee838594025ca3f82a4155ec970482ff3/app/views/donations/_donation_form.html.erb#L33

sampoder:
done!

exu3 (approved):
Thanks Sam!

garyhtou:
Thanks for the PR @sampoder, and thanks for reviewing @exu3!

PR #2395, opened 2022-03-29
Users can rename transactions from attach_receipt page

exu3:
This PR adds the pending transaction history partial to the hcb code attach_receipt page. The edit button is disabled when the organizer is not signed in.


app/views/hcb_codes/attach_receipt.html.erb
```diff
@@ -3,9 +3,10 @@
...
 <% if @hcb_code.event %>
   <%= link_to @hcb_code.event do %>
-    <h1 class="center border-none h1 mt0 mb0 pt3 pb1 underline primary"><%= @hcb_code.event.name %></h1>
+    <h1 class="center border-none h1 mt0 mb0 pt3 pb1 link-reset primary"><%= @hcb_code.event.name %></h1>
   <% end %>
 <% end %>
 
 <%= render partial: "hcb_codes/rich_metadata" %>
-<%= render "receipts/form_v2" %>
+<%= render partial: "hcb_codes/pending_transaction_history" if organizer_signed_in? %>
```
maxwofford:
```suggestion
<%= render partial: "hcb_codes/pending_transaction_history" %>
```

Do we want this to only show for signed in organizers?

exu3: The edit page only works for signed in organizers, so I thought this partial which contains the edit button should only be visible if that condition is met.

exu3: As mentioned in #2400—let's show this partial, but have the button disabled with a tooltip prompting the user to sign in to rename the tx.

exu3:
I think this is ready for a review. The edit button should be disabled when organizer is not signed in.

garyhtou (approved):
looks good for now.

in the future, as a part of https://github.com/hackclub/bank/issues/1855, we should remove this pending transaction history partial and just have a single button/interface on this page for renaming the transaction. if the user wants to see the fully transaction history, they can visit the HCB show page

PR #2396, opened 2022-03-29
Fix The Overflow Bug

sampoder:
After: Screenshot 2022-03-29 at 10 50 00 PM
Before: Screenshot 2022-03-29 at 10 49 48 PM


maxwofford:
Because the header component is used in so many part of the app, let's check that this works on other pages too– both on mobile and on desktop

cjdenio:
There are a few weird side-effects to using `overflow: hidden`:

Screen Shot 2022-03-29 at 2 19 19 PM

There's probably a better way to solve this, but I'm not sure how yet

PR #2398, opened 2022-03-29
Admin international events filter

exu3:


exu3:
Thoughts on checkbox vs select? The form select would probably fit in better with the other filters, but it'd make more sense to make it a checkbox if it's a true/false filter.

https://github.com/hackclub/bank/pull/2398/commits/850784d4e5372a67d7c6c5b3937968ec8aa20b81 is the commit that should be reverted if a form select is desired.

Checkbox:
Screen Shot 2022-03-29 at 11 11 38

garyhtou (approved):
good work!

> Thoughts on checkbox vs select?

No preference by me. however, it would be cool if "International" was a option in the Country select dropdown

exu3:
> it would be cool if "International" was a option in the Country select dropdown

I like this; how would this work? Would "international" have to go into the country enum or would it be possible to just add an option into the select field?

garyhtou:
> how would this work? Would "international" have to go into the country enum or would it be possible to just add an option into the select field?

I think something like this might work for adding it to the select options:
```ruby
# I'm not too sure how rails enums work under the hood, but if their similar to hashes then this should work
{International: 'international'}.merge(Event.countries)
```
then in the controller, handle the filtering differently if `@country == 'international'`

garyhtou:
let's try not to add "international" in the country enum since it's not a country

hackliv:
Would it be possible to change "International" to "Outside of the U.S." Just some phrasing I think Zach prefers. Nice job on this!

PR #2401, opened 2022-03-30
Filter ledger by User for Stripe Card Transactions

garyhtou:
https://hackclub.slack.com/archives/C026RKHLPNJ/p1648586258404309


app/controllers/admin_controller.rb
```diff
@@ -401,6 +402,26 @@ def ledger
...
+
+        join "canonical_hashed_mappings" chm on ct.id = chm.canonical_transaction_id
+        join "hashed_transactions" ht on chm.hashed_transaction_id = ht.id
+        join "raw_stripe_transactions" rst on ht.raw_stripe_transaction_id = rst.id
+
+        where rst.stripe_transaction->>'cardholder' = '#{sch_sid}'
+        SQL
+
+        ids = ActiveRecord::Base.connection.execute(query).values.flatten
+        relation = relation.where(id: ids)
```
garyhtou: hey @albertchae if you get a chance, can you see if it's possible to convert this into a Rails AR query?

I initially had a simple `select` (https://github.com/hackclub/bank/pull/2401/commits/9220ae0d746bceff9e19295274f9435c9b5eeb02), but there were two issues:
1. wayy too slow
2. returns an array, not an active collection

I tried converting the `select` into a `where` query, but I got stuck messing around with `includes` and `references`

garyhtou: example of expected results:

http://localhost:3000/admin/ledger?q=&unmapped=0&exclude_top_ups=0&mapped_by_human=0&event_id=&user_id=1&commit=Search

image

albertchae: Hey @garyhtou 

I think this line worked for me
```ruby
relation = relation.joins(hashed_transactions: :raw_stripe_transaction).where("raw_stripe_transactions.stripe_transaction->>'cardholder' = ?", sch_sid)
```

alternatively if you want to keep it a separate query for fetching the ids, you can do 
```ruby
ids = CanonicalTransaction.joins(hashed_transactions: :raw_stripe_transaction).where("raw_stripe_transactions.stripe_transaction->>'cardholder' = ?", sch_sid).pluck(:id)
relation = relation.where(id: ids)
```
There's probably a way to join it directly to the user's `stripe_id` too without having to do the `find` and `user&.stripe_cardholder&.stripe_id`.


BTW, I think this part of the join wasn't doing anything right? since you're not using it in the where or in subsequent joins
`join "canonical_event_mappings" cem on ct.id = cem.canonical_transaction_id `

garyhtou: that's exactly what i'm looking for! thanks!

> BTW, I think this part of the join wasn't doing anything right?

yeah, that's right. i was using it to filter out `CanonicalTransactions` that don't have an event mapping. but it's not necessary since that was handled when the `relation` variable was initialized

---

is there good way to know when to use `joins` over `includes`? I've read that `includes` is generally more performant than `joins` since it can eagar load. but I'm still not too sure when `join` would be better

albertchae: I think `joins` and `includes` are for separate (but related) problems. `includes` is for preloading of association data so that if you have to loop over a list and look at an association it won't do an N+1 and `joins` makes associated table columns available for filtering in the where. You can use both depending on what you're trying to do.

I get confused by this a lot and usually refer to
https://www.bigbinary.com/blog/preload-vs-eager-load-vs-joins-vs-includes
but I also just came across this post by scout which is pretty good
https://scoutapm.com/blog/activerecord-includes-vs-joins-vs-preload-vs-eager_load-when-and-where

PR #2406, opened 2022-03-30
International events filter take 2: Electric Boogaloo

exu3:
Follow up of #2398 

Changes "international" to "outside US" - https://github.com/hackclub/bank/pull/2398#issuecomment-1083194448 

Adds it to the form select - https://github.com/hackclub/bank/pull/2398#issuecomment-1082927376

Screen Shot 2022-03-30 at 12 05 52


exu3:
@hackliv Thoughts on "non US" vs "outside US"?

hackliv:
I like Outside U.S.

exu3:
It currently doesn't work, but I think it might have to do with: 
https://github.com/hackclub/bank/blob/d5a3965c8be8154ac1f9086246be5cc7306f6dea/app/controllers/admin_controller.rb#L231

I'm not quite sure how to fix 😅

garyhtou:
I have no idea if this is the case, but what is `@country` is a string and it's doing `string == integer`

garyhtou:
To see if `@country` is a string, you can have it print or display `@country.class`

or do `@country == 999.to_s` and see if that works

app/controllers/admin_controller.rb
```diff
@@ -1075,7 +1075,13 @@ def filtered_events(events: Event.all)
...
     @organized_by_hack_clubbers = params[:organized_by_hack_clubbers].present? ? params[:organized_by_hack_clubbers] : "both" # both by default
     @international = params[:international] == "1" ? true : nil # unchecked by default
-    @country = params[:country].present? ? params[:country] : "all" # all by default
+    if params[:country] == 9999.to_s
+      @country = 9999
+    elsif params[:country].present?
+      @country = params[:country]
+    else
+      @country = "all"
+    end
```
garyhtou: if you'd like to clean this up a little bit, I think we can do:
```ruby
@country = params[:country].present? ? params[:country] : "all"
```
then below on line `1099`, do:
```ruby
if @country == 9999.to_s

# or the following would work too:
if @country == "9999"
```

---

It's totally possible to have a variable store different types (Integer, String, etc.) since ruby is dynamically typed (similar to JavaScript & Python, and unlike C++ & Java). But, we generally try to avoid it because sometimes it can confusing to tell what type it is (someone might accidentally call a string method like`gsub` on an integer).

garyhtou (approved):
Yeahhhh! Good work here! good to merge

hackliv:
yay ya yay!

PR #2409, opened 2022-03-31
Deletion Request Page Edits

kunalbotla:
This PR edits the copy of the deletion request page and removes the email checkpoint for deletion requests. 

The copy is edited to fix a typo and to clarify that the form doesn't remove organizers from the contract and that they'll need to do that by reaching out to bank@hackclub.com. 

The email checkpoint is removed since removing an organizer doesn't impact their Google Workspace account or any they've created. Having that checkpoint there creates confusion about what a deletion request does to Google Workspace.


kunalbotla:
I didn't want to rewrite the text, but I think the second paragraph could use some revision

MelSmith104 (commented):
This looks good to me - please run by @garyhtou  but should be good to merge

garyhtou (commented):
Should we also remove the admin side of things? (all new deletion requests will have it defaulted to `false`)

https://github.com/hackclub/bank/blob/772a97f21d3b3b5de117f22f2a4b17bbea416c47/app/views/organizer_position_deletion_requests/show.html.erb#L50-L54

kunalbotla:
I think so; I'm removing it. Does everything else look good?

maxwofford:
Not sure what the current ops process is for deletion requests, but the previous way we handled them coming in was deleting their account on bank and removing their gsuite account (by default). If I'm an organizer, how do I deal with the following situations:

- a team member ages out. we're on good terms, but they shouldn't have direct access anymore
- my sponsorship director goes rogue and starts taking org documents to start a new hackathon– I need to get rid of their access to everything ASAP, but they have active comms with our donors
- my technical director goes rogue– I need to get rid of their access to everything ASAP
- I accidentally added the wrong email for a user and want to hide that user from my account (i've already sent an invite to the right email, so the wrong one is redundant)

garyhtou (approved):
Code portion looks good!

kunalbotla:
@maxwofford our current one is similar. We still have instant suspension of Google Workspace accounts for most of those use cases. The only function we don't have is instant account or org suspension but I think that functionality should come in a different PR.

maxwofford (approved):
Cool with the code changes here if ops team is fine with the form changes

kunalbotla:
@MelSmith104 approved in the comment of https://github.com/hackclub/bank/pull/2409#pullrequestreview-953927749 so I'm merging!

PR #2411, opened 2022-03-31
Add ability to select multiple receipts during upload

cjdenio:
Closes #2410

https://user-images.githubusercontent.com/34525547/160972709-edb59bc7-3bee-45b6-87b2-2019c1d6dd87.mov


exu3 (approved):
image

Ohyeahwoofyeah! Looks good to me. Thanks for adding this!

app/services/hcb_code_service/receipt/create.rb
```diff
@@ -12,7 +12,6 @@ def initialize(hcb_code_id:,
       end
 
       def run
-        hcb_code.update_column(:marked_no_or_lost_receipt_at, nil)
```
garyhtou: Just wondering what your thoughts were for moving this out of the service

cjdenio: Yeah, so otherwise it gets run unnecessarily for _each receipt you upload_, when it only needs to be run once.

garyhtou: Yeah that's true. However, it might be better to keep it within the service so that it's self-contained (encapsulation). This way, anytime we call this service, we don't need to worry about setting `marked_no_or_lost_receipt_at`.

It seems like we aren't using this service anywhere else. but it's still good to keep relevant code contained/together

cjdenio: Good point

garyhtou (approved):
Good to go! Thanks @cjdenio!

PR #2413, opened 2022-03-31
Rename "Inbox" nav item to "Receipts"

hackliv:
Screen Shot 2022-03-31 at 1 26 58 PM

Rename "Inbox" navigation item to "Receipts". 
This change would allow there to be clear and concise labelling on the Bank dashboard where users understand where they can see their transactions missing receipts. There is currently some confusion between users and myself behind the connotation of "Inbox," (meaning messages received via email). 

FYI: This only changes the wording on the page, the URL still uses the word inbox. We may want to change that.

YAYYYYY this is my first pull request. :)


exu3 (changes requested):
Nice! Congrats on your first PR 

There are a couple other places where it says "Inbox" which should probably be changed to maintain consistency.

Tooltip on the icon in the header: https://github.com/hackclub/bank/blob/af85e616291c1b9d8ccf69f1c4b4a58be426eaa2/app/views/application/_header.html.erb#L40

Title of the page itself: https://github.com/hackclub/bank/blob/e04d68422727dc3982c2647370804da8f13e1cb6/app/views/static_pages/my_inbox.html.erb#L1

---

For the URL change, that would be [here](https://github.com/hackclub/bank/blob/451c2c92e9313c9d14aa9922a69be393b6ed0067/config/routes.rb#L36), & it'd probably be a good idea to redirect `/my/inbox` to `/my/receipts` to retain backwards compatibility in case anyone has the link saved. This can be done in a separate PR though so it's not a blocker here.

maxwofford (approved):
This is looking great, thanks for the review @exu3. Go ahead and merge it for @hackliv once you're happy with how it's looking.

Re: URL route changes– let's leave the current routes & not create new ones. Keeps stuff simple.

exu3 (approved):
Nice work here, looks good to merge! 

Screen Shot 2022-03-31 at 13 41 28

PR #2414, opened 2022-03-31
Add notice about codespace setup being recommended setup flow

maxwofford:
This came up while setting up @hackliv's dev environment today! It's hard to install ruby/bundler/gem/pg/etc... let's just use docker!


README.md
```diff
@@ -14,6 +14,8 @@ Instructions on how to setup can be found [here](/codespace-steps.md). Premade `
 
 ## Getting Started Locally
 
+_⚠️ If you already have Docker installed locally we recommend using the Codespaces setup instructions above to get HCB running locally. It should work out-of-the-box and is how most contributors work on HCB._
```
exu3: I don't think you need Docker installed locally to run a Codespace; you could also run it in your browser.

maxwofford: what i'd like to communicate is that I use the codespace config scripts on my local macbook rather than on codespaces & it works great.

exu3: Ah gotcha, I missed that part—that makes sense!

There are also instructions on running it locally with Docker https://github.com/hackclub/bank#alternative-with-docker below this section.

exu3: Thoughts on renaming the Codespace config scripts to something like `setup.sh` so it makes it more clear that it can be used for general purpose setup and not just Codespaces?

exu3:
Closing this PR as this change was addressed in #2430 and the new wiki documentation page on setting up a dev environment. https://github.com/hackclub/bank/wiki/Development

Feel free to add to the wiki if anything should be further clarified!

PR #2418, opened 2022-04-01
User Generated Custom CSS

garyhtou:
heck yeah! users can now spice up their event pages with their own custom CSS by proving a link to a CSS file (similar to Scrapbook).

We plan on having a default subtle CSS link applied to all events for april fools. That can be found in `public/dinobbq.css`

https://hackclub.slack.com/archives/GBXBFEED9/p1648763273801899

---

#### On Deployment, run:
```ruby
Event.update_all custom_css_url: "https://bank.hackclub.com/dinobbq.css"
```


app/views/events/_form.html.erb
```diff
@@ -6,6 +6,20 @@
   </div>
 
   <% if !@event.unapproved? %>
+
+  <h3 class="mb1">Custom CSS</h3>
+
+    <div class="card">
+    <p class="mt0 pl0 mb2">To customize the CSS on your dashboard, paste a link below to a CSS file or a <a href="https://gist.github.com/" target="_blank">GitHub Gist</a>.</p>
```
garyhtou: I think it needs to be an actual .CSS fille (it can be hosted on GitHub Gist tho)

Example:
- https://gist.github.com/exu3/8476d05e625537e7136fed1119e70667 won't work
- https://gist.githubusercontent.com/exu3/8476d05e625537e7136fed1119e70667/raw/6268cf0c6d5c2676ae1c620dffa05b10b3df292b/bank.css would work

if we wanted to, we could update that scrapbook endpoint so that it handles gists pages (and gets the raw CSS from the first file)

PR #2419, opened 2022-04-01
Adjust styles on CSS URL card

exu3:
Right now it's **all bold** (it feels weird, sort of like screaming text)


exu3:
Screen Shot 2022-04-01 at 03 25 05

PR #2422, opened 2022-04-02
Don't show custom CSS after April fools (unless ?jank)

exu3:


app/views/events/_nav.html.erb
```diff
@@ -63,8 +63,8 @@
 <% content_for(:container) do %>
 <% end %>
 
-
-<% if params[:jank].present? || DateTime.now >= Time.new(2022, 4, 1, 9, 30, 00, "-04:00") %>
+<!-- 2022 April Fools :p -->
```
cjdenio: super tiny thing: if you make this an ERB comment (`<%# 2022 April Fools :p %>`) it won't show up in the HTML output

otherwise this happens:
Screen Shot 2022-04-02 at 12 12 32 PM

exu3: Ah I see, thanks!

exu3:
I'm not sure if this is the way we want to disable this feature but I'm going to merge this for now in case people start thinking that their dashboard is messed up because of custom CSS.

PR #2423, opened 2022-04-03
Disable/tooltip CPT edit button if organizer not signed in

exu3:
Fixes https://hackclub.slack.com/archives/C02AN8HRS0H/p1648942615143059


garyhtou (changes requested):
Ah, looks like this one's a tricky bug.

So, preferably, we'd use `organizer_signed_in?` because:

|`organizer_signed_in?` | `current_user` or `signed_in?` |
|--|--|
| - There is a user signed in.<br/>- AND the user has access to the current event. | - There is a user signed in.<br/>(it doesn't care whether the user has access to the current event) |

The tricky thing here is that `organizer_signed_in?` depends on an `@event` instance variable being set. This usually happens in the controller.

https://github.com/hackclub/bank/blob/master/app/helpers/sessions_helper.rb#L46-L49

If we take a look at the `attach_receipt` action in the `HcbCodeController`, we can see that it's not setting the `@event` instance variable.

https://github.com/hackclub/bank/blob/master/app/controllers/hcb_codes_controller.rb#L54-L59

Try adding
```ruby
@event = @hcb_code.event
```
to that `attach_receipt` action

exu3:
Thanks for that explanation. That seems to have fixed it—https://bank--pr-2423.herokuapp.com/hcb/QmHlGm/attach_receipt tested again.

app/views/hcb_codes/_pending_transaction_history.html.erb
```diff
@@ -22,7 +22,7 @@
                   edit_canonical_pending_transaction_path(ct),
                   class: "right tooltipped tooltipped--w", 
                   "aria-label": "Rename pending transaction",
```
garyhtou: untested:
```suggestion 
                  "aria-label": organizer_signed_in? ? "Rename pending transaction" : "Sign in to rename" %>
```

garyhtou (approved):
Looks good to merge as is!

If you'd like to add a tooltip for when it's disabled, then take a look at the suggestion

app/views/hcb_codes/_pending_transaction_history.html.erb
exu3: I don't think it works when the tooltip is applied to the button & the button is disabled. It does however work on the parent element, so I've applied it there.
Screen Shot 2022-04-03 at 13 10 41
Screen Shot 2022-04-03 at 13 10 28

garyhtou: ahhh okay, interesting. i guess it makes sense tho

thanks!

PR #2430, opened 2022-04-03
Updates to Documentation & README

exu3:
> Edited: 
> All documentation previously in the README has been moved to a [wiki](https://github.com/hackclub/bank/wiki). This PR rewrites the README to be _a little less technical_. Additionally, this PR renames the setup/start scripts accordingly for general purpose setup with Docker (instead of just Codespaces). 

Changes may be opinionated, so if anyone has any thoughts around this, would love to hear them.

---

Original content of this comment:

Opinionated edits to the documentation around development setup instructions. Related to #2414

Most notable change is renaming setup scripts to be for general purpose setup with Docker (instead of just Codespaces).

<details>

<summary> Notes about local development currently unrelated to this PR </summary>

- Setting up with dotenv is currently blocking because the project has reached the max # of collaborators (3).

- Nokogiri is throwing a bunch of errors when I try installing HCB locally (M1) with Docker. Not sure if this is an isolated case or if others are experiencing the same.
  - Everything is still fine in Codespaces.

</details>


garyhtou:
> Setting up with dotenv is currently blocking because the project has reached the max # of collaborators (3).

dotenv isn't necessary (it hasn't been tightly integrated into Bank and falls back to .env files). it is fine to proceed without setting it up. Just use normal `.env*` files

> Nokogiri is throwing a bunch of errors when I try installing HCB locally (M1) with Docker. Not sure if this is an isolated case or if others are experiencing the same.

yep, it's due to https://github.com/hackclub/bank/pull/2334 (https://github.com/sparklemotion/nokogiri/pull/2332). I've been playing around with various installation configs. the root cause is that the `ruby:2.7.3` base image seems has a `glibc` version of `< 2.29`

garyhtou:
thoughts on splitting the readme into multiple files with links?

some information like the plaid section I added in https://github.com/hackclub/bank/pull/2424 really doesn't need to be there cluttering up the main readme file

I'm thinking we clean up [this folder](https://github.com/hackclub/bank/blob/0c74273f3b4b1b39f31510f32107236882e0e9e1/documentation#L1-L1) and split the files in there

exu3:
> Throught on splitting the readme into multiple files with links?

Good idea! Maybe a wiki would help? I think [instructure/canvas-lms](https://github.com/instructure/canvas-lms/wiki) does it quite well.

exu3:
> dotenv isn't necessary (it hasn't been tightly integrated into Bank and falls back to .env files). it is fine to proceed without setting it up. Just use normal .env* files

Ah gotcha, this is something we probably should clarify in the README.

garyhtou:
> Good idea! Maybe a wiki would help?

yeah I was thinking that too. i find it kinda weird how it's technically a separate repo. However, it think the trade-offs are worth it

I also wonder how easy it is to search

re separate repo: https://docs.github.com/en/communities/documenting-your-project-with-wikis/adding-or-editing-wiki-pages#adding-or-editing-wiki-pages-locally

exu3:
> I also wonder how easy it is to search

With descriptive-enough headings & page titles, it seems like it should be fairly easy.

Screen Shot 2022-04-02 at 21 39 56

A wiki would probably be substantially easier to view than md files hosted on the main repo/branch.

garyhtou:
> A wiki would probably be substantially easier to view than md files hosted on the main repo/branch.

good point. i think let's go with wiki then!

garyhtou:
fyi:  image

https://docs.github.com/en/codespaces/prebuilding-your-codespaces
https://github.com/hackclub/bank/settings/codespaces

could be helpful

garyhtou:
> Nokogiri is throwing a bunch of errors when I try installing HCB locally (M1) with Docker

Oh yeah, the work around right now is to build native gems from scratch when installing. To do this add `RUN bundle config force_ruby_platform true` above `RUN bundle install` in the `dockerfile`.

It will be really really slow. Took about 20 minutes just to `bundle install` on my mac (M1 pro, 32 gb)

I haven't found a way to only build the `Nokogiri` gem from scratch (and leave the rest as pre-compiled).

---

the long term solution might be to upgrade our ruby image/version to that supports `glibc >= 2.29`

exu3:
> fyi: image
> 
> [docs.github.com/en/codespaces/prebuilding-your-codespaces](https://docs.github.com/en/codespaces/prebuilding-your-codespaces) [hackclub/bank/settings/codespaces](https://github.com/hackclub/bank/settings/codespaces)
> 
> could be helpful

Oh interesting, also just saw their [blog post](https://github.blog/2022-02-23-codespaces-largest-repositories-faster/) on that.

I enabled it and tried it out & it seemed to make it a little faster. Bank is not _super huge_ (I think), but no harm in turning that feature on!

exu3:
I also moved the technical documentation in the README and Codespace setup documentation into the wiki.

For the https://github.com/hackclub/bank/blob/0c74273f3b4b1b39f31510f32107236882e0e9e1/documentation folder (which contains documentation around Emburse configuration on the admin side), is that cool to delete? Or add a deprecation note?

garyhtou:
> For the [`0c74273`/documentation](https://github.com/hackclub/bank/blob/0c74273f3b4b1b39f31510f32107236882e0e9e1/documentation) folder (which contains documentation around Emburse configuration on the admin side), is that cool to delete? Or add a deprecation note?

yeah, it looks like everything's outdated. should be fine to delete the entire folder

exu3:
Cool, I think this is ready for review, & I've added a TLDR in the top-level comment.

docker_dev_setup.sh
```diff
@@ -45,12 +45,12 @@ fi
 echo "$(tput setaf 9)Hack Club Bank:$(tput sgr0) $(tput setaf 10)Done$(tput sgr0)"
 
 echo "
-$(tput setaf 9)Hack Club Bank:$(tput sgr0) Step 5/7: Copy Dockerfile to Docker Container"
+$(tput setaf 9)Hack Club Bank:$(tput sgr0) Step 5/7: Copy .env file"
 cp .env.docker.example .env.docker
```
garyhtou: somewhat unrelated to this PR, but we should do `cp -n`. otherwise, it might override someone's `.env.docker` file if they've made changes to it

`cp -n` copies if it doesn't already exist

garyhtou (approved):
looks good!

Thoughts on making the script names shorter?
e.g. `docker_dev_start` -> `docker_start`

maxwofford (approved):
Love it! Thanks for updating all of this. Only change is to mention that we use GH wiki in the README so people know to check up there for documentation!

PR #2433, opened 2022-04-03
Organization balance report

garyhtou:
https://bank--pr-2433.herokuapp.com/admin/balances

image

image

also added `JSON` export cause it was easy


garyhtou:
fyi @exu3, this PR will most likely cause merge conflicts with https://github.com/hackclub/bank/pull/2406

maxwofford:
This looks great & addresses the issue! Go ahead and merge this once you're ready!

app/controllers/admin_controller.rb
```diff
@@ -1012,9 +1012,8 @@ def balances
       [:end_date, ->(_) { @end_date }]
     ]
     serializer = ->(event) do
-      template.reduce({}) do |hash, (header, field)|
+      template.each_with_object({}) do |(header, field), hash|
```
garyhtou: dang, cool method

garyhtou: makes it so that you don't need to return the hash on every loop

PR #2436, opened 2022-04-03
Handle nil users for user_mention and avatar_for

exu3:
Closes #2427 
Screen Shot 2022-04-03 at 08 07 51


exu3:
This should be ready for review. Is there a more appropriate fallback image to use? Currently, it's: 

image

garyhtou (approved):
look good! consider adding option to change "Anonymous" to "no user"

PR #2438, opened 2022-04-04
`/:event/edit` -> `/:event/settings`

cjdenio:
li'l opinionated change to make the URL line up with the nav item's label.

Screen Shot 2022-04-04 at 3 59 37 PM


garyhtou (approved):
good idea!

PR #2448, opened 2022-04-08
Add period to `help_message`

kunalbotla:
The help message seems like a typo when in a sentence. Adding a trailing period fixes that.

image


exu3 (commented):
Good call! Could you also remove instances of `help_message` with a trailing period so that there aren't double periods anywhere?

One such example is https://github.com/hackclub/bank/blob/51db5f1121bff565100c7e2059cb95c3793debd3/app/views/events/transfers.html.erb#L33

exu3 (approved):
LGTM!

PR #2450, opened 2022-04-10
Preload CanonicalTransaction and attach to CanonicalTransactionGrouped in events#show

albertchae:
This is the first of several PRs to address https://github.com/hackclub/bank/issues/2446

Query/allocation count before: SQL Queries: 1520 (520 cached) | Allocations: 4797691
Query/allocation count after: SQL Queries: 1421 (520 cached) | Allocations: 4333447

I noticed this query in the log
```
CanonicalTransaction Load (0.6ms)  SELECT "canonical_transactions".* FROM "canonical_transactions" WHERE "canonical_transactions"."id" = $1 ORDER BY date desc, id desc LIMIT $2  [["id", 208083], ["LIMIT", 1]]
  ↳ app/models/canonical_transaction_grouped.rb:158:in `ct'
```

Which eventually calls
https://github.com/hackclub/bank/blob/896783570d8331066c66d7e54a8c3580c7c43008/app/models/canonical_transaction_grouped.rb#L31-L33
this is an N+1 because we'll query `CanonicalTransaction` again for each `CanonicalTransactionGrouped` on the page

To mitigate this, we can load all the associated `CanonicalTransaction`s
in one query and then attach them to the appropriate
`CanonicalTransactionGrouped` instances. Then the memoized instance
variable will already be set so we won't need to issue a new SQL query.


app/models/canonical_transaction_grouped.rb
```diff
@@ -84,6 +85,10 @@ def local_hcb_code
     @local_hcb_code ||= HcbCode.find_or_create_by(hcb_code: hcb_code)
   end
 
+  def canonical_transaction_ids
+    JSON.parse(raw_canonical_transaction_ids)
+  end
+
```
albertchae: Had to make this method public, which didn't seem like a big deal to me since `raw_canonical_transaction_ids` is already public

app/models/canonical_transaction_grouped.rb
```diff
@@ -4,6 +4,7 @@ class CanonicalTransactionGrouped
   include ActiveModel::Model
 
   attr_accessor :hcb_code, :date, :amount_cents, :raw_canonical_transaction_ids
+  attr_writer :canonical_transactions
```
albertchae: I don't love having this writer be public and would prefer to only set this on the constructor or something, but this is the easiest change for now. I want to revisit later whether we can keep setting `@canonical_transactions` internal to this class

app/controllers/events_controller.rb
```diff
@@ -29,6 +29,18 @@ def show
...
+    canonical_transaction_ids = @transactions.flat_map(&:canonical_transaction_ids)
+    canonical_transactions_by_id = CanonicalTransaction.where(id: canonical_transaction_ids).index_by(&:id)
+
+    @transactions.each do |t|
+      t.canonical_transactions = canonical_transactions_by_id.slice(*t.canonical_transaction_ids).
+        values.
+        sort_by do |ct|
+          [ct.date, ct.id]
+        end.
+        reverse
```
albertchae: This sort and reverse is essentially this SQL query https://github.com/hackclub/bank/blob/896783570d8331066c66d7e54a8c3580c7c43008/app/models/canonical_transaction_grouped.rb#L32 but in Ruby

Unfortunately sort and reverse means we are doing 2 passes over the array and I can't trivially add a `-` sign in front to do a single pass sort in reverse order because I can't "negate" a date (e.g. `[-ct.date, -ct.id]`). A proposal to add a reverse sort to the standard library was shot down a couple years ago https://bugs.ruby-lang.org/issues/15725. There might be a way to do this in a single pass but I'll revisit after reducing the queries first since it might not be a big effect given that the page size is only 100 elements.

app/controllers/events_controller.rb
```diff
@@ -29,6 +29,18 @@ def show
...
+    end
   end
```
albertchae: I don't think this preloading logic belongs in the controller and will move it out to a service class in the future (and maybe call it from https://github.com/hackclub/bank/blob/896783570d8331066c66d7e54a8c3580c7c43008/app/services/transaction_grouping_engine/transaction/all.rb#L15-L19)

albertchae:
@maxwofford I didn't see any tests for this controller so I didn't add/modify any, but let me know if I missed them

app/controllers/events_controller.rb
albertchae: Forgot I could use `sort` instead of `sort_by` to sort in the correct order from the get go. That change is in https://github.com/hackclub/bank/pull/2450/commits/1bd86bd31abd39c07949af4b9b7fd02ae4be4014

maxwofford (approved):
Looks good! I just skimmed over it & double checked everything looked right in the preview deploy. Let's def change that `attr_writter`, but I'm cool with that getting fixed in a future PR. Great work here, feel free to merge once you're ready

albertchae:
@maxwofford sounds good. The `attr_writer` issue might get worse before it gets better 😅 , but I'll keep thinking of a better way to handle that.

One idea I had, if we don't have a good test suite covering this, I've "tested in prod" before with feature flags or https://github.com/github/scientist, or we can even just have a query param `?preload=true" which I can use to conditionally run the preload logic or not. Then we can deploy to prod and do a curl with and without the query param and compare the results (both output and time). The negative is that we won't start seeing incremental improvements to the page speed and only at the end when we finally remove the query param. Any thoughts?

maxwofford:
I love the `?preload=true` idea for its simplicity! Let's do that so we can more easily benchmark the before and after. No need to gate the stuff in this PR, just for optimizations going forward.

PR #2451, opened 2022-04-10
Make name validation work properly during onboarding

cjdenio:
Updates the username validation to only fail if
- The name is blank AND
- The name was not blank previously

(fixes #2449)


maxwofford:
Does this handle the case a user sets their name during onboarding, then tries to blank out their name?

cjdenio:
Yep, it'll fail if the user tries to go from `"Caleb"` -> `nil`

(but `nil` -> `nil` is allowed)

maxwofford (approved):
Awesome! If this works in staging & you can create a new account that goes through the onboarding just fine, go for merge!

cjdenio:
It works!

For posterity, this also still prevents users from removing their own name:

Screen Shot 2022-04-10 at 6 06 33 PM

PR #2452, opened 2022-04-10
Preload HcbCode and attach to CanonicalTransactionGrouped in events#show

albertchae:

Query/allocation count before: SQL Queries: 1421 (520 cached) | Allocations: 4333447
Query/allocation count after: SQL Queries: 1322 (520 cached) | Allocations: 4136073

From this query:
```
  HcbCode Load (0.3ms)  SELECT "hcb_codes".* FROM "hcb_codes" WHERE "hcb_codes"."hcb_code" = $1 LIMIT $2  [["hcb_code", "redacted"], ["LIMIT", 1]]
  ↳ app/models/canonical_transaction_grouped.rb:85:in `local_hcb_code'
```

Which will be an N+1 here

https://github.com/hackclub/bank/blob/896783570d8331066c66d7e54a8c3580c7c43008/app/models/canonical_transaction_grouped.rb#L84

To mitigate this, we can load all the associated `HcbCode`s
in one query and then attach them to the appropriate
`CanonicalTransactionGrouped` instances.

Technically we should also do an `insert_all` (which is available since
Rails 6 https://www.bigbinary.com/blog/bulk-insert-support-in-rails-6)
to create any non existing ones (since the original is `find_or_*create*_by`, but for now falling back on the
individual creates should be fine because it should only happen the
first time the transaction is loaded.


albertchae:
Please review https://github.com/hackclub/bank/pull/2462 before this one

maxwofford (approved):
Solid, LGTM!

PR #2453, opened 2022-04-10
Attach CanonicalTransaction to HcbCode in events#show

albertchae:

Query/allocation count before: SQL Queries: 1322 (520 cached) | Allocations: 4136073
~~Query/allocation count after: SQL Queries: 1222 (520 cached) | Allocations: 3869178~~
After adding query https://github.com/hackclub/bank/pull/2453/commits/5f15b393d170ed5586d8bdc48ea3c6ebd912ab35: SQL Queries: 1223 (520 cached) | Allocations: 3611538

From this query:
```
  CanonicalTransaction Load (2.3ms)  SELECT "canonical_transactions".* FROM "canonical_transactions" WHERE "canonical_transactions"."hcb_code" = $1 ORDER BY date desc, id desc LIMIT $2  [["hcb_code", "redacted"], ["LIMIT", 1]]
  ↳ app/models/hcb_code.rb:215:in `ct'
```

which will be an N+1 here
https://github.com/hackclub/bank/blob/896783570d8331066c66d7e54a8c3580c7c43008/app/models/hcb_code.rb#L70

~~This approach uses the already loaded `CanonicalTransaction`s, but I'm
not sure whether we need to actually query again for them by `hcb_code`
since that might be a different set of `CanonicalTransaction`s...~~


albertchae:
Please review https://github.com/hackclub/bank/pull/2452 before this

app/services/transaction_grouping_engine/transaction/association_preloader.rb
```diff
@@ -5,31 +5,43 @@ module Transaction
...
+                                     .values
+                                     .sort { |ct1, ct2| compare_date_id_descending(ct1, ct2) }
 
           t.local_hcb_code = hcb_code_by_code[t.hcb_code]
         end
       end
 
+      # comparator that can be used in Array#sort for canonical_transactions
+      # https://ruby-doc.org/core-2.7.5/Array.html#method-i-sort
+      def compare_date_id_descending(ct1, ct2)
```
albertchae: I believe I could write this as `@compare_date_id_descending = ->(ct1, ct2)` and then pass `&@compare_date_id_descending` directly to the sort, but in my experience lambdas/procs have significant performance issues (sorry I don't have a definitive source on this and might be wrong) and in this case don't add a lot of readability anyway

app/services/transaction_grouping_engine/transaction/association_preloader.rb
```diff
@@ -5,31 +5,43 @@ module Transaction
     class AssociationPreloader
       def preload_associations!(transactions)
         canonical_transaction_ids = transactions.flat_map(&:canonical_transaction_ids)
-        canonical_transactions_by_id = CanonicalTransaction.where(id: canonical_transaction_ids).index_by(&:id)
+        canonical_transactions = CanonicalTransaction.where(id: canonical_transaction_ids)
+        canonical_transactions_by_id = canonical_transactions.index_by(&:id)
+        canonical_transactions_by_hcb_code = canonical_transactions.group_by(&:hcb_code)
```
albertchae: I don't know enough about the data model to know the answer to this: should I instead move this a few lines below and write it as

`canonical_transactions_by_hcb_code = CanonicalTransaction.where(hcb_code: hcb_code_codes).group_by(&:hcb_code)`

That is, if I go from `CanonicalTransaction` -> `HcbCode` -> `CanonicalTransaction`, can the last set of `CanonicalTransaction`s be greater than the first?

garyhtou: I think both should work (both `CT` sets should be the same). however, hcb codes are lazily created and associated to `CanonicalTransaction`s/`CanonicalPendingTransaction`s (either by calling `HcbCode#local_hcb_code` or by the	`TransactionGroupingEngineJob::Nightly` job that runs every 5 mins). so maybe there could be a world where they're different due to concurrency (if new associations are created in between)

maxwofford: @garyhtou aren't they different in the following example:

- donation with 2 `CanonicalTransaction`s (fee refund & payout), spaced out by a couple days
- enough transactions between the two cause an event page (with `LIMIT 100`) to catch the payout `CanonicalTransaction`, but not the fee refund `CanonicalTransaction`

I believe the "CanonicalTransaction -> HcbCode -> CanonicalTransaction" approach would also pick up the fee refund.

maxwofford: Ah, I think I'm wrong after reading this: it seems like `TransactionGroupingEngine::Transaction::All` called [just above](https://github.com/hackclub/bank/blob/00332a77b5d5534d4efded3fdb302a0210d04c74/app/controllers/events_controller.rb#L31) is already going from `HcbCode` -> `CanonicalTransaction`. https://github.com/hackclub/bank/blob/00332a77b5d5534d4efded3fdb302a0210d04c74/app/services/transaction_grouping_engine/transaction/all.rb#L25

albertchae: Hm very interesting, I think you're right though @maxwofford in https://github.com/hackclub/bank/pull/2453#discussion_r848829935, even if `TransactionGroupingEngine::Transaction::All` does go from `HcbCode` -> `CanonicalTransaction` since Kaminari will truncate it to 100 entries. I'm going to add the query to be safe, it's still only one additional query so shouldn't be a big deal.

Also @garyhtou what you are saying about the lazy creation of `HcbCode`s is pretty interesting. What's the reasoning behind that? If we're running a batch job to create them every 5 minutes it seems like it might be worth creating them as soon as they're needed (i.e. when the `CanonicalTransaction` is created). Also side note that `TransactionGroupingEngineJob::Nightly` seems to be a small N+1 since it's doing creates in a loop instead of one big `INSERT`

albertchae: Implemented the additional query in https://github.com/hackclub/bank/pull/2453/commits/5f15b393d170ed5586d8bdc48ea3c6ebd912ab35

maxwofford (approved):
Seems like https://github.com/hackclub/bank/pull/2453#discussion_r849110008 is resolved, so merge when ready!

app/services/transaction_grouping_engine/transaction/association_preloader.rb
garyhtou: > Also @garyhtou what you are saying about the lazy creation of HcbCodes is pretty interesting. What's the reasoning behind that

tbh I've got no idea. not sure why they were designed that way. we have ran into a couple issues with depending on the 5 min job. creating when the CT is created does seem like a better idea

it looks like we're doing that for CPT (canonical pending transactions)
https://github.com/hackclub/bank/blob/c83f9023e4f1009e4ea30571231b68aaee3d136b/app/models/canonical_pending_transaction.rb#L58

but not for CT

PR #2461, opened 2022-04-12
Fix missing receipt count

cjdenio:
👋 Howdy! The main point of this PR is to fix #1999 (a rather annoying and long-standing bug), though it also introduces a healthy bit of refactoring and a probably-useful card mention 

Screen Shot 2022-04-11 at 10 29 10 PM


maxwofford:
@cjdenio This is looking awesome! LMK when you want a review.

cjdenio:
@maxwofford this is ready for review!

app/controllers/static_pages_controller.rb
```diff
@@ -64,10 +64,8 @@ def my_missing_receipts_list
   end
 
   def my_inbox
-    @authorizations = current_user.stripe_authorizations.includes(stripe_card: :event).awaiting_receipt
-    @transactions = current_user.emburse_transactions.includes(emburse_card: :event).awaiting_receipt
-    @txs = @authorizations + @transactions
-    @count = @txs.size
+    @cards = current_user.stripe_cards.includes(:event).filter { |card| card.hcb_codes.missing_receipt.length > 0 }
```
maxwofford: Looks like this removes support for transactions on emburse cards. I think we'll want to keep that around.

app/views/static_pages/my_inbox.html.erb
```diff
@@ -12,32 +12,35 @@
...
-              <% else %>
-                <% hcb_code.canonical_pending_transactions.each do |pt| %>
-                  <%= render partial: "canonical_pending_transactions/canonical_pending_transaction", locals: { pt: pt, force_display_details: true } %>
-                <% end %>
-              <% end %>
-            <% end %>
-          </tbody>
-        </table>
+      <div class="flex items-center">
+        <%= inline_icon "card", class: "muted mr1", size: 25 %>
```
maxwofford: This link is a great touch! Love it. Do we want the icon to also be linked?

maxwofford (commented):
Great work! Added a couple comments. Go ahead and re-request review once they're ready.

app/views/static_pages/my_inbox.html.erb
cjdenio: Good catch!

app/controllers/static_pages_controller.rb
cjdenio: @maxwofford Hmm, it looks like the current setup doesn't support Emburse cards either (`EmburseCard` doesn't have an `hcb_codes` method):

https://github.com/hackclub/bank/blob/fdf01740d831d34a0679e1fc2e2315cc6555003e/app/views/static_pages/my_inbox.html.erb#L25

How important is it to have support for Emburse card transactions?

maxwofford: I think it's worth including support– eventually we'll have to clamp down on people not uploading receipts and I'd like the list of missing stuff to not surprise anyone.

Check out https://github.com/hackclub/bank/commit/480d96af11b53eb7f07e97021114cafe4cd04032 for a reference.

maxwofford: If it's what you're looking for feel free to [cherry pick!](https://git-scm.com/docs/git-cherry-pick)

cjdenio:
@maxwofford This should be all set– I'm not able to test Emburse card support, but it should work fine as long as your fancy new `hcb_codes` method works.

maxwofford (approved):
Just tested this & can confirm the emburse change works. Thanks for including that! Great work on this. Merge when ready.

cjdenio:
Awesome, thanks!

PR #2462, opened 2022-04-12
Move preloading logic into separate class

albertchae:
There is a significant amount of database querying + suturing associated
objects together in the upcoming improvements to this endpoint, so move
it all into a separate preloader class for better separation.

Also per the discussion in https://github.com/hackclub/bank/pull/2450#issuecomment-1095609557
add a preload param to gate the optimizations (although it was easier to
gate the existing optimization too because it'll all be in the same class)

This PR is mostly an [S](https://medium.com/@kentbeck_7670/bs-changes-e574bc396aaa) change


albertchae:
Re: https://github.com/hackclub/bank/pull/2450#issuecomment-1095609557

@maxwofford I ended up adding the `preload` query param here for the existing optimization because it'll be easier since all of the logic will live in this `AssociationPreloader` class

maxwofford:
Thanks for [the read](https://medium.com/@kentbeck_7670/bs-changes-e574bc396aaa)! I like that framework for units of change.

maxwofford (approved):
Great use of a service!

Want to flag that most of our services use a `run` method by default & arguments are taken in through the constructor– let's keep that unless there's a reason to deviate here. Happy for this to be merged though!

albertchae:
@maxwofford ah cool thanks for the heads up, no reason to deviate here. Do you distinguish between `run` and `run!` for pure functions/mutations? (I know this isn't consistent in the ruby community anyway)

maxwofford:
> Do you distinguish between `run` and `run!`

Not that I'm aware of, but that'd be a nice to add in a future refactor.

albertchae:
@maxwofford updated the preloader to follow the service class pattern and decided to use `run!` for now to signal mutation. I'll merge this later if you don't have objections.

BTW, do merges to `main` get deployed to production automatically or is there a manual step someone is running?

garyhtou:
> Do you distinguish between run and run!

> Not that I'm aware of, but that'd be a nice to add in a future refactor.

Yeah, all of our services are run `run`. although I'm down for picking up that convention

<br/>

>  (I know this isn't consistent in the ruby community anyway)

haha this is so true

---

> BTW, do merges to main get deployed to production automatically or is there a manual step someone is running?

Heroku will automatically deploy all changes pushed to main (including migrations)

PR #2465, opened 2022-04-13
Bulk check for existing HACK CLUB FEE for CanonicalTransaction in events#show

albertchae:

Query/allocation count before: SQL Queries: 1223 (520 cached) | Allocations: 3611538
Query/allocation count after: SQL Queries: 1124 (520 cached) | Allocations: 3290516

From this query:
```
  Fee Exists? (0.5ms)  SELECT 1 AS one FROM "fees" INNER JOIN "canonical_event_mappings" ON "fees"."canonical_event_mapping_id" = "canonical_event_mappings"."id" WHERE "canonical_event_mappings"."canonical_transaction_id" = $1 AND "fees"."reason" = $2 LIMIT $3  [["canonical_transaction_id", redacted], ["reason", "HACK CLUB FEE"], ["LIMIT", 1]]
  ↳ app/models/canonical_transaction.rb:164:in `fee_payment?'
```

which will be an N+1 here
https://github.com/hackclub/bank/blob/fdf01740d831d34a0679e1fc2e2315cc6555003e/app/models/canonical_transaction.rb#L164

Also fixed a subtle problem with memoizing `@fee_payment`. Since it stores
a boolean, if the value is `false`, it doesn't get cached with `||=` and
instead runs the `exists?` query again. So use a guard return with
`defined?` instead.

It might be worth fixing this in other places with memoization as a defensive
measure (`||=` will also not cache `nil`s properly), but that should be a separate task.


app/services/transaction_grouping_engine/transaction/association_preloader.rb
```diff
@@ -28,6 +29,16 @@ def preload_associations!
                                       .sort { |ct1, ct2| compare_date_id_descending(ct1, ct2) }
         end
 
+        hack_club_fees_by_canonical_transaction_id = Fee
+                                                     .includes(:canonical_event_mapping)
+                                                     .where(canonical_event_mappings: { canonical_transaction_id: canonical_transaction_ids })
+                                                     .hack_club_fee
+                                                     .index_by { |fee| fee.canonical_event_mapping.canonical_transaction_id }
```
albertchae: Subtle point here, `fee.canonical_transaction.id` will kick off another N+1 because of https://github.com/hackclub/bank/blob/fdf01740d831d34a0679e1fc2e2315cc6555003e/app/models/fee.rb#L50-L52, so get the id off `canonical_event_mapping` instead

maxwofford (approved):
LGTM!

PR #2466, opened 2022-04-13
Preload HashedTransaction and attach to CanonicalTransaction in events#show

albertchae:
Query/allocation count before: SQL Queries: 1124 (520 cached) | Allocations: 3561919

HashedTransaction was responsible for 2 N+1s
```
   (0.6ms)  SELECT COUNT(*) FROM "hashed_transactions" INNER JOIN "canonical_hashed_mappings" ON "hashed_transactions"."id" = "canonical_hashed_mappings"."hashed_transaction_id" WHERE "canonical_hashed_mappings"."canonical_transaction_id" = $1  [["canonical_transaction_id", 208036]]
  ↳ app/models/canonical_transaction.rb:255:in `hashed_transaction'
```

This was coming from
https://github.com/hackclub/bank/blob/0fd9f60a918aae4fe11ffdb8834e4aa0f34b33dd/app/models/canonical_transaction.rb#L251
a subtlety here is that `count` will issue a `COUNT` query to the DB
every time and not load the objects into memory, which can be more
performant but in this case since we're doing a `first` right after,
`length` or `size` may have been better to load the objects into memory
first.

```
  HashedTransaction Load (0.2ms)  SELECT "hashed_transactions".* FROM "hashed_transactions" INNER JOIN "canonical_hashed_mappings" ON "hashed_transactions"."id" = "canonical_hashed_mappings"."hashed_transaction_id" WHERE "canonical_hashed_mappings"."canonical_transaction_id" = $1 ORDER BY "hashed_transactions"."id" ASC LIMIT $2  [["canonical_transaction_id", 208036], ["LIMIT", 1]]
  ↳ app/models/canonical_transaction.rb:257:in `hashed_transaction'
```

Coming from
https://github.com/hackclub/bank/blob/0fd9f60a918aae4fe11ffdb8834e4aa0f34b33dd/app/models/canonical_transaction.rb#L253

Query/allocation count after: SQL Queries: 925 (520 cached) | Allocations: 2997399

There's also an additional N+1 for the `RawStripeTransaction` association on `HashedTransaction`

```
  RawStripeTransaction Load (0.3ms)  SELECT "raw_stripe_transactions".* FROM "raw_stripe_transactions" WHERE "raw_stripe_transactions"."id" = $1 LIMIT $2  [["id", 16155], ["LIMIT", 1]]
  ↳ app/models/canonical_transaction.rb:103:in `raw_stripe_transaction'
```

https://github.com/hackclub/bank/blob/0fd9f60a918aae4fe11ffdb8834e4aa0f34b33dd/app/models/canonical_transaction.rb#L100-L102

Query/allocation count after: SQL Queries: 887 (520 cached) | Allocations: 2832752


app/services/transaction_grouping_engine/transaction/association_preloader.rb
```diff
@@ -35,8 +35,17 @@ def preload_associations!
                                                      .hack_club_fee
                                                      .index_by { |fee| fee.canonical_event_mapping.canonical_transaction_id }
 
+        hashed_transactions_by_canonical_transaction_id = HashedTransaction
+                                                          .includes(:canonical_hashed_mapping, :raw_stripe_transaction)
+                                                          .where(canonical_hashed_mappings: { canonical_transaction_id: canonical_transaction_ids })
+                                                          .group_by { |ht| ht.canonical_hashed_mapping.canonical_transaction_id }
```
albertchae: same issues as https://github.com/hackclub/bank/pull/2465#discussion_r849209843

app/services/transaction_grouping_engine/transaction/association_preloader.rb
```diff
@@ -35,8 +35,17 @@ def preload_associations!
                                                      .hack_club_fee
                                                      .index_by { |fee| fee.canonical_event_mapping.canonical_transaction_id }
 
+        hashed_transactions_by_canonical_transaction_id = HashedTransaction
+                                                          .includes(:canonical_hashed_mapping, :raw_stripe_transaction)
```
albertchae: `.includes(:raw_stripe_transaction)` is the usual fix for N+1 type problems but it's the first tIme I got to use it for https://github.com/hackclub/bank/issues/2446 because we were mostly dealing with a struct (`CanonicalTransactionGrouped`) instead of real ActiveRecord associations

app/services/transaction_grouping_engine/transaction/association_preloader.rb
```diff
@@ -35,8 +35,17 @@ def preload_associations!
...
+                                                          .where(canonical_hashed_mappings: { canonical_transaction_id: canonical_transaction_ids })
+                                                          .group_by { |ht| ht.canonical_hashed_mapping.canonical_transaction_id }
+
         canonical_transactions.each do |ct|
           ct.fee_payment = hack_club_fees_by_canonical_transaction_id[ct.id].present?
+
+          hashed_transactions = hashed_transactions_by_canonical_transaction_id[ct.id]
+          Airbrake.notify("There was more (or less) than 1 hashed_transaction for canonical_transaction: #{ct.id}") if hashed_transactions.length != 1
```
albertchae: Original line https://github.com/hackclub/bank/blob/0fd9f60a918aae4fe11ffdb8834e4aa0f34b33dd/app/models/canonical_transaction.rb#L251

As I mentioned this can be an N+1 because of the subtle differences between [`count`, `length`, and `size`](https://blazarblogs.wordpress.com/2019/07/27/activerecord-size-vs-count-vs-length/) (`count` issues a new DB query, `length` loads objects into memory then counts them, `size` tries to use objects already loaded into memory if they exist, otherwise issues a `COUNT` to the db)

To further complicate things, `count` is a method on both [Enumerable](https://ruby-doc.org/core-2.7.5/Enumerable.html#method-i-count) and [ActiveRecord](https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html). It only exhibits the DB vs application behavior above when we're calling the `ActiveRecord` version.

So on this line, `hashed_transactions` is already an array in memory (due to the `group_by`, which will load the objects into memory and then group them into arrays), so using `count` will always hit the `Enumerable` version. But because this can be confusing, I just changed it from `count` to `length`.

BTW I think this is one of the leakiest abstractions with ActiveRecord and I kind of hate that I'm even aware of this distinction 😭 . I have to look up the differences between count, length and size every time.

albertchae:
Please review https://github.com/hackclub/bank/pull/2465 before this one

maxwofford (approved):

app/services/transaction_grouping_engine/transaction/association_preloader.rb
albertchae: @maxwofford btw, this seems like a data integrity issue that could be solved with a database constraint. Has this been considered?

maxwofford: That's a good observation. I think this is an artifact from when we switched from the first transaction engine (just using the `Transaction` model) to the second transaction engine (`RawXTransaction` -> `HashedTransaction` -> `CanonicalTransaction`). We can probably switch this over to a db constraint now.

cc'ing @garyhtou who may know more.

garyhtou: i'm not too familiar with this part of the TX engine, although it could have something to do with transactions ending up with the same hash (duplicates)

this folder contains much of the code for handling Hashed Transaction duplicates
https://github.com/hackclub/bank/tree/c83f9023e4f1009e4ea30571231b68aaee3d136b/app/services/transaction_engine/hashed_transaction_service

PR #2468, opened 2022-04-13
Change tab from Requests to Feedback

hackliv:
There was some feedback on #bank channel that "Requests" be changed to "Feedback" to be more clear for users.


cjdenio (approved):
This is much better, thanks!

PR #2470, opened 2022-04-14
Let there be 1Password

exu3:
This PR adds the events promo and admin card for 1Password.

Screen Shot 2022-04-24 at 00 21 57

Screen Shot 2022-04-24 at 00 21 52

The copy could probably use a pass.


maxwofford:
Could we mention that Hack Club uses 1Password ourselves?

maxwofford:
Forgot to mention, but really great job on the background- love how that looks in both themes.

MelSmith104:
Looks great! Good to ship :)

maxwofford:
And so there was 1Password, and it was good.

Orpheus 1:3

PR #2472, opened 2022-04-14
Explicitly memoize UserSession in SessionsHelper

albertchae:
This improvement is independent from the other association preload
related changes for https://github.com/hackclub/bank/issues/2446

Prior to this change, we were relying on ActiveRecord's built in SQL
caching https://guides.rubyonrails.org/caching_with_rails.html#sql-caching

```
  CACHE UserSession Load (0.0ms)  SELECT "user_sessions".* FROM "user_sessions" WHERE "user_sessions"."deleted_at" IS NULL AND "user_sessions"."session_token" IS NULL LIMIT $1  [["LIMIT", 1]]
  ↳ app/helpers/sessions_helper.rb:63:in `current_session'
```

While this is fine in terms of not hitting the DB so many times, we
still end up instantiating many ActiveRecord objects (in this case
`UserSession`, which can still adversely affect performance).

With ?preload=true
```
Completed 200 OK in 4301ms (Views: 3764.3ms | ActiveRecord: 275.3ms | SQL Queries: 709 (426 cached) | Allocations: 2379017)
Completed 200 OK in 2715ms (Views: 2180.3ms | ActiveRecord: 266.1ms | SQL Queries: 340 (57 cached) | Allocations: 1450341)
```

Without ?preload=true
```
Completed 200 OK in 6792ms (Views: 6021.3ms | ActiveRecord: 622.1ms | SQL Queries: 1342 (426 cached) | Allocations: 4111404)
Completed 200 OK in 5300ms (Views: 4551.6ms | ActiveRecord: 602.9ms | SQL Queries: 973 (57 cached) | Allocations: 3175519)
```

So this change explicitly memoizes `@potential_session` to reduce object
instantiation. I believe this should be safe because SessionHelper is
included on Api::V2Controller and ApplicationController, so the cookie
value should stay the same through the life of a single request (the
fact that ActiveRecord cache intercepted 350+ of these is also a sign of
this). The only behavior change I can see here is that if during the
processing of a request, we cross the threshold of a session expiring,
previously we would expire that session midway whereas now we would run
to the end. Honestly this seems fine to me, but if anyone disagrees let
me know.


app/helpers/sessions_helper.rb
```diff
@@ -59,22 +59,24 @@ def current_user(_ensure_api_authorized = true)
...
+    return nil unless @potential_session
 
     # check if the potential session is still valid
     # If the session is greater than the expiration duration then the current
     # user is no longer valid.
     # (.abs) is added for easier testing when fast-forwarding created_at times
-    if (Time.now - potential_session.created_at).abs > EXPIRATION_DURATION
-      potential_session.set_as_peacefully_expired
-      potential_session.destroy
+    if (Time.now - @potential_session.created_at).abs > EXPIRATION_DURATION
```
albertchae: Before this change, it's possible during the life of a single request that `Time.now` advanced to the point where the session expires. With the memoized value, the session will only be expired if that `Time.now` check was past the expiration point the first time.

maxwofford (approved):
This is a great catch. Happy with the code going in, but the staging environment isn't loading correctly for me. I think that's a db thing (unrelated to your code), but could you sanity check that this doesn't affect signing out a user through the sign out button? That's the only case I'd think needs to change the value of `current_session` during a request.

albertchae:
@maxwofford just messed around with sign in/sign out flow locally a bit and it looked OK, but I'm still concerned since this is a riskier change compared to the other PRs. Is there any particular time I should merge it where it's more likely for someone to be around to revert it if necessary?

maxwofford:
Anytime during EST working hours is good. Going to cc @garyhtou so he gets updates when it goes live

app/helpers/sessions_helper.rb
```diff
@@ -59,22 +59,24 @@ def current_user(_ensure_api_authorized = true)
   end
 
   def current_session
+    return @potential_session if defined?(@potential_session)
```
garyhtou: this isn't really important at all, but should this be memoized as `@current_session` so that it's the same name as the method?

reasoning: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/MemoizedInstanceVariableName

app/helpers/sessions_helper.rb
garyhtou: good catch!

garyhtou:
> Anytime during EST working hours is good. Going to cc @garyhtou so he gets updates when it goes live

I'm PST and usually up pretty late. feel free to text/call if anything goes awry 4255241846

albertchae:
Thanks @garyhtou @maxwofford I'll plan on merging tomorrow morning PT

app/helpers/sessions_helper.rb
albertchae: hm yeah that seems OK to me. Not sure what's "potential" about it except that I guess it could be expired/destroyed? lol

albertchae: I guess it's "potentially" nil (as are most things 😝). Nevertheless, renamed the variable to `@current_session`

albertchae:
@maxwofford @garyhtou merging now 🚀 🤞

PR #2473, opened 2022-04-15
Associate organizer position invites to user

garyhtou:
wip


albertchae:
Hey @garyhtou, @maxwofford and I discussed an alternate way to solve the problem of having to associate a user later by guaranteeing the user exists when the invite is created, which is being worked on in https://github.com/hackclub/bank/pull/2650, so I'm going to close this PR. Thanks for your investigation and honestly I'm not sure why they didn't work... but hopefully we won't have to worry about it happening again 🤞

PR #2475, opened 2022-04-15
Preload some HcbCode associations for events#show

albertchae:
Query/allocation count before: SQL Queries: 666 (426 cached) | Allocations: 2609413

Preloading receipts removes this N+1

```
   (0.3ms)  SELECT COUNT(*) FROM "receipts" WHERE "receipts"."receiptable_id" = $1 AND "receipts"."receiptable_type" = $2  [["receiptable_id", 25571], ["receiptable_type", "HcbCode"]]
  ↳ app/views/canonical_transactions/_canonical_transaction.html.erb:53
```

Query/allocation count after: SQL Queries: 567 (426 cached) | Allocations: 2245983

Preloading comments removes this N+1
```
   (0.4ms)  SELECT COUNT(*) FROM "comments" WHERE "comments"."commentable_id" = $1 AND "comments"."commentable_type" = $2 AND "comments"."admin_only" = $3  [["commentable_id", 25382], ["commentable_type", "HcbCode"], ["admin_only", false]]
  ↳ app/views/canonical_transactions/_canonical_transaction.html.erb:54
```

Completed 200 OK in 16788ms (Views: 2496.5ms | ActiveRecord: 126.3ms | SQL Queries: 468 (426 cached) | Allocations: 1673199)

Both preloading receipts and comments will work when calling size (e.g.
in `_canonical_transaction.html.erb` with `ct.local_hcb_code.receipts.size`
and `ct.local_hcb_code.comments.size`)

BUT comments also has a different count if you're not logged in as an admin
`ct.local_hcb_code.comments.not_admin_only.size`, which doesn't take
advantage of the preloaded comments because `not_admin_only` is a scope,
so it has to run that scope as a SQL query again and then take the
`size` after. Whether it's faster to do the count in memory in the application or
via the DB honestly depends (and has to observed in production), but
given it doesn't seem like there are that many comments per HcbCode, I
chose to go with counting in memory for now. I added a `not_admin_only_comments_count`
method that can use the memoized value if ?preload=true or falls back to
the old behavior (not_admin_only scope + size)


app/views/canonical_transactions/_canonical_transaction.html.erb
```diff
@@ -51,7 +51,7 @@
 
     <% if ct.local_hcb_code %>
       <%= list_badge_for ct.local_hcb_code.receipts.size, "receipt", "payment-docs", optional: !ct.raw_stripe_transaction, required: ct.raw_stripe_transaction && !ct.local_hcb_code.no_or_lost_receipt? %>
-      <%= list_badge_for admin_signed_in? ? ct.local_hcb_code.comments.size : ct.local_hcb_code.comments.not_admin_only.size, "note", "post", optional: true %>
+      <%= list_badge_for admin_signed_in? ? ct.local_hcb_code.comments.size : ct.local_hcb_code.not_admin_only_comments_count, "note", "post", optional: true %>
```
albertchae: `comments.not_admin_only.size` has to query again with the scope. This is what I see in the logs
```
(0.5ms)  SELECT COUNT(*) FROM "comments" WHERE "comments"."commentable_id" = $1 AND "comments"."commentable_type" = $2 AND "comments"."admin_only" = $3
```

maxwofford (approved):
LGTM!

PR #2476, opened 2022-04-15
Fix bad variable renaming for @current_session

albertchae:
Follow up to bad refactoring during https://github.com/hackclub/bank/pull/2472

Sorry, dumb mistake on my part using search and replace. Fortunately the change in #2472 was a no-op change because it was never using the memoized the value and just doing essentially the same thing as before.


albertchae:
BTW do we not require review for merges to `main`? I see this
image

maxwofford (approved):

maxwofford:
@albertchae merging to `main` requires a review approval, _or_ CI to pass. Direct pushes to `main` are blocked.

albertchae:
@maxwofford got it, thanks

PR #2478, opened 2022-04-16
Prefill name, email, and organization on Feedback form

exu3:
Name, Email, and Organization fields are prefilled if the user clicks on the Feedback form link in the navbar.


app/views/users/_nav.html.erb
```diff
@@ -16,7 +16,7 @@
       'payment-docs', my_inbox_path %>
     <%= dock_item 'Feedback',
       'Submit user suggestions.',
-      'message-new', "https://hack.af/bank-userfeedback" %>
+      'message-new', "https://hack.af/bank-userfeedback?prefill_Name=#{current_user.name}&prefill_Email=#{current_user.email}&prefill_Organization=#{current_user.events.first.name}" %>
```
cjdenio: You may want to handle the very-edge-case situation where the current user isn't in any organizations. (`current_user.events.first&.name`)

###### :no-organizations-head-empty:

exu3: Ah good call, thanks!

hackliv:
Hey how does this work if the user is a part of numerous orgs?

exu3:
@hackliv It will prefill with the event whose ID is first numerically. For example, Hack Club HQ (183) comes before Hack Club Bank (636).

hackliv:
@exu3 perfect! thanks for clarification

hackliv (approved):

PR #2481, opened 2022-04-18
More efficient rendering of partials in a loop

albertchae:
Previously, we see many lines repeated of
```
Rendered canonical_transactions/_canonical_transaction.html.erb (Duration: 4.9ms | Allocations: 1405)
Rendered canonical_transactions/_canonical_transaction.html.erb (Duration: 4.3ms | Allocations: 1416)
Rendered canonical_transactions/_canonical_transaction.html.erb (Duration: 4.4ms | Allocations: 1416)
```

This is because the template gets reread from disk every loop iteration.
Fortunately the partials API has a way to read it once and reuse it for
the whole loop.

After:
```
Rendered collection of canonical_pending_transactions/_canonical_pending_transaction.html.erb [0 times] (Duration: 0.0ms | Allocations: 34)
Rendered collection of canonical_transactions/_canonical_transaction.html.erb [100 times] (Duration: 398.0ms | Allocations: 130029)
```

Local timing (absolute values aren't informative but relative are):
Before: 827.9 ms in views `Completed 200 OK in 1409ms (Views: 827.9ms | ActiveRecord: 135.4ms | SQL Queries: 25 (1 cached) | Allocations: 605016)`
After: 714.2 ms in views `Completed 200 OK in 1325ms (Views: 714.2ms | ActiveRecord: 138.8ms | SQL Queries: 25 (1 cached) | Allocations: 603067)`

More reading:

https://scoutapm.com/blog/performance-impact-of-using-ruby-on-rails-view-partials
https://github.com/rails/rails/issues/41452
https://www.justinweiss.com/articles/how-much-time-does-rendering-a-partial-really-take/


maxwofford (approved):
Nice catch! I bet there are a couple instances of this on other pages too.

albertchae:
@maxwofford yeah I wouldn't be surprised, I wasn't aware of this before and feel like a lot of tutorials teach the `each`+`render` partial pattern

Created https://github.com/hackclub/bank/issues/2484 to follow up

PR #2482, opened 2022-04-18
Cache inline_icon to remove unnecessary file I/O

albertchae:
View rendering seemed to be taking a lot of time, especially in the loop of `_canonical_transaction` partials
```
Rendered collection of canonical_transactions/_canonical_transaction.html.erb [100 times] (Duration: 408.7ms | Allocations: 129325)
```

Using the very scientific process 😉  of commenting out chunks of
`_canonical_transaction.html.erb` to see which parts contributed the most
to the time, I tracked it down to https://github.com/hackclub/bank/blob/f649ff1a3a1ceeb38fd3d122ea8c40d6c1c3238b/app/views/canonical_transactions/_canonical_transaction.html.erb#L53-L54

Which calls `list_badge_for` which calls `inline_icon` which loads an svg file 2*100 times

Add a cache for the parsed SVG XML object in `inline_icon` based on filename since we might change the image size dynamically.

Also move `icon` loading until after the optional guard statement, since
it's possible we won't need to read in the SVG at all if the entire page
has a count of 0

Rendering time after:
```
Rendered collection of canonical_transactions/_canonical_transaction.html.erb [100 times] (Duration: 134.0ms | Allocations: 53993)
```


maxwofford (approved):
Nice catch! Wonder if it makes sense to add a cache at the `inline_icon` level? I think that's where the actual icon file is being pulled. https://github.com/hackclub/bank/blob/53e7719f18c10e1229f80521a93a87219e5c9120/app/helpers/application_helper.rb#L146-L148

albertchae:
@maxwofford I moved the cache into `inline_icon` (but kept calling `inline_icon` in `list_badge_for` after the optional check)

garyhtou:
fyi https://github.com/hackclub/bank/issues/2494

PR #2490, opened 2022-04-19
Receipts → Feedback

kunalbotla:
Updating the bell/notification icon to be the new receipts icon

image


hackliv:
I actually was thinking of removing that entirely since it's repetitive. Thoughts?

garyhtou:
yeah i agree, i think we should remove, but then put the feedback button in its place

kunalbotla:
i like that; i'll make that change

kunalbotla:
updated! 
image

garyhtou (changes requested):
please assign for review before merging next time! (no need to revert)

Can we remove the Feedback navigation dock item? (in another pr)
image

kunalbotla:
sorry! i thought it was small enough to merge the change based the original review. i'll make that new pr

PR #2492, opened 2022-04-19
Skip generating pending TXs on invoices manually marked as paid

maxwofford:
Closes #2090 

Before this change the `TransactionEngineJob::Nightly` would generate a $350 charge shown below:

Now, the charge doesn't show:

Screen Shot 2022-04-19 at 12 42 52


garyhtou:
@maxwofford let's jump on a call about this. i think this specific case is a miss use of the marked as paid feature

free to drop something on my calendar after 4pm pst today. or we can cover this during our sync on thursday

maxwofford:
@garyhtou just sent a cal invite for today– b/c this is impacting a user let's try to get it wrapped up today.

garyhtou:
tldr/things to look into before we meet:
- invoices marked as paid are in fact supposed to generate pending tx until a matching CT is found
- for this specific case, we should probably undo the mark as paid and archive it since it was never paid
    - liv said archiving wasnt hiding it. there could be a bug here

maxwofford:
Merging this after review with @garyhtou on a Zoom call

garyhtou:
Context: This prevents creating pending transactions for all future invoices that are manually marked as paid since not all manually marked as paid invoices will have an exact amount matching transactions which "pays" it. because of this, pending tx can be left dangling on the ledger.

we decided that this process of matching a manually marked as paid invoice to a CT should be done by hand (if we do in fact find this information important to know)

PR #2493, opened 2022-04-19
Make WebAuthn login less confusing

cjdenio:
https://github.com/hackclub/bank/pull/2479#issuecomment-1102852022


maxwofford (approved):
Thanks for adding this! Just want to check– does this handle the case of a user having SMS auth setup? If so, go for merge! If not, could we redirect to SMS auth first for users that have opted-in?

cjdenio:
Currently the priority is WebAuthn -> SMS -> Email; are you thinking that SMS should have higher priority over WebAuthn?

maxwofford:
Ah, think i misread the `form` tag change– looks good for me!

cjdenio:
Great, thanks!

PR #2497, opened 2022-04-20
Dupe icon svg from cache before making possible mutations

albertchae:
Fixes https://github.com/hackclub/bank/issues/2494

Bug introduced in https://github.com/hackclub/bank/pull/2482. I didn't
realize we mutate the parsed nokogiri object that's stored in the cache
so we could end up having incorrect styles. This fixes it by creating a
dupe of the parsed object before mutating for styles.

This doesn't seem to have affected endpoint time that much
```
  Rendered collection of canonical_transactions/_canonical_transaction.html.erb [100 times] (Duration: 43.8ms | Allocations: 41525)
```

This line is roughly the same before and after this change, which is
still an improvement over the ~200ms we were seeing without any caching


garyhtou (approved):
this looks good!

---

just wondering: since these instance variables get reset per request, would it be more efficient to preserve the cache across requests with something like `Rails.cache`?
https://guides.rubyonrails.org/caching_with_rails.html#low-level-caching

albertchae:
@garyhtou good question! I actually think it's a feature that this icon cache gets reset per request because it means we don't have to think too hard about cache invalidation. I'm sure you've seen or heard of this joke https://martinfowler.com/bliki/TwoHardThings.html but if we used the Rails cache, we'd also have to make sure we invalidate the cache correctly if any of those images ever changed.

Also the main reason to add this cache was to save on file I/O and I believe [Rails cache is configured to redis](https://github.com/hackclub/bank/blob/16d68827c0eb86cf0f1fe959c6dd912e2dcea7c9/config/environments/production.rb#L62) and I'm not sure how fast disk I/O vs network and redis I/O would be (or if the differences would even be consistent).

The more common thing I've seen with images is to host them on a CDN and then point to that via an image tag. That way the browser can fetch images in parallel (or have them cached) with the HTML for the page itself. Actually now that I think about, we're probably inlining the SVG text multiple times on the page...

I just tried doing an inspect element on an icon on https://bank.hackclub.com/hq

```
<svg fill-rule="evenodd" clip-rule="evenodd" stroke-linejoin="round" stroke-miterlimit="1.414" xmlns="http://www.w3.org/2000/svg" aria-label="payment-docs" viewBox="0 0 32 32" preserveAspectRatio="xMidYMid meet" fill="currentColor" width="20" height="20" aria-hidden="true" style=""><g><path fill-rule="evenodd" clip-rule="evenodd" d="M21.048 6.291C19.836 6.05 18.256 6 16 6c-2.256 0-3.836.05-5.048.291-1.11.216-1.577.533-1.853.864-.32.384-.641 1.094-.844 2.597C8.046 11.283 8 13.274 8 16c0 2.725.046 4.717.255 6.248.203 1.503.524 2.213.844 2.597.276.331.743.648 1.853.864C12.164 25.95 13.744 26 16 26c2.256 0 3.836-.05 5.048-.291 1.11-.216 1.577-.533 1.853-.864.32-.384.641-1.094.843-2.597.21-1.531.256-3.523.256-6.248 0-2.726-.046-4.717-.256-6.248-.202-1.503-.523-2.213-.843-2.597-.276-.331-.743-.648-1.853-.864zM26 16c0-11-.833-12-10-12S6 5 6 16s.833 12 10 12 10-1 10-12z"></path><path fill-rule="evenodd" clip-rule="evenodd" d="M16 7a1 1 0 0 0-1 1v2.019l2-.004V8a1 1 0 0 0-1-1zM17 24v-2.015h-2V24a1 1 0 0 0 2 0z"></path><path fill-rule="evenodd" clip-rule="evenodd" d="M21.909 12.004c.107.542-.358.995-.91.995-.552.001-.942-.492-1.25-.95C19.347 11.451 19.25 11 16 11s-4 0-4 2 .606 2 4 2c5.143 0 6 .8 6 4 0 2.95-1.275 3.86-5 3.985-.631.021-1.369.021-2 0-2.872-.096-4.559-.785-4.924-2.988-.091-.545.372-.997.924-.997s.925.463 1.075.995c.208.736.99 1.005 3.925 1.005 4 0 4-.5 4-1.998v-.004C20 17.5 20 17 16 17c-5.143 0-6-1-6-4 0-2.766 1.275-3.832 5-3.981a28.378 28.378 0 0 1 2-.004c3.006 0 4.537 1.096 4.909 2.989z"></path></g></svg>
```

Which is a decent amount of text to be repeated ~80 times on a page... but then again looking at the size of the page after gzip `34.70 KB (331.35 KB size)` it's not that large, so not sure how much of an impact this would have. It's possible gzip is already taking care of a lot of that bloat

maxwofford (approved):
Thanks for the fix! Go for merge

maxwofford:
Just adding for the record that we originally inlined our SVGs b/c the connection overhead of so many requests (even from a CDN) made time to render higher than inlining.

albertchae:
@maxwofford Got it thanks for the context! I actually just thought of another way to slim out the HTML file from repeated SVGs. Looks like there is a way to inline the SVG once and then reference it by id https://css-tricks.com/use-and-reuse-everything-in-svg-even-animations/ . I might try looking into this later

PR #2502, opened 2022-04-23
Disbursement Confirmation Letter

exu3:
Closes #1425

Screen Shot 2022-04-23 at 00 53 08
Screen Shot 2022-04-23 at 00 53 44


app/controllers/disbursements_controller.rb
```diff
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 class DisbursementsController < ApplicationController
-  before_action :set_disbursement, only: [:show, :edit, :update]
+  before_action :set_disbursement, except: [:new, :create, :index]
+  before_action :set_event, only: [:new, :create]
+  skip_before_action :signed_in_user
```
garyhtou: were these changes necessary? it seems like it might break the edit path and relax security in some places (I haven't tested to see if this is the case tho)

garyhtou (changes requested):
overall, really good!

I took a look at my old ACH Transfer Confirmation Letter PR and noticed a few things that could be improved/fixed and made those changes here: https://github.com/hackclub/bank/pull/2505

could you incorporate the changes from that PR into this one as well?

---

just a thought, would you also like to add a link to the confirmation letter from the HCB show page?

example (ach transfer): 
https://bank.hackclub.com/hcb/PVHXVN
https://github.com/hackclub/bank/blob/8a3300905c0c5fb5640ccb5d0e56247f228d0369/app/views/hcb_codes/_ach_transfer.html.erb#L63-L69

it would belong in this file: https://github.com/hackclub/bank/blob/8a3300905c0c5fb5640ccb5d0e56247f228d0369/app/views/hcb_codes/_disbursement.html.erb

app/controllers/disbursements_controller.rb
exu3: I reverted the changes to lines 5 and 6, and added the `:transfer_confirmation_letter` action to line 4 which makes everything work.

---

~~Just line 4 was necessary.~~

~~With the previous `before_action`, I got a `
Pundit::NotDefinedError in DisbursementsController#transfer_confirmation_letter` (unable to find policy `NilClassPolicy` for `nil`)~~

~~I wasn't sure what it was missing, so I just copied the `before_action` from the ACH Transfers controller, and it seemed to work. Do you know what action(s) might be missing from the original `before_action :set_disbursement, only: [:show, :edit, :update]`?~~

garyhtou: that looks good!

app/services/disbursement_service/preview_transfer_confirmation_letter.rb
```diff
@@ -14,7 +14,7 @@ def run
     private
 
     def pdf_string
-      @pdf_string ||= ActionController::Base.new.render_to_string pdf: "transfer_confirmation_letter", template: "disbursement/transfer_confirmation_letter.pdf.erb", encoding: "UTF-8", locals: { :@disbursement => @disbursement, :@event => @event }
+      @pdf_string ||= ApplicationController::Base.new.render_to_string pdf: "transfer_confirmation_letter", template: "disbursement/transfer_confirmation_letter.pdf.erb", encoding: "UTF-8", locals: { :@disbursement => @disbursement, :@event => @event }
```
garyhtou:
```suggestion
      @pdf_string ||= ApplicationController.new.render_to_string pdf: "transfer_confirmation_letter", template: "disbursement/transfer_confirmation_letter.pdf.erb", encoding: "UTF-8", locals: { :@disbursement => @disbursement, :@event => @event }
```

app/controllers/disbursements_controller.rb
```diff
@@ -15,6 +15,30 @@ def show
     @hcb_code = HcbCode.find_or_create_by(hcb_code: @disbursement.hcb_code)
   end
 
+  def transfer_confirmation_letter
+    authorize @disbursement
+
+    respond_to do |format|
+      unless @disbursement.fulfilled?
+        redirect_to @ach_transfer and return
```
```suggestion
        redirect_to @disbursement and return
```

garyhtou (approved):

garyhtou:
good work!

just a few small edits

app/controllers/disbursements_controller.rb
exu3: Oops, my bad. Thanks for catching!

app/controllers/disbursements_controller.rb
garyhtou: some context to help explain what these callbacks do:

---

```ruby
before_action :set_disbursement, except: [:new, :create, :index]
```
Before every single action, except the `new`, `create`, and `index` actions, run the method called `set_disbursement`. The `set_disbursement` method gets the `id` from the URL params and sets the instance variable `@disbursement`.
https://github.com/hackclub/bank/blob/af69f600ca0faa62d7d8d163cf752aec2f4c896b/app/controllers/disbursements_controller.rb#L110-L115

This means we no longer need to repeat `@disbursement = Disbursement.find(params[:id] || params[:disbursement_id])` in all those methods.

---

```ruby
  before_action :set_event, only: [:new, :create]
```
Before the `new` and `create` actions, the run the method called `set_event`. however, there's no `set_event` method in the DisbursementsController

---

```ruby
  skip_before_action :signed_in_user
```

This one's a little harder to understand. In the ApplicationController, we have `before_action :signed_in_user`. In other words, for every single action in the whole application, run the method `signed_in_user` before the action. The method `signed_in_user` will check if there is a user signed in. If not, redirect them to the sign in page.

https://github.com/hackclub/bank/blob/af69f600ca0faa62d7d8d163cf752aec2f4c896b/app/controllers/application_controller.rb#L9-L12
https://github.com/hackclub/bank/blob/af69f600ca0faa62d7d8d163cf752aec2f4c896b/app/helpers/sessions_helper.rb#L82-L87

However, there are some pages that we don't want to require a user to be signed in (e.g. receipt uploading on HCB page). So, we can use the `skip_before_action :signed_in_user` callback to prevent the `signed_in_user` method from running.

Just saying `skip_before_action :signed_in_user` will skip it for EVERY action in the controller. If we wanted to skip it for just one action, we could do something like `skip_before_action :signed_in_user, only: :transfer_confirmatioon_letter`

exu3: Ah that's super helpful. Thanks for the detailed breakdown of what each filter does!

PR #2503, opened 2022-04-24
Use Monetize/BigDecimal instead of Float for ledger search amount

albertchae:
Fixes https://github.com/hackclub/bank/issues/2416

The reason this is happening is
```
> "319.78".to_f
=> 319.78
> "319.78".to_f * 100
=> 31977.999999999996
> ("319.78".to_f * 100).to_i
=> 31977
```

Floats cannot represent all possible numbers and especially as you get
to larger numbers there are more and more gaps in precision
(see https://en.wikipedia.org/wiki/File:FloatingPointPrecisionAugmented.png)

I originally fixed this using BigDecimal directly, but I saw the project
is including money-rails, so I instead switched it to Monetize.parse
because it expresses the intent of this line a bit clearer and also still
uses BigDecimal under the hood, which can represent numbers with arbitrary precision.

```
> BigDecimal("319.78") * 100
=> 0.31978e5
> (BigDecimal("319.78") * 100).to_i
=> 31978
> Monetize.parse("319.78").cents
=> 31978
```

This is why it's generally recommended to use a data type with arbitrary
precision when dealing with situations like money
https://dzone.com/articles/never-use-float-and-double-for-monetary-calculatio
https://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency


app/controllers/admin_controller.rb
```diff
@@ -383,7 +383,7 @@ def ledger
 
     if @q
       if @q.to_f != 0.0
```
albertchae: I considered changing this line but what this line really seems to be checking "is `@q` a string that represents a monetary amount?", which unfortunately isn't very clean in ruby.

Options:
- we keep `to_f` and compare against 0.0 which has gotchas like `"12.34blahblah".to_f => 12.34` (maybe that's actually a feature for the use case of an admin only search)
- Use regex https://stackoverflow.com/questions/1286870/ruby-implementation-is-numeric-for-strings-need-better-alternatives
- Convert to Float using the constructor and catch exception, which gets into the whole "exceptions are not control flow" debate https://mentalized.net/journal/2011/04/14/ruby-how-to-check-if-a-string-is-numeric/

albertchae: oh I didn't realize Ruby 2.6 added a way to get around the exception https://www.bigbinary.com/blog/ruby-2-6-adds-option-to-not-raise-exception-for-integer-float-methods

```
> Float("12.34asdf", exception: false)
=> nil
> Float("12.34", exception: false)
=> 12.34
```

garyhtou (approved):
amazing! thanks for walking through your thought process and the various options

PR #2504, opened 2022-04-24
Allow Users to choose Session length

albertchae:
Closes https://github.com/hackclub/bank/issues/2444

**Best reviewed by commits**

I'm not sure if storing this on `users` makes sense or if there is some kind of table like user_settings where these sorts of things belong. But I made that assumption for now and ran with it. Given that assumption, the easiest place to put the setting was with the rest of the user edit form, but it's possible it might make sense lower on the page under `Login Details` (maybe in the `Sessions` section?). One drawback of moving it there would be that it could complicate the UX of the form (2 separate form sections on the page with separate submits, or a modal like the security key)

- Add `session_duration_seconds` column to users with a default of 30 days (to replace the hardcoded constant)
- Use `session_duration_seconds` when checking for session expiration instead of hardcoded constant
- Add `session_duration_seconds` to user settings form with the options specified in #2444

Screen Shot 2022-04-24 at 12 00 15 PM
Screen Shot 2022-04-24 at 12 00 20 PM


app/helpers/sessions_helper.rb
```diff
@@ -70,7 +77,7 @@ def current_session
     # If the session is greater than the expiration duration then the current
     # user is no longer valid.
     # (.abs) is added for easier testing when fast-forwarding created_at times
-    if (Time.now - @current_session.created_at).abs > EXPIRATION_DURATION
+    if (Time.now - @current_session.created_at).abs > current_session.user.session_duration_seconds
```
albertchae: I just realized the original issue actually said to store an `expiration` on `UserSession`, which is definitely doable, but possibly a little weird for existing sessions if a user changes their `session_duration_seconds`?

app/helpers/sessions_helper.rb
```diff
@@ -1,16 +1,23 @@
...
 
   # DEPRECATED - begin to start deprecating and ultimately replace with sign_in_and_set_cookie
-  def sign_in(user:, fingerprint_info: {}, impersonate: false, duration: EXPIRATION_DURATION, webauthn_credential: nil)
+  def sign_in(user:, fingerprint_info: {}, impersonate: false, webauthn_credential: nil)
     session_token = SecureRandom.urlsafe_base64
-    cookies.encrypted[:session_token] = { value: session_token, expires: Time.now + duration }
+    cookies.encrypted[:session_token] = { value: session_token, expires: Time.now + user.session_duration_seconds }
     user_session = user.user_sessions.create(
       session_token: session_token,
       fingerprint: fingerprint_info[:fingerprint],
```
albertchae: add expiration time here based on result of discussion in https://github.com/hackclub/bank/pull/2504#discussion_r857168579

app/helpers/sessions_helper.rb
```diff
@@ -1,16 +1,23 @@
...
 
 module SessionsHelper
-  EXPIRATION_DURATION = 30.days
+  SESSION_DURATION_OPTIONS = {
+    "1 hour" => 1.hour.to_i,
+    "1 day" => 1.day.to_i,
+    "3 days" => 3.days.to_i,
+    "7 days" => 7.days.to_i,
+    "14 days" => 14.days.to_i,
+    "30 days" => 30.days.to_i
```
albertchae: wasn't sure if there's a good way to do this string to seconds mapping. From https://stackoverflow.com/a/9916691, I found

- https://apidock.com/rails/ActionView/Helpers/DateHelper/distance_of_time_in_words
which says "about 1 hour" and "about 1 month" for the first and last options
- `ActiveSupport::Duration.build(total_seconds).inspect` which says "1 week", "2 weeks", "4 weeks and 2 days" for the last 3 options

So I just chose to do it manually instead for the greatest control of the display text

app/helpers/sessions_helper.rb
albertchae: Maybe we should expire all existing sessions if the user changes their `session_duration_seconds`?

app/views/users/edit.html.erb
```diff
@@ -64,6 +64,10 @@
         <%= form.label :pretend_is_not_admin, 'Pretend not to be an admin on Bank' %>
       </div>
     <% end %>
+    <div>
+      <%= form.label :session_duration_seconds, "How long your session lasts before you have to login again" %>
```
albertchae: I'm sure the copy here could be better, if anyone has suggestions

app/helpers/sessions_helper.rb
```diff
@@ -1,16 +1,23 @@
...
+  }
 
   def impersonate_user(user)
     sign_in(user: user, impersonate: true)
   end
 
   # DEPRECATED - begin to start deprecating and ultimately replace with sign_in_and_set_cookie
-  def sign_in(user:, fingerprint_info: {}, impersonate: false, duration: EXPIRATION_DURATION, webauthn_credential: nil)
```
albertchae: all call sites of `sign_in` I could find weren't passing in a `duration`, so it seemed safe to remove

app/helpers/sessions_helper.rb
```diff
@@ -70,7 +77,7 @@ def current_session
     # If the session is greater than the expiration duration then the current
     # user is no longer valid.
     # (.abs) is added for easier testing when fast-forwarding created_at times
-    if (Time.now - @current_session.created_at).abs > EXPIRATION_DURATION
+    if (Time.now - @current_session.created_at).abs > current_session.user.session_duration_seconds
       @current_session.set_as_peacefully_expired
       @current_session.destroy
```
albertchae: I realize we are setting the expiration on the browser side cookie, but on server side do we have any periodic reaper or cleanup job that goes and peacefully expires/destroys old sessions? Or does that only happen the next time a user with such a session tries to access the site?

garyhtou: The latter. It only happens the next time a user signs in with that expired session/cookie

app/views/users/edit.html.erb
garyhtou: > Stay signed in for

> Session expiration

cc @exu3

app/helpers/sessions_helper.rb
garyhtou: The intention was for this setting to only affect future sessions. The user settings would change `session_duration_seconds`, and newly created sessions would use `user.session_duration_seconds` to calculate an `expiration_at`. otherwise, a user changing their `session_duration_seconds` can cause them to get kicked out of their current session. to be fair, the issue could have been more descriptive!

I'm totally down with any way you implement it. however, I think from a UX perspective, a user shouldn't get kicked out of their current session after updating their session duration

---

example of getting signed out after changing session duration:
image

I logged in a few days ago. when to settings to change session duration to 1 hour and got signed out

app/helpers/sessions_helper.rb
garyhtou: since we're allowing users to choose their session lengths, go ahead and remove this usage of the `duration` param:

https://github.com/hackclub/bank/blob/bf821447068eb0c77c413446926eb0cd01cd9636/app/controllers/api/v2_controller.rb#L43-L46

app/helpers/sessions_helper.rb
garyhtou: i think what you have is totally fine!

garyhtou:
> I'm not sure if storing this on users makes sense or if there is some kind of table like user_settings where these sorts of things belong.

yeah, stuff like this is just in the `users` table. i think in the future, we should maybe split `User` into multiple models: `User`, `Profile`, `Email`, `Phone` to declutter `User` and also support multiple email addresses

> Given that assumption, the easiest place to put the setting was with the rest of the user edit form, but it's possible it might make sense lower on the page under Login Details (maybe in the Sessions section?). One drawback of moving it there would be that it could complicate the UX of the form (2 separate form sections on the page with separate submits, or a modal like the security key)

Yeah, I agree with you here. It would be nice to have it with the Sessions section, but I'm not sure what the UI/UX would look like for that. Maybe we can leave it where you have it for the time being and plan to split settings into multiple pages with a navigation (similar to github: https://github.com/settings)

app/helpers/sessions_helper.rb
albertchae: Ah yeah good point, I didn't think that even this current implementation could lead you to getting logged out right away. I'll add the `expiration_at` change to this PR

app/helpers/sessions_helper.rb
albertchae: oops I totally missed this one. Multiline arguments got me 😅

app/views/users/edit.html.erb
exu3: Maybe

> Expire sessions after

app/helpers/sessions_helper.rb
albertchae: added `expiration_at`

app/helpers/sessions_helper.rb
albertchae: Done in https://github.com/hackclub/bank/pull/2504/commits/cf8f57ebfcb9ddddcb4bce6939147245bfb1107b

app/views/users/edit.html.erb
albertchae: I like either `Stay signed in for` or `Expire sessions after`. I committed `Stay signed in for` for now, but still open to changing this

app/helpers/sessions_helper.rb
albertchae: Removed it with the comment https://github.com/hackclub/bank/pull/2504/commits/5370d7f54d042f308562e99eee34507b99fc175f

BTW does anyone know why the api controller logged in for half the session duration?

albertchae:
> Yeah, I agree with you here. It would be nice to have it with the Sessions section, but I'm not sure what the UI/UX would look like for that. Maybe we can leave it where you have it for the time being and plan to split settings into multiple pages with a navigation (similar to github: https://github.com/settings)

👍 , that github page is a good example because you can check some of the boxes in a section but hit the submit button of a different section to see that it only saves that section

app/helpers/sessions_helper.rb
```diff
@@ -69,8 +78,7 @@ def current_session
     # check if the potential session is still valid
     # If the session is greater than the expiration duration then the current
     # user is no longer valid.
-    # (.abs) is added for easier testing when fast-forwarding created_at times
-    if (Time.now - @current_session.created_at).abs > EXPIRATION_DURATION
+    if Time.now > @current_session.expiration_at
```
albertchae: I wasn't sure what the `.abs` meant even with the comment, so I removed it for now since this condition is a lot clearer IMO, but I can add it back if someone can explain what fast-forwarding `created_at` times enabled

db/migrate/20220425061358_add_expiration_at_to_user_sessions.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class AddExpirationAtToUserSessions < ActiveRecord::Migration[6.0]
+  def change
+    add_column :user_sessions, :expiration_at, :timestamp, null: true
+
+    reversible do |dir|
+      dir.up do
+        UserSession.only_deleted.update_all('expiration_at = deleted_at')
+        UserSession.update_all("expiration_at = created_at + INTERVAL '30 days'")
```
albertchae: It's not ideal to be doing so called "data migrations" inside a "schema migration", but this table seems small enough that it won't cause any downtime to do it here (unless we have a better way to do this already). I've heard of https://github.com/ilyakatz/data-migrate as an option but haven't really used it before.

Also not 100% sure the logic makes sense but it seemed reasonable to me, and much better than adding application logic to do something special if the column is null

albertchae: Hmm actually the safest thing to do might be to break up this PR to add the column and set the default value before adding the change to check [for expiration based on it](https://github.com/hackclub/bank/pull/2504/files/cf8f57ebfcb9ddddcb4bce6939147245bfb1107b#diff-22710d7b454bd5dff8e2d0c2adee848807d903bb98fbbe901c1429acd52a86dcR81), but not sure if Bank has the traffic to require all those steps. What's the practice been so far?

app/helpers/sessions_helper.rb
garyhtou: eh no solid reason. those logins are done via "magic link" authentication (links generated by a trusted 3rd party that authenticates users into Hack Club Bank). the idea behind making it half was that it's slightly more secure

app/helpers/sessions_helper.rb
garyhtou: yeah, let's remove it. i'm not exactly sure what it's for, but it was added in https://github.com/hackclub/bank/commit/f68279ffc2d603e5b342e5fa5f3101a541d51d96 as a part of https://github.com/hackclub/bank/pull/2049

AFAIK it doesn't have any real purpose. seems like it lets you set `created_at` far into the future to test expiration, which is kinda a pointless way of testing (since you should just set `created_at` back in time)

db/migrate/20220425061358_add_expiration_at_to_user_sessions.rb
garyhtou: [ilyakatz/data-migrate](https://github.com/ilyakatz/data-migrate) looks pretty cool! for data migrations not related to schema changes, we've just been running them via the console. that gem looks like a good way for us to keep track of those.

i think what u have is fine. we've been doing data migration inside schema migrations. @maxwofford thoughts?


logic looks alright, however, this might be better:
```ruby
UserSession.with_deleted.update_all("expiration_at = created_at + INTERVAL '30 days'")
# the expiration_at of user sessions that have already been deleted should technically still be `created at + 30 days`
# I think one point in time, session durations were only a few days, however, doing `created at + 30 days` is probably more accurate than using `deleted_at`
```
Also, I think https://github.com/ActsAsParanoid/acts_as_paranoid might reject updates to deleted objects. maybe this is fine since it `update_all` doesn't call callbacks. not sure

app/helpers/sessions_helper.rb
albertchae: Cool got it, also now we can set `expiration_at` too if we need to test

db/migrate/20220425061358_add_expiration_at_to_user_sessions.rb
albertchae: ok, updated to `UserSession.with_deleted.update_all("expiration_at = created_at + INTERVAL '30 days'")`

I think you're right that `update_all` will skip callbacks and thus bypass any `acts_as_paranoid` protection. At the very least, the migration succeeded on my local db and I could see that soft deleted `UserSession`s were updated.

garyhtou (approved):
looks good!

maxwofford (approved):
This looks great! Seems like everything has been covered by @garyhtou, but just want to check this worked with the prod database in development in under a matter of seconds– we have some endpoints that have uptime requirements (specifically, stripe-issued cards decline transactions if we take longer than 2 seconds to respond) and if this seems like a migration would lock us for 10+ seconds we'll need to find another solution.

albertchae:
@maxwofford I'll double check on a fresh database pull later. Also I can try to deploy this during off hours if that's worth doing?

maxwofford:
> Also I can try to deploy this during off hours

@albertchae I wouldn't sweat it if it's pretty quick on dev, but if there's any doubt go for it. We mostly get transactions noon to afternoon est & pst. Would love to hear how long the migration ends up taking for my own education!

albertchae:
@maxwofford I tried locally and rolled backward/forward few times and it was all consistently much less than a second

```
-- add_column(:user_sessions, :expiration_at, :timestamp, {:null=>true})
   -> 0.0024s
-- change_column_null(:user_sessions, :expiration_at, false)
   -> 0.0021s
0.044980875
```
That last number is just taking `Time.now` at the beginning and end of the migration and subtracting

albertchae:
I keep seeing builds fail with this message `Build timed out while waiting to start.` with nothing on the status page 🤔 https://status.heroku.com/

albertchae:
Ok now I finally see an incident https://status.heroku.com/incidents/2421

garyhtou:
herokus been having a rough few weeks lol 😵‍💫

albertchae:
@garyhtou thanks for following up! also good to see that the local migration time was indicative of speed on prod

PR #2508, opened 2022-04-25
Fix feedback button in header

cjdenio:
Currently:

Screen Shot 2022-04-25 at 1 52 44 PM

(Interpolation only works in double-quoted strings)


exu3 (approved):
Nice catch!

PR #2511, opened 2022-04-25
Redirect to desired page after login

cjdenio:
Screen Shot 2022-04-25 at 6 40 23 PM


maxwofford (approved):
This looks great! Go for merge if it works locally

PR #2514, opened 2022-04-26
Add flipper opt-in for user disbursements

maxwofford:
~~This PR is a design experiment, not currently planning to merge at time of writing:~~
EDIT: This was originally a design experiment, but I'm going forward with this b/c it's a great platform for testing #2558 and similar features that need to exist in production.

This adds a general `feature_enable` and `feature_disable` endpoint. As an example UI, it includes a section for the user edit page where users can choose from the list of feature previews, and opt-in to the ones they want.

Screen Shot 2022-04-26 at 09 28 27


maxwofford:
Going to merge this so I'm not blocking on #2558, but happy to make changes to the design here. @exu3, could I still get your thoughts on the frontend for this?

exu3:
> Going to merge this so I'm not blocking on #2558, but happy to make changes to the design here. @exu3, could I still get your thoughts on the frontend for this?

I think the design looks great for the Feature Previews section! 

It feels like there's a bit too much on the page though; maybe having `Feature Previews` a separate page / item in the sidebar? 
or perhaps, it'd be a good idea to abstract the user settings page into separate pages (eg. `Profile, Feature Previews, Security (login codes, security keys), Sessions`?)

maxwofford:
@exu3 That's a good point! How about we make a feature previews section that only has toggles? You could click through to a modal with all the description/details, but by default it'd be a small section.

exu3:
I think putting the details in a modal would be good idea.

PR #2520, opened 2022-04-29
Transparency for HCB Code show pages

garyhtou:
Unauthenticated users are now redirected to the ledger to view the transaction details


garyhtou:
closes https://github.com/hackclub/bank/issues/2302

maxwofford:
This looks great! I love the recording. One thing we'll want to do before this is merged is check for anything that relies on the HCB code hash ID being private– at the very least the receipts upload endpoint (sent in the transaction mailer) assumes hash ID is private. ~~Could you create a ticket along the lines of `HCB hashID shouldn't be used for anything sensitive` to track this?~~

UPDATE: Added it here! https://github.com/hackclub/bank/issues/2542

exu3:
This PR should be merged after #2543

maxwofford (changes requested):
Just tested locally. The visuals look great! Couple things:

- Worked for some HCB codes (0eH9NK), but others redirected to login (PVHxXp)
- I lost the TX b/c the highlight went away and I scrolled. Could we keep an orange border around it similar to how our direct-linked comments look?

cjdenio:
Little note: you can do the `.hash-highlight` thing with pure CSS: https://developer.mozilla.org/en-US/docs/Web/CSS/:target

maxwofford:
On a call, @garyhtou mentioned he was too busy with school to close this out atm. @cjdenio want to close this out together? Seems like the stuff left over is...
- Highlighting (switch to CSS solution, and keep highlight around for longer)
- Fix broken cases (PVHxXp) that still redirect to login

maxwofford:
Ok, just fixed the issue for hcb codes that didn't seem to work (they were all lacking CTs). For the CSS, want to just swap out the current logic @cjdenio?

maxwofford (approved):
I'm fine with this being merged now– https://github.com/hackclub/bank/pull/2520#issuecomment-1142157886.

app/assets/stylesheets/components/_transactions.scss
```diff
@@ -28,6 +28,11 @@
   }
 }
 
+:target {
```
cjdenio: Currently this also targets non-transactions

cjdenio (changes requested):

app/assets/stylesheets/components/_transactions.scss
maxwofford: @cjdenio if you have a fix for this please add it! I don't think anyone is directly working on the branch rn, so feel free to push directly here.

cjdenio (approved):

app/controllers/hcb_codes_controller.rb
```diff
@@ -1,14 +1,28 @@
...
+    hcb_id = @hcb_code.hashid
 
     authorize @hcb_code
+  rescue Pundit::NotAuthorizedError => e
+    raise unless @event.is_public?
 
-    @event = @hcb_code.event
+    if @hcb_code.canonical_transactions.any?
+      txs = TransactionGroupingEngine::Transaction::All.new(event_id: @event.id).run
+      pos = txs.index { |tx| tx.hcb_code == hcb } + 1
```
garyhtou: `index` should be `||` with `0` since it may return `nil`.

```ruby
pos = (txs.index { |tx| tx.hcb_code == hcb } || 0) + 1
```
I think `index` would return `nil` if it's a pending tx (pending transactions are always on the first page, but aren't in the array)

@cjdenio would you mind testing/making this change?

PR #2527, opened 2022-04-30
Fix preloading when no canonical_pending_transactions for an hcb_code

albertchae:
On a recent database dump, I saw some stray `CanonicalPendingTransaction` loads

```
  CanonicalPendingTransaction Load (1.4ms)  SELECT "canonical_pending_transactions".* FROM "canonical_pending_transactions" WHERE "canonical_pending_transactions"."hcb_code" = $1 ORDER BY "canonical_pending_transactions"."id" ASC LIMIT $2  [["hcb_code", "HCB-000-208981"], ["LIMIT", 1]]
    ↳ /usr/src/app/app/models/hcb_code.rb:213:in `pt'
```

It appears there is no such `CanonicalPendingTransaction` with that
`hcb_code`, so `canonical_pending_transactions_by_hcb_code` would be `nil`
and cause this line to run again https://github.com/hackclub/bank/blob/d8517d2ceea6f00df16547d10fad6b6644a845c6/app/models/hcb_code.rb#L68

To avoid this, set `canonical_pending_transactions` to `[]` instead in the preloader

Before: SQL Queries: 43 (4 cached) | Allocations: 475946
After: SQL Queries: 37 (1 cached) | Allocations: 458735


albertchae:
Does it make sense that there might be no `canonical_pending_transactions` for a given `hcb_code` or does that represent a data integrity problem?

maxwofford:
> Does it make sense that there might be no canonical_pending_transactions for a given hcb_code or does that represent a data integrity problem?

At the time of writing, it makes sense. Examples:

- Disbursements don't have pending transactions, so a disbursement's HCB won't ever have a `CanonicalPendingTransaction`.
- The pending transaction engine only creates `RawPendingXTransaction` immediately, then runs a nightly job (ever 5 min) to create `CanonicalPendingTransaction` records: https://github.com/hackclub/bank/blob/main/app/jobs/pending_transaction_engine_job/nightly.rb

app/services/transaction_grouping_engine/transaction/association_preloader.rb
```diff
@@ -29,7 +29,7 @@ def preload_associations!
         hcb_code_objects.each do |hc|
           hc.canonical_transactions = canonical_transactions_by_hcb_code[hc.hcb_code]
                                       .sort { |ct1, ct2| self.class.compare_date_id_descending(ct1, ct2) }
-          hc.canonical_pending_transactions = canonical_pending_transactions_by_hcb_code[hc.hcb_code]
+          hc.canonical_pending_transactions = canonical_pending_transactions_by_hcb_code[hc.hcb_code] || []
```
maxwofford: I haven't looked into the context where we're using this, but do we want to use `CanonicalPendingTransaction.none` here to keep it as an activerecord association?

maxwofford (approved):
Added a comment, but don't want that to block the fix you're making here– feel free to merge!

garyhtou (approved):
for additional context: i'd say only abt 2/3 of our hcb codes have pending txs

app/services/transaction_grouping_engine/transaction/association_preloader.rb
albertchae: Oh cool, I wasn't aware of `.none` as a way to represent empty results in AREL! However in this case I believe `canonical_pending_transactions_by_hcb_code` has already been taken from AREL to activerecord objects after running the `group_by` on line 27, so I think using an array is the right type

PR #2528, opened 2022-05-01
Autofill Organization name on hcb-stickers form

exu3:
This is helpful context for hand-written messages included in sticker envelopes. (Currently I look up the user by email from the Users admin page for people I don't know by name which is quite tedious.)


maxwofford (approved):
Does this handle users without events?

exu3:
It does handle users with no events.
`current_user.events.first&.name`

PR #2535, opened 2022-05-05
[WIP] pagy spike

albertchae:
For https://github.com/hackclub/bank/issues/2446

https://ddnexus.github.io/pagy/migration-guide

https://ddnexus.github.io/pagy/extras/array


albertchae:
Hey @garyhtou I tried out `pagy` and anecdotally I don't see much of a difference (rough average of ~500ms a request on /hq locally with or without `pagy`). I pushed up this WIP branch in case you want to try it out. I do like some of the design principles of `pagy` (https://ddnexus.github.io/pagy/index#gsc.tab=0) so it might be worth using for API and those reasons alone vs just performance.

albertchae:
Going to close this for now since it doesn't seem to achieve performance gains, although as I said before we can consider it later if we prefer the API

PR #2537, opened 2022-05-05
Fix disbursement from/to

cjdenio:
Fixes #2536!


maxwofford (approved):
Thanks for the fix, LGTM!

PR #2543, opened 2022-05-06
Sign Receipt Upload endpoint

garyhtou:
Closes https://github.com/hackclub/bank/issues/2542 and #2559

WIP. TODO:
- only require valid signature for unauthenticated
- update mailer to use new URL with signature
- 10 day grace period for existing urls


garyhtou:
Now that we're signing these urls, we can also associate the uploaded receipt (when the user's not authenticated) with the user who received the email.

For example, you click on the link (not signed in) and upload the receipt. Instead of the receipt showing as uploaded by `anonymous`, it would be you. This would include some weird edge cases when sharing the link with someone else (such as by forwarding the email). It would show that the receipt was uploaded by you if someone else uses your link.

garyhtou:
@maxwofford, can we get another set of eyes on this to double check that its runs properly? (we have to resolve some credential conflicts)

app/policies/hcb_code_policy.rb
```diff
@@ -14,7 +14,7 @@ def receipt?
   end
 
   def attach_receipt?
-    user&.admin? || present_in_events? || record.date > 10.days.ago
+    user&.admin? || present_in_events? || record.date > Date.parse('2022-05-12') - 10.days
```
garyhtou: Update grace period date before merging

app/views/canonical_pending_transaction_mailer/notify_approved.html.erb
```diff
@@ -1,7 +1,9 @@
+<% upload_url = HcbCodeService::Receipt::SigningEndpoint.new.create(@cpt.local_hcb_code) %>
```
maxwofford: Could we drop this into a view helper or class method? I don't see many examples of calling services directly in our view layer.

app/controllers/hcb_codes_controller.rb
```diff
@@ -61,6 +61,11 @@ def attach_receipt
     @event = @hcb_code.event
 
     authorize @hcb_code
+
+  rescue Pundit::NotAuthorizedError
```
maxwofford: Great that we attempt to handle this, love it!

maxwofford (approved):
Looks great! Left a couple comments, but if this is functional I'm fine with it being merged.

maxwofford (commented):
Hang tight– just realized I hadn't fetched. Looks like there's some rebasing issues with the credentials when pulling in main. Could you merge in the latest from `main` & lmk once it's ready to be tested?

app/views/canonical_pending_transaction_mailer/notify_approved.html.erb
garyhtou: Would it make more sense to pass it via the mailer call?

maxwofford: That'd work too!

exu3:
It seems to be up to date with `main` & works locally for me. It should be ready to be tested.

maxwofford:
Testing locally works for me– played with the dates too and everything seems to be working. LGTM!

PR #2544, opened 2022-05-06
Deploy to Heroku with GH Actions

exu3:


exu3:
~~Actually I think my Heroku API key just changed w/ the password reset; I should fix that before merging.~~

It should be good now!

I also tested this workflow on a separate repo, and it seemed to work.

PR #2551, opened 2022-05-07
Update Plaid tokens

maxwofford:
This reverts commit 8d7d907.

⚠️ This should only be merged once https://github.com/hackclub/bank/pull/2550 is merged, deployed, and confirmed working with the new Plaid tokens.

It should also have the new env tokens added to the Rails encrypted credentials file.


maxwofford:
This deployed successfully & background jobs that rely on Plaid (ie `TransactionEngineJob::Nightly` are completing successfully. I'm now deleting the old API tokens.

PR #2556, opened 2022-05-09
Temporarily Disable Sticker Mule

exu3:
We don't have enough credits currently and are currently working on renewing the contract.

Screen Shot 2022-05-09 at 10 24 20


maxwofford (approved):
Thanks for moving on this. Go ahead and merge when you're ready!

PR #2558, opened 2022-05-09
Spike on receipt uploading via email reply

maxwofford:
This spikes out #2557.


maxwofford:
While this is being developed, you can get a transaction's receipt upload URL in the admin tool on the HCB code show page:

Screen Shot 2022-05-09 at 16 40 25

maxwofford:
Added a feature preview section for emails too!

Screen Shot 2022-05-10 at 09 41 59
Screen Shot 2022-05-10 at 09 42 07

maxwofford:
I've just gated the feature preview to admins until we can test the action mailbox stuff in production:
Screen Shot 2022-05-10 at 12 55 54

If this code does what I think it does, regular users won't see anything different in their emails or bank UI.

kunalbotla:
this looks amazing! i really like the feature preview page

maxwofford:
Merging this to make sure everything is wired up in prod (ie. DNS & API keys). New feature may be non-functional for the first 10 minutes while I do that, but no downtime on other stuff is expected.

app/services/receipt_service/create.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module ReceiptService
```
garyhtou: @maxwofford, I believe this might not handle the edge case where a receipt is uploaded for a receiptable that was previously `marked_no_or_lost_receipt`.

FYI, similar functionality (for HCB Codes) exists here: https://github.com/hackclub/bank/blob/b2f6c940283add4c46cfad9b76433207991644c3/app/services/hcb_code_service/receipt/create.rb

PR #2566, opened 2022-05-10
Regular users can enable feature flag for receipt uploads

maxwofford:
Part of #2557. This should probably only be merged in after https://github.com/hackclub/bank/pull/2567, which adds permissions checking for receipt uploads.


maxwofford:
The requested review here is less for the line changes in this PR, more for a sanity check before releasing this on the public. It doesn't need to be good yet, it just needs to work well enough for it to be a feature preview. I'll merge once there are review approvals.

cjdenio (approved):
This looks great to me!

exu3 (approved):
Looks good to me! Can confirm the feature works well.

garyhtou (approved):
good to go!

app/views/users/edit.html.erb
```diff
@@ -98,37 +98,32 @@
...
           </div>
-        </div>
-        <p>
-          We send an email once a card is charged along the lines of "Please
-          upload a receipt for your $10 charge" with a link to a form on HCB
-          to upload receipts.
-          After opting-in to this feature preview, you can reply directly to
-          that email with your receipt attached to upload it.
-        </p>
+          <%# link_to "Shipped!", enable_feature_path(feature: :transfers_2022_04_21), method: :post, class: 'btn bg-accent' %>
```
garyhtou: ~~should this be disabled? or just be text?~~

nvm! i see its commented out

maxwofford: This is commented out for the time being while getting some styling correct. It's kinda weird, but https://github.com/hackclub/bank/pull/2566/files#diff-7a664d6e0b1d39425223431cffbf717843cbac2e33ac4fe6a58ecb959e38f6abR104 is a check to see if the feature is generally available, and doesn't check for the current user.

PR #2567, opened 2022-05-10
Add invalid permissions for email receipt upload

maxwofford:
Part of #2557. This updates the logic to respond with a simple error in-case the transaction can't be found, or the user doesn't have permission to upload to that transaction:
Screen Shot 2022-05-10 at 19 15 06


app/mailboxes/receipt_uploads_mailbox.rb
```diff
@@ -5,9 +5,11 @@ class ReceiptUploadsMailbox < ApplicationMailbox
   # inbound_email => ActionMailbox::InboundEmail record  --> the active storage record
 
   def process
-    ensure_user
-    ensure_hcb
-    ensure_attachment
+    return unless ensure_user?
+    return unless ensure_hcb?
+    return unless ensure_attachment?
+    return unless ensure_permissions?
```
garyhtou: This totally works, but it might be better to do:
```ruby
before_processing :ensure_user # if it fails, it'll automatically stop processing
```


personally, it feels kinda wonky to have a boolean/question mark method with side effects. not sure if that's commonly agreed upon best practice or not

PR #2568, opened 2022-05-11
Hide declined transactions from /my/inbox list

maxwofford:
Fixes for https://github.com/hackclub/bank/issues/2460 / https://github.com/hackclub/bank/issues/2533

I'd love to get your eyes on this @garyhtou. This code accomplishes the intent of my UI changes, but I want a sanity check on how I'm writing it:
Screen Shot 2022-05-11 at 08 59 39
Screen Shot 2022-05-11 at 09 01 49

Recommend reaching each commit separately– they're each an individual change.


app/models/hcb_code.rb
```diff
@@ -237,6 +237,25 @@ def smartish_custom_memo
...
 
+  def receipt_required?
+    case type
+    when :card_charge
+      if pt.declined?
+        false
+      else
+        true
+      end
+    when :unknown
```
cjdenio: Ruby's `case` conditions don't fall through (like they do in other languages), so maybe do something like this?

```rb
if type == :card_charge && !pt.declined?
  true
else
  false
end
```

cjdenio (changes requested):

app/views/static_pages/my_inbox.html.erb
```diff
@@ -33,6 +33,7 @@
               <% end %>
             <% else %>
               <% hcb_code.canonical_pending_transactions.each do |pt| %>
+                <% next if pt.declined? %>
```
cjdenio: Is this necessary if transactions are already filtered in the controller?

maxwofford: Yep! The controller filters for missing receipts, but we only have `declined?` context on the actual PT or CT. That said, that should probably be dropped into the controller– will do that now!

maxwofford: Just pushed code that should be clearer. Also replaced the `declined?` usage with `receipt_required?` provided by the Receiptable concern.

cjdenio (approved):

PR #2573, opened 2022-05-12
Track receipt upload method

cjdenio:
This PR adds an `upload_method` field to `Receipt` that tracks the way that certain receipts were uploaded, for example, `email` or `receipts_page_drag_and_drop`. I discussed this briefly with @maxwofford earlier.


maxwofford (approved):
Looks great! To keep things a bit more "railsy" I'd recommend using an `enum` here– helps to future-proof stuff if we add/change upload methods in the future. We use one already here: https://github.com/hackclub/bank/pull/2300/files.

That said, if this functions feel free to merge.

cjdenio:
Good call!

PR #2576, opened 2022-05-15
Fix shadow bug on event cards

exu3:
This PR fixes the shadow bug by using `max-width` instead of `overflow-hidden`.

https://hackclub.slack.com/archives/CN523HLKW/p1652628705978339

Screen Shot 2022-05-15 at 10 28 40


cjdenio (approved):
Thanks for fixing this!

PR #2578, opened 2022-05-16
Restrict attach_receipt page to images and PDFs

cjdenio:
### TL;DR: Mobile Firefox won't open the camera unless `image/*` is explicitly allowed

About 99.5% of receipts are either images or PDFs, so I think it's reasonable to enforce this requirement just on the `attach_receipt` page.


maxwofford (approved):
Good to know! This is good to merge if it works locally on your devices. After deploy please reply to Zach's slack message to confirm this fixes it for him.

PR #2579, opened 2022-05-18
Add missing receipt count badge to nav

cjdenio:
Painstakingly tested across light/dark mode, and desktop/mobile.

Screen Shot 2022-05-17 at 11 42 09 PM
Screen Shot 2022-05-17 at 11 42 28 PM


maxwofford (commented):
Love this addition!

This may have to be updated with similar transaction logic to https://github.com/hackclub/bank/pull/2568– the implementation I'm reading in this PR doesn't filter out declined transactions.

cjdenio:
Yeah, good point– I'll wait 'til #2568 is merged

maxwofford (approved):
Great! This looks like the count should work, but I'm a little worried this query is going to slow down loading for some users. Try it on `max@hackclub` as an example case. If it's slow, want to try refactoring it? If it's fine, go for merge!

maxwofford:
Hmmmm, looking back at this, maybe the fix is to make this async? In a perfect world it loads instantly, but in a less-than perfect world the pages still load quickly even if the notification pops up after a little bit.

We could make a missing receiptable service that handles these types of things, then make a new endpoint that serves it's data.

PR #2580, opened 2022-05-18
Comment out secret key base in credentials.yml.enc because heroku will override

maxwofford:
I believe this will disable active user sessions, so before merge we should consider public messaging to warn users


maxwofford:
@albertchae Have you ever rotated the secret key base for an app? Am I missing anything here?

albertchae:
@maxwofford I haven't really used Rails encrypted credentials (in most of my past projects we deployed different environment variables based on dev/staging/prod) so I'm not sure. This will require distributing a new `config/master.key` right? Also I believe there may be some downtime since you'll have to change that on heroku also (unless you're able to tie the deployment of this PR exactly with changing that heroku variable), and probably all user sessions will be expired because `cookies.encrypted[:session_token]` will be different after, so we should communicate that to bank users.

maxwofford:
This is just the secret key base, not the master.key– you should be able to see the content change by running `rails credentials:edit` on this branch, then another branch for comparison. Prod gets it's secret key base from the master.key file so I don't expect downtime. That's a good point about communicating the signout to users. Wish there was a way to have new users on the new secret key base, but let old sessions use the old one for a month until their sessions expire.

albertchae:
@maxwofford ah oops, I see what you mean by you're using the same master.key but you changed the value inside credentials. Then yeah the only concern is communicating that we're logging out everyone's session

> Wish there was a way to have new users on the new secret key base, but let old sessions use the old one for a month until their sessions expire.

💯 I've wished that for a number of situations where an old secret and new secret can both work for a short period of time. I think the only place I've seen this so far is stripe

albertchae:
@maxwofford I found this article which talks about a way to rotate the key but allow old sessions for a temporary time

https://blog.kmkonline.co.id/rotating-ruby-on-rails-secret-key-base-9a4dcdf0d817

official docs https://guides.rubyonrails.org/security.html#rotating-encrypted-and-signed-cookies-configurations

maxwofford:
I was playing around with the secret key config & getting confusing results. I learned that 1.) we aren't using the secret key found in our encrypted credentials, there's a SECRET_KEY_BASE env token set on Heroku which takes precedence and 2.) SECRET_KEY_BASE is auto-generated in development/testing environments.

albertchae:
@maxwofford that's very interesting. Should we remove one or the other in production so we're not confused about which one is being used? Also if we go with the heroku env variable route, we should still be able to try one of those rotation methods right?

albertchae:
Follow up from June 16 call with @maxwofford , I'll take over removing the heroku environment variable and then timebox investigation into rotating the key. If the investigation exceeds the timebox we'll just expire everyone's session and rotate

albertchae:
OK I did some investigation here
- To corroborate Max's findings in https://github.com/hackclub/bank/pull/2580#issuecomment-1156360631 , here are the official Rails docs https://api.rubyonrails.org/classes/Rails/Application.html#method-i-secret_key_base
```
In development and test, this is randomly generated and stored in a temporary file in tmp/development_secret.txt.

In all other environments, we look for it first in ENV["SECRET_KEY_BASE"], then credentials.secret_key_base, and finally secrets.secret_key_base. For most applications, the correct place to store it is in the encrypted credentials file.
```
- So based on this I wanted to test the rotation locally. Since local environments use `tmp/development_secret.txt`, I tried running locally with `RAILS_ENV=production` which triggers https://github.com/hackclub/bank/blob/0ad7ccfe6c8bc1148afbd1682821011cfcccfd52/config/environments/production.rb#L55-L56 so I am unable to access https://localhost:3000 without doing some more SSL finagling. I then tried to find if there was a way to change this secret behavior on locally temporarily, but it's hardcoded to a Rails.env check in the framework https://github.com/rails/rails/blob/6-1-stable/railties/lib/rails/application.rb#L433-L441

- More importantly, I found several posts detailing that Heroku sets `SECRET_KEY_BASE` automatically even if it's not explicitly set in the config vars (and it'll change every deploy if it's not set to a value in config vars). Although it seems possible to force a heroku deployment to look at the credentials value instead by explicitly setting the environment variable `SECRET_KEY_BASE` to `nil`, I feel like we shouldn't fight against the platform here.
  - https://github.com/heroku/heroku-buildpack-ruby/issues/1143
  - https://github.com/anycable/anycable-rails/issues/135
  - https://canolcer.com/post/rails-prevent-users-from-logging-out/
  - https://old.reddit.com/r/rails/comments/mni259/rails_613_on_heroku_wrong_secret_key_base/
  
So in conclusion, my recommendation is that we actually delete `secret_key_base` from credentials because it's unused and can be misleading, don't bother with adding the rotation logic because it's hard to test locally, and rotate `SECRET_KEY_BASE` by changing the heroku config variable. Besides notifying users they will be logged out, it might be good to set some maintenance period and also warn users not to have any forms in progress during that time because the submit will likely fail and they could lose in progress work.

albertchae (approved):
@maxwofford I can't ask you to approve it since you're still technically the author so I'm approving it here, but if you think we can go ahead with this I'll merge it. I changed the title of the PR to accurately reflect the change that's happening here now.

maxwofford:
Just check this out locally to ensure we aren't overwriting anything from other recent PRs. Looks good!

maxwofford:
Follow up task to this: actually changing our secret key base in Heroku & posting in the Slack that users will have to re-auth on next load.

PR #2584, opened 2022-05-21
Fix floating point error in invoice create service

maxwofford:
Fixes ﻿[#2583](https://github.com/hackclub/bank/issues/2583). Previously invoices for $144.20 were created with $144.19. This was a floating point error in parsing the [user's input here](https://github.com/hackclub/bank/blob/227806bab0f18bd157dd1eb42e9f0fc2b7841f5b/app/services/invoice_service/create.rb#L137) (`("144.20".gsub(",", "").to_f * 100.to_i).to_i` returns `14419`).

This isn't the first time we've had floating point issues with handling amounts from users and APIs, so I've gone ahead and added RubyMoney/monetize for parsing values.


app/services/invoice_service/create.rb
```diff
@@ -111,7 +111,7 @@ def sponsor_attrs
     end
 
     def clean_item_amount
-      @clean_item_amount ||= cleanse(@item_amount)
+      @clean_item_amount ||= (Monetize.parse(@item_amount) * 100).to_i
```
cjdenio: I think it's better to use `Monetize.parse(@item_amount).cents`. LGTM otherwise!

cjdenio (changes requested):

cjdenio (approved):

PR #2585, opened 2022-05-22
Clarify errors on receipt upload mailer

maxwofford:


cjdenio (approved):
LGTM! We should add a dino to the email sometime in the future

PR #2589, opened 2022-05-22
Only load Headway widget js if signed in

exu3:
Screen Shot 2022-05-21 at 22 23 01

Currently, this message shows up in the console in transparency mode. This PR makes it so that widget.js is only loaded if the widget is visible/when user signed in.


maxwofford:
Nice catch! Thanks for adding that. This code looks good to merge.

PR #2590, opened 2022-05-23
Change shebang to use /usr/bin/env for NixOS compatibility

zachlatta:
Using `/bin/bash` in scripts doesn't work on NixOS because that's not where the bash executable is. Instead, use the OS to look up the bash binary location by using `/usr/bin/env bash` as the shebang.


maxwofford:
This worked locally for me when testing in a codespace, so I'm going ahead with merge!

PR #2593, opened 2022-05-24
Upload from file system on iOS

exu3:
Seems to fix #2591; although, I'm not sure if this change would cause problems elsewhere (eg. on Android browsers?)
IMG_8576B02D4200-1


cjdenio (approved):
It works on Android if you add `accept: "image/*,.pdf"`. Thank you so much for fixing this!

PR #2597, opened 2022-05-24
Add mailer previews

maxwofford:
An index is available in development at http://localhost:3000/rails/mailers:

Screen Shot 2022-05-24 at 12 33 47


exu3 (approved):
Awesome! LGTM.

PR #2598, opened 2022-05-24
Create example missing receiptable service

maxwofford:
This is an example to demonstrate https://github.com/hackclub/bank/pull/2579#issuecomment-1136326776


app/services/receiptable_service/missing.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+# Get all missing receipts + their cards for a given user:
+# ReceiptableService::Missing.new(user: current_user).run
+# Get last 5 missing receipts for a user:
+# ReceiptableService::Missing.new(user: current_user, limit: 5).run
```
albertchae: Is "last" 5 true? It seems like the query below returns `hcb_code`s with missing receipts for `stripe_card`s/`emburse_card`s, so I'm imagining it'll just be in first 5 that come up in the order of the stripe card association and then the emburse card association on the user?

app/services/receiptable_service/missing.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def run
+      results = {}
+      count = 0
+
+      stripe_cards = @user.stripe_cards
+      emburse_cards = @user.emburse_cards
+      (stripe_cards + emburse_cards).each do |card|
+        card.hcb_codes.missing_receipt.each do |hcb_code|
```
albertchae: This ends up being an N+1 on `hcb_code`s per card (so it's fine when the user only has one card, but not when there are multiple). I poked around a little but it seems like it should be able to get all the relevant `hcb_code`s for all the cards in one go (i.e. combine https://github.com/hackclub/bank/blob/180671544b484a16f384f075865e780003e51445/app/models/stripe_card.rb#L232-L234 and https://github.com/hackclub/bank/blob/180671544b484a16f384f075865e780003e51445/app/models/emburse_card.rb#L146-L156 for all the cards somehow). Unfortunately because `hcb_code` isn't a proper ActiveRecord association we can't just add a bunch of `.includes` to solve it and probably need to do something similar to the preloader added in https://github.com/hackclub/bank/issues/2446

once we have some variable of `all_stripe_and_emburse_card_hcb_codes`, then this can be collapsed into `HcbCode.where(hcb_code: all_stripe_and_emburse_card_hcb_codes).missing_receipt`. It might even be possible to fold in the `limit` and `receipt_required?`, although because `receipt_required?` uses `pt` which isn't an association (https://github.com/hackclub/bank/blob/180671544b484a16f384f075865e780003e51445/app/models/hcb_code.rb#L225-L227) this might be more tricky than it's worth. I think loading all the hcb_codes into memory and then filtering `receipt_required?` and taking a limit in memory should be fine.

app/services/receiptable_service/missing.rb
```diff
@@ -0,0 +1,36 @@
...
+          next unless hcb_code.receipt_required?
+          next unless @limit.nil? || count < @limit
+
+          results[card] ||= []
```
albertchae: if you initialize results as `results = Hash.new { |h, k| h[k] = [] }` so that it sets the default value to a new array each time, you can avoid this line. Debatable if that's any clearer than this line though 😉 

Although just want to point out that it's very important not to do `results = Hash.new([])`, because that sets every default value to the SAME array in memory.

app/services/receiptable_service/missing.rb
```diff
@@ -0,0 +1,36 @@
...
+    def run
+      results = {}
+      count = 0
+
+      stripe_cards = @user.stripe_cards
+      emburse_cards = @user.emburse_cards
+      (stripe_cards + emburse_cards).each do |card|
+        card.hcb_codes.missing_receipt.each do |hcb_code|
+          next unless hcb_code.receipt_required?
+          next unless @limit.nil? || count < @limit
```
albertchae: A potential alternative here is to filter the array of `HcbCode`s with `missing_receipt`s and then take the limit. So something like

`card.hcb_codes.missing_receipt.filter(&:receipt_required?).take(@limit)`

Although this would also require setting `@limit` to max int or the size of the initial array when it's `nil`. The other negative of this is that it'll loop over all the hcb_codes instead of stopping once we hit limit, so another potential way around that is to do a lazy enumeration https://ruby-doc.org/core-2.7.0/Enumerator/Lazy.html

This would look something like `card.hcb_codes.missing_receipt.lazy.filter(&:receipt_required?).take(@limit)`, which would only call need to iterate over the `@limit` number of `hcb_code`s before stopping

I think what you have already is fine though. Just showing some alternatives

app/services/receiptable_service/missing.rb
```diff
@@ -0,0 +1,36 @@
...
+
+          results[card] ||= []
+          results[card] << hcb_code
+
+          count += 1
+        end
+      end
+
+      results
```
albertchae: It might be a little clearer if we return a custom object that represents this data instead of the raw hash. IMO these 4 lines from `app/controllers/static_pages_controller.rb` all illustrate that the clients of this service know too much about the structure of this returned value

```
    @count = @missing_receipts.values.map(&:size).sum

    @missing_receipt_ids = missing_receipts.values.flatten.map(&:id)

    @cards = @missing_receipts.keys
    @count = @missing_receipts.values.map(&:size).sum
```

It's not obvious that the hash is keys of `StripeCard` objects mapping to arrays of `HcbCode`s

albertchae (commented):
@maxwofford sorry for taking so long to get around to this PR. I think the general approach is good and I think this is a great example of encapsulating all this logic in a service. I left some subjective comments on how to avoid N+1s plus some other other thoughts you might want to explore.

maxwofford:
Closing this b/c it's an example & the functionality this was to support is now in #2981

PR #2603, opened 2022-05-25
Fix deletion requests page

cjdenio:
The issue:

Screen Shot 2022-05-25 at 6 05 06 PM

See [this](https://github.com/ActsAsParanoid/acts_as_paranoid#associations) for an explanation.


cjdenio:
👀

maxwofford (approved):
Thanks for catching that! Once you test it locally go for merge. Also, do we use the lambda anywhere else?

cjdenio:
Just checked and this isn't used anywhere else.

PR #2606, opened 2022-05-26
Better Stripe decline emails

cjdenio:
Adds the following cases:

**Insufficient funds**

Screen Shot 2022-05-25 at 9 16 56 PM

**Suspected fraud (incorrect CVC)**

Screen Shot 2022-05-25 at 9 16 53 PM

**Card frozen**

Screen Shot 2022-05-25 at 9 16 49 PM

Closes #2529.


maxwofford:
This is so great! Thanks for making it.

Couple things:
- Does this break if other [reasons](https://stripe.com/docs/api/issuing/authorizations/object#issuing_authorization_object-request_history-reason), ie. `cardholder_verification_required`, come through?
- Does this handle the rejection history object not being the most recent event?
- Is there a way we can run this in production in parallel to see if it breaks for anyone over a couple days?

cjdenio:
Approved by @maxwofford in Slack.

PR #2607, opened 2022-05-26
Pass arguments through to docker_start.sh

maxwofford:
Now you can do this: `docker_start.sh /bin/bash` or `docker_start.sh /bin/bash --with-solargraph`


maxwofford:
Merging b/c this works locally & it's a dev tool so nothing will break in prod.

PR #2614, opened 2022-05-31
Change event_id on organizer_position_invites to NOT NULL

albertchae:
This PR is steps 1 and 2 of https://github.com/hackclub/bank/issues/2415
- Delete `OrganizerPositionInvite`s with null `event_id`
- Make `event_id` NOT NULL on `OrganizerPositionInvite`

I investigated a bit how these could have been orphaned. AFAICT, the
`dependent:` option on the event has_many to organizer_position_invites
was never set to `nullify` according to git, so I tried to see if it was
possible to set a null event

- https://github.com/hackclub/bank/blob/ac8ce08818c16b0df3d74918699abc0a54997096/app/controllers/organizer_position_invites_controller.rb#L9-L20 here the `Event.friendly.find` would throw an exception if the event wasn't found
- https://github.com/hackclub/bank/blob/ac8ce08818c16b0df3d74918699abc0a54997096/app/models/partner.rb#L39-L44 `add_user_to_partnered_event!` doesn't seem to be called anywhere
- https://github.com/hackclub/bank/blob/ac8ce08818c16b0df3d74918699abc0a54997096/app/services/event_service/partner_invite_user.rb#L19-L23 is called in 2 spots
  - https://github.com/hackclub/bank/blob/ac8ce08818c16b0df3d74918699abc0a54997096/app/controllers/admin_controller.rb#L138-L150 creates an event just before, so this shouldn't be a culprit
  - https://github.com/hackclub/bank/blob/ac8ce08818c16b0df3d74918699abc0a54997096/app/controllers/api/v2_controller.rb#L65-L72 it's possible `current_partner.events.find_by_public_id` can return nil if passed an id for an event that doesn't fall under the current partner

I'm still not convinced it's the last callsite, but it's still good to
"make impossible states impossible", so adding this NOT NULL constraint
should help.


db/migrate/20220531000103_change_event_on_organizer_position_invite_to_not_null.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class ChangeEventOnOrganizerPositionInviteToNotNull < ActiveRecord::Migration[6.1]
+  def change
+    OrganizerPositionInvite.where(event_id: nil).each do |opi|
+      opi.destroy!
+    end
```
albertchae: There's only 8 of these in a recent data dump so deleting one by one should be OK

maxwofford (approved):
LGTM!

albertchae:
@maxwofford btw is https://github.com/hackclub/bank/blob/ac8ce08818c16b0df3d74918699abc0a54997096/app/models/partner.rb#L22 used only for admin use in the console or not used at all? I couldn't see where it was called and would like to delete it if it's unused

maxwofford:
That's not used currently, but is part of a larger set of things that may need to be removed in the future (all the partnered API stuff). Forgot to mention, but could you future-proof that migration in case we change the model names? ex. https://github.com/hackclub/bank/blob/main/db/migrate/20220114190026_add_has_untracked_edit_to_comments.rb#L8-L9

albertchae:
@maxwofford added the future proofing. This is a pretty interesting way to tackle that problem and the first time I've seen it! What I've seen in other rails codebases is to collapse old migrations (say from 1 or 2 calendar years ago) into the schema and then move them into an archived migrations folder so they are there for reference but don't run on any db:migrate

> That's not used currently, but is part of a larger set of things that may need to be removed in the future (all the partnered API stuff)

In that case I'll hold off on deleting it for now

PR #2616, opened 2022-06-01
Set cookie, csrf, and feature policy 6.1 defaults

maxwofford:
Last changes that are part of https://github.com/hackclub/bank/pull/2588. Waited on these a week b/c it's hard to roll these changes back and the Rails migration guide recommends only setting these after ensuring 6.1 is running and stable in production.

Things this changes:
- [sessions cookies are now SameSite](https://github.com/rails/rails/issues/35822), which can be tested by logging into the stage deploy and ensuring session logic, sign in/out, still work
- [updates csrf generation](https://guides.rubyonrails.org/configuring.html#config-action-controller-urlsafe-csrf-tokens), which can be tested similarly. need to also check if this will invalidate/logout existing sessions.
- [feature policy headers are now set](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy/), which needs to be tested on an individual-feature basis. some shouldn't matter (ie. microphone), but ones that come to mind:
	- [x] camera: does this affect our receipt upload buttons that take a file from the camera?
	- [x] fullscreen: does this affect someone trying to fullscreen bank? what about fullscreening our embeddable donation pages?
	- [x] payment: does this affect our embeddable donation pages?
	- [x] geolocation: does this affect our analytics?
	- [x] usb: does this affect hardware signing keys (I don't have a yubikey, @cjdenio could you test this out or find someone who can test it on staging?)
	- [x] gyroscope: i don't think this affects anything
	- [x] microphone: i don't think this affects anything


cjdenio:
Can confirm that Yubikeys still work (covered by [this policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy/publickey-credentials-get) for future reference)

maxwofford:
File upload still works! Seems like camera permission is just for JS requests to system media.
IMG_7596

maxwofford:
Looks like the `payment` permission controls access to the [JS `PaymentRequest` class](https://w3c.github.io/payment-request/#dom-paymentrequest).

Stripe's docs [mention they wrap `PaymentRequest` in non-safari browsers](https://stripe.com/docs/js/payment_request/create), so I'm going to look into opening up payment permissions on those specific endpoints we use Stripe's library.

maxwofford:
I poked around with `geolocation` & AFAICT Ahoy just relies on geocoding via IP, so it shouldn't be affected

maxwofford:
I don't think anyone is fullscreening our embeded donation pages, but I granted permissions just in case.

maxwofford:
I didn't have any signin/signout issues, but assigning @cjdenio for a sanity check on what I'm doing. Planning to merge after greenlight

app/controllers/donations_controller.rb
```diff
@@ -10,6 +10,13 @@ class DonationsController < ApplicationController
   before_action :set_event, only: [:start_donation, :make_donation, :finish_donation, :qr_code]
   before_action :allow_iframe, except: [:show, :index]
 
+  permissions_policy do |p|
+    # Allow stripe.js to wrap PaymentRequest in non-safari browsers.
+    p.payment    :self
+    # Allow embeded donation pages to be fullscreened
```
cjdenio:
```suggestion
    # Allow embedded donation pages to be fullscreened
```

cjdenio (approved):
Looks great to me!

PR #2621, opened 2022-06-02
Disbursements create pending transactions

maxwofford:
The first change required for https://github.com/hackclub/bank/issues/2521

A couple notes & design decisions:

- `PendingTransactionEngine::RawPendingDisbursementTransactionService::Disbursement::ImportSingle` also canonizes and eventmaps the transactions it creates. This isn't consistent with the other services I saw, but doing that in `DisbursementService::Create` would involve having extra search logic to pull up the newly created records. Is there a better way to do this?
- `PendingTransactionEngine::RawPendingDisbursementTransactionService::Disbursement::ImportSingle` imports a single disbursement, but creates multiple `RawPendingDisbursementTransaction`s. I'm thinking this should be refactored, but I'm a little lost on the right approach to take here.
- This creates `RawPendingDisbursementTransaction`, but now that I've gone through it, I'm wondering if that should be broken into separate models. This is already a 300+ line change to add something we already have into the pending TX engine, so I'm hesitant to increase complexity that much.
	- `RawPendingOutgoingDisbursementTransaction`
	- `RawPendingIncomingDisbursementTransaction`
- `PendingTransactionEngine::CanonicalPendingTransactionService::ImportSingle::Disbursement` has a flipped name from the `PendingTransactionEngine::RawPendingDisbursementTransactionService` sibling– didn't refactor b/c it follows the convention of the other services, and that's out of the scope of this PR
- I don't have a good way to check for `Disbursement.pending`. For now I'm just checking `Disbursement`s where `fullfilled_at` is not null, but that feels a bit hacky. Good opportunity to refactor disbursements over to aasm state once this PR is done.

Would love to get your eyes on this @garyhtou!


maxwofford:
Closing in favor of https://github.com/hackclub/bank/pull/2642

PR #2623, opened 2022-06-02
Order cards from `/my/cards`

cjdenio:
This PR adds the ability to order cards directly from `/my/cards`.

https://user-images.githubusercontent.com/34525547/171735181-b7dd5ed2-48e3-4ed6-85ab-32118b1fff28.mov


app/views/stripe_cards/_form.html.erb
```diff
@@ -1,4 +1,12 @@
 <%= form_with(model: StripeCard.new, local: true) do |form| %>
+  <% if @event || params[:event_id] %>
```
maxwofford: I like this approach of keeping it all in the partial. Just to sanity check, this still works on the event cards page?

app/views/static_pages/my_cards.html.erb
```diff
@@ -2,7 +2,22 @@
 <% page_md %>
 <%= render 'users/nav' %>
 
-<h1>Cards</h1>
+<h1 class="heading">
+  <span class="flex items-center">
+    Cards
+    <%= badge_for @stripe_cards.count, class: "bg-muted" %>
```
maxwofford: Could we include emburse card count in this too? Feel free to push back if that doesn't work with the design of the page.

app/controllers/stripe_cards_controller.rb
```diff
@@ -76,11 +76,11 @@ def create
     }
     ::StripeCardService::Create.new(attrs).run
 
-    redirect_to event_cards_overview_path(event_id: event.id), flash: { success: "Card was successfully created." }
+    redirect_to event_cards_overview_path(event_id: event.slug), flash: { success: "Card was successfully created." }
```
maxwofford: Thanks for catching that! I'm under the assumption that the previous thing works, but this is a better looking URL. This is fine here, but in the future can we move this sort of thing to a separate PR?

maxwofford (approved):
Looks great! Thanks for dropping this into the page. I left a couple comments, but I'm happy with this getting merged.

app/views/stripe_cards/_form.html.erb
cjdenio: Yep!

app/views/static_pages/my_cards.html.erb
cjdenio: Done!

PR #2625, opened 2022-06-03
Stripe charges show international currencies

maxwofford:
Screen Shot 2022-06-03 at 10 58 19

Screen Shot 2022-06-03 at 10 58 33

cc @sampoder. could you give context on what sorts of issues you run into with HCB currently only supporting USD?


cjdenio (approved):

exu3 (approved):
LGTM!

PR #2628, opened 2022-06-03
Users can delete receipts from their team

maxwofford:
A user on bank just reported having trouble removing a receipt from their transaction. Looks like the cause was b/c we only authorize the uploader to destroy a receipt.


app/policies/receipt_policy.rb
```diff
@@ -2,7 +2,7 @@
 
 class ReceiptPolicy < ApplicationPolicy
   def destroy?
-    user&.admin? || (user && record.user == user)
+    user&.admin? || record.user == user || record&.event&.users&.include?(user)
```
cjdenio: Could this just be
```suggestion
    user&.admin? || record&.event&.users&.include?(user)
```
?

Looks great to me otherwise!

cjdenio (approved):

app/policies/receipt_policy.rb
maxwofford: Good point!

PR #2631, opened 2022-06-05
Paginate deletion requests page

cjdenio:
One more little improvement to the deletion requests page.

This PR also makes open requests show up on top, to make sure they're always on the first page.


exu3 (approved):
Nice, thanks; it looks great!

One suggestion: The `Displaying deletion requests ... in total` text looks a bit out of place. Thoughts on putting it in a `small` tag in a muted colour with some vertical padding?

exu3:
Looks great!

PR #2633, opened 2022-06-05
Don't show transparency message on attach_receipt page when not signe…

exu3:
…d in

closes #2592


cjdenio (changes requested):
Hmm, this also makes the header disappear (which might be desired)

app/views/layouts/application.html.erb
```diff
@@ -57,7 +57,7 @@
         </article>
       </main>
     <% else %>
-      <%= render 'application/header' unless @no_app_shell %>
+      <%= render 'application/header' unless (@no_app_shell || @no_transparency_header) %>
```
cjdenio: Oop I think you meant to put this on line 32

exu3: https://github.com/hackclub/bank/pull/2633/commits/62b55ca6850908de3cc3a2ed6604890372485285 Just fixed, thanks! I should test things before I commit :festpee:

app/views/events/show.html.erb
```diff
@@ -1,5 +1,6 @@
 <% title @event.name %>
 <% include_modals %>
+<% no_transparency_header %>
```
cjdenio: hmmmmmmmm i think you meant to remove this

exu3: Ah I did, thank you.

cjdenio (approved):
Looks great! Thanks for fixing this!

PR #2634, opened 2022-06-05
[UI change] Showing event UI alongside the user-facing disbursements UI

ykdojo:
- First PR for #2531
- Depends on the PR #2627


ykdojo:
I think this one's ready for a review after #2627 is merged.

I don't have all the context around how disbursements are used in other places, so pls let me know if there's anything to change.

Also, it still has the minor issue of the event name being repeated in the URL like:

`devhacks-test-event/disbursements/new?source_event_id=devhacks-test-event`

but I don't think it'll be a huge issue for users - let me know if you want me to change that, also, though.

PR #2635, opened 2022-06-05
Fix point-of-contact admin bug

cjdenio:
Fixes a bug where events become invalid if Mel pretends not to be an admin


maxwofford:
Merging with confirmation from @cjdenio in Slack

PR #2638, opened 2022-06-06
Add initial fronted balance to v2 balance

maxwofford:
Co-Authored-By: Ella <72365100+exu3@users.noreply.github.com>

This depends on https://github.com/hackclub/bank/pull/2621, which should implement a `raw_pending_incoming_disbursement_id` on `PendingCanonicalTransaction`


garyhtou (approved):

app/models/event.rb
```diff
@@ -290,6 +291,18 @@ def settled_outgoing_balance_cents(start_date: nil, end_date: nil)
       end
   end
 
+  def fronted_incoming_balance_v2_cents(start_date: nil, end_date: nil)
```
garyhtou: Feedback: rename to `pending_incoming_balance_v2_cents`. "fronted" is more of a Bank ops term and may be confusing for those who aren't familiar with bank ops.

exu3: `pending_incoming_balance_v2_cents` is already a thing, I think https://github.com/hackclub/bank/blob/9dd0fafad248add04c00183eef4158253dcd14dc/app/models/event.rb#L312

I agree that "fronted" might not be the best term here though

maxwofford: Going forward with not addressing this right now because I think the wording will change once we add our instant deposit feature. In the context of instant disbursements the current wording makes sense (this is specifically fronted money from the ops team).

PR #2642, opened 2022-06-07
Disbursements create pending transactions

maxwofford:
A new pass in place of https://github.com/hackclub/bank/pull/2621.

There's a whole lot in here, so I created a file to help spin it up locally: https://github.com/hackclub/bank/pull/2642/files#diff-54bf6dd72357b503ce605397fce59d10ae38ad9e9d9d4aca855c491259b070d4. I'll remove it before merging.

This PR introduces `RawPendingIncomingDisbursementTransaction`s and `RawPendingOutgoingDisbursementTransaction`s.

Now, HCB codes tied to disbursements will have 2 pending transactions tied to them:
Screen Shot 2022-06-09 at 14 02 33

Those pending transactions will show up on their subsequent event lists:
Screen Shot 2022-06-09 at 14 05 36
Screen Shot 2022-06-09 at 14 06 13

If a disbursement is rejected by an admin, or errors on processing, the pending transactions will be declined, and be hidden from the event's transaction list:
Screen Shot 2022-06-09 at 14 03 10
Screen Shot 2022-06-09 at 14 03 22


garyhtou (approved):
Looks good!

garyhtou:
Now that we have a better separation between incoming/outgoing disbursement, it'll be easier to open up disbursement's HCB Code show pages for normal users.

When a user visits a disbursement's HCB Code show page:

- Verify user has access to at least one of the events
- Use `referrer` to determine which Event context the page should be displayed in
- Provide a button to switch to the other Event's context

cjdenio:
What if incoming/outgoing disbursement transactions each had their own separate HCB codes? It seems like that would fix all those weird event context issues

PR #2645, opened 2022-06-09
Move Stripe pending transaction creation to `issuing_authorization.created` webhook

cjdenio:
This actually fixes 2 bugs:

- An issue where Stripe-declined transactions (e.g. with frozen cards) don't instantly show up on the pending ledger
- race condition thing #2644
  - Screen Shot 2022-06-08 at 11 10 36 PM wheeee

Fixes #2644


maxwofford (approved):
This looks great! I'm cool with this being merged. We might want to move this into a `::StripeAuthorizationService::Webhook::HandleIssuingAuthorizationCreated` service, but the logic flow here looks good.

cjdenio:
Once the v1 stuff is removed we can have this one-liner:

```ruby
::StripeAuthorizationJob::CreateFromWebhook.perform_later(event[:data][:object][:id])
```

PR #2649, opened 2022-06-09
Make email receipt upload case insensitive

cjdenio:
Fixes #2639


app/mailboxes/receipt_uploads_mailbox.rb
```diff
@@ -35,7 +35,7 @@ def process
   private
 
   def user
-    @user ||= User.find_by email: mail.from
+    @user ||= User.find_by("email ILIKE ?", User.sanitize_sql_like(mail.from[0]))
```
maxwofford: If these are already being stored in downcase only, could we just do `User.find_by email: mail.from.downcase`?

cjdenio: Hmm, that's true, though I'm a little worried about writing code that relies on emails being all-lowercase

maxwofford: User emails are stored in downcase https://github.com/hackclub/bank/blob/eb0129324841730c660578018ad97ba4231200be/app/models/organizer_position_invite.rb#L54

app/mailboxes/receipt_uploads_mailbox.rb
```diff
@@ -35,7 +35,7 @@ def process
   private
 
   def user
-    @user ||= User.find_by email: mail.from
+    @user ||= User.find_by(email: mail.from[0].downcase)
```
maxwofford: How was this code running with `mail.from` before if `mail.from[0]` is required now?

cjdenio: `mail.from` is an array, and `find_by` accepts an array (e.g. `User.find_by(email: ["email1@gmail.com", "email2@gmail.com"])`)

cjdenio:
```ruby
mail.from.map(&:downcase)
```

could also work, but I think the `From` header can only contain one value

maxwofford (approved):
LGTM!

PR #2650, opened 2022-06-09
Create user on invite if they don't exist

albertchae:
https://github.com/hackclub/bank/issues/2415

Previously, if a non existing user was invited, we would not associate
user to OrganizerPositionInvite until the user logged in. This led to
some complex logic trying to dynamically associate the user later, see

- https://github.com/hackclub/bank/pull/2327
- https://github.com/hackclub/bank/pull/2473

While it's not entirely clear why this approach didn't work, we can just
circumvent it by creating the user when creating the invite. This
doesn't break any existing assumptions about users (i.e. the presence of
a user record does NOT imply that the user has signed in).

For implementing this change, I could have tried using callbacks on the
`OrganizerPositionInvite` model to create the user, but it is generally a
bad idea to have callbacks touch another model as it tends to get
messy quickly (see https://canolcer.com/post/active-record-callbacks/)

So I went with a service object to follow the convention of the rest of
the codebase. This had a few consequences:

- the service still has to expose the model for use with form_with and
  some error rendering. I will explore whether we can avoid passing the
  model to `form_with` and if we can render errors directly from the service
  in a follow up.
- we were calling the ! version of persisting (create/save) in some
  places and not in others, so ~I added a boolean to the `run` method to
  designate throwing an exception on failure or not.~ This is now `run` vs `run!`
- I moved the `normalize_email` method from the model to the service
  since we need it to happen before creating the user

Follow ups:
- [ ] Backfill existing OrganizerPositionInvite with null user and make user_id on organizer_position_invites NOT NULL
- [ ] Remove methods for non existing user on organization_position_invite
- [x] Add tests
- [ ] Drop user email on organizer position invite (can users change email?)


maxwofford (commented):
As mentioned on our call, go ahead and assign me for review once the changes we talked about are in.

spec/services/organizer_position_invite_service/create_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    partner = Partner.create!({
+                                slug: SecureRandom.hex(30)
+                              })
+    Event.create!({
+                    name: 'test-event',
+                    partner: partner,
+                    sponsorship_fee: 0,
+                    organization_identifier: SecureRandom.hex(30)
+                  })
+  end
```
albertchae: I will refactor this to factorybot after https://github.com/hackclub/bank/pull/2660 lands

albertchae:
@maxwofford this is ready for review now, although there will be a few follow up PRs after

maxwofford (approved):
Looks good to merge. Thanks for breaking up the changes into each commit. Excited to see this setup with factories!

PR #2652, opened 2022-06-09
Remove unused session_token on users

albertchae:
After multiple user sessions were implemented in https://github.com/hackclub/bank/pull/2049, session_token on the users table is no longer used, so drop the column to prevent confusion with the session_token column on user_sessions.

I believe the has_session_token scope and the create_session_token
callback were the only 2 uses of this column

Closes https://github.com/hackclub/bank/issues/2624


maxwofford (approved):
Looks good to merge!

PR #2654, opened 2022-06-10
Fix `uninitialized constant` error in when rails c in production and staging

exu3:
`heroku run rails c -a  bank-hackclu-exu-uninit-ilwgum`

https://github.com/RubyMoney/money-rails/issues/614


exu3:
Merging after getting verbal approval from Max.

PR #2657, opened 2022-06-10
Initial pass on incoming/outgoing disbursements tests

maxwofford:
(PR description by @albertchae)

I paired with @maxwofford on the initial commit here, where we were mainly just trying to machete a path through the jungle to get a test working at all. All the follow up commits here are to bring the test style more in line with what I've seen on other rails projects and I recommend checking them one by one to see the progression. 

Here is a summary of the bigger points though:
- Because of this transactional setting https://github.com/hackclub/bank/blob/d0afdee66e7b107b27e629a33c02f8bb38272f58/spec/rails_helper.rb#L47, each test is running in a separate database transaction so there is no need to teardown at the end because the transaction will just be rolled back
- Tests should generally follow a 3 step pattern (Arrange/Act/Assert or Setup/Exercise/Assert). Some people disagree on this but I like to also assert things in the preconditions just to be very explicit, for example https://github.com/hackclub/bank/pull/2657/files#diff-9f0084498b471fbcc3368f7092dccaf3f12a02b0c7d0df8712000b5aa03e2bb6R54-R55 to guarantee the associations are nil on creation
- We should avoid instance variables in tests to not cross contaminate tests. The most common way of doing this in most [Rails projects is using `let`](https://www.codewithjason.com/difference-let-let-instance-variables-rspec/), however I recommend against overusing them because I've seen tests become unmaintainable as a result of not being clear about what data is required for each test.
- Although it goes against our instincts in almost every other part of writing code, I recommend resisting DRYing up tests because it will end up making tests too coupled to each other and hard to maintain. The acronym I've seen used is DAMP (Dry and Meaningful Phrases) which refers to preferring some duplication to make it more obvious what is happening in each test. However, this doesn't mean don't write helpers or other shared bits of code. It's still worth DRYing up some of the setup code so that each test can change only what matters to it and avoid [irrelevant information](http://xunitpatterns.com/Obscure%20Test.html#Irrelevant%20Information) (e.g. in this PR, `sponsorship_fee`and `organization_identifier` are required to create a valid `event` because of db constraints, but otherwise are not relevant to the test, so it's good to have that in a helper method)

Anyway, I think tests are as much an art as a science and it will also kind of come down to what sort of cost you're willing to pay in maintenance versus how much confidence you gain that the code is working properly. Here's some additional reading and video material:

- https://freecontent.manning.com/making-better-unit-tests-part-1-the-aaa-pattern/
- https://enterprisecraftsmanship.com/posts/dry-damp-unit-tests/
- https://thoughtbot.com/blog/mystery-guest
- https://thoughtbot.com/blog/lets-not
- https://thoughtbot.com/blog/my-issues-with-let
- https://github.com/thoughtbot/guides/blob/b1399bbfd33acca56746098a22fe8b4aaae51110/testing-rspec/avoid_let_spec.rb


albertchae:
@maxwofford this is ready for review now, ironically I can't assign you as a reviewer because you are technically the author. I recommend checking it out commit by commit and especially the part about using `let` vs not.

This PR also didn't introduce factorybot because IMO that's another level of complexity, so I made a separate PR for that https://github.com/hackclub/bank/pull/2660

maxwofford:
Super helpful going over this with you! Merging into https://github.com/hackclub/bank/pull/2642's branch

PR #2660, opened 2022-06-12
Introduce factory_bot for creating test data instead of fixtures

albertchae:
This PR introduces `factory_bot` as an alternative to creating test data with fixtures.
See https://semaphoreci.com/blog/2014/01/14/rails-testing-antipatterns-fixtures-and-factories.html for more opinionated thoughts on using fixtures (from the linked article: `Do not use fixtures.` 😆 )

I made this as a separate PR from https://github.com/hackclub/bank/pull/2657 to explicitly show the difference between creating test data with the model vs factories. IMO the biggest benefit is that you can rely on some sane defaults to wire up associations automatically as well as have some setup that is required to make the object valid (but is not relevant to the test). The negatives are that the factory_bot DSL can be pretty weird and it's sometimes difficult to go from a factory creation to where it's defined, and also if creating a set of data in a certain state is managed by a service object in the application and that service object is already well tested, I would recommend using that instead to setup the test data instead of trying to duplicate that complexity in factory definitions

An orthogonal issue I discovered while working on this is that because of the way we load [environment variables when running docker](https://github.com/hackclub/bank/blob/d0afdee66e7b107b27e629a33c02f8bb38272f58/docker_start.sh#L52), doing `./docker_start bash` and then running `bundle exec rspec` actually is running against the `development` database (and rails environment) which can lead to a lot of confusion (@maxwofford I think this might also have contributed to some of the teardown issues we were running into). One of the reasons we want to have a separate DB for running tests is that we can [drop and reload the schema at will](https://guides.rubyonrails.org/v2.3/testing.html#preparing-your-application-for-testing).

To validate that setting to test works, I tried https://github.com/hackclub/bank/pull/2660/commits/4505f295f2bc2e1997b9064ad2910f0b1fd7b6db, but we need to think of a better way to manage running tests before merging this PR. I would also double check whether the CI setup on github is running in `test` vs `development`


spec/disbursement_wip.rb
```diff
@@ -35,21 +20,21 @@ def create_disbursement(source_event, destination_event)
     end
 
     it "creates valid setup data" do
-      partner = create_partner
-      source_event = create_event('source hacks', partner)
-      destination_event = create_event('destination hacks', partner)
-      disbursement = create_disbursement(source_event, destination_event)
+      partner = create(:partner)
+      source_event = create(:event, partner: partner)
+      destination_event = create(:event, partner: partner)
```
albertchae: we have to supply `partner` to `event` instead of rely on the default factory `association` we defined or these events will end up tied to different `partner`s

spec/disbursement_wip.rb
```diff
@@ -35,21 +20,21 @@ def create_disbursement(source_event, destination_event)
...
+      disbursement = create(:disbursement)
```
albertchae: technically in this test we don't care that the disbursement is linking source and destination event. I would argue that checking the models are valid belongs in model specs which we can move this to later

spec/factories/disbursement_factory.rb
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
+  factory :disbursement do
+    association :event
+    association :source_event, factory: :event
```
albertchae: we have to specify the factory for `source_event` because the association name is not 1 to 1 with the model, although it wasn't necessary for `event` because it does map to the model

spec/factories/event_factory.rb
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
+  factory :event do
+    name { Faker::Name.unique.name }
+    association :partner
```
albertchae: This will create a new partner and associate it to this event UNLESS a partner is explicitly provided

.env.docker
```diff
@@ -1,3 +1,3 @@
-RAILS_ENV=development
+RAILS_ENV=test
 PORT=3000
-DATABASE_URL=postgres://postgres@db:5432/bank_development
+DATABASE_URL=postgres://postgres@db:5432/bank_test
```
albertchae: DO NOT MERGE, but we need to figure out a way to deal with this

maxwofford (commented):
Thanks for this! Chatted over call, but let's get the docker stuff figured out before merging.

Could we setup docker to create both `bank_development` & `bank_test` and not use `DATABASE_URL` in `.env.docker`?

Ideally I can run `bundle exec rspec` directly in the development docker context & it should automatically use `bank_test`.

albertchae:
@maxwofford I pulled out the docker changes into a separate PR https://github.com/hackclub/bank/pull/2698

albertchae:
@maxwofford this is ready for re-review since I merged https://github.com/hackclub/bank/pull/2698

spec/disbursement_wip.rb
maxwofford: Now that I'm thinking about what we're testing, I think it's fine if they're under different partners, correct? Having one partner was done for simplicity when tests were directly using ActiveRecord.

maxwofford (approved):
Thanks for adding this and making it pretty verbose– `factory_bot` seems awesome. Added a comment, but it's not blocking merge.

.env.docker
albertchae: resolved in https://github.com/hackclub/bank/pull/2698

spec/disbursement_wip.rb
albertchae: yeah I think the test still passes, so I'll remove it

PR #2662, opened 2022-06-12
Add annotate gem to improve developer workflow

albertchae:
Closes https://github.com/hackclub/bank/issues/2656

Install https://github.com/ctran/annotate_models gem, which automatically comments model files with the database fields. In my experience this is nicer for developers than having to check `schema.rb` or the database directly, so I recommend we try it out and see if that's the case.

These annotate settings will update the files whenever migrations are run, so they should stay in sync with the db schema.


lib/tasks/auto_annotate_models.rake
```diff
@@ -0,0 +1,59 @@
+# NOTE: only doing this in development as some production environments (Heroku)
```
albertchae: This file was auto generated by the gem installer, but I left a few comments on important aspects to take note of

lib/tasks/auto_annotate_models.rake
```diff
@@ -0,0 +1,61 @@
...
+      'frozen'                     => 'false',
+      'classified_sort'            => 'true',
+      'trace'                      => 'false',
+      'wrapper_open'               => nil,
+      'wrapper_close'              => nil,
+      'with_comment'               => 'true'
+    )
+  end
+
+  Annotate.load_tasks
```
albertchae: This line hooks annotate into the db migrate tasks

lib/tasks/auto_annotate_models.rake
```diff
@@ -0,0 +1,61 @@
...
+      'show_complete_foreign_keys' => 'false',
+      'show_indexes'               => 'true',
+      'simple_indexes'             => 'false',
+      'model_dir'                  => 'app/models',
+      'root_dir'                   => '',
+      'include_version'            => 'false',
+      'require'                    => '',
+      'exclude_tests'              => 'true',
+      'exclude_fixtures'           => 'true',
+      'exclude_factories'          => 'true',
```
albertchae: these are default `false`, but I changed them to `true` because I find annotating those files adds clutter and having them in the models is sufficient, but we can change our mind if we find that isn't the case!

maxwofford (approved):

PR #2664, opened 2022-06-13
Fix donation iframe

cjdenio:
This PR makes the donation iframe work again. (fixes #2658!)

### The problem

Cookies act a little weird inside cross-origin iframes. Actually they only work if you set the `SameSite=None` attribute. And they don't work at all in Safari. aaaaaaaaaaaaaa

### The fix

CSRF token verification and `invisible_captcha`'s spinner validation thing both write to the session (a cookie!), so I disabled them.

I don't have context on why `invisible_captcha` was originally added (or what exactly it's doing), so I'm not exactly sure how badly this affects our spam-blocking abilities. I guess the biggest concern would be that this makes it easier to make automated `make_donation` requests (whereas before you would have had to spoof the CSRF token)

Tested in Firefox, Chrome, and Safari!


garyhtou:
Ah hm interesting. I think this should be okay! Thanks for patching this

cjdenio:
Gonna merge this to make sure it's fixed for http://kiddobyte.org/donate – we should maybe look into the spam stuff more this week

exu3:
Thanks for fixing this!

PR #2665, opened 2022-06-13
Remove 10 day check on receipt upload

exu3:
Users can upload receipts from and access the attach_receipt page as long as the signature/hash matches.


app/models/hcb_code.rb
```diff
@@ -18,6 +18,10 @@ def url
     "/hcb/#{hashid}"
   end
 
+  def receipt_secret
+    HcbCodeService::Receipt::SigningEndpoint.new.create(self)
+  end
+
```
garyhtou: This totally works, however, it would be better to move this into a HcbCodeHelper since this is not really related to the model itself.

app/controllers/hcb_codes_controller.rb
```diff
@@ -62,8 +64,14 @@ def receipt
     if params[:show_link]
       redirect_to params[:redirect_url], flash: { success: { text: "Receipt".pluralize(params[:file].length) + " added!", link: hcb_code_path(@hcb_code), link_text: "View" } }
     else
-      redirect_to params[:redirect_url], flash: { success: "Receipt".pluralize(params[:file].length) + " added!" }
```
garyhtou: Do you happen to know why we're redirecting to `:redirect_url`? What you have looks good, but let's make sure we don't break any existing functionality.

app/controllers/hcb_codes_controller.rb
```diff
@@ -62,8 +64,14 @@ def receipt
...
+      redirect_to @hcb_code.receipt_secret, flash: { success: "Receipt".pluralize(params[:file].length) + " added!" }
     end
+
+  rescue Pundit::NotAuthorizedError
+    unless HcbCodeService::Receipt::SigningEndpoint.new.valid_url?(@hcb_code.hashid, params[:s])
+      raise
+    end
+
```
garyhtou: I think there's a chance you might not need to rescue here since you already checked for a signature above. It should be okay to just let it error.

app/controllers/hcb_codes_controller.rb
exu3: Ah I think it might be because the redirect url can be the attach_receipt page, hcb code show page, or my/inbox page depending on where the receipt is uploaded from.

exu3: I reverted this change.

garyhtou (approved):

app/controllers/hcb_codes_controller.rb
exu3: Ah currently it redirects to `:redirect_url` (attach_receipt page) without the signature in the url param, so users get a "not authorized" error after their receipt successfully uploads. https://hackclub.slack.com/archives/CN523HLKW/p1655733330691369

PR #2667, opened 2022-06-13
Add fields to event_new form

exu3:
This PR adds 2 fields to the admin organization creation form.
Screen Shot 2022-06-13 at 4 28 48 PM


hackliv:
Can we also add the type of org filter? thinking nonprofit, event, hack club, & hackathon as categories. or are we thinking this as a separate pr?

exu3:
> Can we also add the type of org filter? thinking nonprofit, event, hack club, & hackathon as categories. or are we thinking this as a separate pr?

Thinking of adding that in a separate PR as it adds a field that doesn't already exist

hackliv (approved):

PR #2671, opened 2022-06-13
Hotfix for #2670

maxwofford:
Fixes instances of #2670 data leak


garyhtou (approved):

maxwofford:
This also patches an instance of account number showing in TX history partials (see https://bank.hackclub.com/hcb/lZHvak)

garyhtou:
> This also patches an instance of account number showing in TX history partials (see [bank.hackclub.com/hcb/lZHvak](https://bank.hackclub.com/hcb/lZHvak))

I think the fix for friendly memo should handle the account numbers in the TX history partials— unless the friendly memo was already saved to the DB (and not using one generated by the service)

PR #2675, opened 2022-06-14
Remove email receipt upload from feature previews

cjdenio:
(also fixed a few stray HTML tags; I recommend clicking "hide whitespace changes" to see what was actually changed)


maxwofford (approved):
Thanks for taking the initiative on cleaning this up! I was kinda thinking we could use the section to also show stuff that's already shipped, kinda like [Steam Labs](https://store.steampowered.com/labs). Thoughts on just removing the flipper flag stuff and leaving the HTML, but dropping it to the bottom?

Just playing with the idea, I don't have strong opinions about it. Other then that thought this is good to merge!

cjdenio:
Oh, we can totally leave it in there.

(At some point it might be good to break that section off into its own page, so it doesn't get too cluttered)

PR #2676, opened 2022-06-15
Update UI to communicate fronted balance

exu3:
This PR updates the disbursement hcb code partial and new transfer form.
Screen Shot 2022-06-15 at 9 10 09 AM
Screen Shot 2022-06-15 at 9 01 44 AM


app/views/hcb_codes/_disbursement.html.erb
```diff
@@ -21,15 +21,9 @@
 
   <section class="card__banner card__darker details-horiz border-top border-bottom">
     <p>
-      <strong>Requested at</strong>
+      <strong>Transferred at</strong>
       <%= format_datetime @hcb_code.disbursement.created_at %>
```
garyhtou: Would the `approved_at` time be more accurate for "Transferred at"? I believe the pending transactions will only be created once the disbursement has been approved

exu3: I don't see a `approved_at` for disbursements; there's a `fulfilled_at` time, but it seems to be the same as the `created_at` time. https://bank.hackclub.com/hcb/yXH4Eb

maxwofford (approved):
Looks great! Go for merge

PR #2689, opened 2022-06-16
Use OrganizerPositionInviteService::Create service when creating an event and inviting organizers

albertchae:
I missed this during https://github.com/hackclub/bank/pull/2650 because
I forgot about the association based create syntax. Hopefully this
should be the last place this could happen.

For https://github.com/hackclub/bank/issues/2415


maxwofford (approved):
Thanks for the fix! Also another place to swap in a service, nice. Please let @hackliv know once it's merged.

albertchae:
> Also another place to swap in a service, nice.

Yep it plugged in quite nicely! I kind of wanted to go to the extreme of preventing `OrganizerPositionInvite.create` from anywhere outside of the service. I've seen ideas for doing that using private active record models https://kellysutton.com/2019/10/29/taming-large-rails-codebases-with-private-activerecord-models.html, investigating the backtrace to only allow calls from the service, or https://github.com/Shopify/packwerk. But all of those seemed a bit overkill for now.

PR #2690, opened 2022-06-17
"Event not found" -> "Organization not found"

cjdenio:
I think the "event" terminology is a little confusing here, and it seems like "project" is being used more now.

Screen Shot 2022-06-16 at 8 11 31 PM


exu3 (commented):
Good call! I think we should use "organization" instead of "project" to be consistent with the Bank marketing page (https://github.com/hackclub/site/pull/458)

cjdenio:
Oh hmm. "Project" is already used a bunch throughout the UI (but so is "Organization").

Screen Shot 2022-06-16 at 9 51 53 PM
Screen Shot 2022-06-16 at 9 51 01 PM
Screen Shot 2022-06-16 at 9 50 54 PM

I guess I'd think of it as "You're signing up your organization for a project on Bank", so the thing on Bank is a "project" but an "organization" is using Bank. Not sure.

garyhtou:
I think we're trying to move towards using "organization" in place of "event" and "project". if you see any instances of "event" or "project", feel free to switch them over (especially in the UI)

garyhtou (approved):
go for it! thanks :)

PR #2692, opened 2022-06-17
Add fields to events

exu3:
This PR:
- brings back the previously deprecated "point of contact" field (but doesn't do anything yet at the time of writing)
- introduces a new "category" field (enum)
- brings back the previously deprecated start and end date fields for events (which also isn't used anywhere at the time of writing)
- adds `category` and `point_of_contact` as filters on the admin events page

closes #2685 
closes #2518


hackliv:
this looks great! were you able to see about removing some of the admins and just making it mel, dev, and I?

garyhtou:
Do we have a plan to backfill some data? Also, for the `end_date`, would it make sense for it to be `nil` if the event is still running? that would make it easy for us find events that are "dead" (where end_date is not nil)

exu3:
> were you able to see about removing some of the admins and just making it mel, dev, and I?

This PR only adds the point of contact field but doesn't touch the data there; Since removing admins is just a data change / not a code change, we could do that in the UI after this PR is merged. Existing features (eg. Documents page, organizer position invites, etc.) depend on certain admin users for context—it might feel a little weird if someone you didn't know uploaded a legal document on your Documents page and they aren't an admin.

exu3:
> Do we have a plan to backfill some data? Also, for the end_date, would it make sense for it to be nil if the event is still running? that would make it easy for us find events that are "dead" (where end_date is not nil)

This PR just touches the code to add fields @hackliv requested; we should probably backfill some data though. Should we discuss this on Slack after this PR is merged?

I was thinking start and end date would be the date/time the particular hackathon/event starts/ends, so that the Ops team can front money, anticipate more communication/support around the time of the event.

maxwofford (approved):
The code changes look good! Go for merge once you're ready!

PR #2694, opened 2022-06-17
Gate updated balance to feature flag while testing

maxwofford:
Co-Authored-By: Ella <72365100+exu3@users.noreply.github.com>

This allows us to enable the new balance calculation on a per-event basis using...
- `Flipper.enable_actor(:fronted_balance_2022_06_17, Event.find(6))`
- `Flipper.disable_actor(:fronted_balance_2022_06_17, Event.find(6))`
_examples use devhacks (id=6)_


maxwofford:
Here we see devhacks:
Screen Shot 2022-06-17 at 14 28 51

Here is devhacks after I created an incoming disbursement and enabling the feature flag:
Screen Shot 2022-06-17 at 14 29 19

Upon disabling the feature flag, the transactions stay around, but the balance goes back to its original value:
Screen Shot 2022-06-17 at 14 30 05

exu3 (approved):

app/models/event.rb
```diff
@@ -251,10 +251,14 @@ def emburse_balance
...
-      settled_balance_cents(start_date: start_date, end_date: end_date) +
-      pending_outgoing_balance_v2_cents(start_date: start_date, end_date: end_date) +
-      fronted_incoming_balance_v2_cents(start_date: start_date, end_date: end_date)
+    @flipper_balance ||= Flipper.enabled?(:fronted_balance_2022_06_17, self)
+    @balance_v2_cents ||= begin
+      sum = 0
+      sum += settled_balance_cents(start_date: start_date, end_date: end_date)
+      sum += pending_outgoing_balance_v2_cents(start_date: start_date, end_date: end_date)
+      sum += fronted_incoming_balance_v2_cents(start_date: start_date, end_date: end_date) if @flipper_balance
+      sum
```
exu3: Is line 260 necessary?

maxwofford: It's not required, but added to increase readability (the if statement starts trailing off the 80 character limit).

PR #2696, opened 2022-06-17
Add birthday field to users

cjdenio:
Closes #2678! Currently the birthday field is optional, but it could be made required with some fancy validation logic.


maxwofford:
Let's review this IRL now that you're in VT– could you coordinate to fine 15-30 min we're both free over Slack?

cjdenio:
Approved IRL by @maxwofford.

PR #2698, opened 2022-06-18
Create test db and use it for tests

albertchae:
This PR is for addressing https://github.com/hackclub/bank/pull/2660#pullrequestreview-1009930872

Even though config/database.yml specifies both development and test DBs
https://github.com/hackclub/bank/blob/0ad7ccfe6c8bc1148afbd1682821011cfcccfd52/config/database.yml#L8-L14
the current docker setup was only creating `bank_development`, which was
causing issues when trying to run rspec and also ran the risk of running
tests messing with development data. This test creates the Rails standard
test database for running tests.

- Add a `rails db:test:prepare RAILS_ENV=test` to create `bank_test`,
based on https://guides.rubyonrails.org/testing.html#maintaining-the-test-database-schema

I think RAILS_ENV=test isn't strictly required, but it does prevent this
warning
```
rails aborted!
ActiveRecord::ProtectedEnvironmentError: You are attempting to run a destructive action against your 'production' database.
If you are sure you want to continue, run the same command with the environment variable:
DISABLE_DATABASE_ENVIRONMENT_CHECK=1
/usr/src/app/bin/rails:7:in `<top (required)>'
/usr/src/app/bin/spring:12:in `require'
/usr/src/app/bin/spring:12:in `block in <top (required)>'
/usr/src/app/bin/spring:9:in `tap'
/usr/src/app/bin/spring:9:in `<top (required)>'
Tasks: TOP => db:test:load => db:test:purge => db:check_protected_environments
(See full trace by running task with --trace)
```

Regardless of using it or not, `ar_internal_metadata` in `bank_test`
reports `test`
```
bank_test=# select * from ar_internal_metadata;
     key     |                  value                   |         created_at         |         updated_at
-------------+------------------------------------------+----------------------------+----------------------------
 environment | test                                     | 2022-06-18 10:34:53.244898 | 2022-06-18 10:34:53.244898
```

- Remove RAILS_ENV from .env.docker. The test environment will
  automatically run in `test` and server + console run in `development`
  by default, so this is not necessary.

- Remove specifying `bank_development` from the DATABASE_URL, since
  rails can use config/database.yml and the Rails.env to figure out
  which db to connect to.

- Also replace `rails db:create db:migrate` with `db:prepare` which runs all of those commands
https://blog.saeloun.com/2019/04/11/rails-6-rails-db-prepare.html


albertchae:
From https://github.com/hackclub/bank/pull/2660#pullrequestreview-1009930872

> Could we setup docker to create both bank_development & bank_test 

✅ 

> not use DATABASE_URL in .env.docker?

I believe DATABASE_URL is required here https://github.com/hackclub/bank/blob/0ad7ccfe6c8bc1148afbd1682821011cfcccfd52/config/database.yml#L6 because otherwise Rails tries to look for postgres at localhost:5432, but I was able to remove specifying the `bank_development` database

> Ideally I can run bundle exec rspec directly in the development docker context & it should automatically use bank_test.

✅ , without `RAILS_ENV` in the `.env.docker` and having `bank_test` available, we got this for free

maxwofford (approved):
Great! Once this is merged, please post in the #hq-bank channel on Slack saying people will want to rerun `docker_setup.sh` if they're running tests locally.

PR #2699, opened 2022-06-21
Fix bug that marked disbursements as already fulfilled

maxwofford:
At the time of writing, you can see the issue for yourself on https://bank.hackclub.com/admon/disbursements/ – all the disbursements are processing, but haven't completed successfully since Jun 15th. This was caused by a misunderstanding of what `fulfilled_at` was used for in https://github.com/hackclub/bank/commit/51db5f1121bff565100c7e2059cb95c3793debd3.

We'll also need to fix the disbursements that were created from Jun 15, 2022 - June 21, 2022 by zeroing out their `fulfilled_at` field.


garyhtou (approved):
Oh yeah, that is very confusing. I think we should jump on https://github.com/hackclub/bank/issues/1294 if someone gets a chance

PR #2701, opened 2022-06-21
Fix receipt upload

cjdenio:
Closes #2700


exu3 (approved):
Can confirm it works locally. Thanks for the fix!

cjdenio:
_looks both ways_

PR #2702, opened 2022-06-22
Create one time job to fix disbursements with incorrect state

maxwofford:
This addresses the historical data issue fixed in https://github.com/hackclub/bank/commit/020b9eda563de4b017790204a2e01eb24a88db9e. Disbursements 965 through 979 were created before the fix was in & this brings them back into the correct state.

Once in prod this should be queued up (ie. `OneTimeJobs::ResetAffectedDisbursementsJob.perform_later`)


garyhtou (approved):

maxwofford:
I just queued up the job and it looks like it ran successfully:

Before
Screen Shot 2022-06-22 at 16 07 35

After
Screen Shot 2022-06-22 at 16 07 57

garyhtou:
yayyy!

PR #2705, opened 2022-06-22
Display invoice number instead of StripeInvoiceId

exu3:
Invoices paid through Bill.com/similar platforms refer to the invoice using the invoice number. This PR changes the StripeInvoiceId column on the admin index page to display the invoice number instead.

Screen Shot 2022-06-22 at 11 37 53 AM


exu3:
Just to sanity check, is the `StripeInvoiceId` on the admin invoices page used for anything important?

hackliv:
> Just to sanity check, is the `StripeInvoiceId` on the admin invoices page used for anything important?

not a lot, I think that the "view in Stripe" admin button is usually used

exu3:
Reviewed on Slack!

PR #2706, opened 2022-06-22
Change user_id on organizer_position_invite to NOT NULL

albertchae:
https://github.com/hackclub/bank/issues/2415

Now that we create the user before creating organizer_position_invite as
of https://github.com/hackclub/bank/pull/2650, and fixed all the
existing ones with null user with https://github.com/hackclub/bank/pull/2695,
we can set this column to NOT NULL.

We can also remove several checks for non existent user as a result of
this as well.

- [x] Reminder: notify #hq-bank to rebuild their DB or run one time job as this migration will fail otherwise


albertchae:
✅ ~~Can't merge this until we run the job from https://github.com/hackclub/bank/pull/2695~~

albertchae:
This is ready to review after https://github.com/hackclub/bank/pull/2709 and https://github.com/hackclub/bank/pull/2713

maxwofford (approved):

PR #2707, opened 2022-06-22
Make Stripe transactions renameable

cjdenio:
https://user-images.githubusercontent.com/34525547/175091466-707c8eaf-9cf2-465d-8c56-4b467abde79a.mov


maxwofford (approved):
Nice! Does this work if I edit a pending HCB, before the actual transaction hits? If it handles that case, feel free to merge!

cjdenio:
Yep!

There's still a bit of weirdness with custom memos belonging to individual transactions, though it's not serious. We can solve that stuff when we make other transaction types renameable.

PR #2709, opened 2022-06-22
Refactor organizer_position_invite_spec to use factory bot

albertchae:
This required creating a minimal user factory as well.

I need this spec to be using factories instead of fixtures before fixing
the spec issues on https://github.com/hackclub/bank/pull/2706, so I
decided to pull out a pure structural refactor into its own PR before.


spec/services/static_page_service/index_spec.rb
```diff
@@ -3,7 +3,7 @@
 require "rails_helper"
 
 RSpec.describe StaticPageService::Index, type: :model do
-  fixtures "users", "events", "g_suites", "organizer_positions"
```
albertchae: this test must have passed due to cross test db pollution before

maxwofford (approved):

PR #2714, opened 2022-06-23
Redirect to new organization URL when renamed

cjdenio:
This works by storing a slug "history" inside the table generated by `friendly_id`.

https://user-images.githubusercontent.com/34525547/175351561-e09eeef7-36b9-4de4-bbf0-7aa8a70f47cd.mov

Screen Shot 2022-06-23 at 12 39 48 PM

Closes #2228

(note to self: run `Event.all.each { |event| FriendlyId::Slug.create(slug: event.slug, sluggable_id: event.id, sluggable_type: "Event") }` after merge)


exu3 (approved):
Looks good to me!

It worked locally after running `Event.all.each { |event| FriendlyId::Slug.create(slug: event.slug, sluggable_id: event.id, sluggable_type: "Event") }`

app/controllers/concerns/set_event.rb
```diff
@@ -0,0 +1,36 @@
...
+      # Redirect to the new slug
+      if id == params[:event_name]
+        params[:event_name] = @event.slug
+      elsif id == params[:event_id]
+        params[:event_id] = @event.slug
+      elsif id == params[:id]
+        params[:id] = @event.slug
+      end
+
+      redirect_to params.to_unsafe_h
```
albertchae: I'm curious what error you get if you don't `to_unsafe_h` or try using `to_h` instead? IIRC, you can call [`params.permit(...)`](https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-permit) to permit parameters you're expecting so you don't have to use the unsafe version for this. I'm not familiar with how redirecting with params works though, so I might be wrong here

app/views/events/_form.html.erb
```diff
@@ -91,12 +91,17 @@
...
+      <li>
+        We'll automatically redirect most pages to your organization's new URL, including:
+        <ul>
+          <li>Your donation page</li>
+          <li>Transparency Mode pages</li>
+          <li>Transaction bookmarks</li>
+        </ul>
+      </li>
+      <li>However, another organization may claim your current URL and you could be
+        <span class="bold">unable to get it back.</span>
```
albertchae: @maxwofford Is this something we should consider preventing? Because it means a different organization could try to squat on your old slug to get donations. Seems like bank isn't really catering to malicious users but just in case...

If we do want to prevent this, I think we could by adding a unique index on `FriendlyId::Slug` on a combination of the `sluggable_type` and `slug`columns

albertchae (commented):
Hey @cjdenio this looks and works great! I like how you factored out all the set event logic into a shared concern.

However, I was curious about what the `FriendlyId::Slug` table is for anyway and it looks like the intent of it is actually for a built in history feature
https://norman.github.io/friendly_id/file.Guide.html#History__Avoiding_404_s_When_Slugs_Change

As far as I can tell, if you set `friendly_id :name, use: :slugged` to `friendly_id :name, use: [:slugged, :history]`, the gem will handle creating the history entries for you automatically and also do the look up for you. Unless there's some behavior difference I'm missing, I would suggest letting the gem handle all that logic for you. Although I think keeping the `SetEvent` concern is still a good idea, but then you could delegate to the gem to do all the complicated stuff to find the historical slug

app/controllers/concerns/set_event.rb
cjdenio: So `to_unsafe_h` is needed because we want to make sure all the parameters are kept during the redirect. For example, `https://bank.hackclub.com/hq/transfers?filter=deposited` would look like:
```json
{
  "controller": "events",
  "action": "transfers",
  "filter": "deposited"
}
```
(or something)

app/views/events/_form.html.erb
cjdenio: That's a good point– maybe if an organization keeps a slug for one month it's "theirs" and nobody else can take it?

cjdenio:
@albertchae good call on FriendlyId's history feature– I actually used that initially, but I think I ran into some weird default behavior. I'll switch back over to it and see if I can remember what that behavior was (maybe it wasn't that bad)

app/views/events/_form.html.erb
maxwofford: Would that prevent an org from going back to their old slug? Just brainstorming here, but what about the following implementation:

- we _**do not**_ redirect donation pages
- we **do** redirect all other event pages
- we track domains that iframe donation pages (is this possible?) & list them in the "danger zone" dialogue

---

if we really want to redirect donation pages, maybe we can do this?

- the "danger zone" dialogue also shows a list of the previous slugs that redirect
- we only redirect the last X slugs (3? 5?) to prevent spam
- changing your slug more than X times in Y timespan sends a silent email to bank@hackclub.com for human inspection

app/controllers/concerns/set_event.rb
albertchae: Oh got it and since this is a shared concern you don't know which parameters are being used in each request right (since they can be different per controller using this concern)? Thinking about it more now, even if there are params permitting happening they can just be done after the redirect, so this seems great to me

app/views/events/_form.html.erb
albertchae: Is the idea behind not redirecting donation pages and redirecting all others to force users to take another step to realize what org they're donating to before doing so? That seems reasonable to me although it'll kind of complicate the `SetEvent` concern (or how it's used) 😅 

> we track domains that iframe donation pages (is this possible?) & list them in the "danger zone" dialogue

I think it could be possible if we listen to [`window.parent`](https://developer.mozilla.org/en-US/docs/Web/API/Window/parent) and report it somewhere, but it'd be hard to get this list live. This idea prompted me to dig around a little more though and apparently there's a content security policy flag we can add https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors. I wonder if this is something we can allow organizations to configure and then we'd add the header to their donation pages. Very out of scope for this PR btw

> if we really want to redirect donation pages, maybe we can do this?

Those ideas all sound reasonable to me, although would definitely complicate the implementation (but would be a good justification for using handrolled logic over the gem's built in feature)

albertchae:
@cjdenio Oh OK that makes sense how you knew about that table in the first place then 😆 Yeah I think either we're able to switch to it and not have to worry about maintaining our own implementation, or if you can identify the behavior differences we want from the handrolled implementation, we can document them somewhere for future readers.

cjdenio:
@albertchae Ah, I remembered the issue. FriendlyId also redirects numeric IDs, so https://bank.hackclub.com/6 would redirect to https://bank.hackclub.com/devhacks-test-event. Because redirection happens before authorization, this would let someone discover non-transparent organization slugs by brute-forcing sequential IDs and seeing where they redirect to.

albertchae:
@cjdenio ah ok, very interesting. I've seen enumeration attacks cause big problems at some organizations so it's good to be mindful of that. I tried both your latest commit and the one before and saw the issue you were describing, however I believe FriendlyId's support for numeric ids is kind of orthogonal to the history feature right? Ultimately I think this check `if @event.slug != id` ends up going down the redirect branch because a numeric id will never match the slug. So I have a couple suggestions here in no particular order:

1. Keep the current implementation but add a comment in `SetEvent` for why we aren't using the built in history feature
2. In `SetEvent`, if `id` is numeric, do not compare it against event.slug or try to redirect. This seems consistent with current production with transparent https://bank.hackclub.com/183/ vs non transparent https://bank.hackclub.com/133/, and also with the behavior of your latest commit on this PR [1485b44](https://github.com/hackclub/bank/pull/2714/commits/1485b449283467967fffc0f8bd74753e86435db0)
3. Try to add some basic authorization in `SetEvent` to check that current_user has permissions to access the event before doing any redirect.

cjdenio:
Thanks. I gave it some thought, and I'll switch it back to FriendlyId's module with #2 (the numeric ID checking) implemented.

maxwofford (commented):
I'm happy with the proposed implementation. Just tested locally and found weird results in a couple cases:

## Flipped priority on rename history

Made the following renames, in order:

1. Devhacks (`dev`) -> (`devhacks`)
2. Devhacks (`devhacks`) -> (`dev`)
3. HQ (`hq`) -> (`devhacks`)
4. HQ (`devhacks`) -> (`hq`)

I expect that bank.hackclub.com/devhacks would redirect to bank.hackclub.com/hq (more recent change), but instead redirects me to bank.hackclub.com/dev (earlier change).

## Missing initial slug

I renamed `hq` to `hq-hacks`, and expected the admin-tool of event slug history to mention `hq`, but it only had `hq-hacks`.

cjdenio:
Huh, I'll take a look at that first issue.

As for the second, you fix that by running
```ruby
Event.all.each { |event| FriendlyId::Slug.create(slug: event.slug, sluggable_id: event.id, sluggable_type: "Event") }
```
(it's in the description of this PR)

cjdenio:
okay so that priority thing is a bug (or weird feature) in FriendlyId. I'm switching back to the custom implementation once and for all (`implementation-v2-FINAL-real.doc` so to speak)

albertchae (approved):
Sounds good from my end. Thanks for working through the back and forth with some of these tricky cases!

maxwofford (approved):
LGTM!

cjdenio:
(Merging without CI because it seems to be stuck, and I know it'll pass)

PR #2717, opened 2022-06-24
V3 API

garyhtou:
https://bank.hackclub.com/docs/api/v3


Gemfile
```diff
@@ -187,6 +187,13 @@ gem "byebug", platforms: [:mri, :mingw, :x64_mingw]
 gem "chronic"
 gem "dry-validation"
 gem "geocoder" # lookup lat/lng for Stripe Cards shipment tracking
+gem "grape" # API v3
+gem "grape-swagger" # API v3
+# For Grape::Entity ( https://github.com/ruby-grape/grape-entity )
+gem 'grape-entity'
+gem 'grape-swagger-entity', '~> 0.3'
+# For representable ( https://github.com/apotonick/representable )
+gem 'grape-swagger-representable', '~> 0.2'
```
garyhtou: I think it might be okay to remove the `grape-swagger-representable` gem since we're going with `Grape::Entity` instead. Those two lines were added because of: https://github.com/ruby-grape/grape-swagger#model-parsers-

app/models/disbursement.rb
```diff
@@ -39,6 +39,9 @@ class Disbursement < ApplicationRecord
 
   has_paper_trail
 
+  include PublicIdentifiable
+  set_public_id_prefix :txf # Transfer
```
exu3: Thoughts on using `xfr`?

I feel like `txf` is too similar to `txn` but it's a small thing; I don't think it'd be a problem either way.

app/models/hcb_code.rb
```diff
@@ -18,6 +18,9 @@
 class HcbCode < ApplicationRecord
   has_paper_trail
 
+  include PublicIdentifiable
+  set_public_id_prefix :txn # TODO: should we prefix this as `txn` or `hcb`
```
exu3:
```suggestion
  set_public_id_prefix :txn
```

app/models/disbursement.rb
garyhtou: good idea!

config/application.rb
```diff
@@ -52,5 +52,8 @@ class Application < Rails::Application
 
     config.exceptions_app = routes
 
+    # TODO: Pre-load grape API
+    # ::API::V3.compile!
```
exu3: Should this be done before merging or is it alright to do this later?

garyhtou: would be nice to have before merge. don't spend too much time on that tho :)

PR #2718, opened 2022-06-24
Refactor partner_spec to use factorybot

albertchae:
Representative is required, so add it to the factory


albertchae:
This one is very small and straightforward so I will merge it now to unblock https://github.com/hackclub/bank/pull/2697 @maxwofford

PR #2724, opened 2022-06-30
Fix and refactor receipt_uploads_mailbox_spec

albertchae:
https://github.com/hackclub/bank/issues/2703

Best reviewed by commits, I fixed the issue by slowly converting the test to use factorybot instead of fixtures, and then also uncommented a skipped test by upgrading to a newer version of rspec-rails (which is still on RC, so maybe we don't want to do this).

- Use event factory instead of fixture
- Create minimal canonical_transaction factory
- Create minimal canonical_event_mapping factory
- Refactor receipt_uploads_mailbox_spec to use factories
- Upgrade rspec-rails to 6.0.0.rc1 and uncomment text broken by older version


maxwofford (approved):
Looks great! I'm not too experienced with running release candidate gems, will defer to your judgement there.

albertchae:
@maxwofford turns out we can use the previous version of rspec-rails (which is still newer than what was there before and not a release candidate). I was confused because the numbers don't match up exactly with rails versions

`rspec-rails 5 supports Rails 5.2 to 6.1`

PR #2725, opened 2022-06-30
Add presence validation for event.organization_identifier

albertchae:
I noticed this while adding the factory for events. We have a database
constraint on this column already but application validators tend to
have friendlier error messages. Unfortunately ActiveRecord does not have
a good way to avoid this duplication, so I would argue we try to add
missing validations everywhere or if are only going to pick one, the
database constraints are more valuable.

Some reading on this topic:
- https://thoughtbot.com/blog/validation-database-constraint-or-both
- https://naildrivin5.com/blog/2015/11/15/rails-validations-vs-postgres-check-constraints.html
- https://boringrails.com/tips/rails-check-constraints-database-validations
- https://bignerdranch.com/blog/dont-get-tangled-in-your-database-constraints/


maxwofford (approved):
tenor

I think both is the ideal situation for this particular example– it's an improvement for the user & the DB validation should stick around no matter what, but I don't think this will end up extending to all our presence validators; ex. our AASM models are using conditional validations which aren't (& shouldn't be) in the schema.

albertchae:
@maxwofford sounds good, I think I come across these most commonly when adding new factories, so I'll keep that in mind

PR #2726, opened 2022-06-30
User edit page has different title while onboarding

maxwofford:
Previously the title always said "Bank":
Screen Shot 2022-06-30 at 10 21 24

Now it says "Setup your account" / "Edit your account" depending on if the user is onboarding:
Screen Shot 2022-06-30 at 10 21 34


cjdenio (commented):
What about "Account settings" to better match the page header?

app/views/users/edit.html.erb
```diff
@@ -1,14 +1,11 @@
 <% if @onboarding %>
+  <% title "Setup your account" %>
   <% page_sm %>
 <% else %>
+  <% title "Edit your account" %>
```
exu3: Thoughts on?

```suggestion
  <% title "Account Settings" %>
```

exu3 (approved):
LGTM!

PR #2727, opened 2022-06-30
Tweak donation form to accomodate the header image

cjdenio:
Attempts to solve the contrast issues from before. Design feedback very, very welcome.

(also the stuff in this PR is not gated behind a feature flag)

<table>
<tr>
	<td> Before
	<td> After
<tr>
	<td> Screen Shot 2022-06-30 at 10 56 14 AM
	<td> Screen Shot 2022-06-30 at 10 56 31 AM

</table>


maxwofford (approved):
Happy to sit down and jam on design review, but maybe in the interest of shipping could you feature flag this & drop it in https://bank.hackclub.com/my/settings#feature-previews? Not the full donation page changes– just the editor. That way we can get people actually using this & see what sorts of images they use ASAP.

Marking approved for the code changes in the interest of getting this out for user feedback

cjdenio:
Talked with @maxwofford IRL– closing this until we see how many organizations end up uploading an image

PR #2728, opened 2022-06-30
Include dirty head and query count in footer

maxwofford:
Screen Shot 2022-06-30 at 11 18 44

Now the footer will tell you if running on a dirty HEAD (there are uncommitted changes in git) and your SQL query count (while running non-production).


exu3 (approved):
Neat!

PR #2729, opened 2022-06-30
Override the theme with `@dark`

cjdenio:
This PR makes it possible to override the UI theme with `@dark = true` or `@dark = false`.

### Reasoning

This'll eventually let users choose which theme they want to use for their embedded donation form, regardless of the donor's Bank preference


app/views/layouts/application.html.erb
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html data-dark="<%= !!@dark %>">
+<html data-dark="<%= !!@dark %>" <%= "data-ignore-theme" if !@dark.nil? %>>
```
maxwofford: Could we make this an `unless @dark.nil?`?

PR #2730, opened 2022-06-30
Show footer on nav'd pages

maxwofford:
Pages with a navbar would ignore footers. Now, they show up:

Screen Shot 2022-06-30 at 11 25 00


exu3 (approved):
I think it looks better without the footer, but I don't have very strong opinions here. Approving bc the code changes look good and work as intended.

Screen Shot 2022-07-01 at 11 49 45 AM

maxwofford:
@exu3 Totally agree it looks weird on that page w/o any invoices to put it at the bottom of the page. Do you have any thoughts about how we can fix that?

maxwofford:
Going to go forward with merging & we can figure something out to make those pages better in a future PR.

PR #2731, opened 2022-06-30
Prepare to encrypt ach_transfer#account_number and g_suite_accounts#initial_password with lockbox

albertchae:
https://github.com/hackclub/bank/issues/2541

Apparently we were already using lockbox to encrypt partner#stripe_api_key, but it was a pretty old version so the first commit is updating that to 1.0.0.

- Update lockbox to 1.0.0
Used `bundle update --conservative lockbox`
Changed old function to newer syntax `has_encrypted` (this was added to
avoid conflicting with the built in encryption feature with rails 7)

- Prepare to encrypt ach_transfer#account_number and g_suite_accounts#initial_password with lockbox
We need to add these ciphertext columns (Lockbox requires them to be
text `Note: Use a text column for the ciphertext in migrations,
regardless of the type`) and add `has_encrypted ..., migrating: true`.

After this change gets deployed, then in production console we will run
```
Lockbox.migrate(AchTransfer)
Lockbox.migrate(GSuiteAccount)
```

Afterwards we can ignore the old column, drop it, then delete the
ignored column

Similar PRs will be coming for other columns


maxwofford (approved):
Haven't looked too closely, but does this change [how we should be using papertrail](https://github.com/ankane/lockbox/#model-changes)?

I'm cool with the schema & the version bump though.

albertchae:
@maxwofford that's a good point about papertrail. I'll have to look into what happens with historical records, but I guess going forward what do we want to do? It looks like lockbox accounts for tracking ciphertext changes, but it also seems like these are fields we could skip entirely

maxwofford:
@albertchae I looked at this again and think I'm getting ahead of myself– was wondering how this would affect [the future "export fuzzed dev database" feature](https://github.com/hackclub/bank/issues/2712). Go for merge & migrate!

albertchae:
I think how this interacts with papertrail is still important but can be investigated as a follow up, so merging now 🚀

albertchae:
I finished running these commands on prod

```
Lockbox.migrate(AchTransfer)
Lockbox.migrate(GSuiteAccount)
```

Follow up PRs for next steps are coming!

PR #2734, opened 2022-06-30
API: Cache results from helpers + better date formatter

garyhtou:


app/api/api/v3.rb
```diff
@@ -171,7 +195,7 @@ def type_expansion(expand: [], hide: [])
             use :expand
           end
           get do
-            present transactions, with: Api::Entities::Transaction, **type_expansion(expand: %w[transaction linked_objects])
```
garyhtou: By default, we aren't expanding linked objects. However, they can be expanded by passing `linked_objects` or one of the types (`donation`, `ach_transfer`, etc.) via the `expand` query param

PR #2735, opened 2022-07-01
Add Pundit to the API

cjdenio:
Oh yeah.

This'll also make things easier when we add authentication.


app/policies/api/donation_policy.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Api
+  class DonationPolicy < ApplicationPolicy
+    def show?
+      record.event.is_public?
```
garyhtou: and not pending

garyhtou (approved):

exu3 (approved):

garyhtou (approved):

PR #2736, opened 2022-07-01
Hide `pending` donations from API

cjdenio:
`pending` represents unsubmitted donations, so it doesn't make sense to expose them through the API for privacy reasons (also, they're not shown in the normal user UI)


garyhtou:
good catch!

in the future, let's make sure we also restrict the donations you can get (Get a Single Donation endpoint shouldnt let you get a pending donation)

garyhtou (approved):

PR #2739, opened 2022-07-01
API: Override `amount_cents` for Transfer

garyhtou:
By default, linked objects use the HcbCode#amount_cents method. However, for Disbursements, this will always result in an amount_cents of 0 (zero) since there are two equal, by opposite, Canonical Transactions. Therefore, for the API, we are overriding the default amount_cents exposure defined in the LinkedObjectBase.


cjdenio (approved):
Thanks!

garyhtou:
Go ahead and merge on my behalf when ready

exu3 (approved):

garyhtou:
actually one sec

garyhtou:
I think we need to consider the context of the Event (we may need to negate the amount)

garyhtou:
I think we'll just need to leave this as a weird edge case for now. In some situations, we're missing the necessary context to evaluate whether the amount should be negative or positive.

If the org_id is in the path, then that's easy (if that org is the source event, then make it negative). However, when we're just getting a single transfer. there is no org_id in the path (we have no idea which event's context we're viewing this transfer from), so it's impossible to determine if it should be negative or positive. I think it's better to keep it consistent for now (always positive). However, if you'd like and have some time, feel free to try to tackle this issue. We're also running into a very similar issue with the UI (for Hcb Code show pages that belong to Disbursements)

garyhtou:
```ruby
          if hcb_code.disbursement?
            # Disbursement#amount will always be positive, however, from an
            # Event's transaction perspective, this amount may need to be
            # negated (if that Event is the source event for the disbursement).
            disbursement = hcb_code.disbursement
            org = options[:organization]

            is_source = org&.present? && disbursement.source_event == org
            sign = is_source ? -1 : 1
            amount = hcb_code.disbursement.amount * sign

            next amount
          end
```

PR #2742, opened 2022-07-04
Add index on stripe_transaction->card->>id on raw_stripe_transactions

albertchae:
Same as https://github.com/hackclub/bank/pull/2741
but for raw_stripe_transactions instead of raw_pending_stripe_transactions


maxwofford (approved):
LGTM!

PR #2749, opened 2022-07-05
Refresh on event cards page

exu3:
- The event cards page shows the cards in a grid (previously a `table`)
- The sunset background image is more subtle in dark mode (previously it was super bright in dark mode)

part of #2619

Screen Shot 2022-07-06 at 11 35 16 AM


maxwofford (approved):
Love the light/dark mode color change– that's great! The "your cards" section also really makes this readable. I'm approving review on the color change, but there are a couple other things I'd like to see for #2619 to be closed. Don't worry about covering this in your PR– just typing it here so I can link it to the ticket after.

- Orgs with a ton of cards (`/hq` are kinda long already, and the interactive cards list is even longer). I love how Finder on Mac does this Screen Shot 2022-07-06 at 12 37 57
  List files as icons:
  Screen Shot 2022-07-06 at 12 38 08
  List files in list:
  Screen Shot 2022-07-06 at 12 38 13
- The "cardholder" section is still a list. That makes sense for admins to see, because that correlates to `StripeCardholder`s. Does it make sense for regular users?
- Part of why the Emburse cards looked so good was because they had a good texture– wonder if we can do something fancy with the Stripe cards? Particularly virtual cards (what if we had some cool pixelation filter or something).

cjdenio (commented):
This is awesome! I had a few little comments:

- The "See all your cards" button feels very prominent and a little out of place. I wonder if it would look better as a normal button floating to the right of the "Your cards" heading.
- I think Max mentioned this already, but is the "cardholders" section still needed?
- There's a big empty space in Transparency Mode:
Screen Shot 2022-07-06 at 4 40 46 PM

app/views/events/card_overview.html.erb
```diff
@@ -32,51 +32,49 @@
     <%= render "stripe_cards/form", stripe_card: StripeCard.new %>
   </section>
 
-  <% if @session_user_stripe_cards.blank? and @stripe_cards.blank? %>
-    <%= blankslate "You haven’t requested any cards yet." %>
-  <% end %>
-
   <% if signed_in? %>
```
cjdenio: I think this would be better as `organizer_signed_in?`

cjdenio (approved):
Thanks, looks great to me!

---

I just pushed up small fixes to these bugs:

https://user-images.githubusercontent.com/34525547/178162380-da12f91d-0361-4eeb-9446-0f159b22c705.mov

Screen Shot 2022-07-10 at 5 13 52 PM
!

exu3:
Thanks for the fix!

PR #2750, opened 2022-07-05
Transactions can be marked as unmappable

cjdenio:
@MelSmith104 mentioned she wished there was a way to hide certain unmapped transactions from the ledger, mostly old ones from 2021. This PR lets admins mark transactions as "unmappable", meaning they can be left unmapped while not showing up on the unmapped ledger.


maxwofford (approved):
This looks great, thanks for adding it! Marking approved for the code changes, but I'd like to sanity check with @MelSmith104 & Bence that this change is going to be handled by our bookkeeping side of things, so please only merge once she approves.

garyhtou:
Couple of things to note:

- Some unmappable transactions were historically mapped to Event `999` or Clearinghouse (@MelSmith104 might have a better idea of where some existing transactions sit)
- CanonicalPendingTransactions use SettledMapping and DeclinedMapping. This isn't exactly the same, but it's another design choice to consider (I think what you have in this PR is fine!)
- `mark_foo` is kinda synonymous with `AASM` in our code base, it would be nice if we kept the `mark` word for AASM-related processes/states

MelSmith104:
this looks good to me! thank you so much for building this :) i spoke with max and i think the only thing that might be added is that unmappable transactions end up in event 999. also nothing breaks for bookkeeping as these transactions are all from 2021

cjdenio:
Ah, I didn't know about event 999. The button now just maps the transaction there.

image

garyhtou (approved):

garyhtou (approved):

maxwofford:
@garyhtou does anything prevent these from being pulled out of 999 NoEvent during a nightly mapping?

garyhtou:
> @garyhtou does anything prevent these from being pulled out of 999 NoEvent during a nightly mapping?

I believe the tx engine's mapping logic won't touch anything that's already been mapped

PR #2760, opened 2022-07-05
Move Stripe declined email to v2 engine

cjdenio:
This PR, well, moves the Stripe declined email to the v2 engine. All tested locally.

Closes #2647!


maxwofford (approved):
Awesome, thanks for adding this! Approving the code changes.

I'd be curious to see the primary reason charges are flagged as fraud– is it really CVC or something like wrong zip code?

cjdenio:
Looking through the authorization list, it looks like most are invalid CVC. The wording in the email should definitely mention address/zip code though

PR #2761, opened 2022-07-05
Instantly decline Stripe transactions

cjdenio:
Closes #2646.


app/services/stripe_authorization_service/create_from_webhook.rb
```diff
@@ -22,6 +22,11 @@ def run
 
         # 4. idempotent map to event
         ::PendingEventMappingEngine::Map::Single::Stripe.new(canonical_pending_transaction: cpt).run
+
+        # 5. instantly mark the transaction as declined if it was declined on Stripe's end
+        unless remote_stripe_transaction.approved
```
maxwofford: What are we using `remote_stripe_transaction.approved` for? Is there a `rejected`? Are there other states ie. `refunded`? Or is `approved` the only boolean for this?

maxwofford (approved):
Thanks for this improvement! Let's def either ship this or include it in a future changelog.

Approving the code changes, but left a comment about clarity.

app/services/stripe_authorization_service/create_from_webhook.rb
cjdenio: `approved` is just a boolean indicating whether the authorization has been approved or declined, either on our end or Stripe's end. I think everything else comes later, after the merchant captures the authorization (or doesn't)

cjdenio:
@maxwofford is this good to merge?

PR #2764, opened 2022-07-06
Birthday flavor text

maxwofford:


app/helpers/static_pages_helper.rb
```diff
@@ -350,6 +353,21 @@ def flavor_text
     end
   end
 
+  def birthday_flavor_text
+    return nil unless current_user.birthday.present? &&
+                      current_user.birthday.to_date == Date.today &&
+                      rand(100) <= 10
```
cjdenio: Could this be given a higher chance of happening? I'm interpreting this as a 1/10th chance of seeing a birthday message

app/helpers/static_pages_helper.rb
```diff
@@ -350,6 +353,21 @@ def flavor_text
     end
   end
 
+  def birthday_flavor_text
+    return nil unless current_user.birthday.present? &&
```
cjdenio: the `user_birthday?` helper also exists

app/helpers/static_pages_helper.rb
maxwofford: increasing!

app/helpers/static_pages_helper.rb
maxwofford: Refactored!

maxwofford: What's that provided by? I can't find a `def user_birthday?`

cjdenio (approved):
Thanks!

app/helpers/static_pages_helper.rb
cjdenio: It's in `users_helper.rb`, on the `cjdenio/party-mode` branch

PR #2767, opened 2022-07-07
Admin: filter HCB codes by 'has receipt?'

cjdenio:


cjdenio:
Actually, hang on– I'm gonna make some changes

exu3 (commented):
Would it make sense to put this filter on the Admin Ledger page rather than the HCB CODES page? The HCB CODES page has a lot of irrelevant stuff (eg. unsubmitted donations); I'm not sure it's very useful given the Ledger page exists.

cjdenio:
Hmm, I think it makes more sense to have it here because receipts are uploaded to HCB codes.

exu3 (approved):
LGTM!

PR #2770, opened 2022-07-07
Edits to the Transfers Tab and Transfer Page

kunalbotla:
this PR makes edits to the transfers tab and disbursements page

| | old | new |
| --- | --- | --- |
| renames the "Send" tab to "Transfers" to follow the same types of descriptions as the other tabs and match the page header. | image | image |
| consolidates transfers (disbursements) into the "Send" dropdown. previously there was a button and an option in the dropdown. | image | image |
| change heading and field titles to be more similar to ACH and better reflect the instant nature of transfers | image image image | image image image |


MelSmith104:
this looks great! i love the clarification this adds :)

MelSmith104 (commented):
LGTM!

maxwofford (approved):
This looks great! Thanks for the change & the clear PR description.

Once point of confusion is that we now have a specific feature called Transfers– the `Disbursement` model. What if we kept the tab called "Send", and updated the page header to something like "Send funds".

Your call on wording. The code changes are good, so I'm marking this as approved.

kunalbotla:
I thought about that, but I think there's enough clarity on the page itself and with the hover description about what each type of transfer does and the page so that it's not confusing. making it `Transfers` instead of `Send` makes it more similar to `Invoices`, `Cards`, etc

PR #2794, opened 2022-07-12
Users can toggle between list/grid view on cards page

exu3:
This PR adds a toggle on the cards page for events.

Screen Shot 2022-07-12 at 9 46 52 AM


cjdenio (approved):
This looks great; could you just add a little ✅ icon or something to the selected view option?

PR #2795, opened 2022-07-12
Instant Deposits

garyhtou:
Donations and Invoices (and also Disbursement) will now be instantly deposited into organization’s balances after being paid. their transactions will show up as as settled on the event ledger.

---

### TODO:

- [x] List
  - Two services
- [x] Modifying Balance
  - In `Event` model
- [x] UI
  - "Your funds will arrive in ..."
  - mailer
- [x] Unfronting transactions
  - [x] Unfronting a single pending transactions
  - [x] Disabling fronting for an entire org
- [x] Automatically front invoices and donations
- [x] Listen to webhooks for refunds
- [x] Fix association preloader
- [x] Also front disbursement using the same system
- [x] Anomaly (nice to have, not required)
  - Detect if Hcb Code isn't fully settled within 2 weeks
- [x] Instantly create pending transaction for donations
- [x] Instantly create pending transaction for invoices
- [ ] Actual ship post

### TEST for after merging

- Try making a card charge (test 2 sec respond time to stripe auth)
- Make a donation
- Make an invoice


app/services/pending_transaction_engine/canonical_pending_transaction_service/import_single/donation.rb
```diff
@@ -0,0 +1,32 @@
...
+        end
+
+        def run
+          return existing_canonical_pending_transaction if existing_canonical_pending_transaction
+
+          attrs = {
+            date: @raw_pending_donation_transaction.date,
+            memo: @raw_pending_donation_transaction.memo,
+            amount_cents: @raw_pending_donation_transaction.amount_cents,
+            raw_pending_donation_transaction_id: @raw_pending_donation_transaction.id
```
garyhtou: do we need `fronted: true` here?

cjdenio: oops, good catch

garyhtou:
Hey @albertchae would you mind taking a look at the association preloader in this PR?

We're adding some CanonicalPendingTransactions to `TransactionGroupingEngine::Transaction::All`. I tried adding some additional preloading and would love your eyes on it.

I also attempted some preloading in `Event#fronted_incoming_balance_v2_cents`, however ran into an issue:
https://github.com/hackclub/bank/blob/3dc7ffe65f020e7160929263f2ddd987fafbd882/app/models/event.rb#L389-L397

app/services/transaction_grouping_engine/transaction/association_preloader.rb
```diff
@@ -80,6 +85,10 @@ def preload_associations!
                                      .slice(*t.canonical_transaction_ids)
                                      .values
                                      .sort { |ct1, ct2| self.class.compare_date_id_descending(ct1, ct2) }
+          t.canonical_pending_transactions = canonical_pending_transactions_by_id
+                                             .slice(*t.canonical_transaction_ids)
```
albertchae: should this be `*t.canonical_pending_transaction_id`? I believe that's the id that this hash is keyed by and the id you queried for here https://github.com/hackclub/bank/pull/2795/files#diff-6b8a837914a31f99b84a9acd9db1cb86388dd70b3a086ae67720acaa185a7f9fR37

app/services/transaction_grouping_engine/transaction/association_preloader.rb
```diff
@@ -31,12 +31,17 @@ def preload_associations!
         canonical_transactions_by_id = canonical_transactions.index_by(&:id)
         canonical_transaction_ids = canonical_transactions.pluck(:id)
 
-
-        canonical_pending_transactions_by_hcb_code = CanonicalPendingTransaction.where(hcb_code: hcb_code_codes).group_by(&:hcb_code)
+        canonical_pending_transactions = CanonicalPendingTransaction
+                                         .where(hcb_code: hcb_code_codes)
+                                         .or(
+                                           CanonicalPendingTransaction.where(id: @transactions.flat_map(&:canonical_pending_transaction_ids))
+                                         )
```
albertchae: This looks reasonable but IIRC I added this OR because [`hcb_codes` are lazily created for `CanonicalTransaction`](https://github.com/hackclub/bank/pull/2795/files#diff-6b8a837914a31f99b84a9acd9db1cb86388dd70b3a086ae67720acaa185a7f9fR22-R23) and I think you told me that they are eagerly created for `CanonicalPendingTransaction`? So maybe we add making the `canonical_pending_transactions_by_id` hash but without this OR?

I don't have as much context with the rest of this PR though so it's possible I missed another reason why this is needed, so please let me know 🙏

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -46,16 +47,43 @@ def canonical_transaction_ids
...
         return "" unless @search.present?
 
-        "and (ct.memo ilike '%#{@search}%' or ct.friendly_memo ilike '%#{@search}%' or ct.custom_memo ilike '%#{@search}%')"
+        type = type.to_s
+        "and (#{type}.memo ilike '%#{@search}%' or #{type}.friendly_memo ilike '%#{@search}%' or #{type}.custom_memo ilike '%#{@search}%')"
       end
 
       def canonical_transactions_grouped
-        group_sql = <<~SQL
+        pt_group_sql = <<~SQL
```
albertchae: I would say this is out of the scope of this PR especially since it was already a raw SQL query, but it might be worth looking into refactoring this query into a query object as a follow up https://thoughtbot.com/blog/a-case-for-query-objects-in-rails

app/models/event.rb
```diff
@@ -367,14 +368,36 @@ def settled_outgoing_balance_cents(start_date: nil, end_date: nil)
...
+        canonical_transactions_by_hcb_code = canonical_transactions.group_by(&:hcb_code)
+
+        canonical_pending_transactions = CanonicalPendingTransaction.where(hcb_code: hcb_codes)
+        canonical_pending_transactions_by_hcb_code = canonical_pending_transactions.group_by(&:hcb_code)
+
+        pts.each do |pt|
+          pt.local_hcb_code = hcb_code_by_code[pt.hcb_code]
+
+          # TODO: using group_by converts the AR Collection into an array.
+          # Existing code depends on it being a collection.
```
albertchae: Yes, this is a big problem with the manual/explicit preloading strategy I used in `AssociationPreloader` and why it might not be a good idea to use it outside of the events#show page 🤔 

A couple ideas here:

- Do you remember any specific examples where an array was expected vs the AR relation? If there are only a few maybe it'd be easy to refactor them into working with either.
- One reason I went with the manual approach is that there seem to be several "fake" associations in the bank codebase where instead of using an AR has_many/belongs_to etc we use an instance method with an instance variable that queries for what we want. I've been wanting to unwind some of those and replace them with vanilla AR associations so we can just use `includes` or `preloads` etc instead of all the logic in `AssociationPreloader` to slice things up and attach the manually. I believe doing this for `canonical_transactions` and `canonical_pending_transactions` on `HcbCode` should be relatively straightforward. I have a pending branch where I was messing around with this a bit so I can try to clean that up and see if you can incorporate into yours here, so I'll ping you when that is ready.

albertchae: Ok I think this https://github.com/hackclub/bank/pull/2818 should do it – it will give us preloading but keep it an AR relation so we chain whatever else we need to do for your use case

app/services/transaction_grouping_engine/transaction/association_preloader.rb
garyhtou: good catch!

app/models/event.rb
garyhtou: thank you! #2818 would be amazing. I'll be removing the "manual" preloading for `Event#fronted_incoming_balance_v2_cents`. I think we're hoping to merge in this PR (#2795) sometime today.

Sometime after this is merged, would you mind helping preload `Event#fronted_incoming_balance_v2_cents` using the associations created in #2818 and #2819?

app/services/event_service/create.rb
```diff
@@ -47,7 +47,8 @@ def attrs
         expected_budget: 100.0,
         point_of_contact_id: @point_of_contact_id,
         partner_id: partner.id,
-        organization_identifier: organization_identifier
+        organization_identifier: organization_identifier,
+        can_front_balance: true
```
maxwofford: Should this be `false` until shipped for all users?

app/views/hcb_codes/_invoice.html.erb
```diff
@@ -268,6 +229,17 @@
       </section>
       <%# if @invoice.payment_method_type == 'ach_credit_transfer' %>
     <% end %>
+
+    <% unless @invoice.deposited? %>
+      <section class="card__banner card__darker secondary border-top">
+        <p class="my0 italic">
+          The funds from this invoice payment are immediately reflected in your account balance.
+          (<a href="https://hack.af/important">learn more</a>)
```
maxwofford: This is a 404.

app/views/invoice_payouts_mailer/notify_organizers.html.erb
```diff
@@ -8,7 +8,7 @@
 </p>
 
 <p>
-  The funds (<strong><%= render_money @payout.invoice.item_amount %></strong>) will be available on approximately <strong><%= @payout.arrival_date.to_date %></strong>.
+  The funds (<strong><%= render_money @payout.invoice.item_amount %></strong>) are now available to spend.
```
maxwofford: Does this work if an event doesn't have fronted balances enabled?

maxwofford (approved):
Great! Love how simple the change is. Left a couple comments. tl;dr, it seems like this gets enabled for new users & not all changes are feature-flagged. Would prefer that this only affects events we enable it after merging.

app/views/hcb_codes/_invoice.html.erb
exu3: Thinking of linking the changelog post there once that's written. Right now, it's hack.af/vault-incident.

app/services/event_service/create.rb
garyhtou: https://github.com/hackclub/bank/pull/2795/commits/672cc4d3af862c17f792e15de9c226f0e8bd307a

app/views/hcb_codes/_invoice.html.erb
garyhtou: Will link to change log post once we have that!
https://github.com/hackclub/bank/pull/2795/commits/515c78ecb90d484c78529a28442422d8447f9d1f

app/views/invoice_payouts_mailer/notify_organizers.html.erb
exu3: Thanks for catching , just feature-flagged this here https://github.com/hackclub/bank/pull/2795/commits/73c492891e372d242587979385470db46ff91564#diff-30a8b0529ec68a4ab8badeb44c023ae962022ecc0b9ecab3cf91d490f86e20f1

app/models/disbursement.rb
```diff
@@ -172,7 +180,8 @@ def state_text
   end
 
   def state_icon
-    "checkmark" if fulfilled? || processed?
+    "checkmark" if fulfilled?
+    "checkmark" if processed? && @destination_event.can_front_balance?
```
garyhtou:
```suggestion
    "checkmark" if fulfilled? || processed? || (pending? && @destination_event.can_front_balance?)
```

app/services/pending_transaction_engine/canonical_pending_transaction_service/import/donation.rb
```diff
@@ -6,17 +6,7 @@ module Import
       class Donation
         def run
           raw_pending_donation_transactions_ready_for_processing.find_each(batch_size: 100) do |rpoct|
-
-            ActiveRecord::Base.transaction do
```
garyhtou: any idea why this was wrapped in a transaction?

garyhtou:
very basic anomaly detection: https://bank.hackclub.com/blazer/queries/272-unsettled-fronted-hcb-code

app/models/event.rb
albertchae: Yep! I'll take care of following up in `Event#fronted_incoming_balance_v2_cents` after those PRs get merged

PR #2798, opened 2022-07-13
Hack The Bank!

sampoder:
Assemble Easter egg


cjdenio (commented):
Nice! To make this work, you'll want to add a rule to `config/routes.rb`:
```ruby
  get "4e7thoir7eyt7ae5i7tygaer-cash-export", to: "static_pages#4e7thoir7eyt7ae5i7tygaer-cash-export"
```

(you could also rename the view to `cash-export-easter-egg.html.erb` to make it something like
```ruby
  get "4e7thoir7eyt7ae5i7tygaer-cash-export", to: "static_pages#cash-export-easter-egg"
```
)

sampoder:
thanks Caleb!

maxwofford:
@sampoder You'll also want_to_use_underscores instead of dashed-filenames-and-methods

PR #2799, opened 2022-07-13
Get Bank's commit hash through the API

exu3:
This PR adds an endpoint for getting the commit hash and time of the latest build accessible at `/api/v3/version`.


garyhtou (approved):
Thoughts on changing it to `/api/v3/git`?

PR #2802, opened 2022-07-13
Fix floating point parsing on disbursement creation

maxwofford:
Fixes #2801. Tested by creating a $2.07 disbursement successfully in dev.


albertchae (approved):
Seems like it might be worth trying to find all the `to_f`'s or similar float related stuff and changing to BigDecimal/Money

PR #2804, opened 2022-07-13
Fix floating point parsing on check creation

maxwofford:
Fixes #2803. This was tested in development and works while creating a check for $2.07.


app/services/check_service/create.rb
```diff
@@ -6,21 +6,21 @@ class Create
...
-                   payment_for:, memo:, amount_cents:, send_date:,
+                   payment_for:, memo:, amount:, send_date:,
                    current_user:)
       @event_id = event_id
       @lob_address_id = lob_address_id
 
       @payment_for = payment_for
       @memo = memo
-      @amount_cents = amount_cents
+      @amount = amount
```
albertchae: small nit: I do like the `_cents` suffix but I get why you did it since the helper method you added to parse it is now called `amount_cents`. Looks like the param from the frontend is called `amount` but should we maybe call it `amount_cents_string`? although that is a bit cumbersome 🤷

albertchae (approved):

app/services/check_service/create.rb
albertchae: renamed to `amount_cents_string`

albertchae:
@maxwofford I took over this PR per our slack conversation. I added a test as well. Can you take a quick look at the last 2 commits and then I'll merge?

maxwofford:
Thanks for adding the tests! GitHub won't let me review my own PR, but this comment is an approval. It'll be nice once we can just switch out all the lob address stuff for a factory, but makes sense that's out of the scope of this PR.

albertchae:
@maxwofford yep agreed, I got burned by lob address factory on another change too. It might be time to really invest in fixing it up...

PR #2807, opened 2022-07-14
Qualify created_at column with table by using hash syntax

albertchae:
Fixes https://github.com/hackclub/bank/issues/2789

Previous syntax generated this query
```
SELECT "users".* FROM "users" INNER JOIN "organizer_positions" ON "users"."id" = "organizer_positions"."user_id" WHERE "organizer_positions"."deleted_at" IS NULL AND "organizer_positions"."event_id" = 183 ORDER BY created_at desc LIMIT 100 OFFSET 0
```

New syntax specifies `created_at` on users
```
SELECT "users".* FROM "users" INNER JOIN "organizer_positions" ON "users"."id" = "organizer_positions"."user_id" WHERE "organizer_positions"."deleted_at" IS NULL AND "organizer_positions"."event_id" = 183 ORDER BY "users"."created_at" DESC LIMIT 100 OFFSET 0
```

I'm not entirely sure why this only happened now as I don't see what
changed about these models recently that could have triggered this


maxwofford (approved):
That's funny– thanks for the fix on this though!

PR #2814, opened 2022-07-15
Fix common document deletion

cjdenio:
Fixes #1806


exu3 (approved):
Nice, thanks!

Thoughts on also flashing a success message after the document is deleted?

PR #2816, opened 2022-07-15
Webpacker -> `jsbundling-rails`

cjdenio:
Webpacker was [retired](https://github.com/rails/webpacker#webpacker-has-been-retired-) as part of the Rails 7 release . This somewhat convoluted PR makes the switch over to vanilla Webpack + [jsbundling-rails](https://github.com/rails/jsbundling-rails), the newly recommended way of using a bundler.

The big difference is that you have to run `./bin/dev` instead of `rails server`.

And React still works!


webpack.config.js
```diff
@@ -0,0 +1,34 @@
...
+        exclude: /node_modules/,
+        use: {
+          loader: 'babel-loader',
+          options: {
+            presets: ['@babel/preset-react']
+          }
+        }
+      }
+    ]
+  },
```
albertchae: Looks like this replaces the config in `babel.config.js` (which you're deleting in this PR) to support react, is that correct?

albertchae: Do we need the second step from here in `package.json` https://github.com/rails/jsbundling-rails/blob/main/docs/switch_from_webpacker.md#optional-babel ?

app/javascript/application.js
```diff
@@ -15,14 +15,14 @@
 // const imagePath = (name) => images(name, true)
```
albertchae: It looks like lines 2-15 are leftover comments from the webpacker install. Should we remove those since pack files aren't a thing aanymore?

app/javascript/application.js
```diff
@@ -15,14 +15,14 @@
...
 
 // Support component names relative to this directory:
-var componentRequireContext = require.context('components', true)
+var componentRequireContext = require.context('./components', true)
```
albertchae: resolves
```
ERROR in ./app/javascript/application.js 18:30-65
Module not found: Error: Can't resolve 'components' in '/usr/src/app/app/javascript'
```

app/javascript/application.js
```diff
@@ -15,14 +15,14 @@
...
 import ReactRailsUJS from 'react_ujs'
 ReactRailsUJS.useContext(componentRequireContext)
 
 ReactRailsUJS.handleEvent('turbo:load', ReactRailsUJS.handleMount)
 ReactRailsUJS.handleEvent('turbo:before-render', ReactRailsUJS.handleUnmount)
 
-import 'controllers'
+import './controllers'
```
albertchae: resolves
```
ERROR in ./app/javascript/application.js 25:0-20
Module not found: Error: Can't resolve 'controllers' in '/usr/src/app/app/javascript'
Did you mean './controllers'?
```

webpack.config.js
```diff
@@ -0,0 +1,34 @@
+const path = require('path')
+const webpack = require('webpack')
+
+module.exports = {
+  mode: 'production',
+  devtool: 'source-map',
+  entry: {
+    bundle: './app/javascript/application.js'
```
albertchae: are you renaming this to `bundle` to avoid a collision with https://github.com/hackclub/bank/blob/main/app/assets/javascripts/application.js ?

app/javascript/controllers/index.js
```diff
@@ -5,5 +5,5 @@ import { Application } from '@hotwired/stimulus'
 import { definitionsFromContext } from '@hotwired/stimulus-webpack-helpers'
 
 const application = Application.start()
-const context = require.context('controllers', true, /_controller\.js$/)
+const context = require.context('.', true, /_controller\.js$/)
```
albertchae: resolves
```
18:02:02 js.1   | ERROR in ./app/javascript/controllers/index.js 7:16-72
18:02:02 js.1   | Module not found: Error: Can't resolve 'controllers' in '/usr/src/app/app/javascript/controllers'
18:02:02 js.1   | resolve 'controllers' in '/usr/src/app/app/javascript/controllers'
```

webpack.config.js
albertchae: Hmm on further investigation, it looks like the `devDependencies` section might handle that https://github.com/hackclub/bank/pull/2816/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R25-R27

webpack.config.js
albertchae: I see, it looks like we end up with this error if not

```
Sprockets::DoubleLinkError in Users#auth

Showing /usr/src/app/app/views/layouts/application.html.erb where line # raised:

Multiple files with the same output path cannot be linked ("application.js")
In "/usr/src/app/app/assets/config/manifest.js" these files were linked:
  - /usr/src/app/app/assets/javascripts/application.js
  - /usr/src/app/app/assets/builds/application.js
  ```

**Review by albertchae — APPROVED** [review]
This looks great! Looks like it was mostly following https://github.com/rails/jsbundling-rails/blob/main/docs/switch_from_webpacker.md and then resolving errors. I left comments on some changes so other people can follow along if they're curious. @maxwofford we should figure out a good time to merge and deploy this since it'll be somewhat risky and we also want to do it at a time when we don't have to keep resolving merge conflicts.

An interesting thing I noticed is that trying to remove the line in assets.rb here https://github.com/rails/jsbundling-rails/blob/main/docs/switch_from_webpacker.md#2-remove-webpacker
```
# Remove from your config/initializers/assets.rb
-# Add Yarn node_modules folder to the asset load path.
-Rails.application.config.assets.paths << Rails.root.join('node_modules')
```
doesn't work because it looks like we're pulling those in here still https://github.com/hackclub/bank/blob/main/app/assets/javascripts/application.js? So it looks like bank is still using some mix of sprockets and jsbundling?

Another thing I noticed in the migration guide was this section on setting up development vs production mode, which I think is out of scope of this PR but is something we should consider https://github.com/rails/jsbundling-rails/blob/main/docs/switch_from_webpacker.md#optional-add-support-for-development-environment

maxwofford:
We're aiming to merge this in by Thursday (Apr 22nd) & holding off on merging other PRs that may cause merge conflicts until then

albertchae:
@maxwofford sounds good, I pushed up a recent merge with main. Only issue is that `yarn.lock` has a conflict but `yarn install` knows how to resolve git conflicts which is nice.

PR #2817, opened 2022-07-19
Revert "Remove non preloaded version of not_admin_only_comments_count"

albertchae:
Reverts hackclub/bank#2743

This is the cause of https://github.com/hackclub/bank/issues/2812. Reverting for now to fix the immediate problem and I'll revisit how we can remove this and not break counts on the receipts page


albertchae:
@cjdenio added you as a reviewer as an FYI

PR #2818, opened 2022-07-19
Use vanllla has_many AR association from HcbCode to CanonicalTransaction

albertchae:
This PR shows how to take what I'm calling a "manual association" i.e. an instance method that does a query manually and then memoizes to an instance variable and replace it with an ActiveRecord association and how that enables us to use standard AR preloading techniques like `includes`. I'm not 100% sure it's possible but my goal would be to get to the point where we don't even need the `AssociationPreloader` class anymore because everything is just done via [`includes` or `preloads` or etc](https://www.bigbinary.com/blog/preload-vs-eager-load-vs-joins-vs-includes).

**Best reviewed by commits**. This shows the progression and how the query counts change depending on how much of the migration we do. I diffed the generated HTML from `/hq` at each step and it was the same (except for query count, flavor text and other random elements)


garyhtou (approved):
thanks for making this so easy to follow!

PR #2820, opened 2022-07-19
Small footer (& header) changes

exu3:
* Moved 'Feedback' to the footer and added a link to `my/inbox` in the header partial based on feedback from [this email](https://secure.helpscout.net/conversation/1946042692/4526/)
* Added small hover detail to the footer https://hackclub.slack.com/archives/C1C3K2RQV/p1658158617367619


cjdenio (changes requested):
This is great! Two comments:

- It looks like the links aren't transitioning at the same rate as the rest of the footer when un-hovering:

https://user-images.githubusercontent.com/34525547/179764897-685c40af-0e00-42c7-8123-95b33ec88374.mov

- I think maybe the "Feedback" link should say "Leave feedback", but that's just random personal preference.

I'm just requesting changes for the first item.

cjdenio (approved):
Thanks!

PR #2822, opened 2022-07-19
Fix footer for users without events

cjdenio:
- Footer (& header) changes
- "Share Feedback"
- Fix transition
- Fix footer for users without events

Fixes #2821


exu3:
Thanks 🙌

PR #2826, opened 2022-07-20
Cache organizer_signed_in in sessions_helper by event

albertchae:
I noticed we were making several calls to this method

```
  CACHE User Exists? (0.0ms)  SELECT 1 AS one FROM "users" INNER JOIN "organizer_positions" ON "users"."id" = "organizer_positions"."user_id" WHERE "organizer_positions"."deleted_at" IS NULL AND "organizer_positions"."event_id" = $1 AND "users"."id" = $2 LIMIT $3
  ↳ app/helpers/sessions_helper.rb:56:in `organizer_signed_in?'
```

Total for the page
```
Completed 200 OK in 1404ms (Views: 969.5ms | ActiveRecord: 110.6ms | SQL Queries: 153 (116 cached) | Allocations: 564102)
```

After adding memoization
```
Completed 200 OK in 749ms (Views: 439.3ms | ActiveRecord: 84.9ms | SQL Queries: 39 (2 cached) | Allocations: 382730)
```

The memoize by hash pattern was taken for #4 here
https://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/


app/helpers/sessions_helper.rb
```diff
@@ -53,7 +53,10 @@ def current_user=(user)
   end
 
   def organizer_signed_in?(event = @event)
-    (signed_in? && event&.users&.include?(current_user)) || admin_signed_in?
+    @organizer_signed_in ||= Hash.new do |h, event_key|
+      h[event_key] = (signed_in? && event_key&.users&.include?(current_user)) || admin_signed_in?
+    end
+    @organizer_signed_in[event]
```
albertchae: I guess I could also write this as

```
@organizer_signed_in ||= {}
@organizer_signed_in[event] ||= (signed_in? && event_key&.users&.include?(current_user)) || admin_signed_in?
```

maxwofford (approved):
I'm cool with either way of writing it. Thanks for caching this!

PR #2829, opened 2022-07-20
Temporarily track how many missing hcb_codes we create

albertchae:
I want to verify that this count is regularly 0 for `CanonicalPendingTransaction`
because we are already eagerly creating `HcbCode`s for them, and i also want to
track this number for `CanonicalTransaction` to get a sense for how many there are
and to see if it drops to 0 after we eagerly create `HcbCode`s there too.

https://github.com/hackclub/bank/issues/2783


app/services/transaction_grouping_engine/nightly.rb
```diff
@@ -9,21 +9,37 @@ def initialize(start_date: nil)
...
         ct.update_column(:hcb_code, hcb_code)
 
         HcbCode.find_or_create_by!(hcb_code: hcb_code)
+
+        canonical_transaction_missing_hcb_count += 1
       end
+      Ahoy::Event.create!({
+                            name: "canonicalTransactionMissingHcbCount",
+                            properties: { count: canonical_transaction_missing_hcb_count }
+                          })
```
albertchae: This is my first time using ahoy so I want to sanity check I'm not doing something wrong. I tried running this job a few times locally and saw the events getting created

garyhtou: If you'd like, feel free to Airbrake them. either works tho

```ruby
Airbrake.notify("There were #{canonical_transaction_missing_hcb_count} CTs missing HcbCode")
```

albertchae: nice, I think I want to merge the ahoy events here and have it run for about a week to see the historical trend in the database, then airbrake will be good after we add eager creation to `CanonicalTransaction`s too since it should be a sign something is wrong or unexpected

maxwofford (approved):
This looks fine! Seems pretty straight forward to clear them out once we're done with them too.

albertchae:
As of a db pull from today
```
Ahoy::Event.where(name: "pendingTransactionMissingHcbCount").count
   (2.3ms)  SELECT COUNT(*) FROM "ahoy_events" WHERE "ahoy_events"."name" = $1  [["name", "pendingTransactionMissingHcbCount"]]
=> 567
```
All of these are at count 0


But for `CanonicalTransaction`, we see roughly 173-190. It looks like the query for these in the nightly job looks for [unknown or missing from 1 month ago](https://github.com/hackclub/bank/blob/cb920faf0e1055ed90bd496d6ab11e088501e71f/app/services/transaction_grouping_engine/nightly.rb#L32), so it seems ~173 are "stuck" in an unknown state.

PR #2830, opened 2022-07-20
Hotfix for transaction searchbar

cjdenio:
Fixes #2828


cjdenio:
Gonna merge this so the searchbar (kinda) starts working again. @garyhtou, I'd love your thoughts on the best way to fix this

PR #2837, opened 2022-07-21
Notify organizers on invoice paid webhook

exu3:
duck sweat


app/controllers/stripe_controller.rb
```diff
@@ -82,6 +82,9 @@ def handle_invoice_paid(event)
     # Mark invoice as paid
     InvoiceService::OpenToPaid.new(invoice_id: invoice.id).run
 
+    # Email organizers
+    InvoiceService::NotifyOrganizers.new(invoice: invoice).run
```
cjdenio: Ideally this would be run inside the `OpenToPaid` service, which also runs periodically (just in case we miss a webhook from Stripe)

app/mailers/invoice_payouts_mailer.rb
```diff
@@ -2,13 +2,13 @@
 
 class InvoicePayoutsMailer < ApplicationMailer
```
cjdenio: I guess this would be `InvoiceMailer` now

app/services/invoice_service/notify_organizers.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module InvoiceService
```
cjdenio: Hmmmmmmm, does this need to be a separate service?

garyhtou: i guess not really needed. either way is fine

cjdenio:
Looks like https://github.com/hackclub/bank/blob/main/app/views/invoice_payouts_mailer/notify_organizers.html.erb needs to updated from `@payout.invoice` to `@invoice`

app/mailers/invoice_payouts_mailer.rb
exu3: Hmm `InvoiceMailer` seems to already exist; should I edit that or maybe rename this one to `InvoicePaidMailer`?

cjdenio: Ah, I think I'd add it to `InvoiceMailer` instead

app/views/invoice_mailer/notify_organizers.html.erb
```diff
@@ -0,0 +1,29 @@
+<% can_front_balance = @invoice.sponsor.event.can_front_balance? %>
+
+<p>
+  Hi there,
+</p>
+
+<p>
+  The funds from <%= link_to "a recently paid invoice to #{@invoice.sponsor.name}", @invoice %>
+  <% if can_front_balance %>
```
cjdenio: This condition looks like it's backwards

cjdenio (changes requested):

app/views/invoice_mailer/notify_organizers.html.erb
exu3: Oops thanks for catching that!

cjdenio (approved):
Wahoo!

PR #2855, opened 2022-07-22
Enable `can_front_balance` on event creation

exu3:


app/services/event_service/create.rb
```diff
@@ -12,6 +12,7 @@ def initialize(name:, point_of_contact_id:, emails: [], country: [], category: [
       @sponsorship_fee = sponsorship_fee ? sponsorship_fee.to_f : 0.07
       @organized_by_hack_clubbers = organized_by_hack_clubbers
       @omit_stats = omit_stats
+      @can_front_balance = true
```
cjdenio: I think this should be `@can_front_balance = can_front_balance`

exu3: Ah right, sorry. Fixed in https://github.com/hackclub/bank/pull/2855/commits/86f4f7c962fcf4240ba38c7c1b63392a0f84a56e

garyhtou (approved):

garyhtou:
After this is merged, please check to make sure all events have `can_front_balance = true` (there could be some new events created before this pr is merged in)

PR #2860, opened 2022-07-23
Ignore empty receipt uploaded

garyhtou:
Closes https://github.com/hackclub/bank/issues/2773

It's kinda easy to accidentally click and drag a link. This would trigger the receipt dropzone to appear and attempt a drag-and-drop receipt upload— causing a page reload and an error to be flashed.

https://user-images.githubusercontent.com/20099646/180582910-5ddacd2b-cf66-496c-b990-33e5d9caff1e.mov

This PR prevents submitting a drag and drop receipt upload if no actual files were dropped, and also prevents the receipt upload endpoint from erroring if no files were submitted.

https://user-images.githubusercontent.com/20099646/180583178-ad11db27-3e1d-4560-bdd7-258b3db1fb77.mov


cjdenio (approved):
Thanks for fixing this!

PR #2863, opened 2022-07-24
Create HcbCodes for CanonicalTransactions eagerly

albertchae:
**Best reviewed by commits**

For https://github.com/hackclub/bank/issues/2783

This PR changes `HcbCode` to be created eagerly when creating `CanonicalTransaction`s instead of lazily via a background job. 

This work was initially proposed as a simplification and to make the implementation similar to `CanonicalPendingTransaction` (which was already eagerly created), but it turns out it was causing user facing issues such as https://hackclub.slack.com/archives/GBXBFEED9/p1661437942100129. I believe this is because even though we try to lazily create a missing `HcbCode` if the job hasn't run yet at this point https://github.com/hackclub/bank/blob/0eceaa762393ff208d95d7374c153d13a07490fc/app/models/canonical_transaction.rb#L261, the `hcb_code` string being passed to the `HcbCode` model initializer will be `nil` because these lines https://github.com/hackclub/bank/blob/cb920faf0e1055ed90bd496d6ab11e088501e71f/app/services/transaction_grouping_engine/nightly.rb#L13-L15 haven't run yet. So even more justification to make this change.

This PR includes some slightly tangential work – some minimal factories to support creating a dependent association for `canonical_transaction` as well as fixing a reference to an undefined variable.


app/models/canonical_transaction.rb
```diff
@@ -270,9 +271,21 @@ def memo_hcb_code_likely_invoice?
...
+    safely do
+      # require 'pry'
+      # binding.pry
+      code = ::TransactionGroupingEngine::Calculate::HcbCode.new(canonical_transaction_or_canonical_pending_transaction: self).run
+
+      self.update_column(:hcb_code, code)
+
+      ::HcbCodeService::FindOrCreate.new(hcb_code: code).run
+    end
+  end
```
albertchae: This is copied from

https://github.com/hackclub/bank/blob/66a7463cf5f2e180e308af46c04b2a2920d2d764/app/models/canonical_pending_transaction.rb#L309-L317

https://github.com/hackclub/bank/blob/cb920faf0e1055ed90bd496d6ab11e088501e71f/app/services/transaction_grouping_engine/nightly.rb#L13-L17

app/models/canonical_transaction.rb
```diff
@@ -75,6 +75,7 @@ class CanonicalTransaction < ApplicationRecord
   validates :friendly_memo, presence: true, allow_nil: true
   validates :custom_memo, presence: true, allow_nil: true
 
+  after_create :write_hcb_code
```
albertchae: using `after_create` for same reasons as https://github.com/hackclub/bank/pull/2827

spec/models/canonical_transaction_spec.rb
```diff
@@ -51,4 +51,12 @@
       expect(canonical_transaction).to be_valid
     end
   end
+
+  describe "hcb_code" do
+    it "is reachable from the canonical transaction" do
+      canonical_transaction = create(:canonical_transaction)
+      expect(canonical_transaction.local_hcb_code).to be_present
```
albertchae: this test is currently failing because of a missing `hashed_transaction`

app/models/canonical_transaction.rb
```diff
@@ -270,9 +271,21 @@ def memo_hcb_code_likely_invoice?
...
+      self.update_column(:hcb_code, code)
+
+      ::HcbCodeService::FindOrCreate.new(hcb_code: code).run
+    end
+  end
+
   def hashed_transaction
     @hashed_transaction ||= begin
-      Airbrake.notify("There was more (or less) than 1 hashed_transaction for canonical_transaction: #{canonical_transaction.id}") if hashed_transactions.count != 1
+      Airbrake.notify("There was more (or less) than 1 hashed_transaction for canonical_transaction: #{self.id}") if hashed_transactions.count != 1
```
albertchae: it appears this line never gets run in production, which is why it didn't just crash for referencing a non existent variable

I ran this query on local console and got 0 results

```
CanonicalHashedMapping.pluck(:hashed_transaction_id).tally.filter {|k, v| v > 1}
```

albertchae: I'm wondering if we can remove `CanonicalHashedMapping` entirely and replace https://github.com/hackclub/bank/pull/2863/files#diff-2d7450f73f9a1e1269dfc34a02e9ce75637c66409ee92316b414fc987283e0f2R66 with a `has_one :hashed_transaction`

albertchae: It looks like the answer to this is yes https://github.com/hackclub/bank/issues/2783#issuecomment-1213450383, but I'll address this in a follow up or separate issue

spec/models/canonical_transaction_spec.rb
albertchae: resolved in https://github.com/hackclub/bank/pull/2863/commits/06a04460d5c08103550f0cdc5a3f4b7aef52d134

cjdenio (approved):
This looks great to me— thanks!

albertchae:
Merging now, will track if this PR improves this count
`Ahoy::Event.where(name: "canonicalTransactionMissingHcbCount").last(20).pluck(:properties)`

PR #2865, opened 2022-07-25
Transaction tags

cjdenio:
This PR launches a _slightly buggy, but mostly functional_ implementation of transaction tags.

It isn't ready for widespread usage, but this gets it running in production for testing on HQ.

https://user-images.githubusercontent.com/34525547/185470252-714573a3-a528-4517-8b55-35d6e9b73f69.mov


app/models/hcb_code.rb
```diff
@@ -28,6 +30,8 @@ class HcbCode < ApplicationRecord
 
   monetize :amount_cents
 
+  belongs_to :tag, optional: true
```
garyhtou: would we want a single hcb code to have many tags?

cjdenio: Oh, maybe. I was thinking they'd act more like categories than tags. @exu3 thoughts?

exu3: I think yeah, it might be helpful for a single HCB code to have many tags, sort of like how blog posts have many tags / GitHub issues/prs have many labels.

cjdenio:
ohyeah i kinda like how this turned out

https://user-images.githubusercontent.com/34525547/180891106-45b026fe-7a9a-4e60-8254-6d8fe17d2be9.mov

app/models/tag.rb
```diff
@@ -0,0 +1,24 @@
...
+#
+# Indexes
+#
+#  index_tags_on_event_id  (event_id)
+#
+class Tag < ApplicationRecord
+  belongs_to :event
+  has_and_belongs_to_many :hcb_codes
+
+  validates :label, presence: true, uniqueness: true
```
garyhtou: unique within a single event?

db/schema.rb
```diff
@@ -693,10 +693,17 @@
     t.datetime "updated_at", precision: 6, null: false
     t.datetime "marked_no_or_lost_receipt_at"
     t.text "short_code"
+    t.bigint "tag_id"
     t.index ["hcb_code"], name: "index_hcb_codes_on_hcb_code", unique: true
+    t.index ["tag_id"], name: "index_hcb_codes_on_tag_id"
```
garyhtou: do we wanna drop these now that we have the `hcb_codes_tags` many-to-many join table?

app/models/tag.rb
cjdenio: oops  good catch

cjdenio:
_'Tis time_

db/migrate/20220725220556_create_join_table_hcb_code_tag.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class CreateJoinTableHcbCodeTag < ActiveRecord::Migration[6.1]
+  def change
+    create_join_table :hcb_codes, :tags do |t|
+      # t.index [:hcb_code_id, :tag_id]
```
maxwofford: Are these supposed to be commented?

maxwofford (approved):
Exciting WIP! Left a comment on one migration that looked unfinished, but other then that I'm happy with this going in for the sake of user-testing on HQ/meta orgs.

Please have a design review before flipping to general availability

db/migrate/20220725220556_create_join_table_hcb_code_tag.rb
cjdenio: Oop, this was left over from `rails generate:migration`

app/controllers/hcb_codes_controller.rb
```diff
@@ -130,4 +130,22 @@ def dispute
...
+    hcb_code = HcbCode.find(params[:id])
+    tag = Tag.find(params[:tag_id])
+
+    authorize hcb_code
+    authorize tag
+
+    raise Pundit::NotAuthorizedError if tag.event != hcb_code.event
+
+    if hcb_code.tags.exists?(tag.id)
+      hcb_code.tags.delete(tag)
```
```suggestion
      hcb_code.tags.destroy(tag)
```

`destroy`, since that will run callbacks

garyhtou (approved):
Good to merge in for testing with HQ events. There are a handful of things that'll need to be polished up before GA.

p.s. keep a close eye on the db after merging, I ran into some migration issues on staging and dev.

good work!

sampoder:
[It's time!](https://www.youtube.com/watch?v=a4RbVFXjJf4) I'm excited :D

sampoder:
Yay!!

PR #2883, opened 2022-08-12
Implement local_hcb_code on canonical_pending_transaction as an association

albertchae:
**best reviewed by commits**

## commit 1:  Implement local_hcb_code on CanonicalPendingTransaction as an AR association
If local_hcb_code is represented as an association, we can use standard preloading techniques like `includes`

We can compare the following in rails console before and after this change to see

N+1:
```
CanonicalPendingTransaction.last(10).each {|cpt| cpt.local_hcb_code}
```

includes possible with association, no N+1:
```
CanonicalPendingTransaction.includes(:local_hcb_code).last(10).each {|cpt| cpt.local_hcb_code}
```

## commit 2: Preload local_hcb_code in Event#fronted_incoming_balance_v2_cents

This is a follow up to https://github.com/hackclub/bank/pull/2795#discussion_r924792353

This change reduces these SQL calls
```
CACHE HcbCode Load (0.0ms)  SELECT "hcb_codes".* FROM "hcb_codes" WHERE "hcb_codes"."hcb_code" = $1 LIMIT $2  [["hcb_code", "HCB-500-1042"], ["LIMIT", 1]]
↳ app/models/canonical_pending_transaction.rb:127:in `fronted_amount'
```

Before: SQL Queries: 280 (114 cached)
After: SQL Queries: 267 (106 cached)

## commit 3:  Preload local_hcb_code in PendingTransactionEngine::PendingTransaction::All

Apparently defining an association enables a writer method too, which is
why `pt.local_hcb_code = hcb_code_by_code[pt.hcb_code]` didn't break
after previous commits (although there was funny behavior where it would
load the association first before trying to write over it).

But it turns out we don't need this stitching together of `hcb_code` to
`canonical_pending_transaction` to happen in `AssociationPreloader` anymore
now that `local_hcb_code` is an association, so we can do it in the query
instead (in `PendingTransaction::All`).  Since we are now querying for `hcb_code`
there, move nested preloads that we had in `AssociationPreloader`.

Because we no longer try to assign `pt.local_hcb_code`, which as I
mentioned before would also query it once, we remove several of
these queries

```
HcbCode Load (0.3ms)  SELECT "hcb_codes".* FROM "hcb_codes" WHERE "hcb_codes"."hcb_code" = $1 LIMIT $2  [["hcb_code", "HCB-600-iauth_1LVnTRFSaumjmb9rULcSIpNW"], ["LIMIT", 1]]
↳ app/services/pending_transaction_engine/pending_transaction/association_preloader.rb:33:in `block in preload_associations!'
```


garyhtou (approved):
looks good!

maxwofford (approved):
I'm surprised this change is so few lines. I guess there are going to be a couple places that need to be updated for the `includes` later, but they'll still work with this change. Makes sense to have that in a separate PR.

albertchae:
I will merge this after verifying that https://github.com/hackclub/bank/pull/3104 works

albertchae:
Rebased and incorporated the changes for preloading tags based on feature flag https://github.com/hackclub/bank/pull/2925 and improving comment preloading https://github.com/hackclub/bank/pull/3113

Since we haven't seen any new errors since deploying #3104 , (https://github.com/hackclub/bank/issues/3103#issuecomment-1301625030), I'm going to merge this now.

PR #2884, opened 2022-08-13
Users can set start and end date

exu3:
Fixes https://hackclub.slack.com/archives/CN523HLKW/p1660341803588399


cjdenio (approved):
Thanks for fixing this!

PR #2885, opened 2022-08-13
Disbursement metadata partial shows Sent By user

exu3:
Screen Shot 2022-08-13 at 10 27 05 AM

(Disbursements created before #2359 don't have a created_by or fulfilled_by user.)


maxwofford (approved):
Wahoo! Go ahead and merge when ready.

Doesn't have to be in this PR, but I'd love to also have an admin-only one for `fulfilled_by`

PR #2887, opened 2022-08-13
Drop unencrypted columns on AchTransfer and GSuiteAccount

albertchae:
Follow up to https://github.com/hackclub/bank/pull/2886

Because these columns are ignored, it is safe to drop them


kunalbotla:
will this retain account numbers from previously processed ACH transfers? i think it's important to keep them visible for reference by organizations

albertchae:
Hey @kunalbotla This PR is only to drop the unencrypted column. We're still keeping the data (but encrypted in `account_number_ciphertext`). This is the next step in this process https://github.com/ankane/lockbox#migrating-existing-data

maxwofford (approved):
LGTM!

albertchae:
Because there is still some risk to any column dropping, I created backup `b549` on heroku just in case. Merging now

PR #2888, opened 2022-08-13
Unignore dropped columns on AchTransfer and GSuiteAccount

albertchae:
Follow up to https://github.com/hackclub/bank/pull/2887

These columns are dropped, now we can remove them from ignored_columns


maxwofford (approved):
The long chain is finally complete!

albertchae:
@maxwofford get ready for the chain at least 2 more times 😛

PR #2681, opened 2022-06-16
Users can rename transactions

cjdenio:
Closes #2674, possibly closes #1855

PR #2889, opened 2022-08-13
Bank account transactions are user-renameable

exu3:


exu3:
Whoops, just saw #2681; feel free to close this PR if the implementation doesn't seem quite right.

PR #2891, opened 2022-08-15
Add a space and update instant transfer message

kunalbotla:
| old | new |
| --- | --- |
| image | image |
| image | image |
| `Funds should be available[DATE]` | `Funds should be available from [DATE]` |


cjdenio (approved):
Thanks!

PR #2892, opened 2022-08-15
Fix styling on hackathons promo card

exu3:
Now:

Screen Shot 2022-08-15 at 2 53 24 PM

Prev:
sc


cjdenio (approved):
oops, thanks for fixing this!

PR #2902, opened 2022-08-19
Only show fancy styling on grant receiving end

cjdenio:
no need for this much celebration

Screen Shot 2022-08-19 at 9 38 28 AM


exu3 (approved):
🎉 :x:

PR #2904, opened 2022-08-19
Prepare to encrypt 4 additional columns with lockbox

albertchae:
The additional columns are:
- bank_accounts.plaid_access_token
There is some debate about whether this needs to be kept secret, but I
opted to go ahead with it since some people think it should be
https://stackoverflow.com/questions/40473684/is-a-plaid-access-token-a-secret

- invoices.payment_method_ach_credit_transfer_account_number
If ach_transfers.account_number is being encrypted, it seems this should
be too

- partners.api_key
This is like storing a password as plaintext which a malicious actor
could use to perform API calls as a partner, so encrypting this now.

- user_sessions.session_token
Malicious actors could use this to fashion a valid session cookie.

Part of https://github.com/hackclub/bank/issues/2541


albertchae:
Best reviewed with whitespace off because of all the `annotate` changes
image

albertchae:
There is one more obvious candidate for encryption https://github.com/hackclub/bank/blob/9d972e06f2fd99c2f8de5fbd6a187ffab8f4977a/app/models/user.rb#L9 but this column is indexed, so it requires a bit more work and so will come in a separate PR

Besides that there are several memo type fields that we may or may not want to encrypt. I'll start a discussion about that in the related issue.

maxwofford (approved):
Awesome, thanks for catching these fields!

albertchae:
Had to deploy twice due to a lock timeout on the UserSession migration
https://dashboard.heroku.com/apps/bank-hackclub/activity/releases/9dceca2e-6dee-43fa-aa52-cc2dfbbd2b70
https://dashboard.heroku.com/apps/bank-hackclub/activity/releases/0c19d88e-01fa-41a4-859c-5526350416fe

It succeeded the second time with no problems

albertchae:
Ran these migrations locally and then on prod
```
Lockbox.migrate(BankAccount)
Lockbox.migrate(Invoice)
Lockbox.migrate(Partner)
```

However I'm having some problems with `Lockbox.migrate(UserSession)` locally, so I need to investigate what's going on here

albertchae:
The error I was getting with `UserSession` was
```
[1] pry(main)> Lockbox.migrate(UserSession)
  UserSession Load (3.4ms)  SELECT "user_sessions".* FROM "user_sessions" WHERE "user_sessions"."session_token_ciphertext" IS NULL ORDER BY "user_sessions"."id" ASC LIMIT $1  [["LIMIT", 1000]]
ActiveRecord::RecordNotSaved: Failed to save the record
from /bundle/ruby/2.7.0/gems/activerecord-6.1.6/lib/active_record/persistence.rb:507:in `save!'
```
Apparently this stems from lockbox not playing well with paranoid. Turns out there's a closed issue with some workarounds https://github.com/ankane/lockbox/issues/84. The easiest one for now is `Lockbox.migrate(UserSession.all)` to only migrate non soft deleted records. Although it seems like there are convoluted ways to migrate soft deleted records, it seems like it might be OK to just drop `session_token` from `UserSession`s that have been soft deleted. What do you think @maxwofford ?

PR #2914, opened 2022-08-20
Ignore migrated encrypted columns on bank_accounts, invoices, partners, user_sessions

albertchae:
Follow up to https://github.com/hackclub/bank/pull/2904

Similar to https://github.com/hackclub/bank/pull/2886 but for these 4 columns.

As a result of ignoring `api_key` on `partners`, which has a default attribute setting, we now get this warning which advises us to use `after_initialize`:
```
[lockbox] WARNING: attributes with `:default` option are not supported. Use `after_initialize` instead.
```
Related issue:
https://github.com/ankane/lockbox/issues/143

We also have to enable blind_index because the search column is now being ignored, so queries by `api_key`/`session_token` will fail on new records. Enabling blind_index should be safe after the migration here https://github.com/hackclub/bank/pull/2916#issuecomment-1223296668


albertchae:
do not merge until https://github.com/hackclub/bank/pull/2916 is merged and `BlindIndex` migration is done

EDIT: `BlindIndex` migration is done and it was required to turn `migrating` off for it https://github.com/hackclub/bank/pull/2914/commits/1eb8802347dfa6178c8da1761beac201169dfdb0

maxwofford (approved):

albertchae:
Tested that a previously logged in session still worked fine and that I was able to log in with a new session.

PR #2916, opened 2022-08-21
Add blind_index gem and use it for partners.api_key and user_sessions.session_token

albertchae:
Part of https://github.com/hackclub/bank/issues/2541

We are moving to encrypt these columns with lockbox and thus won't be
able to search these columns by plaintext anymore. lockbox author's
recommended approach is to use the blind_index gem (also from same author).
https://github.com/ankane/lockbox#active-record

These are both set to `migrating: true` for now until we can run
```
BlindIndex.backfill(UserSession)
BlindIndex.backfill(Partner)
```
on production

Although we encrypted 4 columns in https://github.com/hackclub/bank/pull/2904, this PR only contains the 2 columns that are used in search queries.

partners.api_ley
https://github.com/hackclub/bank/blob/5b11e27e2e3bff30ad544d3fd58dd5a0e109e9d7/app/services/auth_service/current_partner.rb#L10

user_sessions.session_token
https://github.com/hackclub/bank/blob/5b11e27e2e3bff30ad544d3fd58dd5a0e109e9d7/app/helpers/sessions_helper.rb#L77

Also add tests for both models that would fail if we dropped the
plaintext column without the corresponding blind index.

Because there's a application level uniqueness validation on
partners.api_key, I made the blind index unique to guarantee it at the
database level.


maxwofford (approved):

albertchae:
I finished running

```
BlindIndex.backfill(UserSession)
BlindIndex.backfill(Partner)
```

It appears it didn't touch soft deleted records in `UserSession` which seems in line with the behavior of `Lockbox` https://github.com/hackclub/bank/pull/2904#issuecomment-1221364069

maxwofford:
That seems like the behavior we want– soft deleted stuff is now encrypted, but not indexed (b/c it's not searched by anything), right?

albertchae:
@maxwofford actually the least work for us is to both not encrypt or index soft deleted records. We'd just lose the `session_token` on them entirely. More details in https://github.com/hackclub/bank/pull/2904#issuecomment-1221364069

If you think it's worth keeping `session_token`s on soft deleted records (but encrypted), we'll have to implement one of the hacky workarounds here https://github.com/ankane/lockbox/issues/84 (either use a fork or remove paranoid from a console instance)

albertchae:
Recapping meeting with @maxwofford on August 26, 2022 – We are okay dropping  `session_token` data on soft deleted columns since they are unlikely to be used for debugging. We discussed an edge case with users who haven't used a logged in session on a browser for over a year and then they return, but that should be OK because those `user_session`s will have been encrypted.

PR #2917, opened 2022-08-22
Use any_word option for pg_search_scope on transactions

albertchae:
search_memo was being called here
https://github.com/hackclub/bank/blob/5b11e27e2e3bff30ad544d3fd58dd5a0e109e9d7/app/services/pending_transaction_engine/pending_transaction/all.rb#L37

Apparently the default (non any_word) behavior is to require every search term to match, although I would still think go would match goshippo...
https://github.com/Casecommons/pg_search#any_word

Without any_word, the query uses `&&`
```
bank_development=# select to_tsvector('POSTAGE GOSHIPPO.COM') @@ (to_tsquery('english', ''' ' || 'go' || ' ''' || ':*') && to_tsquery('english', ''' ' || 'shippo' || ' ''' || ':*'));
 ?column?
----------
 f
(1 row)
```

With any_word, the query uses `||`
```
bank_development=# select to_tsvector('POSTAGE GOSHIPPO.COM') @@ (to_tsquery('english', ''' ' || 'go' || ' ''' || ':*') || to_tsquery('english', ''' ' || 'shippo' || ' ''' || ':*'));
 ?column?
----------
 t
(1 row)
```

Although the bug was specifically for CanonicalPendingTransaction, it
seemed like a good idea to add it to CanonicalTransaction too

Closes https://github.com/hackclub/bank/issues/2809


albertchae:
Notes on an alternative approach https://github.com/hackclub/bank/issues/2809#issuecomment-1221652079

maxwofford (approved):
Thanks for including tests!

PR #2924, opened 2022-08-22
Expose tags in API

exu3:
Closes #2915


cjdenio (changes requested):
Thanks, this is great!

Could you:
- Include `PublicIdentifiable` on `Tag`
- make a Tag API entity and only expose the `label` and `id` fields (for now)
Screen Shot 2022-08-22 at 1 28 49 PM

cjdenio (approved):
yeah

exu3:
image

PR #2925, opened 2022-08-22
Make transaction tags speedier

cjdenio:
speeeeeeeeed

Tags can still be optimized further, but this is a massive improvement.

Things I did:
- Preload `HcbCode#tags`
- Remove the `tags/add_tag` partial since it added a bunch of rendering time


albertchae (approved):
Nice! I did a comparison before and after and the preloading definitely helped. Inlining the partial I'm not so sure about though, and it's hard to tell locally because local view rendering time fluctuates quite a bit for me at least. I would imagine that as long as we are rendering the transaction partial in a collection a la https://github.com/hackclub/bank/pull/2481, the `_add_tag` partial would only be loaded into memory a handful of times too 🤔

I did see quite a few of these cached queries
```
  CACHE  (0.0ms)  SELECT COUNT(*) FROM "tags" WHERE "tags"."event_id" = $1  [["event_id", 183]]
  ↳ app/views/canonical_transactions/_canonical_transaction.html.erb:93
  CACHE  (0.0ms)  SELECT COUNT(*) FROM "tags" WHERE "tags"."event_id" = $1  [["event_id", 183]]
  ↳ app/views/canonical_transactions/_canonical_transaction.html.erb:98
```
which could be worth removing. I surprisingly saw speedup from removing cached queries (by caching it explicitly and not having it drop to the activerecord cache). However now that I think about it I wonder if it was just slow on local because of logging the queries to STDOUT and might not be a big deal in prod...

PR #2930, opened 2022-08-22
Decline pending transactions once an invoice is marked as paid

maxwofford:


app/services/pending_event_mapping_engine/decline/single/invoice.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        def run
+          return unless @canonical_pending_transaction.unsettled?
+
+          return unless invoice
+
+          return unless invoice.manually_marked_as_paid?
+
+          if @canonical_pending_transaction.settled?
+            return Airbreak.notify("Invoice #{invoice.id} is manually marked as paid but is settled")
```
cjdenio:
```suggestion
            return Airbrake.notify("Invoice #{invoice.id} is manually marked as paid but is settled")
```

app/services/invoice_service/mark_paid.rb
```diff
@@ -57,5 +59,10 @@ def remote_invoice
       @remote_invoice ||= ::Partners::Stripe::Invoices::Show.new(id: invoice.stripe_invoice_id).run
     end
 
+    def decline_pending_transactions!
+      canonical_pending_transaction = invoice&.raw_pending_invoice_transaction&.canonical_pending_transaction
```
cjdenio:
```suggestion
      canonical_pending_transaction = invoice&.canonical_pending_transaction
```

`RawPendingInvoiceTransaction` doesn't have a `.canonical_pending_transaction`, oddly enough

app/services/invoice_service/mark_paid.rb
```diff
@@ -41,6 +41,8 @@ def run
           remote_invoice.paid = true
           remote_invoice.save
         end
+
+        decline_pending_transactions!
```
cjdenio: Potential bug: the pending transaction probably doesn't exist at this point.

We create the pending transaction when we receive the `invoice.paid` webhook (which gets sent when an invoice is marked as paid outside of Stripe).

https://github.com/hackclub/bank/blob/ea3e10e31104194cf77f89fda49c7c26259fd905/app/controllers/stripe_controller.rb#L78-L89

cjdenio: I think a better solution would be to either:
- Not create a pending transaction if the invoice was paid outside of Stripe
- Decline the transaction immediately in `handle_invoice_paid` if the invoice was paid outside of Stripe

maxwofford: That's a good point about not creating a pending transaction. Is there any world an invoice is marked as paid after stripe reports payment? Just wondering what sorts of edge-cases there are to test out.

cjdenio:
Closing since I think this is superseded by #2932.

PR #2932, opened 2022-08-23
Don't create pending transactions for manually-marked-as-paid invoices

cjdenio:
Part of #2931.


maxwofford:
Could https://github.com/hackclub/bank/blob/ea3e10e31104194cf77f89fda49c7c26259fd905/app/services/pending_transaction_engine/raw_pending_invoice_transaction_service/invoice/import.rb still run and import PTs?

cjdenio:
@maxwofford, just fixed that!

(our cron jobs also have a safeguard against importing manually-marked-as-paid invoices)

maxwofford:
i'd assume (based on naming) that `InvoiceService::OpenToPaid` could handle manually `marked_as_paid` invoices. Maybe we still handle those incoming webhooks, but make sure to process invoices correctly?

PR #2939, opened 2022-08-24
Fix transaction memo in export

exu3:
Instead of `ct.smart_memo`, csv/json should now use `ct.local_hcb_code.memo` which is what the event show page uses.

Screen Shot 2022-08-24 at 9 07 07 AM


maxwofford (approved):
Nice catch, thanks!

PR #2942, opened 2022-08-24
Make transaction tags work with disbursements

cjdenio:
Fixes #2909


cjdenio:
doot

###### merging because this is all feature flagged

PR #2944, opened 2022-08-25
Switch over to JS-positioned dropdowns

cjdenio:
This mostly-not-janky PR adds a neat Stimulus controller that enables fancy JS positioning for dropdowns.

Eventually the entire dropdown setup should be moved over to this, _but_ certain ones don't work properly with it right now.

Fixes #2913


YodaLightsabr:
I'm not sure if this is within the scope of this PR but scrolling within the drop-down is kind of janky on mobile.

image

cjdenio:
doot

PR #2953, opened 2022-08-28
Drop migrated encrypted columns on bank_accounts, invoices, partners, user_sessions

albertchae:
Follow up to https://github.com/hackclub/bank/pull/2914

Now that these columns are ignored, we can safely drop them.

Similar to https://github.com/hackclub/bank/pull/2887


maxwofford (approved):
LGTM!

albertchae:
Manually created backup `b566` before deploying this

albertchae:
https://dashboard.heroku.com/apps/bank-hackclub/activity/releases/b95c6be1-67a7-4e05-8730-7edfc1e14174

First deployment try succeeded on 3 out of 4 migrations but timed out on `remove_column(:invoices, :payment_method_ach_credit_transfer_account_number)`, so trying again now.

albertchae:
Redeploying succeeded https://dashboard.heroku.com/apps/bank-hackclub/activity/releases/90b05d62-67ff-4ed7-8c8a-9b5627274d8b

maxwofford:
https://github.com/hackclub/bank/pull/2953#issuecomment-1236402553

That's funny– any thoughts what could cause that?

albertchae:
@maxwofford actually it can be fairly common to have lock contention during a schema migration. I believe remove column requires an `ACCESS EXCLUSIVE` lock on the table it's altering, which means both that it has to wait for other queries accessing the table to finish AND it'll block other queries from accessing the table while it's running. So in this case it probably waited 10 seconds for other queries to finish and it didn't happen.

See `Adding a column` section in https://medium.com/doctolib/stop-worrying-about-postgresql-locks-in-your-rails-migrations-3426027e9cc9 for a more depth explanation

I believe the lock migration timeout is a feature of strong_migrations https://github.com/hackclub/bank/blob/240818fa902cee9f9333f43124156d6234a4f3c2/config/initializers/strong_migrations.rb#L8. 

Anyway, It's generally better to let the migration timeout and then try again, otherwise you can end up taking down the site like the doctolib blog post mentions. That's also why it's a good idea to split up migrations to multiple tables into separate files (in case any one of them fail) and also to babysit any deploys that include migrations.

PR #2954, opened 2022-08-28
Unignore dropped columns on bank_accounts, invoices, partners, user_sessions

albertchae:
Follow up to https://github.com/hackclub/bank/pull/2953

These columns are dropped, now we can remove them from ignored_columns

Similar to https://github.com/hackclub/bank/pull/2888


maxwofford (approved):
Looks good– matches up with #2953

PR #2956, opened 2022-08-28
Begin migration for blind_index and lockbox on users.api_access_token

albertchae:
It looks like this index is only used for guaranteeing uniqueness, so I
added a test for that and not one for searching.


maxwofford (approved):
Looks great!

albertchae:
TODO on prod:
- [x] `BlindIndex.backfill(User)`
- [x] `Lockbox.migrate(User)`

albertchae:
Both backfills complete! onto https://github.com/hackclub/bank/pull/2957

PR #2957, opened 2022-08-28
Ignore users.api_access_token for lockbox and blind_index

albertchae:
Follow up to https://github.com/hackclub/bank/pull/2956

Test needs to check for unique constraint on the blind index instead of the unencrypted index now


maxwofford (approved):
LGTM!

PR #2959, opened 2022-08-29
Unignore users.api_access_token after dropping it

albertchae:
Follow up to https://github.com/hackclub/bank/pull/2958


maxwofford (approved):
Nice, another one down!

PR #2971, opened 2022-08-31
Instantly create PTs for ACHs

cjdenio:
This PR fixes the famed ACH 404 error by instantly creating a `CanonicalPendingTransction` when an ACH transfer is created (which then, in turn, creates an `HcbCode`).


exu3 (commented):
Is the ACH supposed to appear on the event show after being created? I created a couple ACHs but they didn't seem to show up on the ledger or affect the balance.

Screen Shot 2022-08-31 at 2 35 46 PM

Screen Shot 2022-08-31 at 2 36 12 PM

cjdenio:
oops, thanks for catching that

exu3 (changes requested):
###### wahoo!

One more thing: if the ACH gets rejected, it still stays on the ledger and affects the balance.

app/services/ach_transfer_service/create.rb
```diff
@@ -23,7 +23,11 @@ def initialize(event_id:,
     def run
       raise ArgumentError, "You don't have enough money to send this transfer." unless ample_balance?(@amount_cents, event)
 
-      AchTransfer.create!(create_attrs)
+      ach_transfer = AchTransfer.create!(create_attrs)
+
+      rpoat = PendingTransactionEngine::RawPendingOutgoingAchTransactionService::OutgoingAch::ImportSingle.new(ach_transfer: ach_transfer).run
+      pt = PendingTransactionEngine::CanonicalPendingTransactionService::ImportSingle::OutgoingAch.new(raw_pending_outgoing_ach_transaction: rpoat).run
+      PendingEventMappingEngine::Map::Single::OutgoingAch.new(canonical_pending_transaction: pt).run
```
albertchae: It might be worth wrapping this in a [database transaction](https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html) so it's all or nothing?

cjdenio:
> ###### wahoo!
> One more thing: if the ACH gets rejected, it still stays on the ledger and affects the balance.

It will disappear from the ledger, just on a cron job instead of instantly. I agree it's not the ideal behavior, but I think it's out of scope since that was the behavior before this PR

exu3 (approved):

PR #2974, opened 2022-09-01
Don't use hcb code amount for achs and donations

exu3:
Fixes https://hackclub.slack.com/archives/CN523HLKW/p1662006955468239


cjdenio (approved):
k0jjkk

exu3:
Many are saying this........

PR #2978, opened 2022-09-01
add spending limits?

bellesea:


app/controllers/stripe_cards_controller.rb
```diff
@@ -87,6 +89,20 @@ def create
...
+    authorize @card
+  end
+
+  def update
+    @card = StripeCard.includes(:event, :user).find(params[:id])
+    authorize @card
+
+      flash[:success] = "Updated successfully"
+      redirect_to @card
+
```
cjdenio: There should be some updating done here

config/routes.rb
```diff
@@ -221,9 +221,10 @@
     resources :comments
   end
   resources :stripe_cardholders, only: [:new, :create, :update]
-  resources :stripe_cards, only: %i[create index show] do
+  resources :stripe_cards, only: %i[create index show edit update] do
     post "freeze"
     post "defrost"
+    post "update"
```
cjdenio:
```suggestion
```

This is unnecessary because `update` is already set by the `resources` thing

app/views/stripe_cards/show.html.erb
```diff
@@ -98,12 +98,20 @@
...
+    <section class="flex flex-row items-end pt2">
       <div class="stat tooltipped tooltipped--s" aria-label="This is your account balance">
-        <strong class="stat__label">Spending limit</strong>
-        <span class="stat__value">
-          <%= render_money_amount @event.balance_v2_cents %>
-        </span>
+      <strong class="stat__label">Spending limit</strong>
+      <div class="flex flex-row">
+      <span class="stat__value">
+      <% if @card.spending_limit_amount == 2000000 %>
```
cjdenio: What if the user manually sets the spending limit to $20,000?

app/views/stripe_cards/_form.html.erb
```diff
@@ -23,11 +23,21 @@
...
+    <div class="field">
+    <%= form.label :stripe_lifetime_limit, "Lifetime spending limit (USD)" %>
+    <%= form.text_field :stripe_lifetime_limit %>
+    <small>Total amount that can be spent on this card. Leave empty if no limit. Can be changed later.</small>
+  </div>
+  <div class="field">
+    <%= form.label :stripe_daily_limit, "Daily spending limit (USD)" %>
+    <%= form.text_field :stripe_daily_limit %>
+    <small>Daily max amount that can be spent on this card. Leave empty if no limit. Can be changed later.</small>
+  </div>
```
cjdenio: I don't know about having this here— spending limits feel like an "advanced" option, and this makes the card order form a bit long

cjdenio (changes requested):
This is a great start!

app/controllers/stripe_cards_controller.rb
bellesea: thanks!

app/views/stripe_cards/show.html.erb
bellesea: hmm i'm not sure, any suggestions? cuz this is the default spending limit... or maybe we could set the default to the event balance?

app/views/stripe_cards/_form.html.erb
bellesea: We cut it down from two types of limits to one, maybe that could help? cuz i think being able to immediately set limit could be helpful?

maxwofford: What if it's a 2 step process– first order your card, then on the "your card has been created!" page giving the spending limits?

maxwofford: Errr, hit send before I finished– the intent is to make it so spending limits are a bonus thing users can opt-into on top of the card ordering flow instead of making the current flow longer.

bellesea: Hey Max 👋 Ahh yeah that makes sense! What if it's something like

> Set spending limit (with the link to the edit page)?

PR #2980, opened 2022-09-04
Progress on Rubocop todo

maxwofford:
This PR mostly consists of rubocop auto-corrections from the `.rubocop_todo.yml`. It included some modifications to the Donations model I added tests for, so this also includes [adding those tests](https://github.com/hackclub/bank/commit/fe72f0474d65122e577236b94d39decf45b726c2).


app/controllers/fee_reimbursements_controller.rb
```diff
@@ -5,11 +5,7 @@ class FeeReimbursementsController < ApplicationController
 
   # GET /fee_reimbursements/1
   def show
-    if @fee_reimbursement.invoice
-      @event = @fee_reimbursement.event
-    else
-      @event = @fee_reimbursement.event
-    end
```
albertchae: 😆

spec/models/donation_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    expect do
+      donation = create(:donation, event: event)
+      donation.status = "succeeded"
+      donation.save
+
+      donation2 = create(:donation, event: event)
+      donation2.status = "succeeded"
+      donation2.save
+    end.to change(enqueued_jobs, :size).by(1)
+  end
```
albertchae: nice, great tests! What about one more for the case where you have a donation that was already in the "succeeded" state and then it gets saved again?

app/models/donation.rb
```diff
@@ -267,13 +267,12 @@ def send_payment_notification_if_needed
...
-        DonationMailer.with(donation: self).first_donation_notification.deliver_later
-        return
-      end
-    end
+    # send special email on first donation paid
+    return if was == "succeeded"
+    return unless now == "succeeded"
+    return unless self.event.donations.succeeded.size == 1
+
+    DonationMailer.with(donation: self).first_donation_notification.deliver_later
```
albertchae: I think this change obscures the business logic which the original expressed better IMO. Maybe this would be better written with some helper methods. What about something like:

```ruby
def status_changed_to_succeeded?(saved_status_changes)
    was = saved_status_changes[0] # old value of status
    now = saved_status_changes[1] # new value of status
    
    was != "succeeded" && now == "succeeded"
end

def first_donation?
   self.event.donations.succeeded.size == 1
end

def send_payment_notification_if_needed
   return unless saved_changes[:status].present?
   
   if status_changed_to_succeeded?(saved_changes[:status]) && first_donation?
     DonationMailer.with(donation: self).first_donation_notification.deliver_later
   end
end
```

Not sure if rubocop would then make you change the last part to
```ruby
def send_payment_notification_if_needed
   return unless saved_changes[:status].present?
   
   return unless status_changed_to_succeeded?(saved_changes[:status])
   return unless first_donation?

   DonationMailer.with(donation: self).first_donation_notification.deliver_later
end
```

Either way I think the helper methods improve expressing the intent. What do you think?

albertchae (approved):
Thanks for the cleanup and the additional tests. LGTM but I had a suggestion on the rubocop changes for `Donation`

PR #2981, opened 2022-09-04
Add missing receipt count badge to nav

maxwofford:
This PR updates https://github.com/hackclub/bank/pull/2598 to run in an `async_frame`.

Screen Shot 2022-09-03 at 21 53 03


app/controllers/static_pages_controller.rb
```diff
@@ -63,6 +63,32 @@ def my_missing_receipts_list
...
+  # async frame
+  def my_missing_receipts_icon
+    @missing_receipt_count ||= begin
+      count = 0
+
+      stripe_cards = current_user.stripe_cards.includes(:event)
+      emburse_cards = current_user.emburse_cards.includes(:event)
+
+      (stripe_cards + emburse_cards).each do |card|
+        card.hcb_codes.missing_receipt.each do |hcb_code|
```
albertchae: I know the impact is reduced with changing this to be async, but just highlighting again this is an N+1 like https://github.com/hackclub/bank/pull/2598#discussion_r898800332. But if most users don't have that many cards, not a problem for now

albertchae: I got curious and ran `User.all.map {|u| [u.id, u.stripe_cards.count + u.emburse_cards.count]}.sort_by {|k, v| -v}` and here were the top few users
```
[[476, 24],
 [2596, 18],
 [892, 17],
 [1785, 17],
 [1115, 17],
 [1711, 17],
 [2232, 17],
 [893, 14],
 [720, 13],
 [2385, 12],
 [34, 11],
 [980, 11],
 [1585, 11],
 [453, 10],
 ```

I also ran `User.all.map {|u| u.stripe_cards.count + u.emburse_cards.count}.tally.sort` to get this
```
[[0, 2399],
 [1, 490],
 [2, 207],
 [3, 71],
 [4, 30],
 [5, 17],
 [6, 6],
 [7, 7],
 [8, 2],
 [9, 6],
 [10, 1],
 [11, 3],
 [12, 1],
 [13, 1],
 [14, 1],
 [17, 5],
 [18, 1],
 [24, 1]]
 ```
 So looks like the vast majority of users have 5 cards or less

app/helpers/events_helper.rb
```diff
@@ -3,11 +3,18 @@
...
-  def dock_item(name, tooltip, icon, url, lg = false)
+  def dock_item(name, tooltip, icon, url, lg = false, async_badge = nil)
     link_to url,
             class: "dock__item #{lg && 'dock__item--lg'} tooltipped tooltipped--e",
             'aria-label': tooltip do
-      inline_icon(icon, size: 32, class: "primary") + content_tag(:span, name.html_safe, class: "line-height-3")
+      (content_tag :div, class: "line-height-0 relative" do
+        if async_badge
+          inline_icon(icon, size: 32, class: "primary") +
+          content_tag(:div, nil, 'data-src': async_badge, 'data-behavior': :async_frame, as: :div)
```
albertchae: I'm not familiar with `'data-behavior': :async_frame`, is this a turbo thing?

albertchae (approved):
LGTM and works nicely locally. It looks like you didn't use the service from https://github.com/hackclub/bank/pull/2598 though, is that something you're intending to do in a future change?

app/helpers/events_helper.rb
cjdenio: It's a little utility thing:
https://github.com/hackclub/bank/blob/4c520deaf759c7b78c85a90403de025d42c033df/app/assets/javascripts/ui.js#L39-L43

albertchae: oh cool, good to know

maxwofford:
Updated the description to just mention https://github.com/hackclub/bank/pull/2598 now!

app/helpers/events_helper.rb
cjdenio: Not necessarily in the scope of this PR, but it would be nice if this supported non-async badges

cjdenio (approved):
This looks good to me!

app/helpers/events_helper.rb
maxwofford: Def! Didn't see any usages of this for other stuff, but let's add it once we do!

maxwofford:
_meep boop_ –Caleb

PR #2982, opened 2022-09-05
Rebuild db/schema.rb to reflect current prod database

albertchae:
I noticed on a few `docker-setup.sh` runs that the schema.rb kept
changing so I decided to investigate.

Using `heroku pg:psql -a bank-hackclub`, I verified these changes exist
in the prod schema.


app/models/canonical_pending_transaction.rb
```diff
@@ -20,6 +20,7 @@
 #  raw_pending_outgoing_ach_transaction_id          :bigint
 #  raw_pending_outgoing_check_transaction_id        :bigint
 #  raw_pending_outgoing_disbursement_transaction_id :bigint
+#  raw_pending_partner_donation_transaction_id      :bigint
```
albertchae: Not sure how this got missed before, but it's there in prod
```
bank-hackclub::DATABASE=> \d canonical_pending_transactions;
                                                             Table "public.canonical_pending_transactions"
                      Column                      |              Type              | Collation | Nullable |        >
--------------------------------------------------+--------------------------------+-----------+----------+-------->
...
 raw_pending_donation_transaction_id              | bigint                         |           |          |
```

db/schema.rb
```diff
@@ -16,6 +16,7 @@
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
+  enable_extension "pg_stat_statements"
```
albertchae:
```

bank-hackclub::DATABASE=> \dx
                                     List of installed extensions
        Name        | Version |   Schema   |                        Description
--------------------+---------+------------+-----------------------------------------------------------
 citext             | 1.5     | public     | data type for case-insensitive character strings
 pg_stat_statements | 1.6     | public     | track execution statistics of all SQL statements executed
 plpgsql            | 1.0     | pg_catalog | PL/pgSQL procedural language
(3 rows)
```

db/schema.rb
```diff
@@ -30,8 +31,8 @@
     t.datetime "updated_at", null: false
     t.string "recipient_tel"
     t.datetime "rejected_at"
-    t.text "payment_for"
     t.datetime "scheduled_arrival_date"
+    t.text "payment_for"
```
albertchae: minor reordering

db/schema.rb
```diff
@@ -542,7 +544,7 @@
     t.boolean "omit_stats", default: false
     t.datetime "transaction_engine_v2_at", default: -> { "CURRENT_TIMESTAMP" }
     t.datetime "last_fee_processed_at"
-    t.datetime "pending_transaction_engine_at", default: "2021-02-13 22:27:44"
+    t.datetime "pending_transaction_engine_at", default: "2021-02-13 22:49:40"
```
albertchae: Matches this value as well as https://github.com/hackclub/bank/blob/240818fa902cee9f9333f43124156d6234a4f3c2/app/models/event.rb#L33

```
bank-hackclub::DATABASE=> \d events;
                                                              Table "public.events"
             Column              |            Type             | Collation | Nullable |                          Default
---------------------------------+-----------------------------+-----------+----------+-----------------------------------------------------------
...
 pending_transaction_engine_at   | timestamp without time zone |           |          | '2021-02-13 22:49:40.981965'::timestamp without time zone
 ```

File: db/schema.rb [hunk 5]
```diff
@@ -1115,18 +1117,6 @@
...
-    t.integer "amount_cents"
-    t.datetime "created_at", precision: 6, null: false
-    t.datetime "updated_at", precision: 6, null: false
-    t.string "aasm_state"
-    t.text "reimbursement_for"
-    t.bigint "event_id", null: false
-    t.index ["event_id"], name: "index_reimbursements_on_event_id"
-    t.index ["user_id"], name: "index_reimbursements_on_user_id"
-  end
-
```
albertchae: Looks like this was accidentally added in https://github.com/hackclub/bank/pull/2865/files#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3R1117-R1128

db/schema.rb
```diff
@@ -256,6 +257,7 @@
     t.bigint "raw_pending_invoice_transaction_id"
     t.text "hcb_code"
     t.bigint "raw_pending_bank_fee_transaction_id"
+    t.bigint "raw_pending_partner_donation_transaction_id"
```
albertchae: same as https://github.com/hackclub/bank/pull/2982#discussion_r962398325

db/schema.rb
maxwofford: Should we also update https://github.com/hackclub/bank/blob/main/db/migrate/20210213222744_change_default_for_pending_transaction_engine_at.rb#L7 to keep it inline with production?

maxwofford (approved):
Thanks for catching all this! Left a comment, but I'm happy with this merging when you're ready.

db/schema.rb
albertchae: good call, no wonder there was a discrepancy 😆

PR #2983, opened 2022-09-05
 Copy yarn install hack into docker_setup.sh too 

albertchae:
I've run into this a few times when running docker_setup.sh, so just
copying over the hack from https://github.com/hackclub/bank/pull/2457


maxwofford (approved):
lol, i think that's fine for now

PR #2984, opened 2022-09-05
Upgrade postgres to 11.16 to match heroku production version with trust

albertchae:
image

Just changing the version led to this error
```
Error: Database is uninitialized and superuser password is not specified.
       You must specify POSTGRES_PASSWORD to a non-empty value for the
       superuser. For example, "-e POSTGRES_PASSWORD=password" on "docker run".

       You may also use "POSTGRES_HOST_AUTH_METHOD=trust" to allow all
       connections without a password. This is *not* recommended.

       See PostgreSQL documentation about "trust":
       https://www.postgresql.org/docs/current/auth-trust.html
```

This PR uses trust for postgres instead of a password.

According to the documentation, this should be safe due to

`trust authentication is appropriate and very convenient for local connections on a single-user workstation`

as well as

`This method should only be used when there is adequate operating-system-level protection on connections to the server.`


albertchae:
I didn't have a specific issue where I ran into a need to upgrade, but when I was working on cleaning up the database schema I noticed the versions are out of sync. I usually try to keep local and prod versions in sync to avoid weird errors in the future.

albertchae:
@maxwofford I mentioned this last week but this is one of 2 approaches along with https://github.com/hackclub/bank/pull/2985, let me know what you think

maxwofford:
I'm partial to going with trust on this over passwords (https://github.com/hackclub/bank/pull/2985) to keep things simple, but I want to make sure I understand the trade-off. Would a password just protect against local network port scanning? Is our development DB responding to external connections?

albertchae:
@maxwofford I think because we aren't exposing ports in the `docker-compose.yml`, the postgres instance will only be accessible within the docker network.

when I have bank_db_1 running and I try nmap
```
$ nmap -v -p 5432 127.0.0.1

Starting Nmap 7.92 ( https://nmap.org ) at 2022-09-23 10:58 PDT
Initiating Ping Scan at 10:58
Scanning 127.0.0.1 [2 ports]
Completed Ping Scan at 10:58, 0.00s elapsed (1 total hosts)
Initiating Connect Scan at 10:58
Scanning localhost (127.0.0.1) [1 port]
Completed Connect Scan at 10:58, 0.00s elapsed (1 total ports)
Nmap scan report for localhost (127.0.0.1)
Host is up (0.00024s latency).

PORT     STATE  SERVICE
5432/tcp closed postgresql

Read data files from: /opt/homebrew/bin/../share/nmap
Nmap done: 1 IP address (1 host up) scanned in 0.04 seconds
```

maxwofford (approved):
> I think because we aren't exposing ports in the docker-compose.yml, the postgres instance will only be accessible within the docker network.

Sounds good, go for merge!

PR #2985, opened 2022-09-05
Upgrade postgres to 11.16 to match heroku production version with pgpassword

albertchae:
Alternative approach to https://github.com/hackclub/bank/pull/2984

Just changing the version led to this error
```
Error: Database is uninitialized and superuser password is not specified.
       You must specify POSTGRES_PASSWORD to a non-empty value for the
       superuser. For example, "-e POSTGRES_PASSWORD=password" on "docker run".

       You may also use "POSTGRES_HOST_AUTH_METHOD=trust" to allow all
       connections without a password. This is *not* recommended.

       See PostgreSQL documentation about "trust":
       https://www.postgresql.org/docs/current/auth-trust.html
```

This PR uses password for postgres instead of trust
We can hardcode and commit a password just for local dev, or check in a
blank environment variable and recommend that devs generate their own
password.

NB: env variable for setting up the DB is called POSTGRES_PASSWORD but
the one for accessing via pg_restore/psql is PGPASSWORD


config/database.yml
```diff
@@ -8,10 +8,12 @@ default: &default
 development:
   <<: *default
   database: bank_development
+  password: <%= ENV['PGPASSWORD'] %>
 
 test:
   <<: *default
   database: bank_test
+  password: <%= ENV['PGPASSWORD'] %>
```
albertchae: It turns out these aren't necessary because all `libpq` based tools read this environment variable https://stackoverflow.com/a/18282613, but maybe better to keep for explicitness?

docker-compose.yml
```diff
@@ -1,9 +1,13 @@
...
 services:
   db:
-    image: "postgres:11.4"
+    env_file:
+      - .env.docker
+    image: "postgres:11.16"
     volumes:
       - pg-data:/var/lib/postgresql/data
+    environment:
+      POSTGRES_PASSWORD: ${PGPASSWORD}
```
albertchae: there is a nuanced difference to these environment variables https://github.com/docker-library/postgres/issues/402#issuecomment-359915775

albertchae:
Closing in favor of https://github.com/hackclub/bank/pull/2984

PR #2987, opened 2022-09-05
Update missing HcbCode tracking

albertchae:
Follow up to https://github.com/hackclub/bank/pull/2863 and https://github.com/hackclub/bank/pull/2829

- Remove unnecessary lazy creation of hcb_code for canonical_pending_transactions
This count has been 0 for the past few weeks, which means the
after_create callback is successfully creating a hcb code every time.

```
> Ahoy::Event.where(name: "pendingTransactionMissingHcbCount").pluck(:properties).uniq
   (8.6ms)  SELECT "ahoy_events"."properties" FROM "ahoy_events" WHERE "ahoy_events"."name" = $1  [["name", "pendingTransactionMissingHcbCount"]]
=> [{"count"=>0}]
```

It should also be safe to run this on production after
`Ahoy::Event.where(name: "pendingTransactionMissingHcbCount").delete_all`
- Only count nil hcb_code as missing for canonical_transactions
The previous way would count "unknown" also, so it wasn't informative as
to whether https://github.com/hackclub/bank/pull/2863 was working


maxwofford (approved):
Awesome to see we're down to 0 instances!

PR #2989, opened 2022-09-05
Ignore migrated encrypted columns on comments and checks

albertchae:
Follow up to https://github.com/hackclub/bank/pull/2988

Similar to https://github.com/hackclub/bank/pull/2914


maxwofford (approved):
Added another comment about the fields in https://github.com/hackclub/bank/pull/2988, but after that change it's good!

albertchae:
@maxwofford I updated this PR with removing `transaction_memo` as discussed, can you take another quick sanity check even though it's already approved?

maxwofford (approved):

PR #2992, opened 2022-09-05
Re-enable skylight.io

albertchae:
- Revert "Remove skylight.io"
- Rename skylight.yml.erb to skylight.yml so we trigger the correct warning from skylight


albertchae:
@maxwofford I'm going to merge this so I can babysit if it sends events to skylight. I'll revert if it doesn't seem to work

albertchae:
I set `SKYLIGHT_AUTHENTICATION` in heroku to the correct value from skylight app settings

PR #2993, opened 2022-09-05
Delete skylight.yml to avoid "could not load config file"

albertchae:
Despite what I said in https://github.com/hackclub/bank/pull/2992,
apparently having a blank skylight.yml causes this error
```
/app/vendor/bundle/ruby/2.7.0/gems/skylight-5.3.3/lib/skylight/config.rb:269:in `load': could not load config file; msg=empty file (Skylight::ConfigError)
```

https://www.skylight.io/support/troubleshooting#authentication-token-required--could-not-load-config-file


albertchae:
hooray, it's working now 
image

PR #2995, opened 2022-09-06
Prevent fronting declined transactions

cjdenio:
This only seems to be affecting rejected disbursements at the moment.

There's currently one non-HQ event whose balance is higher than it should be— see [this disbursement](https://bank.hackclub.com/hcb/HCB-500-1116).


cjdenio:
Gonna hold off on merging since this'll make their balance negative

cjdenio:
beep boop

PR #3004, opened 2022-09-09
First-time activation for cards

cjdenio:
Part of #2994 

## to test
- manually issue a replacement card through the Stripe API or CLI
- import the replacement through the Rails console (see #3032 for an example)
- test it out


cjdenio:
Reviewed IRL by @maxwofford!

PR #3015, opened 2022-09-12
Calculate fees on fronted transactions

cjdenio:
Fixes #3014


app/models/event.rb
```diff
@@ -444,6 +444,20 @@ def fee_balance
...
+  def fronted_fee_balance_v2_cents
+    feed_fronted_balance = canonical_pending_transactions
+                           .incoming
+                           .fronted
+                           .not_declined
+                           .where(raw_pending_incoming_disbursement_transaction_id: nil) # We don't charge fees on disbursements
+                           .sum(&:fronted_amount)
+
+    # TODO: make sure this has the same rounding error has the rest of the codebase
+    fee_balance_v2_cents + (feed_fronted_balance * sponsorship_fee)
```
garyhtou: see `app/services/fee_engine/hourly.rb`

garyhtou (commented):
good work. See `app/services/fee_engine/hourly.rb` for existing fee calculation to match rounding.

cjdenio:
bleep borp

PR #3019, opened 2022-09-14
Don't show missing receipt list if no receipts are missing

exu3:
Screen Shot 2022-09-14 at 12 05 05 PM


app/views/static_pages/my_cards.html.erb
```diff
@@ -29,7 +29,7 @@
   <%= blankslate "No cards found!" %>
 <% end %>
 
-<%= async_frame_to my_missing_receipts_list_path %>
+<%= async_frame_to my_missing_receipts_list_path if @missing %>
```
cjdenio: Hmm, I think this check should be done inside the frame for performance.

The weird thing is, it already is?
https://github.com/hackclub/bank/blob/f542e14b7adc1d534a90af8a3bedc4a1f5cfb424/app/views/static_pages/my_missing_receipts_list.html.erb#L1


maybe we need a uniform `MissingReceiptService` so these endpoints all work the same way

maxwofford:
I'm really excited for this! What needs to happen for this to be ready for merge?

cjdenio:
Closing since I can't reproduce the original bug.

PR #3020, opened 2022-09-14
Make incoming fees show up on Bank

cjdenio:
1. A transaction is created from Operating to Main.
   - It has a _different memo_ than the original (`incoming_memo`), which includes an HCB code that's pre-mapped to the "Incoming Fees" org (see `hcb_code.rb`)
2. The magical short code mapper (`EventMappingEngine::Map::HcbCodes::Short`) handles the rest.
3. Profit
   - ###### seriously


app/models/hcb_code.rb
```diff
@@ -126,7 +126,8 @@ def events
         partner_donation.try(:event).try(:id),
         ach_transfer.try(:event).try(:id),
         check.try(:event).try(:id),
-        disbursement.try(:event).try(:id)
+        disbursement.try(:event).try(:id),
+        hcb_i1 == ::TransactionGroupingEngine::Calculate::HcbCode::INCOMING_BANK_FEE_CODE ? 1671 : nil # Map to "Incoming Fees"
```
maxwofford: Is there a way we can better document the intent of the magic number here? Is there a fallback when in development & not running with a prod database?

PR #3035, opened 2022-09-21
Refactor flavor text into `FlavorTextService`

cjdenio:
basically moves all the flavor text stuff into a service that works reliably everywhere, not just inside views.

(this makes https://bank.hackclub.com/api/v3/flavor work again!)

Fixes #2772


exu3 (approved):
This looks great to me, yum!

PR #3042, opened 2022-09-23
Replace StripeCardholders table in card_overview with a button to view cardholder in stripe

albertchae:
Closes https://github.com/hackclub/bank/issues/2554

The card_overview page (e.g. https://bank.hackclub.com/hq/cards) was very slow for admin users because there was a cardholders table on the bottom that would require N calls to stripe to show whether the cardholder was active. Per the conversation on https://github.com/hackclub/bank/issues/2554#issuecomment-1249927956, this table wasn't very useful for admin users, so instead remove that table and add a button per card to view cardholders directly on stripe.

image


maxwofford (approved):
Thanks! The admins will be happy with this.

PR #3047, opened 2022-09-26
More elegantly handle hard-coded event IDs

cjdenio:


exu3 (approved):

cjdenio:
@albertchae I'm wondering if there's a better way to handle these production-specific constants— I'd love your thoughts on this

albertchae (commented):
hey @cjdenio, 🤔 while this is a bit awkward, honestly I think it's fine. I've seen similar setups in other codebases before. The only alternative I could think of is tying it to event slug instead, but then you're relying on an event with that slug being in the database and also would require a DB call in each of these places, so it's not really better.

Maybe one thing we can do (especially if we work on https://github.com/hackclub/bank/issues/2965) is ensure that these IDs are mapped to the same events in the seed file so that most local environments will match

maxwofford (approved):
I'm happy with this as a way to bring everything into one place. +1 on having a working seed file though!

PR #3048, opened 2022-09-26
Anytime is an adverb

exu3:


cjdenio (approved):
verb is such a weird word... verb verb verb

exu3:
verb verb

PR #3050, opened 2022-09-26
Update connect_gofundme.html.erb

hackliv:
Since some processes have changed, this includes updates to the UI text about how users can connect their GoFundMe fundraisers to their orgs on Hack Club Bank. The PayPal Giving Fund is no longer a withdrawal option for more than one account on GoFundMe, so that has been removed as an option for users. The new, best way to connect fundraisers has been added as the recommended option. Since the original publishing, ops team has also noticed the first option does not deduct fees, so that language is updated as well.


exu3 (approved):
Looks good to me; feel free to merge when ready!

PR #3052, opened 2022-09-26
Eagerly create `HcbCode` for disbursements

cjdenio:
Fixes #3051


albertchae (approved):
This seems reasonable to me. How were they currently being created? A periodic background job?

cjdenio:
They're usually created on-demand by calling `local_hcb_code`, but that call was removed from the Transfers page for performance

PR #3056, opened 2022-09-27
Show session creation date

YodaLightsabr:
#3003
Adds the session creation date to the sessions list.
Before:
image
After:
image


cjdenio (changes requested):
This is great, thanks!

One piece of advice: the time currently always shows up as UTC, regardless of the user's timezone. `session.created_at.in_time_zone(session.timezone).strftime("...")` should work.

YodaLightsabr:
Thanks! I don't know how that slipped past me.

cjdenio (approved):

maxwofford (approved):
This is functional but should have a design pass– currently lists a bunch of info which is good, but the list of info is now overflowing the previous design we had for these cards. Go ahead and assign @cjdenio for review on this PR going forward. I'm good with whatever you two come up with.

maxwofford:
Wanted to check in here & see what has to happen to merge this!

YodaLightsabr:
I was thinking about a tooltip design, but it's not immediately clear that you should hover over the card.

I have another design idea that will require a bit of a rework- I can play around with it later today.

maxwofford:
Screenshot 2022-11-10 at 10 58 27

maxwofford:
@cjdenio @YodaLightsabr cool if we merge this? just found a method we're using in other places that automatically generates a nice human-readable human time, but also includes timestamp in local timezone on hover

YodaLightsabr:
This looks good!

PR #3057, opened 2022-09-27
Splash text promo code

YodaLightsabr:
#2947 Add a small easter egg to one of the promo texts linking to the stickers form.


maxwofford (approved):
Love it!

PR #3058, opened 2022-09-27
Fix wording in the HcbCode disbursement partial

cjdenio:
This:
- Fixes an edge case where fronted disbursements say "transferring" until the transfer is started on SVB
- Updates "Transferred at" to actually show when the transfer happened, _not_ when the disbursement was created
- Makes the state text on the right-hand side uppercase

Resolves #3037.


maxwofford (approved):
Looks great!

exu3 (approved):

PR #3061, opened 2022-09-27
Start adding AASM to disbursements

cjdenio:
This PR adds AASM to the Disbursement model, without actually making Bank rely on it. Everything still works the way it did before.

Part of #1294


maxwofford (approved):
Great work taking this on! From our discussion, it sounds like this is just implementing AASM state & in a future PR we'll take on replacing existing systems with the new methods.

Couple notes:
- I'd love to make it clear which methods are aasm related. Seems like `GSuite` implements their events prefixed with `mark_`– let's do that here!
- From our chat the next steps are 1) making a one-time job that backfills aasm data & 2) updating existing code to use aasm fields.

app/models/disbursement.rb
```diff
@@ -76,6 +78,41 @@ class Disbursement < ApplicationRecord
   scope :rejected, -> { where.not(rejected_at: nil) }
   scope :errored, -> { where.not(errored_at: nil) }
 
+  aasm create_scopes: false, whiny_transitions: false do
```
garyhtou: Do we plan to keep `whiny_transitions: false` for the long term? Or will it be removed once we fully migrate to Disbursements AASM?

The reason I'm asking is that we should keep things consistent. Other AASM models have `whiny_transitions: true` (the default)

app/models/disbursement.rb
```diff
@@ -112,7 +149,14 @@ def processed?
   def fulfilled?
     # two transactions, one coming out of source event and another
     # going into destination event
-    canonical_transactions.size == 2
+    fulfilled = canonical_transactions.size == 2
+
+    # Set AASM state
+    if may_mark_deposited? && fulfilled
+      mark_deposited!
```
garyhtou: Is there a way to `mark_deposited!` elsewhere? (maybe here https://github.com/hackclub/bank/blob/main/app/services/pending_event_mapping_engine/settle/incoming_disbursement_hcb_code.rb — there might be a better spot)

`fulfilled?` is kinda like an accessor method, we wouldn't really expect it to change/update anything in the database. In addition, if `fulfilled?` is rarely ran, then the AASM state may be "out of date" in the database, making SQL queries incorrect

app/services/disbursement_service/approve.rb
```diff
@@ -13,6 +13,7 @@ def run
       ActiveRecord::Base.transaction do
         # 1. Approve the disbursement
         @disbursement.update(attrs)
```
garyhtou: Seems like this `update` statement is setting `fulfilled_by`. it might be redundant now that we're calling `mark_approved!`

app/models/disbursement.rb
garyhtou: We might wanna also say `timestamps: true`, this way it'll fill in fields such as `fulfilled_at`, `errored_at`, etc.

unless we plan to drop these fields in the future

app/views/admin/disbursements.html.erb
```diff
@@ -63,6 +64,9 @@
             AUTOMATED
           <% end %>
         </td>
+        <td>
+          <%= disbursement.aasm_state %>
```
garyhtou:
```suggestion
          <%= disbursement.aasm.human_state %>
```

app/models/disbursement.rb
cjdenio: This is just to make sure Bank doesn't die if there's a bug that causes an invalid state change

cjdenio: Good point— I originally had it off because of a discrepancy between `fulfilled_at` and the `fulfilled` state, but the state's called `deposited` now

config/schedule.yml
```diff
@@ -55,6 +55,11 @@ disbursement_nightly_job: # PERFORMS SVB TRANSFERS
   class: "DisbursementJob::Nightly"
   queue: "default"
 
+disbursement_hourly_job: # PERFORMS SVB TRANSFERS
+  cron: "*/30 * * * *" # run every 30 minutes
```
garyhtou: Since this job performs an SVB transfer, we'll want to stagger it from other jobs that perform an SVB transfer (such as `bank_fee_nightly_job`). Maybe we can run it at `15,45 * * * *`.

This is cause SVB transfers are done with selenium and we likely don't want to concurrently login to SVG.

app/services/disbursement_service/nightly.rb
```diff
@@ -30,11 +30,13 @@ def run
           # there was an error so mark this disbursement as in a bad/mixed state
           Airbrake.notify("Disbursement #{disbursement.id} in mixed/bad error state. Partially processed remotely. Investigate and fix by hand.")
           disbursement.update_column(:errored_at, Time.now)
```
garyhtou: we might be able to remove the `errored_at` column update since AASM will handle updating timestamps for us

app/services/disbursement_service/hourly.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module DisbursementService
+  class Hourly
+    def run
+      Disbursement.where(aasm_state: :in_transit).each do |disbursement|
+        if disbursement.canonical_transactions.size == 2
+          disbursement.mark_deposited!
+        end
```
garyhtou: optional: this would be a great place to flag (Airbrake) disbursements that have > 2 CTs. otherwise, they may sit undiscovered until someone mentions that their disbursement has been "in_transit" forever

app/services/disbursement_service/reject.rb
```diff
@@ -11,6 +11,7 @@ def run
       raise ArgumentError, "Disbursement is already processed" unless @disbursement.reviewing?
 
       @disbursement.update!(attrs)
```
garyhtou: I'm not sure if this `update!` is necessary anymore now that the logic has been moved into `mark_rejected!`

garyhtou (approved):
looks good! just a couple of small things

config/schedule.yml
cjdenio: oh I accidentally copied the `# PERFORMS SVB TRANSFERS` comment 😆

app/services/disbursement_service/reject.rb
cjdenio: I'm gonna leave this since I'm aiming to change as little existing code as possible

cjdenio:
wheeee

PR #3067, opened 2022-09-29
 Unskip g_suite_mailer_spec and convert fixtures to factories

albertchae:

The main issue here is that the URL being checked didn't conform to the actual structure for the event's gsuite overview. With factories we have to make it more dynamic, but because the implementation is using the `event_g_suite_overview_url` helper, it'd be kind of a tautological test, so I checked for an expected URL built with File.join instead.


albertchae:
Per slack discussion with @maxwofford , will merge this PR because it has structural changes only.

PR #3068, opened 2022-09-29
Unskip auth_service/token_spec and convert fixtures to factories

albertchae:

Just unskipping this test by itself would fail the 2 checks for
`UnauthenticatedError` when `login_token` didn't exist, which is exactly
what this begin-rescue-ensure block seems to be trying to implement, so
I was confused for a while.

It turns out if `login_token` didn't exist, then the rescue would run, but
then before the `raise UnauthenticatedError`, the ensure block would try
to run and throw `ActiveRecord::RecordNotFound` yet again when doing the
`used?` check. It seems like there's no point in running the code in the
ensure block if the `login_token` doesn't already exist, so I moved it
into the normal method body before returning user.

I also wanted to check that user (which runs login_token.user) wouldn't
also possibly be a culprit for `ActiveRecord::RecordNotFound`, but
because `login_tokens` has a FK constraint on the db with `users`, this
should never happen.

I ended up removing the inline rescue because it only seems to be
catching an invalid AASM transition error which I wasn't sure if it made
sense to do. The scary thing is that inline rescue can apparently catch
a lot of errors that we usually didn't intend to. Reference:
https://thoughtbot.com/blog/don-t-inline-rescue-in-ruby

The rest of the PR is creating a login token factory and using that in the login token model spec and the auth_service/token spec


app/services/auth_service/token.rb
```diff
@@ -11,12 +11,11 @@ def initialize(token:, ip: nil)
 
     # @return [[User, TrueClass | FalseClass]]
     def run
+      @already_used = login_token.used?
+      login_token.mark_used!(@ip)
```
albertchae: this line previous had an inline `rescue nil`, which I think was intended to catch an invalid AASM transition (`used` to `used`). I'm not entirely sure what's semantically correct here, but if we want to treat previously used login tokens as unauthenticated, I think we should instead add a line above this `return if @already_used`

garyhtou: I like what you have here. Yeah, the intention of `rescue nil` was to catch an invalid AASM transition (used to used).

Maybe we can say
```suggestion
      login_token.mark_used!(@ip) unless @already_used
```

garyhtou (approved):
Good catch with the weird `ensure`!

app/services/auth_service/token.rb
albertchae: Sounds good. I like that instead of the guard because it keeps us returning the `user` from this method. Another possibly is to make `used` to `used` a legal transition, but I guess that could potentially allow some bad logic to continue silently, whereas here we have to explicitly check and skip changing state.

PR #3071, opened 2022-09-29
Pass user birthdays to Stripe

cjdenio:
part of #2775


maxwofford (approved):

garyhtou:
Just wondering, do we know how Stripe will behave if we pass a `dob` that's < 18 years old?

maxwofford:
@garyhtou I totally didn't think of that. @cjdenio – could you test this out in prod? We might need a conditional to skip providing this info in that case.

maxwofford:
I just updated my own & it didn't stop me. Will report if my account is deactivated.

Screen Shot 2022-10-05 at 15 40 13

exu3:
There are cardholders on Stripe currently that we have manually entered in DOBs for that are under 18. For example: https://dashboard.stripe.com/issuing/cardholders/ich_1LlBcYFSaumjmb9r2A7Utnps

cjdenio:
It doesn't work if you're under 13:

Screen Shot 2022-10-05 at 3 41 15 PM

maxwofford:
@cjdenio I don't think we have anyone under 13 on Bank, but go ahead and add a condition w/ comment that links to this thread for context.

PR #3073, opened 2022-09-30
Identify users on FullStory

cjdenio:
The big thing here is tracking who's an admin so we can exclude them in some metrics. It'll also be nice for FullStory to reliably identify sessions as belonging to the same user.


maxwofford (approved):
Looks great!

PR #3076, opened 2022-10-03
Associate HashedTransaction with CanonicalTransaction during create, not afterwards

albertchae:
Fixes https://github.com/hackclub/bank/issues/3028

See https://github.com/hackclub/bank/issues/3028#issuecomment-1264771076 specifically for what the bug was

**Best reviewed by commits**

Several commits are to make the change easy, before the last 2-3 commits make the easy change
https://twitter.com/kentbeck/status/250733358307500032?lang=en

We first improve test setup to use factories instead of fixtures for several deeper nodes in the object graph (e.g. `raw_plaid_transaction`, `raw_emburse_transaction`, `hashed_transaction`), then I fixed a bug in the previous definition of `canonical_transaction_factory`, then more fixtures to factories. Finally, we are able to write a failing test for the airbrake issue, then fix the implementation. Unfortunately there is some test pollution going on so the last commit addresses that by destroying any record in the DB that could interfere with the test.

- Convert raw_plaid_transaction_spec to factories only
- Add raw_emburse_transaction_spec with factory
- Add emburse trait to hashed_transaction and remove fixtures from hashed_transaction_spec
- Fix canonical_transaction_factory to create hashed_transactions association correctly
- Convert canonical_hashed_mapping_spec to factories
- Improve test for hcb_code eager creation
- Convert canonical_transaction_spec from fixtures to factories
- Create failing test for CanonicalTransactionService::Import::Simple which reproduces the Airbrake error
- Attach HashedTransaction to CanonicalTransaction during create so HcbCode can be created correctly
- Hackily clean up the DB before running import test


maxwofford (approved):
This looks great! Love the approach here. Sounds like the db cleanup we're doing gets fixed once we convert more of our tests over to factories, so that seems totally fine for now.

PR #3079, opened 2022-10-03
Add foreman in gemfile and use bundle to run it

albertchae:
This way after a ./docker_setup.sh, we know the docker container has all
the dependencies we need to run a local instance

I ran into this during a flight yesterday so I thought this might be a
small improvement.


cjdenio (approved):
Thanks!

PR #3083, opened 2022-10-03
Unskip fee_relationship_spec and fix

albertchae:

First 2 commits are adding dependent factories for `bank_account` and `transaction` (and updating their model specs at the same time), then the last one creates a `fee_relationship` factory and uses it to refactor the test significantly.

- Replace bank_account fixture with factory
- Replace transaction fixture with factories
- Replace fee_relationship fixture with factory


spec/factories/transaction_factory.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :transaction do
+    association :bank_account
+    amount { Faker::Number.number(digits: 4) }
```
albertchae: should we add a validates presence of this field? It isn't required now but fee_relationship would fail without it

spec/models/fee_relationship_spec.rb
```diff
@@ -3,73 +3,57 @@
...
-  let(:fee_relationship) { fee_relationships(:fee_relationship1) }
+  let(:fee_relationship) { create(:fee_relationship) }
 
   it "is valid" do
     expect(fee_relationship).to be_valid
   end
 
   describe "fee_amount" do
-    before do
-      fee_relationship.update_column(:fee_amount, nil) # hacky because of the design of this model and schema. validations that are calculating values should move to a service in the future
```
albertchae: I agree with this comment but I removed this before/do since it was doing nothing. the default value of fee_amount is already nil

spec/models/fee_relationship_spec.rb
```diff
@@ -3,73 +3,57 @@
...
-        expect(fee_relationship.fee_amount).to eql(10010)
-      end
-    end
-
-    context "when fee_applies is false" do
-      before do
-        fee_relationship.update_column(:fee_applies, false)
-      end
+      let(:transaction) { create(:transaction, amount: 100) }
+      let(:event) { create(:event, sponsorship_fee: 5) }
```
albertchae: factories let us more obviously see what values get used in the business logic being tested

spec/models/fee_relationship_spec.rb
```diff
@@ -3,73 +3,57 @@
...
+      expect(fee_relationship.fee_amount).to be_nil
     end
 
     context "when before_validation is called" do
-      it "sets value to transaction.amount" do
-        expect(fee_relationship).to receive(:calculate_fee).and_call_original
-
-        fee_relationship.save # calls calculate_fee
-
-        expect(fee_relationship.fee_amount).to eql(10010)
```
albertchae: this test was lying. 10010 is coming off a previously set fee_amount https://github.com/hackclub/bank/blob/b46cc4ad8afe6cf2d963578cffb611c382c005dc/spec/fixtures/fee_relationships.yml#L3

spec/models/fee_relationship_spec.rb
```diff
@@ -3,73 +3,57 @@
...
-      it "sets to nil" do
-        expect(fee_relationship).to receive(:calculate_fee).and_call_original
+      context "when fee_applies is true" do
+        let(:fee_applies) { true }
 
-        fee_relationship.save # calls calculate_fee
+        it "calculates fee_amount off the transaction and sponsorship fee" do
+          expect(fee_relationship).to receive(:calculate_fee).and_call_original
 
-        expect(fee_relationship.fee_amount).to eql(nil)
```
albertchae: this test was also lying. fee_applies=false does not set fee_amount to nil, but leaves it unchanged.

spec/factories/transaction_factory.rb
maxwofford: This should have a validates, I think that was just an oversight.

maxwofford (approved):

spec/factories/transaction_factory.rb
albertchae: added!

albertchae: Whoops, should have let rspec finish before merging. Fix here https://github.com/hackclub/bank/pull/3187

PR #3086, opened 2022-10-03
 Unskip generate_login_token_spec and fix tests 

albertchae:

- Unskip generate_login_token_spec and fix tests
  - login_url format changed
  - organization_identifier was changed to public_id, so change tests to
    match and also implementation to throw an exception when not found to
    match the test
  - one test is no longer relevant because partner is passed in instead of
    partner_id
- replace fixtures with factories
  - Add a different test than missing partner_id from the previous commit,
  where the partner is not affiliated with the event


albertchae:
Merging this because it is mostly structural changes @maxwofford

PR #3087, opened 2022-10-03
Drag-and-drop event reordering

cjdenio:
Adds the ability for users to re-order the organizations on their homepage.


db/migrate/20221003160538_add_sort_index_to_organizer_position.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddSortIndexToOrganizerPosition < ActiveRecord::Migration[6.1]
+  def change
+    add_column :organizer_positions, :sort_index, :float
```
maxwofford: Seems like this can just be a regular-old integer.

app/javascript/controllers/event_sort_controller.js
```diff
@@ -0,0 +1,31 @@
...
+
+    const id = copy[oldIndex]
+
+    copy.splice(oldIndex, 1)
+    copy.splice(newIndex, 0, id)
+
+    this.organizerPositionsValue = copy
+
+    const csrfToken = document
+      .querySelector('meta[name="csrf-token"]')
```
maxwofford: Seems like we might already have this available to us in the codebase. Should this be using a helper or util?

app/controllers/organizer_positions_controller.rb
```diff
@@ -16,4 +16,29 @@ def destroy
...
+
+    index = params[:index]
+
+    # get all the organizer positions as an array
+    organizer_positions = StaticPageService::Index.new(current_user: current_user).organizer_positions.not_hidden.to_a
+
+    return render status: :bad_request if index < 0 || index >= organizer_positions.size
+
+    # move the organizer position to the desired index
+    organizer_positions.delete organizer_position
```
maxwofford: This looked like an AR deletion from a quick skim. I see the `to_a` [above](https://github.com/hackclub/bank/pull/3087/files#diff-ececbb94349d1077a96c46f9be7f778976c99539393f4388620aa061ec8a3e9bR26), but I wonder if there's a way to make it more immediately clear we aren't reordering AR's index

app/views/static_pages/index.html.erb
```diff
@@ -14,9 +14,15 @@
 <main class="flex flex-column items-center center pb3">
   <% if signed_in? %>
     <% if @events.any? %>
-      <ul class="grid grid--<%= "medium-" if Flipper.enabled?(:detailed_events_list_2022_10_04, current_user) %>narrow left-align w-100 mt0">
+      <ul
+        class="grid grid--<%= "medium-" if Flipper.enabled?(:detailed_events_list_2022_10_04, current_user) %>narrow left-align w-100 mt0"
+        data-controller="<%= "sortable event-sort" if Flipper.enabled?(:event_reordering_2022_10_10, current_user) %>"
```
maxwofford: I don't see anything in the UI to enable this. Should we include it on the `my/settings#feature-flags` page?

maxwofford (commented):
~~I have the branch checked out & the feature enabled, but it doesn't seem to work?~~

Nvm, this was my devserver not recompiling js

https://user-images.githubusercontent.com/5891442/195198268-840f324e-121b-4f65-a64c-a4d4de42b22a.mov

app/views/static_pages/index.html.erb
cjdenio: The feature flag is just so it can be disabled quickly in case something breaks. I don't think it's worth making this opt-in since it's pretty unobtrusive

maxwofford: I think this would be sick to roll into @YodaLightsabr's "event page overhaul" changes from a user perspective

maxwofford: Feel free to merge without either though, just a thought

cjdenio:
blorp

app/controllers/organizer_positions_controller.rb
```diff
@@ -16,4 +16,29 @@ def destroy
...
+    # switch the position *in the in-memory array*
+    organizer_positions.delete organizer_position
+    organizer_positions.insert index, organizer_position
+
+    # persist the sort order
+    ActiveRecord::Base.transaction do
+      organizer_positions.each_with_index do |op, idx|
+        op.update(sort_index: idx)
+      end
+    end
```
garyhtou: I think we need some kind of Mutex Lock here so that code does not run concurrently.

The lock would lock (start) before u get the array. and unlock after the transaction. That section of code is called the critical section and should be as small as possible.

https://github.com/hackclub/bank/issues/3146

garyhtou: @cjdenio this dives into multiprocessing and concurrency/parallelism. lmk if you're stuck here

app/controllers/organizer_positions_controller.rb
```diff
@@ -16,4 +16,29 @@ def destroy
...
+
+    # switch the position *in the in-memory array*
+    organizer_positions.delete organizer_position
+    organizer_positions.insert index, organizer_position
+
+    # persist the sort order
+    ActiveRecord::Base.transaction do
+      organizer_positions.each_with_index do |op, idx|
+        op.update(sort_index: idx)
+      end
```
garyhtou: We don't need to update every organizer's position. Only the ones that have changed.

if `i` = old position and `j` = new position, then only update the ones from `[min(i, j), max(i, j)]` (inclusive).

PR #3098, opened 2022-10-06
Unskip raw_pending_partner_donation_transaction_spec and replace fixtures with factories

albertchae:

This spec passed without any changes after unskipping, but I replaced
the fixtures with factories


albertchae:
Per slack conversation with @maxwofford , will merge this one because it's only a structural change

PR #3102, opened 2022-10-06
Create fee revenue transactions

cjdenio:
This PR makes Bank aggregate weekly fees into a single "Fee revenue" transaction that gets mapped to https://bank.hackclub.com/bank.

Screen Shot 2022-10-06 at 2 28 24 PM
Screen Shot 2022-10-06 at 2 27 09 PM

Things that have been tested:

- [x] `FeeRevenue` objects are created weekly
- [ ] Fee revenue TXs are properly created on SVB
  - [x] It generates the correct memo + amount
  - [ ] it actually transfers
- [x] Those transactions are mapped to /bank
- [x] `FeeRevenue` objects are marked as settled
- [x] `HcbCode#show` works for fee revenue TXs
- [x] Fee revenue TXs have fancy memo + icon


app/models/fee_revenue.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def hcb_code
+    "HCB-#{::TransactionGroupingEngine::Calculate::HcbCode::FEE_REVENUE_CODE}-#{id}"
+  end
+
+  def local_hcb_code
+    @local_hcb_code ||= HcbCode.find_or_create_by(hcb_code: hcb_code)
+  end
+
+  def canonical_transaction
```
maxwofford: Know we do this in other places, but if we expect only 1 let's start switching back to regular rails conventions ie. `find_by`

app/models/fee_revenue.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+  end
+
+
+  def hcb_code
+    "HCB-#{::TransactionGroupingEngine::Calculate::HcbCode::FEE_REVENUE_CODE}-#{id}"
+  end
+
+  def local_hcb_code
+    @local_hcb_code ||= HcbCode.find_or_create_by(hcb_code: hcb_code)
```
maxwofford: Because we're working towards eagerly creating HCB codes (ex. https://github.com/hackclub/bank/issues/2783), do we need to be doing this, or can we have an AR association?

maxwofford (approved):
Chatted IRL!

PR #3104, opened 2022-10-07
Associate HashedTransaction with CanonicalTransaction during create, not afterwards (part 2)

albertchae:
https://github.com/hackclub/bank/issues/3103

**Best reviewed by commits**

Turns out there are 3 other import services that need similar fixes to https://github.com/hackclub/bank/pull/3076

The first 2 were straightforward (`EmburseThatLookLikeDuplicates` and `StripeThatLookLikeDuplicates`). They seem to ignore the "potentially duplicate" issue and just create `CanonicalTransaction`s the same as the `Simple` service does.

The third one `PlaidThatLookLikeDuplicates` looks like it was intending to skip certain duplicate transactions but I believe it may not be functioning as intended. See my comment on that file. This is why this PR currently has a failing test.

- Test and fix emburse_that_look_like_duplicates
- Create raw_stripe_transaction spec and factory, and add as trait to hashed_transaction_factory
- Test and fix stripe_that_look_like_duplicates
- Create test for plaid_that_look_like_duplicates
- Fix plaid_that_look_like_duplicates, exposes that duplicates are being handled the same


spec/services/transaction_engine/canonical_transaction_service/import/plaid_that_look_like_duplicates_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+        # transactions are duplicate if they have same hash but different raw_plaid_transactions
+        create(:hashed_transaction, primary_hash: duplicate_hash, raw_plaid_transaction: create(:raw_plaid_transaction))
+      end
+    end
+
+    it 'does not create any CanonicalTransactions' do
+      expect do
+        service.run
+      end.to change(CanonicalTransaction, :count).by(0)
+    end
```
albertchae: This test is currently failing. See my comment on the service implementation for why

app/services/transaction_engine/canonical_transaction_service/import/plaid_that_look_like_duplicates.rb
```diff
@@ -13,20 +13,13 @@ def run
 
             next unless not_actual_duplicate(ht)
```
albertchae: I'm having a hard time parsing the double negative here mentally and would like to refactor it to say

`next if actually_duplicate(ht)`


However, I also think the code for the current `not_actual_duplicate` is incorrect. I believe it'll return true every time because there will always be at least one `HashedTransaction` that satisfies that criteria, namely the one that was passed into the method.

I believe this block of code
```ruby
::HashedTransaction.where(raw_plaid_transaction_id: hashed_transaction.raw_plaid_transaction_id,
primary_hash: hashed_transaction.primary_hash).any?
```
should be written as
```ruby
::HashedTransaction.where(raw_plaid_transaction_id: hashed_transaction.raw_plaid_transaction_id,
primary_hash: hashed_transaction.primary_hash).count > 1
```

albertchae: @maxwofford I investigated how many instances of these are on prod. Fortunately there aren't any:

```
bank-hackclub::DATABASE=> select raw_plaid_transaction_id, count(*) from hashed_transactions where raw_plaid_transaction_id is not null group by raw_plaid_transaction_id, primary_hash having count(*) > 1;
 raw_plaid_transaction_id | count
--------------------------+-------
(0 rows)
```

spec/services/transaction_engine/canonical_transaction_service/import/plaid_that_look_like_duplicates_spec.rb
albertchae: This test passes now because we changed the implementation to properly detect plaid duplicates

albertchae: There is still an open question whether this should be 1 or 0.

app/services/transaction_engine/canonical_transaction_service/import/stripe_that_look_like_duplicates.rb
```diff
@@ -6,20 +6,13 @@ module Import
...
-              }
-              ::CanonicalHashedMapping.create!(attrs)
-            end
+            attrs = {
+              date: ht.date,
+              memo: ht.memo,
+              amount_cents: ht.amount_cents,
+              canonical_hashed_mappings: [CanonicalHashedMapping.new(hashed_transaction: ht)]
+            }
+            ::CanonicalTransaction.create!(attrs)
```
albertchae: Just to confirm that the most recent incident of this on airbrake was due to stripe, not emburse or plaid

https://heroku-app97991095.airbrake.io/projects/288439/groups/3369663817220586731/notices/3379085929442358847?filters=%7B%22order%22%3A%22last_notice%22%2C%22resolved%22%3A%22any%22%2C%22search%22%3A%22hashed_transaction%22%7D&tab=notice-detail

spec/services/transaction_engine/canonical_transaction_service/import/plaid_that_look_like_duplicates_spec.rb
maxwofford: This should be ~~0~~ 1. To give an example:

- admin Alice uses their SVB account for plaid, so HCB takes in raw TX data
- admin Alice changes their SVB password, which causes plaid syncing to fail
- admin Bob steps in to fix by using their SVB account for plaid, so HCB takes in raw TX data with the same values, but a different `raw_plaid_transaction_id`
- for a given charge in SVB, the transaction engine ends up with 2 `RawPlaidTransaction`s, 2 `HashedTransaction`s, but only 1 `CanonicalTransaction` to prevent the double sync creating duplicate charges

EDIT: Fixed my off-by-one error

albertchae: > only 1 CanonicalTransaction to prevent the double sync creating duplicate charges

By your reasoning here I think this should be 1. Because what's happening in the current implementation is that both `HashedTransaction`s sees each other as a duplicate and doesn't create the `CanonicalTransaction`, so none get created in the end.

albertchae: discussed this with @maxwofford on zoom, we want 1 here

albertchae: The fix for this is in commit https://github.com/hackclub/bank/pull/3104/commits/a7dc81a67f452843e9ed7aa8eb3122785e713a6e. Basically when checking for duplicates, we also check whether they have been processed before skipping the current one. If they haven't been processed, it's safe to process the current hashed transaction.

app/services/transaction_engine/canonical_transaction_service/import/plaid_that_look_like_duplicates.rb
```diff
@@ -52,11 +40,23 @@ def previously_processed_ht_ids
...
-        def not_actual_duplicate(hashed_transaction)
-          # true if it's in the same plaid account
-          # false if it's in a different plaid account
-          ::HashedTransaction.where(raw_plaid_transaction_id: hashed_transaction.raw_plaid_transaction_id,
-                                    primary_hash: hashed_transaction.primary_hash).any?
+        def actual_duplicate_already_processed(hashed_transaction)
+          # For some reason we imported transactions from the same bank
+          # account through 2 different plaid connections. Some historical reasons for this happening
+          # - Plaid authentication info changed
+          # - ????
```
albertchae: @maxwofford when we discussed this 2 weeks ago I remember you described another scenario here but I forgot the details. Can you recall? otherwise i'll delete this placeholder

app/services/transaction_engine/canonical_transaction_service/import/plaid_that_look_like_duplicates.rb
maxwofford: I think this will only show up once we start syncing transactions using 2 separate plaid instances running on the same underlying SVB bank account ([outlined in this comment](https://github.com/hackclub/bank/pull/3104#discussion_r995992240)). I'm on the same page about switching `.any?` -> `.count > 1`.

app/services/transaction_engine/canonical_transaction_service/import/plaid_that_look_like_duplicates.rb
maxwofford: Second scenario is that Plaid has downtime so we manually import using CSVs from SVB, leading to duplicates once Plaid is back up and running.

maxwofford (approved):

PR #3108, opened 2022-10-08
Unskip event spec and fix

albertchae:
### Unskip event_spec and fix

- default state is now approved https://github.com/hackclub/bank/pull/1938
- there were two transactions of 100 and 300, so balance_v2_cents was 400
- eq is preferred over eql for comparing numbers because it'll do the
  proper type conversions

### Replace fixtures with factories in event_spec
- plaid_id must be unique on transaction
- created the appropriate fee_relationships for the private fee methods


albertchae:
@maxwofford this is the last set of skipped specs 🥁

maxwofford (approved):

PR #3112, opened 2022-10-09
Remove new_framework_defaults initializers for 5.2, 6.0, and 6.1

albertchae:
These files are generated by the `rails app:update` task as a way to
incrementally set new default config values. According to
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#configure-framework-defaults and
https://discuss.rubyonrails.org/t/configure-framework-defaults-difficulty-with-wording-in-the-guides/80649,
after the load_defaults line in application.rb is updated to a
particular version, we can delete the initializer files.

Bank currently loads 6.1 defaults https://github.com/hackclub/bank/blob/96cd5972e6536dd0d0bc9ced7e1eb7aa42e6bbdd/config/application.rb#L14

I verified that the config values in these initializers matched the
intended default values as documented in
https://guides.rubyonrails.org/configuring.html#versioned-default-values

and actual reference to the load_defaults code
https://github.com/rails/rails/blob/6-1-stable/railties/lib/rails/application/configuration.rb

A few exceptions:
From new_framework_defaults_6_1
`Rails.application.config.active_job.skip_after_callbacks_if_terminated = true` is not documented in the guide, but in the 6.1 rails/application/configuration.rb, we see it being set to true.

From new_framework_defaults_6_0
`Rails.application.config.action_dispatch.return_only_media_type_on_content_type = false`
`Rails.application.config.active_job.return_false_on_aborted_enqueue = true`
are not documented nor are they in the rails/application/configuration.rb.
I'm not sure why and they did seem to exist in the past (documented in
this blog post
https://lilyreile.medium.com/rails-6-0-new-framework-defaults-what-they-do-and-how-to-safely-uncomment-them-586146f371e8),
but if it's not in the rails default configuration.rb anymore, it seems
safe enough...


albertchae:
@maxwofford I came across while starting look at the rails 7 upgrade. Let me know if you disagree with any of my reasoning.

maxwofford (approved):
Yep! Thanks for catching this

PR #3116, opened 2022-10-11
SPOOKY MODE

cjdenio:
i'm shivering in fright
Screen Shot 2022-10-11 at 1 52 41 PM


app/helpers/seasonal_helper.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module SeasonalHelper
+  SEASONS = {
+    # format:
+    # [[start_month, start_day], [end_month, end_day]]
+
+    fall: [[10, 1], [10, 31]],
+    holiday: [[12, 1], [12, 31]],
```
maxwofford: Maybe we could update `holiday` -> `winter` so it's a bit more clear it's snow-themed?

app/services/flavor_text_service.rb
```diff
@@ -11,6 +12,7 @@ def initialize(user: nil, env: Rails.env)
   def generate
     return development_flavor_texts.sample if @env == "development"
     return holiday_flavor_texts.sample if holiday?
+    return spooky_flavor_texts.sample if fall?
```
maxwofford: What if we did 50% chance of spooky w/ 50% chance of regular so it's not cycling through the same 5 flavors for the month?

app/views/application/_header.html.erb
```diff
@@ -26,21 +26,25 @@
       class: "header-center text-decoration-none flex items-center #{'tooltipped tooltipped--s' if !Rails.env.production? || user_birthday?}",
       'aria-label': user_birthday? ? "Happy birthday! #{['🎂', '🎈', '🎉', '🥳'].sample}" : "You’re in #{Rails.env} mode. #{development_mode_flavor}",
       data: {  "controller" => "fullstory", "fullstory-event-name-value" => "Icon right-clicked", "action" => "contextmenu->menu#toggle contextmenu->fullstory#event click@document->menu#close keydown@document->menu#keydown", "menu-target" => "toggle" } do %>
-      <%= image_tag user_birthday? ? "logo-party-dark.gif" : "logo-#{Rails.env}-dark.svg",
+      <%= image_tag user_birthday? ? "logo-party-dark.gif" : by_season("logo-#{Rails.env}-dark.svg", fall: "https://cloud-qmskqc293-hack-club-bot.vercel.app/0bank-pumpkin.png"),
```
maxwofford: This area is starting to feel kinda overloaded. `by_season` itself is a helpful method, but we're already doing tertiary statements, string interpolation, and duplicating the `image_tag`. Wonder how we can simplify this?

app/views/application/_header.html.erb
```diff
@@ -26,21 +26,25 @@
...
+      <%= image_tag user_birthday? ? "logo-party.gif" : by_season("logo-#{Rails.env}.svg", fall: "https://cloud-qmskqc293-hack-club-bot.vercel.app/0bank-pumpkin.png"),
           name: 'header-logo',
           class: 'logo',
-          width: @home_size || 36,
           height: @home_size || 36,
           alt: 'Bank logo' %>
     <% end %>
 
     <div data-menu-target="content" class="menu__content menu__content--2 menu__content--compact">
+      <% if fall? %>
```
maxwofford: <3

Love the attention to detail

app/helpers/seasonal_helper.rb
```diff
@@ -0,0 +1,41 @@
...
+    define_method "#{key}?" do
+      today = Date.today
+      low = SEASONS[key][0]
+      high = SEASONS[key][1]
+
+      if today.day >= low[1] && today.day <= high[1] && today.month >= low[0] && today.month <= high[0]
+        true
+      else
+        false
+      end
```
garyhtou:
```suggestion
      today.day >= low[1] && today.day <= high[1] && today.month >= low[0] && today.month <= high[0]
```
we can drop the if statement here and leave the condition to be returned

garyhtou: Also, try out the `between?` or `covers?` methods!

PR #3122, opened 2022-10-13
Disable event sorting on mobile

cjdenio:
There's currently a bug that means you can't scroll while your finger's on an event card.

weird stuff

will figure out later


cjdenio:
bonk

PR #3123, opened 2022-10-13
Let users toggle seasonal themes

cjdenio:
Screen Shot 2022-10-13 at 9 57 14 AM
Screen Shot 2022-10-17 at 2 57 54 PM


maxwofford:
@cjdenio don't want to overscope this, but how about we just make this a setting for all users?

cjdenio:
doot

PR #3129, opened 2022-10-20
Map outgoing ACH transactions using confirmation number

cjdenio:
Fixes #2403


app/models/canonical_transaction.rb
```diff
@@ -226,6 +226,10 @@ def ach_transfer
     nil
   end
 
+  def likely_ach_confirmation_number
+    memo.match(/BUSBILLPAY TRAN#(\d+)/)&.[](1)
```
garyhtou: Great job accounting for `nil`!

garyhtou (approved):
Looks great!

Do we have a plan for back-filling in ACH transfers that are in the `000` HCB group? It would also be nice to check for transactions in the `300` HCB group just to make sure they have the right confirmation number (this would be a one-time thing).

cjdenio:
Transactions < 1 month old should be automatically fixed by the `TransactionGroupingEngine`— I'll just pass a different `start_date` in the Rails console to fix the older ones.

cjdenio:
There are now 79 ungrouped ACH transactions— I think it's because some older `AchTransfers` don't have a confirmation number

```
irb(main):026:0> CanonicalTransaction.likely_achs.missing_or_unknown_hcb_code.count
=> 79
```

Note to self: the settling logic still relies on janky `amount` matching

cjdenio:
> ~~WHY~~
> 
> ~~there is ONE transaction whose memo doesn't have a confirmation number and it happened YESTERDAY~~
> 
> 
> ```ruby
> irb(main):047:0> CanonicalTransaction.likely_achs.where("memo not like '%TRAN#%'").count
> => 1
> 
> irb(main):048:0> CanonicalTransaction.likely_achs.where("memo not like '%TRAN#%'").first.memo
> => "ALEXANDRA MITCHE BUSBILLPAY"
> 
> irb(main):049:0> CanonicalTransaction.likely_achs.where("memo not like '%TRAN#%'").first.date
> => Wed, 19 Oct 2022
> ```

Looks like this happens all the time but Bank just fixes itself

garyhtou:
> Looks like this happens all the time but Bank just fixes itself

Yeah, once the conf # is included by SVB in the memo, plaid picks it up and it shows up on Bank

garyhtou:
> I'll just pass a different `start_date` in the Rails console to fix the older ones.

Has this been done?

Here's an easy way to see the remaining ACHs in the 000 HCB Code
https://bank.hackclub.com/blazer/queries/264-wack-ach-hcb-group

cjdenio:
I just fixed all the ones from the past year— there are currently 3 ungrouped ACHs that are over a year old.

```ruby
CanonicalTransaction.likely_achs.missing_or_unknown_hcb_code.filter {|t| t.ach_transfer.present?}.size
=> 3
```

A lot of them (59 to be exact) can't easily be auto-grouped because they don't have `AchTransfer`s with matching confirmation numbers.

PR #3135, opened 2022-10-21
Improvements to dev setup

cjdenio:
### ⚠️ Breaking change

Developers using the production database need to set `USE_PROD_CREDENTIALS=true` when running Bank.

---

- Disabled Sidekiq in development (because it requires Redis)
- Created a credentials file specifically for development: (Bank won't boot without a few secrets/config options)
```yaml
secret_key_base: 00f0c2444a37a2538f4083070c33ca621e1c72f03db214362487914f34da6542

default_url_host:
  live: bank.hackclub.com

smtp:
  username:
  password:
  address:
  domain:
  port:

ipinfo_io:
  api_key:

lockbox:
  master_key: 6eb9a8a92d091b15092a89c7ac5816d7cc1b5840bfdf73eab9a9e16a7fd51e51
```
- Created a seed file that:
  - Makes the user an admin
  - Invites them to a test org
- Made users default to admin, regardless of their admin status in the Hack Club API
  - I TESTED THIS (actually)

Closes #2965


db/seeds.rb
```diff
@@ -1,9 +1,35 @@
...
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
+user = User.first
+
+if user.nil?
+  puts "Woah there, there aren't any users! Please sign in first."
+else
+  puts "Continuing with #{user.email}..."
+
+  user.update!(admin_at: Time.now) if !user.admin?
```
garyhtou: would this get overridden by the Hack Club API? Or has it been decoupled already?

garyhtou (approved):
Good start! In the future, we might want to include some other test data such as ACHs and disbursements since its difficault to create that test data in development

db/seeds.rb
cjdenio: oop, good call

garyhtou:
Zeitwerk check now works!

> development.yml.enc takes priority over credentials.yml.enc in development, so people using the production database will run into Lockbox issues

Is there a way for those using the prod database specify Rails to use credentials.yml.enc in development?

cjdenio:
@garyhtou you can now force Bank to use the production credentials with `USE_PROD_CREDENTIALS=true` (though it doesn't work with `rails console` for some reason— looks like it interprets config differently)

garyhtou:
Just wanted to double check that the `secret_key_base` and `lockbox  master_key` are not the same as prod

cjdenio:
Yep! I generated random ones for dev

albertchae:
I was just stumped for a bit over getting some lockbox errors locally but finally realized this PR added `config/credentials/development.yml.enc`. Thanks for adding the `USE_PROD_CREDENTIALS` workaround. It worked great when adding to `.env.docker` 🎉 

> (though it doesn't work with rails console for some reason— looks like it interprets config differently)

I believe this worked for me when I made the `.env.docker` change, started rails console by doing `./docker_start.sh bash` and then running `./bin/rails c`

cjdenio:
Glad you figured out how to make it work again 😅

PR #3136, opened 2022-10-24
Don't highlight unmapped ACHs

cjdenio:
With the new confirmation code-based ACH mapping, some ACHs sit in the ledger for about 24 hours before the confirmation code shows up in the memo.

This PR updates the ledger UI to make that clear.

Screen Shot 2022-10-24 at 10 40 04 AM

(the memo would just be `LEO MCELROY BUSBILLPAY` if this was "real" and not an example)


cjdenio:
doot

PR #3137, opened 2022-10-25
`Event#balance_v2_cents` performance improvement

garyhtou:
This helps address slow response time for Stripe Issuing authorization webhooks by improving performance. Also adds `stackprof` gem for profiling with `rack-mini-profiler`.

**Test on `/hq`:**
- Before: 1220 milliseconds. 863 unique SQL queries.
- After: 747 milliseconds. 597 unique SQL queries.

Drastic performance improvements will likely come from implementing `local_hcb_code` as an association on PT (https://github.com/hackclub/bank/pull/2883).


garyhtou:
@cjdenio could I get your eyes on this to double-check that the output hasn't changed (calculations are still correct)

maxwofford:
@cjdenio @garyhtou let's bump this to priority 1. Card declines are happening on Stripe b/c of this delay & that's causing downtime.

cjdenio (approved):
Looks great! I've verified that this doesn't change any events' balances.

albertchae:
@garyhtou I finally merged https://github.com/hackclub/bank/pull/2883, can you check how much impact it had on performance? I'm not sure if it'll have that much of a change (I saw a drop of 314 to 256 queries), but every bit helps!

PR #3139, opened 2022-10-25
Fix the tooltip text for current_users

maxwofford:
Previously hovering over users in an event's show page wouldn't show custom flavor_text on the currently signed in user. This fixes that!

Screen Shot 2022-10-25 at 13 33 24


app/helpers/users_helper.rb
```diff
@@ -50,34 +78,14 @@ def user_mention(user, options = {}, default_name = "No User")
 
     klasses = ["mention"]
     klasses << %w[mention--admin tooltipped tooltipped--n] if user&.admin?
-    klasses << "mention--current-user" if user == current_user
+    klasses << %w["mention--current-user tooltipped tooltipped--n"] if user == current_user
```
cjdenio:
```suggestion
    klasses << %w[mention--current-user tooltipped tooltipped--n] if user == current_user
```

cjdenio (approved):

PR #3140, opened 2022-10-25
Add demo mode to events

maxwofford:


maxwofford:
Screen Shot 2022-10-25 at 14 27 52

Just added a placeholder design for organizer invites to demo orgs

maxwofford:
Switched from a pill to a popdown:

_Before_
Screen Shot 2022-10-25 at 14 30 16

_After_
Screen Shot 2022-10-25 at 18 24 25
Screen Shot 2022-10-25 at 18 24 20
Screen Shot 2022-10-25 at 18 24 08

app/models/event.rb
```diff
@@ -176,6 +178,8 @@ class Event < ApplicationRecord
     where("(last_fee_processed_at is null or last_fee_processed_at <= ?) and id in (?)", 5.days.ago, self.event_ids_with_pending_fees_greater_than_0_v2.to_a.map { |a| a["event_id"] })
   end
 
+  scope :filter_demo_mode, ->(demo_mode) { demo_mode.nil? || demo_mode.blank? ? all : where(demo_mode: demo_mode) }
+
   aasm do
```
cjdenio: Thoughts on using AASM for demo mode? That way we could more easily advance through `demo_mode` -> `application_submitted` -> `approved`/`rejected` or something

app/views/events/edit.html.erb
```diff
@@ -4,6 +4,22 @@
...
 <h1>Settings</h1>
 
+<% if @event.demo_mode? %>
+  <div class="card border b--info pb0 mb2 max-width-2">
+    <div class="flex items-center info">
+      <%= inline_icon 'info', size: 32, class: 'mr1' %>
+      <p class="bold mt0 mb0">Welcome to Demo Mode</p>
+    </div>
+    <ul class="mt1 mb2">
+      <li>When you're ready to sign up for a full account, click here</li>
```
exu3: Would "click here" link to the modal?

---

also does it make sense for this line to be visible in demo mode? It's confusing to say that they're on the "full fiscal sponsorship plan" when they're in demo mode / haven't yet signed a contract.
Screen Shot 2022-11-02 at 1 39 55 PM

exu3 (approved):
Lgtm! All the features that come to mind for putting behind a contract appear to be gated.

app/views/events/edit.html.erb
maxwofford: Just fixed the link & hid the plan section!

app/models/event.rb
maxwofford: I'm up to switch over to that but would like to merge this in as-is. The AASM states for events seem like they have some context in them I don't want to get caught in while prototyping currently.

PR #3150, opened 2022-10-28
Add async frame for admin events list

maxwofford:
This drops the upfront SQL queries on loading https://bank.hackclub.com/admin/events & caches the balance of each event. In the long term we'll want to actually speed up the balance calculation (for things like #3149, but this should close #3147 for now.


app/views/admin/events.html.erb
```diff
@@ -37,8 +37,10 @@
             <%= event.name.upcase %>
           <% end %>
         </td>
-        <td style="<%= 'color: lightgray;' if event.balance === 0 %>">
-          <%= render_money event.balance %>
+        <td>
+          <%= async_frame_to event_balance_admin_path(event) do %>
```
garyhtou: if you wanted to play around with turbo frames, this could be a good case to do so

maxwofford: Love that idea! I'm going to merge this as-is so the ops team has their improvement, then send a PR that rewrites this to turbo frames when I can this week.

PR #3151, opened 2022-10-29
Instantly create `Fee` objects for event mappings

cjdenio:
This makes pending fees appear immediately after a transaction is mapped, instead of every hour.

Closes #3142


app/services/fee_engine/create.rb
```diff
@@ -0,0 +1,51 @@
...
+      reason = "HACK CLUB FEE" if @canonical_event_mapping.canonical_transaction.likely_hack_club_fee?
+
+      reason = "REVENUE" if @canonical_event_mapping.canonical_transaction.amount_cents > 0
+
+      reason = "REVENUE WAIVED" if @canonical_event_mapping.canonical_transaction.likely_check_clearing_dda? # this typically has a negative balancing transaction with it
+      reason = "REVENUE WAIVED" if @canonical_event_mapping.canonical_transaction.likely_card_transaction_refund? # sometimes a user is issued a refund on a transaction
+      reason = "REVENUE WAIVED" if @canonical_event_mapping.canonical_transaction.likely_disbursement? # don't run fees on disbursements
+
+      # don't run fee if other transactions in it's HCB Code have fees waived
+      reason = "REVENUE WAIVED" if @canonical_event_mapping.canonical_transaction.local_hcb_code.canonical_transactions.map { |ct| ct.fees }.flatten.any? { |fee| fee.reason == "REVENUE WAIVED" }
```
garyhtou: i think this might be more performant

```suggestion
      reason = "REVENUE WAIVED" if @canonical_event_mapping.canonical_transaction.local_hcb_code.canonical_transactions.joins(canonical_event_mapping: :fees).where("fees.reason = ?", "REVENUE WAIVED").any?
```

garyhtou: for these here, it might be okay to say `return "REVENUE WAIVED" if ...` to bypass the rest of the checks. we can do this since (I assume) if it qualifies as waived, then it overrides any other possible reasons

garyhtou (approved):
a couple of suggestions on the old existing code. feel free to apply if you'd like

PR #3170, opened 2022-10-31
Format donation query param to the hundredth

YodaLightsabr:
Tiny change to the formatting when using the `amount` query parameter on the donation page

Before | After
-------------------- | --------
image | image


cjdenio (approved):
Thanks!

PR #3175, opened 2022-11-01
Experiments with calculation methods for Event's Fronted Balance

garyhtou:
Calculating an Event's balance is a common bottleneck for many page loads. It has also been causing timeouts for [Stripe authorization responses](https://github.com/hackclub/bank/issues/3149). The `Event#fronted_incoming_balance_v2_cents` method is the slowest portion, making up nearly 99% of the time required to retrieve the balance of a given Event.

This PR introduces two new methods for calculating the fronted balance of an Event:
- Bulk calculation using HcbCode grouping
- Raw SQL

I've also put together a Raw SQL query for calculating the total balance of an Event. [Check it out on Blazer](https://bank.hackclub.com/blazer/queries/278-compute-balance-with-just-sql?event_id=183).
```SQL
SELECT SUM(amount_cents) / 100.0 as balance
FROM
(
    ( -- Settled Balance
        SELECT SUM(amount_cents) as amount_cents FROM "canonical_transactions" ct
        JOIN "canonical_event_mappings" cem ON cem.canonical_transaction_id = ct.id
        WHERE cem.event_id = 183
    )
    UNION
    ( -- Fronted Balance
        SELECT GREATEST((coalesce(cpt_amount_cents, 0) - coalesce(ct_amount_cents, 0)), 0) as amount_cents
        FROM
        (
            SELECT SUM(amount_cents) as cpt_amount_cents, hcb_code
            FROM "canonical_pending_transactions" cpt
            JOIN "canonical_pending_event_mappings" cpem ON cpem.canonical_pending_transaction_id = cpt.id
            WHERE cpem.event_id = 183 AND cpt.amount_cents > 0 AND cpt.fronted = true AND cpt.id NOT IN (
                SELECT cpdm.canonical_pending_transaction_id
                FROM "canonical_pending_declined_mappings" cpdm
            )
            GROUP BY hcb_code
        ) cpts
        FULL JOIN
        (
            SELECT SUM(amount_cents) as ct_amount_cents, hcb_code
            FROM "canonical_transactions" ct
            JOIN "canonical_event_mappings" cem ON cem.canonical_transaction_id = ct.id
            WHERE cem.event_id = 183
            GROUP BY hcb_code
        ) cts
        ON cpts.hcb_code = cts.hcb_code
        WHERE cpts.cpt_amount_cents > 0
    )
    UNION
    ( -- Pending Outgoing Balance
        SELECT SUM(amount_cents)
        FROM "canonical_pending_transactions" cpt
        JOIN "canonical_pending_event_mappings" cpem ON cpem.canonical_pending_transaction_id = cpt.id
        WHERE cpt.amount_cents < 0 AND cpem.event_id = 183 AND cpt.id NOT IN (
            SELECT canonical_pending_transaction_id
            FROM "canonical_pending_declined_mappings"
            UNION
            SELECT canonical_pending_transaction_id
            FROM "canonical_pending_settled_mappings"
        )
    )
) q
```

We want to sure that the calculation of an Event's balance is both quick and accurate. So, I've decided to use https://github.com/github/scientist to perform experiments on these two new methods. `Scientist` will allow us to test their accuracy and record observations of their runtime durations. Since balance calculations need to be quick, I've placed these experiments in a job so they do not block/delay the calculation of balances.

So far, I've ran these experiments in development on every Event (total of 2K+ runs) with the following results:
- 100% accuracy for both methods
- on `HQ`
	- Control
		- Median: 2.367 secs
		- Average: 3.653 secs
	- Bulk
		- Median: 0.070 secs
		- Average: 0.060 secs
	- SQL
		- Median: 0.030 secs
		- Average: 0.031 secs
- on all events
	- Control
		- Median: 0.032 secs
		- Average: 0.112 secs
	- Bulk
		- Median: 0.058 secs
		- Average: 0.064 secs
	- SQL
		- Median: 0.025 secs
		- Average: 0.031 secs

My current plan is to leave these experiments running in production for a few days, then we'll check in on the results.


app/jobs/experiment_job/event_fronted_balance.rb
```diff
@@ -0,0 +1,89 @@
...
+              )
+            GROUP BY hcb_code
+        ) cpts
+        FULL JOIN
+        (
+            SELECT SUM(amount_cents) as ct_amount_cents, hcb_code
+            FROM "canonical_transactions" ct
+            JOIN "canonical_event_mappings" cem ON cem.canonical_transaction_id = ct.id
+            WHERE cem.event_id = #{@event_id}
+            GROUP BY hcb_code
```
garyhtou: there might be a way we can optimize this further: only get CT of HcbCodes which include fronted PTs

maxwofford (approved):
This is a good setup for testing these out. Can't wait to see the results!

PR #3177, opened 2022-11-01
Make `git diff` work with encrypted credentials

cjdenio:
Turns out Rails automatically enrolls a diff provider when you run `rails credentials:edit`! See `rails credentials:help` for info on how this works.


cjdenio:
*looks one way\*

PR #3181, opened 2022-11-03
Copy custom_memo from CanonicalPendingTransaction to CanonicalTransaction at settlement time

albertchae:
Per discussion in https://github.com/hackclub/bank/issues/3089#issuecomment-1279453709, to resolve weird searching issues when custom_memo is on canonical_pending_transaction but not on the canonical_transaction, instead of adding canonical_pending_transaction's custom_memo to the set of things we are searching on, we will copy over the custom_memo to the canonical_transaction at the time of settlement (if the canonical_transaction hasn't already had a custom_memo set on it)

- Implement CanonicalPendingTransactionService:Settle service to settle pending transactions and potentially copy over custom_memo
- Replace every instance of CanonicalPendingSettledMapping.create! with the new service


spec/services/canonical_pending_transaction_service/settle_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  it "creates a canonical_pending_settled_mapping" do
+    service.run!
+
+    canonical_transaction.reload
+    canonical_pending_transaction.reload
+
+    canonical_pending_settled_mapping = CanonicalPendingSettledMapping.last
+    expect(canonical_transaction.canonical_pending_settled_mapping).to eq(canonical_pending_settled_mapping)
+    expect(canonical_pending_transaction.canonical_pending_settled_mappings).to eq([canonical_pending_settled_mapping])
```
albertchae: N.B. this is an 1-element array with the `canonical_pending_settled_mapping`

BTW, is there any reason this association is a `has_many` while the one on `canonical_transaction` is a `has_one`? Can a `canonical_pending_transaction` have multiple `canonical_transaction`s that settle it?

maxwofford: A `canonical_pending_transaction` can have multiple `canonical_transaction`s that settle it– one example is a donation:
- 1 `canonical_pending_transaction` created when payment is first made
- the payout `canonical_transaction` for slightly less (with Stripe's fees removed)
- a fee refund `canonical_transaction` to pay back Stripe's fees

maxwofford (approved):
LGTM

spec/services/canonical_pending_transaction_service/settle_spec.rb
albertchae: got it, thanks for the explanation

PR #3194, opened 2022-11-03
Prevent disbursements & ACH in demo events

maxwofford:


exu3 (approved):
LGTM

---

Should admins be able to transfer money to demo events? I can't think of a case where that would need to happen.

maxwofford:
I can't think of a reason, so I'm locking it down for now. We can open it back up once we know what that'd be used for.

PR #3204, opened 2022-11-04
Move Demo Mode notice to show page

exu3:
so it's more noticeable; new users don't go to the settings page immediately

This PR also changes some of the copy around Demo Mode according to feedback from Zach. 
- "Request Contract" makes it feel like you have to pay $$ or make some sort of large commitment
- 24hr turnaround time
- 
Screen Shot 2022-11-04 at 1 47 26 PM


maxwofford (approved):
Love this!

PR #3213, opened 2022-11-10
Patch ACH creation for non-demo Events

garyhtou:
closes https://github.com/hackclub/bank/issues/3212


app/policies/ach_transfer_policy.rb
```diff
@@ -10,7 +10,7 @@ def new?
   end
 
   def create?
-    user&.admin? || record.users.include?(user) && record.not_demo_mode # dirty implementation here. record is event (temporary)
+    user&.admin? || record.users.include?(user) && !record.demo_mode # dirty implementation here. record is event (temporary)
```
garyhtou:
```suggestion
    user&.admin? || (record.users.include?(user) && !record.demo_mode)
```
@maxwofford your logic seems right here. is it okay to drop the comment?

PR #3217, opened 2022-11-10
Make tooltip larger for latest build

YodaLightsabr:
Fixes the text expanding out of the tooltip as seen in this image:

image


app/assets/stylesheets/components/_tooltips.scss
```diff
@@ -30,6 +30,10 @@
       transition: 0.125s all ease-in-out;
       width: max-content;
       z-index: 1;
+
+      &.tooltipped--lg {
```
garyhtou: What are your thoughts on increasing the `max-width` on line 21?

YodaLightsabr: That should work; I can check if it breaks other things.

garyhtou (approved):

cjdenio:
This does make certain tooltips very large

Screenshot 2022-12-24 at 1 45 01 PM

garyhtou:
> This does make certain tooltips very large

oh yeah- that's something I didn't consider. @YodaLightsabr maybe your [original commit](https://github.com/hackclub/bank/pull/3217/commits/ca02f9addfcf652ee3f7ed5bbae4a176683b72a1) might make more sense. I'll leave this up to you to decide!

maxwofford:
I've reverted back to @YodaLightsabr's original commit and am now merging because this works locally!

PR #3218, opened 2022-11-11
Fix leak of invited user's email in Transparency mode

tinuh:
user email addresses were being shown for invited users in transparency mode
fixed the bug by parsing the email into initials before encoding it into avatar url

closes https://github.com/hackclub/bank/issues/3184


app/helpers/users_helper.rb
```diff
@@ -5,6 +5,12 @@
 
 module UsersHelper
   def gravatar_url(email, name, id, size)
+    email = email.split("@")[0].split(/[^a-z\d]/).compact_blank
```
```suggestion
    email = email.split("@")[0].split(/[^a-z\d]/i).compact_blank
```

Added `i` to the end of the regex to make it case insensitive

app/helpers/users_helper.rb
```diff
@@ -5,6 +5,12 @@
...
+    if email.length == 1
+      email = email[0].first(2)
+    else
+      email = email[0].first + email[1].first
+    end
     name ||= email
```
```suggestion
    name ||= begin
    	# Let's move your code into here to prevent modifying `email`. This way the computed `hex` (below) is able to use the full email address and stays the same. Alternatively, we can move the hex computation above.
    end
```

garyhtou (changes requested):
good work here! couple of small edits

garyhtou:
I personally find it really cool when code is written in a concise way. u totally don't need to change what you, ur code works great.

if ur interested, here u go:

```ruby
# this def does make it a little harder to read
email.split("@").first.split(/[^a-z]/i).compact_blank.first(2).map(&:first).join

# explanation
email.split("@").first.split(/[^a-z]/i).compact_blank # the line is the same as what you had
.first(2) # get the first two strings. if there are fewer than 2, it'll get what's available.
.map(&:first) # get the first letter of every string in the array. this will result in an array of strings with 1 character each
.join # join the two strings together to form a single two character string
```

tinuh:
would the concise version of the code ignore the case where we get the first two letters of a single string?

garyhtou:
> would the concise version of the code ignore the case where we get the first two letters of a single string?

oh right, forgot about that! it does ignore it

garyhtou (approved):
good to merge!

PR #3223, opened 2022-11-16
Implement fast fronted balance calculation

garyhtou:
Implemented `bulk` method from https://github.com/hackclub/bank/pull/3175/. On average, it is 20 times faster than the old method for HQ.

### Stats

<table>
<tr>
	<td>HQ
	<td>ALL
<tr>
	<td>image
	<td>image
</table>

#### Average
- HQ Event (20 times faster)
  - New: 0.0500 sec
  - Old: 1.0342 sec
- All Events (9 times faster)
  - New: 0.0252 sec
  - Old: 0.2514 sec

#### Median
- HQ Event (27 times faster)
  - New: 0.0154 sec
  - Old: 0.4339 sec
- All Events (2 times faster)
  - New: 0.0070 sec
  - Old: 0.0208 sec


maxwofford (approved):
Looks great! Is there any non-code tear down we need to do for experiments?

PR #3226, opened 2022-11-18
Temporarily disable viewing edited comment history

albertchae:
This seems to be the only user facing feature using PaperTrail. We want
to update PaperTrail's internal storage from YAML to JSON without having
to run a maintenance window and take down the site, which we can do if
we temporarily disable this feature.

https://github.com/hackclub/bank/issues/3110

Looks like this
image


albertchae:
Merging this per our sync on Nov 18 @maxwofford

PR #3227, opened 2022-11-18
Update PaperTrail to jsonb for new updates

albertchae:
We do this by first renaming existing YAML based object and
object_changes columns in the versions table to `old_objects`
and `old_object_changes`. Then we introduce `objects` and
`object_changes` as jsonb columns. PaperTrail gem will automatically
convert to using JSON serialization for all new edits going forward.

https://github.com/paper-trail-gem/paper_trail#postgresql-json-column-type-support
```
If you use the PostgreSQL json or jsonb column type, you do not need to specify a PaperTrail.serializer.
```

As for picking `json` vs `jsonb` for the postgres type, referring to
https://www.postgresql.org/docs/11/datatype-json.html, jsonb is
considered more efficient in most scenarios and allows for indexing.
Although we don't need to index the column now, I picked jsonb mainly on
this recommendation from the postgres docs
`In general, most applications should prefer to store JSON data as jsonb, unless there are quite specialized needs, such as legacy assumptions about ordering of object keys.`

N.B. this means old papertrail history will be temporarily unaccessible
until we convert the old YAML data into the new JSON data. That will be
a separate follow up PR.

https://github.com/hackclub/bank/issues/3110
Should be merged after https://github.com/hackclub/bank/pull/3226, and a short maintenance window is put in place.


albertchae:
Merging since https://github.com/hackclub/bank/pull/3226 is deployed. Per sync with @maxwofford on Nov 18, we will not need to use a heroku maintenance window since the period of potential errors will be short.

albertchae:
Deploy completed successfully https://dashboard.heroku.com/apps/bank-hackclub/activity/releases/2ea8ce98-6c18-47e0-86b5-23331cca67fd, no airbrake errors AFAICT

PR #3228, opened 2022-11-18
Custom stripe card names

tinuh:
closes #2998

## Before
image

## After
image

## Changing the name
image


tinuh:
not sure how to fix the string literal comment from rubocop because the schema was auto-generated.

exu3:
> not sure how to fix the string literal comment from rubocop because the schema was auto-generated.

Run `bundle exec rubocop -A` to autofix Rubocop offenses

db/schema.rb
```diff
@@ -31,8 +30,8 @@
     t.datetime "updated_at", null: false
     t.string "recipient_tel"
     t.datetime "rejected_at"
-    t.datetime "scheduled_arrival_date"
     t.text "payment_for"
+    t.datetime "scheduled_arrival_date"
```
garyhtou: Sometimes these schema changes show up when we run `db:migrate`. See if you can remove this one and the `raw_pending_partner_donation_transaction_id` change from your PR.

garyhtou (changes requested):
The migration itself looks good! I just left a comment about the other changes

I think we could also put together the other code for this feature as a part of this PR. however, if you would prefer to just merge the migration and leave the other code for another PR, that's fine too

app/models/stripe_card.rb
```diff
@@ -116,10 +117,6 @@ def stripe_name
     stripe_cardholder.stripe_name
   end
 
-  def name
```
garyhtou: This method was removed and existing references were refactored to allow the `name` field to be accessed directly

app/views/events/_event_cards_table.html.erb
```diff
@@ -31,7 +31,7 @@
             <%= format_date d.created_at %>
           </td>
           <td>
-            <%= d.name || d.user.full_name if organizer_signed_in? %>
+            <%= d.stripe_name || d.user.full_name if organizer_signed_in? %>
```
garyhtou: https://github.com/hackclub/bank/pull/3228/files#r1033034311

garyhtou (approved):
Amazing work!

good to merge

PR #3230, opened 2022-11-19
Show ACH processor

YodaLightsabr:
My attempt at #3210


app/controllers/admin_controller.rb
```diff
@@ -522,13 +522,14 @@ def ach_approve
...
+      confirmation_number: params[:confirmation_number],
+      processor: current_user
     }
     ach_transfer = AchTransferService::Approve.new(attrs).run
 
     redirect_to ach_start_approval_admin_path(ach_transfer), flash: { success: "Success" }
-  rescue => e
-    redirect_to ach_start_approval_admin_path(params[:id]), flash: { error: e.message }
+  # rescue => e
+  #   redirect_to ach_start_approval_admin_path(params[:id]), flash: { error: e.message }
```
```suggestion
  rescue => e
    redirect_to ach_start_approval_admin_path(params[:id]), flash: { error: e.message }
```

cjdenio (approved):

garyhtou (approved):
Looks good!

cjdenio:
oop @YodaLightsabr could you rename the field to `processed_by` instead of `processor`? That way it's more consistent with other models, and more evident it's referring to a person

YodaLightsabr:
How could I go about doing it now that it's been merged? Would it mess up any ACH transfers made in the last few days?

cjdenio:
There's only been 1 ACH transfer since— I think the easiest solution is to revert this PR and roll back the database schema

PR #3231, opened 2022-11-19
Skip storing PaperTrail history for unencrypted columns

albertchae:
This is the recommended way by lockbox documentation
https://github.com/ankane/lockbox#model-changes

This only redacts them moving forward. We still need to implement a
one time job to redact all existing ones.

https://github.com/hackclub/bank/issues/2541


albertchae:
@maxwofford I'm going to merge this now so that we don't end up with more records to redact in the backfill job. Let me know if you have any concerns post merge.

maxwofford:
This checks out based on the lockbox documentation!

PR #3233, opened 2022-11-21
Add job to migrate simple cases of papertrail YAML to JSON

albertchae:
These are simple cases because they do not have lockbox encrypted
columns. We use the recommended way from PaperTrail documentation
by loading the old YAML format and storing it in the json column.

https://github.com/paper-trail-gem/paper_trail#convert-existing-yaml-data-to-json

Although this is slow, bank's db doesn't have that many records so it
still finishes within a few minutes.

Ironically Rubocop is flagging Security/YAMLLoad which is the very
reason we want to be updating PaperTrail.
https://github.com/paper-trail-gem/paper_trail/blob/master/doc/pt_13_yaml_safe_load.md
Disable it for this job so we can finish the migration though.

I ran this query (which tells you how many records have only the old
YAML columns) before and after
```
$ PaperTrail::Version.where(object: nil, object_changes: nil).count
Before: 158270
After:   44397
```


albertchae:
I didn't write a test for this because I'd have to manually write an old format YAML into the `old_object`/`old_object_changes` columns. Instead I ran it on a local dump of the prod database and ran queries afterwards to verify correctness.

app/jobs/one_time_jobs/paper_trail_convert_yaml_to_json_simple_job.rb
```diff
@@ -0,0 +1,33 @@
...
+  class PaperTrailConvertYamlToJsonSimpleJob < ApplicationJob
+    # These are simple because they don't have lockbox encrypted columns that require redaction
+    SIMPLE_ITEM_TYPES = [PartnerDonation, CanonicalTransaction, FeeReimbursement, Event, HcbCode, StripeAuthorization, Fee, DonationPayout, InvoicePayout, LoginToken, Donation, Disbursement, PartneredSignup, GSuite, Sponsor, OrganizerPositionInvite, OrganizerPositionDeletionRequest, BankFee, StripeCard, CanonicalPendingTransaction].freeze
+
+    def perform
+      PaperTrail::Version
+        .where(object: nil, object_changes: nil, item_type: SIMPLE_ITEM_TYPES.map(&:to_s))
+        .find_each do |version|
+
+        # rubocop:disable Security/YAMLLoad
```
albertchae: Ironically Rubocop is flagging Security/YAMLLoad which is the very reason we want to be updating PaperTrail.
https://github.com/paper-trail-gem/paper_trail/blob/master/doc/pt_13_yaml_safe_load.md
Disable it for this job so we can finish the migration though.

maxwofford (approved):

albertchae:
Finished running this on prod with same results as local
```
irb(main):001:0> PaperTrail::Version.where(object: nil, object_changes: nil).count
=> 158270
irb(main):002:0> OneTimeJobs::PaperTrailConvertYamlToJsonSimpleJob.new.perform
=> nil
irb(main):003:0> PaperTrail::Version.where(object: nil, object_changes: nil).count
=> 44397
```

PR #3234, opened 2022-11-21
Add job to migrate more complex PaperTrail records with lockbox columns

albertchae:
As part of https://github.com/hackclub/bank/issues/2541, we also need to
redact historical entries of columns that were later encrypted. We do
this by calling `Model.generate_column_name_ciphertext(plaintext)`, e.g.
for the `User` model and `api_access_token` column, we are calling
`User.generate_api_access_token_ciphertext(plaintext)`. We get the
plaintext value from the old PaperTrail entry, which we remove from the
hash we store in the new JSON based PaperTrail columns.

Because the column name is baked into the Lockbox API, this required
some metaprogramming on method name and column name that I do not
recommend as a pattern for future code unless there is a very good
reason.

For reference, this is the PR that redacted these columns going forward
https://github.com/hackclub/bank/pull/3231

I ran this query (which tells you how many records have only the old
YAML columns) before and after
```
$ PaperTrail::Version.where(object: nil, object_changes: nil).count
Before: 44397
After:      0
```

These queries are also 0
```
PaperTrail::Version.where(object_changes: nil).where.not(old_object_changes: nil).count
PaperTrail::Version.where(object: nil).where.not(old_object: nil).count
```


albertchae:
No tests for same reasoning as https://github.com/hackclub/bank/pull/3233#issuecomment-1321487808

app/jobs/one_time_jobs/paper_trail_convert_yaml_to_json_encrypted_columns_job.rb
```diff
@@ -0,0 +1,77 @@
...
+            encrypted_columns = MODELS_TO_ENCRYPTED_COLUMNS[model_class]
+            old_object_redacted = old_object
+                                  .except(*encrypted_columns) # remove non ciphertext versions of columns that are now encrypted from the papertrail history
+                                  .tap do |hash|
+              encrypted_columns.each do |column|
+                next unless old_object[column]
+
+                # Adds ciphertext version of encrypted column, so we are still storing history but it is encrypted at the PaperTrail level too
+                # e.g. calls User.generate_api_access_token_ciphertext(old_object["api_access_token"])
+                hash["#{column}_ciphertext"] = model_class.send("generate_#{column}_ciphertext", old_object[column])
```
albertchae: N.B. Calling `User.generate_api_access_token_ciphertext` (or any of the other generate column ciphertext methods) on the same plaintext multiple times will actually result in different encrypted strings because each call uses a different [nonce](https://en.wikipedia.org/wiki/Cryptographic_nonce). Reference from lockbox author https://github.com/ankane/lockbox/issues/51

albertchae:
Besides this and https://github.com/hackclub/bank/pull/3233, there are a small number of `PaperTrail::Version` records that created data in the new JSON format but didn't redact sensitive columns (i.e. the PaperTrail records created between the deploys of https://github.com/hackclub/bank/pull/3227 and https://github.com/hackclub/bank/pull/3231). I will follow up with another PR for those, but the range of affected records are `id: 160358, item_type: "UserSession"` to `id: 160400, item_type: "OrganizerPositionDeletionRequest"`

maxwofford (approved):

albertchae:
Finished running this on prod

```
irb(main):001:0> PaperTrail::Version.where(object: nil, object_changes: nil).count
=> 44397
irb(main):002:0> PaperTrail::Version.where(object_changes: nil).where.not(old_object_changes: nil).count
=> 43651
irb(main):003:0> PaperTrail::Version.where(object: nil).where.not(old_object: nil).count
=> 35369
irb(main):004:0> OneTimeJobs::PaperTrailConvertYamlToJsonEncryptedColumnsJob.new.perform
...
irb(main):005:0> PaperTrail::Version.where(object: nil, object_changes: nil).count
=> 0
irb(main):006:0> PaperTrail::Version.where(object_changes: nil).where.not(old_object_changes: nil).count
=> 0
irb(main):007:0> PaperTrail::Version.where(object: nil).where.not(old_object: nil).count
=> 0
```

PR #3239, opened 2022-11-22
Redact encrypted columns from paper trail versions that were already using json format

albertchae:
Per https://github.com/hackclub/bank/pull/3234#issuecomment-1321545073

Between the deploys of https://github.com/hackclub/bank/pull/3227 and https://github.com/hackclub/bank/pull/3231, i.e. after PaperTrail was changed to use JSON columns but before we omitted encrypted columns from PaperTrail, there were a few records created that the job in https://github.com/hackclub/bank/pull/3234 would not have redacted. This PR implements a job to redact those.

Because we are looking at records that already were written to JSON, we can just look at `object` and `object_changes` columns and ignore any `YAML.load`ing

Running this job is the last step for https://github.com/hackclub/bank/issues/2541 as well as unblocks upgrading the gem https://github.com/hackclub/bank/issues/3110


maxwofford (approved):

albertchae:
Finished running this on prod

PR #3247, opened 2022-11-23
Delete all fixtures and db cleanup hacks due to fixture pollution

albertchae:
Now that all tests have been converted to factorybot, we don't need
these anymore


maxwofford (approved):
Wahoo! PRs like this feel so good

PR #3254, opened 2022-11-26
Fix _canonical_transaction.html.erb partial to take canonical_transaction instead of hcb_code

albertchae:
When an HcbCode had multiple canonical_transactions associated with it,
we were seeing duplicate transactions on my/inbox and the stripe card
show page. This is because instead of passing in the
`canonical_transaction` to the `_canonical_transaction.html.erb` partial, we
were passing in the `hcb_code` and thus rendering the first
`canonical_transaction` on the `hcb_code` twice.

It looks like the original intent of the `_canonical_transaction.html.erb`
partial was to take `canonical_transaction` as `ct`, both based on the
variable name and
https://github.com/hackclub/bank/blob/1211f55f0057d1cfd2667a7759be5e0d2d6e5cd8/app/views/stripe_cards/show.html.erb#L147-L149
but it looks like this was inadvertently changed in
https://github.com/hackclub/bank/pull/1872/files#diff-bc63b7f9f2db69007ac0065edaaf33690c5155c1ca34eef0bbffb74fdd191d09
and masked by the fact that `HcbCode` actually delegates a lot of its
methods to `CanonicalTransaction`, so it was difficult to see the type
error.

Since that change, it looks like the partial diverged a bit in assuming
it was receiving an `HcbCode`, so this PR includes changes to fix that
(mainly getting the `HcbCode` off the `ct` with `local_hcb_code`)

Fixes https://github.com/hackclub/bank/issues/3171
Fixes https://github.com/hackclub/bank/issues/3172


albertchae:
The examples from the issues

### Before:
Screen Shot 2022-10-31 at 18 29 04

### After (I had to take this from stripe card because these transactions aren't in inbox anymore):
image

### Before:
Screen Shot 2022-10-31 at 18 25 28

### After:
image

maxwofford (approved):
Thanks for this fix!

PR #3256, opened 2022-11-26
Speed up stripe card show

albertchae:
https://github.com/hackclub/bank/issues/2955

This PR improves the speed of `stripe_cards#show` by reducing the number of SQL queries used to render the page. The individual changes are best reviewed by commits, but overall this PR takes a sample page `stripe_cards/brxhw1` from
 
```
Before: SQL Queries: 708 (105 cached)
After: SQL Queries: 162 (50 cached)
```

Hypothetically I also checked what would happen if we didn't have pagination implemented (https://github.com/hackclub/bank/pull/3033)

```
Before: SQL Queries: 1954 (554 cached)
After: SQL Queries: 905 (355 cached)
```

So there is still a lot of room for improvement, but I avoided trying to go too heavy handed on the preloader approach like https://github.com/hackclub/bank/blob/main/app/services/transaction_grouping_engine/transaction/association_preloader.rb, since our goal is actually to get rid of that and use vanilla AR preloads/includes. The only case where I set an association manually was the first commit, where I added `local_hcb_code` on `CanonicalTransaction`. This should be converted once https://github.com/hackclub/bank/issues/3255 is completed

- Cache local_hcb_code instead of querying for it again from canonical_transaction
- Optionally preload tags if the feature flag is enabled. Unlike in https://github.com/hackclub/bank/blob/857386beb6e0f8bfb1dca9e48b4f81c7f674b395/app/services/transaction_grouping_engine/transaction/association_preloader.rb#L17-L23, we can't use a splat array because calling `includes` does not work with no arguments, and we would be splatting an empty array if the feature flag is off
- Preload canonical_transactions and canonical_pending_transactions on HcbCode
- Preload hashed_transactions on canonical_transactions
- Use inverse_of so CanonicalPendingTransaction uses the existing HcbCode in memory. More on inverse_of here https://www.viget.com/articles/exploring-the-inverse-of-option-on-rails-model-associations/
- Preload raw_stripe_transaction on hashed_transactions
- Preload raw_pending_stripe_transaction on canonical_pending_transactions. This one had no impact on the query count but should still be an improvement.
- Count hashed_transactions with size to use association already loaded in memory. `count` always issues a query to the db whereas `size` will use the in memory array length if it exists, otherwise it will issue a db query. In this case, the in memory array should always exist.


app/views/stripe_cards/show.html.erb
```diff
@@ -188,6 +188,7 @@
         <% @hcb_codes.order(created_at: :desc).each do |hcb_code| %>
           <% if hcb_code.canonical_transactions.any? %>
             <% hcb_code.canonical_transactions.each do |ct| %>
```
maxwofford: Is there a reason we aren't switching this to a collection?

maxwofford (approved):
Thanks for laying out all the changes in the description– that was a good read & I learned more about how `inverse_of` works under the hood. Left a comment about the `.each do` vs using a collection, but happy with this improvement overall– feel free to merge!

app/views/stripe_cards/show.html.erb
albertchae: Hm, when I was thinking about that originally, I thought it wouldn't be possible at the `@hcb_codes` level because we have to choose a different partial based on whether to display canonical_transactions or not. I think that's still true, however, we could probably still do it once we know whether we're displaying `canonical_transactions` or `canonical_pending_transactions`, but I don't think it'll have a large effect – it seems we have 1-5 `canonical_transactions`/`canonical_pending_transactions` per `hcb_code`, and we'd still need to load the partial 25 times for the whole page.

But TL;DR I'll make the change anyway, I don't think it'll hurt.

albertchae: Oh oops, now I remember why. Because we have the line after to assign the `hcb_code` on `canonical_transaction` in memory so it doesn't get loaded from database again (similar strategy to the `AssociationPreloader` classes), we have to loop over the `canonical_transactions` anyway. I think we can investigate making this improvement once `local_hcb_code` is a vanilla association (https://github.com/hackclub/bank/issues/3255)

Just to verify what the breakdown would be, I ran this query
```
HcbCode.includes(:canonical_transactions).all.map{|h| h.canonical_transactions.size}.tally
=> {0=>14838, 1=>21119, 2=>3810, 3=>230, 4=>8, 5=>2, 6=>3, 7=>2, 9=>2, 10=>3}
```
so the vast majority of hcb_codes have 3 or fewer canonical_transactions, so we're not missing out on too much of an improvement for now.

albertchae:
Merging for now because we'll address the loop over the partial in a follow up

PR #3259, opened 2022-11-27
Add event name to statement descriptors

cjdenio:
Tested in Stripe's test environment.

Before:

Screenshot 2022-11-26 at 11 33 55 PM

After:
Screenshot 2022-11-26 at 11 34 08 PM

(closes #2108)


app/models/donation.rb
```diff
@@ -285,6 +285,7 @@ def create_payment_intent_attrs
       amount: amount,
       currency: "usd",
       statement_descriptor: "HACK CLUB BANK",
+      statement_descriptor_suffix: event.name.truncate(STATEMENT_DESCRIPTOR_SUFFIX_CHAR_LIMIT),
```
garyhtou: The statement descriptor limit is 22, including the prefix.
```ruby
STATEMENT_DESCRIPTOR_SUFFIX_CHAR_LIMIT  = 22 - ("HCB* ".length)
```

app/models/donation.rb
```diff
@@ -285,6 +285,7 @@ def create_payment_intent_attrs
       amount: amount,
       currency: "usd",
       statement_descriptor: "HACK CLUB BANK",
```
garyhtou: I believe `statement_descriptor` is a fallback for when the charge is not paid via card. Do we also want to use the event's name here? (not sure if there is a char limit).

```suggestion
      statement_descriptor: "#{event.name} via Hack Club Bank"
```

garyhtou:
It looks like rspec caught a bug! lol

statement descriptors can't have a `'`. I think we should filter out invalid characters, then truncate

garyhtou:
@cjdenio would u mind testing this out?

cjdenio:
It works great for both donations and invoices!

garyhtou:
let's go ahead and merge this then!

PR #3262, opened 2022-11-28
placeholder and button to clear for custom card names

tinuh:
## Before (no placeholder or clear button):
image

## After (placeholder):
image

## Clear Button (only when there is already a name):
image


app/views/stripe_cards/_name_form.html.erb
```diff
@@ -2,7 +2,8 @@
   <section class="grid grid--split">
   <div class="field">
     <%= form.label :name, "Name" %>
-    <%= form.text_field :name, value: @card.name %>
+    <%= form.text_field :name, value: @card.name, placeholder: "eg. Online Retailers" %>
```
garyhtou:
```suggestion
    <%= form.text_field :name, value: @card.name, placeholder: "Online Retailers" %>
```

garyhtou (commented):
What if we did something like this:
Untitled

Add a delete button with a tooltip on hover ("Remove name"). The delete button would only show up if the card has a name

---

if we wanna get fancy, maybe we can find a way to have the delete button inside the field. sometime similar to 
image

app/views/stripe_cards/_name_form.html.erb
```diff
@@ -1,12 +1,30 @@
...
     <%= form.submit "Save" %>
   </div>
-<% end %>
\ No newline at end of file
+<% end %>
+
+<script>
+  const clearName = () => {
+    document.querySelector("#stripe_card_name").value = ""
+    document.querySelector("#clear_name_btn").style = "display: none;"
```
garyhtou: maybe try disabling the button instead

garyhtou (approved):
Amazing!

PR #3273, opened 2022-12-01
Prevent fronting of in-review disbursements

cjdenio:
1. Disbursement instantly shows up on the ledger as pending
2. If approved, it's marked as fronted
3. If rejected, it's declined (and thus disappears from the ledger)

Closes #3202


maxwofford (dismissed):
This PR changes some underlying logic in disbursements counter to what our interface displays:
Screenshot 2022-12-02 at 09 28 17

I'm totally onboard with the change if this is to prevent someone from stealing money, but let's also update the interface to communicate this!

app/services/disbursement_service/create.rb
```diff
@@ -28,6 +28,20 @@ def run
           disbursement_id: disbursement.id,
           fulfilled_by_id: requested_by.id
         ).run
+
+        # No need to create the pending transactions since `DisbursementService::Approve` will do it
```
garyhtou: Would it make sense to refactor so that the pending transaction creation is only located in this file? having it in multiple location gets a bit messy

PR #3279, opened 2022-12-02
Only send SMS notifs on IRL charges

maxwofford:
This is based on the description of [authorization methods published by stripe](https://stripe.com/docs/api/issuing/authorizations/object#issuing_authorization_object-authorization_method). Now we'll only send SMS notifs for IRL charges, which should also filter out monthly subscriptions.


maxwofford:
Edit: this should also reduce the number of panic attacks caused by the feature
Screenshot 2022-12-02 at 11 38 10

app/jobs/canonical_pending_transaction_job/send_twilio_message.rb
```diff
@@ -12,11 +12,25 @@ def perform(cpt_id:, user_id:)
...
       TwilioMessageService::Send.new(@user, message, hcb_code: hcb_code).run!
     end
 
+    private
+
+    def auth_method
+      @cpt&.raw_pending_stripe_transaction&.authorization_method
+    end
+
+    def in_person_auth_methods
```
cjdenio: This could be a constant:

```ruby
class SendTwilioMessage < ApplicationJob
  IN_PERSON_AUTH_METHODS = %w[keyed_in swipe chip contactless]
```

cjdenio (approved):

cjdenio (approved):

PR #3281, opened 2022-12-03
Remove yarn integrity check hack

albertchae:
Closes https://github.com/hackclub/bank/issues/2445

This was originally added to run yarn so we'd avoid the integrity check
during docker initialization of the rails app. However, this is a
webpacker specific check and webpacker was removed in
https://github.com/hackclub/bank/pull/2816 so we can get rid of this
now.


maxwofford (approved):
Wahoo!

PR #3283, opened 2022-12-04
Implement local_hcb_code on canonical_transaction as an association

albertchae:
Closes https://github.com/hackclub/bank/issues/3255

The `CanonicalTransaction` version of https://github.com/hackclub/bank/pull/2883

I was hoping this change would allow me to get rid of this line https://github.com/hackclub/bank/blob/94aa10e9261c589c991dd7521878add1b12f4bbc/app/services/transaction_grouping_engine/transaction/association_preloader.rb#L86 and use `includes` instead, but I forgot this preloader is operating on `CanonicalTransactionGrouped`...

This change still gives us some benefit with the stripe card changes in https://github.com/hackclub/bank/pull/3256

- Implement local_hcb_code on CanonicalTransaction as an AR association
- Add inverse_of for hcb_code.canonical_transactions
- Render partials as a collection now that we don't need to assign local_hcb_code
  This is part of https://github.com/hackclub/bank/issues/2484


app/models/hcb_code.rb
```diff
@@ -107,9 +107,16 @@ def amount_cents
...
+  has_many :canonical_pending_transactions,
+           foreign_key: 'hcb_code',
+           primary_key: 'hcb_code',
+           inverse_of: :local_hcb_code
+
+  has_many :canonical_transactions,
+           -> { order("canonical_transactions.date desc, canonical_transactions.id desc") },
+           foreign_key: 'hcb_code',
+           primary_key: 'hcb_code',
+           inverse_of: :local_hcb_code
```
albertchae: this is the only actual change here. the rest was formatting

app/views/stripe_cards/show.html.erb
```diff
@@ -200,14 +200,9 @@
       <tbody data-behavior="transactions">
         <% @hcb_codes.order(created_at: :desc).each do |hcb_code| %>
           <% if hcb_code.canonical_transactions.any? %>
-            <% hcb_code.canonical_transactions.each do |ct| %>
-              <% ct.local_hcb_code = hcb_code %>
```
albertchae: now that we don't need this line (because inverse_of gives it to us), we can use render collection

maxwofford (approved):

PR #3287, opened 2022-12-05
Add custom css to donation page

InternetRamen:
Solves #3257

This PR: 
 - Creates `donation_page_css` on the `events` table
 - Creates `custom_css` component
 - Creates input field in organization settings.

This PR aims to solve the issue of customization of the donation page brought by this thread: https://hackclub.slack.com/archives/CN523HLKW/p1669443832748739

This PR does not provide a form element that users can drop into their page but allows users to edit the CSS that sponsors see when viewing the donation page.

**Best reviewed by security vulnerabilities**
There might be security vulnerabilities with storing CSS and applying CSS. Another vulnerability could be the usage of pseudo-elements on input text elements.

**Future improvements**
Create a form element that users can drop into their own pages. 

**To-Do**
- [ ] Syntax highlighting
- [ ] You have to do `!important` and hyperspecify for styles

Would love for any feedback (I don't even know if this helps the customization issue)

**Screenshots**
image
image


InternetRamen:
image
XSS

InternetRamen:
I could probably put the if statement in the component call rather than inside the component itself

app/views/events/_form.html.erb
```diff
@@ -97,6 +97,17 @@
...
+        <%= form.label :donation_page_css, "Custom CSS for donation page" %>
+        <p class="h5 muted mt0 mb1">
+            "!important" tags may be needed.
+        </p>
+        <div data-controller="codemirror" data-codemirror-doc-value="<%= form.object.donation_page_css %>">
+        <div data-codemirror-target="editor"></div>
+        <%= form.hidden_field :donation_page_css, data: {codemirror_target: "input"} %>
+        </div>
+        </div>
+      </div>
```
cjdenio:
```suggestion
      <div class="field">
        <%= form.label :donation_page_css, "Custom CSS for donation page" %>
        <p class="h5 muted mt0 mb1">
            "!important" tags may be needed.
        </p>
        <div data-controller="codemirror" data-codemirror-doc-value="<%= form.object.donation_page_css %>">
          <div data-codemirror-target="editor"></div>
          <%= form.hidden_field :donation_page_css, data: {codemirror_target: "input"} %>
        </div>
      </div>
```

cjdenio (changes requested):
indentation + removed extraneous `</div>`

InternetRamen:
- [ ] Style donations box to match HCB themes (aside from Dracula idk we can play around with it)

northeastprince:
There are a infinite ways this could be abused/messed up, though my main concerns are around security (even if we implement some sort of CSS sanitization) and people (even those with good intentions) modifying something that potentially threatens HC/HCB's reputation. Thoughts on just adding limited customization options like colors/patterns of certain things, maybe custom images, etc.?

InternetRamen:
That's probably a good idea considering I'm unable to find a way to theme code mirror for both light mode and dark mode. 

I do believe we allow custom images and icons already, what do you think should be an option to change?

I think this PR was meant to solve the issue of not being able to embed the donation form into an existing website, so this feature may not solve the actual problem

YodaLightsabr:
The donation page can be embedded with an iframe, and there's a workaround for building a custom form and prefilling the values.

Custom CSS would be awesome, but there's a lot of room for vulnerabilities. Unless we look over the CSS or only allow modifying certain elements, it would be easy to impersonate another organization, trick someone into donating more, or secretly switch to a monthly donation.

northeastprince:
It'd probably be better to have a separate view for the embeddable form.

garyhtou:
Closing as stale, we can reopen later

PR #3288, opened 2022-12-05
Use `acts_as_paranoid` for Google Workspace deletion

cjdenio:
This
- de-jankifies things and prevents deletion-related bugs from appearing if someone forgets a `not_deleted`
- makes google workspace applications actually disappear on the admin side, thus preventing situations like this:
Screenshot 2022-12-05 at 4 36 12 PM


app/models/g_suite.rb
```diff
@@ -30,6 +30,8 @@
 class GSuite < ApplicationRecord
   VALID_DOMAIN = /[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,5}(:[0-9]{1,5})?(\/.*)?\z/ix.freeze
 
+  acts_as_paranoid
+  validates_as_paranoid
```
garyhtou: I think might be okay if we drop the `validates_as_paranoid` here.

and keep the original `validates_uniqueness_of :domain` below.

garyhtou (approved):
just a comment about the `validates_as_paranoid`

app/models/g_suite.rb
cjdenio: I tried that initially, and you do in fact need the `validates_as_paranoid`: https://github.com/ActsAsParanoid/acts_as_paranoid#validation

PR #3289, opened 2022-12-06
Spike on using view_component to improve performance of events#show

albertchae:
As mentioned in https://github.com/hackclub/bank/issues/3280, vanilla rails partials/views can be slow. The `view_component` gem is an alternative way to use rails views that has many benefits, one of which is improved performance.

This PR adds the `view_component` gem, then uses it to reimplement https://github.com/hackclub/bank/blob/main/app/views/canonical_transactions/_canonical_transaction.html.erb as a ViewComponent, then uses that ViewComponent on the events show page.


cjdenio:
Closing because of inactivity— feel free to reopen later!

PR #3298, opened 2022-12-07
Lint ERB files

cjdenio:
Best reviewed by commits. Most of these changes were applied automatically by erb-lint.

Fixes #2962


garyhtou:
Can we set up a script that will run both erb_lint and rubocop?

app/services/flavor_text_service.rb
```diff
@@ -35,18 +35,18 @@ def development_flavor_texts
   def birthday_flavor_texts
     [
       "Happy birthday! #{['🎂', '🎈', '🎉', '🥳'].sample}",
-      "<a href='https://www.youtube.com/watch?v=XNV2EyC1NrQ' target='_blank' style='color: inherit'>happy birthday</a>",
+      "<a href='https://www.youtube.com/watch?v=XNV2EyC1NrQ' target='_blank' style='color: inherit'>happy birthday</a>".html_safe,
```
maxwofford: Is this intentionally added? If I'm reading this right won't this add the literal `<a>` text to the page (not render as an html link)?

cjdenio: This is intentional. Rails will escape the flavor text without `.html_safe`, rendering a literal `<a></a>` (previously there was a blanket `.html_safe` on all flavor text)

maxwofford (approved):
Thanks for adding this!

PR #3299, opened 2022-12-08
Push tab state to browser history

YodaLightsabr:
Fixes #3294 by pushing the respective tab path (/my/settings/previews, etc.) to browser history after clicking on a tab.

https://user-images.githubusercontent.com/76178582/206368646-6c171e5d-8fcd-40b6-8fec-8d45599e227e.mov


cjdenio (commented):
You could set `data-turbo-action="advance"` on each page's `<turbo-frame>` element: https://turbo.hotwired.dev/reference/frames#frame-that-promotes-navigations-to-visits

YodaLightsabr:
Interesting. Since it uses `edit_user_path(@user)`, it puts what looks to be some identifier for the user in the URL. 


image


Would swapping it out for `my_settings_path` keep the same functionality?

maxwofford:
Updated it to use the /my/settings prefixed URLs!
Screenshot 2023-02-08 at 12 03 27

cjdenio:
i think this breaks the settings page for admins viewing other users

PR #3305, opened 2022-12-12
Fix timebomb in invoice_service/create_spec.rb

albertchae:
Now that we're past 2022-12-12 in real life, this spec would fail with
`Validation failed: Due date can't be in the past`

I replaced it with code to pick a day 14 days in the future and in the
same string format YYYY-MM-DD


albertchae:
Merging this before review to unblock new PRs @maxwofford

PR #3311, opened 2022-12-13
Use --data-only pg_dump for creating review apps

albertchae:
This is an attempt to fix https://github.com/hackclub/bank/issues/3205 based on https://help.heroku.com/ZOFBHJCJ/heroku-postgres-extension-changes-faq

As mentioned in the issue, the problem centers around installing extensions. Something I noticed during my investigation is that `bin/rails db:create` was succeeding (which would create the `g_suites` table), but then the `pg_dump`/`pg_restore` would try to drop and recreate everything. So based on this, I wondered if we could do a data only dump/restore from postgres so that we use the schema from the rake task and just load data into it. Turns out there is a `--data-only` flag for `pg_dump`.

This seems to produce a working review app!


maxwofford (approved):
Nice catch on this! I'm trying to think of any edgecases this causes (ie. migrations), but can't think of any so let's go for merge!

albertchae:
@maxwofford that's a good point about migrations. I think it is possible a migration will change the schema in a way that can cause the `pg_restore` to fail. Since I think having working review apps for at least some set of PRs is still worth it, I'm going to merge this for now but let's keep an eye out for PRs with migrations to see what happens.

The possible longer term fix is to change all our environments to install the extensions in the `heroku_ext` schema, but there are a number of negatives associated with that which the blog posts I linked in the issue discuss. Mainly that it's a very platform specific change and that we'd have to first undo all dependencies on existing extensions, then reinstall them in `heroku_ext` then change the table schema to use the extensions in the new location.

PR #3315, opened 2022-12-13
Use `has_one` for `CanonicalPendingDeclinedMapping`s

cjdenio:
i don't think there's a case where a transaction should be declined multiple times.

(all of these changes were tested locally)


maxwofford (approved):
I think this makes semantic sense– you can't decline twice. Thanks for the improvement!

I'm cool if this is already tested, but I could have sworn we had other instances of 'includes' and 'where' statements in services outside the model.

albertchae:
Love changes like this! Should we also consider making this index unique to enforce this at the database level? https://github.com/hackclub/bank/blob/5fa1e92a1182179180fc95000c3171ca09d870a7/app/models/canonical_pending_declined_mapping.rb#L14

cjdenio:
That would be awesome— I think a unique constraint would also let us do this with a single query:

https://github.com/hackclub/bank/blob/dbed1b2febf19ca7c6bb509849a613c31d25a564/app/models/canonical_pending_transaction.rb#L121-L123

Can we just add a unique constraint and call it a day? or there anything else to consider

albertchae:
I think if you are able to add the constraint on a local dump of the production database you should be OK. This would mean we don't have bad data where there are multiple `CanonicalPendingDeclinedMapping`s for a single `CanonicalPendingTransaction`. If it turns out we do have this bad data, there would probably be need to be some manual investigation into which of the multiple records we should pick and also some migration script to clean that all up.

I think the other thing to consider would be whether to `ALTER INDEX` the existing index or create a new unique index and drop the old non unique one. It seems one advantage of dropping/creating is that the [new index can be created concurrently as recommended by strong_migrations](https://github.com/ankane/strong_migrations#adding-an-index-non-concurrently). *However*, I just learned that in postgres, a unique constraint and a unique index are different 🤯 
- https://stackoverflow.com/questions/17837010/postgresql-should-i-alter-an-existing-index-or-create-a-new-one
- https://stackoverflow.com/questions/23542794/postgres-unique-constraint-vs-index

Apparently they are implemented the same in postgres and while it used to be recommended to use [constraints in the postgres 9.4 docs](https://www.postgresql.org/docs/9.4/indexes-unique.html), that advice has been removed in subsequent versions. In my experience, I've always used unique indexes for this purpose.

The last small thing to consider is whether you want to add a Rails unique validation, which usually tends to be a duplicate of whatever you have in the database. This article sums up some of the tradeoffs of doing one or both, but typically if you're only going to have one, you should have the one in the database because that'll work properly in concurrent environments.
- https://thoughtbot.com/blog/validation-database-constraint-or-both

PR #3316, opened 2022-12-13
Add `canonical_pending_transaction.decline!`

cjdenio:
Previously:

```ruby
CanonicalPendingDeclinedMapping.create!(
	canonical_pending_transaction_id: cpt.id,
)
```

Now:

```ruby
cpt.decline!
```


maxwofford (commented):
Love simplifying the interface like this. Could we just use a `find_or_create` here to push the logic down into the sql level?

PR #3317, opened 2022-12-15
Upgrade to Rails 7

albertchae:
Copy of https://github.com/hackclub/bank/pull/3261 but with a fresh review app so that https://github.com/hackclub/bank/pull/3311 is applied

Fixes https://github.com/hackclub/bank/issues/3111

Best reviewed by commits. This PR starts by updating the Rails gem and then fixing a few warnings/errors that were happening on boot. Then we go through a few of the recommended rails 7 migration steps (removing spring, running `app:update`, updating active storage).

- Update rails gem from 6.1.7 to 7.0.4
- Silence net-http warnings by adding to Gemfile
- Force uri version to 0.10.0 (default version in Ruby 2.7)
- Update maildown gem
- Remove spring dependency
- Ran bin/rails app:update
- Run ActiveStorage migration added by bin/rails app:update


albertchae:
Hey @garyhtou, I noticed on a recent rebase + deploy that I was getting this error
```
 Your lockfile doesn't include a valid resolution.

 You can fix this by regenerating your lockfile or trying to manually editing the bad locked gems to a version that satisfies all dependencies.

 The unmet dependencies are:

 * rails (>= 5.1.0, < 7.0), depended upon lab_tech-0.1.9, unsatisfied by rails-7.0.4
```

It looks like lab_tech technically doesn't support rails 7. There's an issue for it here https://github.com/RealGeeks/lab_tech/issues/42. Are we still depending on this gem or can we remove it until the next experiment (hopefully the gem is updated by then, or maybe we can take on updating the gem)? I didn't see anything in the codebase when grepping for `LabTech` but not sure if there's some other way it's being used.

garyhtou:
Let's remove lab_tech. thanks for checking!

PR #3318, opened 2022-12-15
Make transparency mode default in db

exu3:
This PR adds a migration that changes the `is_public` column default to true. Currently, demo mode accounts aren't created in transparency mode. This is one way to fix that.


cjdenio (approved):
looks great to me!

PR #3322, opened 2022-12-16
Switch Disbursement model over to relying on AASM

cjdenio:
- Best reviewed commit-by-commit.
- This is fully backwards-compatible— other parts of the code can still use a Disbursement object the same way as before
- The old state fields are still used in some parts of the codebase.

Every change in this PR is certified tested™


cjdenio:
Gonna merge this so I can continue to work on this migration. Everything here has been tested against production data.

PR #3335, opened 2022-12-21
Fix `managed_events` association in User

garyhtou:
stumbled across an error due to this


garyhtou:
wut, interesting... i don't get why those tests are failing

cjdenio:
Doesn't this conflict with line 57

PR #3341, opened 2022-12-23
Add user to transparency API transaction

YodaLightsabr:
Adds the user property to transparency API transactions if the type is `card_charge`.

image

This will be useful for counting the amount spent for Bank Wrapped.


garyhtou:
I'll review this later, but one consideration is that we'll want to add this to the API docs. API docs are generated from the v3.rb file

garyhtou (approved):
Let's go ahead and merge this so it isn't blocking for bank wrapped. However, I'm going to restructure this endpoint at some point to move `user` under a `card_charge` object; so expect breaking changes (I'll give u a heads up!)

PR #3347, opened 2022-12-24
API: Add type docs for Tags

garyhtou:


app/api/api/entities/tag.rb
```diff
@@ -3,7 +3,6 @@
 module Api
   module Entities
     class Tag < Base
-      expose :id
```
garyhtou: Not necessary since `Base` exposes `public_id` as `id`. Removing this will expose the same info (no breaking changes)

PR #3349, opened 2022-12-24
Add Card Charges to the API

garyhtou:
Adds Card Charges as a resource to the Transparency API.

image

---

Card Charges are not an actual linked object within the transaction engine. It's pretty messy. In the TX engine, they're represented with:
- StripeAuthorization
- RawPendingStripeTransaction
- RawStripeTransaction

and accessed via HcbCode using:
- HashedTransaction
- CanonicalTransaction
- CanonicalPendingTransaction

There isn't a great Model for representing Card Charges. However, since HcbCodes have easy access to all the card charge data, we'll be using HcbCodes to represent Card Charges in the API.

This PR adds an Api::Models::CardCharge class that is an ActiveRecord model and inherits from HcbCode. This definitely feels pretty hacky, but it seems to be the simplest way to give card charges their own public ID and to have "stricter" types. I've tried my best to prevent database modifications by this CardCharge model.


app/api/api/models/card_charge.rb
```diff
@@ -0,0 +1,52 @@
...
+    # HcbCodes to represent Card Charges in the API.
+    #
+    # This class is an ActiveRecord model and inherits from HcbCode. This
+    # definitely feels pretty hacky, but it seems to be the simplest way to give
+    # card charges their own public ID and to have "stricter" types. I've tried
+    # my best to prevent database modifications by this CardCharge model.
+    # ~ @garyhtou
+    class CardCharge < HcbCode
+      # Changes the Public ID Prefix to `chg` (instead of `txn` set by HcbCode).
+      set_public_id_prefix :chg
```
garyhtou: I've set the public id prefix as `chg` for Card Charges. lmk if u have a better 3 character abbreviation!

YodaLightsabr (approved):
Looks good! From the API perspective, it seems to work as expected.

garyhtou:
fyi @YodaLightsabr, now that was merged, you might need to update bank-wrapped

YodaLightsabr:
Yep! Thanks for such a quick PR. This is a much cleaner way of doing it.

PR #3350, opened 2022-12-24
Add Partner To No-No Words

sampoder:
It's slightly annoying when organisations claim to be partnered with Hack Club, for example: 

Screenshot 2022-12-24 at 3 15 59 PM


cjdenio (approved):

exu3:
I've noticed that the ops team uses "partner" in their communications with fiscally sponsored groups.
Screenshot 2022-12-24 at 9 59 10 AM

sampoder:
Yeah, I've seen that a bit but I still think the wording should be avoided externally

garyhtou:
Find it difficult to describe the relationship between HCB and a fiscally sponsored org without using technical jargon such as "fiscal sponsor". I think "partner" is a reasonable term to use.

"Sponsor" and "Donor" implies some sort of financial contribution, however, "partner" does not

sampoder:
I disagree. Hack Club is a brand that covers more things than Bank, we do a lot of work in coding related things. Bank also supports a lot of coding related organisations. Partner implies that we're collaborating at a deeper level, such as working with them on their programs etc.

I worry that if organisations continue to claim they partner with Hack Club we put our reputation at risk. Take this story for example: https://hackclub.slack.com/archives/C0266FRGV/p1671290035969289?thread_ts=1671285911.495129&cid=C0266FRGV. 

I think that we should encourage people to use the fiscal sponsor term, because that's the most truthful. 

The above screenshot is another example, I'd read that as Hack Club is collaborating with these Project Empower folks. A layman's reading of that advertisement would not conclude that we're simply offering fiscal sponsorship.

maxwofford:
I'm cool with us adding this, but maybe we can offer a better alternative on our branding page so people use that instead?

@cjdenio: thoughts on adding a branding page step to the tutorial? If so just open a ticket

cc @hackliv @daisybanks to loop in ops about this.

garyhtou:
@sampoder this is good to merge (chatted with Max and Caleb)

PR #3355, opened 2022-12-24
Add /wrapped redirect for HCB Wrapped

YodaLightsabr:
Publish Bank Wrapped by redirecting https://bank.hackclub.com/wrapped to each user's personalized wrapped page.


app/controllers/users_controller.rb
```diff
@@ -364,4 +364,8 @@ def initialize_sms_params
     end
   end
 
+  def wrapped
```
garyhtou: Try moving this above `private` on line 342

app/controllers/users_controller.rb
```diff
@@ -364,4 +364,8 @@ def initialize_sms_params
...
+    redirect_to "https://bank-wrapped.hackclub.com/wrapped?user_id=#{current_user.public_id}&orgs=#{current_user.events.select{ |event| event.is_public }.map{ |event| event.slug }.join("+")}"
```
```suggestion
    redirect_to "https://bank-wrapped.hackclub.com/wrapped?user_id=#{current_user.public_id}&org_ids=#{current_user.events.transparent.map(&:public_id).join(",")}"
```

i'd suggest using `public_id` for orgs since slugs could contain `+` i think. the transparency API will take either IDs or slugs for orgs

YodaLightsabr: Good catch, thanks!

garyhtou (approved):

PR #3366, opened 2022-12-27
Configure ActiveStorage to automatically preview HEIC files as pngs

albertchae:
Fixes https://github.com/hackclub/bank/issues/1792

**Best reviewed by commits**

### Commit 1
From https://guides.rubyonrails.org/active_storage_overview.html#transforming-images

```
If a variant is requested, Active Storage will automatically apply transformations depending on the image's format:

    Content types that are variable (as dictated by config.active_storage.variable_content_types) and not considered web images (as dictated by config.active_storage.web_image_content_types), will be converted to PNG.
```

`config.active_storage.web_image_content_types` only contains `["image/png", "image/jpeg", "image/jpg", "image/gif"]`, so by adding `"image/heic"` to `config.active_storage.variable_content_types`, we should trigger ActiveStorage's automatic conversion (using the image_processing gem, which was added several months ago in https://github.com/hackclub/bank/commit/c26a78bae964deca52bfc0e8caeebf10953fc138)

### Commit 2
We also have to pick a heroku-20 buildpack with support for HEIC because the default imagemagick is not compiled with HEIC support. There are several imagemagick with HEIC buildpacks in the heroku marketplace, so I used the same one recommended by https://mariochavez.io/desarrollo/2021/06/22/heic-support-active-storage/

Also it seems to be associated with a reputable company.

### Commit 3
Lastly, I had to fix the upload dialog to allow selecting heic files
image

We need this explicitly because although `image/*` should handle all
image MIME types https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept

as of today, heic support in browsers is non-existent https://caniuse.com/heif


albertchae:
You can see an example of a previously unpreviewable HEIC receipt on prod vs this review app

https://bank.hackclub.com/hcb/wNHkXd

https://bank-hackclu-1792-previ-obyxv2.herokuapp.com/hcb/wNHkXd

albertchae:
I initially tried the approach mentioned in https://mariochavez.io/desarrollo/2021/06/22/heic-support-active-storage/ and https://hashtagjohnt.com/how-to-generage-heic-previews-in-rails-using-activestorage.html of using a custom previewer as detailed here https://api.rubyonrails.org/classes/ActiveStorage/Preview.html, but it turned out wasn't necessary.

Hey @garyhtou , I see in https://github.com/hackclub/bank/issues/1792 that you mentioned converting heic before uploading, but this approach actually accepts heic as is but converts and uploads a separate preview png. Let me know if you think this still solves the goal of the original issue or if I should still look into converting before upload, which is the approach https://github.com/hackclub/bank/pull/2269 was taking

garyhtou (approved):
amazing!

maxwofford (approved):
This looks great, thanks for noticing the file uploader accepted types bug too– I wasn't aware that was happening. LGMT!

app.json
```diff
@@ -45,6 +45,9 @@
     {
       "url": "https://github.com/maxwofford/heroku-buildpack-sourceversion"
     },
+    {
+      "url": "https://github.com/HiMamaInc/heroku-buildpack-imagemagick-heif"
+    },
```
albertchae: Apparently I still had to manually edit this in the heroku UI for prod

image

albertchae: Confirmed this works after adding the buildpack and redeploying

garyhtou: Ah yes, i think some of the config in the repo only applies to staging apps

PR #3367, opened 2022-12-28
Cache missing receipt badge

cjdenio:
To prevent flashes of nothing while navigating between pages.


maxwofford (approved):
Great work on this! Just tested in the preview deployment and LGTM.

PR #3368, opened 2022-12-28
Shame people with 100+ missing receipts

cjdenio:
we're really just shaming sam

closes #3353


cjdenio:
imagine signing into bank one day and it's just

Screenshot 2022-12-27 at 11 37 24 PM

garyhtou:
as we should :) hehe

cjdenio:
👀

PR #3374, opened 2022-12-30
Refactor `dock_item` to use keyword arguments

cjdenio:
Before:

```ruby
<%= dock_item 'Donations',
  'Receive donations online',
  'support',
  event_donation_overview_path(event_id: @event.slug),
  nil,
  nil,
  data: { "tour-step" => "donations" } %>
```

Now:

```ruby
<%= dock_item 'Donations',
  event_donation_overview_path(event_id: @event.slug),
  tooltip: 'Receive donations online',
  icon: 'support',
  data: { tour_step: "donations" } %>
```

Inspired partially by `link_to`.

(also removed the useless `lg` option)


garyhtou:
if we're removing the lg option, do we also want to remove its css? (if it exists)

cjdenio:
It doesn't exist in CSS

garyhtou (approved):

PR #3378, opened 2022-12-31
Opt-in public listing for transparent orgs

exu3:
This PR adds a `is_indexable` column to events. If an event has transparency mode enabled, on their settings page, they can opt into public listing. It's enabled by default for new events.

(`is_indexable` will be true for new events, but `false` for existing events. Once this PR is merged, I think we can update `is_indexable` to be true for high school hackathons, robotics teams, and hq events)

In addition to the above changes, there is now also an api route http://localhost:3000/api/v3/organizations that lists events that are transparent and indexable.


cjdenio (commented):
this looks great! you can make it default to false for existing events like this:
https://github.com/hackclub/bank/blob/e04b5880e341b2447e2f9512b0e93ea903280ec1/db/migrate/20220908210922_add_activated_to_stripe_card.rb#L3-L10

app/api/api/v3.rb
```diff
@@ -206,6 +206,24 @@ def type_expansion(expand: [], hide: [])
...
+      success Entities::Organization
+      tags ["Organizations"]
+      nickname "list-transparent-organizations"
+    end
+    params do
+      use :pagination, per_page: 50, max_per_page: 500
+      use :expand
+    end
+    get :organizations do
+      present Event.where(is_public: true, is_indexable: true), with: Api::Entities::Organization, **type_expansion(expand: %w[organization user])
```
cjdenio: You might extract this into an `Event.indexable` scope

app/views/events/_form.html.erb
```diff
@@ -119,8 +119,12 @@
       <%= form.check_box :is_public %>
       <%= form.label :is_public, "Enable Transparency Mode" %>
     </div>
-    <% if false %> <%# Holiday features disabled site-wide %>
-      <div class="field field--checkbox" data-behavior="holiday_features_transparency <% if !event.is_public? %>autohide<% end %>">
+    <div class="field field--checkbox" data-behavior="additional_transparency_settings <% if !event.is_public? %>autohide<% end %>">
```
cjdenio: Eventually I think this should be a dropdown ("Private", "Unlisted", "Transparent" perhaps), but that's more complicated to implement and this is fine for now

db/schema.rb
```diff
@@ -571,6 +571,7 @@
     t.boolean "can_front_balance", default: false, null: false
     t.boolean "demo_mode", default: false, null: false
     t.datetime "demo_mode_request_meeting_at"
+    t.boolean "is_indexable", default: true
```
cjdenio: maybe there's a better name for this field— `indexable` makes me think of boring database optimizations. but idk what the better name is

cjdenio (approved):
left a few comments to consider, but this is great as-is!

garyhtou:
amazing @exu3!

app/api/api/v3.rb
```diff
@@ -206,6 +206,24 @@ def type_expansion(expand: [], hide: [])
...
 
+    desc 'Return a list of transparent organizations' do
+      summary "Get a list of transparent organizations"
+      detail "Returns a list of organizations in <a href='https://changelog.bank.hackclub.com/transparent-finances-(optional-feature)-151427'><strong>Transparency Mode</strong></a> that have opted in to public listing."
+      failure [[404]]
+      produces ['application/json']
+      consumes ['application/json']
+      success Entities::Organization
+      tags ["Organizations"]
+      nickname "list-transparent-organizations"
```
garyhtou: Can we add an `is_array true`

cc @exu3

db/schema.rb
garyhtou: i do agree with you here. maybe `is_listable`. although this would be solve by moving to an enum as mentioned in your other comment

PR #3386, opened 2023-01-03
Fix indexable scope, add hq as a category, update api docs

exu3:
Screenshot 2023-01-03 at 8 47 06 AM


app/api/api/v3.rb
```diff
@@ -210,6 +210,7 @@ def type_expansion(expand: [], hide: [])
       summary "Get a list of transparent organizations"
       detail "Returns a list of organizations in <a href='https://changelog.bank.hackclub.com/transparent-finances-(optional-feature)-151427'><strong>Transparency Mode</strong></a> that have opted in to public listing."
       failure [[404]]
+      is_array true
```
garyhtou: thank you!

PR #3390, opened 2023-01-03
Expose pagination headers to /api/*

YodaLightsabr:
https://hackclub.slack.com/archives/C047Y01MHJQ/p1672575369699689

Pagination headers aren't currently exposed to the `fetch` API on other hosts. This adds them to `/api/v3/*`.


config/initializers/cors.rb
```diff
@@ -17,4 +17,12 @@
              headers: :any,
              methods: %i[get post put patch delete options head]
   end
+
+  allow do
+    resource "/api/v3/*",
```
garyhtou: maybe we can do
```suggestion
    resource "/api/*",
```

garyhtou:
fyi, got this error in the logs
```
config/initializers/cors.rb:26: syntax error, unexpected end-of-input, expecting `end'
```

YodaLightsabr:
image

It seems to be working now @garyhtou

config/initializers/cors.rb
```diff
@@ -13,6 +13,11 @@
       end
     end
 
+    resource "/api/*",
+             headers: :any,
+             methods: %i[get post put patch delete options head],
```
garyhtou: i wonder if this is necessary

garyhtou (approved):
perfect! good to merge as is!

config/initializers/cors.rb
YodaLightsabr: I can revisit this after Bank Wrapped is done and try that out.

PR #3391, opened 2023-01-04
Make sorting less jank again!

exu3:
Related to #3381 
for these card things
Screenshot 2023-01-03 at 8 52 59 PM


cjdenio (commented):
Super picky, but would it look nicer to stack the avatars in reverse? So the most recent team members are on top

PR #3394, opened 2023-01-04
Expose org category in API

exu3:


garyhtou:
i think we should underscore the categories. in some languages it's more difficult to work with data that contain spaces

garyhtou (approved):
cool!!

garyhtou:
interesting, it looks like they aren't being underscored

PR #3395, opened 2023-01-06
Fix pdf to image preview conversion by upgrading poppler library

albertchae:
https://github.com/hackclub/bank/issues/3389

Related to https://github.com/hackclub/bank/pull/3397, which
demonstrates this working in the local docker environment.

The airbrake error is different, but if I try uploading that same pdf
locally, I get this error
Syntax Error: Gen inside xref table too large (bigger than INT_MAX)

This appears to be this bug, especially because the pdf is from docusign
https://bugs.launchpad.net/ubuntu/+source/poppler/+bug/1944453
https://stackoverflow.com/questions/66636441/pdf2image-library-failing-to-read-pdf-signed-using-docusign/74313286

This PR updates the poppler version on the deployed heroku environment
to 21.03.0, specified by

https://github.com/hackclub/heroku-buildpack-poppler

The README covers why we built our own buildpack

~~https://elements.heroku.com/buildpacks/k16shikano/heroku-buildpack-poppler~~
~~https://github.com/k16shikano/heroku-buildpack-poppler~~

~~Criteria for picking this buildpack:~~
~~This is the version recommended by the stackoverflow answer above.~~
~~There are only 2 heroku buildpacks in the marketplace with poppler versions new enough, and the other one has its git repo archived.~~

We can verify the poppler version by running the following command on
the heroku deployment
```
$ pdftoppm -v
pdftoppm version 21.03.0
Copyright 2005-2021 The Poppler Developers - http://poppler.freedesktop.org/
Copyright 1996-2011 Glyph & Cog, LLC
```

N.B. Upgrading from heroku-20 to heroku-22 would solve this also,
because heroku-22's poppler version is 22.02.0. But heroku-22 also
requires a minimum of ruby 3.1.3.

One nuance, even with the upgraded poppler version, because
ActiveStorage automatically creates and stores preview images, for
existing pdfs with this issue, we have to go and delete ActiveStorage
blobs with corrupted previews. Then they will be converted again the
next time they are previewed. Any new uploaded pdfs will work fine.


albertchae:
We can see this works in the review app for this PR
https://bank-hackclu-3389-updat-nbzqg6.herokuapp.com/the-innovation-circuit/documents

But it doesn't work in the review app for this PR that only made this fix locally
https://github.com/hackclub/bank/pull/3397
https://bank-hackclu-3389-updat-dsneur.herokuapp.com/the-innovation-circuit/documents

app.json
```diff
@@ -48,6 +48,9 @@
     {
       "url": "https://github.com/HiMamaInc/heroku-buildpack-imagemagick-heif"
     },
+    {
+      "url": "https://github.com/k16shikano/heroku-buildpack-poppler.git"
```
albertchae: I'm not 100% comfortable using this buildpack since it has such few numbers plus the `tar.gz` here isn't the exact same as the one served from https://poppler.freedesktop.org/releases.html (these are `tar.xz` files, not `gz`). I might recommend making our own buildpack off a fork...

garyhtou: if you think it might be worth creating our own fork, then let's go for it

garyhtou (approved):

app.json
albertchae: Yeah I talked to @maxwofford and I think we should. I don't have permissions to fork this under the hackclub org, can you or @maxwofford do that and I can make the necessary changes to pull in a newer version of poppler?

garyhtou: which repo would you like to fork?

albertchae: Can you fork this one https://github.com/k16shikano/heroku-buildpack-poppler?

albertchae: Max forked it for me in https://github.com/hackclub/heroku-buildpack-poppler

garyhtou: a whoops, sorry missed your message! please @garyhtou next time 😄

albertchae: All good no worries @garyhtou

maxwofford (approved):

garyhtou (approved):

albertchae:
Updated buildpack for prod deployment too 
image

PR #3396, opened 2023-01-06
Add Flipper flag for disabling Google Workspace

cjdenio:
Only affects orgs that don't have it set up already.


app/views/events/_nav.html.erb
```diff
@@ -42,7 +42,7 @@
       event_g_suite_overview_path(event_id: @event.slug),
       tooltip: (@event.hardware_grant? ? 'Not available for Winter Hardware Wonderland accounts' : 'Set up domain Gmail'),
       icon: 'google',
-      disabled: @event.hardware_grant? %>
+      disabled: @event.hardware_grant? unless !@event.g_suites.exists? && Flipper.enabled?(:disable_google_workspace, @event) %>
```
garyhtou: thoughts on making the flag :google_workspace this way we dont need to work with double negatives?

cjdenio: The tricky thing is that Flipper flags are disabled by default, so dev environments would have to opt-in to Google Workspace. It does look weird though

garyhtou (changes requested):
Let's also remove the reference to Google Workspace here if the flag is flipped

image

garyhtou (approved):
ty!

PR #3397, opened 2023-01-06
[DO NOT MERGE local docker only] Use updated version of poppler PDF library to handle converting certain pdfs to images

albertchae:
https://github.com/hackclub/bank/issues/3389

The airbrake error is different, but if I try uploading that same pdf
locally, I get this error
`Syntax Error: Gen inside xref table too large (bigger than INT_MAX)`

This appears to be this bug, especially because the pdf is from docusign
https://bugs.launchpad.net/ubuntu/+source/poppler/+bug/1944453
https://stackoverflow.com/questions/66636441/pdf2image-library-failing-to-read-pdf-signed-using-docusign/74313286

To verify that updating poppler would fix this, this commit is just a
POC in the local docker. It turns out updating the poppler version is
pretty complex, but I was able to follow the manual build instructions
here to do it in docker

https://stackoverflow.com/questions/66636441/pdf2image-library-failing-to-read-pdf-signed-using-docusign


albertchae:
@garyhtou @maxwofford as an FYI only, related to https://github.com/hackclub/bank/issues/3389

The real fix is https://github.com/hackclub/bank/pull/3395

albertchae:
Closing this PR now since a lot of the steps are included in https://github.com/hackclub/heroku-buildpack-poppler/pull/5 now, and we aren't going to merge this anyway.

PR #3407, opened 2023-01-10
Fully enable Turbo

cjdenio:
behind a feature flag!


cjdenio:
Would anyone like to review this before I merge it?

maxwofford:
Yes– can you grab me IRL to review?

app/views/users/edit_featurepreviews.html.erb
```diff
@@ -9,6 +9,29 @@
 <%= render 'settings_nav', active:'previews' %>
 <section>
     <h2 id="feature-previews">Feature previews</h2>
+        <% unless Flipper.enabled?(:turbo) %>
+            <% admin_tools "mb3" do %>
+                <div class="card pb0">
+                    <div class="card__banner card__banner--top flex justify-between items-center feature-detailed-events-list">
+                        <h3 class="h1 p2 my0 color-black">Turbo</h3>
+                        <% if Flipper.enabled?(:turbo, current_user) %>
```
garyhtou: let's include a date in the feature flag. other than that, let's merge it into main in a few days!

(the reason i think we should wait a few days is so that we can work out any Rails 7 regressions. it might get confusing to debug regressions when we have two large changes)

garyhtou:
im excited for this!

maxwofford (approved):
Reviewed IRL with @cjdenio. Please address the comment that @garyhtou added. Other than that LGTM!

cjdenio:
approved by @maxwofford IRL. @albertchae, FYI that this is a large frontend change and could possibly affect parts of #3469

albertchae:
@cjdenio thanks for the heads up. Exciting change!

garyhtou:
Are we now able to remove `data: { turbo: true }` from our links?

cjdenio:
Not yet, since Turbo is still disabled globally for regular users

garyhtou:
ah, right. forgot this was feature flagged

PR #3408, opened 2023-01-10
Set up ESLint

cjdenio:
also known as the 'react best practice enforcer 9000'


maxwofford (approved):
Thanks for adding this!

PR #3409, opened 2023-01-11
Require explicit intent to view card details

InternetRamen:
Closes #3385 
this PR
 - Requires button press to view card details such as CVC and billing address

**Best reviewed by code standards?**
I don't know if taking a controller from a pending PR is a good idea.

I would like some help with tracking events with ahoy. What's the best way to do it?


cjdenio (commented):
I think it's better to not even send the card details to the browser until they're requested. One possible option might be:

- Only load the card details if `?show_details=true` is set
- Wrap the card info box in a `<turbo-frame>`
- Make the "Show more details" button link to `/stripe_cards/<id>?show_details=true`
  - (it won't reload the whole page, just the `<turbo-frame>`)
- Track every request with `?show_details=true` on the backend, like
```rb
if params[:show_details]
  ahoy.track "Card details shown", stripe_card_id: @stripe_card.id
end
```

garyhtou (commented):
Thanks for helping review this @cjdenio! 

Sorry. should have given a little more detail in https://github.com/hackclub/bank/issues/3385. I had the same design in mind as Caleb. Let's move this to the turbo frame design.

Also, as for the UI design, I like what Stripe has. It let you see exactly what information will be revealed (and what's currently hidden).
image

Only the card number and CVC need to stay hidden. The rest (such as address) can remain visible at all times.

Also, this only needs to apply to virtual cards since we don't have card numbers and CVC for physical cards

YodaLightsabr:
image
I'm not able to log in on the heroku deploy

cjdenio (changes requested):
This is an awesome start! For the UI, I like Stripe's design— specifically, what if there was a clickable 👁️ (`view` or `view-fill`) icon next to the card number?

app/controllers/stripe_cards_controller.rb
```diff
@@ -70,8 +70,14 @@ def activate
...
 
-    authorize @card
 
+    if params[:show_details]
+      ahoy.track "Card details shown", stripe_card_id: @card.stripe_id
+    end
+
+    @show_card_details = params[:show_details] == "true"
+
+    authorize @card
```
garyhtou:
```suggestion
    authorize @card

    if params[:show_details] == "true"
      ahoy.track "Card details shown", stripe_card_id: @card.id
    end

    @show_card_details = params[:show_details] == "true"
```

We'll want to `authorize @card` first.

Switch `stripe_card_id` to use our ID rather than stripe id. we can add stripe's add too if we want.

Also, I made both conditions the same `params[:show_details] == "true"`. This way when track an ahoy event, we can also be sure that `@show_card_details` was `true` and that details were viewed.

app/views/stripe_cards/show.html.erb
```diff
@@ -88,79 +88,98 @@
...
+            </p>
+          <% end %>
+          <% if @card.virtual? %>
+
+              <% if @card.user == current_user %>
+                <div class="fs-mask">
+                  <strong>
+                    Card number
+                  </strong>
+                  <div class="flex justify-between">
```
garyhtou: not necessary: i wonder if we can use a monospaced font here to prevent layout shift

garyhtou (approved):
looks good!

just one change

app/views/stripe_cards/show.html.erb
InternetRamen: Will try

YodaLightsabr: There's a monospaced font that appears on the cards overview page, maybe we could use that

garyhtou:
> @garyhtou Is this what you're looking for?

My goal is for the last 4 digits to not move on the screen when switching between showing/hiding details. The reason it moves is because the font's character width for `•` vs. numbers such as `1`, `2`, `3`, etc. are different. This can be fixed by changing the font-family to one that is monospaced. Monospaced fonts have the same width for all characters.

<table>
<tr>
	<td>Monospaced font
	<td>Mon-monospaced font
<tr>
	<td><pre>•••• •••• •••• ••••
1234 5678 9012 3456</pre>
	<td>•••• •••• •••• ••••<br/>
1234 5678 9012 3456
</table>

<details>
<summary>Here's a screenshot of what the table above should look like in case github renders it funny</summary>

</details>

however, this is totally unnecessary! so don't let this block this pr. please feel free to merge as is

---

as for the flexbox and justify-between, I'm fine with it either way!

cjdenio (approved):

PR #3411, opened 2023-01-11
Fix alignment issue on card names

YodaLightsabr:
Fixes the alignment issue on the cards page with CSS `vertical-align`

Before:
image

After:
image


garyhtou (approved):
Heck yeah! thanks for fixing this!

Should this CSS class be applied to all `user_mention`

YodaLightsabr:
I think it depends on the parent element and if there are other user mentions that aren't aligned properly

YodaLightsabr:
I just checked a few other instances and it appears to be aligned properly

PR #3416, opened 2023-01-12
API Transactions speed up: Pluck + Remove N+1 query

garyhtou:
Improves `/api/v3/organizations/hq/transactions` endpoint from 2.3 secs to 0.8 sec in development.

BEFORE
```
Response timings (50 runs):
Average = 2282.62 ms
Max = 4869 ms
Min = 1786 ms
```

AFTER
```
Response timings (50 runs):
Average = 840.66 ms
Max = 2467 ms
Min = 660 ms
```

This improvement comes from
- Plucking only `hcb_code` from results (return less data)
- Removing N+1 query when getting HcbCode objects


garyhtou:
closing as stale as part of triage call on 6/23/2023

there are plans to hopefully rewrite the ledger listing code, which should invalidate the need for this

PR #3417, opened 2023-01-12
Revert "API Card Charge speed up: Pluck + Remove N+1 query"

garyhtou:
Reverts hackclub/bank#3415


garyhtou:
something's a lil fishy

PR #3418, opened 2023-01-12
API Card Charge speed up: Pluck + Remove N+1 query

garyhtou:
Improves from 17 sec to 3 sec in development.

but it's kinda a mess due to the old SQL code


app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -206,13 +216,16 @@ def canonical_transactions_grouped
             #{event.can_front_balance? ? "#{pt_group_sql}\nunion" : ''}
             #{ct_group_sql}
           ) q1
-          #{tag_modifier}
+          #{filter_tags}
+          #{filter_hcb_code_type}
           order by date desc, pt_ids[0] desc, ct_ids[0] desc
```
garyhtou: :( this `order` isn't "strict enough". results are still accurate in the sense that it returns the same objects in chronological order. however, the order between the new and old results can be different due to having different query plans and the `order` not being strict/specific enough to keep objects with the same rank from shifting

garyhtou:
hm, im considering completely rewriting the code for listing an org's transactions

app/services/pending_transaction_engine/pending_transaction/all.rb
```diff
@@ -3,10 +3,11 @@
 module PendingTransactionEngine
   module PendingTransaction
     class All
```
garyhtou: The code in this class works perfectly fine. It's settled transactions class that is the issue

garyhtou:
closing as stale as part of triage call on 6/23/2023

there are plans to hopefully rewrite the ledger listing code, which should invalidate the need for this

PR #3424, opened 2023-01-13
Handle force capture transactions (+refunds) from Stripe

cjdenio:
Refunds right now are super wack because they (sometimes) don't belong to a Stripe authorization. This fixes that!

Here's a force capture transaction:
Screenshot 2023-01-13 at 10 27 44 PM

And here's a refund:
Screenshot 2023-01-13 at 10 28 01 PM

And here's that refund in the ledger!
Screenshot 2023-01-13 at 10 21 35 PM

Resolves #2555


app/models/canonical_transaction.rb
```diff
@@ -141,6 +141,10 @@ def stripe_card
     end
   end
 
+  def stripe_refund?
+    raw_stripe_transaction&.stripe_transaction&.dig("type") == "refund"
```
garyhtou: I'd move this logic into `RawStripeTransaction` and then call `raw_stripe_transaction&.stripe_transaction&.refund?` here

app/models/hcb_code.rb
```diff
@@ -163,6 +163,12 @@ def stripe_cardholder
     pt.try(:stripe_cardholder) || ct.try(:stripe_cardholder)
   end
 
+  def stripe_refund?
+    # hcb code must be 000 (unknown) for this to be considered a standalone refund
+    # note that HCB-600- hcb codes sometimes contain refunds grouped in with the original transaction
```
garyhtou: In my opinion, standalone stripe refunds should be `601` (or in the `6xx` range ). This is because `000` should be reserved for transactions that basically have no additional metadata or linked objects.

but that's gonna require a decent amount of work. if you're able to make it `601`, that'd be great! but if not, I think this should be okay for now without creating too much tech debt

https://github.com/hackclub/bank/blob/94a91292ec187a16d0f125c55f1e837cdf549144/app/services/transaction_grouping_engine/calculate/hcb_code.rb#L8-L20

app/views/canonical_transactions/_canonical_transaction.html.erb
```diff
@@ -123,7 +127,7 @@
       </div>
 
       <% if ct.local_hcb_code %>
-        <%= list_badge_for ct.local_hcb_code.receipts.size, "receipt", "payment-docs", optional: !ct.raw_stripe_transaction, required: ct.raw_stripe_transaction && !ct.local_hcb_code.no_or_lost_receipt? %>
+        <%= list_badge_for ct.local_hcb_code.receipts.size, "receipt", "payment-docs", optional: !ct.raw_stripe_transaction || ct.local_hcb_code.stripe_refund?, required: ct.raw_stripe_transaction && !ct.local_hcb_code.no_or_lost_receipt? %>
```
garyhtou: this receipt require/optional logic should all be moved into `HcbCode`, but won't worry about that for this PR

app/views/hcb_codes/_rich_metadata.html.erb
```diff
@@ -24,7 +24,7 @@
 <% end %>
 
 <%# stripe card %>
-<% if @hcb_code.stripe_card? && @hcb_code.stripe_card.present? %>
+<% if (@hcb_code.stripe_card? && @hcb_code.stripe_card.present?) || (@hcb_code.unknown? && @hcb_code.ct.raw_stripe_transaction.present?) %>
```
garyhtou: having a `601` can help us consolidate all this logic into one place and avoid having to handle these edge cases in multiple places

app/views/hcb_codes/_stripe_card.html.erb
```diff
@@ -85,52 +85,56 @@
...
-        <% end %>
-      </p>
-      <p>
-        <strong>Charge method</strong>
-        <% if organizer_signed_in? %>
-            <action data-behavior="modal_trigger" data-modal="verification_details" class="pointer" tabindex="0">
-              <span>
-              <%= @hcb_code.pt.raw_pending_stripe_transaction.stripe_transaction["authorization_method"].humanize %>
+                <%= country_to_emoji (@hcb_code.pt&.raw_pending_stripe_transaction&.stripe_transaction || @hcb_code.ct.raw_stripe_transaction.stripe_transaction)["merchant_data"]["country"] %>
+                <%= (@hcb_code.pt&.raw_pending_stripe_transaction&.stripe_transaction || @hcb_code.ct.raw_stripe_transaction.stripe_transaction)["merchant_data"]["name"] %>
```
garyhtou: Let's move this into `HcbCode#merchant` that returns a hash with `county` and `name`

garyhtou (changes requested):
Glad to see you working on improving this!

Some feedback; the most important being creating a `601` HcbCode type for standalone stripe refunds. This is because 000 should be reserved for transactions that basically have no additional metadata or linked objects.

cjdenio:
For posterity: ran this to update the HCB code for the affected transactions:

```ruby
CanonicalTransaction
  .joins(canonical_hashed_mappings: { hashed_transaction: [ :raw_stripe_transaction ] })
  .where("raw_stripe_transactions.stripe_authorization_id IS NULL")
  .each(&:write_hcb_code)
```

PR #3425, opened 2023-01-14
Improve organization team page

YodaLightsabr:
This PR adds a second column to the team page (`/:org/team`) on larger screens.

Before:

After:

I tested it signed out, in admin mode, and in dark mode.
It also adds the [`.grid--ultrawide` class](https://github.com/hackclub/bank/blob/f020334677c9df22385d9d6810f37c1571a55938/app/assets/stylesheets/components/_layouts.scss#L128).


garyhtou (commented):
I love the two columns! Yeah, some of these pages are getting kinda long for large events and this is a great solution.

Some feedback:
image

- I think the grid column gap could be a little smaller (the area boxed in green)
- On both the left and right sides, I think the edge of the cards should line up with the width of the heading (similar to the Before image). For example, the left cards should line up with the purple line
- Lastly, I know this looks kinda weird, but for the `creator_bar`, usually, the avatar hangs out to the left a little; so that the start of the name (or admin badge) aligns with the left of the card. I think we should keep that so other pages look consistent (see the Before image as an example)

would love to hear your thoughts on those!

garyhtou:
Also, here's what the mobile view looks like:
image

YodaLightsabr:
definitely, the spacing is pretty janky. I'll play around with the CSS and try to find a better way of spacing them.

maxwofford:
I just rebased from main and wasn't able to replicate the mobile issues– I think some other recent changes fixed that issue.
IMG_9149

maxwofford:
Screenshot 2023-02-08 at 11 27 08

Looks like it is off sides on mobile for admin users, but that's not introduced by this PR, so I'm going to go ahead with this change.

maxwofford:
Hmmmm.... now that i'm playing around further it looks like there is an issue with sizing that happens with long emails or names pushing the `<section>` to overflow it's `grid` parent. I'm not sure how to contain it properly & overflow isn't fixing this the way I expect it to.

maxwofford:
Going to close this for now because it hasn't been touched for a month, but feel free to re-open if it should be!

PR #3429, opened 2023-01-15
Delete redundant commented out mini_magick

albertchae:
Looks like this line has been commented out since the initial commit
https://github.com/hackclub/bank/commit/29d929e3b57240dfb126391127c16db2e7afe8c9#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55fR31

but then was added back later in the Gemfile here
https://github.com/hackclub/bank/commit/47943de686ba1610934dc67b119444b3c75eb21a


albertchae:
Super small change, so I'm going to merge without review

PR #3431, opened 2023-01-16
Implement activity since filter on admin events page

albertchae:
Closes https://github.com/hackclub/bank/issues/3243

I interpreted "activity" to mean canonical_transactions only, but we
could expand this definition later.

I also added an index to `canonical_transactions.date`. See commit message in https://github.com/hackclub/bank/pull/3431/commits/441e186afa395bd4000a1e0a6bd1787437ce8f13 for explain plan results.


hackliv (approved):

db/schema.rb
```diff
@@ -1467,15 +1466,14 @@
   add_foreign_key "donations", "donation_payouts", column: "payout_id"
   add_foreign_key "donations", "events"
   add_foreign_key "donations", "fee_reimbursements"
-  add_foreign_key "emburse_card_requests", "emburse_cards"
```
maxwofford: Were these changes intentional?

albertchae: good catch, no they weren't. I've noticed sometimes restoring from a heroku db dump sometimes messes up emburse related tables (and thus messes up the local schema) but i haven't investigated why yet.

albertchae: Fixed the schema.rb, going to merge now

PR #3435, opened 2023-01-18
Try removing asset pipeline to see what's broken and how much slug size we save

albertchae:
Remove sassc rails


albertchae:
OK neither commit built successfully on heroku and overall this is a bit more involved than I realized, plus it looks like we need to be on rails 7 first. I'll investigate more later

PR #3446, opened 2023-01-20
Run Rubocop inside views

cjdenio:
This improves a lot of inconsistent formatting inside ERB tags.

###### _what's better than normal code? immaculate code!_


app/views/emburse_transactions/show.html.erb
```diff
@@ -47,7 +47,7 @@
           <% if @emburse_transaction.emburse_card.nil? %>
             -
           <% else %>
-            <%= "#{@emburse_transaction.emburse_card.full_name}" %>
+            <%= @emburse_transaction.emburse_card.full_name.to_s %>
```
maxwofford: I'm seeing a lot of these autocorrections that add an extra `to_s`.

app/views/admin/partner_organizations.html.erb
```diff
@@ -34,7 +34,7 @@
             <%= event.name.upcase %>
           <% end %>
         </td>
-        <td style="<%= 'color: lightgray;' if event.balance === 0 %>">
+        <td style="<%= 'color: lightgray;' if event.balance == 0 %>">
```
maxwofford: Small style change, but should this be `event.balance.zero?`?

app/views/emburse_transactions/show.html.erb
cjdenio: Good catch— that's just because `"#{var}"` is identical to `.to_s` and the linter can't be sure that `full_name` is a string.

PR #3448, opened 2023-01-22
Upgrade to rails 7 (try again after slug size increase)

albertchae:
Same PR as https://github.com/hackclub/bank/pull/3317 (essentially a revert of a revert https://github.com/hackclub/bank/pull/3433)

- Update rails gem from 6.1.7 to 7.0.4
- Silence net-http warnings by adding to Gemfile
- Force uri version to 0.10.0 (default version in Ruby 2.7)
- Update maildown gem
- Remove spring dependency
- Ran bin/rails app:update
- Run ActiveStorage migration added by bin/rails app:update
- Remove lab_tech because it's not compatible with rails 7


albertchae:
Hooray, deploy succeeded!

```
-----> Compressing...

       Done: 492M

-----> Launching...

 !     Warning: Your slug size (492 MB) exceeds our soft limit (300 MB) which may affect boot time.

       Released v3248

       https://bank-hackclub.herokuapp.com/ deployed to Heroku
```

PR #3451, opened 2023-01-22
Preview docx files with activestorage-office-previewer

albertchae:
Requires installing libreoffice too


albertchae:
lol, libreoffice from apt takes the slug size to 1 GB
` !     Compiled slug size: 1G is too large (max is 500M).`

Tried using unoconv as an alternative (this gist https://gist.github.com/zealot128/5534cb70a06dcce96dba1fea7274fae5), but still too big
` !     Compiled slug size: 793.8M is too large (max is 500M).`

Looks like we have to stick to no docx previews for now

PR #3452, opened 2023-01-23
Smaller hidden projects

InternetRamen:
image
Closes #3422


cjdenio (approved):
Thanks! I'd also apply the `muted` class, but this looks great otherwise!

PR #3454, opened 2023-01-24
Fix more deprecated Rails 7 render templates with extensions

albertchae:
Same change as https://github.com/hackclub/bank/pull/3450 but in a few
other places I found

Fixes https://github.com/hackclub/bank/issues/3449

Rails changed this behavior in https://github.com/rails/rails/pull/39164


maxwofford (approved):
Looks good and runs fine on staging!

PR #3466, opened 2023-01-27
Add AI-generated transaction custom memo

kognise:
Adds an admin-only experimental "AI Rename" button to the transaction list. Clicking it suggests a new name, and a save button must be clicked to confirm.

Benefits: for simple transactions it's now absurdly fast to give them names! Everything updates via Turbo Frames too. Workflow go brrr.

Considerations:

- The model isn't fantastic right now, I have a couple improvements to make. Perhaps it should only suggest renames with a high enough level of confidence.
- This could be put under a Flipper gate, but I decided not to because it's admin-only.
- I have no idea how Bank works and probably screwed something up royally here.

Cost: should be next to negligible; absolutely maximum is $3 per month but that's already very conservative.


app/controllers/hcb_codes_controller.rb
```diff
@@ -44,6 +44,33 @@ def show
...
+      response = conn.post(
+        "https://api.openai.com/v1/completions",
+        model: "davinci:ft-hack-club-2023-01-27-18-45-40",
+        prompt: "Amount: $#{amount}\nTransaction Description: #{description}\nHuman-Readable Memo:",
+        echo: false,
+        stop: "\n",
+        temperature: 0,
+        top_p: 1
+      )
+      @ai_memo = response.body["choices"][0]["text"].strip
```
garyhtou: Let's move this to a service. See the `app/services/hcb_codes` folder

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -58,14 +58,23 @@
...
+            <span title="<%= pt.local_hcb_code.memo %>">
+              <% if pt.declined? %>
+                DECLINED:
+              <% else %>
+                PENDING:
+              <% end %>
+
+              <%= pt.local_hcb_code.memo %>
+
+              <% if current_user&.admin? && pt.amount_cents < 0 && /^\p{Emoji}/ !~ pt.local_hcb_code.memo %>
```
garyhtou: If `/^\p{Emoji}/ !~ pt.local_hcb_code.memo` is to check if a custom memo has been set, then I think we can use
```ruby
pt.custom_memo.nil?
```

app/controllers/hcb_codes_controller.rb
```diff
@@ -44,6 +44,33 @@ def show
...
+      conn = Faraday.new(headers: {"Authorization" => "Bearer #{Rails.application.credentials.openai.api_key}"}) do |f|
+        f.request :json
+        f.request :retry
+        f.response :json
+      end
+  
+      amount = "%.2f" % (-@hcb_code.amount_cents / 100 / 100)
+      description = @hcb_code.ct.try(:less_smart_memo) || @hcb_code.pt.try(:friendly_memo)
+  
+      response = conn.post(
```
garyhtou: We should handle a non-2XX response

app/controllers/hcb_codes_controller.rb
```diff
@@ -44,6 +44,33 @@ def show
...
+    authorize @hcb_code
+
+    if params[:gen_memo]
+      conn = Faraday.new(headers: {"Authorization" => "Bearer #{Rails.application.credentials.openai.api_key}"}) do |f|
+        f.request :json
+        f.request :retry
+        f.response :json
+      end
+  
+      amount = "%.2f" % (-@hcb_code.amount_cents / 100 / 100)
```
```suggestion
      amount = @hcb_code.amount.abs.to_s
```

app/controllers/hcb_codes_controller.rb
```diff
@@ -44,6 +44,33 @@ def show
...
+      description = @hcb_code.ct.try(:less_smart_memo) || @hcb_code.pt.try(:friendly_memo)
```
garyhtou: great job using friendly memo rather than just `memo`! (sometimes `memo` can contain sensitive info such as our bank account number)

app/controllers/canonical_pending_transactions_controller.rb
```diff
@@ -29,12 +29,14 @@ def set_custom_memo
...
     ::CanonicalPendingTransactionService::SetCustomMemo.new(attrs).run
 
-    if @current_user.admin?
+    if @current_user.admin? && !params[:canonical_pending_transaction][:fronted].nil?
       @canonical_pending_transaction.update(fronted: params[:canonical_pending_transaction][:fronted])
     end
 
-    flash[:success] = "Renamed pending transaction"
-    redirect_to @canonical_pending_transaction.local_hcb_code
+    if !params[:no_flash]
```
garyhtou: sorry is this very picky lol
```suggestion
    unless params[:no_flash]
```

garyhtou (commented):
amazing @kognise! I love this.

a couple of comments

app/controllers/canonical_transactions_controller.rb
```diff
@@ -31,8 +31,10 @@ def set_custom_memo
     }
     ::CanonicalTransactionService::SetCustomMemo.new(attrs).run
 
-    flash[:success] = "Renamed transaction"
-    redirect_to @canonical_transaction.local_hcb_code
+    if !params[:no_flash]
```
maxwofford:
```suggestion
    unless params[:no_flash]
```

kognise:
[fyi, putting this pr on hold for ~1wk to prioritize other work]

maxwofford:
Amazing work here @kognise! @cjdenio @garyhtou because most of the issues here are code-style/CI issues, let's just take on amending the PR to be good to go so @kognise doesn't have to.

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
maxwofford: I think the intent here is to check for custom memos prefixed with emojis, because the trained model always prefixes memos with emojis.

app/controllers/hcb_codes_controller.rb
maxwofford: Just added this! It's simple, but at least we'll get all the details surfaced up in Airbrake so we can start deciding which error cases to spend eng time on handling.

maxwofford:
Want to flag, I found I hit a 429 ratelimit after generating only about 20 memos in the span of about 5 minutes. That was with paid API tokens. Depending on how this ends up getting used we might want to move inference into a background job that inferences new transactions in advance throughout the day.

cjdenio (approved):

garyhtou (approved):

cjdenio:
Rolling back since this breaks memo overflow (for non-admins and Transparency Mode too)

Screenshot 2023-01-31 at 2 51 26 PM

PR #3470, opened 2023-01-30
Let admins delete demo accounts

cjdenio:
This PR adds a button that lets admins clean up demo accounts. It's all `acts_as_paranoid`-ified.

I'm thinking users might be able to delete their demo accounts in the future as well.

Screenshot 2023-01-30 at 12 46 14 PM


app/policies/event_policy.rb
```diff
@@ -53,7 +53,7 @@ def update?
   end
 
   def destroy?
-    user&.admin?
+    user&.admin? && record.demo_mode?
```
maxwofford: Was there an admin destroy route that used to exist for regular orgs?

maxwofford (commented):
Thanks! I'm sure the ops team will be happy to have this.

How does this handle things like https://github.com/hackclub/bank/blob/b7f7e5f64e9bb23052f69404a51ffe5105798d6e/app/services/user_service/can_open_demo_mode.rb#L18?

app/policies/event_policy.rb
cjdenio: Yeah, there already was one (though it's currently broken and I can't find it in the UI)

cjdenio:
Since the event is effectively deleted, another demo account slot will be freed up when a demo account is deleted.

garyhtou (approved):
LGTM. just to double check, it's not possible for demo mode events to have financial data, right? (no disbursements or card charges, etc.)

cjdenio:
Yep! Demo mode accounts don't have any associated financial data

cjdenio:
Reviewed by @maxwofford IRL.

PR #3476, opened 2023-01-31
Create Stripe top-ups for refunded donations

cjdenio:
Fixes #3474, fixes #2064, fixes #3739


app/controllers/stripe_controller.rb
```diff
@@ -136,6 +136,22 @@ def handle_charge_dispute_funds_withdrawn(event)
     head 200
   end
 
+  def handle_charge_refunded(event)
+    charge = event[:data][:object]
+
+    donation = Donation.find_by(stripe_payment_intent_id: charge[:payment_intent])
+    return unless donation
+
+    donation.mark_refunded! unless donation.refunded?
```
garyhtou:
```suggestion
		return if donation.refunded?
```

This should be idempotent. So, we don't want to continue on (and make a topup) if the donation was already refunded

garyhtou (commented):
(i see that the pr might not be complete, just wanted to add a small comment tho!)

app/controllers/stripe_controller.rb
cjdenio: This is somewhat intentional since `DonationService::Refund` already marks the donation as refunded. Maybe the service shouldn't mark it as refunded, and just wait for the webhook?

PR #3478, opened 2023-01-31
New signup flow

cjdenio:


app/models/organizer_position.rb
```diff
@@ -32,14 +32,14 @@ class OrganizerPosition < ApplicationRecord
   belongs_to :user
   belongs_to :event
 
-  has_one :organizer_position_invite
+  has_one :organizer_position_invite, required: false
```
garyhtou: I'm somewhat a fan of using `Partner.representative` (where the Partner is Bank) as the inviter in these scenarios (when users create a new event for themselves). This has a few benefits:

- Every user has an invite to their orgs. Lets us keep Organizer Position Invite as required
	 - Provides us with a more consistent UI. No need to check if invite is `nil?` when showing `creator_bar`
- For initial users, it'll show that Hack Club Bank invited them.
- If/when we integrated with partners (e.g. GiveButter), it will show that GiveButter invited the initial user.

cjdenio: Hmm. I agree with keeping things consistent, but I don't think the user should receive an "invite" to an org they just created on our interface.

garyhtou: could we auto accept their invite?

PR #3483, opened 2023-02-01
Batch of small or relatively safe Rails 7 framework defaults

albertchae:

This PR is best reviewed by commits, but combines several settings which I think are fairly safe to turn on together. If anyone disagrees, I can split it out into a separate PR.

- [TEST ONLY] turn on fixture foreign key checks - bank specs aren't using fixtures anymore, so this setting is inconsequential
- Enable smarter video previewer, we only have 4 videos in bank's activestorage
- isolation_level as thread, since bank is using puma, which is not fiber based
- Set smtp timeout default to 5. There was no timeout before
- [TEST ONLY] Enable executor_around_test_case to make tests more realistic, test suite still passes
- Delete commented out cookie serializer, since we already use `json` explicitly
- Remove deprecated `ActiveSupport::TimeWithZone.name`, which isn't being called anywhere in bank (maybe by a gem, but I doubt it)
- Remove legacy media=screen for stylesheet_link_tag, this shouldn't affect bank because there aren't any stylesheets that are wholly intended for one type of media or another (there are some rules that use the @media, but those should still work as before)


maxwofford (approved):
Just went through them and this all makes sense to me. I haven't tested the preview app, but I see the tests are passing. LGTM!

PR #3484, opened 2023-02-01
Enable raise_on_open_redirects and set allow_other_host on external redirects

albertchae:

There is a security issue with the old way rails handled external
redirects https://github.com/rails/rails/issues/39643, mainly when the
url was user provided.

For example in https://github.com/hackclub/bank/blob/051928cbedeb7532facf2e47b5029f763f006265/app/controllers/hcb_codes_controller.rb#L61,
the `redirect_url` param is technically a (hidden) form field, so it's
possible a malicious user could manipulate this.

I went through every redirect in the codebase and added
`allow_other_host: true` where it seemed we were intentionally allowing
external redirects


app/controllers/events_controller.rb
```diff
@@ -357,7 +357,7 @@ def toggle_hidden
         "https://cloud-is6jebpbb.vercel.app/disappearing_doge_meme.mp4"
       ].sample
 
-      redirect_to file_redirects
+      redirect_to file_redirects, allow_other_host: true
```
albertchae: This was the easiest place to test how this argument worked 😆

garyhtou:
Should we also remove:
https://github.com/hackclub/bank/blob/5a77d06642f2f560a34a7cd6ba4b105d50d9276e/app/controllers/users_controller.rb#L129-L130
https://github.com/hackclub/bank/blob/5a77d06642f2f560a34a7cd6ba4b105d50d9276e/app/controllers/users_controller.rb#L163-L164

albertchae:
@garyhtou 🤔 hmm, from my understanding, as long as we do not add `allow_other_host: true` to these, these redirects will only work with `return_to` urls that are from bank

oh wait, I just tried https://bank.hackclub.com/users/auth?return_to=https%3A%2F%2Fwww.google.com and I see it doesn't work because of the `params[:return_to].start_with?(root_url)` check, so were you asking whether we should remove the check because of the new behavior? Hmm, I like that the existing code is explicit about the check and doesn't require future bank devs to be familiar with the `raise_on_open_redirects` default, so my vote is to keep it as is

maxwofford (approved):
I like this new default– a bit more explicit but less things I need to think about as an attack vector.

I'm cool with skipping changes to https://github.com/hackclub/bank/pull/3484#issuecomment-1411535123 for the scope of this PR, but if we do want to look at changes my 2 cents are something that is explicit about intent, rather than explicit about logic. ex `allow_other_host: false`

albertchae:
@maxwofford I'm not sure I understood your comment, are you saying that `allow_other_host: false` is explicit about intent or about logic?

garyhtou:
> @garyhtou 🤔 hmm, from my understanding, as long as we do not add `allow_other_host: true` to these, these redirects will only work with `return_to` urls that are from bank
> 
> oh wait, I just tried [bank.hackclub.com/users/auth?return_to=https%3A%2F%2Fwww.google.com](https://bank.hackclub.com/users/auth?return_to=https%3A%2F%2Fwww.google.com) and I see it doesn't work because of the `params[:return_to].start_with?(root_url)` check, so were you asking whether we should remove the check because of the new behavior? Hmm, I like that the existing code is explicit about the check and doesn't require future bank devs to be familiar with the `raise_on_open_redirects` default, so my vote is to keep it as is

I was thinking we could remove those checks because `raise_on_open_redirects` would protect us from redirecting to a non-Bank domain. But something I didn't consider is that if `return_to` is a non-Bank URL (ex. google.com), currently, Bank would redirect to `root_path` rather than raising an error. If we were to remove the check, it would raise an error.

Personally, I'm in favor of redirecting to `root_path` in the case of a non-Bank URL (instead of erroring). So yeah, I'm also in favor of keeping those checks.

Alternatively, for the same functionality, we could remove the checks and add:
```ruby
# https://blog.saeloun.com/2022/02/08/rails-7-raise-unsafe-redirect-error.html#after
class ApplicationController < ActionController::Base
  rescue_from ActionController::Redirecting::UnsafeRedirectError do
    redirect_to root_url
  end
end
```
This would have the added benefit of "soft failing" by redirecting to `root_url` rather than throwing a 500 error for ALL `redirect_to`

albertchae:
@garyhtou nice find on the fallback! I was waffling a bit on picking between the explicit check or the fallback or even keeping both, but I decided to go with the fallback only because in the context of `UsersController`, having a `return_to` param that is an external site would be abnormal/malicious behavior and so I think relying on a parent controller fallback (which will also handle other places this happens) will be fine.

I did add an airbrake notification though and tested it https://heroku-app97991095.airbrake.io/projects/288439/groups/3473647596703806985?tab=overview

garyhtou (approved):
Perfect! Thank you 😄

PR #3485, opened 2023-02-01
Disable browsers' flawed legacy XSS protection

albertchae:

Rails PR explains more https://github.com/rails/rails/pull/41769, but
these headers were non standard and had several flaws before they were
deprecated.

This one is a separate PR because I could see it possibly causing issues.


maxwofford (approved):
Makes sense, this header doesn't look well supported:
Screenshot 2023-02-01 at 10 23 22

Just in case, when this is merged could you post an accompanying post in #bank just giving a heads up that we've upgraded a security header and people should reach out if they have issues. Specifically, it seems like Safari is still deprecating usage & may still have some users.

albertchae:
Posted to #bank https://hackclub.slack.com/archives/CN523HLKW/p1676422353732989

PR #3495, opened 2023-02-07
Fix deletion requests for deleted events

cjdenio:


garyhtou:
Would it make sense to cascade an Event deletion down to OrganizerPositionDeleteRequest? (`dependent: :destroy`)

PR #3498, opened 2023-02-08
Fix settings tabs for admins

cjdenio:
related to #3299 

also swapped out `history.pushState` for turbo's built-in `data-turbo-action="advance"`


maxwofford:
Thanks for fixing this!

FYI, I'm running into state issues now that I didn't encounter before:
Screenshot 2023-02-08 at 14 13 16

cjdenio:
Huh, that happens to me in production but not development. Thanks for reporting!

cjdenio:
OH @maxwofford it's because this PR was rolled back on Heroku. This change actually fixes the issue in the screenshot

PR #3514, opened 2023-02-13
Paginate team positions

maxwofford:
Currently the team page isn't paginated:
Screenshot 2023-02-13 at 12 33 21
Now it is!
Screenshot 2023-02-13 at 12 35 44

It also drops our SQL query count from 51 (variable depending on team size) to 20 (independent of team size).

Creating this as a draft because this is the first time I'm directly using turbo & wanted to get a sanity check this is the right approach


app/controllers/events_controller.rb
```diff
@@ -72,8 +72,11 @@ def by_airtable_id
   def team
     authorize @event
 
+    max_positions = 5 # If over 9 members, start paginating
     all_positions = @event.organizer_positions.includes(:user).order(created_at: :desc)
-    @positions = Kaminari.paginate_array(all_positions).page(params[:page]).per(params[:per] || 3)
+    @all_positions_count = all_positions.size
+    @should_paginate = all_positions.size > max_positions
+    @positions = @should_paginate ? Kaminari.paginate_array(all_positions).page(params[:page]).per(params[:per] || 4) : all_positions
```
```suggestion
    @positions = @should_paginate ? all_positions.page(params[:page]).per(params[:per] || 4) : all_positions
```

From my understanding, since `all_positions` is already an active record relation, we can call `page` and `per` directly on the relation rather than using `paginate_array`. I think this would be more performant since it can use the database to perform `LIMIT` and `OFFSET`, rather than loading all the data into memory and doing it in Ruby

app/controllers/events_controller.rb
```diff
@@ -71,7 +71,12 @@ def by_airtable_id
 
   def team
     authorize @event
-    @positions = @event.organizer_positions.includes(:user).order(created_at: :desc)
+
+    max_positions = 5 # If over 9 members, start paginating
+    all_positions = @event.organizer_positions.includes(:user).order(created_at: :desc)
+    @all_positions_count = all_positions.size
+    @should_paginate = all_positions.size > max_positions
+    @positions = @should_paginate ? Kaminari.paginate_array(all_positions).page(params[:page]).per(params[:per] || 4) : all_positions
```
albertchae: I haven't done too much kaminari recently, but I thought the library handled the "should paginate" logic so you don't have to be counting how many items there are. Is this a consequence of using turbo?

app/controllers/events_controller.rb
```diff
@@ -71,7 +71,12 @@ def by_airtable_id
 
   def team
     authorize @event
-    @positions = @event.organizer_positions.includes(:user).order(created_at: :desc)
+
+    max_positions = 5 # If over 9 members, start paginating
```
albertchae: is the comment outdated?

garyhtou (approved):
works well!

PR #3518, opened 2023-02-15
Allow Bank Wrapped redirect

garyhtou:


albertchae:
woops, thanks for fixing

PR #3526, opened 2023-02-17
Don't Notify Users Of Declined Charges on Frozen Cards

sampoder:
I've probably done this wrong but this would be sick to have.... I get about 100 of these each week:

Screenshot 2023-02-17 at 9 39 31 PM

Once, I got 200 in a day.


app/services/stripe_authorization_service/create_from_webhook.rb
```diff
@@ -39,7 +39,8 @@ def run
           end
           # rubocop:enable Naming/VariableNumber
         else
-          CanonicalPendingTransactionMailer.with(canonical_pending_transaction_id: cpt.id).notify_declined.deliver_later
+          if cpt.raw_pending_stripe_transaction.stripe_transaction["request_history"][0]["reason"] != "card_inactive":
```
sampoder:
```suggestion
          if cpt.raw_pending_stripe_transaction.stripe_transaction["request_history"][0]["reason"] != "card_inactive"
```

app/services/stripe_authorization_service/create_from_webhook.rb
```diff
@@ -39,7 +39,8 @@ def run
           end
           # rubocop:enable Naming/VariableNumber
         else
-          CanonicalPendingTransactionMailer.with(canonical_pending_transaction_id: cpt.id).notify_declined.deliver_later
+          if cpt.raw_pending_stripe_transaction.stripe_transaction["request_history"][0]["reason"] != "card_inactive"
```
cjdenio: This includes cards that haven't been activated yet, which I think decline emails should still be sent for. You might change this to `unless cpt&.stripe_card&.frozen?` and update [`frozen?`](https://github.com/hackclub/bank/blob/8ec63112665d3ea59b4750b24f4fa4680d505648/app/models/stripe_card.rb#L178-L180) to be
```ruby
def frozen?
  activated? && stripe_status == "inactive"
end
```

sampoder: sure!

sampoder: done :D

app/services/stripe_authorization_service/create_from_webhook.rb
```diff
@@ -39,7 +39,8 @@ def run
           end
           # rubocop:enable Naming/VariableNumber
         else
-          CanonicalPendingTransactionMailer.with(canonical_pending_transaction_id: cpt.id).notify_declined.deliver_later
+          unless cpt&.stripe_card&.frozen?
+            CanonicalPendingTransactionMailer.with(canonical_pending_transaction_id: cpt.id).notify_declined.deliver_later
```
sampoder:
```suggestion
            CanonicalPendingTransactionMailer.with(canonical_pending_transaction_id: cpt.id).notify_declined.deliver_later
          end
```

cjdenio (approved):
Thanks!

PR #3534, opened 2023-02-19
Properly sort transactions within date

cjdenio:
array indexes in postgres start at 1


garyhtou:
OH MY GOSH REALLY?! whyyyyy postgres 😭😭

garyhtou:
ty for catching this bug!

PR #3545, opened 2023-02-22
Refactor exchange_login_code_for_user_spec to scope the stubbing closer to the relevant test

albertchae:
https://github.com/hackclub/bank/issues/3442

This is a small structural change

PR #3544, opened 2023-02-22
LoginCode model and migration

albertchae:
https://github.com/hackclub/bank/issues/3442

**Best reviewed by commits**

- Copy login code model and migration verbatim from hackclub/api

   model: https://github.com/hackclub/api/blob/86ae4362373cf05234a8e266ead33ad76547ecb6/app/models/login_code.rb
   migration: https://github.com/hackclub/api/blob/86ae4362373cf05234a8e266ead33ad76547ecb6/db/migrate/20191219235853_create_login_codes.rb

- Tests and improvements to login_code

  - added active? helper method
  - test that generate_code calls SecureRandom multiple times if a code exists
  - test that DB guarantees unique codes can't be saved (currently failing
    because there is no constraint)

- Add partial unique index migration to ensure active unique codes can't be saved

PR #3546, opened 2023-02-22
Implement LoginCode request and exchange logic in bank

albertchae:
https://github.com/hackclub/bank/issues/3442

**Best reviewed by commits**

- Copy mailer from hackclub/api

  mailer: https://github.com/hackclub/api/blob/86ae4362373cf05234a8e266ead33ad76547ecb6/app/mailers/applicant_mailer.rb#L4-L13
  view: https://github.com/hackclub/api/blob/86ae4362373cf05234a8e266ead33ad76547ecb6/app/views/applicant_mailer/login_code.html.erb

  Slightly refactored from the original to include only the fields needed
  to send the email

- Add association from user to login_codes

- Copy login code generation and mailer logic from hackclub/api

  original: https://github.com/hackclub/api/blob/86ae4362373cf05234a8e266ead33ad76547ecb6/app/controllers/v1/users_controller.rb#L18-L39

  I moved the code into a service for use in bank. This also required
  adding `ip_address` and `user_agent` as explicit parameters, since `request`
  isn't available in the service

  Partly to test for login_codes but also to surface issues with saving
  invalid IPs to postgres inet columns, I added a validation on
  `ip_address` using `Resolv` from the ruby stdlib. It's possible this is
  overly strict and we may need to allow invalid ip addresses to still
  request and generate login codes.

- Add feature flag (and SMS check) to either call hackclub/api or service in bank for login code

- Reimplement logic for exchanging login code and make it conditional on a feature flag

  original: https://github.com/hackclub/api/blob/86ae4362373cf05234a8e266ead33ad76547ecb6/app/controllers/v1/users_controller.rb#L41-L61

  I chose not to make a new service for the new logic because
  `ExchangeLoginCodeForUser` seems like a good place for that logic to
  live once the hackclub/api path is deleted. However, I kept the existing
  interface for `exchange_login_code_resp` now to allow turning on/off the
  feature flag easily, but once the feature flag is off we should try to
  refactor this.

- Cleanup logic from hackclub/api

  `login_code.save` is redundant because Rails saves nested associations
  automatically (and in the same transaction) if they are created this way
  (i.e. `user.login_codes.new`). As a related consequence, it's impossible
  for the `!login_code.valid?` check to be reached because an invalid
  `login_code` would have prevented `user` from saving properly and thus
  hit the first condition (this is shown in the spec for this logic, since
  the error is on `login_codes` of `user`)


albertchae:
* main
  * **PR #3544** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3544" target="_blank">Graphite</a> 
    * **PR #3545** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3545" target="_blank">Graphite</a> 
      * **PR #3546** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3546" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/hackclub/bank/3546?utm_source=stack-comment).

app/services/login_code_service/request.rb
```diff
@@ -0,0 +1,36 @@
...
+  class Request
+    def initialize(email:, ip_address:, user_agent:, sms: false)
+      @email = email
+      @sms = sms
+      @ip_address = ip_address
+      @user_agent = user_agent
+    end
+
+    def run
+      user = User.find_or_initialize_by(email: @email.downcase)
```
albertchae: One caveat with this implementation in bank for emails being seen for the first time.

Previously if a user tried to login with an email:

1. a `user` with that email would NOT be created in `bank`
2. request is sent to `api`, a `user` is created in `api`
3. login code is emailed to user
4. user logs in with login code, if valid, a `user` is created in `bank`

New implementation:
1. a `user` with that email is created in `bank` as soon as login code is requested
2. login code is emailed to user
3. user logs in with login code, if valid, the `user` created in step 1 is returned

app/services/user_service/exchange_login_code_for_user.rb
```diff
@@ -11,7 +11,12 @@ def initialize(user_id:, login_code:, sms: false)
     def run
       raise ::Errors::InvalidLoginCode, error_message if exchange_login_code_resp[:errors].present? || exchange_login_code_resp[:error].present?
 
-      user = User.find_or_initialize_by(email: remote_email)
+      if Flipper.enabled?(:login_code_2023_02_21) && !@sms
+        user = exchange_login_code_resp[:user]
+      else
+        user = User.find_or_initialize_by(email: remote_email)
```
albertchae: creating a user is no longer needed in new implementation

app/services/login_code_service/request.rb
albertchae: While this might mean we end up with some junk users if people typo their emails, it simplifies the logic quite a bit IMO and is similar in philosophy to the approach we took with https://github.com/hackclub/bank/pull/2650, where we create a user on invite instead of waiting for their first login.

albertchae:
I'll fix the rubocop errors later before merging

app/services/user_service/exchange_login_code_for_user.rb
```diff
@@ -11,7 +11,14 @@ def initialize(user_id:, login_code:, sms: false)
...
-      user = User.find_or_initialize_by(email: remote_email)
+      # rubocop:disable Naming/VariableNumber
+      if Flipper.enabled?(:login_code_2023_02_21) && !@sms
+        user = exchange_login_code_resp[:user]
+      else
+        user = User.find_or_initialize_by(email: remote_email)
+      end
+      # rubocop:enable Naming/VariableNumber
+
       user.api_access_token = remote_access_token
```
albertchae: this will be `nil` when the feature flag is off and it seems that we can delete this column once we shut off `hackclub/api`, right?

maxwofford (approved):

albertchae:
When I tested with the review app, I noticed the emailing styling is different. Mainly the icon in the header and the footer, which seem more bank themed than hack club in general (because we are using bank's email header/footer by definition of moving into this repo). I assume this is a good thing but just wanted to call it out.

## Before
image

## After
image

PR #3548, opened 2023-02-22
change applicable "2022" references to "2023"

northeastprince:
Not sure if we even wanna keep these, since they're just redundant parts of copyright(?) statements that are only in 2 obscure locations, but I guess just a year update to start. Lemme know if there are any other locations!


cjdenio (approved):
Thanks, feel free to go ahead and merge!

garyhtou (approved):

PR #3552, opened 2023-02-24
 Implement SMS LoginCode request and exchange logic by calling Twilio directly in bank 

albertchae:
https://github.com/hackclub/bank/issues/3442

This PR does **NOT** include the logic to enroll/toggle sms auth (see https://github.com/hackclub/bank/blob/2fffd3ab2e8240580d5f6d654e01407f53bf8df2/app/controllers/users_controller.rb#L312-L346)

This is to keep the scope of this PR smaller and also it looks like SMS auth is not that widely used anyway (97 out of 4401 total users have `use_sms_auth = true`)

Also this portion of the code isn't as extensively tested as the email path, partly due to time constraints but also because unit testing is a bit less valuable to mock out when it's mostly checking for third party integrations (in this case calling Twilio properly)

**Best reviewed by commits**

- Copy TwilioVerificationService verbatim from hackclub/api
  original: https://github.com/hackclub/api/blob/e95c4341586d4da20fb74888d58fd771b72bf901/app/services/twilio_verification_service.rb

  The only changes are how we pick out the twilio credentials from Rails credentials.
- Implement request and exchange of login code via Twilio SMS
- Fix specs for sms cases to expect calling Twilio directly instead of hackclub/api
- Add credentials files with twilio creds copied from hackclub/api
  
  There is an existing set of twilio credentials in bank's credentials, but they do not match api's. It looks like we need to use credentials that match up with `TwilioVerificationService::VERIFY_SERVICE_ID`, so while we figure out why we have a separate set of twilio credentials, add new ones with `_sms_verify` suffix (e.g. `account_sid_sms_verify` and `auth_token_sms_verify`) that copy the hackclub/api values for use by `TwilioVerificationService`.

  Also add the same keys to `development.yml.enc` with dummy replaceme
values


maxwofford:
cc @cjdenio for context on the credentials and why they're different. I'm not sure where the keys that were already in bank come from

albertchae:
* main
  * **PR #3552** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3552" target="_blank">Graphite</a>  👈
    * **PR #3554** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3554" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/hackclub/bank/3552?utm_source=stack-comment).

albertchae:
@maxwofford I'll merge for now due to the 2/25 deadline for https://github.com/hackclub/api/issues/482 and we can resolve the credential issue in a follow up

albertchae:
Tried logging out and in via SMS a few times while tailing the `hackclub/api` logs with `heroku logs --tail -a api-hackclub`. The first time I saw `/v1/users/sms_auth` and `/v1/users/29868/sms_exchange_login_code` get hit, but the next 3 times I didn't. I think this might be due to some weirdness with heroku's preboot feature and I hit the old dyno the first time.

PR #3559, opened 2023-02-27
Fix settled ACH payment memo

cjdenio:
Screenshot 2023-02-27 at 9 51 25 AM


northeastprince:
what happened with this?

cjdenio:
We default to "virtual" memos for most transaction types (see `HcbCode#memo`). In this case, since there wasn't one set, it fell back on the transaction's "real" memo, which is just a random code (specifically an HcbCode `short_code`) used to map the transaction to the right organization

PR #3563, opened 2023-02-27
Add helpers to users model for setting admins

maxwofford:
These are handy helpers to promote and demote admin status from users in the heroku console when we onboard new devs & team members


app/models/user.rb
```diff
@@ -114,6 +114,16 @@ def admin_override_pretend?
     self.admin_at.present?
   end
 
+  def make_admin!
+    self.admin_at = Time.now
+    self.save!
```
cjdenio:
```suggestion
    update!(admin_at: Time.now)
```

PR #3564, opened 2023-02-27
Add Brakeman gem for static analysis

maxwofford:
Now you can spin up docker-compose and run `bundle exec brakeman` for some handy pointers!

In the future we can play with adding this as a CI check, but for now I'm adding it just to development to play with locally.


northeastprince (approved):
🔐🔍🥳

PR #3565, opened 2023-02-28
Add performance monitoring via airbrake

maxwofford:


northeastprince (approved):
Nice! Does this mean we can phase out New Relic? Hopefully Skylight in the future if it's redundant now.

maxwofford:
Let's check back in a couple days and see if anyone on the team is using new relic. If we haven't heard anything by then I think it's fine to take it out.

PR #3568, opened 2023-02-28
improve organization of Gemfile and update/reduce dependencies

maxwofford:


northeastprince:
@garyhtou any ideas on this?

garyhtou:
which gems were removed or updated?

northeastprince:
@garyhtou all of them 😆 - I did a little debugging in the console with the gem changes and it seems the cause of the 5 tests not passing has something to do with either the `PublicIdentifiable` concern and/or the `friendly_id` gem it uses.

As with this change, running the service doesn't throw an error:

image

northeastprince:
Yep, it was `friendly_id` overriding the `find(hashid)` method and looking for a slug instead.

.rubocop.yml
```diff
@@ -132,3 +132,6 @@ Layout/DotPosition:
 
 # Metrics/LineLength:
 #   Max: 80
+
+Bundler/OrderedGems:
+  Enabled: false
```
albertchae: I didn't realize this was enabled before. Any particular reason for disabling it?

albertchae (commented):
Thanks for taking this on! I've definitely been confused about where to add gems in the Gemfile and haven't been disciplined about it, so this should be a big improvement.

Can you describe the organization rules you applied so that other devs can try to adhere to them? Related to that, is there some way to use a linter to enforce those rules? (see my comment on the rubocop rule)

I would also separate this PR into one where you just update the Gemfile organization and `run bundle install`, and then a follow up PR to actually update the gem versions. This first PR would convince me that the previous version of the Gemfile and your updated one would produce the same Gemfile.lock (modulo any deleted gems and versions getting unpinned). In the spirit of [SB changes](https://medium.com/@kentbeck_7670/bs-changes-e574bc396aaa), the first PR would be an S change and the second would be a B change.

I tried this quickly on my own and it looks like we're getting rid of maildown, brakeman, libv8node, and mini_racer. Didn't we just add brakeman recently?

.rubocop.yml
northeastprince: It enforces alphabetical ordering within groups, which is useless and kinda weird if we're trying to make things more organized.

northeastprince:
Oh, sorry, it must've undone the changes from adding `brakeman` automatically since the diffs are confusing.

garyhtou (approved):
thanks!

PR #3579, opened 2023-03-02
Remove newrelic_rpm

maxwofford:
As described in https://github.com/hackclub/bank/pull/3535, we're stepping away from new relic for the time being.


garyhtou (commented):
Since we're on a free New Relic plan, I think it makes sense to leave the gem so that New Relic can continue to collect data. In the case that we decide to go back to New Relic, we'd have existing data we can refer to

maxwofford:
@garyhtou another intent of this is to remove the extra dependency so preview builds aren't failing as much. (https://github.com/hackclub/bank/issues/3410)

PR #3585, opened 2023-03-03
bugfix for first grant background bleed on safari

InternetRamen:


InternetRamen:
imageimage

InternetRamen:
It may be a temporary work around because some data cells don't render the background as shown in the screenshot

YodaLightsabr (approved):
I'm no longer getting the visual glitch on Safari, this looks good!

PR #3588, opened 2023-03-06
Use more direct db:schema:load for CI testing database

northeastprince:
There's no need to run all of the migrations one after another, so this will just speed up runs, if anything.


cjdenio (approved):

albertchae:
Thanks for this change!

PR #3589, opened 2023-03-06
Improve organization of Gemfile

northeastprince:
(doesn't do anything to `Gemfile.lock`)


maxwofford:
Thanks for submitting this @northeastprince! Does this new structure follow a specific pattern we can all follow? Asking just to make sure we document it– the file may be improved by this, but if the next 10-15 PRs that update our gemfile don't know how to add stuff in the same style as this it'll get pretty messy pretty quick.

northeastprince:
I really liked the structure of [HEY's Gemfile](https://gist.github.com/dhh/782fb925b57450da28c1e15656779556), albeit much lighter, so I tried to mimic that - sections and categories with comments on their own line, and gem descriptions on the same line. I also think the nomenclature of "drivers" to describe the high-level gems that provide an interface for things like database(s), jobs, etc. is great.

maxwofford:
@northeastprince is there any doc that we can link to that's more specific? I want to avoid bike shedding as much as possible or the mental load of having to choose which category to add something. In that linked gist it's already happening [in comments](https://gist.github.com/dhh/782fb925b57450da28c1e15656779556?permalink_comment_id=3366300#gistcomment-3366300)

re: alphabetical order, the intent of alphabetical order is to reduce merge conflicts. maybe we want to use the categories, but then within each group alphabetically order the group?

northeastprince:
I'm not sure - the categories are definitely not set in stone, as those are just my recommendations for groupings. Open to input though.

Regarding alphabetical ordering, that's a good point - I'll remove the Rubocop exemption and fix that.

northeastprince:
@maxwofford regarding the Rubocop rule, it seems that it actually enforces any gems not separated by a new line, not just environment groups.

northeastprince:
I confirmed that this still doesn't change the current `Gemfile.lock`.

maxwofford:
I'm biasing for no on this– I'd like for our Gemfile to be easy to add to without having to read a blog or detailed writeup. I didn't understand what `Drivers` were until I read through the whole comments section & I'd like to avoid extra context/overhead for our contributors. How about we close it for now?

northeastprince:
I'd at least like to clean up the list of gems there are all clumped at the bottom, as we shouldn't have environment-specific groups stuck in the middle. We don't have to have those comments, but I also think some sense of related gems going together would make it a lot less confusing.

maxwofford:
@northeastprince yeah, that'd def be an improvement. How about you break up that list at the bottom so the grouped gems (`:production`/`:development`, etc.) can be at the bottom. Try to break them up similar to how we're currently doing it (let's skip the groups that hey.com uses for now) and ideally comment them to give a simple overview like we have in our other gems (I'm noticing lots at the bottom aren't).

maxwofford:
Whoops! sorry to rebase your stuff– tabbed over to the wrong PR.

garyhtou:
Heads up, some gems have a strict ordering constraint. See https://github.com/MiniProfiler/rack-mini-profiler as an example

northeastprince:
@maxwofford is this better? I still put some extra spacing between different categories of gems - now we don't have to adhere to certain definitions, but it's still easier to mentally parse.

maxwofford:
@northeastprince I'm fine with this going in now that we're not touching the lockfile in the same PR. Could you resolve conflicts & rebase it in?

maxwofford (approved):

PR #3597, opened 2023-03-07
Upgrade Redis to v6

northeastprince:
Closes #3525 once upgraded.


garyhtou (commented):
Can we add a comment in the files to explain why we're disabling SSL verification (and mention Heroku in the comment. This way if/when we move off Heroku, it'll be easy to find Heroku-dependent configurations)

app.json
```diff
@@ -40,7 +40,18 @@
...
+      "options": {
+        "version": "11"
+      }
+    },
+    "heroku-redis": {
+      "options": {
+        "version": "6"
+      }
+    }
+  },
```
garyhtou: > Latest build failed
> app.json validation error: Expected array for key: addons but got: Hash

I think Heroku might be expecting something like this:
```suggestion
  "addons": [
    {
      "plan": "heroku-postgresql",
      "options": {
        "version": "11"
      }
    },
    {
      "plan": "heroku-redis",
      "options": {
        "version": "9"
      }
    }
  ]
```

garyhtou:
Also i saw that our dynos have a `REDIS_TLS_URL` env var added by heroku. not sure if that might solve the SSL termination issues that I'm reading about

northeastprince:
We're not really "disabling" SSL verification per se, it's more that there's not really a certificate we can verify against - it's still using SSL/TLS, but because we're on Heroku and the app and Redis server are in the same place, a MITM attack isn't really a concern.

northeastprince:
> Also i saw that our dynos have a `REDIS_TLS_URL` env var added by heroku.

I think it's the same since for higher plans they enforce TLS only - but I'll switch it to that just to be safe.

garyhtou:
> I think it's the same since for higher plans they enforce TLS only - but I'll switch it to that just to be safe.

Cool, if you do switch it. Have it fall back to `REDIS_URL` if `REDIS_TLS_URL` doesn't exist

cjdenio:
@northeastprince is this ready for merge?

northeastprince:
Not sure - @garyhtou?

garyhtou:
Staging deploy is failing
image

Please see this comment: https://github.com/hackclub/bank/pull/3597#discussion_r1128773400

garyhtou:
@northeastprince would you mind testing in staging?

here's an error that I got in staging
image

maybe we need to remove `ENV["REDIS_URL"] = ENV["REDIS_TLS_URL"] if ENV["REDIS_TLS_URL"].present?`

northeastprince:
Hmm, maybe it's something with the load order? If the two env vars are the same, do we need that?

northeastprince:
It doesn't seem like hiredis supports SSL at all, and the performance increase is negligible - maybe we can just remove it?

garyhtou:
Go ahead and do whatever you think is best (and share your thought process along the way).

from a little bit of research, it seems like upgrading to `redis-rb` version 5 and removing `hiredis` might do the trick
- https://github.com/redis/hiredis-rb/issues/58#issuecomment-1214152718

northeastprince:
.deploy atm2

northeastprince:
@garyhtou this is currently deployed on [ATM1](https://atm1.teleport.hcb.gg) with the latest Redis server version and seems to work fine. Can you test this on Heroku and then we can finally merge this?

garyhtou:
Would you mind running the following on ATM1's rails console?
```ruby
Sidekiq.redis(&:info)
```

On the Heroku staging deploy I'm getting:
```ruby
irb(main):001:0> Sidekiq.redis(&:info)
Traceback (most recent call last):
        2: from (irb):1
        1: from (irb):2:in `rescue in irb_binding'
Redis::CannotConnectError (SSL_connect returned=1 errno=0 state=error: certificate verify failed (self signed certificate in certificate chain))
```

The expected output is something like this:
```ruby
# Ran in production
irb(main):001:0> Sidekiq.redis(&:info)
=> {"redis_version"=>"4.0.14", "redis_git_sha1"=>"18572a92", "redis_git_dirty"=>"1", ... }
```

garyhtou:
I'm going to send you the staging deploy's REDIS_URL via Slack in case that helps you debug this

garyhtou:

albertchae:
Thanks for working on this! One thing I want to caution about is that while `heroku redis:upgrade` does seem to upgrade in place so we won't lose any queued sidekiq jobs, it looks like there's some risk of data loss if jobs get queued during the upgrade process https://devcenter.heroku.com/articles/heroku-redis-version-upgrade#upgrade-using-redis-upgrade. I think we should set a maintenance window so we can monitor the existing redis for a bit or drain the queue before doing the upgrade, or if the maintenance window is unacceptable, temporarily save sidekiq jobs we want to queue somewhere else before we move them to the upgraded redis instance.

garyhtou:
I just confirmed that Sidekiq's able to connect to Redis in staging. 

---

Thanks for looking into this @albertchae. I plan on following the [guide](https://devcenter.heroku.com/articles/heroku-redis-version-upgrade#upgrade-using-redis-upgrade) you linked:
1. [Prepare the upgrade](https://devcenter.heroku.com/articles/heroku-redis-version-upgrade#step-1-prepare-the-upgrade)
	`heroku redis:upgrade redis-rectangular-95728 --version 6 -a bank-hackclub`
	I'm expecting version `6.2.11` since that's the version Heroku chose for the staging deploy
2. [Wait for the Maintenance to Be Created](https://devcenter.heroku.com/articles/heroku-redis-version-upgrade#step-2-wait-for-the-maintenance-to-be-created)
	`heroku data:maintenances:info redis-rectangular-95728 -a bank-hackclub` - Make sure you have Data Maintenance CLI [install](https://devcenter.heroku.com/articles/data-maintenance-cli-commands#installation) first.
	We can also check the maintenance window [here](https://data.heroku.com/datastores/a014b5ee-ac89-43d6-ac2d-7a2dfa8a972f) and by running `heroku redis:info -a bank-hackclub`
3. Disable all Sidekiq cron at https://bank.hackclub.com/sidekiq/cron
4. Turn maintenance mode on
  `heroku maintenance:on -a bank-hackclub`
5. Check that Sidekiq queues are clear
	Rails console: `Sidekiq::Stats.new.queues`
6. Merge this PR (heroku will build and deploy during maintenance mode)
	> Enabling or disabling maintenance mode does not alter your app’s dyno formation. Dynos continue to accrue billing hours while in maintenance mode. Web dynos continue to run, but Heroku’s [routers](https://devcenter.heroku.com/articles/http-routing) block all incoming HTTP requests to them. Dynos of other types (such as worker dynos) also continue to run, and you can run [one-off dynos](https://devcenter.heroku.com/articles/one-off-dynos) as usual.
	https://devcenter.heroku.com/articles/maintenance-mode
7. Wait until the build finishes and deploys
	- Upgrading the client before the server because we can always rollback the client version. But we can't rollback the server version
9. [Perform Upgrade](https://devcenter.heroku.com/articles/heroku-redis-version-upgrade#step-4-perform-upgrade)
	`heroku data:maintenances:run redis-rectangular-95728 -a bank-hackclub`
10. Turn maintenance mode off
  `heroku maintenance:off -a bank-hackclub`
11. Enable all Sidekiq cron jobs
12. Test to make sure everything's working: Using a devhacks card, make a donation to the devhacks org
	- Verify that the charge succeed
	- Verify that charge shows on devhacks' ledger in realtime ([an async sidekiq job](https://github.com/hackclub/bank/blob/fee772d4a840f66627b59f89f7f2e04482e7c5e7/app/controllers/stripe_controller.rb#L50-L51) is responsible for this)

@albertchae let me know if you've got any feedback on this runbook process

Disabling cron jobs probably isn't necessary since it's okay if we lose those queued jobs

## Timing

April 12th, 12:30am PDT.
- A lot of jobs run at 6 and 7am UTC (PDT midnight)
- So we'll do this upgrade after the large cron jobs run

albertchae:
@garyhtou that plan looks solid. One thing is I'm not sure if you'll be able to access `/sidekiq` in step 5 after you turn on maintenance mode, so you might need to check the sidekiq queues via console access?

> Disabling cron jobs probably isn't necessary since it's okay if we lose those queued jobs

Agree with that, the main concern are jobs that get queued up as a result of a user action that could get lost.

garyhtou:
> @garyhtou that plan looks solid. One thing is I'm not sure if you'll be able to access `/sidekiq` in step 5 after you turn on maintenance mode, so you might need to check the sidekiq queues via console access?

Ah good catch. Yeah I'll use the console to do that then

PR #3598, opened 2023-03-08
Deploy to Heroku with Docker

garyhtou:
TODO:

- [ ] Heroku Buildpacks aren't supported by `container` stack. These configs would need to happen inside the dockerfile
		image
- [ ] Set up multi-stage build Dockerfile
	- https://docs.docker.com/develop/develop-images/dockerfile_best-practices/
	- https://evilmartians.com/chronicles/ruby-on-whales-docker-for-ruby-rails-development


cjdenio:
This looks like an experiment now that our slug size (https://github.com/hackclub/bank/issues/3410) is reduced (https://github.com/hackclub/bank/pull/3698). Closing for now as stale!

PR #3606, opened 2023-03-08
Block viewing of HcbCodes without transactions (return 404) 

garyhtou:
they're kinda ugly
image


cjdenio:
Does this break pages for unpaid invoices?

garyhtou:
ah yes it does. reverting

PR #3607, opened 2023-03-08
Revert "Block viewing of HcbCodes without transactions (return 404) "

garyhtou:
Reverts hackclub/bank#3606


northeastprince:
What happened?

garyhtou:
HcbCodes are supposed to contain at least a CanonicalTransaction or CanonicalPendingTransaction. however, there's a quirk where there are HcbCodes without any transactions being used to display unpaid Invoices (which shouldn't be the case). 404ing would prevent people from viewing unpaid invoices

PR #3611, opened 2023-03-08
Improve new user flow

malted:
@maxwofford


malted:
Closed because tests were playing up due to main checks.
Instead opened #3750 into temp branch
#3751 was temp branch into main

PR #3621, opened 2023-03-10
Add service to accept terms for stripe cardholders

maxwofford:


garyhtou:
Are we good to close out this PR because of https://github.com/hackclub/bank/issues/3599#issuecomment-1468872546?

northeastprince:
I think so, yeah.

PR #3622, opened 2023-03-10
Import transactions from Increase

cjdenio:
tested, believe it or not


app/services/transaction_engine/hashed_transaction_service/raw_increase_transaction/import.rb
```diff
@@ -0,0 +1,47 @@
...
+              raise e
+            end
+          end
+        end
+
+        private
+
+        def primary_hash(it)
+          attrs = {
+            unique_bank_identifier: "INCREASE",
```
garyhtou: Increase `account_id` from transaction endpoint should probably be used as the `unique_bank_identifier`. you can prefix it with "INCREASE". For example `INCREASE-account_in71c4amph0vgo2qllky`

app/services/transaction_engine/canonical_transaction_service/import/simple.rb
```diff
@@ -19,7 +19,7 @@ def run
         private
 
         def hashed_transactions_ready_for_processing
-          ::HashedTransaction.where("id not in (?)", duplicate_hashed_transaction_ids + previously_processed_hashed_transaction_ids)
+          ::HashedTransaction.where.not(id: duplicate_hashed_transaction_ids + previously_processed_hashed_transaction_ids)
```
cjdenio: Context: `where("id not in (?)", [])` generates `WHERE (id NOT IN (NULL))`, which doesn't work !!!

app/services/transaction_engine/hashed_transaction_service/raw_increase_transaction/import.rb
```diff
@@ -33,7 +33,7 @@ def run
 
         def primary_hash(it)
           attrs = {
-            unique_bank_identifier: "INCREASE",
+            unique_bank_identifier: "INCREASE-#{it.increase_account_id.upcase}",
```
garyhtou: If `upcase` is not necessary, let's remove it

```suggestion
            unique_bank_identifier: "INCREASE-#{it.increase_account_id}",
```

cjdenio: It is required, there's a check somewhere that forces bank identifiers to be uppercase

garyhtou: cool, thanks!

app/services/transaction_engine/hashed_transaction_service/raw_increase_transaction/import.rb
```diff
@@ -0,0 +1,47 @@
...
+    module RawIncreaseTransaction
+      class Import
+        include ::TransactionEngine::Shared
+
+        def initialize(start_date: nil)
+          @start_date = start_date || last_1_month
+        end
+
+        def run
+          ::RawIncreaseTransaction.where("date_posted >= ?", @start_date).find_each(batch_size: 100) do |pt|
```
garyhtou: super picky (i don't mind whether you change it or not): I think you mean `rit` or `it` here

garyhtou (approved):
logic looks good to me! great work here

PR #3627, opened 2023-03-10
Increase Stripe Issuing balance to $80k

maxwofford:
We currently float $40k in our Stripe Issuing Balance at any given time as a buffer. Given the instability caused by the SVB bank run and size of our Stripe monthly spending (we're at about $30k at the time of writing) I'm doubling the buffer to $80k. This is an arbitrary higher number that is double our previous.


maxwofford:
This should only be merged once Stripe is wired up to our FRB account https://github.com/hackclub/bank/issues/3624.

garyhtou:
I'm cool with 80k. Here are some Stripe Issuing stats that might help:

- Largest monthly volume: Jan 2022 with 163.9k
- Last month's volume (Feb 2023): 123.6k
- All-time volume: $1.77M
image

maxwofford:
Ooooh! That is helpful @garyhtou – thanks for posting.

> our Stripe monthly spending (we're at about $30k at the time of writing)

I'm re-reading my earlier comment and realizing the "monthly" wording is confusing. What I meant to say is that Stripe top-ups take about 3-5 days to arrive (I believe), and we're usually waiting on ~$30k in the 3-5 rolling average to arrive on the busiest times.

PR #3637, opened 2023-03-13
Setup MRSK for (branch) deployments

northeastprince:
This is a big step towards getting off Heroku, even if it only runs in development at the moment - however, to test out branch deploys, this has to be in `main` to listen to PR comment events.


northeastprince:
.deploy noop

garyhtou:
I like what you're working on here!

I do have a very tiny concern about merging this PR since we're inching ever closer to making staging deploys undeployable. This PR adds 1 MB, which is really nothing, but we are now sitting at 499 MB lol- meaning that future PRs have 1 MB to spare before failure.

It's important to try out these new deployment options, so **I'm okay with merging this** as long as these development deployments are not publicly accessible. The reason is that in development mode, anyone who signs in is automatically granted admin permissions. It's not a super high risk, but I'd prefer that we don't have that available on the public internet. If we put some sort of authentication on top of the deployment (even basic auth), I'd be okay with that

---

Speaking of MRSK, I'd like to see us prioritize the following two deployment-related tickets:
- https://github.com/hackclub/bank/issues/3410
Staging deploys are going to be very important as we're adapting to new banks under the hood
- https://github.com/hackclub/bank/issues/3609
Having a production docker image will allow us to deploy MRSK in production, and also deploy Heroku Review apps without a slug limit (see https://github.com/hackclub/bank/issues/3410#issuecomment-1452736923).

Putting together a production dockerfile or reducing slug size is going to be a _quick_ short-term solution

northeastprince:
Yeah, I was going to add some sort of authentication, but our current branch deploys aren't behind a wall, so I thought it was fine. Nevertheless, I put the [services](https://atm1.teleport.hcb.gg) behind [Teleport](https://goteleport.com) and firewalled the direct servers.

garyhtou:
perfect! lgtm

if you'd like, we can also use the bank.engineering domain

garyhtou (approved):

northeastprince:
.deploy noop

PR #3638, opened 2023-03-13
Add flipper flag to disable sending of checks

YodaLightsabr:
Disables outgoing checks as outlined in #3623


garyhtou (approved):
perfect!

garyhtou:
go ahead and merge it! i've already added the `outgoing_checks` feature in production (as fully enabled).

when we do need to disable Checks, anyone with production access an flip the switch here: https://bank.hackclub.com/flipper/features/outgoing_checks

PR #3645, opened 2023-03-14
Create manually_decline_check.rb

maxwofford:
This check is pending and locking up an org funds. It looks like our aasm_states don't support cancelling a check, so this manually voids it and declines pending transactions that are tying up their funds.


maxwofford:
Merging this after IRL review with @cjdenio

northeastprince:
This feels wrong to be circumventing the model and kind of doing this in just a job.

maxwofford:
@northeastprince completely agree in an ideal world this shouldn't be needed, but the current aasm implementation doesn't have a way to cancel checks. This is time sensitive because a hackathon is trying to pay for their venue and the pending check is taking up their budget. I'm putting it as a `OneTimeJob` instead of running this in `rails c`, but the alternative to this is running directly in `rails c`.

maxwofford:
@northeastprince i'd love to chat about adding check cancelling after creation– it seems like we lost it in the refactor to aasm but can still be implemented. [I see other instances](https://github.com/hackclub/bank/blob/main/app/models/check.rb#L229-L235) that I believe would be fixed by adding it. cc'ing @garyhtou to correct me– I don't have full context.

garyhtou:
declining check was lost when moving from V1 to V2 tx engine. Scott had the intention of implementing it, but just never got around to doing it since Checks are such a low priority

PR #3659, opened 2023-03-14
SMS auth doesn't work in development right now. This is a temp fix

garyhtou:
I'm not too up-to-date with our current status on migrating our login code system.

this is a temp fix for the following error in development.
image

I think we don't have valid Twilio keys in our development credentials


garyhtou:
@albertchae just wanted to get your thoughts here on if we plan to support SMS auth in development

garyhtou:
nevermind! turns out I was using the wrong master key. SMS auth works fine in development

PR #3666, opened 2023-03-15
Improve (deploy) command Action

northeastprince:
This PR improves naming to avoid confusion, as the workflow gets triggered on any new PR comment, optimizes run time by only checking out the repo if it's a deploy command, and configures a "warehouse" server to (hopefully) speed up Docker builds.


northeastprince:
.deploy web

garyhtou (approved):

PR #3667, opened 2023-03-15
Improve performance of stripe issuing authorization request by reducing query count

albertchae:
This is the result of investigating several failed stripe authorizations for a particular user due to timeout (e.g. https://dashboard.stripe.com/issuing/authorizations/iauth_1MhyokFSaumjmb9rNXxfOdyB). While this user did eventually get lucky once with the webhook completing in time, anyone in this user's organization could run into this problem again until we improve the performance here.

Tracing through the webhook code a bit and using `Benchmark.measure`, we get from
https://github.com/hackclub/bank/blob/000801f9083b97eb936c0047e28e4a91fbaa08c5/app/controllers/stripe_controller.rb#L34
https://github.com/hackclub/bank/blob/000801f9083b97eb936c0047e28e4a91fbaa08c5/app/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb#L41
https://github.com/hackclub/bank/blob/84da6e838230af413080b5d3564ddd26a178237a/app/models/event.rb#L488
https://github.com/hackclub/bank/blob/84da6e838230af413080b5d3564ddd26a178237a/app/models/event.rb#L519-L530
This last method is the largest component of this request's computation and it appears to be several N+1's while computing `fronted_amount` on each of the `canonical_pending_transactions`. 

Copying this query manually, we can see that the affected user's organization has a high number of pending transactions
```
Event.find(1567).canonical_pending_transactions.incoming.fronted.not_waived.not_declined.where(raw_pending_incoming_disbursement_transaction_id: nil).count
=> 114
```

For reference, hq's event
```
Event.find(183).canonical_pending_transactions.incoming.fronted.not_waived.not_declined.where(raw_pending_incoming_disbursement_transaction_id: nil).count
=> 48
```

I was curious about every organization so I also ran this
```
Event.all.map { |e| e.canonical_pending_transactions.incoming.fronted.not_waived.not_declined.where(raw_pending_incoming_disbursement_transaction_id: nil).count }.tally
```
The top 4 events have 48, 57, 62 and 114 pending transactions.

The individual commit messages detail the improvements and how many queries are reduced. At each commit, we can check that the computation is the same by running
```
Event.all.filter {|e| e.fast_fronted_fee_balance_v2_cents != e.fronted_fee_balance_v2_cents}
```


albertchae:
* main
  * **PR #3667** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3667" target="_blank">Graphite</a>  👈
    * **PR #3668** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3668" target="_blank">Graphite</a> 
      * **PR #3669** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3669" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/hackclub/bank/3667?utm_source=stack-comment).

garyhtou:
Very cool! Thanks 😄

PR #3668, opened 2023-03-15
Filter for canonical_transactions belonging to event in memory

albertchae:
This is more of an iffy improvement which is why I have split it out into a separate PR, but I still think is a desirable change. Reducing queries generally improves performance but not always if we end up pulling too much data into memory. In this case I think we are OK because we do loop through the objects later anyway. 

It's also slightly questionable to be using Arel since it's considered private API, although the `gt`/`lt` cases are
at least on the more common end of the spectrum. More context in https://discuss.rubyonrails.org/t/what-has-happened-to-arel/74383/

**Best reviewed by commits**
- First commit uses Arel to resolve an ambiguous column issue
- Second commit reduces the number of queries because we already have these
preloaded and don't have to issue a new `sum` query to the db per `HcbCode`. 

```
> Event.find(1567).fast_fronted_fee_balance_v2_cents
> QueryCount::Counter.reset_counter
=> 121
```


albertchae:
* main
  * **PR #3667** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3667" target="_blank">Graphite</a> 
    * **PR #3668** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3668" target="_blank">Graphite</a>  👈
      * **PR #3669** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3669" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/hackclub/bank/3668?utm_source=stack-comment).

maxwofford (approved):
LGTM!

garyhtou (approved):
Looks good!

albertchae:
Discussed with @maxwofford on slack that we wanted to wait and see what #3667 's impact on stripe webhook was. #3667 was deployed on March 16 at 10:22 AM PDT. AFAICT, we can only look at 6 hour ranges at a time on skylight, so scanning through a couple different 6 hour ranges to this deploy, the P95 response time was hovering around 900 ms to 1.4s

### Screenshot of the 6 hour range just before the deploy
image


After #3667, the P95 for various 6 hour ranges seems to be around 500-600 ms

### Screenshot of the 6 hour range just after the deploy
image

There probably isn't enough traffic/data to be completely sure, but looking at some of the slower instances on skylight, the N+1 on `canonical_transactions` seems to be a decent chunk of the response time, so I'm guessing this will still help.
Screenshot 2023-03-19 at 10 42 05 PM

albertchae:
Just found out you can can look at up to 24 hour ranges on skylight by zooming out on the timeline, but that doesn't change anything meaningful from my last post so I didn't update the screenshots.

PR #3673, opened 2023-03-16
Normalize Puma settings in development

northeastprince:
We've changed some Puma (server) defaults since `rails new bank`. This optimizes things in production, but causes unwanted behavior in development, such as having zombie processes that don't properly exit when the server is shut down (due to it running in cluster mode). This can cause cryptic IP bind errors if you don't know something is still running, and just seems like a good deviation to disable unless needed.


northeastprince:
.deploy atm1

northeastprince:
.deploy atm1

garyhtou (approved):

PR #3675, opened 2023-03-17
Revert "Use redis in production for rack mini profiler"

garyhtou:
Reverts hackclub/bank#3618
Solves https://github.com/hackclub/bank/issues/3674

This is draining our Redis memory!

image


garyhtou:
We may still need to look into cache eviction

PR #3680, opened 2023-03-17
Add break-word to sidebar so longer names wrap

InternetRamen:
image


maxwofford:
Hmmmm... mixed feelings about this because it wraps user content. If we mess with our own interface that's one thing, but I don't want to mess with the name the user chose. What does ellipse or hiding overflow look like?

maxwofford (approved):
Going to approve this despite my comment because I think this is an improvement over the current setup which kinda looks broken.

northeastprince:
Maybe we could extend the sidebar (up to a point) and/or lower the font size based on the length, and then just truncate it (... at the end) as a last resort? The current change is definitely the lesser of 2 evils but if there's a relatively straightforward way to implement either of these options that'd be way better.

InternetRamen:
Hmm, good points. I'm gonna keep working on this.

maxwofford:
@InternetRamen cool if we merge this? Doesn't mean you can't make the other improvements, but I'd love to get this issue fixed so navs don't break pages.

InternetRamen:
> @InternetRamen cool if we merge this? Doesn't mean you can't make the other improvements, but I'd love to get this issue fixed so navs don't break pages.

Yup, just saw this but. I'll go ahead and merge

PR #3682, opened 2023-03-17
Process disbursements on Increase

cjdenio:
_tested!_

This change is intentionally minimal. A few things that should (and will) be implemented in the future:
- Map transactions using the transfer ID from Increase's API (instead of an identifier in the memo)
- Instantly process disbursements instead of in a cron (also remove `pending` and `in_transit` AASM states)
  - Somewhat dependent on #3681

Closes #3625


maxwofford (approved):
Love how simple and small this change is! Looks good to go, but haven't tested locally.

cjdenio:
I know we're building out a proper Ruby client for Increase, but I'm gonna merge this now to get things working again (albeit in a quick-and-dirty way)

PR #3683, opened 2023-03-19
Bring Turbo into the light

northeastprince:
This removes the version constraint and brings turbo-rails from 0.8.3 to 1.4.0 (as of now), therefore updating Turbo and fixing a bunch of current bugs and issues with it (especially Drive).


cjdenio:
Thanks! We're using the NPM release on the frontend— could you also update the version in `package.json`?

cjdenio:
Also, the change in https://github.com/hotwired/turbo/pull/648 may break Bank's UI in spots, since we assume links in frames will trigger a full page reload by default. Let's make sure `data-turbo-frame="_top"` is set on all relevant links

northeastprince:
Isn't the whole point of the gem to provide the Turbo library? Why are we referring to it in a different way on the frontend?

cjdenio:
The NPM release plays a lot nicer with Webpack. It's also part of the [turbo-rails installation process](https://github.com/hotwired/turbo-rails/blob/main/lib/install/turbo_with_node.rb). The gem still provides some backend integration

northeastprince:
.deploy atm1

northeastprince:
.deploy atm1

.github/workflows/test-mrsk-without-full-gemset.yml
```diff
@@ -0,0 +1,28 @@
+name: Test MRSK deploy without all gems
```
maxwofford: Did you mean to include this in the PR?

Gemfile
```diff
@@ -28,7 +28,7 @@ gem "react-rails"
 # roll back to mini_racer 0.4.0: https://github.com/rubyjs/mini_racer/issues/218.
 gem "mini_racer", "~> 0.4.0", platforms: :ruby
 # Turbo makes navigating your web application faster. Read more: https://github.com/hotwired/turbo
-gem "turbo-rails", "~> 0.8.3"
+gem "turbo-rails"
```
maxwofford: This is a bit of a jump from our current version tagging– thoughts on this?

```suggestion
gem "turbo-rails", "~> 1.4"
```

maxwofford (dismissed):
Added some quick comments, but happy with this upgrade on the high-level! Thanks for this.

Gemfile
northeastprince: If that's the case shouldn't we optimistically lock every gem that doesn't have a version constraint?

maxwofford: I'm up for optimistically locking to any 1.X version here, but want to find something that's a change scoped to turbo (not changing all the gemfile format in a PR meant for turbo updates) & doesn't differ too much from our current setup.

maxwofford: > doesn't differ too much from our current setup

err.... to clarify, i'm ok with updating our current setup in a separate PR, but don't want to block this merge on waiting for that.

northeastprince:
.deploy atm1

northeastprince:
.deploy atm1

northeastprince:
.deploy atm1

maxwofford (approved):
Thanks for fixing this up! This is looking good to go.

northeastprince:
@cjdenio I got permission from @maxwofford to merge this, but I wasn't able to test every Frame that might be impacted because I didn't have the relevant data locally - let me know if there's any other instances that should be fixed!

PR #3691, opened 2023-03-21
Fix exchange_login_code_for_user_spec to use real logic instead of stubbed method

albertchae:


albertchae:
OK the test failed because the ids aren't the same... awkward...

maxwofford (approved):

albertchae:
@maxwofford FYI the ids didn't match because the test was stubbing the logic inside `ExchangeLoginCodeForUser` which was necessary before when it was relying on calling hackclub API, but now that it's not I just refactored the test to execute the real logic.

PR #3692, opened 2023-03-21
Change calls to "render" to play well with Turbo

northeastprince:
[Renders within the same action have to return 422 (Unprocessable Entity) in order for Turbo to process the change(s).](https://turbo.hotwired.dev/handbook/drive#redirecting-after-a-form-submission) This also implements `flash.now` where necessary to set the message for the current action.


northeastprince:
.deploy atm1

maxwofford (approved):
lgtm!

PR #3696, opened 2023-03-22
Refactor ExchangeLoginCodeForUser now that it's not calling hackclub api

albertchae:
- No need to save/reload user outside of development since we're no
  longer saving api token
- No need to wrap exchange_login_code_resp in a hash to match the
  previous return shape. Raise errors directly instead of returning
  error in a hash that then triggers a raise


app/services/user_service/exchange_login_code_for_user.rb
```diff
@@ -9,23 +9,24 @@ def initialize(user_id:, login_code:, sms: false)
...
+      user = exchange_login_code
 
-      user = exchange_login_code_resp[:user]
-
-      user.admin_at = Time.now if Rails.env.development? # Make all users admin in development mode
-      user.save!
-
-      user.reload
+      if Rails.env.development?
+        user.admin_at = Time.now
```
maxwofford: Looks like https://github.com/hackclub/bank/blob/99957ae52bf8859bf4e1e2d1ade612ff220c387e/db/seeds.rb#L10 is irrelevant (not caused by this change, but only noticing it now)

maxwofford (approved):
Love the simplicity!

app/services/user_service/exchange_login_code_for_user.rb
```diff
@@ -9,23 +9,24 @@ def initialize(user_id:, login_code:, sms: false)
...
+        user.save!
+        user.reload
```
northeastprince: No reason to reload - even before it's saved to the database, the attributes are still changed on the same model.

cjdenio:
@albertchae Cool to go ahead and merge this?

(this is @maxwofford on @cjdenio's account)

((@cjdenio says hi too))

app/services/user_service/exchange_login_code_for_user.rb
albertchae: good call, was just moving existing code without thinking

app/services/user_service/exchange_login_code_for_user.rb
albertchae: oh actually if we have that in the seeds file and we're preferring having users use the seed file instead of the production database, I'd rather remove this whole development only condition here. I'm guessing most devs who use the production database will already be admin or can easily do it via rails console if they need to upgrade someone. What do you think?

albertchae:
@maxwofford @cjdenio just replied to a few comments on this PR, once we address https://github.com/hackclub/bank/pull/3696#discussion_r1153608086 we can merge

app/services/user_service/exchange_login_code_for_user.rb
maxwofford: Yeah, let's remove for the time being & see how it affects dev experience!

albertchae: removed!

PR #3698, opened 2023-03-22
Drop mini_racer

northeastprince:
I don't this will affect anything, but from what I gather, it was never really needed in the first place?


northeastprince:
.deploy atm1

maxwofford (approved):
Let's go ahead and merge this in!

maxwofford:
I'm merging this with the understanding that there may be some issues with our react components that didn't show up when testing this locally. All our react components are non-critical to the app so we can revert this if there are UI bugs that start showing up.

albertchae:
hooray, goodbye mini 🏎️

PR #3701, opened 2023-03-23
Remove "Export Finances" feature

northeastprince:
This is an admin tool that's not being used and relies on multiple dependencies - removing this reduces build size and additional complexity.


northeastprince:
@maxwofford should I get rid of the underlying `Export` model too? We currently don't use it for anything else, even though we have other types of exports.

garyhtou:
Thanks for this!

> should I get rid of the underlying Export model too? We currently don't use it for anything else, even though we have other types of exports.

yes please! We do have 106 records in production, but it's honestly junk data. I've created an export of that data and saved it to google drive. you can drop the table if you'd like

maxwofford (approved):
Thanks for doing this @northeastprince! Good to merge

PR #3703, opened 2023-03-24
Add option to bypass authentication in development

northeastprince:
No more searching for login codes on every deploy!


northeastprince:
Not sure if there's a better way to implement this on the frontend side without having a separate email field.

northeastprince:
.deploy atm3

garyhtou (changes requested):
Since staging contains a copy of the production database it must have authentication! We like to keep staging as similar to production as possible

maxwofford:
I'd kinda like to still have authentication in development– all the recent login improvements have come from devs dealing with this and getting frustrated, which led to improvements to the signup flow. I'm worried if we alleviate that pain just for us we're going to leave that burden on regular users.

northeastprince:
Ok, no problem! Thought it'd be a convenient thing to have, but that's a good point.

maxwofford:
@northeastprince I can def see scenarios this would be nice though! thoughts on having an env var that allows bypass, then check for env var AND that we're running in dev?

PR #3704, opened 2023-03-24
Install Increase gem

garyhtou:
Refactor codebase to use [Increase gem](https://github.com/garyhtou/increase-ruby)

_tested!_


app/services/increase_service.rb
```diff
@@ -2,63 +2,21 @@
 
 class IncreaseService
   module AccountIds
-    def increase_environment
-      if Rails.env.production?
-        :production
-      else
-        :sandbox
-      end
+    def self.fs_main
```
cjdenio: These could maybe be constants (`FS_MAIN = Rails.application.credentials.dig(...)`)

config/initializers/increase.rb
```diff
@@ -0,0 +1,25 @@
...
+  # Your Increase API Key!
+  # Grab it from https://dashboard.increase.com/developers/api_keys
+  config.api_key = Rails.application.credentials.dig(:increase, increase_envrionment, :api_key)
+
+  # The base URL for Increase's API.
+  # You can use
+  # - :production (https://api.increase.com)
+  # - :sandbox (https://sandbox.increase.com)
+  # - or set an actual URL
+  config.base_url = increase_envrionment
```
cjdenio:
```suggestion
  increase_environment = Rails.env.production? ? :production : :sandbox

  # Your Increase API Key!
  # Grab it from https://dashboard.increase.com/developers/api_keys
  config.api_key = Rails.application.credentials.dig(:increase, increase_environment, :api_key)

  # The base URL for Increase's API.
  # You can use
  # - :production (https://api.increase.com)
  # - :sandbox (https://sandbox.increase.com)
  # - or set an actual URL
  config.base_url = increase_environment
```

cjdenio (approved):
This looks awesome, thanks! I just left a few comments

config/initializers/increase.rb
garyhtou: thanks for catching that

app/services/increase_service.rb
garyhtou: good idea!

PR #3706, opened 2023-03-24
Install (Rails') ActionText

northeastprince:
`rails action_text:install`

PR #3707, opened 2023-03-24
Implement product announcements

northeastprince:


northeastprince:
* main
  * **PR #3706** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3706" target="_blank">Graphite</a> 
    * **PR #3707** <a href="https://app.graphite.dev/github/pr/hackclub/bank/3707" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/hackclub/bank/3707?utm_source=stack-comment).

garyhtou:
hey @northeastprince! I'd love to have a chat about your visions and plans for this feature. feel free to share what you're thinking here or over slack

garyhtou:
I @northeastprince, just checking in on the progress of this. Should we close this as stale?

northeastprince:
Completely forgot about this, but will keep working on it (#3706 is a dependency for this one)

northeastprince:
Proposed domain model:

- any authoritative content from the Bank team is a `Bulletin`
- a `Bulletin` can be (through [STI](https://api.rubyonrails.org/classes/ActiveRecord/Inheritance.html)) an `Announcement` (really important stuff like planned downtime, ~~a financial collapse~~, etc.), an `Update` (could be like a changelog-type thing?), potentially a `Resource` (for branding guidelines and such), things like that which are displayed in different ways throughout Bank (like recent announcements displaying on a banner at the top with a link to the full content, changelogs more off to the side, etc.)
- I don't think they should be namespaced under `Bulletin` to avoid confusions with things like `Bulletin::Update`s,  but that also means we might not wanna have a root-level model called `Resource` which could mean basically anything - thoughts on this would be very appreciated
- I think using Single Table Inheritance appropriately simplifies things, but it also means we can't (or shouldn't, really) have additional fields on a `Bulletin` if only a `Resource` would use it, for instance, which shouldn't be a problem especially with the flexibility of Action Text + Trix

northeastprince:
@maxwofford I'm thinking of adding optional email blasts for `Announcement`s on publication so this would be great for putting out the word on our upcoming downtime.

PR #3723, opened 2023-03-29
Set rack profiler to only run in dev

maxwofford:
Closes https://github.com/hackclub/bank/issues/3617


Gemfile
```diff
@@ -154,6 +154,10 @@ group :development do
 
   # for running webpack-dev-server and rails server together via Procfile
   gem "foreman"
+
+  # for recording performance and sql queries on the page
+  gem "rack-mini-profiler", "~> 3.0", require: false
```
garyhtou: To use this in development, would `require: false` make us explicitly require the gem before use?

garyhtou (approved):
good to merge

northeastprince:
This broke Bank for admins because it was referencing a gem in production that didn't exist in production. Fixed and redone in #3815.

PR #3730, opened 2023-03-31
Validate auth code length

YodaLightsabr:
Fixes #3729 by ensuring the login code is the right length before validating with Twilio.


garyhtou (approved):
looks good! if this works in staging, go ahead and merge

PR #3731, opened 2023-03-31
Fix rendering error on /my/inbox

YodaLightsabr:
This is a draft/working branch for #3167

For now, I've found a "solution" that I don't quite understand.

Removing the icon has fixed the visual glitch for me.

image


YodaLightsabr:
Update: by only blurring when the tooltip is visible, the page seems to be back to normal.

image

garyhtou (approved):
really good work here! tough bug to identify and solve

PR #3744, opened 2023-04-05
Create Increase account numbers for orgs

cjdenio:
Closes #3632


cjdenio:
let's-a go!

(merging behind a feature flag)

app/services/transaction_grouping_engine/calculate/hcb_code.rb
```diff
@@ -27,6 +27,7 @@ def initialize(canonical_transaction_or_canonical_pending_transaction:)
       end
 
       def run
+        return unknown_hcb_code if @ct_or_cp.is_a?(CanonicalTransaction) && @ct_or_cp.raw_increase_transaction&.increase_route_type == "account_number" && IncreaseAccountNumber.exists?(increase_account_number_id: @ct_or_cp.raw_increase_transaction&.increase_route_id) # Don't attempt to group transactions posted to an org's account/routing number
```
garyhtou: Should these transactions have their own HcbCode? such as `010`?

cjdenio: I considered that, but decided not to since these transactions aren't linked to a Bank object (donation/invoice/disbursement). This feature is really just an auto-mapper for the bank account transaction importing we already do. I'd love to hear if you have thoughts though!

PR #3748, opened 2023-04-06
[TEST] Attempt to cache bundle between GH action tasks

maxwofford:
This is a change to our GitHub actions config, so I'm testing it by opening a PR. This will only be ready for review once [TEST] is removed from the PR description.


maxwofford:
Closing this as not needed/not working. It caches the build dependencies, but reading and writing all that to disk ends up slowing down the process so much it's not worth the change. Additionally, this is only helpful for the first commit on a new branch that introduces Gemfile.lock changes because our ruby action already has good cache support

Here's an example of a current large change that slows down our builds because multiple runners are bundle installing in parallel (and throwing warnings for multiple writes to cache at the same time):

Screenshot 2023-04-06 at 12 27 29

Here's the same branch on a second push:

Screenshot 2023-04-06 at 12 31 51

This PR staggered the build which saves on network i/o, but with the added reading and writing from cache it ends up taking even longer (I guess GH actions don't run on SSDs?):

Screenshot 2023-04-06 at 12 32 26

PR #3757, opened 2023-04-07
Show running balance when ?show_running_balance=true

albertchae:
https://github.com/hackclub/bank/issues/3464

It might be easiest to review by commits, because the first one just rearranges existing code with no behavior change and then the second commit adds the new functionality.

Since we already query for all the transactions when showing an organization (even though the UI is paginated), the general approach is sum up the amounts of every transaction that comes on the next page up to the last (i.e. every transaction before the oldest one on the current page). Then we add each transaction's amount per row cumulatively.

Currently we only show the running balance if we pass the query param `show_running_balance=true` and the currently logged in user is an admin. This triggers a huge N+1 because the `amount` method looks up `canonical_transactions` and `hcb_code` per `CanonicalTransactionGrouped`, but as the issue stated, I didn't do any performance improvements for this first pass. Because we are still using `CanonicalTransactionGrouped`, we will probably have to use a similar strategy to the preloader classes.

How it looks like for my local copy of `/hq`

image

### TODO
- [x] ~~Should we add pending transactions to the running balance too?~~ NO
- [ ] Do we want some kind of UI toggle for adding `show_running_balance=true` or is requiring the user to manually add it enough?


app/controllers/events_controller.rb
```diff
@@ -39,9 +39,30 @@ def show
       @hide_holiday_features = true
     end
 
-    @transactions = Kaminari.paginate_array(TransactionGroupingEngine::Transaction::All.new(event_id: @event.id, search: params[:q], tag_id: @tag&.id).run).page(params[:page]).per(params[:per] || 75)
+    @all_transactions = TransactionGroupingEngine::Transaction::All.new(event_id: @event.id, search: params[:q], tag_id: @tag&.id).run
```
albertchae: It turns out the existing query was fetching all transactions already, so no performance hit here even if the param `show_running_balance` isn't set. I want to fix that with proper database based pagination but that will probably be a follow up after we implement this page using `HcbCode`s instead of `CanonicalTransactionGrouped`s

maxwofford (approved):
Nice! Glad to see this isn't adding much weight to the page load either due to all the preloading.

> Should we add pending transactions to the running balance too?

Nope! I want to be super clear to the bookkeepers what is actually touching our balance vs what isn't, so let's keep that hidden.

> Do we want some kind of UI toggle for adding show_running_balance=true or is requiring the user to manually add it enough?

This is fine to merge for now, but let's add that. Maybe even a user setting? We have a spot for admin settings:
Screenshot 2023-04-07 at 23 22 36

garyhtou:

albertchae:
> Nice! Glad to see this isn't adding much weight to the page load either due to all the preloading.

@maxwofford that's not quite true, what I meant is there was no performance hit just for loading and creating all the `CanonicalTransactionGrouped` initially. There's a huge performance issue with loading all the associations needed to calculate the balance though. I have a PR I'll put up for that shortly.

> This is fine to merge for now, but let's add that. Maybe even a user setting? We have a spot for admin settings:

With an admin setting, this means with that set any admin would always see the running balance on an organization page right? And they'd have to turn that setting off globally? Just wondering if it makes sense for them to be able to toggle it while viewing an organization just for that session or if they'd want it always on. From our use case, I'm guessing only a small subset of the admins would need this functionality anyway, and they'd probably want to see it on every organization page?

albertchae:
Following up here, the performance PR is https://github.com/hackclub/bank/pull/3780 and discussed with @maxwofford in zoom that the user setting will be global for that user and we can remove the `show_running_balance` param

PR #3766, opened 2023-04-08
Fix remaining instances of demo -> playground mode

maxwofford:
Also standardize on "Playground Mode" as the capitalization used everywhere


garyhtou:

PR #3742, opened 2023-04-04
Remove selenium/webdriver related gems as well as the heroku buildpacks

albertchae:

PR #3768, opened 2023-04-08
Remove redundant `rubyzip` from Gemfile.lock

garyhtou:
idk why `rubyzip` is being removed from Gemfile.lock on `bundle install`. could I get some eyes on this?


northeastprince (approved):
It's not depended on after #3742, so this is fine to merge, but wack that it wasn't removed when bundler was run then.

garyhtou:
Thanks for looking into this @northeastprince

PR #3771, opened 2023-04-09
Allow for easily toggling mock data

northeastprince:
It seems that #3755 is affecting every org with mock data enabled, and as of now, the only way to toggle it is to completely reset the session.


northeastprince:
Ah, didn't realize that `?show_mock_data=false` will change it - this still seems like a better way to do it (imo) since we shouldn't rely on changing session data within an action to toggle a behavior that is now throwing errors.

garyhtou:
Looks like this has been handled by https://github.com/hackclub/bank/pull/3798!

PR #3772, opened 2023-04-09
Update schema comments for recurring_donations to match database

albertchae:
Investigated this as a result of https://github.com/hackclub/bank/pull/3741#discussion_r1158556962.
It looks like this column actually does have default = FALSE in the
database but the model annotations nor schema.rb were updated correctly,
so this PR is to update those.

Migration: https://github.com/hackclub/bank/blob/65343778752022c1d9f112ac0c388281af8f386b/db/migrate/20230209001142_add_migrated_from_legacy_stripe_account_to_recurring_donations.rb#L5

Prod schema

```
bank-hackclub::DATABASE=> \d recurring_donations;
                                                      Table "public.recurring_donations"
               Column                |              Type              | Collation | Nullable |                     D>
-------------------------------------+--------------------------------+-----------+----------+---------------------->
 id                                  | bigint                         |           | not null | nextval('recurring_do>
 email                               | text                           |           |          |
 name                                | text                           |           |          |
 event_id                            | bigint                         |           | not null |
 amount                              | integer                        |           |          |
 created_at                          | timestamp(6) without time zone |           | not null |
 updated_at                          | timestamp(6) without time zone |           | not null |
 stripe_customer_id                  | text                           |           |          |
 stripe_subscription_id              | text                           |           |          |
 stripe_payment_intent_id            | text                           |           |          |
 stripe_client_secret                | text                           |           |          |
 stripe_current_period_end           | timestamp(6) without time zone |           |          |
 stripe_status                       | text                           |           |          |
 url_hash                            | text                           |           |          |
 last4_ciphertext                    | text                           |           |          |
 canceled_at                         | timestamp(6) without time zone |           |          |
 migrated_from_legacy_stripe_account | boolean                        |           |          | false
```


albertchae:
Merging this without review since it's comments only and I want to test whether this works https://github.com/hackclub/bank/pull/3742#issuecomment-1501221252

PR #3775, opened 2023-04-10
Don't assume $20 donation(s) (as a placeholder)

northeastprince:
> [Is it possible to increase the money placeholder in Hack Club Bank? The $20.00 makes it seem like we are only requesting a little bit for our hackathon.](https://hackclub.slack.com/archives/CN523HLKW/p1681085694439149)

I thought about replacing it with `1.00` (the minimum amount), but that gave off the wrong vibe as well.

If anyone has a better idea feel free to share!


maxwofford:
This feels weird from the UI side based on the user's feedback. They're hoping for donations larger than $20 by setting donor expectation, but $0 removes donor expectation.

northeastprince:
Maybe `1.00+` to signify the minimum donation amount?

maxwofford:
@northeastprince the user's message says that `$20` feels too low. Would `$1.00+` do that?

northeastprince:
I was thinking it would just say "1 dollar (minimum) and up" instead of recommending $20

maxwofford:
@northeastprince i think the user is talking about [anchoring](https://www.pon.harvard.edu/daily/negotiation-skills-daily/what-is-anchoring-in-negotiation/) their donors at a higher number, not a misunderstanding of the minimum donation allowed.

Reading between the lines, they want donors to think "ah, a $100 donation form. I could donate $100, or go above and beyond to donate $150" instead of "a $1 donation minimum. I could donate $1, or go above and beyond to donate $5"

garyhtou:
Hey @northeastprince, it looks like https://github.com/hackclub/bank/pull/3776 fixed the issue for the user

PR #3776, opened 2023-04-10
Donation page: Amount placeholder influenced by median donation amount

garyhtou:
Here are examples of suggested donation amounts for some transparent orgs.
```ruby
{
  "Hack Club HQ"  => {
    :donation  => "$20.00",
    :recurring => "$20.00"
  },
  "Hackathon ATL" => {
    :donation  => "$100.00",
    :recurring => "$20.00"
  },
  "Aid2Day"       => {
    :donation  => "$20.00",
    :recurring => "$20.00"
  },
  "Reboot"        => {
    :donation  => "$100.00",
    :recurring => "$20.00"
  },
  "Mesto"         => {
    :donation  => "$100.00",
    :recurring => "$20.00"
  },
  "Verses"        => {
    :donation  => "$100.00",
    :recurring => "$20.00"
  }
}
```

---

This is my take on https://github.com/hackclub/bank/pull/3775


northeastprince:
I don't love this approach for a few reasons - mainly that we're relying on sorting every successful donation, but also that we're still suggesting a donation at all (within a placeholder which could also be confusing), and even within that it's still limited between 20 and 100, which might be completely wrong depending on the type or size of event/org. It just feels like a lot of unnecessary complexity for a small detail. I'm not for moving forward with this, but I do wanna know if other people see a real need for this.

maxwofford:
This is awesome @garyhtou! Curious to see the runtime for large orgs like HQ. I'm 100% for making this feel customized and am happy with this implementation. Another option I'd accept is an event setting that users can customize.

Where'd the $100 upper-bound come from? Feels kinda low for the larger orgs we have.

northeastprince:
An option in the donation page settings is a great idea.

maxwofford:
@garyhtou Want to merge this for now so things are more dynamic, then we can add in a donation page setting with a designer's help?

malted:
~~median~~ buy vulns in all major banking services and calculate proportion (e.g. 0.1%) of donator's account balance for the placeholder text
_That's_ personalised

garyhtou:
Sure go ahead and do whatever with this PR

the reason i chose the $100 upper limit is because personally, it's slightly off putting to see a donation page ask for > $100 if i dont already have some sort of connection with the org. however, that's just my personal experience

garyhtou:
performance isn't too bad. iirc in development it was around 12ms for HQ. It can be improved by computing the median in the database using the 50th percentile and caching it with rails

maxwofford (approved):
Let's merge it in!

PR #3778, opened 2023-04-11
Disable transaction duplicate detection

cjdenio:
This causes a bunch of problems (reported by Liv) and isn't particularly necessary now that we're using Increase.

closes #3715


northeastprince:
@cjdenio does this have something to do with the `HashedTransaction` model? If so, can we remove it? I still don't quite get what it does.

cjdenio:
Yeah, this essentially obsoletes the `HashedTransaction` model. We should totally remove it, just later since it involves migrating live data

cjdenio:
For future reference, these are the `HashedTransaction` IDs that Bank currently thinks are duplicates:
```
[221120, 223426, 223427, 223428, 223429, 223430, 223431, 224228, 224229, 224230, 224231, 224232, 224233, 224234, 224317, 224318, 224319, 224320, 224321, 224322, 224323, 224324, 224433, 224434, 224435, 224436, 224442, 224443, 224444, 224445, 224446, 224447, 224448, 224452, 224453, 224454, 224455, 224456, 224457, 224458, 224459, 230600]
```

PR #3792, opened 2023-04-12
Stop version controlling docker env file

maxwofford:
This lets developers set port or prod locally without having to worry about messing with other dev's configs.

⚠️ This will remove your local `.env.docker` file. If you have custom changes make sure to save them somewhere else before pulling.


garyhtou:
thank you! i had been thinking about making this change

PR #3797, opened 2023-04-12
Switch to VIPS

maxwofford:
This is an experiment so we can test with how much faster/efficient-er VIPS is.

It is currently working for everything but HEIC files.


albertchae (approved):
This looks good to me and seems to be working on heroku (staging with receipt pics). It turns out the issues I was having on local and staging before are because of this line https://github.com/hackclub/bank/blob/3e4afa04baba6c25a92bb3cc86212770b6aea69f/app/helpers/users_helper.rb#L21, which is relevant if you're using a dump of the prod DB in local or staging.

> It is currently working for everything but HEIC files.

Can you clarify what you meant by this @maxwofford ? It seemed to be working for me besides the small change I made in https://github.com/hackclub/bank/pull/3797/commits/b5b609003c7ffe2cf19a14fa0f263b90244bea97

maxwofford:
> Can you clarify what you meant by this

nvm on that– I think my docker setup was just caching the wrong build. It's working for HEIC files for me now.

PR #3798, opened 2023-04-12
Set mock data to be event specific

maxwofford:


cjdenio:
Rolling back because this breaks the Invoices page (bank.hackclub.com/hq/invoices), because that's InvoicesController and not EventsController

PR #3805, opened 2023-04-13
Add Receipt Bin modal on HcbCode page

YodaLightsabr:
Adds a modal to the HcbCode page to select a receipt from Receipt Bin
image

image


app/javascript/controllers/receipt_select_controller.js
northeastprince, line 1: Maybe we can make this a more generic controller so it can be used down the road for other things that need selections?

app/controllers/hcb_codes_controller.rb
```diff
@@ -172,4 +172,27 @@ def toggle_tag
...
+    @receipt = Receipt.find(params[:receipt_id])
+
+    authorize @hcb_code
+
+    @receipt.receiptable = @hcb_code
+    @receipt.save!
+
+    flash[:success] = "Receipt added!"
+    redirect_back fallback_location: @hcb_code
+  end
```
garyhtou: Since it's technically possible to link more than just a HcbCode to a receipt, what if we moved this into the Receipts Controller?

garyhtou (approved):
Amazing work!

I love the addition of the search bar. we can implement it with Postgres' full-text search once https://github.com/hackclub/bank/pull/3769 gets merged

app/javascript/controllers/receipt_select_controller.js
```diff
@@ -0,0 +1,30 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = [ "receipt", "select", "confirm" ]
+  static values = { selected: String }
+
+  connect() {
+    console.log("Hello, Stimulus!", this.element)
+  }
```
garyhtou: do we wanna keep this?

YodaLightsabr: Oops lol, I'll fix that

app/views/receipts/_form_v2.html.erb
```diff
@@ -24,12 +24,22 @@
...
                               "action"           => "change->form#submit"
                             } %>
-        <%= form.label :file, class: "btn" do %>
-          <%= inline_icon "share" %>
-          <span>Upload receipt</span>
-        <% end %>
+        <span class="flex items-center justify-center flex-row flex-wrap" style="margin: -8px;">
+          <%= form.label :file, class: "btn m1" do %>
+            <%= inline_icon "share" %>
+            <span>Select from computer</span>
```
cjdenio: @garyhtou / @yodalightsabr Why was this copy changed? Now it's somewhat unclear this is for uploading receipts (select _what_ from computer?)

YodaLightsabr: We changed it to avoid confusion, but I can see how this is still unclear.

image

YodaLightsabr: I modified the copy for all 3 slightly. If you have any ideas I'd love to hear them.

PR #3820, opened 2023-04-14
Fix reset on upload for transactions marked no or lost receipt

YodaLightsabr:
Fixes #3817


YodaLightsabr:
I think this line was lost in the refactor

northeastprince:
Why is this a separate bool being stored in the database? Just have it be lost if there's no receipts attached.

YodaLightsabr:
There's a difference between the states of "no receipts, one must be added" and "there _was_ no receipt given for the transaction, therefore one can't be added"

cjdenio (approved):

app/services/receipt_service/create.rb
```diff
@@ -10,6 +10,8 @@ def initialize(attachments:, uploader:, upload_method: nil, receiptable: nil)
     end
 
     def run!
+      @receiptable&.update_column(:marked_no_or_lost_receipt_at, nil)
```
northeastprince: @YodaLightsabr why not `update!(marked_no_or_lost_receipt_at: nil)`?

cjdenio:
```suggestion
      suppress(ActiveModel::MissingAttributeError) do
        @receiptable&.update_column(:marked_no_or_lost_receipt_at, nil)
      end
```

in case `receiptable` doesn't have a `marked_no_or_lost_receipt_at` column

PR #3827, opened 2023-04-18
Hide transfer buttons in playground mode

YodaLightsabr:
image

Fixes #3825


cjdenio (commented):
Could we just disable the buttons? I do think it's useful to show what transfer types are available

PR #3831, opened 2023-04-19
CHECK DEPOSITS

cjdenio:


db/migrate/20230418200827_add_user_to_check_deposits.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddUserToCheckDeposits < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+  def change
+    add_reference :check_deposits, :created_by, null: false, index: { algorithm: :concurrently }
```
northeastprince: @cjdenio maybe use `creator` instead?

northeastprince (commented):
@cjdenio what are your thoughts on moving this under a `Check`/`Checks` namespace to keep things more organized? e.g. `Check::Deposit` and `Checks::DepositsController`

PR #3833, opened 2023-04-21
🧹 Clean up receipt uploading views and methods

YodaLightsabr:
Combine receipt upload code into a single view/partial and remove uploading to anything that isn't an HcbCode


YodaLightsabr:
How receipt uploading currently works: (pardon my messy handwriting)

Hack Club Bank Receipts!

Architecture decisions:
 - What needs a receipt? Just HcbCodes? ACHs/Stripe Authorizations/other types of transactions?
 - Where should receipt upload/linking logic take place from the web?
   - The Receipts controller makes sense since the common denominator is receipts
 - What about marking a receiptable as having no/lost receipt?
   - This should happen in the HcbCode and other respective receiptable controllers

PR #3836, opened 2023-04-21
Fix poppler build

albertchae:
We need to use this fork of heroku-buildpack-apt
https://github.com/nickrivadeneira/heroku-buildpack-apt/tree/gobject-introspection-support

So that the typelib directory can be found via setting environment
variables correctly
https://github.com/nickrivadeneira/heroku-buildpack-apt/commit/5918ae9c8749ae738ebd4771678f525d65bedf39

In fact, we don't want to set GI_TYPELIB_PATH in the environment variables because it'll override
what we're getting from the buildpack

We need to run apt before hackclub's custom poppler-utils buildpack,
otherwise the older version of pdftoppm comes up in th PATH first

Unfortunately the fork of heroku-buildpack-apt is behind this commit on
mainline https://github.com/heroku/heroku-buildpack-apt/pull/24, so I
had to remove comments for now. I think it might be worth forking
hackclub's own heroku-buildpack-apt that applies the new commit on top
of heroku master


albertchae:
Oh and using our hackclub's fork of poppler-utils for that one error is still OK, because it only contains various binaries (https://ubuntu.pkgs.org/20.04/ubuntu-main-amd64/poppler-utils_0.86.1-0ubuntu1_amd64.deb.html)

PR #3844, opened 2023-04-25
Validate check addresses

cjdenio:


northeastprince:
@cjdenio does the existing data meet this criteria?

cjdenio:
All the existing data is valid (we only have a few checks so far)

PR #3846, opened 2023-04-26
Start to directly associate transaction source objects with `CanonicalTransaction`

cjdenio:
This PR takes steps towards directly associating `CanonicalTransaction`↔️`Raw***Transaction`. This just adds the migration + a hack to fill in the fields for new transactions.

The end goal is to remove `HashedTransaction` entirely, since people tend to find it a confusing (and currently unused) part of the codebase.


app/models/canonical_transaction.rb
```diff
@@ -76,6 +79,8 @@ class CanonicalTransaction < ApplicationRecord
   has_one :local_hcb_code, foreign_key: "hcb_code", primary_key: "hcb_code", class_name: "HcbCode"
   has_many :fees, through: :canonical_event_mapping
 
+  belongs_to :transaction_source, polymorphic: true, optional: true
```
northeastprince: Still don't have a good understanding of the data modeling around this, but this seems like kind of a hacky way to do this. Would [Rails' delegated type feature](https://github.com/rails/rails/pull/39341) work for this?

cjdenio: Delegated types are more useful when you're trying to represent several models as a continuous set of records (e.g. `AchTransfer`, `Disbursement`, and `Check` could all be a `Transfer`). I think a polymorphic association makes sense here since a `CanonicalTransaction` needs to belong to a source-specific object that could be one of multiple types (`RawStripeTransaction`, `RawIncreaseTransaction`, etc)

maxwofford (approved):

app/models/canonical_transaction.rb
northeastprince: Then couldn't `RawStripeTransaction`s and other types all be `Raw/CanonicalTransaction`s and/or `Transactible`?

PR #3848, opened 2023-04-27
Remove dotenv gem

maxwofford:
Is anyone currently using this in their local config? As far as I can tell we don't use it for any production or staging configs


northeastprince:
It'll still be there regardless of the explicit dependency cause [`MRSK` uses a `.env` file for credentials and such](https://github.com/mrsked/mrsk#using-env-file-to-load-required-environment-variables).

maxwofford:
Thanks for pointing that out @northeastprince! I'm totally fine with this being in our Gemfile.lock as a dependency, but remove it from the Gemfile to make our intent clear. When I saw this line in the Gemfile I assumed someone was running with [dotenv cloud](https://www.dotenv.org/docs) or something.

PR #3849, opened 2023-04-27
Hardwire events to use v2

maxwofford:
This is a step towards making the v2 transaction engine the primary financial logic instead of the deprecated stuff being `balance` and the new stuff being `balance_v2`


northeastprince:
@maxwofford is this still being worked on?

garyhtou:
We should also rip this out
https://github.com/hackclub/bank/blob/386995d7ffd23e3682973712efc53781a9198ef5/app/controllers/application_controller.rb#L72-L76

PR #3850, opened 2023-04-27
Run rubocop todo

maxwofford:
This runs rubocop to chip away at a couple of the todo list items.


app/services/pending_event_mapping_engine/settle/invoice.rb
```diff
@@ -59,15 +53,15 @@ def run
               next
             end
 
-            Airbrake.notify("matched more than 1 canonical transaction for canonical pending transaction #{cpt.id}") if cts.count > 1
-            ct = cts.first
 
             # 3. mark no longer pending
```
garyhtou: please move this comment down (and remove the duplicate comment above)

garyhtou:
During the triage meeting, we discussed that our current rubocop config contains rules that we don't agree with. The solution is to update those rule configs or use something like https://github.com/standardrb/standard

PR #3851, opened 2023-04-27
Rename Check -> LobCheck

maxwofford:
This is still very much in progress. Pushing so the work is backed up


northeastprince:
@maxwofford what about something like `OutboundCheck` instead to better reflect what it actually is and make it more service-agonistic/flexible?

maxwofford:
It might make sense to have an abstract class or some other OOP solution to the diff/similarities between this and IncreaseChecks in the future, but this rename is just to free up the "check" namespace before logic refactoring.

This model is tied to lob, and IncreaseCheck is tied to increase, so I'm going to go with LobCheck for the scope of this PR.

PR #3852, opened 2023-04-27
Add developer option to exclude pry

maxwofford:
This gets around the pry issues I have in development for the time being. It's a bit of a hack instead of diagnosing the issue my pry install has, but this also addresses the comments some have raised about wanting the vanilla irb experience.


maxwofford:
I did my best to not interfere with anyone currently using pry in dev, but want to flag in case it behaves differently. You shouldn't need to set any variables as it should default to using pry still.

cc @albertchae

PR #3862, opened 2023-05-01
🧹 Clean up receipt up receipt uploading views and methods

YodaLightsabr:
#3833


app/controllers/receipts_controller.rb
```diff
@@ -4,11 +4,11 @@ class ReceiptsController < ApplicationController
   skip_after_action :verify_authorized, only: :upload # do not force pundit
   skip_before_action :signed_in_user, only: :upload
   before_action :set_paper_trail_whodunnit, only: :upload
-  before_action :find_receiptable, only: [:upload, :mark_no_or_lost]
+  before_action :find_receiptable, only: [:old_upload, :upload, :link, :link_modal]
 
-  def upload
+  def old_upload
```
northeastprince: @YodaLightsabr is there a better way to do this? I don't like the naming, as it'll just leave more things to clean up in the future.

YodaLightsabr: Yeah you're right, this is a bit messy. This is only used by the emburse transactions and stripe authorizations, and I'm not sure if those pages are linked to from anywhere.

northeastprince: cc @cjdenio

YodaLightsabr: Gary made this a few days ago. We haven't had a receipt uploaded to either of those receiptable types for well over a year, and I couldn't find a link to their respective show pages anywhere in the UI.

[image](https://bank.hackclub.com/blazer/queries/335-receipts-receiptable_type-time-series)

app/controllers/receipts_controller.rb
```diff
@@ -40,15 +40,83 @@ def destroy
...
+      if params[:show_link]
+        flash[:success] = { text: "#{"Receipt".pluralize(params[:file].length)} added!", link: (hcb_code_path(@receiptable) if @receiptable.instance_of?(HcbCode)), link_text: "View" }
+      else
+        flash[:success] = "#{"Receipt".pluralize(params[:file].length)} added!"
+      end
+    end
+
+    return redirect_to params[:redirect_url] if params[:redirect_url]
+
+    redirect_back fallback_location: @receiptable.nil? ? my_inbox_path : (@receiptable.try(:url) || @receiptable)
```
garyhtou:
```suggestion
    redirect_back fallback_location: @receiptable&.try(:url) || @receiptable || my_inbox_path
```

app/controllers/receipts_controller.rb
```diff
@@ -40,15 +40,83 @@ def destroy
...
+
+  rescue => e
+    Airbrake.notify(e)
+
+    flash[:error] = e.message
+    return redirect_to params[:redirect_url] if params[:redirect_url]
+
+    redirect_back fallback_location: @receiptable.nil? ? my_inbox_path : (@receiptable.try(:url) || @receiptable)
```
garyhtou: Can be moved into an `ensure` to prevent duplicating code

app/controllers/receipts_controller.rb
```diff
@@ -40,15 +40,83 @@ def destroy
...
+  end
+
+
   private
 
   def receipt_params
-    params.require(:receipt).permit(:file, :uploader, :receiptable_type, :receiptable_id)
+    params.permit(:receipt, :file, :uploader, :receiptable_type, :receiptable_id)
   end
```
garyhtou: create a `receipt_link_params` and `receipt_upload_params` instead

app/controllers/receipts_controller.rb
```diff
@@ -40,15 +40,83 @@ def destroy
...
+    if @receiptable.nil?
+      raise "Receiptable not found"
+    end
+
+    authorize @receipt
+    authorize @receiptable, policy_class: ReceiptablePolicy
+
+    @receipt.update!(receiptable: @receiptable)
+
+    flash[:success] = "Receipt added!"
```
garyhtou: feature idea: undo link

app/policies/receipt_policy.rb
```diff
@@ -2,7 +2,12 @@
 
 class ReceiptPolicy < ApplicationPolicy
   def destroy?
-    user&.admin? || record&.receiptable&.event&.users&.include?(user)
+    user&.admin? || record&.receiptable&.event&.users&.include?(user) ||
+      (record&.receiptable.nil? && record&.user == user) # Checking if receiptable is nil prevents unauthorized deletion when user no longer has access to an org
+  end
+
+  def link?
+    user&.admin? || record&.user == user
```
garyhtou:
```suggestion
    record.receiptable.nil? && record.user == user
```

app/views/receipts/_upload.html.erb
```diff
@@ -1,7 +1,9 @@
+<%# Only used by deprecated receiptable types: emburse transaction and stripe authorization %>
```
garyhtou: we can remove this now

app/views/receipts/link_modal.html.erb
```diff
@@ -0,0 +1,36 @@
...
+
+  <footer class="modal__action-row flex justify-end">
+    <%= link_to my_inbox_path, target: "_blank", class: "btn btn--primary mr2" do %>
+      Manage receipts
+    <% end %>
+    <%= form.submit "Add to transaction", class: "btn bg-success", style: "transition: 0.2s all;", data: {
+          "receipt-select-target" => "confirm"
+        }, disabled: true %>
+
+    <%# TODO: Replace dynamically with the right receiptable, since not just transactions are supported %>
```
garyhtou:
```suggestion
```

app/controllers/receipts_controller.rb
garyhtou: actually `link_receipt_params` sounds better (verb first)

config/routes.rb
```diff
@@ -58,8 +58,22 @@
     get "cards", to: "static_pages#my_cards", as: :my_cards
     get "cards/shipping", to: "stripe_cards#shipping", as: :my_cards_shipping
   end
-  post "receipts/upload", to: "receipts#upload"
-  delete "receipts/destroy", to: "receipts#destroy"
+
+  # post "receipts/upload", to: "receipts#old_upload"
+  # delete "receipts/destroy", to: "receipts#destroy"
+
+  resources :receipts, only: [] do
```
northeastprince: if you're doing `only: []` you might as well just use `scope`

northeastprince (dismissed):

config/routes.rb
YodaLightsabr: I think that would work if I was just using `member`-type routes, but Rails didn't like it with `collection` routes

app/controllers/receipts_controller.rb
YodaLightsabr: I think we can call this out of scope for now, but it would definitely be a cool thing to add later.

garyhtou (approved):

PR #3867, opened 2023-05-04
Fix rspec test suite in both docker setups locally

albertchae:
- I believe the changes in `.env.docker.example` were preventing test
  database setup from happening correctly, since we need to make a db
  called `bank_test` and be running in `RAILS_ENV=test`
  - Because `.env.docker.example` will not clobber an existing
    `.env.docker`, you may need to copy these changes manually.
- `docker_dev_setup.sh` needs to create `bank_test` db also (this line was
  copied from `docker_setup.sh`


albertchae:
@maxwofford I believe these changes worked for me with either setup, can you verify on your machine?

maxwofford (approved):
Just ran it locally and this worked for me! Thanks for getting this fixed up– will def help the beginner experience.

PR #3868, opened 2023-05-04
Re-add wallets perk

recursiveforte:


maxwofford:
Love it! Thanks for submitting the PR. This will technically work, but I'd love this to feel like we're adding on a new "thing" we provide– could we remove the "beta" message and make it feel more legit?

garyhtou (approved):
Amazing! thanks for adding this @recursiveforte, and its great to see this available again!

garyhtou:
feel free to merge when ready

PR #3869, opened 2023-05-04
Add clarification for setup in README

recursiveforte:


garyhtou:
hi @recursiveforte, try running `docker_dev_setup.sh`! That should handle the development key and seeding the database. Let me know if you run into any issues with that script

PR #3874, opened 2023-05-05
Fix keyword arg positioning in preparation for Ruby 3

albertchae:
## TL;DR
I ran the spec suite with `RUBYOPT='-W:deprecated'` and fixed all the issues that seemed like true positives. I propose we set this environment variable in prod for a few days and then grep the logs to find additional cases to fix, since our spec suite is not comprehensive.

## Full details
Ruby 3 introduced new rules around keyword arguments.
- https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/
- https://github.com/ruby/ruby/blob/v3_0_0/NEWS.md
- https://www.fastruby.io/blog/ruby/upgrades/upgrade-ruby-from-2.7-to-3.0.html

Originally in Ruby 2.7, we would see deprecation warnings for these such as `warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call`

However, it turns out they got turned off in 2.7.2 because people complained they were too noisy
- https://www.ruby-lang.org/en/news/2020/10/02/ruby-2-7-2-released/
- https://bugs.ruby-lang.org/issues/17000
- https://discuss.rubyonrails.org/t/new-2-7-3-0-keyword-argument-pain-point/74980

To get the previous deprecation warning behavior, we have to run with `RUBYOPT='-W:deprecated'`. So I ran the spec suite with that environment variable and fixed all the issues (except for a few involving ActiveRecord::Enum which I'm still investigating whether they are false positives). Many of these changes were in spec only but some were in app code too.

Because the spec suite isn't comprehensive, I propose we turn on the environment variable in production as well and examine logs after. We could try to intelligently log deprecation warnings to airbrake in the same vein as https://gitlab.com/gitlab-org/gitlab/-/merge_requests/66815 and https://www.bugsnag.com/blog/managing-warnings-in-ruby, but I think these approaches might be overkill for bank.

P.S. I thought there might be a rubocop rule that could help us, but apparently it was disabled for too many false positives
- https://github.com/rubocop/rubocop/pull/5871
- https://github.com/rubocop/rubocop/issues/5887
- https://github.com/rubocop/rubocop/pull/5892

Gitlab wrote a custom rubocop to fix warnings in their codebase, which I also think is overkill for bank

https://about.gitlab.com/blog/2021/02/03/how-we-automatically-fixed-hundreds-of-ruby-2-7-deprecation-warnings/


maxwofford (approved):
I like the logging approach– I was running into this while playing with https://github.com/hackclub/bank/pull/3745 & found instances all over the codebase.

albertchae:
Set this in heroku so we'll get prod logging. Merging this PR for now

image

albertchae:
Oops, I merged into another branch instead of `main`, so I recreated this PR as https://github.com/hackclub/bank/pull/3877

PR #3875, opened 2023-05-05
Upgrade ruby to 3.0.6 and bundler to 2.4.12

albertchae:
## Update versions in Gemfile and Dockerfile

These changes result in a successful docker build. I had to fight a bit
of a chicken and egg problem because just upgrading the Dockerfile ruby
version would not build, so first I would use the 2.7.7 docker build,
manually installing ruby 3.0.6 in the docker container with
https://github.com/rbenv/rbenv#basic-git-checkout, then updating
`.ruby-version` and rerunning bundle. Then I had to update bundler due to
this error

```
Gem::LoadError: You have already activated set 1.0.1, but your Gemfile requires set 1.0.2. Since set is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports set as a default gem.
```

Fortunately after all that, these changes resulted in a buildable docker image

## Remove ruby-version from github actions so we pick up same version as .ruby-version

According to https://github.com/ruby/setup-ruby#supported-version-syntax
```
If the ruby-version input is not specified, .ruby-version is tried first, followed by .tool-versions
```

## Fix rubocop issue with redundant regex freeze

Regexes are frozen by default in ruby 3

https://bugs.ruby-lang.org/issues/8948

```
$ ruby -ve 'p(/regexp/.frozen?)'
ruby 3.0.6p216 (2023-03-30 revision 23a532679b) [aarch64-linux]
true
```


albertchae:
I think https://github.com/hackclub/bank/pull/3907 was the last of the keyword argument fixes, but I plan on waiting 1-2 days to see if anything new pops up in the logs before deploying this PR.

maxwofford (approved):

albertchae:
The only instance of keyword arguments warnings I see now are `2023-06-06 21:05:59.742 [bank_hackclub] web.4 /app/vendor/bundle/ruby/2.7.0/gems/pg_search-2.3.6/lib/pg_search/model.rb:18: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call`, which is in a gem and I'm pretty sure is a false positive.

PR #3878, opened 2023-05-08
More Ruby 3 keyword args fixes

albertchae:
Continuation of https://github.com/hackclub/bank/pull/3877 after seeing what deprecation warnings we got on prod with `RUBYOPT='-W:deprecated'`

I generally favored converting the hashes into keyword args instead of double splatting, but in a few cases where it would make the formatting quite a bit different I opted to splat for simplicity. This was completely arbitrary and subjective so I can change them to all be consistent


northeastprince:
Yeah, it never really made sense to me why we had seperate `attrs` hashes for arguments.

maxwofford (approved):
I'd like to switch over time to removing the double-splats, but this is good to merge now. Feel free to merge in as-is because we're trying to get up to ruby 3

albertchae:
Yeah I'm concerned this is more all over the codebase than I realized and even logging in prod won't catch everything... the other concern I have is that every case in this PR is a codepath that wasn't covered by the spec suite, so we might have to do more of a manual code audit than I was hoping to.

albertchae:
Hmm actually the code coverage is better than I thought it'd be 
image

PR #3879, opened 2023-05-08
Use postgres 12 for review apps

albertchae:
It looks like heroku stopped supporting postgres 11 on review apps without any warning (https://github.com/hackclub/bank/issues/3876#issuecomment-1537070994).

Regardless, I think we should just bump this to 12 since
a) We were already running review apps on postgres 14 and 15 until just a month ago when they were set to 11 here https://github.com/hackclub/bank/pull/3597/files#diff-96677aa35debfefd031d9d34d9c70369754ee3acb2d9a9d4090e98612efee6f5

b) As part of https://github.com/hackclub/bank/issues/3302, we will probably start looking into upgrading to postgres 12 on prod very soon anyway, so testing it out in review apps is a good idea


garyhtou (approved):
makes sense, thank you Albert!

PR #3903, opened 2023-05-15
Switch some things over from `HashedTransaction` to `transaction_source`

cjdenio:


cjdenio:
@albertchae I'd love your thoughts on this— I'm working on switching from `Raw???Transaction -> HashedTransaction -> CanonicalTransaction` to a polymorphic `Raw???Transaction -> CanonicalTransaction` association

app/models/canonical_transaction.rb
```diff
@@ -298,7 +302,7 @@ def disbursement
   end
 
   def unique_bank_identifier
-    @unique_bank_identifier ||= hashed_transactions.first.unique_bank_identifier
+    @unique_bank_identifier ||= transaction_source.try(:unique_bank_identifier)
```
albertchae: is this a `try` because this column doesn't exist on `RawIncreaseTransaction`?

app/models/canonical_transaction.rb
```diff
@@ -117,27 +121,27 @@ def likely_card_transaction_refund?
...
+    transaction_source_type == "RawStripeTransaction"
   end
 
   def linked_object
     @linked_object ||= TransactionEngine::SyntaxSugarService::LinkedObject.new(canonical_transaction: self).run
   end
 
   def raw_plaid_transaction
-    hashed_transaction.raw_plaid_transaction
+    transaction_source if transaction_source_type == "RawPlaidTransaction"
```
albertchae: minor but when comparing to class names, I like to do `RawPlaidTransaction.name` because it guarantees the class exists and is loadable before getting the string name out of it. It's technically less efficient than a frozen string but I think is a bit more defensive against typos. I have heard arguments against this too though, so not a big deal either way.

app/models/emburse_card.rb
```diff
@@ -178,11 +178,8 @@ def sync_from_emburse!
   def hcb_codes
     @emburse_transaction_emburse_ids ||= emburse_transactions.pluck(:emburse_id)
     @raw_emburse_transaction_ids ||= RawEmburseTransaction.where(emburse_transaction_id: @emburse_transaction_emburse_ids).pluck(:id)
-    @hashed_transaction_ids = HashedTransaction.where(raw_emburse_transaction_id: @raw_emburse_transaction_ids).pluck(:id)
 
-
-    @canonical_transaction_ids ||= CanonicalHashedMapping.where(hashed_transaction_id: @hashed_transaction_ids).pluck(:canonical_transaction_id)
-    @canonical_transactions ||= CanonicalTransaction.where(id: @canonical_transaction_ids)
+    @canonical_transactions ||= CanonicalTransaction.where(transaction_source_type: "RawEmburseTransaction", transaction_source_id: @raw_emburse_transaction_ids)
```
albertchae: I think you can reuse the scope you added in this PR, so this can be `CanonicalTransaction.emburse_transaction.where(transaction_source_id: @raw_emburse_transaction_ids)`

app/models/raw_increase_transaction.rb
```diff
@@ -44,4 +44,8 @@ def category
     increase_transaction.dig("source", "category")
   end
 
+  def unique_bank_identifier
+    "INCREASE-#{increase_account_id.upcase}"
```
albertchae: oh do you need the `try` from before if you add this?

app/models/stripe_card.rb
```diff
@@ -327,18 +327,6 @@ def canonical_transaction_hcb_codes
...
-  end
-
-  def hashed_transaction_ids
-    @hashed_transaction_ids ||= HashedTransaction.where(raw_stripe_transaction_id: raw_stripe_transaction_ids).pluck(:id)
-  end
-
-  def raw_stripe_transaction_ids
-    @raw_stripe_transaction_ids ||= RawStripeTransaction.where("stripe_transaction->'card'->>'id' = ?", stripe_id).pluck(:id)
-  end
-
```
albertchae: 👏

albertchae (commented):
First of all, I forgot to comment on https://github.com/hackclub/bank/pull/3846 but at a high level I like this change and I'm excited you're working on it. For these types of migrations I like to follow the "expand and contract" approach (described here https://martinfowler.com/bliki/ParallelChange.html and also a similar one here https://stripe.com/blog/online-migrations).

I personally didn't like the `HashedTransaction` model because it seemed like unnecessary hops through join tables (like `canonical_hashed_mappings`) but also the `has_many` was confusing since it looks like we assumed one in most cases (as a lot of the code you deleted that had `hashed_transactions.first` shows). I also ran this on the database and verified it was empty.  `CanonicalTransaction.includes(:canonical_hashed_mappings).all.filter_map {|ct| [ct.id, ct.canonical_hashed_mappings.size] if ct.canonical_hashed_mappings.size > 1}`

It looks you already did most of the "expand" work in https://github.com/hackclub/bank/pull/3846 and this PR is about using the new association. 

- I was going to ask about a backfill job but looks like you covered that in https://github.com/hackclub/bank/pull/3866. 

- Don't `RawCsvTransaction` and `RawEmburseTransaction` need the reverse `has_one :canonical_transaction` association also? EDIT: I see you added it to `RawEmburseTransaction` in this PR, but still wondering about `RawCsvTransaction`

- Does it now make sense to change the transaction_source columns to not null since you ran the one time job? (This is fine for another PR).

- I'm not 100% about the business logic with the duplicate removal, but IIRC when I last looked at them they were either no-ops or very rudimentary checks. I trust you or hope someone else can verify it makes sense to remove them.

app/models/emburse_card.rb
cjdenio: Good call, thanks!

app/models/canonical_transaction.rb
cjdenio: The `try` isn't technically necessary, but since `unique_bank_identifier` is only used by the transaction hashing system, I wanted to play it safe in case future transaction sources don't include it

cjdenio:
Thanks for reviewing!

Re: not-null— yeah, at this point we could safely add a constraint in another PR. All of the CanonicalTransactions in production have a transaction_source set.

Re: duplicate removal— it was originally created to fix issues where Plaid would send us duplicates, though it isn't really necessary now that we're mostly using Increase (and the duplicate removal feature mostly caused false positives anyway)

PR #3907, opened 2023-05-16
Manual sweep through remaining Ruby 3 keyword argument changes

albertchae:
Continuation of https://github.com/hackclub/bank/pull/3897

We aren't getting too many hits via logged warnings anymore (which might
be a sign some of this is dead code), but to be extra careful I went
through all instances of methods with keyword parameters being called
and converted them from hash to keyword arguments.

This is the script I used to identify all methods with keyword arguments in the bank codebase
https://gist.github.com/albertchae/01b144bc8ea017a08dfd9c11b2570129

I also double checked by looking everywhere "attrs" was being used, and
it looks like after this change the only instances are for ActiveRecord
create and update.

Risks with this change:
- I used a macro to do the transformation, but it's possible the change
  is buggy if the starting text did not match what the macro was looking for.
- I searched for constant/method references manually which could be
  error prone if there is metaprogramming (which I think bank has little
  to none of) or because of scoping, since a class under a module
  hierarchy doesn't need to be fully scoped depending on where it's
  being called from.
- As much as I tried to tool my process, I was still manually looking at
  ripgrep output so I could have made a mistake.


maxwofford (approved):
I manually read through these and they all look good.

PR #3909, opened 2023-05-16
Update uri to address CVE

albertchae:
Resolves https://github.com/hackclub/bank/security/dependabot/95

This will be removed very soon once bank is on ruby 3 anyway, but it's
an easy fix to ship now.


albertchae:
Well I was wrong about the easy fix since upgrading to `0.10.0.1` triggers the same error in CI that made us put this in the Gemfile explicitly anyway... since we're going to have Ruby 3 ready in a week, I'm going to punt on this after all

PR #3913, opened 2023-05-17
Remove hacks unneeded after Ruby 3 upgrade

albertchae:
- Remove default gems that are no longer needed with ruby 3 upgrade (see https://github.com/hackclub/bank/pull/3448 for context)
- Remove default suffix: false to silence ruby 3 keyword argument warnings (see https://github.com/hackclub/bank/pull/3898 for context)


northeastprince:
When I looked at this PR based on the title I expected to find something along the lines of "remove Ruby 3 preparations, don't need that version anyway 😤" 💀

albertchae:
@northeastprince oh sorry this was supposed to be a draft PR but I think hackclub's github account wasn't allowing drafts at the time. The title just came from the branch name which is `gh`'s default behavior IIRC

albertchae:
@maxwofford in accordance with the PR experiment we discussed on Tuesday June 13, I consider this PR low risk and you can merge after approving.

maxwofford:
Merging now while staying near my computer so I can monitor the deploy

PR #3923, opened 2023-05-19
Only show receipt upload for card transactions

cjdenio:


garyhtou (commented):
im cool with this. ~~can we move the logic into the partial?~~ (EDIT: actually nvm)

maxwofford (approved):
I'm cool with this as long as we're comfortable with `receipt_required?`'s definition.

wouldn't it be nice if `admin_tools do` took a conditional for this sort of situation?

PR #3931, opened 2023-05-24
Slightly refactor LoginCode model

northeastprince:
- remove redundant validations (including `ip_address` cause it literally comes from the request and we shouldn't be validating our own data, at least in this instance)
- cleaned up the code generation logic (removed duplicate retries cause login codes are checked against the user and there's no need to have a different one on the extremely small chance of two users having the same active code)


garyhtou:
@albertchae, would you mind taking a look at this?

cjdenio (approved):
This looks good to me.

PR #3932, opened 2023-05-25
Return 404 for private orgs

northeastprince:
Don't have much experience with Pundit, so let me know if there's a better way to do this.


maxwofford:
Thanks for the quick fix on this!

garyhtou (approved):
thank you!

PR #3933, opened 2023-05-26
Add ability to send grants through virtual HCB cards

cjdenio:
this is... a lot, i know.


northeastprince:
@cjdenio can you elaborate on what exactly this is? I think it's some sort of way to have only one card be able to be use a grant?

cjdenio:
It's a feature that lets you send money to somebody through an HCB card, most useful if you want to restrict/monitor how the funds are used. This is being developed specifically for [OnBoard](https://onboard.hackclub.com), but I plan on making it an actual feature for larger orgs that need it in the future.

cjdenio:
Merging because of an event launch. I know this touches a lot of things and I'm happy to talk through any of the changes in this PR later.

PR #3947, opened 2023-06-07
Fix attach receipts page when not logged in

YodaLightsabr:
Fixes https://hackclub.slack.com/archives/CN523HLKW/p1686166250102209


app/controllers/hcb_codes_controller.rb
```diff
@@ -99,6 +99,7 @@ def comment
   def attach_receipt
     @hcb_code = HcbCode.find(params[:id])
     @event = @hcb_code.event
+    @s = params[:s]
```
northeastprince: Maybe `token` or `secret` instead of just `s`?

app/views/receipts/_form_v3.html.erb
```diff
@@ -27,10 +27,11 @@
       <% end %>
       <div class="container container--md pt2">
 
-        <%= form.hidden_field :redirect_url, value: url_for(only_path: false) %>
+        <%= form.hidden_field :redirect_url, value: request.fullpath %>
```
northeastprince: Couldn't this potentially give a URL to redirect back to that's not ours?

YodaLightsabr: It shouldn't, unless the user modifies the hidden field

app/controllers/hcb_codes_controller.rb
YodaLightsabr: Fixed, thank you for the suggestion

app/views/receipts/_form_v3.html.erb
YodaLightsabr: I removed it for now; since it was passing in the page URL to begin with, it likely isn't necessary if it redirects back by default

YodaLightsabr: #3945 removes the `redirect_url` param by default unless specified in the partial parameters, so we should be able to leave it as-is.

PR #3953, opened 2023-06-08
Receipt Center soft launch

YodaLightsabr:
Soft launch for Receipt Center


app/controllers/receipts_controller.rb
```diff
@@ -49,23 +49,22 @@ def link
...
+    if @receiptable.instance_of?(HcbCode)
+      receipt_distances = @receipts.map do |receipt|
+        {
+          receipt: receipt,
+          distance: SuggestedPairing.find_by(receipt: receipt, hcb_code: @receiptable).distance
+        }
+      end.sort_by { |receipt| receipt[:distance] }
 
-    authorize @receiptable, policy_class: ReceiptablePolicy
+      @receipts = receipt_distances.map { |receipt| receipt[:receipt] }
```
garyhtou:
```suggestion
      @receipts = receipt_distances.pluck(:receipt)
```

app/controllers/receipts_controller.rb
```diff
@@ -49,23 +49,22 @@ def link
...
+    authorize @receiptable, policy_class: ReceiptablePolicy
 
+    @receipts = Receipt.where(user: current_user, receiptable: nil)
+    @suggested_receipt_ids = []
 
+    if @receiptable.instance_of?(HcbCode)
+      receipt_distances = @receipts.map do |receipt|
+        {
+          receipt: receipt,
+          distance: SuggestedPairing.find_by(receipt: receipt, hcb_code: @receiptable).distance
```
garyhtou:
```suggestion
          distance: SuggestedPairing.find_by(receipt: receipt, hcb_code: @receiptable)&.distance || Float::INFINITY
```
although it shouldn't happen, it might be possible for find_by to return `nil`. this sets the default value to infinity so that it's at the end of the list when sorted

app/services/receipt_service/create.rb
```diff
@@ -14,14 +14,25 @@ def run!
...
+
+          unless pairings.nil?
+            pairings.each do |pairing|
+              SuggestedPairing.create!(
+                receipt: receipt,
+                hcb_code: pairing[:hcb_code],
+                distance: pairing[:distance]
+              )
+            end
+          end
```
garyhtou:
```suggestion
          unless pairings.nil?
            pairs = pairings.map do |pairing|
              {
                receipt: receipt,
                hcb_code: pairing[:hcb_code],
                distance: pairing[:distance]
              }
            end
            SuggestedPairing.insert_all(pairs)
          end
```
I _think_ this should work (and be faster)

app/services/receipt_service/extract.rb
```diff
@@ -7,6 +7,14 @@ def initialize(receipt:)
     end
 
     def run!
+      if @receipt.textual_content.nil?
+        @receipt.extract_textual_content!
+
+        if @receipt.textual_content.nil?
+          return nil
+        end
+      end
```
garyhtou:
```suggestion
      if @receipt.textual_content.nil? && @receipt.extract_textual_content!.nil?
          return nil
      end
```
Since `extract_textual_content!` returns the textual content (or `nil`), we can consolidate the if statements

app/services/receipt_service/create.rb
```diff
@@ -14,13 +14,25 @@ def run!
         @receiptable&.update(marked_no_or_lost_receipt_at: nil)
       end
 
-      receipt_ids = []
       ActiveRecord::Base.transaction do
```
garyhtou: tbh, i'm not sure why the DB transaction is here. maybe we get rid of it

app/controllers/static_pages_controller.rb
```diff
@@ -104,7 +104,37 @@ def my_inbox
...
+      recent_receipt = @receipts.order(created_at: :desc).first      
+
+      if params[:uploaded_receipts]
+        receipts_by_id = @receipts.map { |receipt| [receipt.id, receipt] }.to_h
+        @pairings = params[:uploaded_receipts].map do |receipt_id|
+          receipt = receipts_by_id[receipt_id.first.to_i]
+          next unless receipt
+
+
+          pairing = receipt.suggested_pairings.order(distance: :asc).first
```
```suggestion
          pairing = receipt.suggested_pairings.order(distance: :asc).first
          next if pairing.nil?
```

app/controllers/static_pages_controller.rb
```diff
@@ -104,7 +104,37 @@ def my_inbox
...
+
+          {
+            hcb_code: pairing.hcb_code,
+            receipt: receipt
+          }
+        end
```
```suggestion
        end.compact
```
removes possible `nil` values from `next`.

`next` essentially "returns" early from the block (although it's different than a return). since we aren't giving `next` a value (such as `next "example"`), the default value for that `map` iteration is `nil`

app/services/receipt_service/create.rb
YodaLightsabr: I don't think it's doing any harm, + it's nice to be able to completely abort if something goes wrong

app/services/receipt_service/extract.rb
YodaLightsabr: Ohh, right

app/controllers/receipts_controller.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require "uri"
+require "timeout"
```
northeastprince: Why do these need to be `require`d?

app/models/hcb_code.rb
```diff
@@ -46,6 +49,10 @@ def url
     "/hcb/#{hashid}"
   end
 
+  def popover_url
```
northeastprince: `url` -> `path`?

app/models/suggested_pairing.rb
```diff
@@ -0,0 +1,20 @@
...
+#
+# Table name: suggested_pairings
+#
+#  distance    :float
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#  hcb_code_id :bigint           not null
+#  receipt_id  :bigint           not null
+#
+class SuggestedPairing < ApplicationRecord
```
northeastprince: Love a separated entity, but does this have to be an *ActiveRecord* model?

app/controllers/receipts_controller.rb
YodaLightsabr: I think you’re right, we don’t have to require these.

app/models/suggested_pairing.rb
YodaLightsabr: Thanks! Let’s keep `SuggestedPairing` as an ActiveRecord model for efficiency and consistency.

app/services/receipt_service/create.rb
YodaLightsabr: For posterity: we ended up getting rid of the DB transaction because of issues with ActiveRecord

app/assets/stylesheets/components/_shimmer.scss
```diff
@@ -0,0 +1,89 @@
...
+    0% {
+        background: rgba($red, 0.4);
+    }
+    50% {
+        background: rgba($red, 0.9);
+    }
+    100% {
+        background: rgba($red, 0.4);
+    }
+}
```
garyhtou:
```suggestion
}

```

app/controllers/receipts_controller.rb
```diff
@@ -83,14 +103,32 @@ def upload
   rescue => e
     notify_airbrake(e)
 
+    puts e.message
+    puts e.backtrace
+
```
garyhtou:
```suggestion
```

app/controllers/static_pages_controller.rb
```diff
@@ -104,7 +109,28 @@ def my_inbox
...
+      # if params[:pairing_receipt] && params[:pairing_hcb_code]
+      #   begin
+      #     @pairing = {
+      #       receipt: Receipt.find(params[:pairing_receipt]),
+      #       hcb_code: HcbCode.find(params[:pairing_hcb_code])
+      #     }
+      #   rescue ActiveRecord::RecordNotFound => e
+      #   end
+      # end
+
```
```suggestion
```

app/models/receipt.rb
```diff
@@ -98,6 +101,10 @@ def extract_textual_content!
     end
   end
 
+  def extract_features
+    textual_content
+  end
+
```
```suggestion
```

garyhtou (approved):

garyhtou:
good to merge for the soft launch. we'll come back and clean things up

PR #3956, opened 2023-06-08
Fix keyword arguments passed to pg_search_scope

albertchae:
This fixes this warning which I incorrectly thought was a false positive
`/app/vendor/bundle/ruby/2.7.0/gems/pg_search-2.3.6/lib/pg_search/model.rb:18: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call`

https://github.com/hackclub/bank/pull/3875#issuecomment-1581748897

It was manifesting as a "wrong number of arguments" error after ruby 3
upgrade, which at least gave me a better stack trace to debug it.
https://github.com/hackclub/bank/issues/3955

This was breaking transaction renaming, because we use this search scope to find similarly named transactions


albertchae:
Merging now since this is broken on prod

PR #3958, opened 2023-06-09
Receipt previews

YodaLightsabr:
Enables previewing receipts on the same page via a modal. This is a better user experience than opening the receipt in a new tab.


app/views/receipts/_receipt.html.erb
```diff
@@ -1,40 +1,22 @@
...
-    <% end %>
-  <% else %>
-    <span class="muted mb1">
-      <strong>
-        <%= receipt.file.blob.filename %>
-      </strong>
-      (<%= number_to_human_size receipt.file.blob.byte_size %>)
-    </span>
-  <% end %>
+<% instance = (0...16).map { ('a'..'z').to_a.concat(('A'..'Z').to_a).concat((0..9).to_a).sample }.join %>
```
garyhtou: there's likely a better way to get unique ids for modals. i can help u look into it this weekend

garyhtou (approved):

PR #3959, opened 2023-06-09
Use signed IDs to verify unauthenticated receipt attachments

northeastprince:
We don't need to reinvent the wheel on this - I generally think links that bypass auth should have an expiration date on them, but lemme know if a week is too short or if there's a better way to do this.


garyhtou (approved):
amazing! ty for this Matt!

PR looks good. just one change: we'll want to support both signature systems for the next month or so. This way the old links we recent sent out (via email) still work

as for expiration, i think at least 2 weeks is good

YodaLightsabr (approved):
Looks good. Both methods are still working, so we can come back in a few weeks and remove the hmac check.

garyhtou:
Hey @northeastprince, just as a suggestion: when making edits to your PRs, please add additional commits rather than amending your existing ones. It makes reviewing it much quicker since I'll have context of what's new and what I've already seen.

PR #3962, opened 2023-06-12
disable wallets perk because omg people are addicted to these wallets

recursiveforte:


maxwofford:
@recursiveforte hey! in the future please wait on PRs getting reviewed. Know this is a small change, but we just want to really make sure everything is getting a check before going live.

recursiveforte:
Sorry, my bad! Will do in the future.

PR #3965, opened 2023-06-13
Fix title bug

YodaLightsabr:
Fix https://heroku-app97991095.airbrake.io/projects/288439/groups/3559654844659365182


YodaLightsabr:
A bit more info: when rendering mock data, there is no `event` associated with each `HcbCode`. The error occurs when trying to get the `.name` of `nil`, which results in mock data being broken for demo accounts.

garyhtou (approved):

PR #3973, opened 2023-06-14
Sync transactions from Increase accounts besides FS Main

cjdenio:
Dynamically figures out which Increase accounts to sync from based on the org list.

Also adds a validation to prevent mapping transactions between Increase accounts— e.g. mapping a fiscally sponsored transaction to a donor-advised fund


app/services/transaction_engine/raw_increase_transaction_service/increase/import.rb
```diff
@@ -9,27 +9,34 @@ def initialize(start_date: 1.month.ago)
...
+          account_ids.each do |account_id|
+            params = {
+              account_id: account_id,
+              "created_at.on_or_after": @start_date.iso8601,
+              limit: :all,
+            }
+
+            ::Increase::Transactions.list(params) do |transactions|
+              # For each transaction in the page...
+              transactions.each do |transaction|
```
garyhtou: unrelated: I wonder if I should add a `list_each` method. it'd yield individual transactions rather than pages

PR #3976, opened 2023-06-14
Handle missing `email` param for login

northeastprince:
Didn't catch potential additional error coming with #3971.


northeastprince:
Feel free to merge upon approval.

garyhtou:
See https://github.com/hackclub/bank/commit/df437f92a9b06bdef5aa27ce4f9d1a4b787c4a39#r118089754

garyhtou:
Another thing I wanna mention too is that we can use strong params in the controller to handle this

northeastprince:
Yeah, but that only protects against mass assignment, not omitting a required attribute.

garyhtou:
> Yeah, but that only protects against mass assignment, not omitting a required attribute.

With strong params, you can use [`require`](https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-require). It'll raise an `ActionController::ParameterMissing` if the parameter was omitted.

Example:
image

garyhtou:
image

PR #3987, opened 2023-06-16
Refactor feature preview into partial

YodaLightsabr:


app/views/users/_feature_preview.html.erb
```diff
@@ -0,0 +1,25 @@
+<div class="card pb0 mb3" id="<%= local_assigns[:id] %>">
```
garyhtou: maybe include logic to also handle private previews

YodaLightsabr: oh good idea

YodaLightsabr: thoughts on keeping the feature previews as a constant in the users controller?

garyhtou: what do you mean?

app/controllers/users_controller.rb
```diff
@@ -368,7 +368,7 @@ def wrapped
   private
 
   def set_shown_private_feature_previews
-    @shown_private_feature_previews = params[:classified_top_secret]&.split(",") || []
+    @hidden_feature_previews = ["receipt-bin"].reject { |feature| feature.in?(params[:classified_top_secret]&.split(",") || []) }
```
garyhtou: For private feature previews, I prefer the system that we previously had. With the current setup, to make a preview private, we'd need to add its ID to this array (and make sure it's the right ID). In other words, there would be two files to change (the ERB and this array)

Here's what i'd suggested:
To make a preview hidden, pass `private: true` to the partial (it should default to `false`). Then, in the partial, if `private == true`, it will check to see if the `id` is in `@shown_private_feature_previews`

garyhtou (changes requested):

app/controllers/users_controller.rb
YodaLightsabr: I agree, that's cleaner.

garyhtou: in case you're curious, the `@hidden_feature_previews` design leads to the [shotgun surgery code smell](https://refactoring.guru/smells/shotgun-surgery) (where you need to make small changes to multiple files to accomplish a task)

app/views/users/_feature_preview.html.erb
```diff
@@ -0,0 +1,27 @@
+<% unless defined?(private) && defined?(@shown_private_feature_previews) && !@shown_private_feature_previews&.include?(local_assigns[:id]) && !Flipper.enabled?(local_assigns[:feature_flag], current_user) %>
```
garyhtou: To make things easier to read, can we invert this condition? (I inverted it using demorgan's law)

```suggestion
<% if !defined?(private) || !defined?(@shown_private_feature_previews) || @shown_private_feature_previews&.include?(local_assigns[:id]) || Flipper.enabled?(local_assigns[:feature_flag], current_user) %>
```

also, I'm just realizing now, that `private` might not be the best var name to use since it's also a keyword. maybe we call it `unlisted` or `invite_only` 🤷‍♂️

garyhtou (approved):
Perfect!

One thing to note is that you might wanna update the Receipt Bin feature preview link in your slack post, since the URL query's changed from `?classified_top_secret=receipt_bin` to `?classified_top_secret=receipt-bin` (underscore -> hyphen)

YodaLightsabr:
> Perfect!
> 
> One thing to note is that you might wanna update the Receipt Bin feature preview link in your slack post, since the URL query's changed from `?classified_top_secret=receipt_bin` to `?classified_top_secret=receipt-bin` (underscore -> hyphen)

Yep! I updated the message in #bank-internals

PR #3988, opened 2023-06-16
Include avail balance in Stripe Auth metadata

garyhtou:


app/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb
```diff
@@ -38,17 +38,32 @@ def event
...
+
         true
       end
 
+      def set_metadata!(additional = {})
+        default_metadata = {
+          current_balance_available: card.balance_available
+        }
+
+        # Sending request in an async job to prevent blocking webhook response
```
cjdenio: I think we need to do this synchronously to avoid a possible race condition where we receive `issuing_authorization.created` (and attempt to send the declined email) before `declined_reason` has been set
https://github.com/hackclub/bank/blob/e578b7247ea233065d8513c7f1fc48c4457fa5b8/app/mailers/canonical_pending_transaction_mailer.rb#L31

garyhtou: thanks for catching that

garyhtou:
@cjdenio please review if you do get a chance

cjdenio (approved):
Tested and this looks good to me!

app/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb
cjdenio, line 1: Context: `Event#fronted_fee_balance_v2_cents` sometimes returns a float, which causes the balance to show a confusing trailing `.0` in Stripe's dashboard.
Screenshot 2023-06-22 at 3 48 50 PM

PR #3994, opened 2023-06-20
Convert most instances of ** a hash to keyword arguments

albertchae:
In the initial PR https://github.com/hackclub/bank/pull/3877
to fix this issue for Ruby 3, I opted for double splatting since it was
the smallest change. But in https://github.com/hackclub/bank/pull/3878
and after, since I had a macro for inlining a hash of attributes as
keyword arguments, I used that more to avoid unnecessary indirection.

Besides a few cases where the formatting was non trivial (e.g. the result
of a method call gets passed to a block or is used in a ||), this PR
changes most of the ** to keyword arguments instead for consistency.


albertchae:
@maxwofford should be safe to merge after review

app/services/transaction_engine/hashed_transaction_service/raw_increase_transaction/import.rb
```diff
@@ -32,13 +32,12 @@ def run
         private
 
         def primary_hash(it)
-          attrs = {
-            unique_bank_identifier: it.unique_bank_identifier,
+          ::TransactionEngine::HashedTransactionService::PrimaryHash.new(
+            unique_bank_identifier: "INCREASE-#{it.increase_account_id.upcase}",
```
maxwofford: Maybe I'm reading this wrong-- is there a reason we're switching from `it.unique_bank_identifier` to something hardcoded here?

maxwofford (commented):
Left a comment on one line that looks out of place, but everything else looks good to merge! Let me know if that was intentional.

app/services/transaction_engine/hashed_transaction_service/raw_increase_transaction/import.rb
albertchae: Oh oops, I had this branch open from a while ago and this got caught up in a rebase. It looks like https://github.com/hackclub/bank/commit/7e64b496e1f48ffcd28567a7a3a128a06df28e5b#diff-9bfb229e4b75a4711852fc2ec0e1481f88d5e28dfe0627ef8d2bb1114a19ab3b changed it to remove `INCREASE-#`, which I will do now

albertchae: changed it back to `it.unique_bank_identifier`

albertchae:
@maxwofford great catch, that was not intentional. Now the PR should be ready to go

maxwofford (approved):
Thanks for that, looks good now!

PR #3995, opened 2023-06-20
Update webauthn for openssl 3

albertchae:
Ran `bundle update webauthn`

I believe the `ruby:3.1.4` only has openssl 3 and greater, so we need to
make this change before moving to ruby 3.1


albertchae:
I tested I was able to login with my yubikey on the review app, so I think this is good to go.

albertchae:
@maxwofford should be safe to merge after review, although I'd check security key login on prod after deploy just to be safe

albertchae:
Per https://github.com/hackclub/bank/pull/3996#issuecomment-1601215792, I reset this branch to just be the webauthn changes. I'll reopen that branch after we merge this one

maxwofford (approved):

maxwofford:
I can confirm webauthn works in prod after merge!

PR #3996, opened 2023-06-20
Upgrade to Ruby 3.1.4

albertchae:
Upgrade to latest patch version of Ruby 3.1

This should be backwards compatible. With the improved debug gem and irb
changes, we should evaluate if we still need pry

https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/
https://www.ruby-lang.org/en/news/2022/02/18/ruby-3-1-1-released/

3.1.2, 3.1.3, and 3.1.4 are security fixes only

https://www.ruby-lang.org/en/news/2022/04/12/ruby-3-1-2-released/
https://www.ruby-lang.org/en/news/2022/11/24/ruby-3-1-3-released/
https://www.ruby-lang.org/en/news/2023/03/30/ruby-3-1-4-released/


northeastprince:
I'm fine with just IRB, but if people prefer pry we can keep it.

albertchae:
@maxwofford should be safe to merge after review

maxwofford (approved):
New syntax just dropped! Wahoo!

Thanks for adding this. I'll go ahead and merge and keep an eye as it deploys

maxwofford:
Also, re: pry. I'll defer to anyone using pry-- I've been sticking with the byebug env.

albertchae:
@maxwofford oh whoops I should have called it out before, but https://github.com/hackclub/bank/pull/3995 needed to be merged before https://github.com/hackclub/bank/pull/3996. I should be more careful when chaining PRs

EDIT: the good thing is that the ruby 3.1 stuff just went to the webauthn branch, so it won't affect production. I reset that branch to its original state and will reopen this PR after

PR #4000, opened 2023-06-20
Move stats endpoints to `StatsController`

cjdenio:
an attempt to de-clutter `StaticPagesController`!


northeastprince:
Could we use this as a starting point to separate out actions into different controllers to begin to move towards CRUD conventions? So stats#index, events/stats#index, etc., then you can use `resource :stats` in `routes.rb`

cjdenio:
For now I think it makes sense to keep all of the stats endpoints in one place, but it would be great to look into changing that in the future.

PR #4019, opened 2023-06-28
Use dotenv, update dev scripts accordingly

northeastprince:
This way, we don't have to set env vars before every command.


maxwofford (approved):
This looks good! thanks for fixing it-- it's def an improvement to the beginner contributor experience.

I'm reinstalling docker rn and haven't tested this locally, but please go for merge if a fresh clone can run the docker dev setup script and have everything work out of the box.

PR #4033, opened 2023-06-30
Add flavor texts

YodaLightsabr:


YodaLightsabr:
Approved IRL by @maxwofford

PR #4037, opened 2023-07-02
Attempt to fix #2644

maxwofford:
Stripe shouldn't be sending us duplicates in the first place, but this will handle them for the time being to prevent us from going over Airbrake's warning limit


app/jobs/stripe_authorization_job/deprecated/create_from_webhook.rb
```diff
@@ -6,7 +6,7 @@ class CreateFromWebhook < ApplicationJob
       def perform(stripe_transaction_id)
         # DEPRECATED: create stripe auth on v1 engine
         #
-        ::StripeAuthorization.create!(stripe_id: stripe_transaction_id)
+        ::StripeAuthorization.find_or_create_by!(stripe_id: stripe_transaction_id)
```
northeastprince: Do we wanna do `create_or_find_by!` instead since it's usually not gonna be there and to avoid any potential race conditions?

northeastprince:
If this is deprecated, how the code still in use?

app/jobs/stripe_authorization_job/deprecated/create_from_webhook.rb
maxwofford: That's good to know about `create_or_find_by`– I hadn't seen it before. I'm going to stick with `find_or_create_by` here because all the running webhooks are running with their own stripe id so I think the chance of race condition is safely low.

maxwofford:
@northeastprince the code was moved to a deprecated class so we could phase it out. It doesn't mean it's not being called though, so this quickfix is only to make sure we are still getting `StripeAuthorization`s created while we can check the code for instances of it being used. The goal here is still 100% to phase this out.

PR #4052, opened 2023-07-05
Show address errors from Stripe

northeastprince:
This will prevent things like unsupported countries causing 500s.


app/models/stripe_cardholder.rb
```diff
@@ -93,26 +93,28 @@ def remote_requirements_disabled_reason
...
+            state: address_state,
+            postal_code: address_postal_code,
+            country: address_country
+          }.compact_blank
+        }
+      }.compact_blank # Stripe doesn't like blank values
+    )
+  rescue Stripe::StripeError => error
+    if error.message.downcase.include?("address") || error.message.downcase.include?("country")
+      errors.add(:base, error.message)
```
cjdenio:
```suggestion
      Airbrake.notify(error)
      errors.add(:base, error.message)
```

this way we're aware of which countries are unsupported

cjdenio (approved):

PR #4053, opened 2023-07-05
Fix docker port to listen on 3000

YodaLightsabr:


northeastprince (approved):
Imagine running Rails in Docker, smh...

PR #4059, opened 2023-07-06
Allow for preferred names in additional to legal names

northeastprince:
Not quite sure where else we want to explicitly have the legal vs preferred name, so lemme know if there are any other places I should switch it!


app/views/organizer_position_invites/show.html.erb
```diff
@@ -9,7 +9,7 @@
 <% if @organizers.any? %>
   <div class="flex justify-center flex-wrap">
     <% @organizers.each do |position| %>
-      <div class="tooltipped tooltipped--s mr1" aria-label="<%= position.user.full_name %>">
+      <div class="tooltipped tooltipped--s mr1" aria-label="<%= position.user.display__name %>">
```
garyhtou: Is `display__name` defined?

db/schema.rb
```diff
@@ -279,11 +279,11 @@
     t.bigint "raw_pending_invoice_transaction_id"
     t.text "hcb_code"
     t.bigint "raw_pending_bank_fee_transaction_id"
-    t.bigint "raw_pending_partner_donation_transaction_id"
```
garyhtou: can we remove these unrelated changes?

app/models/user.rb
garyhtou, line 1: what are your thoughts on adding a validation to limit the length of the preferred name? it may help with preventing broken UI

garyhtou (changes requested):
great job! just a couple comments

app/models/user.rb
northeastprince: Sure! Not sure what's reasonable - 30 characters? 40?

db/schema.rb
northeastprince: Yeah, Rails reorganized that for some reason,

app/views/organizer_position_invites/show.html.erb
northeastprince: Ah, my bad 😆 - thought about naming the column `display_name` but it sounded too username-y.

app/views/organizer_position_invites_mailer/notify.html.erb
```diff
@@ -1,7 +1,7 @@
 <p>Hi there! 👋</p>
 
 <p>
-  You've been invited by <strong><%= @invite.sender.full_name %></strong> to manage <strong><%= @invite.event.name %></strong>,
+  You've been invited by <strong><%= @invite.sender.display_name %></strong> to manage <strong><%= @invite.event.name %></strong>,
```
garyhtou:
```suggestion
  You've been invited by <strong><%= @invite.sender.name %></strong> to manage <strong><%= @invite.event.name %></strong>,
```

app/helpers/users_helper.rb
```diff
@@ -85,7 +85,7 @@ def user_mention(user, options = {}, default_name = "No User")
            elsif user.id == current_user&.id
              current_user_flavor_text.sample
            elsif user.admin?
-             "#{user.name.split(' ').first} is an admin"
+             "#{user.name} is an admin"
```
``` suggestion
             "#{user.first_name} is an admin"
```
It would be nice to keep the existing functionality of just showing the first name. We can use the existing `first_name` method

app/views/users/edit.html.erb
```diff
@@ -7,61 +7,67 @@
...
-        </div>
-        <div class="field">
+      <div class="field">
+        <%= form.label :legal_name %>
+        <%= form.text_field :full_name, required: true, placeholder: "Fiona Hackworth III", pattern: '^\S+.+' %>
+      </div>
+      <div class="field">
+        <%= form.label :preferred_name %>
+        <p class="h5 muted mt0 mb1">If you go by a different name.</p>
+        <%= form.text_field :preferred_name, placeholder: "FiFi Hacks" %>
```
garyhtou: Let's set the placeholder to be their `full_name` (if it's `nil`, then fallback to a default name like "FiFi Hacks").

This way it shows that if a preferred name is not given, it shows your `full_name`

app/models/user.rb
garyhtou: for posterity: as we talked about in our sync, 30 or 40 sounds good. I'll leave this decision up to you

app/helpers/users_helper.rb
northeastprince: That's used for the `aria-label`, which is utilized instead of the content of an element if present. We already display the full name to the user, so I think we should make it accessible as well. If there's some other case I'm not getting where guessing the first name is better, I can change it.

garyhtou: `aria-label` is used to set the contents of the tooltip

**before (first name)**
image

**after (full preferred name)**
image

in the context of the tooltip, it would be nice to keep it short and just use the first name

northeastprince: Oh, huh. I thought it was only supposed to be used for screenreaders - that seems like more of a `title` attribute thing. Will it still say "Melanie Smith"?

northeastprince: Yeah, [that's what I'm finding](https://stackoverflow.com/a/63726152), so we probably wanna change that instead so we're not misusing an ARIA attribute. It might also make more sense to have some sort of badge next to an employee's name (cause if I have to hover over the name to find out if they're a Bank employee, it's not obvious and would probably be by accident), but I can propose that in a separate PR.

garyhtou: yeah, i think we are misusing the aria attribute.

As for the admin badge, there is currently a lightning bolt badge next to an admin's name (see image)

image

northeastprince: Ah, didn't get the meaning of that. Also, since we're abbreviating their name, it might actually make more sense to have the full name as a tooltip if we wanna keep that. Aside from a clearer admin badge in a future PR, is there anything I should change with this for now?

northeastprince: I could also change it to:

> Melania Smith (admin)

garyhtou: For this PR, let's either keep the current code as is:
```ruby
"#{user.name} is an admin"
```

or, use their first name (which is the existing functionality before this PR)
```ruby
"#{user.first_name} is an admin"
```

---

we can discuss how to make admin status and tooltips more clear in a different PR

garyhtou (approved):
Great work! good to merge 🥳

PR #4072, opened 2023-07-08
Expose total raised to transparency API

YodaLightsabr:
Expose the total amount raised to the Transparency API


sampoder (approved):
Looks good to me!

PR #4073, opened 2023-07-08
Expose Header On API

sampoder:


app/api/api/entities/organization.rb
```diff
@@ -25,6 +25,9 @@ class Organization < Base
         expose :logo do |organization|
           organization.logo.attached? ? Rails.application.routes.url_helpers.url_for(organization.logo) : nil
         end
+        expose :image do |organization|
```
garyhtou: can we call this `donation_header_image` or `donation_header`?

`image` is just a tad bit too broad

garyhtou (commented):
looks great! just a small change

app/api/api/entities/organization.rb
sampoder: let's do it!

sampoder:
```suggestion
        expose :donation_header do |organization|
```

sampoder:
Got Ian and Ben's approval IRL.

PR #4074, opened 2023-07-10
Add OpenGraph images for transparent org homepages

YodaLightsabr:
https://github.com/hackclub/bank-og


YodaLightsabr:
Reviewed IRL w/ @maxwofford

PR #4077, opened 2023-07-11
Fix #4075

northeastprince:
I don't think we need to be notified about invalid location info anymore since we now have the list of supported countries and it's passed back to the user.


northeastprince:
@cjdenio approved IRL

PR #4078, opened 2023-07-11
Don't attempt to notify empty orgs about invoices

northeastprince:
Closes #4069.


app/services/invoice_service/open_to_paid.rb
```diff
@@ -28,7 +28,9 @@ def run
         invoice.save!
       end
 
-      InvoiceMailer.with(invoice: @invoice).notify_organizers.deliver_later
+      if invoice.sponsor.event.users.any?
```
cjdenio:
```suggestion
      if invoice.event.users.any?
```

cjdenio (approved):

PR #4096, opened 2023-07-17
Organization: `description` (Mission Statement) field

garyhtou:
This will primarily be used for the Fiscal Sponsorship Directory. I'm limiting the field to 250 characters for easier/better UI.

image


garyhtou:
@YodaLightsabr what are your thoughts on this? especially with the max length

garyhtou:
Approved by @YodaLightsabr on call

PR #4097, opened 2023-07-17
API: Expose Organization donation link

garyhtou:
@YodaLightsabr, go ahead and merge this only when necessary (if we decide to add a donation button to the FS Directory UI)


YodaLightsabr:
It's just /donations/start/:org, right?

garyhtou:
Yeah, however, some orgs might not have their donation page enabled. Maybe instead of a donation link, we expose a boolean for whether their donation form is enabled

YodaLightsabr (approved):

PR #4105, opened 2023-07-20
Feature: Email receipts to Receipt Bin

YodaLightsabr:
This PR gives users the option to email receipts to a unique email address to make uploading receipts a faster process.

https://github.com/hackclub/hcb/assets/39828164/82477989-e26e-48c4-a454-a3fa39709681


garyhtou (dismissed):
Great start here! I'm going to propose a solution:

- Use [wicked_pdf](https://github.com/mileszs/wicked_pdf) to convert HTML to PDF
	- We already have this installed and it will prevent us from having to send receipts to a third party
	- Something like this should work:
	   ```ruby
	   pdf = WickedPdf.new.pdf_from_string('<h1>Hello There!</h1>')
	   # or
	   pdf = WickedPdf.new.pdf_from_html_file('/your/absolute/path/here')
	   ```
	   https://github.com/mileszs/wicked_pdf#super-advanced-usage
- We should only convert the HTML if it came from an inbound email. It's a bit confusing from the user perspective to upload an HTML receipt and then see that it's now a PDF 🤔

Let me know if it'd be helpful to have a discussion around this or if you have any questions

YodaLightsabr:
@garyhtou and I discussed the points above; I'm going to try moving forward with `wicked_pdf` for now. I'll also leave in the conversion on upload for testing for the time being.

app/views/receipts/_form_v3.html.erb
```diff
@@ -40,7 +40,6 @@
```
sampoder: What's the thinking behind the changes in this file?

sampoder (commented):
Thoughts on including the word "receipt" in the email address? It might be difficult for people to remember the unique identifier so if they can quickly search for the email when they send an in a receipt the second time (it'll show up in the list of previously sent to emails when they type receipt).

YodaLightsabr:
Gary mentioned not using hcb.gg in the past and I can sort of get behind that. We were thinking of using something along the lines of hcb-receipts.hackclub.com and keeping hcb.gg internal as to not confuse users.

sampoder:
What about just receipts.hackclub.com <http://receipts.hackclub.com/> or receipts.hcb.hackclub.com <http://receipts.hcb.hackclub.com/>? But yeah overall that seems like a perfect solution.

app/controllers/users_controller.rb
```diff
@@ -363,6 +364,15 @@ def update
...
 
+  def generate_receipt_identifier
+    # @user = params[:id] ? User.friendly.find(params[:id]) : current_user
+    # authorize @user
+
+    # response.content_type = "text/plain"
+
+    # render plain: @user.generate_receipt_identifier(ensure_validity: true)
+  end
+
```
```suggestion
```

app/models/mailbox_address.rb
```diff
@@ -0,0 +1,77 @@
...
+  def self.generate_address
+    high_end = 10**DISCRIMINATOR_LENGTH - 1
+    discriminator = rand(1..high_end).to_s.rjust(DISCRIMINATOR_LENGTH, "0")
+
+    animal = Faker::Creature::Animal.name.downcase.gsub(/[^a-z]/, "")
+
+    identifier = "#{animal}.#{discriminator}"
+    address = "#{identifier}@#{EMAIL_DOMAIN}"
+
+    return self.generate if self.exists?(address:)
```
garyhtou:
```suggestion
    return self.generate_address if self.exists?(address:)
```

app/models/mailbox_address.rb
```diff
@@ -0,0 +1,77 @@
...
+#
+#  fk_rails_...  (user_id => users.id)
+#
+class MailboxAddress < ApplicationRecord
+  DISCRIMINATOR_LENGTH = 4 # currently, 4 is enough to avoid most collisions, however this can be increased later
+  EMAIL_DOMAIN = "hcb.gg"
+
+  VALIDATION_REGEX = /\A[a-z]+\.\d{#{DISCRIMINATOR_LENGTH}}@#{Regexp.escape(EMAIL_DOMAIN)}\z/
+
+  belongs_to :user
```
garyhtou:
```suggestion
  belongs_to :user, dependent: :destroy
```

app/models/user.rb
```diff
@@ -218,8 +219,18 @@ def onboarding?
...
+  end
+
   private
 
+  def has_one_active_mailbox_address
+    return if mailbox_addresses.activated.count <= 1
+
+    record.errors.add :mailbox_addresses, :too_many_active_mailbox_address, message: "can only have one active mailbox address"
+  end
+
```
```suggestion
```

config/routes.rb
```diff
@@ -51,6 +51,7 @@
     post "receipts/upload", to: "static_pages#receipt", as: :my_receipts_upload
     get "missing_receipts", to: "static_pages#my_missing_receipts_list", as: :my_missing_receipts_list
     get "missing_receipts_icon", to: "static_pages#my_missing_receipts_icon", as: :my_missing_receipts_icon
+    get "generate_receipt_identifier", to: "users#generate_receipt_identifier"
```
garyhtou:
```suggestion
```

garyhtou (approved):

sampoder (approved):

PR #4106, opened 2023-07-20
V4 API

cjdenio:
###### _probably the last version!_

This is a rewritten HCB API for everything, not just viewing transparent orgs. It supports authentication and uses the exact same Pundit policies as the web app

todo:
- authorization scopes
- pagination


db/migrate/20230719194911_create_api_tokens.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class CreateApiTokens < ActiveRecord::Migration[7.0]
+  def change
+    create_table :api_tokens do |t|
+      t.text :token_ciphertext
+      t.string :token_bidx
+      t.index :token_bidx, unique: true
+
+      t.references :user, null: false, foreign_key: true
```
northeastprince:
```suggestion
      t.belongs_to :user, null: false, foreign_key: true
```
Makes the relationship a bit clearer imo, but just a nitpick.

northeastprince:
Thoughts on just using the same controllers, authenticating with an API token if present, and returning JSON for relevant routes so we can keep everything in the same place? Since everything's now in Rails and we can just progressively level up what we already have, no `/api/vX` necessary.

cjdenio:
Good suggestion, though I'd rather use separate controllers so we have the flexibility of structuring the API differently from the UI (e.g. the UI's `events#show` lists transactions while the API's `events#show` doesn't)

cjdenio:
(Merging so I can do funny mobile app experiments in a production environment)

PR #4128, opened 2023-08-07
Tweak tour donation copy to use active voice

cjdenio:


app/javascript/components/tour/TourOverlay.js
```diff
@@ -32,7 +32,7 @@ const tours = {
       if (options.category != 'outernet guild') {
         steps.push({
           attachTo: isMobile ? 'receive' : 'donations',
-          text: 'Share your donation form with others, and get it embedded on your website.',
+          text: 'Share your donation form with others and embed it on your website.',
```
northeastprince:
```suggestion
          text: 'Share your donation form with others and embed it anywhere.',
```

cjdenio: Disagree — I think it's better to emphasize a specific user story, so it feels more personal ("hey, this is a feature I could use because I have a website!")

 \- @kognise borrowing caleb's computer

PR #4148, opened 2023-08-11
Add "superadmins" who can make other people admins

cjdenio:
This PR replaces `admin_at` with an `access_level` enum field (+ `paper_trail` for tracking people)

Closes #4146


app/views/users/edit_admin.html.erb
```diff
@@ -31,6 +34,20 @@
...
         <% end %>
 
+        <% if current_user.superadmin? %>
+          <div class="field">
+            <%= form.label :access_level, "Access level" %>
+            <%= form.select :access_level,
+                            [
+                              ["👤 Regular user", "user"],
+                              ["🛡️ Admin", "admin"],
+                              ["👑 Superadmin (can manage other users' admin status)", "superadmin"],
```
garyhtou: i petition that we call this role Orpheus :) 🦕🪧🗣️📣

PR #4152, opened 2023-08-15
Add Support for Donations in AUD, SGD, EUR, NZD, and the GBP

sampoder:
(this a draft.... I'm trying to preview what has already been changed on this branch & remedy the sins in #3099)


sampoder:
I'm going to grab lunch after dealing with Rubocop. Here are the known issues with this right now:

- [x] Available currencies are wrong (easy fix)
- [x] Refunding donations will probably not work for donations not in USD
- [x] On the frontend, we need to handle currencies that don't act like the dollar. Most pressing one is the Yen. Alternatively, we ship with just "USD", "AUD", "SGD", "EUR" and other similar currencies. I am probs pro-this as we don't have any orgs in Japan.
- [x] Related ^^ the "suggested amount" might need to be changed 20 yen is effectively nothing.
- [x] Need to look into modifying the minimum, 1 SGD < 1 USD, not sure if Stripe will charge 1 SGD

Last three points were done by restricting the currencies to those that act like the US dollar.

sampoder:
for reference: here's how Caleb did it https://github.com/hackclub/bank/commit/e476fb1c92516e074f7eb35194673580fd0eb2f1. 

we seem to have largely done it the same way except: Caleb stores currencies using an enum while I'm directly storing the currency code in the DB and Caleb let's the donor choose the currency while I'm having the organisation set the currency. personally, having the organisation set it makes more sense to me as its a regional thing, @cjdenio whatcha think?

cjdenio:
Thanks for this! Unfortunately I don't have time to review a change this close to Bank's financial infrastructure— Stripe has some dangerous edge cases around multi-currency that would need to be tested (specifically, making sure HCB still treats everything as USD on the backend). I'd love for this to be a thing someday though!

PR #4154, opened 2023-08-16
Maintenance banner for signed in users

garyhtou:
This banner has the same design and location as the Org transparency banner. However, they won't conflict because the transparency banner is only shown when signed out.

image


app/views/layouts/application.html.erb
```diff
@@ -50,6 +50,19 @@
...
 
+    <% if signed_in? %>
+      <div class="w-100 px3 bg-white transparency-banner" data-behavior="hide_iframe">
+        <p class="h5 medium center mt1 mb1">
+          Hack Club Bank will be under maintenance on
+          <strong>Tuesday, August 29th</strong> from 1:00am to 3:00am PT.
+          <br>
+          You will not be able to access your account and
+          <i>card transactions will decline</i> during this time.
+          <%= link_to "Learn more", "" %>
```
garyhtou: link TBD

app/views/layouts/application.html.erb
```diff
@@ -50,6 +50,19 @@
...
     <% end %>
 
+    <% if signed_in? %>
+      <div class="w-100 px3 bg-white transparency-banner" data-behavior="hide_iframe">
+        <p class="h5 medium center mt1 mb1">
+          Hack Club Bank will be under maintenance on
+          <strong>Tuesday, August 29th</strong> from 1:00am to 3:00am PT.
+          <br>
+          You will not be able to access your account and
+          <i>card transactions will decline</i> during this time.
```
garyhtou: we'll come back and update the copy to match the email

app/views/layouts/application.html.erb
```diff
@@ -50,6 +50,19 @@
       </div>
     <% end %>
 
+    <% if signed_in? %>
+      <div class="w-100 px3 bg-white transparency-banner" data-behavior="hide_iframe">
+        <p class="h5 medium center mt1 mb1">
+          Hack Club Bank will be under maintenance on
+          <strong>Tuesday, August 29th</strong> from 1:00am to 3:00am PT.
```
cjdenio: Might be cool to show this in the user's local timezone!

garyhtou: great idea! i had looked into it when i initially threw together this pr but found that https://github.com/basecamp/local_time doesn't support date ranges.

it's a little more involved, but I went ahead and added it [`050027d` (#4154)](https://github.com/hackclub/bank/pull/4154/commits/050027def914e0b818c8a8ebe320687d672d4864)

app/views/application/_maintenance_banner.html.erb
```diff
@@ -0,0 +1,25 @@
+<% if signed_in? && local_assigns.has_key?(:datetime) && local_assigns.has_key?(:duration) %>
+  <% if (datetime + duration) > Time.now %><%# if it hasn't already passed %>
```
albertchae: Hmm, I might be fine requiring a deploy to remove this just in case it takes longer. But this is fine too, not a big deal (especially since we'll literally be using a [heroku maintenance page](https://devcenter.heroku.com/articles/maintenance-mode) until the database upgrade is finished anyway)

app/views/layouts/application.html.erb
```diff
@@ -50,6 +50,10 @@
       </div>
     <% end %>
 
+    <%= render "application/maintenance_banner",
+               datetime: "2023-08-29T01:00:00-0700".to_datetime, duration: 2.hours,
+               learn_more_url: "https://postal.hackclub.com/w/ohsB1xRMhhuwbFeWwVihLQ" %>
```
albertchae: nice, should I use this URL for https://devcenter.heroku.com/articles/maintenance-mode#customizing-your-maintenance-page too?

albertchae (approved):

app/views/layouts/application.html.erb
garyhtou: That could be a good idea. It's currently set to https://hackclub.github.io/bank/maintenance-mode.html which provides no context/info

app/views/application/_maintenance_banner.html.erb
garyhtou: yeah true, that message wouldn't be visible anyways during maintenance. I do plan on making a PR afterwards to remove the partial render.

PR #4156, opened 2023-08-17
Prompt to install as PWA during first visit on Safari mobile

sampoder:
I'm tracking first visits using a cookie and then showing it if they're on Safari / iOS browser

Closes #3864


app/controllers/application_controller.rb
```diff
@@ -17,6 +17,19 @@ class ApplicationController < ActionController::Base
...
+  def handle_cookies
+    if current_user
+      if cookies[:first_visit] != "1"
+        cookies.permanent[:first_visit] = 1
+        @first_visit = true
+      else
+        @first_visit = false
+      end
+    end
+  end
```
garyhtou: i believe this should do the same thing? (untested)

```suggestion
  before_action do
    return current_user.nil?

    @first_visit = cookies[:first_visit] != "1"
    cookies.permanent[:first_visit] = 1
  end
```

sampoder: Yes it would!

app/controllers/application_controller.rb
```diff
@@ -17,6 +17,14 @@ class ApplicationController < ActionController::Base
   # Redirect users to the onboarding page if they haven't completed it yet
   before_action :redirect_to_onboarding
 
+  # This cookie is used for Safari PWA prompts
+  before_action do
+    return current_user.nil?
```
cjdenio:
```suggestion
    return if current_user.nil?
```

garyhtou: whoops thanks for catching that

sampoder: thanks Caleb!

sampoder:
@cjdenio & @garyhtou: do either of you know what's going on with the RSpec test? looks like it came about after https://github.com/hackclub/bank/pull/4156/commits/9c4a69c998e90ffe8aff3a35fb3aa44adaed2cec.

app/controllers/application_controller.rb
```diff
@@ -17,6 +17,14 @@ class ApplicationController < ActionController::Base
   # Redirect users to the onboarding page if they haven't completed it yet
   before_action :redirect_to_onboarding
 
+  # This cookie is used for Safari PWA prompts
+  before_action do
+    return if current_user.nil?
```
cjdenio:
```suggestion
    break if current_user.nil?
```

my bad!

sampoder: all good!

sampoder:
yay! it's fixed!

cjdenio (dismissed):
A few things:
- Most users won't see this because `first_visit` will be set as soon as they open the login page, meaning they won't see it once they're logged in
- The prompt should be dismissible
- Contrast in light mode could be better

app/controllers/application_controller.rb
```diff
@@ -17,6 +17,14 @@ class ApplicationController < ActionController::Base
   # Redirect users to the onboarding page if they haven't completed it yet
   before_action :redirect_to_onboarding
 
+  # This cookie is used for Safari PWA prompts
+  before_action do
+    next if current_user.nil?
```
sampoder: @cjdenio I added this line here to try and prevent your first bullet point, is there a better way / correct way of doing this? my thinking was that on the login page the current_user would be `nil`

agree on point 2 & 3 btw, will fix asap! thanks :D

sampoder:
Screenshot 2024-03-26 at 2 16 48 AM

PR #4157, opened 2023-08-17
Move Transparency Mode Message To Under The Balances

sampoder:
This was originally proposed in https://github.com/hackclub/bank/pull/2110/, and then suggested by @YodaLightsabr in https://github.com/hackclub/bank/issues/3936. This would fix #3936.


app/views/events/show.html.erb
```diff
@@ -93,6 +93,8 @@
   </div>
 </div>
 
+<%= render 'events/public_message' %>
```
sampoder:
```suggestion
<%= render "events/public_message" %>
```
double the quotes! double the fun!

YodaLightsabr (approved):
Looks good!

PR #4158, opened 2023-08-17
SMS Users When IRL Transactions Are Declined

sampoder:
This PR also clarifies that `send_twilio_message` was for receipts by renaming it to `send_twilio_receipt_message`. I'm not sure if it's best to use the Flipper flag for receipts in this case, however, I left it for now. Will wait to hear what others think. I think this fixes #3560.


app/jobs/canonical_pending_transaction_job/send_twilio_declined_message.rb
```diff
@@ -0,0 +1,54 @@
...
+                             "was declined because this card isn't allowed to make purchases at #{@merchant}"
+                           else
+                             "was declined due to insufficient funds on the card"
+                           end
+                         else
+                           "was declined"
+                         end
+
+      hcb_code = @cpt.local_hcb_code
+      message = "Your card #{humanised_reason}."
```
YodaLightsabr:
```suggestion
      humanized_reason = case @reason
                         when "card_inactive"
                           if !@card.activated?
                             "at #{@merchant} because this card hasn't been activated yet"
                           else
                             "at #{@merchant} because this card has been frozen"
                           end
                         when "suspected_fraud"
                           "at #{@merchant} due to suspected fraud"
                         when "webhook_declined"
                           case @webhook_declined_reason
                           when "merchant_not_allowed"
                             "because this card isn't allowed to make purchases at #{@merchant}"
                           else
                             "at #{@merchant} due to insufficient funds"
                           end
                         else
                           "at #{@merchant}"
                         end

      hcb_code = @cpt.local_hcb_code
      message = "Your card was declined just now #{humanized_reason}."
```
Thoughts on changing up the wording a bit? I was thinking
- it might be helpful to show the merchant name even if it isn't the reason why the card declined
- including "just now" might be a helpful indicator that it's a real-time notification
- `humanized` instead of `humanised` just to be consistent with the rest of the codebase

app/jobs/canonical_pending_transaction_job/send_twilio_declined_message.rb
```diff
@@ -0,0 +1,54 @@
...
+    # always correct.
+    IN_PERSON_AUTH_METHODS = %w[keyed_in swipe chip contactless].freeze
+
+    def perform(cpt_id:, user_id:)
+      @cpt = CanonicalPendingTransaction.find(cpt_id)
+      @card = @cpt.raw_pending_stripe_transaction.stripe_card
+      @merchant = @cpt.raw_pending_stripe_transaction.stripe_transaction["merchant_data"]["name"]
+      @reason = @cpt.raw_pending_stripe_transaction.stripe_transaction["request_history"][0]["reason"]
+      @webhook_declined_reason = @cpt.raw_pending_stripe_transaction.stripe_transaction.dig("metadata", "declined_reason")
+      @user = User.find(user_id)
```
YodaLightsabr:
```suggestion
      @cpt = CanonicalPendingTransaction.find(cpt_id)
      @rpst = @cpt.raw_pending_stripe_transaction
      @card = @rpst.stripe_card
      @merchant = @rpst.stripe_transaction["merchant_data"]["name"]
      @reason = @rpst.stripe_transaction["request_history"][0]["reason"]
      @webhook_declined_reason = @rpst.stripe_transaction.dig("metadata", "declined_reason")
      @user = User.find(user_id)
```
This might keep the lines from getting too long

app/jobs/canonical_pending_transaction_job/send_twilio_declined_message.rb
```diff
@@ -0,0 +1,54 @@
...
+      hcb_code = @cpt.local_hcb_code
+      message = "Your card #{humanised_reason}."
+
+      TwilioMessageService::Send.new(@user, message, hcb_code:).run!
+    end
+
+    private
+
+    def auth_method
+      @cpt&.raw_pending_stripe_transaction&.authorization_method
```
YodaLightsabr:
```suggestion
      @cpt.raw_pending_stripe_transaction.authorization_method
```
We might be able to remove the safe navigation down here since we're not doing it at the top

app/jobs/canonical_pending_transaction_job/send_twilio_declined_message.rb
sampoder: Merchant name is a good idea, as is "just now"! And humanized is of course my favourite spelling of that word.

app/jobs/canonical_pending_transaction_job/send_twilio_declined_message.rb
sampoder: sweet!

YodaLightsabr (approved):
Looks good!

PR #4160, opened 2023-08-17
Add The Option To Reverse Admin Search Queries

sampoder:
I believe this fixes #3550; @maxwofford & @MelSmith104 is this what you were looking for? I think this let's you do a query that would return all accounts that aren't hardware accounts:


app/controllers/admin_controller.rb
```diff
@@ -1175,6 +1176,13 @@ def filtered_events(events: Event.all)
     states << "approved" if @approved
     states << "rejected" if @rejected
     relation = relation.where("aasm_state in (?)", states)
+
+    if @inverse
+      relation = Event.all.where.not(id: relation.pluck(:id))
+    else
+      relation
+    end
```
garyhtou: check this out!
https://blog.saeloun.com/2021/04/21/rails-7-invert-where-method/

sampoder: That's sick! So would that look like turning:

```
@events = filtered_events.page(@page).per(@per).reorder(Arel.sql("COALESCE(events.activated_at, events.created_at) desc"))
```

into when inverted is selected:

```
@events = filtered_events.invert_where.page(@page).per(@per).reorder(Arel.sql("COALESCE(events.activated_at, events.created_at) desc"))
```

maxwofford:
To clarify the original issue, we're looking for something like this:

Screenshot 2023-08-18 at 13 16 46

Just like how funded can be "Funded", "Not funded", or "Doesn't matter if funded or not"

sampoder:
Alrighty! I made a checkbox thing for events:

Screenshot 2023-08-18 at 7 03 04 PM

sampoder:
my only worry with the checkboxes is that selecting them all is a pain. maybe I could add an option that unselects them all. thoughts?

sampoder:
Is this still of interest to people or should I close this?

PR #4161, opened 2023-08-18
Add Custom Thank You Messages To Donor Receipts

sampoder:
And also let people receive replies to these emails themselves. Fixes #4147.


YodaLightsabr:
Pulling from main to get the Heroku deploy working again...

sampoder:
thank you!!

YodaLightsabr (approved):
Looks good!

app/models/event.rb
```diff
@@ -19,6 +19,8 @@
 #  description                     :text
 #  donation_page_enabled           :boolean          default(TRUE)
 #  donation_page_message           :text
+#  donation_reply_to_email         :text
+#  donation_thank_you_message      :text
```
garyhtou: Let's add some validations for both the email and thank you message. For example:

- ensure that the email is valid (even though we're validating this on the frontend using HTML 5, it could be beneficial to also validate it on the backend)
- ty message must be less than 500 characters

app/views/donation_mailer/donor_receipt.html.erb
```diff
@@ -10,6 +10,15 @@
 </p>
 <% end %>
 
+<% if @event.donation_thank_you_message.present? && (@initial_recurring_donation || !@donation.recurring?) %>
```
garyhtou: i think it could be nice to send it for all donations, regardless of whether it's recurring. partially because it wouldn't hurt, and the ty message could change between donations

garyhtou (commented):
yayy! couple small things

app/views/donation_mailer/donor_receipt.html.erb
sampoder:
```suggestion
<% if @event.donation_thank_you_message.present? %>
```

app/models/event.rb
```diff
@@ -77,6 +79,9 @@ class Event < ApplicationRecord
   has_paper_trail
   acts_as_paranoid
   validates_as_paranoid
+  
+  validates_email_format_of :donation_reply_to_email
```
```suggestion
  validates_email_format_of :donation_reply_to_email, allow_nil: true, allow_blank: true 
```

app/mailers/donation_mailer.rb
```diff
@@ -5,7 +5,7 @@ def donor_receipt
     @donation = params[:donation]
     @initial_recurring_donation = @donation.initial_recurring_donation? && !@donation.recurring_donation&.migrated_from_legacy_stripe_account?
 
-    mail to: @donation.email, subject: @donation.recurring? ? "Receipt for your donation to #{@donation.event.name} — #{@donation.created_at.strftime("%B %Y")}" : "Receipt for your donation to #{@donation.event.name}"
+    mail to: @donation.email, reply_to: @donation.event.donation_reply_to_email || nil, subject: @donation.recurring? ? "Receipt for your donation to #{@donation.event.name} — #{@donation.created_at.strftime("%B %Y")}" : "Receipt for your donation to #{@donation.event.name}"
```
garyhtou:
```suggestion
    mail to: @donation.email, reply_to: @donation.event.donation_reply_to_email.presence, subject: @donation.recurring? ? "Receipt for your donation to #{@donation.event.name} — #{@donation.created_at.strftime("%B %Y")}" : "Receipt for your donation to #{@donation.event.name}"
```

garyhtou (approved):
great work here! i'm excited to see this used

PR #4171, opened 2023-08-21
Fix PDF receipt uploads when users are signed out

YodaLightsabr:
Fixes #4010, an error where uploading PDF receipts fails when users are signed out.

An error displayed while trying to pair a receipt after uploading


garyhtou (commented):
Let's rename `needs_extraction?` to `has_content?` and move it into the Receipt model

sampoder (approved):

garyhtou (approved):

PR #4172, opened 2023-08-21
Add Exports To Transparent Organisations & Refactor Exports Into a Dedicated Controller 

sampoder:
This PR removes the statements that hid the export feature when viewing transparent organisations. Most of it, however, is focused on adding a page to collect an email for large exports (those viewing transparent organisations aren't always signed-in).

I'd also appreciate if someone could do a quick security sanity check on this one.

Fixes #3371.


garyhtou (commented):
Great start to this PR. Right now, we're handling exports in the `TransactionsController` and now we're adding to the `StaticPagesController` (which I totally understand why). However, I'm nervous about spreading the export logic throughout the codebase more. I would prefer if we focused on refactoring to an `Export` model and controller (and maybe a `FinanicalExport` model) before we add this feature

sampoder:
Alrighty @garyhtou! This PR has ballooned a little but.... this should now refactor everything into an export controller and into export jobs so there shouldn't be any tie back into the transactions controller. 

One thing is that I didn't end up needing a model for what we currently have.

app/controllers/static_pages_controller.rb
```diff
@@ -4,7 +4,7 @@
 
 class StaticPagesController < ApplicationController
   skip_after_action :verify_authorized # do not force pundit
-  skip_before_action :signed_in_user, only: [:branding, :brand_guidelines, :faq]
+  skip_before_action :signed_in_user, only: [:branding, :brand_guidelines, :faq, :email_transparent_export]
```
```suggestion
  skip_before_action :signed_in_user, only: [:branding, :brand_guidelines, :faq]
```

app/controllers/transactions_controller.rb
```diff
@@ -4,6 +4,7 @@
 
 class TransactionsController < ApplicationController
   skip_before_action :signed_in_user
+  skip_after_action :verify_authorized, only: [:export]
```
garyhtou:
```suggestion
```

app/mailers/canonical_transaction_mailer.rb
```diff
@@ -1,17 +1,18 @@
 # frozen_string_literal: true
 
 class CanonicalTransactionMailer < ApplicationMailer
```
garyhtou: Can we move this to `ExportMailer`?

app/views/exports/collect_email.erb
```diff
@@ -0,0 +1,25 @@
...
+
+<div class="container container--sm card py3 md-p3 lg-p4 mt2 mt3 md-mt4 overflow-visible">
+  <h3 class="mt2 md-mt0">
+    This export is too big, so we'll send you an email when it's ready.
+  </h3>
+  <p>
+    We don't store emails submitted here in any of our databases, once you've recieved your file,
+    you won't hear from us again. 
+  </p>
+  <%= form_with(model: nil, local: true, method: "get", url: "/exports/#{@event_slug}.#{@file_extension}") do |form| %>
```
garyhtou: See if we can use a URL helper here

garyhtou: maybe something like this?
```ruby
export_exports_path(event: @event_slug, format: @file_extension)
```

app/controllers/exports_controller.rb
```diff
@@ -0,0 +1,149 @@
+# frozen_string_literal: true
+
+require "csv"
+
+class ExportsController < ApplicationController
+  skip_before_action :signed_in_user
+  skip_after_action :verify_authorized, only: [:export, :collect_email]
+
+  def export
```
garyhtou: rename `export` action to `transactions`

garyhtou (commented):
great job refactoring! just a couple things left

app/views/exports/collect_email.erb
sampoder:
```suggestion
  <%= form_with(model: nil, local: true, method: "get", url: export_exports_path(event: @event_slug, format: @file_extension)) do |form| %>
```

app/controllers/exports_controller.rb
```diff
@@ -0,0 +1,149 @@
+# frozen_string_literal: true
+
+require "csv"
```
garyhtou:
```suggestion
```

app/mailers/export_mailer.rb
```diff
@@ -1,17 +1,18 @@
 # frozen_string_literal: true
 
-class CanonicalTransactionMailer < ApplicationMailer
-  def export_ready(event:, user:, mime_type:, title:, content:)
+class ExportMailer < ApplicationMailer
+  def export_ready(event:, email:, mime_type:, title:, content:)
     @event = event
     @user = user
```
garyhtou:
```suggestion
```

garyhtou (approved):

PR #4177, opened 2023-08-21
Add The Ability To Batch Edit The Ledger

sampoder:
This adds a column of checkboxes and then a little form at the bottom:

Screenshot 2023-08-21 at 2 35 49 PM

Screenshot 2023-08-21 at 2 36 00 PM

@hackliv was this the sort of thing you were looking for in #4113?


hackliv:
yes!! this would be amazing when mapping our payroll :) thanks @sampoder!!

sampoder:
yay!

app/controllers/admin_controller.rb
```diff
@@ -1010,6 +1010,23 @@ def set_event
     redirect_to transaction_admin_path(params[:id]), flash: { error: e.message }
   end
 
+  def batch_set_event
+    error_occurred = false
+    params.each do |key, value|
```
maxwofford: This should probably be in an [ActiveRecord::Transaction](https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html)

sampoder: oh neat, TIL! added

sampoder:
@maxwofford @hackliv this good to merge for both of you?

app/controllers/admin_controller.rb
```diff
@@ -1015,6 +1015,25 @@ def set_event
...
+        begin
+          @canonical_transaction = ::CanonicalTransactionService::SetEvent.new(canonical_transaction_id: id, event_id: params[:event_id], user: current_user).run
+        rescue => e
+          redirect_to transaction_admin_path(id), flash: { error: e.message }
+          error_occurred = true
+          break
+        end
+      end
+    end
+    redirect_back fallback_location: ledger_admin_index_path unless error_occurred
```
```suggestion
    ActiveRecord::Base.transaction do
      params.each do |key, value|
        next unless key.start_with?("checkbox-") && value == "1"

        id = key.sub("checkbox-", "")
        begin
          @canonical_transaction = ::CanonicalTransactionService::SetEvent.new(canonical_transaction_id: id, event_id: params[:event_id], user: current_user).run
        rescue => e
          redirect_to transaction_admin_path(id), flash: { error: e.message }
          return
        end
      end
    end
    redirect_back fallback_location: ledger_admin_index_path
```

```suggestion
  def batch_set_event
    error_occurred = false
    ActiveRecord::Base.transaction do
      cts = params.filter_map do |key, value|
        next unless key.start_with?("checkbox-") && value == "1"

        id = key.sub("checkbox-", "")
        begin
          ::CanonicalTransactionService::SetEvent.new(canonical_transaction_id: id, event_id: params[:event_id], user: current_user).run
        rescue => e
          redirect_to transaction_admin_path(id), flash: { error: e.message }
          error_occurred = true
          break
        end
      end
      flash[:error] = "No transactions provided to bulk map event" if cts.empty?
    end
    redirect_back fallback_location: ledger_admin_index_path unless error_occurred
```

app/views/admin/ledger.html.erb
```diff
@@ -46,45 +46,53 @@
...
+              <br>
+              <%= ct.canonical_event_mapping.try(:user).try(:email) %>
+            <% end %>
+          </td>
+        </tr>
+      <% end %>
+    </tbody>
+  </table>
+  <%= form.collection_select(:event_id, Event.reorder(Event::CUSTOM_SORT).not_demo_mode, :id, :admin_dropdown_description, { include_blank: true }, { width: 300, style: "max-width: 300px; margin-right: 16px;" }) %>
+  <%= form.submit "Set" %>
```
garyhtou:
```suggestion
  <%= form.submit "Bulk Map to Event" %>
```

PR #4178, opened 2023-08-22
Hide donation page configuration options in embed preview

sampoder:
Fixes #3309


YodaLightsabr (approved):
Gone are the days of changing the Hack OC donation page background from within the embed :((((

PR #4183, opened 2023-08-23
Enable donor messages without param and email on donation w/ message

sampoder:
Donor messages are such a fun idea! I thought it'd be nice to make it so people could leave messages. I also added a mailer so that if you get a message, it get sent to all the organisers.


app/views/donation_mailer/donation_with_message_notification.html.erb
```diff
@@ -0,0 +1,14 @@
+<p><%= @emails.length > 1 ? "Hi all," : "Hi there," %></p>
+
+<p>
+  Congrats! We're emailing to let you know that you've received your a donation of <%= render_money @donation.amount %> from <%= @donation.name %>. They left you a message:
```
sampoder:
```suggestion
  Congrats! We're emailing to let you know that you've received a donation of <%= render_money @donation.amount %> from <%= @donation.name %>. They left you a message:
```

app/mailers/donation_mailer.rb
```diff
@@ -15,4 +15,11 @@ def first_donation_notification
     mail to: @emails, subject: "Congrats on receiving your first donation! 🎉"
   end
 
+  def donation_with_message_notification
+    @donation = params[:donation]
+    @emails = @donation.event.users.map { |u| u.email }
```
garyhtou:
```suggestion
    @emails = @donation.event.users.pluck(:email)
```

Pluck is more performant than `map` since it's limited the amount of data `SELECT`ed from the database

app/mailers/donation_mailer.rb
```diff
@@ -15,4 +15,11 @@ def first_donation_notification
...
+
+    mail to: @emails, subject: "A new donor sent you a message! 🎉"
```
garyhtou:
```suggestion
    mail to: @emails, subject: "You've received a donation! 🎉"
```

Organizers will likely want to know that they've received a donation, rather than just a note. Also, this mailer is used for all donations with messages, not just ones from _new_ donors

app/models/donation.rb
```diff
@@ -285,9 +285,15 @@ def send_payment_notification_if_needed
     return unless saved_changes[:status].present?
 
     return unless status_changed_to_succeeded?(saved_changes[:status])
+
+    if first_donation?
+      DonationMailer.with(donation: self).first_donation_notification.deliver_later
+    elsif includes_message?
+      DonationMailer.with(donation: self).donation_with_message_notification.deliver_later
+    end
     return unless first_donation?
```
garyhtou:
```suggestion
```
no longer needed since it's handled by the `if`

app/views/donation_mailer/first_donation_notification.html.erb
```diff
@@ -8,6 +8,16 @@
   The funds are available to spend right away. You can <%= link_to "click here", event_donation_overview_url(@donation.event.slug) %> to see the full details.
 </p>
 
+<% if @donation.message.present? %>
```
garyhtou:
```suggestion
<% if @donation.includes_message? %>
```

garyhtou (approved):
heck yeahhh!!

YodaLightsabr:
Awesome work on donor messages, Sam!

I just got an email from HCB about a donation, and I was a bit confused at first.

Should we maybe add the organization name to the email or a link to the transaction?

PR #4184, opened 2023-08-24
Add the ability to create custom QR codes

sampoder:
Now you can add cool icons on the middle of your QR code!

Screenshot 2023-08-24 at 1 59 26 AM

We might be able to play around with colours as well if we wanted


sampoder:
@YodaLightsabr could you push your changes from last night? I'll go through and make all the classes match the BEM standard 👍

YodaLightsabr (approved):

PR #4188, opened 2023-08-25
Add Dropdown For "per" Param

sampoder:
Closes #3465


app/controllers/events_controller.rb
```diff
@@ -49,11 +49,13 @@ def show
     page = (params[:page] || 1).to_i
     per_page = (params[:per] || 75).to_i
 
-    @transactions = Kaminari.paginate_array(@all_transactions).page(page).per(per_page)
-    TransactionGroupingEngine::Transaction::AssociationPreloader.new(transactions: @transactions, event: @event).run!
+    @transactions = Kaminari.paginate_array(@pending_transactions + @all_transactions).page(page).per(per_page)
+    @settled_transactions_paginated = @transactions.select { |transaction| !@pending_transactions.include?(transaction) }
+    @pending_transactions_paginated = @transactions.select { |transaction| @pending_transactions.include?(transaction) }
```
sampoder: This shouldn't increase page load too much, right?

app/controllers/events_controller.rb
```diff
@@ -49,11 +49,13 @@ def show
...
+    TransactionGroupingEngine::Transaction::AssociationPreloader.new(transactions: @settled_transactions_paginated, event: @event).run!
 
     if show_running_balance?
-      offset = page * per_page
+      offset = page * per_page - _show_pending_transactions.length
```
sampoder: Running balance still seems to be correct after adding this line

app/controllers/events_controller.rb
```diff
@@ -665,7 +667,6 @@ def user_event_params
   end
 
   def _show_pending_transactions
-    return [] if params[:page] && params[:page] != "1"
```
sampoder: Are there any knock-on effects of removing this line? I've removed as we need to know how many pending transactions on every page for the offset

PR #4189, opened 2023-08-25
Fingerprint Donations with UA & IP

sampoder:
Fixes #4129


db/migrate/20230825055850_add_ua_and_ip_to_donation.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddUaAndIpToDonation < ActiveRecord::Migration[7.0]
+  def change
+    add_column :donations, :ua, :text
+    add_column :donations, :ip, :text
```
northeastprince: Maybe make these a little clearer?

```suggestion
    add_column :donations, :user_agent, :string
    add_column :donations, :ip_address, :string
```

Also I think string makes more sense instead of text. Doesn't make a practical difference with a database like Postgres, but I consider those values instead of content like a memo or comment might be.

app/controllers/admin_controller.rb
```diff
@@ -713,6 +715,8 @@ def donations
       end
     end
 
+    relation = relation.where(ip_address: @ip_address) if @ip_address
+    relation = relation.where(user_agent: @user_agent) if @user_agent
```
```suggestion
    relation = relation.where("user_agent ILIKE ?", "%#{Donation.sanitize_sql_like(@user_agent)}%") if @user_agent
```

So we can do partial matches on user agents

cjdenio (approved):
This looks great, thanks! Just left one comment.

PR #4191, opened 2023-08-25
Add a tool to report & be invoiced for "accidental misuse"

sampoder:
For when you tap your HQ card and not the other one...
image

image

Only available to HQ orgs for now!


app/views/hcb_codes/_stripe_card.html.erb
```diff
@@ -249,7 +249,17 @@
...
-  <%= render partial: "hcb_codes/dispute", locals: { type: "card", hcb_code: @hcb_code } %>
+  <div class="flex justify-end align-middle flex-wrap " style="gap: 12px">
+    <div class="flex muted justify-end items-end">
+      <%= inline_icon "sad", size: 20 %>
+      <%= link_to invoice_as_personal_transaction_hcb_code_path(id: @hcb_code.hashid, event_id: @event.id || @hcb_code.event) do %>
+        <small class="mt0 mb0">
+          Flag as accidental misuse
+        </small>
+      <% end %>
+    </div>
```
garyhtou: Can we move this to a `accidental_misuse` partial?

app/controllers/hcb_codes_controller.rb
```diff
@@ -163,4 +162,35 @@ def toggle_tag
...
 
+  def invoice_as_personal_transaction
+    hcb_code = HcbCode.find_by(hcb_code: params[:id]) || HcbCode.find(params[:id])
+    event = Event.friendly.find(params[:event_id])
+
+    authorize hcb_code
+
+    @sponsor = Sponsor.new(event:)
+
+    due_date = Date.current.advance(months: 1)
```
garyhtou: I think this might be the same
```ruby
due_date = 1.month.from_now
```
if so, we can set this directly in the call to `.new`

app/views/hcb_codes/_stripe_card.html.erb
```diff
@@ -249,7 +249,17 @@
     <% end %>
 
   </article>
-  <%= render partial: "hcb_codes/dispute", locals: { type: "card", hcb_code: @hcb_code } %>
+  <div class="flex justify-end align-middle flex-wrap " style="gap: 12px">
+    <div class="flex muted justify-end items-end">
+      <%= inline_icon "sad", size: 20 %>
+      <%= link_to invoice_as_personal_transaction_hcb_code_path(id: @hcb_code.hashid, event_id: @event.id || @hcb_code.event) do %>
```
garyhtou: Is there a case where `@event.id` is nil?

app/controllers/hcb_codes_controller.rb
```diff
@@ -163,4 +162,35 @@ def toggle_tag
     redirect_back fallback_location: @event
   end
 
+  def invoice_as_personal_transaction
+    hcb_code = HcbCode.find_by(hcb_code: params[:id]) || HcbCode.find(params[:id])
```
```suggestion
    hcb_code = HcbCode.find_by(hcb_code: params[:id])
```

Since we're passing the hcb_code in the code below, we likely don't need the 2nd part

app/controllers/hcb_codes_controller.rb
```diff
@@ -163,4 +162,35 @@ def toggle_tag
...
+    event = Event.friendly.find(params[:event_id])
+
+    authorize hcb_code
+
+    @sponsor = Sponsor.new(event:)
```
garyhtou: this appears to be unused?

garyhtou (changes requested):
Can you clean this up and then re-request for review? Also, let's limit this to HQ orgs for now: `event.hack_club_hq?`

app/controllers/hcb_codes_controller.rb
sampoder: It's used on line 181 at the moment

app/controllers/hcb_codes_controller.rb
sampoder: Oh yes it is!

config/routes.rb
```diff
@@ -347,6 +347,7 @@
       get "attach_receipt"
       get "memo_frame"
       get "dispute"
+      get "invoice_as_personal_transaction"
```
garyhtou: Can we make this a POST? we shouldn't modify the database with a GET request

sampoder: oh yes mb

app/controllers/hcb_codes_controller.rb
garyhtou: oops, i missed that, thanks!

garyhtou (approved):
cha-ching! 💸

garyhtou:
@sampoder this is a good one to ship in #hcb-team!

PR #4192, opened 2023-08-26
Add Multi-Selecting Transactions for Tags

sampoder:
Alrighty! Here's how this works:

1. Command-click / Ctrl-click the transactions they want to select
2. This button will appear: Screenshot 2023-08-26 at 2 11 30 PM
3. From there it works like tags usually do: 
Screenshot 2023-08-26 at 2 12 18 PM

Fixes #4174... I think?


app/assets/stylesheets/components/_transactions.scss
```diff
@@ -119,3 +119,23 @@ html[data-dark='true'] {
     background-color: none;
   }
 }
+
+.transaction__icon__when__selected {
```
northeastprince:
```suggestion
.transaction__icon__when-selected {
```

app/assets/stylesheets/components/_transactions.scss
```diff
@@ -119,3 +119,23 @@ html[data-dark='true'] {
     background-color: none;
   }
 }
+
+.transaction__icon__when-selected {
```
northeastprince: Actually, if we're using BEM, probably makes more sense to do something like this:

```suggestion
.transaction__icon--when-selected {
```
or
```suggestion
.transaction__icon--selected {
```
sorry @sampoder 😆

app/assets/stylesheets/components/_transactions.scss
```diff
@@ -119,3 +119,23 @@ html[data-dark='true'] {
...
+  display: table-cell!important;
+}
+
+.selected__transaction > .transaction__icon {
+  display: none;
+}
+
+.transaction__icon {
+  padding-left: 1rem; 
+}
```
northeastprince:
```suggestion
}

```

app/assets/stylesheets/components/_transactions.scss
sampoder: All good!

app/assets/stylesheets/components/_transactions.scss
northeastprince: Not sure why the newline isn't being enforced by linters, but whatevs...

app/controllers/tags_controller.rb
```diff
@@ -10,8 +10,16 @@ def create
...
+    hcb_code_ids = if params[:hcb_code_id]
+                     [params[:hcb_code_id]]
+                   elsif params[:hcb_code_ids]
+                     params[:hcb_code_ids].split(",")
+                   else
+                     []
+                   end
+
+    hcb_code_ids.each do |hcb_code_id|
+      hcb_code = HcbCode.find(hcb_code_id)
```
northeastprince: Rails has built in support for a param as an array, so it's probably more straightforward if we just use that. [See how we use it for allowing people to manage locations they've subscribed to for new hackathons!](https://github.com/hackclub/hackathons-backend/blob/af17b783e21009f6b8e87b0ab739bbe2f3e6d2d0/app/controllers/hackathons/subscriptions/bulk_controller.rb)

app/javascript/controllers/transactions_controller.js
```diff
@@ -0,0 +1,33 @@
...
+					this.taggerTarget.classList.add("display-none")
+				}
+			}
+			this.updateFormFieldValue()
+		}
+	}
+	
+	updateFormFieldValue() {
+		this.formFieldTargets.map(formField => {
+			formField.value = this.selectedHCBCodes.join(",");
```
sampoder:
```suggestion
			formField.value = this.selectedHCBCodes;
```

@northeastprince, does that mean I can just do this  ^ and drop the .split(",") stuff in the controller

northeastprince: Not sure how it'd work in JS, but you just take `hcb_code_ids` as a parameter and if you're using Rails URL helpers it should work automatically (just make it so that you don't use a singular param, even if it's just one transaction) - I believe the format in URLs would be `hcb_code_ids[]=1,2,3` and you can do it seperately too and it'll just append (`?hcb_code_ids[]=1&hcb_code_ids[]=2`. Then, for the query, you do `HcbCode.where(id: hcb_code_ids)`. Lemme know if that doesn't work!

sampoder: hcb_code_ids[] doesn't seem to trigger the conversion when it's passed through in a POST request

app/views/events/show.html.erb
```diff
@@ -19,176 +19,212 @@
...
           <% @event.tags.each do |tag| %>
             <div class="flex items-center">
-              <%= link_to tag.label, { tag: tag.label, q: params[:q] }.compact, class: "flex-auto menu__action" %>
-              <% if organizer_signed_in? %>
-                <%= button_to event_tag_path(@event, tag), class: "menu__action", method: :delete, title: "Delete this tag", form: { "data-turbo" => "true", "data-turbo-confirm" => tag.removal_confirmation_message } do %>
-                  <%= inline_icon "delete", size: 20, style: "margin: 0" %>
-                <% end %>
+              <%= form_with url: event_tags_path(@event), data: { turbo: false }, class: "flex-auto" do |form| %>
+                <button class="menu__action" title="Apply this tag" type="submit"><%= tag.label %></button>
+                <%= form.hidden_field :hcb_code_ids, "data-transactions-target": "formField" %>
```
garyhtou:
```suggestion
                <%= form.hidden_field :hcb_code_ids[], "data-transactions-target": "formField" %>
```

app/controllers/tags_controller.rb
```diff
@@ -10,8 +10,16 @@ def create
...
 
-    if params[:hcb_code_id]
-      hcb_code = HcbCode.find(params[:hcb_code_id])
+    hcb_code_ids = if params[:hcb_code_id]
+                     [params[:hcb_code_id]]
+                   elsif params[:hcb_code_ids]
+                     params[:hcb_code_ids].split(",")
+                   else
+                     []
+                   end
```
garyhtou:
```suggestion
    hcb_code_ids = params[:hcb_code_id] || params[:hcb_code_ids]&.split(",") || []
```

app/controllers/tags_controller.rb
garyhtou:
```suggestion
    HcbCode.where(id: hcb_code_id).find_each do |hcb_code|
```

app/javascript/controllers/transactions_controller.js
garyhtou, line 1: Use spaces instead of tabs

app/javascript/controllers/transactions_controller.js
```diff
@@ -0,0 +1,33 @@
...
+	static targets = ["transaction", "tagger", "formField"];
+	selectedHCBCodes = [];
+
+	trackClick(event) {
+		const clickedElement = event.currentTarget;
+		if (event.metaKey || event.ctrlKey) {
+			if(!this.selectedHCBCodes.includes(clickedElement.id)){
+				this.selectedHCBCodes.push(clickedElement.id);
+				clickedElement.classList.add("selected__transaction")
+				if(this.selectedHCBCodes.length > 0){
```
garyhtou: condition not needed

app/javascript/controllers/transactions_controller.js
```diff
@@ -0,0 +1,33 @@
...
+export default class extends Controller {
+	static targets = ["transaction", "tagger", "formField"];
+	selectedHCBCodes = [];
+
+	trackClick(event) {
+		const clickedElement = event.currentTarget;
+		if (event.metaKey || event.ctrlKey) {
+			if(!this.selectedHCBCodes.includes(clickedElement.id)){
+				this.selectedHCBCodes.push(clickedElement.id);
+				clickedElement.classList.add("selected__transaction")
```
garyhtou: move this into a `selectHcbCode`

app/javascript/controllers/transactions_controller.js
```diff
@@ -0,0 +1,33 @@
...
+				if(this.selectedHCBCodes.length > 0){
+					this.taggerTarget.classList.remove("display-none")
+				}
+			}
+			else{
+				this.selectedHCBCodes = this.selectedHCBCodes.filter(x => x != clickedElement.id);
+				clickedElement.classList.remove("selected__transaction")
```
garyhtou: move this into a `unselectHcbCode`

app/javascript/controllers/transactions_controller.js
```diff
@@ -0,0 +1,33 @@
...
+				if(this.selectedHCBCodes.length == 0){
+					this.taggerTarget.classList.add("display-none")
+				}
+			}
+			this.updateFormFieldValue()
```
garyhtou: Add a `updateTaggerVisbility` method for:

```javascript
				if(this.selectedHCBCodes.length == 0){
					this.taggerTarget.classList.add("display-none")
				} else {
					this.taggerTarget.classList.remove("display-none")
        }
```

```suggestion
      this.updateTaggerVisbility()
			this.updateFormFieldValue()
```

app/javascript/controllers/transactions_controller.js
```diff
@@ -0,0 +1,33 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
+	static targets = ["transaction", "tagger", "formField"];
+	selectedHCBCodes = [];
+
+	trackClick(event) {
```
garyhtou: rename this to `multiselect`

app/views/canonical_transactions/_canonical_transaction.html.erb
```diff
@@ -13,8 +15,12 @@
       drop->file-drop#drop
       dragenter->file-drop#dragenter
       dragleave->file-drop#dragleave
+      click->transactions#trackClick
     "
   <% end %>>
+    <td class="transaction__icon tooltipped tooltipped--e transaction__icon--selected">
+      <%= inline_icon "checkmark" %>
+    </td>
```
garyhtou: Do we need this for pending transactions?

garyhtou (changes requested):

app/views/canonical_transactions/_canonical_transaction.html.erb
sampoder: I think we have it in there

PR #4202, opened 2023-08-29
Upgrade local postgres to version 12 to match production update

albertchae:
This may require running `docker volume rm bank_pg-data` to avoid
incompatibility errors with a postgres 11 data directory vs 12.

I ran the test suite against this local postgres 12 and everything
passed.


maxwofford (dismissed):

garyhtou:
@albertchae is this good to merge?

albertchae:
@garyhtou should be good to merge but I think this will blow away developers' local databases, if that's OK

garyhtou:
I just remembered we've also got GitHub codespaces (devcontainers)
https://github.com/hackclub/bank/blob/0fdb7d4f3adbca5c27006f548a867c93ef8d7890/.devcontainer/docker-compose.yml#L25

albertchae:
Closing this in favor of https://github.com/hackclub/hcb/pull/5731, which allows us to run both 11 and 12 to make the upgrade path smoother for developers

PR #4204, opened 2023-08-29
Update staging databases to restore from new production database

garyhtou:
Staging databases are already running version 12
https://github.com/hackclub/bank/blob/9fbc17306ddcd0c1a32713eb66f797463743458d/app.json#L45-L49


albertchae:
Good catch, thanks!

PR #4206, opened 2023-08-30
Deprecate `StripeAuthorization` jobs

sampoder:
@maxwofford alongside deleting the file and the only call made to it (from what can I see), is there anything else needed to fix #4039?


cjdenio:
We also have to remove references to the `StripeAuthorization` model, for example, it's still used to calculate the amount spent on a card: https://github.com/hackclub/hcb/blob/105bb2fbf9bc39422a7ef180cbb020e0ba0ebeae/app/models/stripe_card.rb#L136-L138

sampoder:
gotcha! on it!

cjdenio (approved):
Thanks! ~~Let's see what this breaks~~

sampoder:
hell yeah

PR #4209, opened 2023-08-30
Rename to HCB

cjdenio:
Things that MUST be done before this is merged:

- [x] Changelog must be moved to `changelog.hcb.hackclub.com`
- [x] `bank-parse.hackclub.com` must be moved to `hcb.hackclub.com`
- [x] All instances of the legacy OG image must be removed
- [ ] Brand guidelines (https://hcb.hackclub.com/branding) needs to be rewritten for HCB


garyhtou:
Ideally, old routes/domains should redirect to new routes to prevent breaking functionality

cjdenio:
Absolutely! I don't think this renames any routes but we're going to redirect `bank.hackclub.com` to `hcb.hackclub.com`

PR #4213, opened 2023-09-01
Add Random HCB Acronyms

sampoder:
Heavily inspired by [`npm-expansions`](https://github.com/npm/npm-expansions/). I guess we'll try to keep on top of the PRs (https://github.com/npm/npm-expansions/issues/3697). 

Screenshot 2023-08-31 at 10 30 25 PM


app/services/static_page_service/index.rb
```diff
@@ -56,6 +58,21 @@ def disbursements_count
       Disbursement.pending.count
     end
 
+    def acronym
+      client = Faraday.new do |builder|
+        builder.use :http_cache, store: Rails.cache
```
northeastprince: I believe this is the first time we're actually caching something in Bank 😆, so we might need to consider finally separating Redis for Sidekiq from everything else.

northeastprince:
Also just throwing it out there that we could just have an `HcbAcronym` and not have it in a separate repo as it adds way more complexity than needed imo.

sampoder:
I hear that but I think the fun of it is that people can add their own expansions

northeastprince:
I'm fine with this as long as there's some way that we're not holding up a request by fetching this and cache storage is kept under control, potentially mitigating the need for a second instance (at least for now). Maybe something like this:

```ruby
class HcbAcronym
  attr_reader :expansion

  def initialize
    @expansion = Rails.cache.read("hcb_acronym_expansion") || "Hack Club Buckaroos"
  end
end
```

```ruby
class RefreshHcbAcronymExpansionJob < ApplicationJob
  def perform
    response = Faraday.get("https://raw.githubusercontent.com/hackclub/hcb-expansions/main/phrases.txt")

    if response.success?
      hcb_acronym_expansions = response.body.split("\n")

      Rails.cache.write "hcb_acronym_expansion", hcb_acronym_expansions.sample
    end
  end
end
```

sampoder:
Agreed @northeastprince, I'll change it up tonight!

sampoder:
@northeastprince thoughts on this now?

app/controllers/static_pages_controller.rb
```diff
@@ -16,6 +16,9 @@ def index
       @invites = @service.invites
 
       @show_event_reorder_tip = current_user.organizer_positions.where.not(sort_index: nil).none?
+
+      @phrase = Rails.cache.read("hcb_acronym_expansion").sample || "Hack Club Buckaroos"
```
northeastprince:
```suggestion
      @hcb_expansion = Rails.cache.read("hcb_acronym_expansion").sample || "Hack Club Buckaroos"
```

northeastprince (approved):

app/views/static_pages/index.html.erb
```diff
@@ -5,9 +5,14 @@
...
 <h1 class="heading items-center md-items-baseline justify-center md-justify-between center md-left-align w-auto">
   <div>
-    <span class="h3 caps secondary block">Welcome to</span>
-    <span class="primary <%= "nowrap" unless fall? %> <%= "warning" if fall? %>">
-      HCB
+    <a
+      class="h4 caps muted block text-decoration-none"
+      href="https://github.com/hackclub/hcb-expansions/tree/main#hcb-what-does-it-stand-for"
+      target="_blank">
+      <%= @phrase %>
```
sampoder:
```suggestion
      <%= @hcb_expansion %>
```

app/jobs/refresh_hcb_acronym_expansion_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class RefreshHcbAcronymExpansionJob < ApplicationJob
+  def perform
+    response = Faraday.get("https://raw.githubusercontent.com/hackclub/hcb-expansions/main/phrases.txt")
+
+    if response.success?
+      hcb_acronym_expansions = response.body.split("\n")
+
+      Rails.cache.write "hcb_acronym_expansion", hcb_acronym_expansions
```
northeastprince: If you wanna have it be random for every request, then it's fine to save the whole array and sample from that, but I would reflect that in the cache key:


```suggestion
      Rails.cache.write "hcb_acronym_expansions", hcb_acronym_expansions
```

sampoder: sounds like a plan

app/controllers/static_pages_controller.rb
```diff
@@ -16,6 +16,9 @@ def index
       @invites = @service.invites
 
       @show_event_reorder_tip = current_user.organizer_positions.where.not(sort_index: nil).none?
+
+      @hcb_expansion = Rails.cache.read("hcb_acronym_expansion").sample || "Hack Club Buckaroos"
```
sampoder:
```suggestion
      @hcb_expansion = Rails.cache.read("hcb_acronym_expansions").sample || "Hack Club Buckaroos"
```

config/schedule.yml
```diff
@@ -127,3 +127,8 @@ stripe_card_nightly_job:
   cron: "0 0 1 * *" # run first day of each month
   class: "DisbursementJob::Distribute128CollectiveFund"
   queue: "default"
+
+refresh_hcb_acronym_expansion_job:
+  cron: "*/5 * * * *" # run every 5 minutes
+  class: "RefreshHcbAcronymExpansionJob"
+  queue: "default"
```
northeastprince:
```suggestion
  queue: "default"

```

EOF

app/jobs/refresh_hcb_acronym_expansion_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class RefreshHcbAcronymExpansionJob < ApplicationJob
```
northeastprince:
```suggestion
class RefreshHcbAcronymExpansionsJob < ApplicationJob
```

app/controllers/static_pages_controller.rb
```diff
@@ -16,6 +16,9 @@ def index
       @invites = @service.invites
 
       @show_event_reorder_tip = current_user.organizer_positions.where.not(sort_index: nil).none?
+
+      @hcb_expansion = Rails.cache.read("hcb_acronym_expansions").sample || "Hack Club Buckaroos"
```
northeastprince:
```suggestion
      @hcb_expansion = Rails.cache.read("hcb_acronym_expansions")&.sample || "Hack Club Buckaroos"
```
OR clause won't work without this.

sampoder:
@maxwofford got the 👍 from Matt, cool if I merge?

cjdenio (approved):

PR #4214, opened 2023-09-01
Update Brand Guidelines for "HCB"

sampoder:
I guess we're no longer a bank... rip.


rluodev (approved):
looks good to me, LGTM!

PR #4216, opened 2023-09-01
Add "Hack Club" to hcb.hackclub.com's homepage

sampoder:
Now when someone goes to `hcb.hackclub.com` they don't see Hack Club any more. Let's show them Hack Club!

(and yes in the commit I meant hcb.hackclub.com...)


maxwofford:
I like the new look!

PR #4217, opened 2023-09-02
Remove Badges That Span Two Lines On Admin Cards

sampoder:
I guess this fixes this:

Screenshot 2023-09-02 at 1 26 06 AM


YodaLightsabr (approved):
Thanks for the fix! Looks good!

PR #4208, opened 2023-08-30
Add Badges to Non-React Admin Cards

sampoder:
Previously, we had a mix of numbers in brackets and numbers in badges. Now everyone gets a badge!

PR #4219, opened 2023-09-02
Fix #4218

sampoder:
This fixes a logic mistake I made in #4208 and add some CSS classes we were referencing but weren't defined in light mode


YodaLightsabr (approved):
Thanks for the fix, looks good!

PR #4220, opened 2023-09-02
Restrict Login Codes To Their Origin Browser

sampoder:
This will fix #3438. How it does it is that it generates a unique token when some signs in (this is the "browser token"), then it stores that token locally as a cookie in the browser as well as in the DB, then, when someone inputs a code it checks that the two match.


app/models/login_code.rb
```diff
@@ -23,6 +25,8 @@
 #  fk_rails_...  (user_id => users.id)
 #
 class LoginCode < ApplicationRecord
+  has_encrypted :browser_token
+  blind_index :browser_token
```
northeastprince: Why does this need to be encrypted if it's already random?

app/controllers/users_controller.rb
```diff
@@ -64,7 +64,10 @@ def login_code
 
     initialize_sms_params
 
-    resp = LoginCodeService::Request.new(email: @email, sms: @use_sms_auth, ip_address: request.ip, user_agent: request.user_agent).run
+    @browser_token = SecureRandom.urlsafe_base64
+    cookies.encrypted[:browser_token] = { value: @browser_token, expires: Time.current + 15.minutes }
```
northeastprince: Can this just be `signed`?

northeastprince:
Before we implement this, I'm wondering if we should consider implementing and switching to login links instead. I think it's more secure anyway and easier to use.

app/models/login_code.rb
sampoder: Good point, I was just copying the session_token logic

app/controllers/users_controller.rb
sampoder: Wdym?

northeastprince: A signed cookie instead of an encrypted one.

sampoder: we can do that!

sampoder:
```suggestion
    cookies.signed[:browser_token] = { value: @browser_token, expires: Time.current + 15.minutes }
```

app/controllers/users_controller.rb
```diff
@@ -154,7 +157,8 @@ def exchange_login_code
     user = UserService::ExchangeLoginCodeForUser.new(
       user_id: params[:user_id],
       login_code: params[:login_code],
-      sms: params[:sms]
+      sms: params[:sms],
+      browser_token: cookies.encrypted[:browser_token]
```
sampoder:
```suggestion
      browser_token: cookies.signed[:browser_token]
```

YodaLightsabr:
@sampoder: what's the status of this PR? what do we have to do to move forward with it? It'd be great to get this out as one of our small wins in the next 5 weeks.

sampoder:
Other than the conflicts, this PR is good to review from my end

sampoder:
I now seem to be getting that error for all of those case, @YodaLightsabr if you're still not getting that error, what does your setup look like?

app/models/login_code.rb
```diff
@@ -23,6 +25,8 @@
 #  fk_rails_...  (user_id => users.id)
 #
 class LoginCode < ApplicationRecord
+  has_encrypted :browser_token
```
garyhtou: maybe we can use https://api.rubyonrails.org/classes/ActiveRecord/SecureToken/ClassMethods.html

or, moving the token generation logic into the model using a `before_validation`

garyhtou (changes requested):

sampoder:
check out #4405 which does this using the fancy rails method!

PR #4226, opened 2023-09-05
Rename 'HCB fee' to 'fiscal sponsorship fee'

zachlatta:
I think this makes the fee feel better from a UX perspective, and less HCB-specific.

I haven't tested this - so please take a look to make sure this won't break anything. The only thing I would check is to make sure that the change in `ui.js` doesn't extend past the page.

I went through to make sure all instances of `HCB fee` (case insensitive) are renamed.


app/views/events/_pending_fee_transaction.html.erb
```diff
@@ -1,9 +1,9 @@
 <tr class="transaction--negative muted">
-  <td class="transaction__icon tooltipped tooltipped--e" aria-label="HCB fee payment">
+  <td class="transaction__icon tooltipped tooltipped--e" aria-label="Fiscal sponsorship payment">
     <%= inline_icon "minus-fill" %>
   </td>
   <td><%= Date.today.strftime("%Y-%m-%d") %></td>
-  <td class="transaction__memo">PENDING: HCB FEE</td>
+  <td class="transaction__memo">PENDING: FISCAL SPONSORSHIP</td>
```
recursiveforte: Could this be `FISCAL SPONSORSHIP FEE` instead of just `FISCAL SPONSORSHIP`?

zachlatta: I was thinking about that, and prefer `FISCAL SPONSORSHIP`. The word `FEE` is just so negative, and the fact that it's a debit instead of a credit makes it clear it's a fee IMO.

sampoder (approved):
tested locally, seems to work well for me!

PR #4227, opened 2023-09-05
Vertically center logo when org name overflows on homepage cards

YodaLightsabr:
Vertically center logo when org name overflows on homepage cards

This
image

Becomes this
image


sampoder (approved):
Looks good to me! LGTM!

PR #4228, opened 2023-09-06
Small Session Page Redesign

sampoder:
The main changes were capitalising Current Session, moving that button down to prevent the overflow, and only showing the compass icon and dot if the device info is present.


maxwofford (approved):
Looks great! Go for merge on this. And for the future, please prefer updating with rebase over updating with merge!

PR #4231, opened 2023-09-06
Don't flip the unflippables

sampoder:
Fixes #3816, I think. One thing to check is that I'm not missing any potential Flipper tags we'd want people to flip.


app/controllers/users_controller.rb
```diff
@@ -215,22 +215,26 @@ def receipt_report
...
   }.freeze
 
   def enable_feature
     @user = current_user
     @feature = params[:feature]
     authorize @user
-    if Flipper.enable_actor(@feature, @user)
-      confetti!(emojis: FEATURE_CONFETTI_EMOJIS[@feature.to_sym])
-      flash[:success] = "Opted into beta"
+    if FEATURES.key?(@feature) || @user.admin?
```
```suggestion
    if FEATURES.key?(@feature.to_sym) || @user.admin?
```

app/controllers/users_controller.rb
```diff
@@ -239,10 +243,14 @@ def disable_feature
     @user = current_user
     @feature = params[:feature]
     authorize @user
-    if Flipper.disable_actor(@feature, @user)
-      flash[:success] = "Opted out of beta"
+    if FEATURES.key?(@feature) || @user.admin?
```
```suggestion
    if FEATURES.key?(@feature.to_sym) || @user.admin?
```

garyhtou (approved):
good to go! great work here

PR #4232, opened 2023-09-07
Add Pinned Admin Cards

devramsean0:
This Pull request closes #4169


sampoder (commented):
Looks like the ESLint check is failing because of the `id` prop, https://github.com/hackclub/hcb/commit/9aeed6642934cc4232ba75a0d9e390778da61e9e seems like a good example of an alternative to using `id`.

sampoder:
Thoughts on using a cookie to store the information and then making the rearrangements server-side?

devramsean0:
I considered it, but cookies expire after a while. And I believe they have a smaller size limit

devramsean0:
Progress Update:
 - Adding and removing cards works for airtable cards
 - Loading those cards on page load doesn't work (I think react is populating the IDs after stimulus connects)

devramsean0:
I can't get this to work reliably :/ I think this would be far easier to implement without LocalStorage / Stimulus
RN I need the ID property to do the ID stuff. But that makes it fail the lint.
I also can't get it to rerender the cards on page load. I'm pushing the code I already have but I would appreciate some help

devramsean0:
I still don't know how I'm gonna deal with the no ID eslint rule. There is no other way for the react based cards that will allow them to work.

northeastprince:
I've been [sleeping on it](https://37signals.com/34), and I'm not sure this is worth implementing. Granted, I don't use admin cards that much, but from what I've seen, they're basically links to different things both inside and outside of HCB. If one needs a few specific things more at the ready than scrolling down a bit, I think bookmarking it would be a better option.

devramsean0:
Yeah, that's fair. but it's so close to fully working lol
(I just need to work out how the ERB / Ruby cards work)

devramsean0:
This took far longer than I'd like to admit, and I still have to work out how to add the Pin icon to the non-react cards (and patch the controller so they work) but I got IDs being given to non-react cards.

devramsean0:
Closing as follow-up for meeting with Gary. I will move this to a different branch which is less broken
Change controllers to just generate params and redirect with params to the admin page
Have ERB / Rails do all the rendering and ignore React until necessary
Upon status changes regenerate query parameters and reload with new

PR #4236, opened 2023-09-10
Fix Money Parsing Error on Admin Ledger Page

sampoder:
Fixes #3890 I think


app/controllers/admin_controller.rb
```diff
@@ -384,9 +384,8 @@ def ledger
     end
 
     if @q
-      if @q.to_f != 0.0
+      if @q.to_f != 0.0 && (money = Monetize.parse(@q) rescue nil)
```
```suggestion
      if @q.match /\A\d+(\.\d{1,2})?\z/
```

sampoder:
oh that's smart! nice!

From Sam

PR #4240, opened 2023-09-11
Allow Dark Mode To Be Forced Used A Query Param

sampoder:
For this use case (in #hcb):

Screenshot 2023-09-11 at 4 41 56 PM

Seems like a nice idea!


cjdenio:
Two things:

- Could this be changed to use the [`@dark` instance variable](https://github.com/hackclub/hcb/blob/6efde57c346cac6c4f5e48cd9e6febd780a931bf/app/views/layouts/application.html.erb#L2) instead of JavaScript? (just set `@dark = true` or `@dark = false` in a controller)
- I don't think this would retain theme state as you click through the donation page (the query parameter would be removed as soon as you clicked "Continue")

sampoder:
Good idea! At the moment, it retains it but through local storage which admittedly isn't great

sampoder:
@cjdenio thoughts about doing it this way? it's persistent and use `@dark` but it also uses cookies

garyhtou (approved):

PR #4243, opened 2023-09-13
Validate Phone Number

sampoder:
Fixes #3982


app/controllers/users_controller.rb
```diff
@@ -312,8 +312,10 @@ def update
 
     if @user.update(user_params)
       confetti! if !@user.seasonal_themes_enabled_before_last_save && @user.seasonal_themes_enabled? # confetti if the user enables seasonal themes
-
-      if @user.full_name_before_last_save.blank?
+      if !((user_params[:phone_number] =~ /\A\+[1-9]\d{1,14}\z/)) && user_params[:phone_number]
```
cjdenio: Could you move this to a [model-level validation](https://guides.rubyonrails.org/active_record_validations.html)?

cjdenio (changes requested):

sampoder:
oh awesome @cjdenio, TIL! just updated, let me know if i did smth wrong :D

sampoder:
Seems like I've messed something up..

cjdenio:
@sampoder it looks like we're already using [this gem](https://github.com/daddyz/phonelib) for phone validation— I wonder if it's configured incorrectly

northeastprince:
Seems to be validating phone numbers correctly already on my end.

cjdenio:
Has that changed since [this comment](https://github.com/hackclub/hcb/issues/3982#issuecomment-1593549955)?

northeastprince:
Hmm, maybe the validations weren't done for some reason on that one or we need to format numbers in E164?

sampoder:
@northeastprince what was the phone number you used that broke it?

sampoder:
I guess I'll close this for now

PR #4248, opened 2023-09-14
Post-Card Grant Surveys

sampoder:
The current use case for this is working out club's favourite pizza toppings!


app/models/card_grant.rb
```diff
@@ -4,19 +4,20 @@
...
-#  sent_by_id      :bigint           not null
-#  stripe_card_id  :bigint
-#  subledger_id    :bigint
-#  user_id         :bigint           not null
+#  id                        :bigint           not null, primary key
+#  amount_cents              :integer
+#  category_lock             :string
+#  email                     :string           not null
+#  merchant_lock             :string
+#  post_grant_survey_answers :jsonb
```
northeastprince: I don't love just sticking the data in a JSON blob. Could this be properly modeled out, maybe in a more general way so we can use it for other feedback things?

sampoder: it's difficult though cause if we want to have different questions for every card grant, that's a lot of columns / tables to add

northeastprince: Why is it a lot of columns and tables to add?

```ruby
class Survey < ApplicationRecord
  validates :name, presence: true

  has_many :questions
  has_many :responses
end

class Survey::Question < ApplicationRecord
  belongs_to :survey

  validates :content, presence: true
end

class Survey::Response < ApplicationRecord
  belongs_to :survey
  belongs_to :respondent, class_name: "User"

  has_many :answers
end

class Survey::Answer < ApplicationRecord
  belongs_to :response
  belongs_to :question

  validates :content, presence: true
end
```

sampoder:
This has gone stale. Let's not let it die a slow and painful death... if anyone ever wants to, we could pick this back up. RIP.

PR #4266, opened 2023-09-29
Update Stripe Phone Number On User Edit

sampoder:
In theory this fixes #4265 but honestly I'm very skeptical


app/controllers/stripe_cardholders_controller.rb
```diff
@@ -53,7 +53,9 @@ def stripe_cardholder_params
       :stripe_billing_address_city,
       :stripe_billing_address_state,
       :stripe_billing_address_postal_code,
-      :stripe_billing_address_country
+      :stripe_billing_address_country,
+      :stripe_billing_address_country,
+      :phone_number
```
sampoder:
```suggestion
      :stripe_phone_number
```

app/controllers/stripe_cardholders_controller.rb
```diff
@@ -53,7 +53,9 @@ def stripe_cardholder_params
       :stripe_billing_address_city,
       :stripe_billing_address_state,
       :stripe_billing_address_postal_code,
-      :stripe_billing_address_country
+      :stripe_billing_address_country,
+      :stripe_billing_address_country,
+      :stripe_phone_number
```
sampoder:
```suggestion
      :phone_number
```

sampoder:
hell yeah third times a charm

app/models/stripe_cardholder.rb
```diff
@@ -51,6 +51,7 @@ class StripeCardholder < ApplicationRecord
   alias_attribute :address_state, :stripe_billing_address_state
   alias_attribute :address_country, :stripe_billing_address_country
   alias_attribute :address_postal_code, :stripe_billing_address_postal_code
+  alias_attribute :phone_number, :stripe_phone_number
```
northeastprince: Why are we aliasing all of these?

sampoder: not sure about the others but in this case it was just cause it's `phone_number` on the form and `stripe_phone_number` in the model

northeastprince: Isn't that implied when you're updating a `StripeCardholder`?

sampoder:
the new implementation:

image

PR #4269, opened 2023-09-29
Fix #4264

sampoder:
hopefully not too hacky! fixes #4264


bin/dev
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+bundle install --gemfile Gemfile
+
```
northeastprince:
```suggestion
```
Unrelated

app/assets/stylesheets/_utilities.scss
```diff
@@ -33,6 +33,10 @@
   margin-right: auto;
 }
 
+.mt6 {
```
garyhtou: can we move this class here?
https://github.com/hackclub/bank/blob/379cd0567728fbf61a6038613b99eb3e95d1ff69/app/assets/stylesheets/space.css#L36-L36

garyhtou (changes requested):

app/assets/stylesheets/_utilities.scss
YodaLightsabr: to add on to what gary said, could we use `96px` instead of `6rem` to stay consistent with the other units used in `space.css`?

app/views/donations/_donation_form.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= turbo_frame_tag :donation do %>
   <% if Flipper.enabled?(:recurring_donations_2023_02_07, @event) %>
-    <ul class="tab-container max-width-1 mx-auto">
+    <ul class="tab-container max-width-1 mx-auto <%= "mt6" if @event.donation_header_image.attached? && !@event.donation_page_message.present? %>">
```
sampoder:
```suggestion
    <ul class="tab-container max-width-1 mx-auto <%= "mt7" if @event.donation_header_image.attached? && !@event.donation_page_message.present? %>">
```

we actually want to jump to 7! (and skip 6)

app/assets/stylesheets/_utilities.scss
sampoder: indeed we should

sampoder:
(i'm gonna keep mt6 is there coz its weird to have mt5 and mt7 but no mt6)

app/assets/stylesheets/_utilities.scss
YodaLightsabr: did you mean [indeed.com](https://indeed.com)?

garyhtou (approved):

PR #4277, opened 2023-10-04
✨ Add Vermont-based org tag

malted:


northeastprince:
I'm not sure if we should add this as a category, cause then orgs can only be marked as Vermont-based and nothing else. That's the fundamental flaw of this enum - personally I think something like [a tagging system](https://github.com/hackclub/hackathons-backend/blob/main/app/models/concerns/taggable.rb) would work much better.

northeastprince:
Ahh, it just sent now, and you already did it 💀

maxwofford (approved):
Thanks for adding this!

PR #4296, opened 2023-10-13
Default to California for Check Address

sampoder:
Idk... feels weird to have San Francisco in Alabama:

Screenshot 2023-10-12 at 5 20 20 PM


sampoder:
Is there a better way to do this so it doesn't look like California is selected? Make it a placeholder value?

sampoder:
oh that's how you do it, sick!!

cjdenio:
Thanks for flagging this!

PR #4307, opened 2023-10-17
Interface for Updating Card Grant Merchant / Category Lock

sampoder:
I think this does what #4292 describes; however, it might make more sense to add two fields to every event that are considered defaults for every card grant associated to that event.


sampoder:
Closing in favour of #4308

PR #4308, opened 2023-10-17
Event-Wide Card Grant Locks

sampoder:
Alternative implementation to #4307, would close #4292.


maxwofford:
@sampoder thanks for making this! I prefer this implementation over https://github.com/hackclub/hcb/pull/4307.

We want this to be something separate from the events codebase, so maybe these settings can go in a `CardGrantSetting` model that gets attached to `Event`?

sampoder:
Oh interesting, would the interface look like for linking an event to a CardGrantSetting or would each event by default have a CardGrantSetting?

maxwofford:
I don't think we want an interface for that because we don't officially support it outside of onboard and pizza party. That said, for those 2 events we want a way for regular users on those events to have control over this sort of thing

sampoder:
Sweet, I can make some sort of model then

db/migrate/20231018175019_create_card_grant_settings.rb
```diff
@@ -0,0 +1,9 @@
+class CreateCardGrantSettings < ActiveRecord::Migration[7.0]
+  def change
+    create_table :card_grant_settings do |t|
+      t.string :merchant_lock
+      t.string :category_lock
+      t.references :event, null: false, foreign_key: true
```
northeastprince:
```suggestion
      t.belongs_to :event, null: false, foreign_key: true
      t.string :merchant_lock
      t.string :category_lock
```
reflects the relationship a bit better

sampoder: lets do it

app/controllers/events_controller.rb
```diff
@@ -176,6 +176,24 @@ def update
...
+    end
+
+    fixed_event_params.delete(:card_grants_category_lock)
+
+    if fixed_event_params[:card_grant_setting_invite_message].present?
+      @event.card_grant_setting.update(invite_message: fixed_event_params[:card_grant_setting_invite_message])
+    end
+
+    fixed_event_params.delete(:card_grant_setting_invite_message)
+
```
northeastprince: Adding conditionals for all of these seems like a bit of a smell. Could we use nested attributes under the event?

sampoder: @northeastprince just had a shot at using nested attributes, can you check if I did it correctly?

northeastprince: LGTM!

app/controllers/events_controller.rb
```diff
@@ -646,6 +646,12 @@ def event_params
       :website,
       :background_image,
       :stripe_card_shipping_type,
+      card_grant_setting_attributes: [
+        :id,
```
northeastprince: Why is this necessary? An event only `has_one :cart_grant_setting`.

maxwofford: We found that without including this it'd try to create a new CardGrantSetting object each time. I think this helps clarify that the update is just patching records.

northeastprince: Ah, you have to set [the `update_only` option](https://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html#:~:text=without%20providing%20the%20id).

maxwofford: Nice! Thanks for digging that up

sampoder: That's sick! Nice find

app/jobs/one_time_jobs/seed_card_grant_setting.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module OneTimeJobs
+  class SeedCardGrantSetting < ApplicationJob
+    def perform
+      onboard = Event.find_by_slug("onboard")
+      process_event onboard unless onboard.card_grant_setting.present?
+      onboard.card_grant_setting.invite_message = <<~MD
```
cjdenio: Could these just be set manually after the feature's available?

app/views/card_grant_mailer/card_grant_notification.html.erb
```diff
@@ -1,7 +1,13 @@
 You've been sent a <%= @card_grant.amount.format %> grant from <%= @card_grant.event.name %>!
 
 <br>
-<br>
+<% unless @custom_invite_message.blank? %>
```
cjdenio:
```suggestion
<% if @custom_invite_message.present? %>
```

app/jobs/one_time_jobs/seed_card_grant_setting.rb
maxwofford: Yep! I'm just dumping them in here as an optional tool for after deployment

app/models/card_grant.rb
```diff
@@ -130,6 +132,14 @@ def create_stripe_card(session)
     save!
   end
 
+  def combined_merchant_lock
+    setting&.merchant_lock&.+ merchant_lock
+  end
+
+  def combined_category_lock
+    setting&.category_lock&.+ category_lock
+  end
```
cjdenio:
```suggestion
  def allowed_merchants
    setting&.merchant_lock&.+ merchant_lock
  end

  def allowed_categories
    setting&.category_lock&.+ category_lock
  end
```

app/views/card_grants/show.html.erb
```diff
@@ -48,6 +48,14 @@
             <strong>Billing address</strong>
             <%= render_address @card.cardholder %>
           </p>
+          <% admin_tools do %>
+            <p><strong>Merchant lock</strong>
+              <%= @card_grant.combined_merchant_lock.join(" ") %>
+            </p>
+            <p><strong>Category lock</strong>
+              <%= @card_grant.combined_category_lock.join(" ") %>
+            </p>
```
cjdenio:
```suggestion
            <p><strong>Allowed merchants</strong>
              <%= @card_grant.allowed_merchants.join(" ") %>
            </p>
            <p><strong>Allowed categories</strong>
              <%= @card_grant.allowed_categories.join(" ") %>
            </p>
```

sampoder: good catch!

app/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb
```diff
@@ -46,8 +46,13 @@ def decline_with_reason!(reason)
       def approve?
         return decline_with_reason!("inadequate_balance") if card.balance_available < amount_cents
 
-        return decline_with_reason!("merchant_not_allowed") if card.card_grant&.allowed_merchants.present? && card.card_grant.allowed_merchants.exclude?(auth[:merchant_data][:network_id]) # Handle merchant locks for restricted grants
-        return decline_with_reason!("merchant_not_allowed") if card.card_grant&.allowed_categories.present? && card.card_grant.allowed_categories.exclude?(auth[:merchant_data][:category])
+        if card&.card_grant&.allowed_categories&.present? && card.card_grant.allowed_categories.exclude?(auth[:merchant_data][:category])
+          return decline_with_reason!("category_not_allowed")
```
cjdenio:
```suggestion
          return decline_with_reason!("merchant_not_allowed")
```

Either this needs to stay `merchant_not_allowed` or the mailer needs to be updated:
https://github.com/hackclub/hcb/blob/44214ab6c3b11a011ac37c067b05c036dad65ed6/app/views/canonical_pending_transaction_mailer/notify_declined.html.erb#L13-L14

cjdenio (approved):

maxwofford:
[It's pizza time!](https://www.youtube.com/watch?v=z9KsJESRx38)

I just ran the one-time-job and it worked!

Screenshot 2023-10-30 at 17 44 51

PR #4309, opened 2023-10-18
Fix #4289

sampoder:
I think this should fix this but I'd appreciate a second pair of eyes here. 

Closes #4289.


app/controllers/application_controller.rb
```diff
@@ -15,7 +15,7 @@ class ApplicationController < ActionController::Base
   before_action :set_paper_trail_whodunnit
 
   # Redirect users to the onboarding page if they haven't completed it yet
-  before_action :redirect_to_onboarding
+  after_action :redirect_to_onboarding
```
cjdenio: This does need to be a `before_action`— `after_action`s can't redirect since they're run after the request is handled

PR #4318, opened 2023-10-19
✨ Introduce user permissions

malted:


app/models/user.rb
```diff
@@ -217,6 +218,10 @@ def onboarding?
     full_name.blank?
   end
 
+  def position(event)
```
northeastprince:
```suggestion
  def position_in(event)
```

app/models/user.rb
```diff
@@ -170,8 +170,9 @@ def initials
     words.any? ? words.map(&:first).join.upcase : name
   end
 
-  def pretty_phone_number
-    Phonelib.parse(self.phone_number).national
+  def pretty_phone_number(international: false)
+    parsed_number = Phonelib.parse(self.phone_number)
+    international ? parsed_number.international : parsed_number.national
   end
```
northeastprince: Unrelated
```suggestion
  def pretty_phone_number(international: false)
    Phonelib.parse(self.phone_number).national
  end
```

malted:
Replaced by #5201

PR #4323, opened 2023-10-21
Split every sidekiq queue into a new dyno

maxwofford:
This is to test and identify which sidekiq jobs are taking up so much memory– I intend to revert the change shortly after this goes into production if I can see which one causes run-away memory usage. I've left the default worker alive in the procfile and intend to just toggle it off/on in production instead of actually touching the config for that worker just in case.


maxwofford:
For posterity, this was the configured dyno size of prod before merging this:

Screenshot 2023-10-20 at 21 42 44

I'm documenting it here because it looks like I can edit these in the dashboard without updating a config file, which seems like a footgun.

maxwofford:
What I learned:

At first it spun up fine, but I didn't see all the queues I expected. I also would enqueue jobs and they wouldn't show up in the queue or be processed:

Screenshot 2023-10-20 at 22 10 28

Looking at the logs, this seems like an ssl issue with our redis provider:

Screenshot 2023-10-20 at 22 31 01

maxwofford:
This explains the errors we're seeing with enqueing and failing cron: if sidekiq sometimes fails to read/write from the db, all sorts of strange things would happen with jobs.

maxwofford:
Now that I've poked around with this I'm going to revert and head to bed soon. Seems like:

- This is a workable way to figure out what jobs are taking up the most memory
- The real underlying issue is communication between sidekiq and redis, not related to memory
- **This underlying issue is a critical issue because jobs are failing and failing to enqueue silently**

maxwofford:
I'm also going to go out on a limb and hypothesize the memory leak we have is due to the redis communication failure; I think something is done in redis to bottleneck jobs that isn't being done currently.

maxwofford:
When running `heroku logs -p heroku-redis -t` I'm not seeing any errors, so I think that's further proof that this is a connection error on sidekiq and redis isn't even aware of the failures happening

PR #4326, opened 2023-10-22
Clean up memory usage of jobs

garyhtou:
- Load records in batches instead of all at once
	- Using ActiveRecord's batching methods also prevents it from caching the queries, which takes up more memory
- Only select the required fields when it's necessary to load all records into memory
- Paginate stripe issuing authorization (transaction) import. It's currently creating an array of 30k items


app/services/stripe_card_service/nightly.rb
```diff
@@ -18,7 +18,7 @@ def run
     #
     #   Check out HcbCode#memo for more info on how this work.
     def rename_canonical_transaction
-      stripe_issuing_card_canonical_transactons_to_rename.each do |ct|
+      stripe_issuing_card_canonical_transactons_to_rename.find_each(batch_size: 100) do |ct|
         ct.update!(custom_memo: "💳 New user card fee")
       end
```
cjdenio:
```suggestion
      stripe_issuing_card_canonical_transactons_to_rename.update_all(custom_memo: "💳 New user card fee")
```

cjdenio (approved):
Thanks! I left one small performance improvement, but this looks great otherwise

app/services/stripe_card_service/nightly.rb
garyhtou: I'm somewhat weary of `update_all` because it doesn't run validations, but in this case, it should be fine

PR #4342, opened 2023-10-24
Disable State Input & Set to "NA" For Non-US Billing Addresses

sampoder:
https://hackclub.slack.com/archives/CN523HLKW/p1698168882359829


sampoder:
This actually doesn't make sense because it seems like Stripe only supports American and Canadian billing addresses.

PR #4343, opened 2023-10-24
Limit Billing Addresses to US & Canada + include Canadian Provinces & American / Canadian Territories

sampoder:
Context: https://hackclub.slack.com/archives/CN523HLKW/p1698168882359829.

It seems like Stripe only supports US & Canada:

Screenshot 2023-10-24 at 1 10 23 PM

fixes #4351 as well


app/views/users/edit_address.html.erb
```diff
@@ -97,7 +106,7 @@
 
         <div class="field">
           <%= form.label :country %>
-          <%= form.country_select :stripe_billing_address_country, nil, { selected: "US" }, { required: true } %>
+          <%= form.country_select :stripe_billing_address_country, nil, { selected: "US", only: ["US", "CA"] }, { required: true } %>
```
sampoder:
```suggestion
          <%= form.country_select :stripe_billing_address_country, nil, { only: ["US", "CA"] }, { required: true } %>
```

sampoder: fixes #4351

app/views/users/edit_address.html.erb
```diff
@@ -32,7 +32,7 @@
             <%= form.text_field :stripe_billing_address_city, placeholder: "West Hollywood", required: true %>
           </div>
           <div class="field flex-auto ml1">
-            <%= form.label :address_state, "State" %>
+            <%= form.label :address_state, "State / Province" %>
```
garyhtou: We have the `countries` gem installed, i'd suggest that we use that instead of hardcoding

here's a starter:
```ruby
ISO3166::Country.new("US").subdivisions.values.map {|s| [s.translations["en"], s.code] }
```

garyhtou (changes requested):

app/views/users/edit_address.html.erb
```diff
@@ -32,61 +32,11 @@
...
-                  ["Utah", "UT"],
-                  ["Vermont", "VT"],
-                  ["Virginia", "VA"],
-                  ["Washington", "WA"],
-                  ["West Virginia", "WV"],
-                  ["Wisconsin", "WI"],
-                  ["Wyoming", "WY"],
-                ] %>
+                  ISO3166::Country.new("US").subdivisions.values.map { |s| [s.translations["en"], s.code] },
+                  ISO3166::Country.new("CA").subdivisions.values.map { |s| [s.translations["en"], s.code] }
```
garyhtou: Move this to controller as `@states`

garyhtou (changes requested):

garyhtou (approved):
yay. we love canada

image

PR #4350, opened 2023-10-25
Remove React prop types in production JS build

cjdenio:
Shaves a whole 776 bytes off of our production build!


northeastprince:
What exactly is this and why don't we need it anymore?

cjdenio:
[PropTypes](https://legacy.reactjs.org/docs/typechecking-with-proptypes.html) is a type checker for React component props— it's used like this:
https://github.com/hackclub/hcb/blob/8a2a0490842be7d70f5895af149f43bdcc69a4dc/app/javascript/components/tour/TourStep.js#L184-L209

This PR configures Babel to remove these lines in production since they're primarily used by ESLint

PR #4352, opened 2023-10-25
Remove unused methods from `Event`

cjdenio:
This PR removes a bunch of unused methods/logic from `Event`, mostly around legacy fee calculation.


northeastprince:
Can we drop the v2 from naming then?

PR #4358, opened 2023-10-27
Admin tool: Orgs amount raised

sampoder:
Add an admin tools that shows the amount raised! Fixes #4356.


app/controllers/admin_controller.rb
```diff
@@ -206,6 +206,17 @@ def events
     render layout: "admin"
   end
 
+  def raised
```
cjdenio: Could this be part of the `organizations` page instead of being a separate page?

sampoder: Yes it could and I think it'd make more sense if it was. #4356 described a separate page, that was the only reason I went with that option. @garyhtou @MelSmith104 thoughts?

garyhtou: I'm totally cool with it being on the orgs page. Mel described it to me as a separate page, but I am in favor of having it on the same page as long as it doesn't affect page load times

sampoder: Coolio, it's now a part of the organizations page :D

garyhtou (dismissed):
sorry it's been a second. I'm thinking we should revisit our approach for this PR. Here are my thoughts:
- `raised_v2_cents` is a good method. but let's name it `raised_cents`. we're trying to get ride of the `v2` naming.
- let's not modify the existing methods. i think we should be able to get the same calculation by subtracting disbursements without a fee; rather than excluding them from the query. Caleb's recent PR https://github.com/hackclub/hcb/pull/4845 might help with this.

garyhtou:
Before merging, let's make sure it's in line with https://github.com/hackclub/hcb/issues/4875

app/models/event.rb
```diff
@@ -436,6 +436,17 @@ def balance_v2_cents(start_date: nil, end_date: nil)
       end
   end
 
+  def raised_cents(start_date: nil, end_date: nil)
+    @raised_cents ||=
+      begin
+        sum = settled_incoming_balance_cents(start_date:, end_date:)
+        sum += pending_incoming_balance_v2_cents(start_date:, end_date:)
+        sum += fronted_incoming_balance_v2_cents(start_date:, end_date:) if can_front_balance?
+        sum -= Disbursement.where(event_id: id, should_charge_fee: false).sum(:amount)
```
garyhtou: Should filter using `start_date` and `end_date`

garyhtou:
@sampoder, imma send you a screenshot of a couple of orgs and their total raised (computed by Daisy). can you compare this implementation and Daisy's results?

PR #4359, opened 2023-10-27
Catch exceptions regarding Bank Syncs on the nightly job

devramsean0:
This should catch exceptions on the nightly Bank Sync Job allowing other banks to continue syncing if one errors.

I suggest we rerun the sync job within the preview deployment to ensure I didn't break anything (I shouldn't have but you never know)


sampoder (changes requested):
This is awesome! Quick thing: it looks like there are some changes from your pinned cards PR still on the branch, we probably want to remove them

devramsean0:
Oh, weird. I switched branches and VScode didn't show any changes on the branch before I started

sampoder (approved):
lgtm!

maxwofford:
@devramsean0 while this functions, there's probably a simpler way to do this in ruby. I'd encourage you to look around a bit and see if there's something built in that can one-line this for us.

devramsean0:
@maxwofford  don't think there is a built-in way to apply an error handler per line in a function (A google search didn't find anything for ruby and I couldn't find anything for rails either) additionally, ChatGPT when I asked just now came up with the same solution to what I've done.

I could however be looking in the wrong places.

devramsean0:
Oh, Interesting! I had no idea you could call a method like that

maxwofford:
EDIT: Reposted because I had a typo from the ChatGPT response (that, ironically, ChatGPT has correct)

Nice job with searching, ChatGPT and Google will usually get results with such an old language. Good ruby code commonly reads like a regular English sentence. For example, I'd start with writing this to make it kinda like English:

```diff
- import_raw_plaid_transactions!
+ try_to import_raw_plaid_transactions!
```

The code will _try to import raw plaid transactions_, but it doesn't say it actually will succeed. Now to figure out how we can write that:

Screenshot 2023-10-30 at 18 44 09

I didn't include the code it spit out because it was kinda verbose, but after cleaning it up I got it down to this:

```rb
def try_to(&block)
  yield
rescue => e
  Airbrake.notify(e)
end
```

With the `&block` syntax, I need to update how I'm running the code:


```diff
- try_to import_raw_plaid_transactions!
+ try_to { import_raw_plaid_transactions! }
```

Now I can `try_to` run any sort of method in a single line! This won't solve all your problems; for example, you'll have to figure out some way of running `import_raw_plaid_transactions!`, but I'm sure you can figure it out from here!

devramsean0:
@maxwofford I've made that change but now we have some failed CI (tests) which I'm not sure how to fix.

app/services/transaction_engine/nightly.rb
```diff
@@ -9,30 +9,31 @@ def initialize(start_date: nil)
...
+        # (2) Hash transactions
+        try_to { hash_raw_plaid_transactions! }
+        # hash_raw_emburse_transactions! TODO: remove completely
+        try_to { hash_raw_stripe_transactions! }
+        try_to { hash_raw_csv_transactions! }
+        try_to { hash_raw_increase_transactions! }
+        # (3) Canonize transactions
+        try_to { canonize_hashed_transactions! }
+        # (4) Fix plaid mistakes
+        try_to { fix_plaid_mistakes! }
```
maxwofford: Seems like we're already wrapping this in a rescue under here https://github.com/hackclub/hcb/pull/4359/files#diff-f9e99fb7aa8b1e445d8bf52e00c10cb35be31ab4d3077a242b8c25822e9799a4L112

Do we need to wrap it again?

app/services/transaction_engine/nightly.rb
```diff
@@ -9,30 +9,31 @@ def initialize(start_date: nil)
...
+        # (5) Fix memo mistakes
+        try_to { fix_memo_mistakes! }
+      rescue => e
```
maxwofford: Same as above, it seems like we don't need to wrap the `run` method if our individual methods are all being rescued individually.

app/services/transaction_engine/nightly.rb
```diff
@@ -9,30 +9,31 @@ def initialize(start_date: nil)
     end
 
     def run
-      # 1 raw imports
-      import_raw_plaid_transactions!
-      import_other_raw_plaid_transactions!
+      begin
```
maxwofford: Small syntax thing, but [you can do this](https://stackoverflow.com/a/1542682)

maxwofford (commented):
Nice work! I added a couple other comments, but these are more suggestions; your code should run fine as is

YodaLightsabr:
Could we possibly use inline `rescue`? It seems to work similarly to our `try_to` method.

image

devramsean0:
EDIT: Added some references and context behind my confusion
Probably! How does rescue work with passing error contents to variables though?

Additionally, I've found at least 1 article saying that using inline rescue is a bad idea [[1]](https://thoughtbot.com/blog/don-t-inline-rescue-in-ruby)
And a Ruby Issue page suggesting that inline rescue can't take arguments (and therefore surely not set a var like the `e` I'm currently using) [[2]](https://bugs.ruby-lang.org/issues/6739)

YodaLightsabr:
Interesting, I hadn't considered that. To address concern 1, I don't think that filtering the type of error would be especially useful here, since we would be sending it to Airbrake anyways. However, because you have to pass a value and not a block, I agree that we should move forward with our clever `try_to` method since inline `rescue` wouldn't work for our use case.

It's certainly interesting syntax that I hadn't seen before, though. TIL!

app/services/transaction_engine/nightly.rb
devramsean0: Good Catch!
I might have gone a bit over the top with the error handling :)

app/services/transaction_engine/nightly.rb
devramsean0: Probably not, tbh I forgot I added error handling there

devramsean0:
I would love another review on this. But hopefully, this should be ready to merge!

app/services/transaction_engine/nightly.rb
```diff
@@ -117,3 +131,9 @@ def fix_memo_mistakes!
 
   end
 end
+
+def try_to(&block)
+  block.call
```
cjdenio:
```suggestion
def try_to
  yield
```

devramsean0: What does `yeid` do?

garyhtou (approved):
great work here! i'm excited to get this in

PR #3360, opened 2022-12-26
SMS receipt upload

maxwofford:
Initial implementation of #3270

PR #4363, opened 2023-10-29
Feature: SMS receipts to Receipt Bin

sampoder:
inspired by #4105, adapted from #3360. very much a WIP.


app/controllers/twilio_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  private
+
+  def user
+    @user ||= begin
+      potential_users = User.where(phone_number: params["From"])
+      return potential_users.first if potential_users.count == 1
+
+      # If we have multiple users with the same phone number, try to find the user via their stripe card
+      user_id = last_sent_message_hcb_code&.canonical_pending_transactions&.last&.stripe_card&.user&.id
```
YodaLightsabr: Should we continue to support multiple accounts linked to the same phone number?

app/controllers/twilio_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+  before_action :attachments, :user
+
+  def webhook
+
+    if @user.present? && Flipper.enabled?(:receipt_bin_2023_04_07, @user)
+
+      # This is a user & they're opted into the receipt bin, continue forward
+
+      if @attachments.none?
+        return
```
YodaLightsabr:
```suggestion
      return if @attachments.none?
```

app/controllers/twilio_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+  include ActionView::Helpers::TextHelper # for `pluralize`
+
+  protect_from_forgery except: :webhook # ignore csrf checks
+  skip_after_action :verify_authorized # do not force pundit
+  skip_before_action :signed_in_user # do not require logged in user
+  before_action :attachments, :user
+
+  def webhook
+
+    if @user.present? && Flipper.enabled?(:receipt_bin_2023_04_07, @user)
```
YodaLightsabr: Maybe we could try something like this?
```suggestion
    return unless @user.present? && Flipper.enabled?(:receipt_bin_2023_04_07, @user)
```
I noticed at the bottom we're responding with XML no matter what, and I'm wondering if we should only show that if the receipts were uploaded.

app/controllers/twilio_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+      else
+        grab_attachments!
+
+        @attachments.each do |attachment|
+
+          attachable = ActiveStorage::Blob.create_and_upload!(
+            io: attachment[:io],
+            filename: attachment[:filename],
+            content_type: attachment[:content_type]
+          )
```
YodaLightsabr: I'm not sure if you have to wrap the attachment data with `ActiveStorage::Blob.create_and_upload!`. In the receipt upload mailbox I just passed the attachment data. 

https://github.com/hackclub/hcb/blob/3df4d72a850ba4d9bf80c151ae7122b8b0938e96/app/mailboxes/receipt_uploads_mailbox.rb#L89-L93

YodaLightsabr (commented):
Nice work!

app/controllers/twilio_controller.rb
sampoder: Oh, sick!

sampoder: That makes this super simple now

app/controllers/twilio_controller.rb
sampoder: Let me investigate. IMO no.

app/controllers/twilio_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    respond_to do |format|
+      format.xml { render xml: "<Response><Message> Added #{pluralize @attachments.count, 'receipt'} to your receipt bin! </Message></Response>" }
+    end
+  end
+
+  private
+
+  def user
+    @user ||= begin
+      potential_users = User.where(phone_number: params["From"])
```
garyhtou: should we also filter by `phone_number_verified`?

app/controllers/twilio_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+        (0..num_media - 1).each do |i|
+          attachment = {}
+          attachment[:media_url] = params["MediaUrl#{i}"]
+          attachment[:filename] = File.basename(URI.parse(attachment[:media_url]).path)
+          attachment[:content_type] = params["MediaContentType#{i}"]
+          results << attachment
+        end
+      end
+
+      results
```
garyhtou: please test this first

```suggestion
      num_media = params["NumMedia"].to_i
      return [] if num_media.zero?

      (0..num_media - 1).map do |i|
        {
          media_url: params["MediaUrl#{i}"],
          filename: File.basename(URI.parse(attachment[:media_url]).path),
          content_type: params["MediaContentType#{i}"]
        }
      end
```

app/controllers/twilio_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+  end
+
+  def grab_attachments!
+    @attachments.each do |attachment|
+      url = URI.parse(attachment[:media_url])
```
garyhtou: if wonder if we should filter by `http/https` only as a security measure

app/controllers/twilio_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+          attachment[:content_type] = params["MediaContentType#{i}"]
+          results << attachment
+        end
+      end
+
+      results
+    end
+  end
+
+  def grab_attachments!
```
garyhtou: should we combine this into the `attachments` method?

app/controllers/twilio_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+        receiptable: nil,
+        uploader: @user,
+        attachments: attachment,
+        upload_method: "sms"
+      ).run!
+    end
+
+    respond_to do |format|
+      format.xml { render xml: "<Response><Message> Added #{pluralize @attachments.count, 'receipt'} to your receipt bin! </Message></Response>" }
+    end
```
garyhtou:
```suggestion
    receipts = ::ReceiptService::Create.new(
      receiptable: nil,
      uploader: @user,
      attachments: @attachments,
      upload_method: "sms"
    ).run!

    respond_to do |format|
      format.xml { render xml: "<Response><Message> Added #{pluralize receipts.count, 'receipt'} to your receipt bin! </Message></Response>" }
    end
```

garyhtou (changes requested):

app/controllers/twilio_controller.rb
sampoder: after combining the grab attachments method and attachments i think it's more readable to use the `.each`. it also lets me use `break` etc.

app/controllers/twilio_controller.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+class TwilioController < ApplicationController
+  include ActionView::Helpers::TextHelper # for `pluralize`
+
+  protect_from_forgery except: :webhook # ignore csrf checks
+  skip_after_action :verify_authorized # do not force pundit
+  skip_before_action :signed_in_user # do not require logged in user
```
garyhtou:
```suggestion
class TwilioController < ActionController::Base
  protect_from_forgery except: :webhook
  include ActionView::Helpers::TextHelper # for `pluralize`
```

app/controllers/twilio_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      else
+        # If we have multiple users with the same phone number, try to find the user via their stripe card
+        user_id = last_sent_message_hcb_code&.canonical_pending_transactions&.last&.stripe_card&.user&.id
+        potential_users.find_by(id: user_id)
+      end
+    end
+  end
+
+  def attachments
+    @attachments ||= begin
```
garyhtou:
```suggestion
  def set_attachments
    @attachments = begin
```

same for `set_user`. and don't forget to update the `before_action`

app/controllers/twilio_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+      (https://hcb.hackclub.com/my/settings).
+    MSG
+
+    return reply_with(<<~MSG.squish) unless Flipper.enabled?(:receipt_bin_2023_04_07, @user)
+      Hey! Looking to upload receipts? Make sure the Receipt Bin feature preview
+      is enabled on your account (https://hcb.hackclub.com/my/settings/previews).
+    MSG
+
+    return reply_with(<<~MSG.squish) if @attachments.none?
+      Hey! We couldn't find any attachments in your message. If you're looking
```
garyhtou: can we hint that this feature is for uploading receipts? i can see someone replying to a "your card was charged" SMS and then getting confused by "We couldn't find any attachments in your message"

garyhtou (approved):
great work on this!

garyhtou (approved):
great work on this!

garyhtou (approved):
great work on this!

PR #4365, opened 2023-10-29
Add unique constraint to phone numbers

sampoder:
Before merging, these cases would need to be handled: https://github.com/hackclub/hcb/pull/4363#discussion_r1375513761

I wonder if we want to design some way for admins to have multiple phone numbers for testing.


sampoder:
I'm going to close this lol... we'll keep the jankness

PR #4366, opened 2023-10-30
Configure bundler to install to the gems docker volume

albertchae:
From what I can tell this never worked properly before, because the gems volume was always mounted at `/usr/local/bundle`
https://github.com/hackclub/hcb/blob/0d71af94edca388ef4406c2be07074f3607c5fd8/docker-compose.yml#L21

but the Dockerfile was configured to install to `/bundle`
https://github.com/hackclub/hcb/commit/d8cb6c8d374dbb5ab54e7bba1438ae1f95d49223#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557

Fixes https://github.com/hackclub/hcb/issues/4315


maxwofford (approved):
Thanks for fixing this! I tested locally and can confirm I wasn't able to cause a similar issue to what I encountered before after messing with my branch and re-entering the container.

PR #4369, opened 2023-10-30
Patch for card overspending glitch

cjdenio:
Fixes #4360


maxwofford:
Screenshot 2023-10-30 at 12 00 30

Seems like this job still only runs once per day unless I'm misreading this. Isn't there still an overspending window?

PR #4383, opened 2023-11-01
Use default credentials for all non prod environments

albertchae:
Fixes https://github.com/hackclub/hcb/issues/4368

After this change, we will use `config/credentials.yml.enc` and
`config/master.key` for all non prod environments (e.g. development,
test, etc). Production will use `config/credentials/production.yml.enc`
and the key provided as the `RAILS_MASTER_KEY` environment variable.

To achieve this, I moved the existing development credentials to the
global files
`config/credentials/development.key` -> `config/master.key`
`config/credentials/development.yml.enc` -> `config/credentials.yml.enc`

The existing global credentials to the production file
`config/credentials.yml.enc` -> `config/credentials/production.yml.enc`

Git doesn't recognize all these changes as moves, but we can verify
even from the diffs of the ciphertexts.

- [x] Can run `./docker_dev_setup.sh` on a fresh checkout without `ActiveSupport::MessageEncryptor::InvalidMessage` errors
- [x] Can run `bin/rails credentials:edit` and edit `config/credentials.yml.enc` (the development/global credentials)
- [x] Can run `bin/rails credentials:edit --environment production` and edit `config/credentials/production.yml.enc`, provided we have the correct `config/credentials/production.key`
- [x] Can set `USE_PROD_CREDENTIALS` to use `config/credentials/production.yml.enc` locally
- [x] When running on heroku with `RAILS_ENV=production` and `RAILS_MASTER_KEY` set to the production key, accesses the right `config/credentials/production.yml.enc` and can decrypt it. We can verify this by running a heroku console and checking for `Rails.application.credentials`


docker_dev_setup.sh
```diff
@@ -4,12 +4,7 @@
 # reach out to Max Wofford (max@hackclub.com) if you have any questions or issues
 
 echo "
-$(tput setaf 9)Hack Club Bank:$(tput sgr0) Step 0/5: Copy .env file"
-cp -n config/credentials/development.key config/master.key
-echo "$(tput setaf 9)Hack Club Bank:$(tput sgr0) $(tput setaf 10)Done$(tput sgr0)"
-
-echo "
-$(tput setaf 9)Hack Club Bank:$(tput sgr0) Step 1/5: Copy development key"
+$(tput setaf 9)Hack Club Bank:$(tput sgr0) Step 1/5: Copy .env file"
```
albertchae: The comment was actually switched up before, but as noted in the issue, the new arrangement of credentials means we don't have to copy this key.

config/master.key
albertchae, line 1: It's a bit odd to commit `config/master.key` at all (by default it is in rails .gitignore), but in our setup it's the same as committing `config/credentials/development.key`.

app/services/gsuite_service.rb
```diff
@@ -12,7 +12,7 @@ class GsuiteService
 
   # this is a hack to work with the google library's requirement that tokens must be in files
   TOKEN_FILE = Tempfile.new("token")
-  TOKEN_FILE << Rails.application.credentials.gsuite[:token]
+  TOKEN_FILE << Rails.application.credentials.dig(:gsuite, :token)
```
albertchae: An alternative to all these dig changes is to add this key to `config/credentials.yml.enc`

.github/workflows/ci.yml
```diff
@@ -94,13 +92,11 @@ jobs:
         env:
           DATABASE_URL: postgres://postgres:postgres@localhost:5432/bank_test
           RAILS_ENV: test
-          RAILS_MASTER_KEY: ${{ secrets.RAILS_MASTER_KEY }}
         run: bundler exec rails db:create db:schema:load RAILS_ENV=test
       - name: Run RSpec
         env:
           DATABASE_URL: postgres://postgres:postgres@localhost:5432/bank_test
           RAILS_ENV: test
-          RAILS_MASTER_KEY: ${{ secrets.RAILS_MASTER_KEY }}
```
albertchae: This explains how it was working in CI, but one of the benefits now is we shouldn't even need RAILS_MASTER_KEY anywhere in github.

app.json
```diff
@@ -19,9 +19,6 @@
     "RAILS_LOG_TO_STDOUT": {
       "required": true
     },
-    "RAILS_MASTER_KEY": {
-      "required": true
-    },
```
albertchae: We also need to remove this in the heroku UI from the pipeline settings for review apps, because currently all apps get the production master key provided as `RAILS_MASTER_KEY`

I also verified `When running on heroku with RAILS_ENV=production and RAILS_MASTER_KEY set to the production key, accesses the right config/credentials/production.yml.enc and can decrypt it. We can verify this by running a heroku console and checking for Rails.application.credentials` on a review app, so I checked that one off

.github/workflows/ci.yml
```diff
@@ -12,8 +12,6 @@ jobs:
         with:
           bundler-cache: true
       - name: Run Rubocop
-        env:
-          RAILS_MASTER_KEY: ${{ secrets.RAILS_MASTER_KEY }}
```
maxwofford: This is hilarious that we're passing in the master key just to run rubocop. Thanks for cleaning this up.

maxwofford (approved):
Thank you for taking this update carefully. You're good for rolling this out!

albertchae:
Deploy finished. Verified I could still access prod credentials with a console and `Rails.application.credentials`

app.json
albertchae: As expected, after merging this PR, subsequent review apps are failing due to presence of `RAILS_MASTER_KEY`. I removed `RAILS_MASTER_KEY` from the review app settings and am trying to rebuild to see if it works after.

albertchae: Verified review app works after `RAILS_MASTER_KEY` was removed

.github/workflows/ci.yml
albertchae: @maxwofford I don't have permissions to access [Github secrets](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions), can you remove `RAILS_MASTER_KEY` from there?

garyhtou: I just removed it from GitHub secrets

albertchae: Thanks @garyhtou !

PR #4355, opened 2023-10-26
Admin Weekly Ledger Audits

sampoder:
Screenshot 2024-01-10 at 5 39 21 PM

Screenshot 2024-01-10 at 5 40 36 PM

PR #4384, opened 2023-11-01
Add public view of card grant spending 

sampoder:
This uses a model that pops up on transparent orgs with card grants:

Screenshot 2023-11-01 at 12 04 44 AM

All private info should be hidden, though I'd appreciate a sanity check on that.

I thought a model was the cleanest way to do this.

Closes #4380.

Happy halloween! 🎃


maxwofford:
Awesome work on this @sampoder! I'm half tempted to merge even with it needing tweaks because this by itself makes everything so much clearer.

sampoder:
@maxwofford awesome! i just changed it up a little so that this view shows up when you click "Details" while signed in as an organiser and added a little icon to keep it visually distinct

app/assets/javascripts/ui.js
```diff
@@ -184,7 +184,7 @@ $(document).on('turbo:load', function () {
 
   $.each(BK.s('async_frame'), (i, frame) => {
     const loadFrame = () => {
-      $.get($(frame).data('src'), data => {
+      $.get(`${$(frame).data('src')}?async_frame=true`,  data => {
```
YodaLightsabr: I think this might interfere with or break the HcbCode popover implementation
https://github.com/hackclub/hcb/blob/837ba812492346d041c5eddf50865b9941097c3d/app/models/hcb_code.rb#L53

app/controllers/card_grants_controller.rb
```diff
@@ -55,6 +55,22 @@ def show
...
 
+  def public
+    @card_grant = CardGrant.find_by_hashid!(params[:id])
+
+    authorize @card_grant
+
+    @card = @card_grant.stripe_card
+    @hcb_codes = @card&.hcb_codes
+
+    @async_frame = params[:async_frame].present?
```
YodaLightsabr:
```suggestion
    @frame = params[:frame].present?
```

Thoughts on using `frame` instead of `async_frame`?

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -2,7 +2,7 @@
 <tr
   class="transaction--<%= pt.amount.negative? ? "negative" : "positive" %> muted"
   id="<%= pt.local_hcb_code.hashid %>"
-  <% if defined?(receipt_upload_button) %>
+  <% if defined?(receipt_upload_button) && receipt_upload_button %>
```
YodaLightsabr:
```suggestion
  <% if local_assigns[:receipt_upload_button] %>
```
This might be a little bit cleaner (for cases like this where you want to be able to pass in `false`)

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -158,9 +158,9 @@
     <td></td>
   <% end %>
   <td>
-    <%= link_to "Details", pt.url, data: Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) ? { turbo_frame: "_top", behavior: "modal_trigger", modal: "transaction_details_#{instance}" } : { turbo_frame: "_top" } if organizer_signed_in? || defined?(force_display_details) %>
+    <%= link_to "Details", pt.url, data: Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (!defined?(@force_no_popover) || !@force_no_popover) ? { turbo_frame: "_top", behavior: "modal_trigger", modal: "transaction_details_#{instance}" } : { turbo_frame: "_top" } if organizer_signed_in? || defined?(force_display_details) %>
   </td>
-  <% if defined?(receipt_upload_button) %>
+  <% if defined?(receipt_upload_button) && receipt_upload_button %>
```
YodaLightsabr:
```suggestion
  <% if local_assigns[:receipt_upload_button] %>
```

app/views/canonical_transactions/_canonical_transaction.html.erb
```diff
@@ -2,7 +2,7 @@
 <tr
   class="transaction--<%= ct.amount.negative? ? "negative" : "positive" %> <%= ct.local_hcb_code.disbursement.special_appearance[:css_class] if !ct.amount.negative? && ct.local_hcb_code.disbursement&.special_appearance? %>"
   id="<%= ct.local_hcb_code.hashid %>"
-  <% if defined?(receipt_upload_button) %>
+  <% if defined?(receipt_upload_button) && receipt_upload_button %>
```
YodaLightsabr:
```suggestion
  <% if local_assigns[:receipt_upload_button] %>
```

app/views/canonical_transactions/_canonical_transaction.html.erb
```diff
@@ -172,9 +180,14 @@
...
+
+    <% if ct.local_hcb_code.card_grant? %>
+      <%= link_to "Details", public_card_grant_path(ct.local_hcb_code.disbursement.card_grant), data: Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (!defined?(@force_no_popover) || !@force_no_popover) ? { turbo_frame: "_top", behavior: "modal_trigger", modal: "card_grant_details_#{instance}" } : { turbo_frame: "_top" } if organizer_signed_in? || defined?(force_display_details) %>
+    <% else %>
+      <%= link_to "Details", ct.local_hcb_code.url, data: Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (!defined?(@force_no_popover) || !@force_no_popover) ? { turbo_frame: "_top", behavior: "modal_trigger", modal: "transaction_details_#{instance}", "modal-auto-open": defined?(@popover) ? @popover == ct.local_hcb_code.hashid : nil } : { turbo_frame: "_top" } if organizer_signed_in? || defined?(force_display_details) %>
+    <% end %>
   </td>
-  <% if defined?(receipt_upload_button) %>
+  <% if defined?(receipt_upload_button) && receipt_upload_button %>
```
YodaLightsabr:
```suggestion
  <% if local_assigns[:receipt_upload_button] %>
```

YodaLightsabr (commented):
Nice work on this! I just wanted to add a few comments to make sure our popover implementation is consistent.

app/assets/javascripts/ui.js
sampoder: that explains a lot omg... lol yes, let's use frame

sampoder:
```suggestion
      $.get(`${$(frame).data('src')}?frame=true`,  data => {
```

app/assets/javascripts/ui.js
```diff
@@ -184,7 +184,7 @@ $(document).on('turbo:load', function () {
 
   $.each(BK.s('async_frame'), (i, frame) => {
     const loadFrame = () => {
-      $.get($(frame).data('src'), data => {
+      $.get(`${$(frame).data('src')}?frame=true`,  data => {
```
sampoder:
```suggestion
      $.get($(frame).data('src')`,  data => {
```

app/assets/javascripts/ui.js
```diff
@@ -184,7 +184,7 @@ $(document).on('turbo:load', function () {
 
   $.each(BK.s('async_frame'), (i, frame) => {
     const loadFrame = () => {
-      $.get($(frame).data('src'), data => {
+      $.get($(frame).data('src')`,  data => {
```
sampoder:
```suggestion
      $.get($(frame).data('src'),  data => {
```

app/views/canonical_transactions/_canonical_transaction.html.erb
```diff
@@ -280,3 +293,12 @@
     <% end %>
   </section>
 <% end %>
+
+<% if (Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) || !organizer_signed_in?) && ct.local_hcb_code.card_grant? %>
+  <section class="modal modal--scroll modal--popover bg-snow" data-behavior="modal" role="dialog" id="card_grant_details_<%= instance %>">
+    <%= modal_header(transaction_memo(ct), level: "h1", external_link: public_card_grant_path(ct.local_hcb_code.disbursement.card_grant)) %>
+    <%= async_frame_to public_card_grant_path(ct.local_hcb_code.disbursement.card_grant, params: { async_frame: true }), as: :div do %>
```
sampoder:
```suggestion
    <%= async_frame_to public_card_grant_path(ct.local_hcb_code.disbursement.card_grant, params: { frame: true }), as: :div do %>
```

sampoder:
@maxwofford @YodaLightsabr: whilst the HCPD may be fictitious, what do we need to do to push this PR over the line? Does it look good to the both of you?

app/controllers/card_grants_controller.rb
```diff
@@ -55,6 +55,22 @@ def show
     redirect_to auth_users_path(email: @card_grant.user.email, return_to: card_grant_path(@card_grant)), flash: { info: "Please sign in with the same email you received the invitation at." }
   end
 
+  def public
```
garyhtou: This action should be renamed. `public` is basically like a keyword in Ruby; we don't want to be overriding that

app/views/canonical_transactions/_canonical_transaction.html.erb
```diff
@@ -69,6 +69,10 @@
           <%= inline_icon(ct.local_hcb_code.disbursement.special_appearance[:icon] || "door-enter") %>
         </td>
       <% end %>
+    <% elsif ct.local_hcb_code.card_grant? %>
+      <td class="transaction__icon tooltipped tooltipped--e" aria-label="Card Grant">
+        <%= inline_icon "like" %>
```
garyhtou: Do we wanna update the transfer page to have the same terminology/icon?

https://github.com/hackclub/bank/blob/c1807ec9e56164a6fca4a1c9b4dfe35aea440340/app/views/events/transfers.html.erb#L81-L83

config/routes.rb
```diff
@@ -525,9 +525,10 @@
     end
   end
 
-  resources :card_grants, only: [:show], path: "grants" do
+  resources :card_grants, only: [:show, :public], path: "grants" do
```
garyhtou: if we have `get "public"` at the bottom, I don't think we need `:public` here

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -158,9 +158,9 @@
     <td></td>
   <% end %>
   <td>
-    <%= link_to "Details", pt.url, data: Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) ? { turbo_frame: "_top", behavior: "modal_trigger", modal: "transaction_details_#{instance}" } : { turbo_frame: "_top" } if organizer_signed_in? || defined?(force_display_details) %>
+    <%= link_to "Details", pt.url, data: Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (!defined?(@force_no_popover) || !@force_no_popover) ? { turbo_frame: "_top", behavior: "modal_trigger", modal: "transaction_details_#{instance}" } : { turbo_frame: "_top" } if organizer_signed_in? || defined?(force_display_details) %>
```
garyhtou: There is no need to check if `@force_no_popover` is defined. Undefined instance variables return `nil`. So, `(!defined?(@force_no_popover) || !@force_no_popover)` I think can be rewritten as just `!@force_no_popover`

app/views/canonical_transactions/_canonical_transaction.html.erb
```diff
@@ -172,9 +180,14 @@
...
+
+    <% if ct.local_hcb_code.card_grant? %>
+      <%= link_to "Details", public_card_grant_path(ct.local_hcb_code.disbursement.card_grant), data: Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (!defined?(@force_no_popover) || !@force_no_popover) ? { turbo_frame: "_top", behavior: "modal_trigger", modal: "card_grant_details_#{instance}" } : { turbo_frame: "_top" } if organizer_signed_in? || defined?(force_display_details) %>
+    <% else %>
+      <%= link_to "Details", ct.local_hcb_code.url, data: Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (!defined?(@force_no_popover) || !@force_no_popover) ? { turbo_frame: "_top", behavior: "modal_trigger", modal: "transaction_details_#{instance}", "modal-auto-open": defined?(@popover) ? @popover == ct.local_hcb_code.hashid : nil } : { turbo_frame: "_top" } if organizer_signed_in? || defined?(force_display_details) %>
+    <% end %>
```
garyhtou: Is the only difference the link? If so, do

```ruby
<% details_link = ct.local_hcb_code.card_grant? ? public_card_grant_path(ct.local_hcb_code.disbursement.card_grant) : ct.local_hcb_code.url %>

<%= link_to "Details", details_link, blath blath blah other stuff %>
```

app/views/card_grants/show.html.erb
garyhtou, line 1: Instead of using the `show` view for both the full-page and "public"/popover version. Let's create a separate page for the public version. It can sit at `card_grants/:id/popover`.

Over all shared content from this page into their own partials. I'm thinking we could have 3:
- `balance`
- `spending_instructions`
- `transactions`

This will also us to get rid of the `@public` instance variable

garyhtou (changes requested):
This is very cool! Two large feedbacks:
- We should find a new name for the controller action (we're currently overriding an existing`public` method)
- Give the popover (public) version its own view file. This means moving shared content into partials.

maxwofford (dismissed):
Hey @sampoder! You or @garyhtou should have a chat with @SerenityUX. HCPD might not be real, but the HCPD: Pizza Crimes Division is a real thing headed by them and they probably have some opinions about this!

As for https://github.com/hackclub/hcb/issues/4380, I think this is missing one big component: on a regular event, I can audit it by looking at the transaction history in linear order. If I check it each day, I can look at the most recent ~5 transactions and stop searching once I reach the previous day's transactions.

On card grant events, if I want to see if someone has done something invalid in the past day, **I have to go through every card's popover every single day**. I think that's too hard to audit / call transparent.

sampoder:
howdy! that makes sense to me. i guess that wasn't quite the problem i was trying to solve with the original PR which is why what we've got here isn't a perfect fit. instead of trying to shoehorn in a fix to what we already have, would it make sense to build out an admin tool like we have with https://github.com/hackclub/hcb/pull/4355? some other ideas I have are: including card grant transactions in exports or having a toggle to list them on the ledger directly.

sampoder:
Oh, and I'll also text Thomas!

sarthaktexas:
What you suggested (daily compliance checks) is what Thomas, Jasper, and I agreed on. I think this PR is fine to merge especially because they need this quickly. cc @maxwofford

sarthaktexas:
@maxwofford said his request to change was in fact not a request to not merge. We are just waiting for an approval from @garyhtou.

garyhtou:
> What you suggested (daily compliance checks) is what Thomas, Jasper, and I agreed on.

Is this their intention behind https://github.com/hackclub/hcb/issues/4932?

sarthaktexas:
Oh we have two options. Webhook works if you want to add that into the codebase -- or what Max suggested is a compliance check feature for organizations (not an internal admin tool) giving out card grants.

garyhtou:
> Oh we have two options. Webhook works if you want to add that into the codebase -- or what Max suggested is a compliance check feature for organizations (not an internal admin tool) giving out card grants.

I think webhooks will be a lighter lift for us. Let's go for that especially since it's a very niche use case (we have few card grant programs). https://github.com/hackclub/hcb/issues/4932 outlines the spec that Jasper and I agreed on

garyhtou (approved):
coolio!

PR #4393, opened 2023-11-01
✨ Make organisational balance table row link to org page

malted:


maxwofford (approved):
We chatted IRL about rewriting this without the loop, but this code functions and is following the previous style so it's not a regression.

PR #4397, opened 2023-11-02
Decrease Sidekiq pool to 5

cjdenio:


garyhtou:
Maybe look into modifying sidekiq's concurrency

https://github.com/hackclub/bank/blob/e6e69b1593e766e48baef778def71d20343c9744/config/sidekiq.yml#L2-L2

PR #4407, opened 2023-11-03
Only do nightly job for one-week old transactions & younger

sampoder:
Is there a shorter way of doing the unix stuff?


app/services/partners/stripe/issuing/authorizations/list.rb
```diff
@@ -7,13 +7,13 @@ module Authorizations
         class List
           include StripeService
```
garyhtou:
```ruby
def initialize(created_after: nil)
  @created_after = created_after
end
```

app/services/pending_transaction_engine/nightly.rb
```diff
@@ -65,7 +65,9 @@ def import_raw_pending_stripe_transactions!
     end
 
     def canonize_raw_pending_stripe_transactions!
-      ::PendingTransactionEngine::CanonicalPendingTransactionService::Import::Stripe.new.run
+      ::PendingTransactionEngine::CanonicalPendingTransactionService::Import::Stripe.new(
+        created_after: 1.week.ago.to_i
```
garyhtou:
```suggestion
        created_after: 1.week.ago.to_i.to_s
```

PR #4410, opened 2023-11-03
Make Lob Check read-only

sampoder:
@garyhtou I think this gets them all. There's a bunch of Check-related services, do we want to strip those out as well?

closes https://github.com/hackclub/hcb/issues/4408

---

hi it's @garyhtou, I went a bit overboard and decided to rip out everything except the stuff needed to view Checks.

Here is what's left:
- Admin `index` page `/admin/checks`
  image

- Check `show` page, which redirects to HcbCode. This page hasn't visually changed
- Check `view_scan` page. This hasn't changed. it just redirects to lob's site for the Check PDF. however, it is mostly broken due to Lob's data retention policy (ends up 404 or 403'ing). Lob no longer has those Check PDFs I think. See comments in https://github.com/hackclub/hcb/issues/3601


garyhtou:
@cjdenio i just wanna get a sanity check for the files in `services` and `jobs` being removed

cjdenio (approved):
Did a look-through and I don't see anything that shouldn't be deleted!

garyhtou:
thanks!

PR #4411, opened 2023-11-03
Drop `attempted_match_at` from Receipts

sampoder:
hopefully this does what it says on the tin


garyhtou (commented):
This is amazing! Would you mind following this guide?

https://andycroll.com/ruby/safely-remove-a-column-field-from-active-record/

Spin up a new PR for ignoring the column. We'll merge that PR first, then we'll merge this one, and then you can follow up with another PR to remove the `ignored_columns`

garyhtou (approved):
good to go after merging https://github.com/hackclub/hcb/pull/4416! 👍

PR #4412, opened 2023-11-03
Rip out PartnerDonation and PartneredSignup

sampoder:
This may be ripping out too much... let me know. There's still more to go.


sampoder:
TODO: figure out RSpec issues. If anyone has any ideas lemme know!

sampoder:
alright RSpec is spec-ing and going green!

garyhtou:
https://github.com/hackclub/hcb/issues/4409

db/migrate/20231103184005_drop_partner_donations_partnered_signups_partners.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class DropPartnerDonationsPartneredSignupsPartners < ActiveRecord::Migration[7.0]
+  def up
+    drop_table :partner_donations, force: :cascade
+    drop_table :partnered_signups, force: :cascade
+    drop_table :partners, force: :cascade
```
garyhtou, line 7: be careful with `force: :cascade`. you can very easily unintentionally delete other tables.

I'd recommend explicitly deleting each table and not using `force: :cascade`.

db/migrate/20231103184005_drop_partner_donations_partnered_signups_partners.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class DropPartnerDonationsPartneredSignupsPartners < ActiveRecord::Migration[7.0]
+  def up
+    drop_table :partner_donations, force: :cascade
```
garyhtou, line 5: I wonder if we need to keep this data. it is financial data. maybe there's an option to convert them to donation objects? (might be a bad idea, not sure)

config/routes.rb
```diff
@@ -17,8 +17,6 @@
 
   # API documentation
   scope "docs/api" do
-    get "v2", to: "docs#v2"
-    get "v2/swagger", to: "docs#swagger"
```
garyhtou, line 21: good catch

garyhtou (commented):
i haven't looked through the whole thing, but here are a few comments. please get my approval before merging since we're dropping tables/data.

sampoder:
yeah, not planning on merging this anytime soon. it frankly scares me a little.

sampoder:
also with the merge conflicts this might need to be redone

garyhtou:
i think all the Contracts can be deleted and `dry-validation` can be uninstalled

PR #4413, opened 2023-11-03
Use `SendTwilioReceiptMessage` to send Receipt SMSs

sampoder:
The other one didn't exist, see: https://hackclub.slack.com/archives/C04LDFJ5X7S/p1699041247697239


garyhtou (approved):
Hmm interesting! I was wondering why this error just popped up now since the code hasn't been touched in a while. But I now see that it's an admin-only feature.

Thanks for fixing!

PR #4419, opened 2023-11-04
Configure Sidekiq queues to be `critical`, `default`, and `low`

sampoder:
Fixes #4415 and https://github.com/hackclub/hcb/issues/4417.

See comment in  [`ff4bf08` (#4419)](https://github.com/hackclub/hcb/pull/4419/commits/ff4bf08d36c8dff11f3876342514d9c388f8753a) for some context on the two deployments


sampoder:
Now also fixes https://github.com/hackclub/hcb/issues/4417

PR #4421, opened 2023-11-04
Auto-rename STRIPE FEE REIMBURSEMENTs

sarthaktexas:
fixes #4354


sampoder:
to help with Rubocop, you can run `bin/lint`. it'll fix most things for you!

sarthaktexas:
> to help with Rubocop, you can run `bin/lint`. it'll fix most things for you!

thank you love you @sampoder

sampoder (approved):
looks good to me!

app/models/canonical_transaction.rb
```diff
@@ -57,6 +57,7 @@ class CanonicalTransaction < ApplicationRecord
   scope :emburse_transaction,  -> { joins("INNER JOIN raw_emburse_transactions  ON transaction_source_type = 'RawEmburseTransaction'  AND raw_emburse_transactions.id  = transaction_source_id") }
 
   scope :likely_hack_club_bank_issued_cards, -> { where("memo ilike 'Hack Club Bank Issued car%' or memo ilike 'HCKCLB Issued car%'") }
+  scope :likely_hack_club_bank_stripe_fee_reimbursement, -> { where("memo ilike 'STRIPE FEE REIMBURSEMENT%'") }
```
garyhtou:
```suggestion
  scope :likely_fee_reimbursement, -> { where(memo: "Stripe fee reimbursement") }
```

Two edits here:
- `likely_hack_club_bank_stripe_fee_reimbursement` -> `likely_fee_reimbursement`. This better reflects the `FeeReimbursement` model.
- We can make the `where` query more strict, the transaction memos are "Stripe fee reimbursement"
	image
	
	Side note: Technically, we have [control over this memo](https://github.com/hackclub/bank/blob/a0da88f1b4c335e4f7923dd4fad3a20905556897/app/services/fee_reimbursement_service/nightly.rb#L20-L20) and could change that line so that it is prettier. However, I like the approach you took here for one key reason: The new pretty memo contains an emoji, which may not be supported by all banks (such as Increase) and could be lost when it goes through their systems.

app/services/stripe_card_service/nightly.rb
garyhtou, line 1: The new code here looks great, however, let's move it into the FeeReimbursement's nightly job. That is located here:
https://github.com/hackclub/bank/blob/a0da88f1b4c335e4f7923dd4fad3a20905556897/app/services/fee_reimbursement_service/nightly.rb

garyhtou (changes requested):
you're on the right track!

app/models/canonical_transaction.rb
sarthaktexas: Why does it capitalize the memo i didn't realize it was lowercase cause of that

app/services/stripe_card_service/nightly.rb
sarthaktexas: that job only runs for unprocessed fee reimbursements, should I move it outside that fragment?

garyhtou:
<table>
<tr>
 <th>Before
 <th>After
<tr>
 <td>
image

 <td>
image

</table>

garyhtou (approved):
Yayyy! amazing job 🎉

PR #4422, opened 2023-11-04
Add simplecov to generate code coverage reports

KiKoS0:
Fixes: https://github.com/hackclub/hcb/issues/4406

A code coverage report should be generated when running tests with:
```bash
root:/usr/src/app# bundle exec rspec
Coverage report generated for RSpec to /usr/src/app/coverage. 4498 / 23784 LOC (18.91%) covered.
```
The report is generated in a /coverage directory and can be viewed:
```bash
# Mac
open coverage/index.html
# Linux
xdg-open coverage/index.html
```

Regarding publishing the results, both [codecov](https://about.codecov.io/pricing/) and [coveralls](https://coveralls.io/pricing) don't offer free multi-user plan for private repositories. Self hosting a solution is also an option but it's obviously way more involved so let's discuss if it's worth it @albertchae.


Gemfile
```diff
@@ -157,6 +157,7 @@ end
 
 group :test do
   gem "factory_bot_rails" # Test data
+  gem "simplecov", "~> 0.22.0", require: false # Code coverage
```
albertchae: i think we're probably fine not even specifying a twiddlewakka version on this, unless you had a reason for adding it?

albertchae (approved):
Looks good, I was able to run the report locally

image


We can figure out hosting later since for now even the local report will be useful.

Gemfile
KiKoS0: no reason, just an elixir habit 👍

PR #4429, opened 2023-11-07
Use `upsert_all` for Stripe transaction import in `PendingTransactionEngine`

cjdenio:
After a bit of investigating, it appears that `PendingTransactionEngine::Nightly#import_raw_pending_stripe_transactions!` is extremely slow— specifically all the ActiveRecord calls in `Stripe::ImportSingle`. This refactor should theoretically speed up the job by reducing allocations. I've tested this locally and confirmed that `RawPendingStripeTransactions` are both created and updated as expected.


cjdenio:
Merging— `PendingTransactionEngine` is currently not working so this is time-sensitive

PR #4432, opened 2023-11-07
Add a command bar

sampoder:
Uses the [kbar](https://kbar.vercel.app) package to create a command bar:

Screenshot 2023-11-07 at 3 58 04 PM

Why? Well, it seems like it's trendy. No, really, having something would make navigating HCB a lot quicker.


sampoder:
latest thing is a temp thing to see if I can get preview deploys working

app/assets/stylesheets/kbar.css
```diff
@@ -0,0 +1,9 @@
+:root {
+	--kbar-background: #fcfcfc;
```
garyhtou: we probably want to use spaces instead of tabs to be consistent

app/assets/stylesheets/application.scss
```diff
@@ -400,5 +400,6 @@ summary {
 
 @import 'basscss';
 @import 'space';
+@import 'kbar';
 @import 'utilities';
 @import 'dark';
```
garyhtou:
```suggestion
@import 'dark';

```

app/assets/stylesheets/kbar.css
```diff
@@ -0,0 +1,9 @@
...
+	--kbar-foreground: #1c1c1d;
+}
+
+html[data-dark='true'] {
+	--kbar-background: #1c1c1d;
+	--kbar-foreground: #fcfcfc;
+}
```
garyhtou:
```suggestion
}

```

app/javascript/components/command_bar/CommandBar.js
```diff
@@ -0,0 +1,682 @@
...
+							</kbd>
+						))}
+					</div>
+				) : null}
+			</div>
+		);
+	},
+);
+
+ResultItem.displayName = "ResultItem"
```
garyhtou:
```suggestion
ResultItem.displayName = "ResultItem"

```

app/controllers/users_controller.rb
```diff
@@ -414,6 +414,24 @@ def toggle_sms_auth
...
+    events = @current_user.events.reorder("organizer_positions.sort_index ASC", "events.id ASC").map { |x|
+      {
+        name: x.name,
+        slug: x.slug,
+        logo: x.logo.attached? ? Rails.application.routes.url_helpers.url_for(x.logo) : "none"
+      }
+    }
+
+    if admin_signed_in?
+      events.concat(Event.all.select([:slug, :name]).map { |e| { slug: e.slug, name: e.name } }.reject { |e| events.any? { |x| x[:slug] == e[:slug] } })
```
garyhtou: We can include the logo if we want. We should just make sure to use

https://edgeapi.rubyonrails.org/classes/ActiveStorage/Attached/Model.html#method-i-with_attached_-2A

```suggestion
      events.concat(Event.all.excluding(events).select([:slug, :name]).map { |e| { slug: e.slug, name: e.name } })
```

app/controllers/users_controller.rb
```diff
@@ -414,6 +414,24 @@ def toggle_sms_auth
     redirect_back_or_to security_user_path(current_user)
   end
 
+  def organizations
```
garyhtou: Let's move this method into `EventsController#index` when the format is JSON

app/javascript/components/command_bar/CommandBar.js
```diff
@@ -0,0 +1,682 @@
...
+import Icon from "@hackclub/icons";
+
+const searchStyle = {
+	padding: "12px 16px",
+	fontSize: "16px",
+	width: "100%",
+	boxSizing: "border-box",
+	outline: "none",
+	border: "none",
+	background: "var(--kbar-background)",
```
garyhtou: contrast is low between the selected search item and the non-selected items
image

app/javascript/components/command_bar/CommandBar.js
```diff
@@ -0,0 +1,682 @@
...
+			perform: () => (window.location.pathname = "/documents"),
+		},
+		{
+			id: "admin_tool_34",
+			section: "Admin Tools",
+			priority: Priority.HIGH,
+			name: "Flipper",
+			icon: <Icon glyph="rep" size={16} />,
+			perform: () => (window.location.pathname = "/flipper"),
+		},
```
garyhtou: Let's remove this. This is more of a developer tool rather than an admin tool

app/javascript/components/command_bar/CommandBar.js
```diff
@@ -0,0 +1,682 @@
...
+		{
+			id: "admin_tool_26",
+			section: "Admin Tools",
+			priority: Priority.HIGH,
+			name: "Domains",
+			icon: <Icon glyph="web" size={16} />,
+			perform: () =>
+				(window.location.href =
+					"https://airtable.com/tbl22cXd3Bo9uo0wp/viwcnZyoctJTFGVY2"),
```
garyhtou: Should we be concerned about leaking these URLs into the front end?

northeastprince:
I would suggest considering a simpler approach with just a few basic keyboard shortcuts to start using GitHub's [hotkey](https://github.com/github/hotkey) library, then seeing how it goes for a few weeks.

app/views/application/_header.html.erb
```diff
@@ -84,4 +84,27 @@
       <% end %>
     <% end %>
   </div>
+
+  <% if signed_in? && current_user&.admin? %>
```
garyhtou:
```suggestion
  <% if admin_signed_in? %>
```

app/javascript/components/command_bar/CommandBar.js
garyhtou: solved

garyhtou (approved):
Let's merge this!

garyhtou:
This is for admins only! If we decide to launch this for everyone, we'll need to polish up the authorization

maxwofford:
This is great! Want to flag though: if I use this and type "ledger" to go to the admin ledger, it takes me, but then when I try to trigger the command bar again it doesn't work! I think the JS isn't being loaded on the admin pages?

sampoder:
Good catch; let's add it!

PR #4436, opened 2023-11-08
Remove Redundant US & Canada only message

sampoder:
At the moment it looks a little awkward where it is:

Screenshot 2023-11-08 at 9 52 00 AM

We already say they are only available in the US & Canada so it's probably safe to remove:

Screenshot 2023-11-08 at 9 52 24 AM

Alternatively, we could shorten the message.


cjdenio (approved):
I'm cool with this!

PR #4439, opened 2023-11-08
Ensure USE_PROD_CREDENTIALS is "true"

sampoder:
So if you have it set as false... you don't use prod credentials.


config/application.rb
```diff
@@ -13,7 +13,7 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 7.0
 
-    if ENV["USE_PROD_CREDENTIALS"].present?
+    if ENV["USE_PROD_CREDENTIALS"].present? && ENV["USE_PROD_CREDENTIALS"].downcase == "true"
```
cjdenio:
```suggestion
    if ENV["USE_PROD_CREDENTIALS"]&.downcase == "true"
```

sampoder: oh that's nice

garyhtou: fun fact, there's a method called [`inquiry`](https://apidock.com/rails/String/inquiry):

```ruby
ENV["USE_PROD_CREDENTIALS"]&downcase.&.inquiry&.true?
```
is the same as
```ruby
ENV["USE_PROD_CREDENTIALS"]&.downcase == "true"
```

It's cool, but I still like the good old `==` for readability

garyhtou (approved):

PR #4442, opened 2023-11-08
Add WebP as a `web_image_content_type`

sampoder:
Based on https://edgeguides.rubyonrails.org/configuring.html#config-active-storage-web-image-content-types, I think this should fix #4441.


garyhtou (approved):
let's see if this work!!!

PR #4444, opened 2023-11-08
Email when Stripe card is lost

sampoder:
sad moment :(


app/mailers/stripe_card_mailer.rb
```diff
@@ -13,6 +13,16 @@ def physical_card_ordered
          subject: "Your new HCB card (ending in #{@card.last4}) for #{@event.name} is on its way"
   end
 
+  def physical_card_lost
```
northeastprince:
```suggestion
  def lost_in_shipping
```

app/mailers/stripe_card_mailer.rb
```diff
@@ -13,6 +13,16 @@ def physical_card_ordered
...
+    @card = StripeCard.find params[:card_id]
```
northeastprince: Don't repeat the logic, just extract it to a `before_action` callback. And I think we can just pass the model in, we don't need the ID.

db/migrate/20231108222523_add_lost_to_stripe_cards.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddLostToStripeCards < ActiveRecord::Migration[7.0]
+  def change
+    add_column :stripe_cards, :lost, :boolean, default: false
```
northeastprince: Maybe call the column `lost_in_shipping` instead? Without context, it sounds like someone lost their own card and reported it.

app/models/stripe_card.rb
```diff
@@ -258,6 +259,10 @@ def sync_from_stripe!
     end
 
     if stripe_obj[:shipping]
+      if (stripe_obj[:shipping][:status] == "returned" || stripe_obj[:shipping][:status] == "failure") && !self.lost
+        self.lost = true
+        StripeCardMailer.with(card_id: self.id).physical_card_lost.deliver_later
+      end
```
northeastprince:
```suggestion
      if stripe_obj[:shipping][:status] == "returned" || stripe_obj[:shipping][:status] == "failure"
        unless lost?
          self.lost = true
          StripeCardMailer.with(card_id: self.id).lost_in_shipping.deliver_later
        end
      end

```

app/mailers/stripe_card_mailer.rb
```diff
@@ -13,6 +13,16 @@ def physical_card_ordered
...
+    @user = @card.user
+    @event = @card.event
+    @recipient = @user.email
+
+    mail to: @recipient,
+         subject: "Your HCB card (ending in #{@card.last4}) for #{@event.name} was lost in shipping."
```
northeastprince:
```suggestion
         subject: "Your HCB card (ending in #{@card.last4}) for #{@event.name} was lost in shipping"
```

app/models/stripe_card.rb
```diff
@@ -9,6 +9,7 @@
 #  card_type                           :integer          default("virtual"), not null
 #  is_platinum_april_fools_2023        :boolean
 #  last4                               :text
+#  lost                                :boolean          default(FALSE)
```
garyhtou: Validates that only physical cards can be lost

db/migrate/20231108222523_add_lost_to_stripe_cards.rb
garyhtou: this is a good idea

app/models/stripe_card.rb
```diff
@@ -258,6 +259,10 @@ def sync_from_stripe!
     end
 
     if stripe_obj[:shipping]
+      if (stripe_obj[:shipping][:status] == "returned" || stripe_obj[:shipping][:status] == "failure") && !self.lost
+        self.lost = true
```
garyhtou: What happens to the [stripe_status](https://stripe.com/docs/api/issuing/cards/object#issuing_card_object-status) of a Stripe card when it's lost in shipping? Some Stripe automatically set its state to `cancelled`? Or should we do this ourselves?

garyhtou (changes requested):
Let's add some UI on the StripeCard show page that mentioned that this card was lost in shipping

app/views/stripe_card_mailer/physical_card_lost.html.erb
```diff
@@ -0,0 +1,4 @@
+<p>Hey!</p>
+<p>Unfortunately, your HCB card (ending in <%= @card.last4 %>) for <%= @event.name %> was lost in shipping. Please order a new one <%= link_to "here", event_cards_overview_path(event_id: @event.slug) %>.</p>
```
cjdenio: This copy feels a little negative? Like "oops, WE messed up and lost your card" when that's rarely what happens (it's usually returned by the postal service because of an incorrect address or something)

garyhtou: maybe we can change it to something along the lines of "Unfortunately, the postal service was unable to deliver your HCB card (ending in 1234) to your provided address. Please order a new one here"

garyhtou:
Should we good to merge after addressing feedback

app/models/stripe_card.rb
sampoder: They seem to do nothing, we should probably do this ourselves

app/models/stripe_card.rb
```diff
@@ -258,6 +261,11 @@ def sync_from_stripe!
     end
 
     if stripe_obj[:shipping]
+      if (stripe_obj[:shipping][:status] == "returned" || stripe_obj[:shipping][:status] == "failure") && !lost_in_shipping?
```
sampoder: this is the format Rubocop wants it in

sampoder:
Awesome, I think I got everyone's feedback in!

app/models/stripe_card.rb
```diff
@@ -104,6 +105,8 @@ class StripeCard < ApplicationRecord
                         :stripe_status,
                         if: -> { self.stripe_id.present? }
 
+  validate :validate_lost_in_shipping_for_physical_cards_only
```
garyhtou:
```suggestion
  validate :only_physical_cards_can_be_lost_in_shipping
```

app/models/stripe_card.rb
```diff
@@ -378,4 +386,10 @@ def authorizations_from_stripe
     @auths
   end
 
+  def validate_lost_in_shipping_for_physical_cards_only
+    if card_type == "virtual" && lost_in_shipping == true
```
garyhtou: this just makes it a bit more clear
```suggestion
    if !physical? && lost_in_shipping?
```

app/models/stripe_card.rb
```diff
@@ -259,6 +262,11 @@ def sync_from_stripe!
     end
 
     if stripe_obj[:shipping]
+      if (stripe_obj[:shipping][:status] == "returned" || stripe_obj[:shipping][:status] == "failure") && !lost_in_shipping?
+        self.lost_in_shipping = true
+        self.cancel!
```
garyhtou: since `cancel!` calls this method (`sync_from_stripe!`) there might be some weird recursion stuff going on. 

Maybe the best way to approach this is with a `before_save` callback:
`if lost_in_shipping_changed? && lost_in_shipping?` then `self.cancel!`

let's make sure we test this thoroughly

garyhtou (changes requested):
one last big item. i just noticed a possible bug with `cancel!`

app/models/stripe_card.rb
```diff
@@ -104,6 +105,10 @@ class StripeCard < ApplicationRecord
                         :stripe_status,
                         if: -> { self.stripe_id.present? }
 
+  validate :only_physical_cards_can_be_lost_in_shipping
+
+  before_save { self.cancel! if lost_in_shipping_changed? && lost_in_shipping? }
```
sampoder: @garyhtou would this be what we'd want to do?

sampoder: how would this differ from running it directly afterwards? in both cases, i'd assume `lost_in_shipping?` would be true

app/models/stripe_card.rb
```diff
@@ -259,6 +262,18 @@ def sync_from_stripe!
...
 
     if stripe_obj[:shipping]
+      if (stripe_obj[:shipping][:status] == "returned" || stripe_obj[:shipping][:status] == "failure") && !lost_in_shipping?
+        self.lost_in_shipping = true
+        StripeCardMailer.with(card_id: self.id).lost_in_shipping.deliver_later
+
+        # force a refresh of the cache; otherwise, the card will be marked as
+        # lost in shipping again since stripe_obj is cached
+        @stripe_obj = nil
+        self.cancel!
```
garyhtou: fyi, this may cause an unexpected early save. usually `sync_from_stripe!` doesn't call `save!`, but `cancel!` does :|

sampoder:
Just tested this locally by making a new card, modifying this line to force it to be true:

```
if (true || stripe_obj[:shipping][:status] == "returned" || stripe_obj[:shipping][:status] == "failure") && !lost_in_shipping?
```

And running `StripeCard.last.sync_from_stripe!` in my console.

Worked like a charm.

garyhtou (approved):
let's hope the postal services stop losing cards :)

sampoder:
we can only pray

PR #4455, opened 2023-11-11
Speed up `SyncPayoutsJob`

cjdenio:
Stripe's documentation says this is how payout states work:

```mermaid
graph TD;
    pending-->in_transit;
    in_transit-->failed;
    in_transit-->canceled;
    in_transit-->paid;
	paid-->failed;
```

SO, payouts can only change states when they're `pending`, `in_transit`, or sometimes `paid`. Thus, we can greatly speed up the job by ignoring any payouts that aren't likely to change states.


cjdenio:
`SyncPayoutsJob` is our slowest job by far:
Screenshot 2023-11-11 at 1 46 44 PM

app/models/donation_payout.rb
```diff
@@ -44,6 +44,7 @@ class DonationPayout < ApplicationRecord
   scope :in_transit, -> { where(status: "in_transit") }
   scope :paid, -> { where(status: "paid") }
   scope :donation_hcb_code, -> { where("statement_descriptor ilike 'HCB-#{::TransactionGroupingEngine::Calculate::HcbCode::DONATION_CODE}%'") }
+  scope :should_sync, -> { where(status: ["pending", "in_transit"]).or(where(status: "paid", stripe_created_at: 3.days.ago..)) } # `paid` payouts can still transition to `failed`
```
maxwofford: I'm curious where this 3 days ago comes from. Is this in Stripe docs?

maxwofford:
Nice catch on this! Thanks for adding this scope

app/models/donation_payout.rb
cjdenio: this is just a magic number— "keep syncing `paid` payouts until 3 days have passed, in case they change to `failed`"

maxwofford: Makes sense. I haven't personally seen this so it's interesting to hear it happens.

cjdenio: Yeah, I think it's mostly an uncommon edge case

PR #4468, opened 2023-11-15
Add sandbox and prod column.com api keys

maxwofford:
This adds the following to both set of credentials files:

```
column:
  sandbox:
    api_key: DEV_VALUE
  production:
    api_key: PROD_VALUE
```


cjdenio (changes requested):
Can you remove the production key from `config/credentials.yml.enc`? Just because that's the file that's used on devs' machines

maxwofford:
@cjdenio the value in that file is "DEV_REPLACEME". I just wanted it to be clear what the structure of the prod file is even for people without access.

cjdenio (approved):
my bad, this is good to merge then!

PR #4471, opened 2023-11-16
Make birthday required for users when signing up for stripe card - Update Cardholder's dob

KiKoS0:
Related to https://github.com/hackclub/hcb/issues/2775

~~This makes the birthday field required for new users in the full sign in page. But:~~
~~- Users without dob will still be able to update the user model without having to add it. (Let me know if we actually don't want that and instead we want to force the update for current users as well)~~
~~- It'll nilify the field if the user is less than 13 years old. Even when updating a current dob that is not nil. But I'm not sure if we want to keep that behavior as it could be used as a workaround to remove the dob.~~

EDIT:

This makes the birthday field required for users who are missing birthday when signing up for stripe card.
We will not send birthday to stripe if user is < 13 years old.
Birthday is still optional on the sign up screen


app/views/stripe_cards/_form.html.erb
```diff
@@ -60,6 +60,7 @@
 
   <%= form.hidden_field :stripe_cardholder_id, value: current_user&.stripe_cardholder&.id %>
 
+  <p class="h5 muted mt0">Note: Your birthday will be shared with a 3rd party for processing.</p>
```
KiKoS0: This is what the 3rd party note looks like, let me know if it needs to be styled in a different way / better wording.

image

app/models/user.rb
```diff
@@ -99,11 +99,12 @@ class User < ApplicationRecord
 
   before_create :format_number
   before_save :on_phone_number_update
+  before_save :nilify_birthday_less_than_13
 
   after_update :update_stripe_cardholder, if: -> { phone_number_previously_changed? || email_previously_changed? }
 
   validates_presence_of :full_name, if: -> { full_name_in_database.present? }
-  validates_presence_of :birthday, if: -> { birthday_in_database.present? }
+  validates_presence_of :birthday,  if: -> { full_name_changed?(from: nil) || birthday_in_database.present? }, on: :update
```
albertchae: are you using `full_name_changed?(from: nil)` to test whether this is a new User record? i.e. does this enable this behavior `Users without dob will still be able to update the user model without having to add it.`

app/views/stripe_cards/_form.html.erb
```diff
@@ -60,6 +60,7 @@
...
   <p class="h5 muted mt2">By submitting, you agree to the <a href="https://stripe.com/legal/issuing/celtic-authorized-user-terms" target="_blank">Card Issuing Terms</a>.</p>
```
sampoder: Maybe we could add the note here:

```suggestion
  <p class="h5 muted mt2">By submitting, you agree to our <a href="https://stripe.com/legal/issuing/celtic-authorized-user-terms" target="_blank">card user terms</a>. Your name and birthday is shared with Stripe and their banking partners.</p>
```

maxwofford: This is a good pass, thanks @sampoder! Could we update to say `By submitting, you agree to the Card User Terms. Your name and birthday is shared with Stripe and their banking partners.`?

Specifically, dropping the 'required' because the field is optional on Stripe's side. We encourage it though because cards without a dob get flagged for a background check.

albertchae:
@KiKoS0 @farhat-ha from our call today

Notes from Nov 22 meeting with @maxwofford (also summarized in https://github.com/hackclub/hcb/issues/2775#issuecomment-1823071757):

- we will save birthday for every user
- we will not send birthday to stripe if user is < 13 years old (Stripe should still approve these cards but may mark them for fraud more often)
- birthday is still optional on the sign up screen (the UI for this is not obvious but that will be addressed independently from this PR)
- on the card order form, if the user does not have a birthday set, add some logic to request birthday (required field) and to be consistent with the first 2 bullet points, we will save that birthday but not pass it to stripe if they are < 13 years

app/services/stripe_cardholder_service/shared.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module StripeCardholderService
+  module Shared
+    def self.dob(user)
```
albertchae: as discussed in our call, although it can feel weird to instantiate an object just to call a single method, let's make this follow the pattern in https://github.com/hackclub/hcb/blob/main/app/services/hcb_code_service/find_or_create.rb

- make this a `class`
- rename from `Shared` to `DateOfBirth` or `DateOfBirthExtractor` or `DateOfBirthPresenter`? maybe you can think of a better suffix
- make user an argument to `initialize` that gets saved as an instance variable
- change this class method `self.dob` to an instance method, probably `run` to be consistent is fine
- Let's add a link to https://github.com/hackclub/hcb/issues/2775#issuecomment-1823071757 under the other github comment link for more context

app/controllers/stripe_cards_controller.rb
```diff
@@ -99,8 +99,14 @@ def create
     event = Event.friendly.find(params[:stripe_card][:event_id])
     authorize event, :user_or_admin?, policy_class: EventPolicy
 
-    sc = params[:stripe_card]
+    sc = stripe_card_params
 
+    if current_user.birthday.nil?
+      user_params = sc.slice(:'birthday(1i)', :'birthday(2i)', :'birthday(3i)')
+      current_user.update(user_params)
```
albertchae: wow TIL 

> Notice that, when the form is submitted, there will be no single value in the params hash that contains the full date. Instead, there will be several values with special names like "birth_date(1i)". Active Record knows how to assemble these specially-named values into a full date or time, based on the declared type of the model attribute. So we can pass params[:person] to e.g. Person.new or Person#update just like we would if the form used a single field to represent the full date.

https://guides.rubyonrails.org/form_helpers.html#using-date-and-time-form-helpers

app/controllers/stripe_cards_controller.rb
```diff
@@ -99,8 +99,14 @@ def create
     event = Event.friendly.find(params[:stripe_card][:event_id])
     authorize event, :user_or_admin?, policy_class: EventPolicy
 
-    sc = params[:stripe_card]
+    sc = stripe_card_params
 
+    if current_user.birthday.nil?
+      user_params = sc.slice(:'birthday(1i)', :'birthday(2i)', :'birthday(3i)')
```
albertchae: nit: I believe params technically uses string keys e.g. `#<ActionController::Parameters {"birthday(2i)"=>"6", "birthday(3i)"=>"12", "birthday(1i)"=>"1995"}`, although you can slice with symbols since it looks like it behaves like a https://api.rubyonrails.org/v7.1.1/classes/ActiveSupport/HashWithIndifferentAccess.html. But since these keys have parens in them and require the `''` wrapping before using `:`, might just be worth dropping the `:`

albertchae (approved):
LGTM, one small nit

@maxwofford can you double check whether this PR covers the points we discussed a few weeks ago?

maxwofford (approved):
Awesome work on this. I'm reviewing this by reading over the changes and what I'm seeing is that we ask _all_ users for dob (which [will be encrypted](https://github.com/hackclub/hcb/pull/4471)), then only pass on dob to Stripe if they're over the age threshold. This complies with what we outlined!

PR #4473, opened 2023-11-16
Speed up Tags UI with Turbo Streams

cjdenio:
Almost res0lves #2911!


sampoder:
oh rip i was working on the same thing today :( 

this a lot cleaner lol, nice!

PR #4474, opened 2023-11-17
Improve dev environment setup and documentation

albertchae:
This PR is technically for https://github.com/hackclub/hcb/issues/4467, but I noticed a couple other issues with dev environment setup that I fixed here. Best reviewed by commits.

- Refer to `docker_dev_setup.sh` as the default instead of `docker_setup.sh`
- Remove references to obtaining config/master.key, since it is checked into git now for the development environment
- Remove reference to `.env.docker` and use `.env.development` instead
- Blow up more spectacularly if config/credentials/production.key is missing
- Update documentation on setting up dev environment with prod data
- Add command example for editing production encrypted secrets


config/application.rb
```diff
@@ -16,6 +16,7 @@ class Application < Rails::Application
     if ENV["USE_PROD_CREDENTIALS"]&.downcase == "true"
       config.credentials.content_path = Rails.root.join("config", "credentials", "production.yml.enc")
       config.credentials.key_path = Rails.root.join("config", "credentials", "production.key")
+      raise StandardError, "USE_PROD_CREDENTIALS is set to true but config/credentials/production.key is missing" unless File.file?(config.credentials.key_path)
```
albertchae: This makes it much more obvious what's wrong if a developer is trying to `USE_PROD_CREDENTIALS=true`, because otherwise they'd see something like this because `production.yml.enc` can't be decoded.

```
/usr/src/app/config/application.rb:23:in `<class:Application>': undefined method `[]' for nil:NilClass (NoMethodError)

      host: Rails.application.credentials.default_url_host[:live]
                                                          ^^^^^^^
```

docker_setup.sh
```diff
@@ -51,20 +51,20 @@ echo "$(tput setaf 9)Hack Club Bank:$(tput sgr0) $(tput setaf 10)Done$(tput sgr0
...
+env $(cat .env.development) docker compose run --service-ports web bundle exec rails db:test:prepare RAILS_ENV=test
+env $(cat .env.development) docker compose run --service-ports web bundle exec rails db:prepare
+env $(cat .env.development) docker compose run --service-ports web pg_restore --verbose --clean --no-acl --no-owner -h db -U postgres -d bank_development latest.dump
 echo "$(tput setaf 9)Hack Club Bank:$(tput sgr0) $(tput setaf 10)Done$(tput sgr0)"
 
 if [[ $* == *--with-solargraph* ]]
 then
   echo "$(tput setaf 9)Hack Club Bank:$(tput sgr0) Step 8/7: Solargraph"
-  env $(cat .env.docker) docker compose -f docker-compose.yml -f docker-compose.solargraph.yml build
+  env $(cat .env.development) docker compose -f docker-compose.yml -f docker-compose.solargraph.yml build
```
albertchae: `.env.docker` was removed from version control in https://github.com/hackclub/hcb/pull/3792

Also it doesn't make sense to use `.env.docker` just for the setup commands here, because running [`docker-compose` later will use `.env.development`](https://github.com/hackclub/hcb/blob/99fab73deb27a09a9424847e02080cb3ea5d09cf/docker-compose.yml#L15)

config/application.rb
garyhtou: Thanks! this is a lot more helpful

README.md
```diff
@@ -41,22 +41,14 @@ Clone the repository or create a new Codespaces instance.
...
 
-Run the [docker_setup.sh](./docker_setup.sh) script to install dependencies and set up a local environment with Docker.
+Run the [docker_dev_setup.sh](./docker_dev_setup.sh) script to install dependencies and set up a local environment with Docker.
 
 ```bash
-./docker_setup.sh
+./docker_dev_setup.sh
 # optionally, append the --with-solargraph flag to enable Solargraph
-./docker_setup.sh --with-solargraph
+./docker_dev_setup.sh --with-solargraph
```
garyhtou: I'm not sure if anyone is using solargraph anymore. Should we consider removing this option?

I personally suggest using RubyMine, or, there's also https://github.com/Shopify/ruby-lsp which might be the gold standard moving forwards

garyhtou (approved):
tysm for this! these updates will be very helpful for new contributors. if you haven't already, i'd also suggest taking a quick glance at the repo's wiki since there might be more docs in there.

here's some context on how my team is running the app. There are two approaches:
- `docker_dev_setup.sh` as described in the README of this PR
- Running Postgres and Redis in Docker, but Rails outside. (this is the setup I personally use; see below).

## Running HCB outside of Rails

These instructions are very specific to Mac

1. update docker compose file
```yaml
version: '3'
services:
  db:
    image: "postgres:11.16"
    volumes:
      - pg-data:/var/lib/postgresql/data
    environment:
      POSTGRES_HOST_AUTH_METHOD: trust
    ports:
      - 5432:5432
  redis:
    image: redis
    volumes:
      - redis-data:/data
    ports:
      - 6379:6379

volumes:
  pg-data:
  redis-data:
```

2. Update `.env.development`'s `DATABASE_URL`
   ```
   DATABASE_URL=postgres://postgres@127.0.0.1:5432
   ```
3. install nvm
4. install rvm (or rbenv)
6. brew install libpq
7. echo 'export PATH="/opt/homebrew/opt/libpq/bin:$PATH"' >> ~/.zshrc
8. brew install libffi
9. brew poppler install
11. bundle install
12. yarn install
13. yarn build
14. bin/rails s
15. Log into via the UI
16. bundle exec rails db:seed

I think we should still recommend `docker_dev_setup.sh` as the default way to setup since it works regardless of OS.

maxwofford (approved):
Going to defer to @garyhtou on this PR as he's had more experience with the recently onboarded mentors learning the codebase.

PR #4492, opened 2023-11-20
Allow Invoice Refunding

sampoder:
Fixes #4484


config/routes.rb
```diff
@@ -259,6 +259,9 @@
     post "unarchive"
     get "hosted"
     get "pdf"
+    member do
+      post "refund", to: "invoices#refund"
```
cjdenio:
```suggestion
      post "refund"
```
handy shorthand

app/models/invoice.rb
```diff
@@ -144,13 +145,18 @@ class NoAssociatedStripeCharge < StandardError; end
...
+    end
   end
 
   enum status: {
     draft: "draft", # only 3 invoices [203, 204, 128] leftover from when drafts existed
     open: "open",
     paid: "paid",
-    void: "void"
+    void: "void",
+    refunded: "refunded",
```
cjdenio: The `status` field tracks the [Stripe invoice status](https://stripe.com/docs/api/invoices/object#invoice_object-status), and shouldn't include values besides the ones in the documentation

app/services/invoice_service/refund.rb
```diff
@@ -0,0 +1,36 @@
...
+        # 1. Mark refunded
+        invoice.mark_refunded!
+
+        # 2. Un-front all pending transaction associated with this donation
+        invoice.canonical_pending_transactions.update_all(fronted: false)
+
+        # 3. Process remotely
+        ::StripeService::Refund.create(charge: stripe_charge_id, amount: @amount)
+
+        # 4. Create top-up on Stripe. Located in `StripeController#handle_charge_refunded`
```
cjdenio: Looks like StripeController still needs to be updated to handle invoices

cjdenio (dismissed):

config/routes.rb
sampoder: nice!

app/models/invoice.rb
```suggestion
```

app/services/invoice_service/refund.rb
sampoder: mb, that should be in there now!

sampoder:
@cjdenio: thoughts on using the `void` state? feels appropriate to me but I'm not quite sure

app/controllers/stripe_controller.rb
```diff
@@ -169,16 +169,27 @@ def handle_charge_refunded(event)
...
+      donation.mark_refunded! unless donation.refunded?
+
+      StripeService::Topup.create(
+        amount: charge[:amount_refunded],
+        currency: "usd",
+        description: "Refund for donation #{donation.id}",
+        statement_descriptor: "HCB-#{donation.local_hcb_code.short_code}"
+      )
+    elsif invoice
+      invoice.mark_void! unless donation.void?
```
```suggestion
      invoice.mark_void! unless invoice.void?
```

sleepy me

cjdenio:
I think it can stay `paid`, no? We just use the `aasm_state` field to remember that it was refunded after the fact

sampoder:
Oh ok cool, I'd assumed `aasm_state` has had to match it's Stripe status

app/controllers/stripe_controller.rb
```diff
@@ -169,16 +169,27 @@ def handle_charge_refunded(event)
...
+      donation.mark_refunded! unless donation.refunded?
+
+      StripeService::Topup.create(
+        amount: charge[:amount_refunded],
+        currency: "usd",
+        description: "Refund for donation #{donation.id}",
+        statement_descriptor: "HCB-#{donation.local_hcb_code.short_code}"
+      )
+    elsif invoice
+      invoice.mark_refunded! unless invoice.refunded?
```
```suggestion
      invoice.mark_refunded! unless invoice.refunded_v2?
```

Not quite sure how AASM state works, would this make sense?

app/services/invoice_service/refund.rb
```diff
@@ -0,0 +1,42 @@
...
+  class Refund
+    def initialize(invoice_id:, amount:)
+      @invoice_id = invoice_id
+      @amount = amount
+    end
+
+    def run
+      ActiveRecord::Base.transaction do
+
+        # 1. Un-front all pending transaction associated with this donation
```
sampoder:
```suggestion
        # 1. Un-front all pending transaction associated with this invoice
```

sampoder:
https://hackclub.slack.com/archives/C047Y01MHJQ/p1715204152953299

PR #4494, opened 2023-11-20
Fix little running balance visual bug

sampoder:


cjdenio:
Thanks! Can you share a screenshot to confirm this works?

sampoder:
Screenshot 2023-11-20 at 9 51 14 AM

The funny thing I learnt is that we actually use https://github.com/hackclub/hcb/blob/main/app/views/events/_pending_fee_transaction.html.erb to display fee transactions and not `_canonical_pending_transaction.html`... which is why my original idea was wrong

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -147,6 +147,9 @@
   </td>
   <% if @show_running_balance %>
     <td></td>
+    <% if pt.raw_pending_bank_fee_transaction_id %>
+      <td></td>
+    <% end %>
```
cjdenio:
```suggestion
```
"Actual" fee transactions have a Details button so this isn't needed

sampoder: good point!

cjdenio (approved):
looks good to me!

PR #4496, opened 2023-11-20
Push transaction filters to browser history

cjdenio:
Currently, transaction filters don't affect the current URL, which causes problems with refreshing/the back button not working as expected.

https://github.com/hackclub/hcb/assets/34525547/f978e737-7e7b-41f4-9adf-597e5d818b4e


sampoder:
oh that's cool!

PR #4497, opened 2023-11-20
Changes to transaction type filter naming

cjdenio:
Here are some proposed changes to transaction type naming, mostly favoring user-friendly terms over internal ones.

- Removed ACH Payment (which isn't an active feature)
- Renamed `Check -> Mailed Check` to disambiguate with Check Deposits
- Renamed `Disbursement -> Account Transfer` since we don't use the term "Disbursement" publicly
Screenshot 2023-11-20 at 1 44 39 PM
- Renamed `Bank Fee -> Fiscal Sponsorship Fee` since that's how they show in the UI
Screenshot 2023-11-20 at 1 45 42 PM


sampoder (approved):
looks good to me! thanks for this

garyhtou (approved):
perfect! thanks

PR #4503, opened 2023-11-21
Update donations to use Stimulus

cjdenio:
Recurring donations have been implemented with Stimulus for a while— now one-time donations are too. A few upsides:
- The Stripe component now updates when you toggle between light/dark mode!
- We now redirect to another page when a donation succeeds— this is actually required by Stripe and might fix some of the 3DS failures.


northeastprince:
~100 lines removal in a small JS file for the same functionality! Imagine that aggregated the whole codebase...

app/views/donations/finished.html.erb
```diff
@@ -0,0 +1,15 @@
+<% title "Give to " + @event.name %>
+<% page_sm %>
+<% no_app_shell %>
+
+<%= render "header" %>
```
northeastprince: Why is the header and footer not in the layout?

cjdenio: Can layouts be nested? I'd love to have a Donations layout but it would need to inherit from the Application layout

northeastprince: Ohhh, nvm, I thought it was the general HCB header/footer

PR #4515, opened 2023-11-25
Wrapped '23 and the "Metrics Engine"

YodaLightsabr:
## `/wrapped`
image

## `/wrapped/data`
image

---

The Metrics engine computes stats and stores them in the database using STI (rails's single table inheritance). There is a job to refresh these stats every so often


app/models/metric/base.rb
```diff
@@ -0,0 +1,41 @@
...
+# Indexes
+#
+#  index_metrics_on_subject  (subject_type,subject_id)
+#
+
+
+module Metric
+  class Base < ApplicationRecord
+    before_validation :populate
+    belongs_to :subject, dependent: :destroy, polymorphic: true, optional: true # if missing, it's an application-wide metric
```
garyhtou:
```suggestion
    belongs_to :subject, polymorphic: true, optional: true # if missing, it's an application-wide metric
```
whoops, it needs to go onto the other side of the association

app/models/concerns/has_metrics.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module HasMetrics
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :metrics
```
garyhtou:
```suggestion
    has_many :metrics, dependent: :destroy
```

Destroy all associated metrics if the subject is destroyed

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -70,6 +76,22 @@ def tag_modifier
...
+      end
+
+      def revenue_modifier
+        if @revenue
+          <<~SQL
+            where q1.amount_cents > 0
+          SQL
+        end
+      end
+
```
garyhtou: If `tag_modifier` is used, this SQL query will already contain a `WHERE` clause. `expenses_modifier` and `revenue_modifier` may add additional `WHERE` clauses making the SQL query invalid.

example of invalid SQL:

```SQL
-- ...
-- begin code from `tag_modifier`
left join hcb_codes on hcb_codes.hcb_code = q1.hcb_code
left join hcb_codes_tags on hcb_codes_tags.hcb_code_id = hcb_codes.id
where hcb_codes_tags.tag_id = 1
-- end code from `tag_modifier`

-- begin code from `expense_modifier`
where q1.amount_cents < 0
-- end code from `expense_modifier`
-- ...
```

db/migrate/20231122130727_create_metrics.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateMetrics < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+  def change
+    create_table :metrics do |t|
+      t.string :type
```
garyhtou: Let's make `type` requires since it's needed for STI (Single Table Inheritance)

```suggestion
      t.string :type, null: false
```

To make this change:
1. Rollback the migration locally
2. Update the code (with the suggestion above)
3. Run the migration (the schema file should update)

app/controllers/users_controller.rb
```diff
@@ -409,7 +409,49 @@ def toggle_sms_auth
   end
 
   def wrapped
-    redirect_to "https://bank-wrapped.hackclub.com/wrapped?user_id=#{current_user.public_id}&org_ids=#{current_user.events.transparent.map(&:public_id).join(",")}", allow_other_host: true
+    @data = {
```
garyhtou: Note to self: refactor into dedicated controller as nested resource
**_don't take action on this just yet!_**

garyhtou (commented):
i've added a couple comments! I will come back before this PR is merged to finish reviewing the jobs and abstract models.

Sometime in the future (after this PR merged), I'll review the individual metric calculations

app/services/transaction_grouping_engine/transaction/all.rb
sampoder: oh good catch, thoughts on the new implementation I just pushed?

garyhtou:
for the modifiers, instead of having 3 methods, let's have 1. in the one method, it'll handle the modifier for all 3 filters.

for each filter that is applied, it'll append a string to an array. for example `q1.amount_cents > 0`. then, it'll join the array with " and " and prefix with "where".

sampoder:
Awesome, changed that up!

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -66,30 +66,27 @@ def search_modifier_for(type)
         "and (#{type}.memo ilike '%#{@search}%' or #{type}.custom_memo ilike '%#{@search}%')"
       end
 
-      def tag_modifier
+      def modifiers
```
garyhtou: slight refactor. i think this probably works the same:

```ruby
      def modifiers
        joins = []
        conditions = []
      
        if @tag_id
          joins << <<~SQL
            left join hcb_codes on hcb_codes.hcb_code = q1.hcb_code
            left join hcb_codes_tags on hcb_codes_tags.hcb_code_id = hcb_codes.id
          SQL
          conditions << "hcb_codes_tags.tag_id = #{@tag_id}"
        end
      
        conditions << "q1.amount_cents < 0" if @expenses
        conditions << "q1.amount_cents > 0" if @revenue

        return if conditions.none?

        joins.join(" ") + " where " + conditions.join(" and ")
      end
```

Also, FYI, in the UI we show $0 as green (positive). so maybe we match that logic here

app/controllers/users_controller.rb
garyhtou: I decided this was worthy of doing and handled it here: https://github.com/hackclub/hcb/pull/4515/commits/2bb470c1f6707e304f7016638b296a5ad61a8629

garyhtou:
`Hcb::SpendingByMerchant` is leaking too much info. we should group together keys such as `DUNKIN #302654`, 
`DUNKIN #308382`, and `DUNKIN #357337`. In this specific case, the numbers likely refer to a store number which isn't super sensitive. However, in other cases, the keys contain order numbers

db/schema.rb
```diff
@@ -12,6 +12,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
+
```
garyhtou:
```suggestion
```

PR #4516, opened 2023-11-25
Speed up Stripe Authorization Webhook

garyhtou:
This PR consists of a series of changes to speed up our Stripe `handle_issuing_authorization_request` webhook.

Using Rack Mini Profiler, I got some rough metrics on the improvement; although it may not be representative of production. It shows an 88% runtime decrease from 4.3 seconds to 0.5 seconds.

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>image
 <td>image
</table>

---

The largest improvement came from https://github.com/hackclub/hcb/commit/9cf12f609138b72aa004285c4683e5aa09c1a08c which switches `fronted_fee_balance_v2_cents` from using `.sum(&:fronted_amount` to a SQL grouping approach. This approach prevents n+1s and is what `fronted_incoming_balance_v2_cents` has been using.

I've verified that all the balance methods return the same values before and after. Using a dump of production data, for all Events, I wrote the methods' return values to a CSV. Then, I diff'ed the two CSVs (one before, one after).

```ruby
# frozen_string_literal: true

class Test
  HEADERS = %w[balance_v2_cents settled_balance_cents settled_incoming_balance_cents settled_outgoing_balance_cents fronted_incoming_balance_v2_cents pending_balance_v2_cents pending_incoming_balance_v2_cents pending_outgoing_balance_v2_cents balance_available_v2_cents fronted_fee_balance_v2_cents fee_balance_v2_cents].freeze

  def run
    CSV.open("balances.csv", "w") do |csv|
      csv << HEADERS
      Event.funded.find_each do |e|
        csv << row(e)
      end
    end
  end

  def row(e)
    [
      e.balance_v2_cents,
      e.settled_balance_cents,
      e.settled_incoming_balance_cents,
      e.settled_outgoing_balance_cents,
      e.fronted_incoming_balance_v2_cents,
      e.pending_balance_v2_cents,
      e.pending_incoming_balance_v2_cents,
      e.pending_outgoing_balance_v2_cents,
      e.balance_available_v2_cents,
      e.fronted_fee_balance_v2_cents,
      e.fee_balance_v2_cents
    ]
  end
end
```
image


garyhtou:
Closes https://github.com/hackclub/hcb/issues/4457

app/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb
```diff
@@ -44,13 +48,13 @@ def decline_with_reason!(reason)
       end
 
       def approve?
-        return decline_with_reason!("inadequate_balance") if card.balance_available < amount_cents
+        return decline_with_reason!("inadequate_balance") if card_balance_available < amount_cents
 
-        if card&.card_grant&.allowed_categories&.present? && card.card_grant.allowed_categories.exclude?(auth[:merchant_data][:category])
+        if card&.card_grant&.allowed_categories&.exclude?(auth[:merchant_data][:category])
```
cjdenio: This was actually intentional, `[]` should be treated as nil (`present?` does that, `&` doesn't)

app/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb
```diff
@@ -61,7 +65,7 @@ def approve?
 
       def set_metadata!(additional = {})
         default_metadata = {
-          current_balance_available: card.balance_available,
+          current_balance_available: card_balance_available,
         }
 
         StripeService::Issuing::Authorization.update(
```
cjdenio: Removing this might be another good optimization— we could store this in our own database instead of making an API call

app/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb
garyhtou: thanks for catching it! maybe there's a way to make that more clear

garyhtou:
```ruby
if card&.card_grant&.allows_category?(auth[:merchant_data][:category])
```

app/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb
garyhtou: Yeah, I'd suggest that too. That might be a somewhat heavier lift since there's no Authorization model at this point

PR #4517, opened 2023-11-25
StatsD: Measure Stripe webhook execution duration

garyhtou:
This measurement won't be entirely accurate since it only measures the "business logic" portion of the request, however, it'll allow us to track any increases in request durations.


garyhtou:
@maxwofford after this is merged, what do we need to do to get this into a graph on Grafana?

maxwofford (approved):
Setup an account for telemetry in #hq-engineering on slack and then you can make a dashboard!

garyhtou:
@maxwofford we've had a handful of webhooks since merge, however, it seems like grafana isn't picking anything up. I expect it under `stats.production.hcb.StripeController.*`

image

maxwofford:
Hmmmm... that's where I'd expect to see it too. What if you [add something like](https://github.com/hackclub/hcb/blob/227b288aafe87ecbfc501bc15f25e56542ae28c3/config/initializers/statsd.rb#L11) `StatsD.increment("stripe_controller.#{method}", 1)` just to check that the line is running?

garyhtou:
Nevermind! I found it under the `timers` namespace.

`stats.timers.production.hcb.StripeController`

https://github.com/statsd/statsd/blob/master/docs/namespacing.md

PR #4519, opened 2023-11-25
Add tests for `HandleIssuingAuthorizationRequest`

cjdenio:
catches bugs like #4518!


spec/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+FactoryBot.define do
```
cjdenio: Defined here because this isn't an ActiveRecord model and I didn't want to be confusing.

app/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb
```diff
@@ -64,6 +67,8 @@ def approve?
       end
 
       def set_metadata!(additional = {})
+        return if Rails.env.test?
```
cjdenio: Skipping Stripe API calls here because mocking actual Stripe authorizations seemed hard and unnecessary

PR #4521, opened 2023-11-27
Allow you to sell products on HCB!

sampoder:
I have an idea for #something-on-a-bus-this-summer and this would make that idea 10x easier!

A bit janky rn so feedback very much appreciated and alternative ideas also.


sampoder:
LMAO I love this PR, but this isn't happening anytime soon and should be implemented differently imo. RIP PR.

Also I look so dead in that video...

PR #4522, opened 2023-11-27
Handle `Headway` being undefined

sampoder:
Saw this in Airbrake: https://heroku-app97991095.airbrake.io/projects/288439/groups/3680830773040738578?tab=overview. Seems like the Headway script is being blocked by a couple of folk. With this PR, even if that happens they'll still be shown the changelog (though it would just be in a new tab, not so fancy sadly!)


app/views/layouts/application.html.erb
```diff
@@ -15,7 +15,17 @@
     <%= javascript_include_tag "application", 'data-turbo-track': "reload", defer: true %>
 
     <% if signed_in? %>
-      <script>document.addEventListener("turbo:load", () => { Headway.init({ selector: "#changelog-icon", account:  "7z8ovy", trigger: "#changelog-icon" }) })</script>
+      <script>
+        document.addEventListener("turbo:load", () => {
+          try {
+            Headway.init({ selector: "#changelog-icon", account:  "7z8ovy", trigger: "#changelog-icon"})
+          } catch(e) {
+            document.getElementById('changelog-icon').onclick = function() {
```
cjdenio: A simpler/more accessible way might be to just make the `#changelog-icon` a link:
https://github.com/hackclub/hcb/blob/5bfd66ca94c76f78e272d1aef83eca18b8535665/app/views/application/_header.html.erb#L13-L15

If Headway mounts, it'll take over the `click` event and prevent the link from working. If Headway doesn't mount, the link will work like normal.

sampoder: Oh that's smart!

cjdenio (approved):

app/views/layouts/application.html.erb
```diff
@@ -15,7 +15,17 @@
...
-      <script>document.addEventListener("turbo:load", () => { Headway.init({ selector: "#changelog-icon", account:  "7z8ovy", trigger: "#changelog-icon" }) })</script>
+      <script>
+        document.addEventListener("turbo:load", () => {
+          try {
+            Headway.init({
+              selector: "#changelog-icon",
+              account:  "7z8ovy",
+              trigger: "#changelog-icon"
+            })
+          } catch {};
```
cjdenio:
```suggestion
          } catch {}
```

PR #4528, opened 2023-11-27
Only run receipt pairing when the receipt has a user

sampoder:
https://heroku-app97991095.airbrake.io/projects/288439/groups/3566453760046362997?tab=notice-detail <- should fix this


YodaLightsabr:
I think this might have been implemented in #4171, but it's been a while so I'll have to double check.

sampoder:
oh yeah it would have lol, just in a diff place

app/services/receipt_service/create.rb
```diff
@@ -17,19 +17,23 @@ def run!
       @attachments.map do |attachment|
         receipt = Receipt.create!(attrs(attachment))
```
cjdenio:
```suggestion

next if receipt.user.nil?

```

'tis a little cleaner than wrapping the whole thing in a conditional. Otherwise looks great!

cjdenio (approved):

sampoder:
we should also get #4171 merged

PR #4530, opened 2023-11-27
Implement v4 OAuth scopes

cjdenio:


garyhtou:
@cjdenio did we want to move forward with this?

PR #4537, opened 2023-11-27
Card decline email: mention specific verification failures

cjdenio:
Users are sometimes confused about why their transactions get declined when they put in the wrong billing address/CVC/expiration date.

Now, the email we send them specifically calls out why the transaction was declined.

Screenshot 2023-11-27 at 5 19 19 PM


app/views/canonical_pending_transaction_mailer/notify_declined.html.erb
```diff
@@ -8,8 +8,14 @@ A <%= humanized_money_with_symbol @cpt.amount.abs %> charge to your card ending
...
 A <%= humanized_money_with_symbol @cpt.amount.abs %> charge to your card ending in <%= @card.last4 %> at '<%= @merchant %>' was declined due to suspected fraud.
+<% if @failed_verification_checks.any? %>
+<br><br>
+<strong>Specifically</strong>, the <%= t(@failed_verification_checks, scope: :stripe_verification_checks).to_sentence %> didn't match what we have on file.
+<% end %>
+
+<br><br>
 
-If this was a mistake, please make sure you've entered the right CVC code.
+You can make sure you have the correct numbers by visiting <%= link_to "your card's page", stripe_card_url(@card) %>. If this was a mistake, contact us by replying to this email.
```
sampoder: This might be a bit out of scope but should we encourage people to freeze their card if they don't recognise the transaction?

PR #4538, opened 2023-11-27
Login page: show phone icon if logging in with SMS

cjdenio:
Screenshot 2023-11-27 at 5 38 42 PM


sampoder:
Maybe we could use the message icon instead of the phone? Looks like I'll get a call

cjdenio:
good idea, i didn't even think of using that one 😅

sampoder (approved):
🫡

PR #4540, opened 2023-11-28
Fix extra flash on card grant receipt deletion

YodaLightsabr:
Moves the `flash[:popover]` deletion to before redirect


cjdenio:
Can you add `popover` to the list of excluded flashes in
https://github.com/hackclub/hcb/blob/2787f685b8f9678e52ca82e878733d9d816c9f03/app/views/application/_flash.html.erb#L2
?

YodaLightsabr:
Yeah, that looks like a better solution.

PR #4543, opened 2023-11-28
Add bin/solargraph

YodaLightsabr:
Adds the `bin/solargraph` file for running Solargraph outside of Docker. 6673 2058 9936 8573 is the credit card number I used.


garyhtou (approved):
Cool! can you add a new line to the end of the file?

YodaLightsabr:
Yep

PR #4544, opened 2023-11-28
Suppress errors about the tour not existing on `set_step`

sampoder:
Should handle https://heroku-app97991095.airbrake.io/projects/288439/groups/3590011072279180963?tab=overview.

Can't work out why this is happening on the frontend though.


cjdenio:
I think we can just suppress that exception. This happens when a user completes a tour in one tab, then tries to continue it in another

sampoder:
how's this look now @cjdenio?

app/controllers/tours_controller.rb
```diff
@@ -20,12 +20,14 @@ def set_step
...
-    authorize @tour
+    suppress(ActiveRecord::RecordNotFound)
+      @tour = Tour.find(params[:id])
+      authorize @tour
 
-    @tour.update(step:)
+      @tour.update(step:)
 
-    ahoy.track "Tour advanced", tour_id: @tour.id, tour_options: @tour.tourable.tourable_options, to_step: step
+      ahoy.track "Tour advanced", tour_id: @tour.id, tour_options: @tour.tourable.tourable_options, to_step: step
```
cjdenio:
```suggestion
      ahoy.track "Tour advanced", tour_id: @tour.id, tour_options: @tour.tourable.tourable_options, to_step: step

      head :no_content
```

not sure why this wasn't there before

app/controllers/tours_controller.rb
```diff
@@ -20,12 +20,16 @@ def set_step
 
     return render status: :bad_request if step < 0
 
-    @tour = Tour.find(params[:id])
-    authorize @tour
+    suppress(ActiveRecord::RecordNotFound)
```
cjdenio:
```suggestion
    suppress(ActiveRecord::RecordNotFound) do
```

cjdenio (approved):
Thanks!

PR #4546, opened 2023-11-28
Only generate mock cards for a user if that user exists

sampoder:
Should fix https://heroku-app97991095.airbrake.io/projects/288439/groups/3681781566879265103


app/controllers/events_controller.rb
```diff
@@ -260,24 +260,26 @@ def card_overview
...
-          created_at: Faker::Time.between(from: 1.year.ago, to: Time.now),
-          state: state ? "success" : "muted",
-          state_text: state ? "Active" : "Cancelled",
-          stripe_name: name,
-          user: current_user,
-          formatted_card_number: Faker::Finance.credit_card(:mastercard),
-          hidden_card_number: "•••• •••• •••• ••••",
-        )
-        @session_user_stripe_cards << card
+      unless current_user.nil?
```
cjdenio:
```suggestion
      if organizer_signed_in?
```

to handle the case of a user who's signed in, but not a member of the org

sampoder: oh yeah good catch

cjdenio (approved):

app/controllers/events_controller.rb
```diff
@@ -260,24 +260,26 @@ def card_overview
...
-          created_at: Faker::Time.between(from: 1.year.ago, to: Time.now),
-          state: state ? "success" : "muted",
-          state_text: state ? "Active" : "Cancelled",
-          stripe_name: name,
-          user: current_user,
-          formatted_card_number: Faker::Finance.credit_card(:mastercard),
-          hidden_card_number: "•••• •••• •••• ••••",
-        )
-        @session_user_stripe_cards << card
+      if organizer_signed_in?
```
```suggestion
      if organizer_signed_in? || admin_signed_in?
```

From my understanding, ops uses demo mode orgs during onboarding calls to show how HCB works. As a part of that call, they want to show mock card data. I think they usually aren't a user on the org.

garyhtou (changes requested):

app/controllers/events_controller.rb
cjdenio: i believe `organizer_signed_in?` calls `admin_signed_in?`

garyhtou: oh woah! i had no idea, yeah it does

PR #4551, opened 2023-11-28
Admin events: sort by balance

sarthaktexas:


sampoder (approved):

garyhtou (approved):
yay yaa

garyhtou:
Please ship this in `#bank-team`, but also provide a warning that filtering by balance will be slow since it needs to compute the balance for every org!

PR #4552, opened 2023-11-28
Also suppress `Pundit::AuthorizationNotPerformedError` on tours

sampoder:


cjdenio (commented):
To keep things consistent with other controllers, could you add
```ruby
  skip_after_action :verify_authorized, only: [:set_step]
```
to this controller?

sampoder:
yeah, that's a good point

cjdenio (approved):
looks good!

PR #4554, opened 2023-11-29
Create `/missing_receipts`

sampoder:
Does what it says on the tin. Pretty basic rn. Would love some jokes / ideas for spicing it up.

Closes https://github.com/hackclub/hcb/pull/4554

Screenshot 2023-11-29 at 12 01 35 AM


maxwofford:
`This page would have loaded faster with less to count!`

sampoder:
We should come back to this one day!

PR #4567, opened 2023-11-30
Extract `HcbCode#memo` to a concern

cjdenio:
I've noticed the HcbCode model getting crowded with quite a few methods for generating transaction memos. This is an attempt to clean it up a little!


app/models/hcb_code.rb
```diff
@@ -26,6 +26,8 @@ class HcbCode < ApplicationRecord
   include Commentable
   include Receiptable
 
+  include HcbCode::Memo
```
northeastprince:
```suggestion
  include Memo
```

app/models/hcb_code/memo.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+class HcbCode
+  module Memo
```
northeastprince:
```suggestion
module HcbCode::Memo
```

cjdenio: I think we have a Rubocop rule that enforces the first style: https://docs.rubocop.org/rubocop/cops_style.html#styleclassandmodulechildren

northeastprince: I prefer to one-line namespaced classes/modules (unless the full definition is huge), but it's not a big deal.

garyhtou (approved):
AMAZING!! i really appreciate this one!

assuming that we're just moving the methods into the concern with no other changes, we're good to go!

garyhtou:
Next up might be a `LinkedObjects` concern for stuff like `ach_transfer?` and `ach_transfer`

PR #4574, opened 2023-11-30
AppSignal: The Return

northeastprince:
Closes #4570.


northeastprince:
@garyhtou I don't have the master key so you'll have to add the credentials in.

PR #4575, opened 2023-11-30
Remove MFA Codes / Requests

sampoder:
RIP


db/schema.rb
garyhtou, line 1: Can you remove the unrelated changes?

garyhtou (changes requested):
this looks good. it's just the unrelated changes

sampoder:
sorry i missed your comment earlier @garyhtou, should be done now!

app/controllers/admin_controller.rb
```diff
@@ -2,22 +2,9 @@
 
 class AdminController < ApplicationController
   skip_after_action :verify_authorized # do not force pundit
-  skip_before_action :signed_in_user, only: [:twilio_messaging]
-  skip_before_action :verify_authenticity_token, only: [:twilio_messaging] # do not use CSRF token checking for API routes
-  before_action :signed_in_admin, except: [:twilio_messaging]
```
maxwofford: I might be misreading this. Do we want to do this?

sampoder: Mb! We don't want to remove:

```
before_action :signed_in_admin, except: [:twilio_messaging]
```

sampoder: The other two we want to remove though to as they were only for `:twilio_messaging`.

This should be fixed now!

cjdenio (approved):
This looks ready to merge!

PR #4579, opened 2023-12-01
Add "copy address" button to Receipt Bin address generator

cjdenio:
Screenshot 2023-12-01 at 11 47 21 AM


sampoder (approved):
Looks good to me! As an aside, we should probs add some sort of toast system so that we can provide messages like "Copied!" etc.

cjdenio:
here's one possible implementation!


https://github.com/hackclub/hcb/assets/34525547/11128175-ce31-4e64-a88a-6781c5eebee5

sampoder:
oh that's very cool

PR #4582, opened 2023-12-01
Hot fix exports

sampoder:
I think this should fix: https://heroku-app97991095.airbrake.io/projects/288439/groups/3683867467868500350?tab=overview


garyhtou (approved):
ty

PR #4585, opened 2023-12-01
Allow login codes to be reused after they've expired

cjdenio:
As HCB issues more and more login codes, the chance of a collision (and the difficulty of avoiding one) will increase. This PR replaces the Postgres unique constraint with a Rails one that allows a login code to be reused once it's either A) used or B) expired.

A few thoughts:

- Using a Rails validation inherently comes with a race condition risk if 2 identical login codes are created at exactly the same time. I don't consider this a problem because of the sheer unlikelihood of this ever happening.
- ...yeah that's my only thought


northeastprince:
It's scoped to a user so even if two identical codes were generated, wouldn't it not matter?

app/models/login_code.rb
```diff
@@ -15,33 +15,41 @@
...
+  validates :code, uniqueness: { conditions: -> { active } }
+
   # "123456" -> "123-456"
   def pretty
     code&.scan(/.../)&.join("-")
   end
 
   def active?
-    used_at.nil?
+    used_at.nil? && created_at > EXPIRATION.ago
```
garyhtou:
```suggestion
    used_at.nil? && created_at >= EXPIRATION.ago
```

The query `where(created_at: EXPIRATION.ago..)` results in the SQL `WHERE "created_at" >= '2023-12-02 08:45:33.438175'`

```
[1] pry(main)> LoginCode.where(created_at: 1.minute.ago..).to_sql
=> "SELECT \"login_codes\".* FROM \"login_codes\" WHERE \"login_codes\".\"created_at\" >= '2023-12-02 08:49:00.533908'"
```

app/models/login_code.rb
```diff
@@ -15,33 +15,41 @@
...
-  scope :active, -> { where(used_at: nil) }
+  EXPIRATION = 15.minutes
+
+  scope :active, -> { where(used_at: nil, created_at: EXPIRATION.ago..) }
 
   belongs_to :user
 
   before_create :generate_code
 
+  validates :code, uniqueness: { conditions: -> { active } }
```
garyhtou: should we validate presence?

spec/models/login_code_spec.rb
```diff
@@ -21,12 +21,21 @@
...
+    it "is false when used_at is present" do
+      login_code = create(:login_code, used_at: DateTime.current)
+      expect(login_code).to_not be_active
+    end
+
+    it "is false when expired" do
+      login_code = create(:login_code)
+      expect(login_code).to be_active
+
+      travel 15.minutes + 1.second
```
garyhtou:
```suggestion
      travel LoginCode::EXPIRATION + 1.second
```

app/models/login_code.rb
```diff
@@ -15,33 +15,41 @@
...
 
   private
 
   def generate_code
-    self.code = SecureRandom.random_number(999_999).to_s.ljust(6, "0") # pad with zero(s) if needed
+    loop do
+      self.code ||= SecureRandom.random_number(999_999).to_s.ljust(6, "0") # pad with zero(s) if needed
+      self.validate
+      break unless self.errors[:code].any?
+    end
```
garyhtou: I think this results in an infinite loop if there's a collision. We can try something like this:

```ruby
def generate_code
  while (self.validate || true) && self.errors[:code].any? do
    self.code = SecureRandom.random_number(999_999).to_s.ljust(6, "0") # pad with zero(s) if needed
  end
end

# or

def generate_code
  while !code_valid? do
    self.code = SecureRandom.random_number(999_999).to_s.ljust(6, "0") # pad with zero(s) if needed
  end
end

def code_valid?
	self.validate
  self.errors[:code].any?
end
```

cjdenio: Good catch. The `||=` was causing the loop— I've refactored the method to remove it, and verified it _doesn't_ loop infinitely now

(also working on writing a test case for this)

northeastprince:
> It's scoped to a user so even if two identical codes were generated, wouldn't it not matter?

@cjdenio

cjdenio:
What's the practical difference? User-scoped uniqueness offers a theoretically smaller chance of a collision, and global uniqueness offers a theoretical security advantage. But I don't think either approach is any different in reality

northeastprince:
What do you mean by a theoretical security advantage? I'm saying that if we're checking login codes under a user, there's no need for a uniqueness validation.

cjdenio:
Yes, but you can pretend to be any user at the login stage by submitting the form with their email

northeastprince:
Yeah but as long as they don't have the login code they won't be authenticated.

garyhtou (approved):

PR #4589, opened 2023-12-01
Hide message field for recurring donations

cjdenio:
## Summary of the problem

There's a "message" field when you go to make a monthly donation to an org, while messages don't seem to be implemented for monthly donations.

## Describe your changes

https://github.com/hackclub/hcb/assets/34525547/316b1fd3-c1c4-46f0-8811-a54ec6ab340e


sampoder:
oh that's good catch, cool if we instead implement it for monthly donors? i could probs whip smth up in an hour

sampoder:
Check out #4590!

PR #4590, opened 2023-12-01
Add messages to recurring donations

sampoder:
## Summary of the problem

We had this field but it did nothing!

## Describe your changes

Now values inputted will be saved in the recurring donation record and they should be emailed out. Do we only want to email messages once? I feel like we should and can edit this PR to do that.


sampoder:
(ignore those schema changes I'll fix them quickly when my laptop gets some juice in it)

cjdenio (changes requested):
Great start! I agree that messages should only be sent for the first donation— you can use `Donation#initial_recurring_donation?` to check

app/models/donation.rb
```diff
@@ -306,7 +310,7 @@ def send_payment_notification_if_needed
 
     if first_donation?
       DonationMailer.with(donation: self).first_donation_notification.deliver_later
-    elsif includes_message?
+    elsif includes_message? && (initial_recurring_donation? || !recurring?)
```
sampoder: @cjdenio that should do this! is there a cleaner way you can think of?

app/models/donation.rb
garyhtou, line 1: Let's remove the changes from this file and include the message here:

https://github.com/hackclub/bank/blob/6ea51d0675f51ed73409a7e185afc5f6f284a816/app/services/recurring_donation_service/handle_invoice_paid.rb#L10-L19

Only include the message if it's the first donation

garyhtou: This is because ideally, `Donation` shouldn't know much about `RecurringDonation`.

garyhtou: addressed

garyhtou (approved):
I think this should be good to go now! please sanity check it first (I haven't tested it)

PR #4592, opened 2023-12-02
Improve mobile responsiveness of mailbox address generation UI

cjdenio:
## Summary of the problem

The address textbox is cut off on mobile:

## Describe your changes

now it's not!

Screenshot 2023-12-01 at 10 09 47 PM


sampoder (approved):
Good catch, looks great!

PR #4593, opened 2023-12-02
Serve `.webp` files directly without conversion.

sampoder:
## Summary of the problem

Every time we attempt to preview a `.webp` image we're running it through image magik's converter.

This piece of code:

```
Rails.application.routes.url_helpers.rails_representation_url(file.variant(resize: "1024x1024").processed, only_path:)
```

(where file is a webp file)

Runs this command:

```
magick convert /tmp/ActiveStorage-55143-20231202-2-g0tqqi.webp[0] -auto-orient -resize 1024x1024 /tmp/image_processing20231202-2-3y0ter.png
```

Which leads to this error:

```
convert: UnableToOpenConfigureFile `delegates.xml' @ warning/configure.c/GetConfigureOptions/702.                
convert: NoDecodeDelegateForThisImageFormat `VIDEO' @ error/constitute.c/ReadImage/572.                          
convert: NoImagesDefined `/tmp/image_processing20231202-2-3y0ter.png' @ error/convert.c/ConvertImageCommand/3322.
```

This doesn't occur for `.heic` files, interestingly.

## Describe your changes

My change here bypasses the conversion and gives a direct URL to the file. Is this ideal? No. Ideally we would be able to run the conversion, but it's better than having no preview and because `.webp` images are already very well optimised I think it won't have any major performance impact.


garyhtou:
Can you take a look into [this PR](https://github.com/hackclub/hcb/pull/3366/files) that handled a similar problem for HEIC? See if we can follow the same approach

PR #4598, opened 2023-12-03
Rename export_ready.erb to export_ready.erb

sampoder:
Notice this bug just before the previous PR got merged but the app wouldn't let me disable auto merge. Matches Rails' conventions on file naming... really not my night tonight


sampoder:
(additional change from extra testing, user isn't defined)

garyhtou (approved):

PR #4599, opened 2023-12-03
Reduce Code Repetition in `ExportsController#transaction`

sampoder:
## Summary of the problem

This thing was annoying to maintain because there's so many cases to consider... now every case uses the same code. 

## Describe your changes

I thought this was the cleanest way of reducing the amount of repetition but I'd be curious to hear feedback / ideas.

Draft WIP because I need to test this rq.


sampoder:
Fixed in another PR

PR #4605, opened 2023-12-04
adds ineligibility for perks

sarthaktexas:


sarthaktexas:
fixes #4548

app/views/events/promotions.html.erb
```diff
@@ -46,12 +46,16 @@
   </p>
 </div>
 
+<% if (@event.country_us?) %>
 <div class="card pb0 mb3" id="event-helper">
+<% else %>
+<div class="card pb0 mb3 muted" id="event-helper">
+<% end %>
```
```suggestion
<div class="card pb0 mb3 <%= "muted" unless @event.country_us? %>" id="event-helper">
```
check this rq for me, but I think this is valid ERB and imo cleaner / more readable

app/views/events/promotions.html.erb
```diff
@@ -111,12 +115,16 @@
   </div>
 <% end %>
 
+<% if (@event.country_us?) %>
 <div class="card pb0 mb3" id="pvsa">
+<% else %>
+<div class="card pb0 mb3 muted" id="pvsa">
+<% end %>
```
sampoder: (if what I sent above works, it'd be nice to include it here as well)

app/views/events/promotions.html.erb
sarthaktexas: i literally googled this and it said this was not valid but i didn't fact check that SO post

sarthaktexas: i agree this is cleaner so i'll check later if this works

sampoder: i searched the bank codebase, we seem to do it a fair bit
eg.
```erb
<span class="<%= "muted " unless defined?(selectable) %>mb1 flex items-center">
```

¯\_(ツ)_/¯

sampoder (approved):

garyhtou (approved):

PR #4606, opened 2023-12-04
Sort sessions by updated_at

sarthaktexas:
session now sorts by last active i believe


sarthaktexas:
fixes #4477

cjdenio (changes requested):
Good start! `updated_at` tracks when any field on the `UserSession` was changed, which doesn't necessarily correlate to when a user opens their device and starts using HCB.

One implementation might be a new `active_at` field that's bumped whenever a user takes an action in the interface (possibly with an [`after_action`](https://guides.rubyonrails.org/action_controller_overview.html#after-filters-and-around-filters) filter)

PR #4607, opened 2023-12-04
Say "HCB" if they haven't been invited to an org but are signing up

sampoder:
## Summary of the problem

I'm not sure what's the edge case causing this but here's the error: https://heroku-app97991095.airbrake.io/projects/288439/groups/3686007377140347301?tab=notice-detail

## Describe your changes

If org_pos is nil, we'll just say "HCB"


sarthaktexas (approved):

app/views/layouts/application.html.erb
```diff
@@ -148,7 +148,9 @@
             <div class="welcome__content">
               <div class="center">
                 <h3 class="h3 caps secondary block m0">Welcome to</h3>
-                <h1 class="primary mt0 border-none"><%= @organizer_position.initial? ? "HCB" : @event.name %></h1>
+                <h1 class="primary mt0 border-none">
+                  <%= @organizer_position.nil? || @organizer_position.initial? ? "HCB" : @event.name %>
+                </h1>
```
garyhtou:
```suggestion
                <h1 class="primary mt0 border-none"><%= @organizer_position&.initial? ? "HCB" : @event.name %></h1>
```

garyhtou (approved):

PR #4619, opened 2023-12-05
Move the `.where` in event filtering on the admin controller

sampoder:
## Summary of the problem

I think this is related to the array / active record stuff we've been talking about. Here's an Airbrake error: https://heroku-app97991095.airbrake.io/projects/288439/groups/3686814272025489859.

## Describe your changes

I moved it down until before we do the balance filtering.


garyhtou (commented):
~~thanks! it looks like the page is also supposed to handle `@start_date`. I'm not sure if I see a `where` for that (might have been lost sometime in the past)~~

nevermind! the start and end dates are handled by the balance methods. the where for `@end_date` only exists to filter out events that weren't created before the end date

garyhtou (approved):

PR #4623, opened 2023-12-06
Setup Codespaces

sampoder:
## Summary of the problem

We have many people who want to code on Bank but.... it's difficult to get started ~~(except if you use Docker)~~.

## Describe your changes

Here's a config file for Codespaces.


garyhtou:
The issue now is that `.env.development` doesn't exist since `cp` is running too late. it's being used in docker compose

garyhtou:
you may get an error about `bundle.js`. if so, run `bin/rails assets:precompile`

garyhtou:
can someone test this to see if it's working for them? On this PR, there's a button on the top right called "Code". it should let you create a codespace

garyhtou:
To get the server running, run `bin/rails s` in the terminal

.env.development.example
```diff
@@ -1,3 +1,4 @@
 PORT=3000
-DATABASE_URL=postgres://postgres@db:5432
+DATABASE_URL=postgres://postgres@localhost:5432
+REDIS_URL=redis://redis:6379
```
garyhtou: TODO: figure out of this is necessary

polypixeldev:
Works great for me after running `bin/rails assets:precompile`!

northeastprince:
It's currently taking ~10 minutes to build the container.

garyhtou:
> It's currently taking ~10 minutes to build the container.

That is normal since docker needs to pull, build, and run containers! However, once this is on `main` GitHub codespaces will pre-build codespaces for us on every push

PR #4624, opened 2023-12-06
Improve The Seed

sampoder:
## Summary of the problem

@sarthaktexas wanted more data! Now he has it...

## Describe your changes

Made a bunch of transactions and different type of events


garyhtou:
@sampoder, two things:
- Let's make it so that you do not need to have a User in the database to run the seed file. If no users exist, create a user with admin@example.com (and continue with the script)
- Let's merge this in by Friday at the latest

garyhtou (approved):

PR #4628, opened 2023-12-06
Rename HCB Wrapped download to `wrapped.json`

cjdenio:


garyhtou:
Can you also update the README that references "test.json"

https://github.com/hackclub/hcb-wrapped-2023

PR #4630, opened 2023-12-06
Use `all` not `default`

sampoder:


garyhtou (approved):
good to go!

ref: https://github.com/rails/rails/blob/c057edaaadbec0cd1f22cbfac9697e8e0b51f73e/actionmailbox/lib/action_mailbox/router/route.rb#L18

there needs to be better docs for actionmailbox 😭

PR #4011, opened 2023-06-22
Try upgrading to heroku-22 for review apps

albertchae:
Now that we're on ruby 3.1, we can try upgrading to heroku-22 for review apps. After confirming the app works ok on 22, we can change it for production as well.

https://devcenter.heroku.com/articles/heroku-22-stack

https://devcenter.heroku.com/articles/ruby-support#ruby-versions

We can remove 2 buildpacks we were using for heroku-20:

It looks like the heroku-22 built in imagemagick is technically older than what the buildpack was giving us, but it has HEIC support and so HEIC conversion works without an explicit buildpack.

heroku-20 with buildpack
```
$ convert -version
Version: ImageMagick 7.0.11-5 Q16 x86_64 2021-03-20 https://imagemagick.org/
Copyright: (C) 1999-2021 ImageMagick Studio LLC
License: https://imagemagick.org/script/license.php
Features: Cipher DPC HDRI OpenMP(4.5)
Delegates (built-in): bzlib djvu fontconfig freetype heic jbig jng jpeg lcms lqr lzma openexr pangocairo png tiff x xml zip zlib
```

heroku-22 no buildpack
```
$ convert -v
Version: ImageMagick 6.9.11-60 Q16 x86_64 2021-01-25 https://imagemagick.org/
Copyright: (C) 1999-2021 ImageMagick Studio LLC
License: https://imagemagick.org/script/license.php
Features: Cipher DPC Modules OpenMP(4.5)
Delegates (built-in): bzlib djvu fftw fontconfig freetype heic jbig jng jp2 jpeg lcms lqr ltdl lzma openexr pangocairo png tiff webp wmf x xml zlib
```

We can remove our custom buildpack for poppler 21.03.0 that we introduced in https://github.com/hackclub/bank/pull/3395, because heroku-22 has poppler 22.02.0 installed by default

```
$ pdftoppm -v
pdftoppm version 22.02.0
Copyright 2005-2022 The Poppler Developers - http://poppler.freedesktop.org
Copyright 1996-2011 Glyph & Cog, LLC
```

PR #4632, opened 2023-12-06
Add WebP Buildpack

sampoder:
## Summary of the problem

We can't convert WebP images with imagemagik!

## Describe your changes

Takes the same approach as #3366 to add support for converting WebP images.


garyhtou:
I just wanted to verify with you that webp currently isn't working on staging. Because looking this comment (https://github.com/hackclub/hcb/issues/4034#issuecomment-1615283439) there's a chance the heroku-22 upgrade might have fixed it for staging (and not prod)

sampoder:
Is staging back up? I remember it wasn't possible to sign into it for a while

sampoder:
Doesn't it look like it is, that said we don't seem to have the HEIC buildpack in `app.json` anymore

maxwofford:
cc'ing @albertchae for context on HEIC and WebP support

albertchae:
The HEIC buildpack was removed from `app.json` in https://github.com/hackclub/hcb/pull/4011. You can read that PR for more details, but TL;DR using the heroku-22 stack means we don't need an explicit buildpack anymore.

Unfortunately I think I forgot about upgrading prod to heroku-22 which we were tracking in https://github.com/hackclub/hcb/issues/3398. This is probably something we can do soon.

There also appears to be some more divergence between review apps and prod after https://github.com/hackclub/hcb/pull/4304/files, which calls `config.load_defaults 7.0` in `config/application.rb`, but sets `config.active_storage.variant_processor = :mini_magick` in `config/environments/production.rb`. This means prod is using imagemagick [but all other environments will be using vips](https://guides.rubyonrails.org/configuring.html#default-values-for-target-version-7-0).

So some action items we should take:
- [x] Finalize upgrading prod to heroku-22. https://github.com/hackclub/hcb/issues/3398
- [ ] Decide on imagemagick or vips everywhere and modify the config to reflect that. I am unclear on vips support for HEIC or webp and unclear on whether the imagemagick from heroku-22 handles webp. We should research that too. We should favor using vips since it's supposed to be faster unless there's a compatibility issue
- [ ] ~~This is orthogonal, but do we have an issue for the review app login issue? I'm in some slack DMs related to it, but we should bring the conversation to github.~~ created https://github.com/hackclub/hcb/issues/4927

maxwofford:
Thanks for adding the context @albertchae! @sampoder are you interested in taking on fully switching to VIPS and checking compatibility?

@albertchae Please create a ticket for that review app issue, I don't think I have much context to that.

sampoder:
Sure thing! I can work on the VIPS stuff

sampoder:
To check VIPS support, we could upload HEICs / WebPs to a review app. Am I right in saying that @albertchae?

albertchae:
> To check VIPS support, we could upload HEICs / WebPs to a review app. Am I right in saying that @albertchae?

@sampoder sorry for the late reply but yeah that should be sufficient. I'll work on fixing staging in https://github.com/hackclub/hcb/issues/4927 and ping you afterwards

albertchae:
@sampoder new review apps should be able to login again via email, SMS login is still in progress though

sampoder:
Awesome!

sampoder:
Upgrading to `heroku-22` worked!

garyhtou:
amazing!

PR #4633, opened 2023-12-06
Fix country specific perks displaying incorrectly in playground mode

sarthaktexas:
## Summary of the problem

During demo mode, some buttons would say ineligible and some would say Purchase, even though all were disabled.

## Describe your changes

This fixes it through a series of implicit if-else statements.

In playground mode:
Screenshot 2023-12-07 at 12 58 12 PM

Out of PM:
Screenshot 2023-12-07 at 12 58 40 PM

Out of PM with US set as Country:
Screenshot 2023-12-07 at 12 59 11 PM


sarthaktexas:
closes #4627

sarthaktexas:
@cjdenio can i get a quick review from you? i'd like to fix this feature quickly

app/views/events/promotions.html.erb
```diff
@@ -46,10 +46,10 @@
   </p>
 </div>
 
-<div class="card pb0 mb3 <%= "muted" unless @event.country_us? %>" id="event-helper">
+<div class="card pb0 mb3 <%= "muted" unless (@event.country_us? || !@perks_available) %>" id="event-helper">
   <div class="card__banner card__banner--top flex justify-between items-center promo-event-helper">
     <%= image_tag "https://cloud-fz66hmje0-hack-club-bot.vercel.app/0theeventhelperbutbetter.png", width: 60 %>
-    <%= link_to @event.country_us? ? "Purchase" : "Ineligible",
+    <%= link_to @perks_available ? (@event.country_us? ? "Purchase" : "Ineligible") : "Purchase",
```
maxwofford: This looks a little funny to have a nested ternary here. Couldn't we do something like `(@event.country_us? && @perks_available) ? "Purchase"  : "Ineligible")`?

By that point we're kinda being redundant with our conditions too. Maybe it'd make sense to drop this into the controller so we can just ask `@event_helper_eligible` or something?

maxwofford: Talked through this over a call with @sarthaktexas and the additional condition he's trying to cover is showing "Purchase" while an org is in demo + transparency, because demo orgs don't have their country set. I think this is fine for now as long as it changes in #4612

maxwofford (approved):
I haven't run this locally, but as mentioned in my comment (https://github.com/hackclub/hcb/pull/4633/files#r1423092247) I'm fine as long as this gets refactored.

PR #4636, opened 2023-12-06
adds grid view to mock cards

sarthaktexas:
## Summary of the problem

onboarding would like grid view to be default when demoing the platform. grid view is currently broken.

## Describe your changes

this fixes it by adding keys to fix render and adding a to_partial_path key to help rails find the partial due to the openstruct not having the model struct. also fixes card muted status and href on cards.


sarthaktexas:
fixes #4547

garyhtou (approved):

PR #4639, opened 2023-12-07
Use `inline` not `mini` and `memo`

sampoder:
## Summary of the problem

When we renamed from `mini-rename` to `memo` I cmd-f'd and I missed that the `memo` param will exist whenever we are editing.

## Describe your changes

I'm using `inline` for everything now


garyhtou:
go ahead and merge this after doing a thorough test!

sampoder:
currently tested: renaming card transactions

sampoder:
Now tested: inline renaming, renaming bank account transactions, renaming ACHs

PR #4654, opened 2023-12-08
HcbCode: Show memo of Check, IncreaseCheck, and Disbursement

garyhtou:
## Summary of the problem

The memos for linked objects such as Checks, IncreaseChecks, and Disbursements are not visible in the info table on HcbCode show pages
- Checks: The memo printed on the check
- Increase check: The memo printed on the check
- Disbursement: The reason for the disbursement

https://hackclub.slack.com/archives/C047Y01MHJQ/p1702009034016009

## Describe your changes

add em back

image
image


sampoder (commented):
Agree with adding back the reason for disbursements but I think having two memos for a check is confusing

garyhtou:
Yeah, tbh check memos aren't really necessary since that info is already visible on the visual mock check

cjdenio:
I think it is still important to show the check memo for readability (and since the mock check isn't shown on mobile). Maybe the naming can be workshopped, but the "transaction memo" and "check memo" are 2 different things that should both be accessible to orgs

PR #4660, opened 2023-12-08
Upgrade to Rails 7.1

cjdenio:


garyhtou:
OHHHHH YES PLEASE 🙏

cjdenio:
This should be ready to go. Two caveats:

- I had to remove the `skylight` gem to prevent a weird version conflict. I assume we're not using it regularly and can try to add it back later.
- This upgrade **will log every HCB user out**. I'm not sure why (looks like they changed the format/algorithm for encrypted cookies), and I can't find anything in the documentation on how to prevent this. It's annoying, but I don't think users will care. (anyone is welcome to look into fixing this)

sampoder:
sorry that's my bad re: the AdminMailer! i wonder why that only just starting breaking now?

cjdenio:
For posterity: the logout issue was was found to only affect development

PR #4662, opened 2023-12-08
Clarify when we redirect from `edit_address`

sampoder:
## Summary of the problem

It's a bit confusing to be randomly redirected, see #4656. 

## Describe your changes

Now we'll flash an info message!


garyhtou:
Is this superseded by https://github.com/hackclub/hcb/pull/4956?

PR #4666, opened 2023-12-09
Waive all fees collected on refunded donations

sampoder:
WIP because I want to test, what'd be a good way of testing?


cjdenio:
Do we want to do this from an operational standpoint? Stripe doesn't waive fees when we refund a payment

sampoder:
Not sure, there was an open issue, I was playing around

rluodev:
@hackliv Would this be something that we'd want to do?

sampoder:
For reference, here's the issue #2093

hackliv:
If there is a way to do this automatically, then yes, would be great for Ops to not have to worry about!

hackliv:
I will share though that I'd prefer prioritizing instant fee waive/un-waive on pending disbursements over this though if you all are starting a dive into bug fixes for fees!

app/services/donation_service/refund.rb
```diff
@@ -15,10 +15,19 @@ def run
         # 2. Un-front all pending transaction associated with this donation
         donation.canonical_pending_transactions.update_all(fronted: false)
 
-        # 3. Process remotely
+        # 3. Waive all fees collected
+
+        donation.canonical_transactions.each do |ct|
+          fee = ct.canonical_event_mapping.fees.first
```
garyhtou: this line may need to be updated due to https://github.com/hackclub/hcb/pull/4773

app/services/donation_service/refund.rb
```diff
@@ -15,10 +15,19 @@ def run
         # 2. Un-front all pending transaction associated with this donation
         donation.canonical_pending_transactions.update_all(fronted: false)
 
-        # 3. Process remotely
+        # 3. Waive all fees collected
+
```
garyhtou:
```suggestion
```

app/services/donation_service/refund.rb
```diff
@@ -15,10 +15,18 @@ def run
         # 2. Un-front all pending transaction associated with this donation
         donation.canonical_pending_transactions.update_all(fronted: false)
 
-        # 3. Process remotely
+        # 3. Waive all fees collected
+        donation.canonical_transactions.each do |ct|
+          fee = ct.fee
```
sampoder: @garyhtou does that work?

cjdenio (changes requested):
https://github.com/hackclub/hcb/blob/13576eb84d5ff7bf7436bec1600c95869c8df91b/app/services/fee_engine/create.rb also needs to be updated for when a CanonicalTransaction comes in _after_ the donation is refunded

app/services/donation_service/refund.rb
garyhtou: i believe so

app/services/fee_engine/create.rb
```diff
@@ -47,6 +47,8 @@ def determine_reason
 
       reason = "REVENUE WAIVED" if canonical_transaction.memo.downcase.include?("acctverify") && canonical_transaction.amount_cents.abs < 100 # Waive fees on account verification transactions from platforms like Venmo
 
+      reason = "REFUNDED" if canonical_transaction.donation&.refunded?
```
sampoder: @cjdenio would this do the trick?

cjdenio: `CanonicalTransaction#donation`, unintuitively, doesn't work for modern donations

```suggestion
      reason = "DONATION REFUNDED" if canonical_transaction.local_hcb_code.donation&.refunded?
```

sampoder: Oh neat good catch

cjdenio (approved):

PR #4685, opened 2023-12-12
Add more development flavor texts

polypixeldev:
## Summary of the problem

This change is being made to a couple more flavor texts to the HCB homepage, specifically for the default set of flavor texts and the development mode set of flavor texts. It helps to solve the problem of there not being enough flavor texts on the HCB homepage, although this problem will never be fully solved.

## Describe your changes

I added a couple lines to `flavor_text_service.rb` which include text that will be displayed for users when they navigate to the HCB homepage. One line is for the set of development flavor texts, and the other is for the set of default flavor texts.

Development mode text:
Screenshot from 2023-12-11 21-27-45

Default mode text:
Screenshot from 2023-12-11 21-40-39


sampoder (approved):
which fraudsters have you been talking to?

polypixeldev:
sam bankman-fried is the only one that I'm able to say at the moment, but rest assured that I have taken every step to make sure that the results of the study are accurate!

app/services/flavor_text_service.rb
```diff
@@ -485,6 +486,7 @@ def flavor_texts
       "<marquee scrollamount='5'>💸💸💸</marquee>".html_safe,
       "Wow, that’s a lot of money. Need some help carrying it?",
       "I would rather check my Facebook than face my checkbook.",
+      "recommended by 100% of fraudsters",
```
maxwofford: Want to make this a dev flavor text instead?

polypixeldev: sure, might not be right for a default flavor text lol - will make the change later today when I get to a computer

garyhtou (approved):
perfect! 💸

app/services/flavor_text_service.rb
garyhtou: thanks!

PR #4702, opened 2023-12-12
Wallets Perk moved to expired promotions

aramshiva:
## Summary of the problem
Solving #4695. Wallets are not being shipped at moment so it has been moved to "Expired Promotions" at least for now.


app/views/events/promotions.html.erb
```diff
@@ -234,7 +213,24 @@
 
 <details class="w-100 left-align">
   <summary><h3 class="inline-block pb0 border-none">Expired Promotions</h3></summary>
+  
+  <div class="card mb3" id="wallets">
+    <div class="card__banner card__banner--top flex justify-between items-center promo-wallets">
+      <h3 class="h1 p2 my0 color-black">Wallets</h3>
+      <%= link_to @wallets_available ? "Request" : "Out of stock",
```
```suggestion
      <%= link_to @wallets_available ? "Request" : "Expired",
```

Let's update the copy here.

sampoder (approved):
looks good! just one quick copy edit

sarthaktexas (approved):

app/views/events/promotions.html.erb
```diff
@@ -234,7 +213,24 @@
 
 <details class="w-100 left-align">
   <summary><h3 class="inline-block pb0 border-none">Expired Promotions</h3></summary>
+  
+  <div class="card mb3" id="wallets">
```
garyhtou: `@wallets_available` is not defined. this wouldn't cause an error since undefined instance variables return `nil`. However, if we plan to use it, it's best practice to explicitly define it.

```suggestion
  <% @wallets_available = false %>
  <div class="card mb3" id="wallets">
```

garyhtou (changes requested):

garyhtou (approved):
great! thank you

PR #4705, opened 2023-12-12
Ignore Chrome Extension Errors in Airbrake

rluodev:
I think this fixes #4667, but I have no way of testing it. Added a filter in `/config/initializers/airbrake.rb` to filter out anything that starts with `chrome-extension://`, but since I don't think I have access to airbrake I can't test this.


config/initializers/airbrake.rb
```diff
@@ -84,6 +84,11 @@
   notice.ignore! if rand(1..10) <= 9
 end
 
+Airbrake.add_filter do |notice|
+  if notice[:errors].any? { |error| error[:file].start_with?("chrome-extension://") }
+    notice.ignore!
+  end
```
sampoder:
```suggestion
  notice.ignore! if notice[:errors].any? { |error| error[:file].start_with?("chrome-extension://") }
```

I believe in Ruby we can just write this which is shorter (and cooler?)

cjdenio (changes requested):
Great start! Since these errors are reported by the JavaScript frontend, a filter has to be added in `app/javascript/airbrake.js`:
https://github.com/hackclub/hcb/blob/8c9d91d290d60d9ff56924a9ed09abeb38e0248b/app/javascript/airbrake.js#L14-L17

config/initializers/airbrake.rb
```diff
@@ -84,7 +84,6 @@
   notice.ignore! if rand(1..10) <= 9
 end
```
sampoder:
```suggestion


```

add back that line

rluodev:
@cjdenio @garyhtou I think this is good to go

sampoder (approved):
LGTM

garyhtou:
@cjdenio did you want to review this one? no worries if you're short on capacity

app/javascript/airbrake.js
```diff
@@ -16,4 +16,9 @@ airbrake?.addFilter(notice => {
   return notice
 })
 
+airbrake?.addFilter((notice) => {
+  if (notice.errors.some((error) => error.file.startsWith('chrome-extension://'))) return null;
```
cjdenio:
```suggestion
  if (
    notice.errors
      .flatMap(e => e.backtrace)
      .some(e => e.file?.startsWith('chrome-extension://'))
  )
    return null
```

`Error` objects don't have a `file`, but they have `Backtraces` that do

cjdenio (approved):

PR #4709, opened 2023-12-12
Return nil when linking to an invalid HCB code & support 7 digit HCB codes

sampoder:
## Summary of the problem

closes https://github.com/hackclub/hcb/issues/4707

## Describe your changes

Now when the HCB code is nil it should return nil


garyhtou (approved):
great!

PR #4716, opened 2023-12-12
Authorise events, cards, and HCB codes when linking them

sampoder:
## Summary of the problem

Security issue, see: https://hackclub.slack.com/archives/C047Y01MHJQ/p1702404175275799

## Describe your changes

Now we authorise each of these with Pundit based on the current user


app/services/markdown_service.rb
```diff
@@ -64,6 +72,7 @@ def try_hcb_autolink(link)
         hcb_code = found_link[2]
         comment = found_link[3]
         hcb = HcbCode.find_by_hashid hcb_code
+        Pundit.authorize(@current_user, hcb, :show?)
```
garyhtou: when we merge the other PR, let's make sure that the `return nil` goes BEFORE this new line

garyhtou (commented):
thank you

garyhtou (approved):

PR #4720, opened 2023-12-13
Make it possible to void invoices

rluodev:
This adds the ability to void an invoice and adds sorting options on both admin- and user-side to sort and view these voided invoices.

Once merged, this will close #4711.

Voided invoice

Screenshot 2023-12-13 at 9 01 18 AM
paid invoice with void button disabled

Screenshot 2023-12-12 at 5 25 35 PM
sent invoice with void button

Screenshot 2023-12-12 at 5 25 54 PM

Screenshot 2023-12-12 at 5 23 52 PM

Screenshot 2023-12-12 at 5 26 56 PM


sarthaktexas:
by rawr???

rluodev:
> by rawr???

that's just what I could think of

sarthaktexas (changes requested):
Could I get captions on the images in the PR description? Specifically, the first few because I cannot tell the difference.

sarthaktexas:
Also, I cannot approve this until I test it locally, so you'll have to wait until tomorrow for an approval review. I've reviewed your code and have some comments. I will submit a review tomorrow when I have a pair of fresh morning eyes. I love the feature!

app/views/hcb_codes/_invoice.html.erb
```diff
@@ -140,6 +145,12 @@
       PDF
     <% end %>
     </div>
+    <%= link_to (@invoice.voided? ? "Already voided" : @invoice.paid_v2? ? "Already paid" : "Void this invoice"),
+      invoice_void_path(@invoice.id),
+      disabled: !organizer_signed_in? || @invoice.voided? || @invoice.paid_v2?,
+      method: :post,
+      data: { confirm: "Are you sure you want to void this invoice? This action cannot be undone." },
+      class: "btn bg-error mr1 right" %>
```
sampoder:
```suggestion
    <%= link_to "Void this invoice",
      invoice_void_path(@invoice.id),
      disabled: !organizer_signed_in? || @invoice.voided? || @invoice.paid_v2?,
      method: :post,
      data: { confirm: "Are you sure you want to void this invoice? This action cannot be undone." },
      class: "btn bg-error mr1 right" unless @invoice.paid_v2? || @invoice.voided?  %>
```

Let's hide this button if the invoice is paid or already voided, reduces clutter imo.

app/views/hcb_codes/_invoice.html.erb
```diff
@@ -140,6 +145,12 @@
       PDF
     <% end %>
     </div>
+    <%= link_to (@invoice.voided? ? "Already voided" : @invoice.paid_v2? ? "Already paid" : "Void this invoice"),
+      invoice_void_path(@invoice.id),
+      disabled: !organizer_signed_in? || @invoice.voided? || @invoice.paid_v2?,
+      method: :post,
+      data: { confirm: "Are you sure you want to void this invoice? This action cannot be undone." },
```
sampoder:
```suggestion
      data: { confirm: "Are you sure you want to void this invoice? Voided invoices can not be paid and this action cannot be undone." },
```

Let's be super clear with folks about what this button does.

app/views/hcb_codes/_invoice.html.erb
```diff
@@ -9,7 +9,12 @@
   <%= link_to "View on Stripe", @invoice.stripe_dashboard_url, class: "btn bg-accent" %>
 <% end %>
 
-<% if @invoice.archived? %>
+<% if @invoice.voided? %>
+  <div class="banner banner--archived mt3">
+    <span class="bold block mb1">This invoice was voided by <%= user_mention @invoice.voided_by %> on <%= format_date @invoice.voided_at %>.</span>
+    <span class="block">It will only display in the Voided tab<span class="xs-hide"> on the Invoices page</span>.</span>
+  </div>
+<% elsif @invoice.archived? %>
```
sampoder: Can an invoice be both voided and archived? It seems like I can still archive voided invoices. Do we want to support this?

app/models/invoice.rb
```diff
@@ -188,6 +194,10 @@ def archived?
     archived_at.present?
   end
 
+  def voided?
+    voided_at.present?
+  end
+
```
sampoder: Let's continue using `void_v2?`. Afaik `mark_void!` should lead to `void_v2?` returning true.

app/controllers/admin_controller.rb
```diff
@@ -857,11 +858,13 @@ def invoices
       end
     end
 
-    relation = relation.open_v2 if @open
+    relation = relation.open_v2.excluding(relation.voided) if @open
```
sampoder:
```suggestion
    relation = relation.open_v2 if @open
```
When we void a invoice they should transition to the `void_v2` state and therefore not be open. Were you observing that when testing this?

app/controllers/admin_controller.rb
```diff
@@ -857,11 +858,13 @@ def invoices
...
     relation = relation.paid_v2 if @paid
     relation = relation.missing_payout if @missing_payout
     relation = relation.missing_fee_reimbursement if @missing_fee_reimbursement
     relation = relation.past_due if @past_due
+    relation = relation.voided if @voided
```
sampoder:
```suggestion
    relation = relation.void_v2 if @voided
```

app/controllers/invoices_controller.rb
```diff
@@ -38,11 +38,13 @@ def index
     when "paid"
       relation = relation.paid_v2
     when "unpaid"
-      relation = relation.unpaid
+      relation = relation.unpaid.excluding(relation.voided)
     when "archived"
-      relation = relation.archived
+      relation = relation.archived.excluding(relation.voided)
+    when "voided"
+      relation = relation.voided
```
sampoder:
```suggestion
      relation = relation.void_v2
```

sampoder (commented):
Couple of comments from me. Main thing is that I think we should be using `.void_v2` instead of creating a separate `.voided`.

app/controllers/admin_controller.rb
rluodev: Yeah, while testing, I had to use .excluding in order to get it to actually exclude voided ones. not sure why.

app/views/hcb_codes/_invoice.html.erb
rluodev: I mean it doesn't change anything visually aside from the invoice archive button changing between archive and unarchive, so I think we can just hide the archive button once voided? Will push a change for this.

app/models/invoice.rb
rluodev: Done!

rluodev:
@sampoder re: the void_v2 thing - it only works for *some* invoices, apparently, and the voided? call is a much more reliable way to keep track of voided status since we keep a record on our end of when the invoice was voided ~~instead of asking stripe, "hey is this void" and maybe not receiving back a correct response.~~

wait no i totally assumed that void_v2 was a call to stripe but anyway it still doesn't always work.

app/controllers/admin_controller.rb
sampoder: Were you running `.mark_void!`?

sampoder:
@Soar1826 how are you testing this locally? I'd like to pull it up on my computer and investigate.

app/controllers/admin_controller.rb
rluodev: yep, 
Screenshot 2023-12-13 at 11 08 10 AM

rluodev:
wait no nvm it was a weird thing that happened to one invoice on my end

rluodev:
void_v2 works fine

sampoder:
Oh hell yeah!

app/controllers/admin_controller.rb
```diff
@@ -857,11 +858,13 @@ def invoices
       end
     end
 
-    relation = relation.open_v2 if @open
+    relation = relation.open_v2.excluding(relation.void_v2) if @open
```
sampoder:
```suggestion
    relation = relation.open_v2 if @open
```
I believe it's implied that `open_v2` invoices are not `void_v2`

app/services/invoice_service/void.rb
sampoder, line 1: In the same way we have `MarkPaid`, let's have `MarkVoid` instead of just `Void`.

app/controllers/invoices_controller.rb
```diff
@@ -161,6 +163,24 @@ def archive
     end
   end
 
+  def void
+    @invoice = Invoice.friendly.find(params[:invoice_id])
+
+    authorize @invoice
+
+    @invoice.voided_at = DateTime.now
+    @invoice.voided_by = current_user
```
sampoder: Would it make sense to move this into the service? Feels more in line with what we do on `MarkPaid`

app/views/hcb_codes/_invoice.html.erb
```diff
@@ -9,7 +9,12 @@
   <%= link_to "View on Stripe", @invoice.stripe_dashboard_url, class: "btn bg-accent" %>
 <% end %>
 
-<% if @invoice.archived? %>
+<% if @invoice.void_v2? %>
+  <div class="banner banner--archived mt3">
+    <span class="bold block mb1">This invoice was voided by <%= user_mention @invoice.voided_by %> on <%= format_date @invoice.voided_at %>.</span>
+    <span class="block">It will only display in the Voided tab<span class="xs-hide"> on the Invoices page</span>.</span>
+  </div>
+<% elsif @invoice.archived? %>
```
sampoder: I could archive an invoice and then void it, should we display that in any way?

app/views/hcb_codes/_invoice.html.erb
```diff
@@ -140,11 +145,21 @@
...
+          method: :post,
+          data: { confirm: "Are you sure you want to void this invoice? Voided invoices can not be paid and this action cannot be undone." },
+          class: "btn bg-error mr1 right" %>
+      <% end %>
+      <%= link_to (@invoice.archived? ? "Un-archive" : "Archive"),
+        (@invoice.archived? ? invoice_unarchive_path(@invoice) : invoice_archive_path(@invoice)),
+        disabled: !organizer_signed_in?,
+        method: :post,
+        class: "btn bg-muted" %>
+    <% end %>
```
sampoder:
```suggestion
    <%= link_to "Void this invoice",
      invoice_void_path(@invoice.id),
      disabled: !organizer_signed_in? || @invoice.void_v2? || @invoice.paid_v2?,
      method: :post,
      data: { confirm: "Are you sure you want to void this invoice? Voided invoices can not be paid and this action cannot be undone." },
      class: "btn bg-error mr1 right"unless @invoice.void_v2? || @invoice.paid? %>
  <%= link_to (@invoice.archived? ? "Un-archive" : "Archive"),
    (@invoice.archived? ? invoice_unarchive_path(@invoice) : invoice_archive_path(@invoice)),
    disabled: !organizer_signed_in?,
    method: :post,
    class: "btn bg-muted" unless @invoice.void_v2? %>
```

Test this rq but I think this is valid and imo cleaner

sampoder (changes requested):
Couple from me! Looking really good.

app/views/hcb_codes/_invoice.html.erb
rluodev: The general idea is that voiding takes precedence - you can't archive after voiding, so it makes no sense to show something as archived when the void tab does the same thing and only shows voided like how archived only shows archived.

app/views/hcb_codes/_invoice.html.erb
rluodev: I feel like we do want to hide the button to archive as well when voided, doesn't really make sense to void and then archive.

app/views/hcb_codes/_invoice.html.erb
sampoder: That makes sense to me; thanks for clarifying! Just to sanity check, after we "void" an archived invoice does it show up as "archived" anywhere?

app/services/invoice_service/void.rb
rluodev: done!

app/views/hcb_codes/_invoice.html.erb
sampoder: Oh yes, that's my bad! I intended for that to be the case. Let me quickly fix my suggestion.

sampoder: Fixed, sorry

app/controllers/invoices_controller.rb
rluodev: I was doing it in-line with the other actions in invoice controller, like archive, etc.

app/views/hcb_codes/_invoice.html.erb
rluodev: thanks! also, one more thing: shouldn't `unless @invoice.void_v2? && @invoice.paid?` be `unless @invoice.void_v2? or @invoice.paid?`?

app/views/hcb_codes/_invoice.html.erb
rluodev: No, the status are all listed as void I believe, will test tho.

rluodev: yep, they all show as void.

sampoder: Perfect!

app/views/hcb_codes/_invoice.html.erb
rluodev: @sampoder

sampoder: Yes, though use `||`

app/controllers/invoices_controller.rb
sampoder: Let's move it into the service, that way we can reuse the service and it'll be consistent

sampoder (approved):
Awesome, LGTM!

app/models/invoice.rb
```diff
@@ -110,6 +114,7 @@ class Invoice < ApplicationRecord
   scope :missing_payout, -> { where("payout_id is null and payout_creation_balance_net is not null") } # some invoices are missing a payout but it is ok because they were paid by check. that is why we additionally check on payout_creation_balance_net
   scope :unpaid, -> { where("aasm_state != 'paid_v2'") }
   scope :past_due, -> { where("due_date < ?", Time.current) }
+  scope :voided, -> { where.not(voided_at: nil) }
```
garyhtou: Technically using `voided_at` works! However, when using AASM, this code pattern of using `*_at` to determine state can be buggy. Here's why!

Let's say an Invoice goes from `open` -> `paid`. Each state transition usually involves setting a timestamp (e.g. `opened_at`, and `paid_at`). We'd expect `where.not(opened_at: nil)` to give us a list of `open` invoices, however, this actually gives us a list of invoices that were _open at some point_. This is because state transitions do not clear the timestamps of other states.

In this specific case, it works fine because `void` is a terminal state. Once it's voided, there is no other state it could transition to.

To prevent possible bugs like these in the future, let's switch it to using the `aasm_state` column for determining state.

```suggestion
  scope :voided, -> { where(aasm_state: "void_v2") }
```

---

Also, I just wanted to mention that AASM automatically provides scopes. In this case, our `voided` scope is the exact same as AASM's provided `void_v2` scope.

app/views/hcb_codes/_invoice.html.erb
```diff
@@ -140,11 +145,17 @@
       PDF
     <% end %>
     </div>
-    <%= link_to (@invoice.archived? ? "Un-archive" : "Archive"),
-      (@invoice.archived? ? invoice_unarchive_path(@invoice) : invoice_archive_path(@invoice)),
-      disabled: !organizer_signed_in?,
+    <%= link_to "Void this invoice",
+      invoice_void_path(@invoice.id),
+      disabled: !organizer_signed_in? || @invoice.void_v2? || @invoice.paid_v2?,
```
garyhtou: Since we have `unless @invoice.void_v2? || @invoice.paid?` on line 153, I think it might not be needed

```suggestion
      disabled: !organizer_signed_in?,
```

app/views/hcb_codes/_invoice.html.erb
```diff
@@ -140,11 +145,17 @@
...
       method: :post,
-      class: "btn bg-muted" %>
+      data: { confirm: "Are you sure you want to void this invoice? Voided invoices can not be paid and this action cannot be undone." },
+      class: "btn bg-error mr1 right" unless @invoice.void_v2? || @invoice.paid? %>
```
garyhtou: I see that we're also checking for `deposited?` on line 51. I'm actually not sure if this is necessary; but imma assume it is 🤷‍♂️ (it's tough to say because the state management in this model needs some work)

```suggestion
      class: "btn bg-error mr1 right" unless @invoice.void_v2? || @invoice.deposited? || @invoice.paid? %>
```

app/views/hcb_codes/_invoice.html.erb
```diff
@@ -140,11 +145,17 @@
...
+  <%= link_to (@invoice.archived? ? "Un-archive" : "Archive"),
+    (@invoice.archived? ? invoice_unarchive_path(@invoice) : invoice_archive_path(@invoice)),
+    disabled: !organizer_signed_in?,
+    method: :post,
+    class: "btn bg-muted" unless @invoice.void_v2? %>
```
garyhtou: It seems like the intention is to not allow `void` -> `archive`. However, at the moment, it's possible to `void` a previously archived invoice (and it'd be in both states!). Is that intentional?

The void button is visible for archived invoices.

app/services/invoice_service/mark_void.rb
```diff
@@ -0,0 +1,28 @@
...
+      return if invoice.paid_v2? # return if already marked paid
+
+      invoice.voided_at = DateTime.now
+      invoice.voided_by = @user
+
+      invoice.remote_invoice.void_invoice
+      invoice.sync_remote!
+      invoice.mark_void!
+
+      invoice.save
```
garyhtou: Three edits:
1. Wrap in a database transaction. If anything fails, it will rollback changes; this way invoices aren't partially voided.
2. use `close_stripe_invoice` which is already defined. You'll need to make `close_stripe_invoice` a public method. If you'd like, feel free to give it a better name (e.g.`void_stripe_invoice`)
3. Use `save!`. If there is a validation error, I'll rollback the changes and show an error to the User

```suggestion
      Invoice.transaction do
        invoice.voided_at = DateTime.now
        invoice.voided_by = @user

        invoice.close_stripe_invoice
        invoice.mark_void!

        invoice.save!
      end
```

app/controllers/invoices_controller.rb
```diff
@@ -38,11 +38,13 @@ def index
     when "paid"
       relation = relation.paid_v2
     when "unpaid"
-      relation = relation.unpaid
+      relation = relation.unpaid.excluding(relation.void_v2)
```
garyhtou: We should update the `unpaid` and `archived` scope rather than excluding voided invoices here

garyhtou (changes requested):

app/views/hcb_codes/_invoice.html.erb
rluodev: Yep, this is intentional, as the invoice can be possibly archived and then voided, at which point it will just show up solely in the voided tab

app/controllers/invoices_controller.rb
rluodev: fixed!

sarthaktexas (approved):
lgtm great job

app/services/invoice_service/mark_void.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def run
+      return if invoice.paid_v2? # return if already marked paid
+
+      Invoice.transaction do
+        invoice.voided_at = DateTime.now
+        invoice.voided_by = @user
+
+        invoice.close_stripe_invoice
+        invoice.mark_void!
```
garyhtou: Let's do `mark_void!` before `close_stripe_invoice`. This is because we can rollback `mark_void!`. But we can't rollback `close_stripe_invoice` since it's an API call to Stripe

```suggestion
        invoice.mark_void!
        invoice.close_stripe_invoice
```

app/services/invoice_service/mark_void.rb
```diff
@@ -0,0 +1,29 @@
...
+      @invoice_id = invoice_id
+      @user = user
+    end
+
+    def run
+      return if invoice.paid_v2? # return if already marked paid
+
+      Invoice.transaction do
+        invoice.voided_at = DateTime.now
+        invoice.voided_by = @user
```
garyhtou: Add `timestamp: true` to the `aasm do` block in the model. AASM will then automatically set the timestamp.

https://github.com/aasm/aasm#timestamps




You can also add a parameter to the AASM transition to take in a user.
See this as an example: https://github.com/hackclub/bank/blob/945eadcce352b789331b3058a8fe1de55e8c727a/app/models/disbursement.rb#L117-L123

sarthaktexas (approved):

app/controllers/invoices_controller.rb
```diff
@@ -161,6 +164,16 @@ def archive
     end
   end
 
+  def void
+    @invoice = Invoice.friendly.find(params[:invoice_id])
+
+    authorize @invoice
+
+    ::InvoiceService::MarkVoid.new(invoice_id: params[:invoice_id], user: current_user).run
```
garyhtou:
```suggestion
    ::InvoiceService::MarkVoid.new(invoice_id: @invoice.id, user: current_user).run
```

app/views/hcb_codes/_invoice.html.erb
```diff
@@ -52,7 +57,7 @@
   </section>
   <!-- recipient info -->
   <section class="details details--wide pt2 pb2 details--tall">
-    <% unless @invoice.paid_v2? or @invoice.deposited? %>
+    <% unless @invoice.paid_v2? or @invoice.deposited? or @invoice.void_v2? %>
```
garyhtou: i _think_ this does the same? i'm not confident about this change
```suggestion
    <% if @invoice.open_v2? %>
```

app/views/hcb_codes/_invoice.html.erb
```diff
@@ -155,11 +160,17 @@
...
     </div>
-    <%= link_to (@invoice.archived? ? "Un-archive" : "Archive"),
-      (@invoice.archived? ? invoice_unarchive_path(@invoice) : invoice_archive_path(@invoice)),
+    <%= link_to "Void this invoice",
+      invoice_void_path(@invoice.id),
       disabled: !organizer_signed_in?,
       method: :post,
-      class: "btn bg-muted" %>
+      data: { confirm: "Are you sure you want to void this invoice? Voided invoices can not be paid and this action cannot be undone." },
+      class: "btn bg-error mr1 right" unless @invoice.void_v2? || @invoice.deposited? || @invoice.paid? %>
```
garyhtou: maybe this works
```suggestion
      class: "btn bg-error mr1 right" if @invoice.open_v2? %>
```

garyhtou (commented):
looks good, just a couple small things then good to merge

PR #4723, opened 2023-12-13
Remove check deposits flipper flag dependency

rluodev:
See issue #4722.


garyhtou:
Oops! Sorry, there are a couple things here:
- Let's not touch Check Deposits just yet (re: https://github.com/hackclub/hcb/issues/4696) since Caleb has some upcoming work that may require us to re-use this flipper flag
- Also, we'd like to keep the Feature Preview card (if it exists) to show that the feature was shipped

Sorry for the confusion. Would you mind if we closed this PR? We may come back to this in the future. In the meantime, `donation_page_customization_2022_06_29` would be a great feature flag to remove

sampoder (approved):
LGTM! As a side note we should turn event preview features into components / partials like we did with user features

rluodev:
> Oops! Sorry, there are a couple things here:
> 
> * Let's not touch Check Deposits just yet (re: [Remove Check Deposits Feature Preview #4696](https://github.com/hackclub/hcb/issues/4696)) since Caleb has some upcoming work that may require us to re-use this flipper flag
> * Also, we'd like to keep the Feature Preview card (if it exists) to show that the feature was shipped
> 
> Sorry for the confusion. Would you mind if we closed this PR? We may come back to this in the future. In the meantime, `donation_page_customization_2022_06_29` would be a great feature flag to remove

@garyhtou sure that's fine! Closing this now.

PR #4733, opened 2023-12-14
Fix Sign out / Logout from the Command Bar

sampoder:
## Summary of the problem

It didn't work...

## Describe your changes

Now we make a DELETE request with fetch using the CSRF token based on: https://guides.rubyonrails.org/v6.0.0/working_with_javascript_in_rails.html#cross-site-request-forgery-csrf-token-in-ajax


app/javascript/components/command_bar/CommandBar.js
```diff
@@ -213,9 +213,18 @@ export default function CommandBar({admin = false, adminUrls = {}}) {
 		},
 		{
 			id: "logout",
```
garyhtou: would we wanna change the `id` too?

app/javascript/components/command_bar/CommandBar.js
```diff
@@ -213,9 +213,18 @@ export default function CommandBar({admin = false, adminUrls = {}}) {
...
+			name: "Sign out",
 			keywords: "sign out logout log out",
-			perform: () => (window.location.pathname = "/users/logout"),
+			perform: () => fetch(
+				"/users/logout", 
+				{ 
+					method: "DELETE", 
+					headers: {
+						'Content-Type': 'application/json',
+						'X-CSRF-Token': document.getElementsByName('csrf-token')[0].content
```
garyhtou: this may be helpful

https://github.com/hackclub/bank/blob/c5acfaeb97d1b9a05550682ce7fc37c369d64538/app/javascript/common/csrf.js#L1

garyhtou (changes requested):

app/javascript/components/command_bar/CommandBar.js
sampoder: oh yeah may as well
```suggestion
			id: "signout",
```

we use logout on the backend but sign out on the front

app/javascript/components/command_bar/CommandBar.js
sampoder: oh perfect
```suggestion
						'X-CSRF-Token': csrf()
```

garyhtou (approved):

PR #4738, opened 2023-12-14
Align feature preview name with description

polypixeldev:
## Summary of the problem

Previously the names of feature previews on the `/my/settings/previews` page were not aligned with the descriptions below the names.
Fixes #4640

## Describe your changes

I modified `_feature_preview.html.erb` to only use padding vertically (top and bottom) instead of on all sides of the name, which was causing the issue.

image


sampoder (approved):
Looks good to me! Thank you for this

garyhtou (approved):
yay it's great when things are aligned

PR #4741, opened 2023-12-14
Create a `command_bar` partial & render it on admin pages

sampoder:
## Summary of the problem

Raised by Max here: https://github.com/hackclub/hcb/pull/4432#issuecomment-1855135637

## Describe your changes

1) Made a partial to prevent repeating code
2) Rendered that partial on admin pages
3) Added a bit of CSS to ensure the colours are correct


app/views/application/_command_bar.html.erb
```diff
@@ -0,0 +1,20 @@
+<% @admin_urls = current_user&.admin? ? {
```
garyhtou: we can use `admin_signed_in?`

app/views/application/_command_bar.html.erb
garyhtou, line 1: it looks like we lost the `if` that hides this for non-admins

app/views/application/_command_bar.html.erb
sampoder: we can!

```suggestion
<% @admin_urls = admin_signed_in? ? {
```

app/views/application/_command_bar.html.erb
sampoder: Good catch, thanks!

garyhtou (approved):
great!

app/views/application/_command_bar.html.erb
```diff
@@ -0,0 +1,22 @@
+<% if admin_signed_in? %>
+   <% @admin_urls = current_user&.admin? ? {
+        "Applications"              => "https://airtable.com/tblctmRFEeluG4do7/viwGhv19cV1ZRj61a",
+        "OnBoard ID"                => "https://airtable.com/app4Bs8Tjwvk5qcD4/tblVZwB8QMUSDAd41/viwJ15CT6VHCZ0UZ4",
+        "Wires"                     => "https://airtable.com/tbloFbH16HI7t3mfG/viwzgt8VLHOC82m8n",
```
sampoder:
```suggestion
```

garyhtou: good catch

PR #4742, opened 2023-12-14
Command Bar: Restrict what demo mode / Outernet Guilds / hardware grants have access to

sampoder:
## Summary of the problem

This the main problem blocking us from making the command bar public. Basically we shouldn't list pages that events / orgs don't have access to.

## Describe your changes

1) Started returning category / demo mode status in the API route
2) Filtered our the subitems based on this
3) I also stopped displaying demo mode accounts in the list for admins as I'm fairly sure they're irrelevant


garyhtou:
Can we pull the `not_demo_mode` another another PR? (that portion is good to merge).

I think we should handle the restrictions in the backend

sampoder:
Sure thing! What's your vision for doing it on the background?

sampoder:
Split it off into #4764

garyhtou:
> Sure thing! What's your vision for doing it on the background?

Here are my ideas/thoughts (not necessarily to say that we should go with them; there could be better ways).

Currently, we have a bunch of conditions sprinkled around the code to control what people/organizations have access to. The concept of "what people have access to" falls under **authorization**; which is what Pundit is for. What I envision is that we move a lot of this logic into Pundit so that it can easily be reused. Afterward, to determine whether we show a route or not in the command bar, we'd ask Pundit whether this user/event has access to it.

This solution would fundamentally change how we pass routes to the frontend. I would like to find a way to add decorators to the top of Controller actions and have them generate command bar actions. We can chat more on a call about this

garyhtou:
Maybe we can use Pundit to determine what pages are accessible. We'd need to move the category logic into `EventPolicy` and then call `EventPolicy.new(current_user, @event).donation_overview?`

> [!NOTE]
> **Unrelated:**
>
> `EventPolicy.new(current_user, @event).donation_overview?` should also be used to conditionally render navigation items. Currently `events/_nav.html.erb` has duplicated code with logic for `@event.outernet_guild?`

garyhtou (commented):
🪄

PR #4744, opened 2023-12-14
Only run Github Actions on `pull_request` & pushes to `main`

sampoder:
## Summary of the problem

We're running into all sorts of funky Github actions: https://github.com/hackclub/hcb/issues/4717

## Describe your changes

If we merged this, we'd only run jobs on the `pull_request` event. Thoughts?


northeastprince:
I would just run CI on push but not pushes to main.

sampoder:
That would recreate the original issue, we need the `pull_request` event so that we can run CI on private forks

garyhtou:
What are your thoughts on keeping `pull_request` and `push`, but filtering pushes to only the `main` branch? I'd assume this would allow us to run checks on pull requests and the main branch.

Running checks on the `main` branch isn't completely necessary, but is nice in case there is a (rare) commit directly to `main`

sampoder:
I like that idea!

northeastprince:
The issue was caused by CI bring run on push and PR events, which is redundant with the former, so we should be fine just having CI run on push events.

sampoder:
CI does not run on push events from private forks, hence we need the pull_request event.

garyhtou (approved):

PR #4747, opened 2023-12-14
Use native `create`/`destroy` actions for receipts

cjdenio:


sampoder (approved):
This looks good to me! Let's get this merged and then update https://github.com/hackclub/hcb/pull/4916

sampoder:
We'll also need to update https://github.com/hackclub/hcb/pull/5062, though that should be easy.

garyhtou (approved):

PR #4750, opened 2023-12-14
Add a visual indicator for admin tools in the cmd bar

sampoder:
## Summary of the problem

See #4745. Closes #4745.

## Describe your changes

We now use the same orange-y overlay for admin tools. Orange!

Screenshot 2023-12-14 at 10 44 02 AM

Screenshot 2023-12-14 at 10 44 12 AM

I guess it's more like yellow on light mode? Yellow!


garyhtou (approved):
snazzy 🆒

PR #4766, opened 2023-12-15
Track The Command Bar! (& add a bit Turbo jazz)

sampoder:
## Summary of the problem

Max describes it well: #4746

## Describe your changes

1) Used Fullstory to track when some navigates using the command bar. Also use it track what people type, that way we can capture failures.

2) Refactored the navigation a bit to use this helper function. The helper function also uses Turbo for inter-HCB navigation!


app/javascript/components/command_bar/CommandBar.js
```diff
@@ -43,6 +43,19 @@ const groupNameStyle = {
   opacity: 0.5,
 };
 
+function navigate(to){
```
garyhtou: It may be nice if `navigate` returned an (uncalled) function, this way we don't need to wrap the `navigate` call with a lambda. what are your thoughts?

sampoder: Done, love that idea!

app/javascript/components/command_bar/CommandBar.js
```diff
@@ -548,7 +537,17 @@ export default function CommandBar({admin = false, adminUrls = {}}) {
...
+              <KBarSearch 
+                style={searchStyle} 
+                defaultPlaceholder={"Search..."} 
+                onChange={
+                  () => {
+                    window?.FS?.event("command_bar_typed", {
+                      query: event.target.value
+                    })
+                  }
+                }
```
garyhtou: Is `onChange` debounced? Or will it be called 234234 times as users type into the search bar?

garyhtou (commented):
Amazing! just one question

app/javascript/components/command_bar/CommandBar.js
sampoder: not really but this was by design, i'd like to have as much data as possible about the usage of the command bar. hopefully this data can tell us what queries people try but don't get any results for

app/javascript/components/command_bar/CommandBar.js
```diff
@@ -76,89 +90,79 @@ function ActionRegister() {
               section: 'Organizations',
             })),
             ...data.map(event => ({
-              id: `${event.slug}-transactions`,
-              name: 'Transactions',
-              perform: () => (window.location.pathname = `/${event.slug}`),
-              icon: <Icon glyph="transactions" size={16} />,
+              id: `${event.slug}-home`,
+              name: 'Home',
```
sampoder:
```suggestion
              id: `${event.slug}-transactions`,
              name: 'Transactions',
```

app/javascript/components/command_bar/CommandBar.js
```diff
@@ -76,89 +90,79 @@ function ActionRegister() {
...
+              perform: navigate(`/${event.slug}`),
+              icon: <Icon glyph="home" size={16} />,
```
sampoder:
```suggestion
              icon: <Icon glyph="transactions" size={16} />,
```

app/javascript/components/command_bar/CommandBar.js
```diff
@@ -241,281 +245,287 @@ export default function CommandBar({ admin = false, adminUrls = {} }) {
...
+      perform: navigate('/admin/grants'),
+    },
+    {
+      id: 'admin_tool_7',
+      section: 'Admin Tools',
+      priority: Priority.HIGH,
+      name: 'Wires',
+      icon: <Icon glyph="bolt" size={16} />,
+      perform: navigate(adminUrls['Wires']),
     },
```
sampoder:
```suggestion
```

PR #4768, opened 2023-12-15
Highlight transaction memos when renamed on homepage

polypixeldev:
## Summary of the problem

Previously, when a transaction was renamed on the homepage (with the "Rename Transactions on the Ledger" feature preview enabled) there was no feedback to the user after the rename was submitted and the user might've not known if the rename had really succeeded.
Fixes #4641 

## Describe your changes

This PR adds a local field called `renamed` to the memo partial, which is set to false by default and set to true once a transaction rename on the homepage succeeds. It uses the same highlight styling as the highlight when a user visits an HCB code, except with some vertical padding to make it look a little better.

[Screencast from 2023-12-14 21-05-46.webm](https://github.com/hackclub/hcb/assets/79737178/7744652a-1b32-4058-8c6e-db6109769557)


sampoder (changes requested):
Oh that's beautiful! Would it be possible to make the whole transaction highlighted? Or make the styles feel less rectangular / sharp?

polypixeldev:
I added a small rounding to the memo, but I'm not sure how to do anything else. Adding horizontal padding would cause the memo to become unaligned, and I don't know how to make the whole transaction highlighted. It might be possible with the `:has` selector in CSS, but it's still new and doesn't have great browser support yet.
Here's a video with the rounding added:
[Screencast from 2023-12-14 21-46-44.webm](https://github.com/hackclub/hcb/assets/79737178/eb146f3a-83a0-4fa8-aa55-2c7558542bd9)

sampoder:
Gotcha! We use `:has` a fair bit in the codebase right now so I'd be comfortable using it in this case. Let me see if I can try to get something working with it.

sampoder (approved):
Great work on this; LGTM!

garyhtou (approved):
Great PR description! love it!

app/assets/stylesheets/components/_transactions.scss
```diff
@@ -117,6 +117,11 @@ html[data-dark='true'] {
   }
 }
 
+.transaction--positive:has(.renamed),
```
garyhtou: one minor feedback is that maybe we should use `.highlight` as the class name. I can totally see this css being used for other stuff as well such as when adding/removing a tag from a transaction

garyhtou:
@sampoder go ahead and merge ur available

PR #4777, opened 2023-12-17
Geocode HCB Wrapped Locations

sampoder:
## Summary of the problem

I wouldn't want to be know as the hometown hero of 90069.... 

## Describe your changes

Now we geocode them! It slows things down a little but it helps with things.


garyhtou:
~~@sampoder I did some quick back-of-the-napkin math. I think it'll take 20 hours to run the full geocode since we're rate limited to 70 requests per hour. I don't think this is gonna be possible unfortunately :(~~

edit: wait!! there's still time. i saw it was sunday and thought it was sunday night. but nope, it's still got more than 24 hours til launch

PR #4786, opened 2023-12-18
Reply to receipt upload emails

cjdenio:
## Summary of the problem

When you email HCB with a receipt, the confirmation back from HCB shows up as a separate thread rather than a reply.

## Describe your changes

This PR sets the `In-Reply-To` header, which should theoretically tell mail clients to treat these messages as replies. I've tested this and verified that the header is set correctly.

Resolves #4769


cjdenio:
Update: it seems to work in Apple Mail but not Gmail

PR #4788, opened 2023-12-18
Drop admin_at column

rluodev:
This closes issue #4767.


sarthaktexas (approved):

garyhtou (changes requested):
Can we please break this up into 3 PRs? Here's some context why: https://github.com/hackclub/hcb/issues/4697#issuecomment-1854569689

PR #4796, opened 2023-12-19
HCB Metrics: Fill in the blanks (date style)

sampoder:
## Summary of the problem

Things break when we don't know if you spent money.

## Describe your changes

Basically we now just mention dates where you spent $0.

There is almost certainly a cleaner way of doing this, suggestions much appreciated.


garyhtou (approved):

garyhtou:
@sampoder FYI, the keys likely won't be returned in a sorted order. I'm not sure if your frontend code depends on that

PR #4797, opened 2023-12-19
HCB Wrapped 2023

garyhtou:
it's time.


app/views/users/wrapped/index.html.erb
```diff
@@ -1 +1,2 @@
-coming soon to a screen near you :)
+<% title("Wrapped 2023") %>
```
sampoder:
```suggestion
<% title("🎁 Wrapped 2023") %>
```
emoji? the title looks a bit plain rn

app/controllers/users/wrapped_controller.rb
```diff
@@ -3,11 +3,31 @@
...
     end
 
     def data
+      headers["Content-Type"] = "application/json"
+      headers["Content-disposition"] = "attachment; filename=wrapped.json"
+      set_streaming_headers
+
+      response.status = 200
+
+      self.response_body = @data.to_json
```
cjdenio: I think this could just be
```suggestion
      headers["Content-Disposition"] = "attachment; filename=wrapped.json"

      render json: @data
```
unless I'm mistaken.

app/controllers/users/wrapped_controller.rb
```diff
@@ -3,11 +3,31 @@
 module Users
   class WrappedController < ApplicationController
     skip_after_action :verify_authorized
+    before_action :set_data
 
     def index
+      unless Flipper.enabled?(:bank_wrapped, current_user)
+        render plain: "HCB Wrapped coming soon! I heard #{current_user.first_name + "'s" || "you're"} on the naughty list 🎅" and return
```
cjdenio:
```suggestion
        render plain: "HCB Wrapped coming soon! I heard #{current_user&.first_name.presence&.concat("'s") || "you're"} on the naughty list 🎅" and return
```

Properly handle missing first names. The current implementation would just output "I heard 's on the naughty list"

app/views/users/wrapped/index.html.erb
```diff
@@ -1 +1,2 @@
-coming soon to a screen near you :)
+<% title("🎁 Wrapped 2023") %>
+<%= react_component "holiday/Wrapped", { data: @data }, { camelize_props: false } %>
```
cjdenio:
```suggestion
<%= react_component "holiday/Wrapped", { data: @data } %>
```

cjdenio: @garyhtou this is weird, I tried both locally and it generated the same HTML

Screenshot 2023-12-19 at 1 03 59 AM

Screenshot 2023-12-19 at 1 03 40 AM

garyhtou: it converts
```ruby
{
  hi_hello: 1,
  "Hack Club HQ": 2
}
```

to something like

```JSON
{
  "hiHello": 1,
  "hack Club HQ": 2
}
```

garyhtou: See https://github.com/reactjs/react-rails/blob/master/docs/component-generator.md#camelize-props

garyhtou: we had ran into an issue where data was being rendered with incorrect capitalization

PR #4798, opened 2023-12-19
Admin: User information page

devramsean0:
## Summary of the problem

This completes Gary's request in #4651 by adding a more complete page to see any relevant information about a user

## Describe your changes

I decided to use the admin controller for this page as I concluded it's an admin page, so it's protected by that authentication level, It also appears to take an id parameter for all its routes too.
I then searched each model using the user ID. and put that information in a table so that it only shows if the user has 1 of those transactions/cards

image


app/controllers/admin_controller.rb
```diff
@@ -1091,6 +1091,18 @@ def check_deposits
     render layout: "admin"
   end
 
+  def user_information
+    @user = User.find(params[:id])
+    @invoices = Invoice.where(creator_id: @user.id)
+    @donations = Donation.where(email: @user.email) # Probably not the best way to fetch this but I don't have any other ideas
```
sampoder: Does it make sense for these to be displayed? These aren't ever really actions taken by the user on HCB.

devramsean0: I think it makes sense to show the Invoices that a user has created, not sure if donations make sense though.

sampoder: Agree with displaying invoices, but I think we can scrap donations.

devramsean0: Yeah, I agree. I've now removed it

sarthaktexas (approved):
Approved! Tried it on my side and worked like a charm. I did not try LOB Checks.

app/views/admin/user_information.html.erb
```diff
@@ -0,0 +1,183 @@
+<h1>Information for <%= @user.full_name %></h1>
```
sampoder: Do we still want this page? Sorry I'm not caught up on this PR. Are we adding the information to both places?

devramsean0: Adding it to both places seems excessive (and hardish to maintain)
I'll remove the new page

config/routes.rb
```diff
@@ -214,6 +214,7 @@
       get "invoice_process", to: "admin#invoice_process"
       post "invoice_mark_paid", to: "admin#invoice_mark_paid"
       get "grant_process", to: "admin#grant_process"
+      get "user", to: "admin#user_information"
```
garyhtou: I think this is no longer being used

app/views/users/edit_admin.html.erb
```diff
@@ -52,6 +50,181 @@
...
+            <% @disbursements.each do |disbursement| %>
+              <tr>
+                <td><%= disbursement.name %></td>
+                <td><%= disbursement.amount %></td>
+                <td><%= disbursement.aasm_state %></td>
+                <td><%= link_to "View", hcb_code_path(disbursement.hcb_code) %></td>
+              </tr>
+            <% end %>
+          </table>
+        <% end %>
```
garyhtou: Currently, it is not possible for this data to be shown because the instance variables are never defined.

For example, `@invoice` is never defined

garyhtou (changes requested):

app/views/users/edit_admin.html.erb
devramsean0: Ack, I must have forgotten to add those in when we changed the page it's on

app/controllers/users_controller.rb
```diff
@@ -331,6 +331,13 @@ def edit_admin
     @onboarding = @user.full_name.blank?
     show_impersonated_sessions = admin_signed_in? || current_session.impersonated?
     @sessions = show_impersonated_sessions ? @user.user_sessions : @user.user_sessions.not_impersonated
+
+    # User Information
+    @invoices = Invoice.where(creator_id: @user.id)
+    @check_deposits = CheckDeposit.where(created_by_id: @user.id)
+    @increase_checks = IncreaseCheck.where(user_id: @user.id)
+    @lob_checks = Check.where(creator_id: @user.id)
```
garyhtou: One cool thing about Rails is that you don't need to explicitly pass the id!

```suggestion
    @invoices = Invoice.where(creator: @user)
    @check_deposits = CheckDeposit.where(created_by: @user)
    @increase_checks = IncreaseCheck.where(user: @user)
    @lob_checks = Check.where(creator: @user)
```

app/views/users/edit_admin.html.erb
garyhtou: all good!

garyhtou: It looks like we're still missing `@ach_transfer` and `@disbursement`

devramsean0:
closes #4651

app/views/users/edit_admin.html.erb
```diff
@@ -52,6 +50,181 @@
...
+              <tr>
+                <td><%= invoice.id %></td>
+                <td><%= invoice.total %></td>
+                <td><%= invoice.status %></td>
+                <td><%= link_to "View", hcb_code_path(invoice.hcb_code) %></td>
+              </tr>
+            <% end %>
+          </table>
+        <% end %>
+        <% if @checks_deposited.present? %>
```
garyhtou:
```suggestion
        <% if @checks_deposits.present? %>
```

app/views/users/edit_admin.html.erb
```diff
@@ -52,6 +50,181 @@
...
+          <h2>Deposited Checks</h2>
+          <table>
+            <tr>
+              <th>Status</th>
+              <th>Rejection reason</th>
+              <th>Amount (cents)</th>
+              <th>Date</th>
+            </tr>
+            <% @checks_deposited.each do |check| %>
```
garyhtou:
```suggestion
            <% @checks_deposits.each do |check| %>
```

garyhtou (approved):
@sarthaktexas Please test by impersonating in prod after merging

app/views/users/edit_admin.html.erb
```diff
@@ -52,6 +50,181 @@
           <%= form.submit "Save settings" %>
         </div>
       <% end %>
+    </section>
+    <% if current_user.admin? %>
```
sarthaktexas:
```suggestion
    <% if admin_signed_in? %>
```

sarthaktexas (changes requested):

sarthaktexas (approved):

PR #4802, opened 2023-12-19
Use "Transactions" instead of "Home"

sampoder:
## Summary of the problem

Issue raised by @dwahbe in #4801. Current PR makes the exact change described in the issue.

An alternative I'd like to throw in is the `bank-account` icon and the word "Ledger"

Screenshot 2023-12-19 at 9 46 49 AM

Though this kind of conflicts with the account number / routing number stuff.

## Describe your changes

Screenshot 2023-12-19 at 9 45 14 AM


sarthaktexas:
I don't like the alternative. I think the Transactions wording with the $ inside of a circle makes the most sense here.

garyhtou (approved):
great!

PR #4803, opened 2023-12-19
Prepare User model for encrypting the birthday column

farhat-ha:
## Summary of the problem

## Describe your changes

- Following the LockBox [documentation](https://github.com/ankane/lockbox#migrating-existing-data) I marked the date of birth as an encrypted field and prepared it for migration
- Create the migration file to add birthday_ciphertext to user model
- Update schema.rb

# Note
Backfill the data in the Rails console:
From rails console.
```
Lockbox.migrate(User)
```


db/schema.rb
albertchae, line 1: @farhat-ha FYI I reverted the changes to this file except for the new column you added. As an orthogonal follow up we should find out why the schema dump had columns etc in a different order.

albertchae (approved):
LGTM, gonna deploy now

albertchae:
Deployed in https://dashboard.heroku.com/apps/bank-hackclub/activity/releases/5a045c9d-f458-4327-8533-f575b5bb2dca

Ran `Lockbox.migrate(User)`

Verified that users with nil birthdays still have nil migrated_birthdays and that users with a birthday set now have one in the encrypted column

PR #4804, opened 2023-12-19
Ignore plaintext birthday column before dropping the column

farhat-ha:
## Summary of the problem

This is a followup to https://github.com/hackclub/hcb/pull/4803

# Before merging! 
- [x] https://github.com/hackclub/hcb/pull/4803 needs to be merged
- [x] migration is ran to add `birthday_ciphertext` to user
- [x] Data backfilled (`Lockbox.migrate(User)` from rails console)

## Describe your changes
Ignore the birthday field in preparation for it to be dropped in a subsequent PR


app/models/user.rb
```diff
@@ -106,7 +108,7 @@ class User < ApplicationRecord
   after_update :update_stripe_cardholder, if: -> { phone_number_previously_changed? || email_previously_changed? }
 
   validates_presence_of :full_name, if: -> { full_name_in_database.present? }
-  validates_presence_of :birthday, if: -> { birthday_in_database.present? }
+  validates_presence_of :birthday, if: -> { birthday_ciphertext_in_database.present? }
```
albertchae: nice catch

albertchae (approved):
LGTM

albertchae:
Verified after deploy that `user.birthday` works and is pulling from the encrypted column

PR #4805, opened 2023-12-19
Drop plaintext birthday column

farhat-ha:
## Summary of the problem

Add migration to drop the plaintext birthday column from the user table. 

# Before merging these PRs need to be merged

- [x] https://github.com/hackclub/hcb/pull/4803
- [x] https://github.com/hackclub/hcb/pull/4804


albertchae (approved):

albertchae:
As with https://github.com/hackclub/hcb/pull/4804#issuecomment-1876697843, verified again after deploy that user.birthday works and is pulling from the encrypted column. Also opened a psql console to see in raw sql that there is no `birthday` column and only a `birthday_ciphertext`

PR #4808, opened 2023-12-19
Create account numbers on Column

cjdenio:


app/controllers/column/webhooks_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    private
+
+    def handle_ach_incoming_transfer_scheduled
+      return if @object[:type].downcase == "credit" || @object[:amount] <= 100 # Allow incoming ACH credits and small debits
+
+      # If this ACH debit is to an org's account number, reject it
+      if AccountNumber.exists?(column_id: @object[:account_number_id])
+        ColumnService.post("/transfers/ach/#{@object[:id]}/return", return_code: "R08")
+      end
```
garyhtou: we should move this into the model

cjdenio: Which model?

garyhtou: AccountNumber

PR #4815, opened 2023-12-20
Add password to example development env DATABASE_URL

polypixeldev:
## Summary of the problem

When creating a new codespace, the container build currently fails due to the database setup commands in `.devcontainer/boot.sh` not being able to connect to the database without a password provided.

## Describe your changes

I added the default postgres password (`postgres`) to the example database connection URL in `.env.development.example`, which is also the password specified in `.devcontainer/docker-compose.yml`. The codespace build succeeds after this change since it copies the values from the `.env.development.example` to `.env.development`.


sampoder (approved):
Looks good to me!

garyhtou:
Can either of yall please make sure this new `.env.development.example` works with the existing `.docker_dev_setup.sh` script?

garyhtou:
I believe `.docker_dev_setup.sh` should work fine since we've got <a href="https://hub.docker.com/_/postgres#:~:text=PostgreSQL%2010%2B).-,POSTGRES_HOST_AUTH_METHOD,-This%20optional%20variable">POSTGRES_HOST_AUTH_METHOD</a> set in `.docker_compose` which should allow connections without authentication (e.g. a password; or the right password)

polypixeldev:
Dev setup script run without issues!

sampoder:
Awesome!

garyhtou (approved):

garyhtou:
Great! thank you!

PR #4822, opened 2023-12-21
Email about ACH transfers to deprecated Increase account number

cjdenio:
Closes #4821


app/views/raw_increase_transaction_mailer/deprecated_account_number.html.erb
```diff
@@ -0,0 +1 @@
+hey there lol
```
garyhtou: _hehe lol_ 🦦

garyhtou:
thank you!

PR #4827, opened 2023-12-22
Disbursement Scheduling

zoya-hussain:
## Summary of the problem

draft pull request for issue #4672
- migration 
- job scheduling
- form for scheduling
- validation states and transition
- name and scheduling for disbursement for admin user in disbursement_controller
- scheduled and approved state

## Describe your changes

closes #4672


sampoder (dismissed):
PR looks good to go to me!

I've tested:

* Creating scheduled disbursements
* Cancelling scheduled disbursements
* Running the daily job

I'd like a bit of feedback on how we're currently creating the canonical transactions, @garyhtou thoughts?

cjdenio (dismissed):
Thanks for this! One more thing: these two services need to be updated to _not_ import pending transactions for disbursements that have been scheduled, but not sent.

https://github.com/hackclub/hcb/blob/main/app/services/pending_transaction_engine/raw_pending_outgoing_disbursement_transaction_service/disbursement/import.rb
https://github.com/hackclub/hcb/blob/main/app/services/pending_transaction_engine/raw_pending_incoming_disbursement_transaction_service/disbursement/import.rb

As an example, here's how this is implemented for ACH transfers:
https://github.com/hackclub/hcb/blob/3cb5616635280241f249e7088d9e5be096fd1005/app/services/pending_transaction_engine/raw_pending_outgoing_ach_transaction_service/outgoing_ach/import.rb#L21
(For disbursements we'd use `reviewing` instead of `pending`)

app/services/pending_transaction_engine/raw_pending_incoming_disbursement_transaction_service/disbursement/import.rb
```diff
@@ -14,6 +14,7 @@ def run
 
         def disbursements
           ::Disbursement.where.missing :raw_pending_incoming_disbursement_transaction
+          ::Disbursement.in_transit.or(::Disbursement.reviewing.where(scheduled__on: nil).where.missing :raw_pending_incoming_disbursement_transaction)
```
sampoder:
```suggestion
          ::Disbursement.in_transit.or(::Disbursement.reviewing.where(scheduled__on: nil)).where.missing :raw_pending_incoming_disbursement_transaction
```

We shouldn't need that line any more! Also moved around the brackets.

app/services/pending_transaction_engine/raw_pending_outgoing_disbursement_transaction_service/disbursement/import.rb
```diff
@@ -14,9 +14,11 @@ def run
 
         def disbursements
           ::Disbursement.where.missing :raw_pending_outgoing_disbursement_transaction
+          ::Disbursement.in_transit.or(::Disbursement.reviewing.where(scheduled__on: nil).where.missing :raw_pending_outgoing_disbursement_transaction)
```
sampoder:
```suggestion
          ::Disbursement.in_transit.or(::Disbursement.reviewing.where(scheduled__on: nil)).where.missing :raw_pending_outgoing_disbursement_transaction
```

Same as above :D

app/models/disbursement.rb
```diff
@@ -289,4 +305,10 @@ def events_are_not_demos
     self.errors.add(:source_event, "cannot be a demo event") if source_event.demo_mode?
   end
 
+  def scheduled_on_must_be_in_the_future
+    if scheduled_on.present? && scheduled_on.before?(Date.today)
```
garyhtou: `Date.today` gives us today at UTC midnight (first second of today). This means it would allow people to schedule a Disbursement for Today.

I think that might be an unintended behavior since it's the same as just sending it normallly.

```suggestion
    if scheduled_on.present? && scheduled_on.before?(Time.now.end_of_day)
```

Using `Time.now.end_of_day` would result in tomorrow being the earlier possible time. What are your thoughts on this?

app/services/disbursement_service/create.rb
```diff
@@ -26,20 +27,26 @@ def run
...
+        i_cpt = ::PendingTransactionEngine::CanonicalPendingTransactionService::ImportSingle::IncomingDisbursement.new(raw_pending_incoming_disbursement_transaction: rpidt).run
+        o_cpt = ::PendingTransactionEngine::CanonicalPendingTransactionService::ImportSingle::OutgoingDisbursement.new(raw_pending_outgoing_disbursement_transaction: rpodt).run
 
-      # 3. Map to event
-      ::PendingEventMappingEngine::Map::Single::IncomingDisbursement.new(canonical_pending_transaction: i_cpt).run
-      ::PendingEventMappingEngine::Map::Single::OutgoingDisbursement.new(canonical_pending_transaction: o_cpt).run
+        # 3. Map to event
+        ::PendingEventMappingEngine::Map::Single::IncomingDisbursement.new(canonical_pending_transaction: i_cpt).run
+        ::PendingEventMappingEngine::Map::Single::OutgoingDisbursement.new(canonical_pending_transaction: o_cpt).run
+      end
```
garyhtou: Since we're duplicating this code, can we extract this to a service?

garyhtou (commented):
I love that we're adding this! sorry it took so long for me to get around to this PR.

I've added a couple comments. I'll finish reviewing this during one of my calls with @sampoder. If you'd to join in on that call, just lmk!

---

Can you please add some screenshots after the changes?

app/services/disbursement_service/create.rb
sampoder: Hmm, wdym? Can you point to the lines we could extract into a service

sampoder: Oh I see now

app/views/admin/disbursement_process.html.erb
```diff
@@ -54,6 +54,11 @@
 
     <p>This is a human review check.</p>
 
+    <% if @disbursement.scheduled_on.present? %>
+      <br>
+      <small>Once approved, this disbursement will automatically send on <%= @disbursement.scheduled_on.strftime("%Y-%m-%d") %>.</small>
```
garyhtou: Copy edit because this message will still show after approval

```suggestion
      <small>When approved, this disbursement will automatically send on <%= @disbursement.scheduled_on.strftime("%Y-%m-%d") %>.</small>
```

garyhtou:
@cjdenio, when you get a chance, can you please take a look at this and approve it?

The goal is to have scheduled disbursements. When a disbursement is scheduled, it will affect the balance of the originating org (as it currently does all disbursements). However, the destination org's balance will only be affected AFTER the scheduled date.

sampoder (approved):

PR #4828, opened 2023-12-22
Improve phone number verification UI

manuthecoder:
## Summary of the problem
See #4642 

## Describe your changes
Adds an event listener to the modal trigger to programmatically click on the "Send verification code" button


app/javascript/components/settings/SmsVerification.js
```diff
@@ -72,35 +91,68 @@ const SmsVerification = ({ phoneNumber, enrollSmsAuth = false }) => {
...
           <p>
-            <a href="#" onClick={handleClick} className={loading ? 'muted wait' : 'pointer'}>{validationSent ? 'Resend code' : 'Send verification code'}</a>
-            {' '}
+            <a
+              href="#"
+              onClick={handleClick}
+              ref={triggerRef}
+              className={loading ? 'muted wait' : 'pointer'}
+            >
+              {validationSent ? 'Resend code' : 'Send verification code'}
```
sampoder: Maybe it'd make sense to have a "Loading" message while we're sending a verification code. Locally, when I test it I first see 'Send verification code' when opening the modal.

sampoder (commented):
The code seems good to me! Though I wonder if there's a cleaner / simpler way of doing it? ~~Maybe we should refactor it to use Turbo / Stimulus instead of React, similar to the receipt email generation part of the settings page.~~

cjdenio:
Closing for now as this isn't quite polished or production-ready.

PR #4839, opened 2023-12-23
Add a "Search" action for the Command Bar

sampoder:
## Summary of the problem

Wouldn't it be incredible if we could use the Command Bar to "search" HCB? Imagine searching for a card or a transaction or a donation or... anything on HCB!

## Describe your changes

This is a draft pull request that models out how this can be achieved using the `kbar` package. Here's a visual demo of how it works:

https://github.com/hackclub/hcb/assets/39828164/a9ad535c-8caf-468b-ac77-0999e50045be

To make this shippable, we would need to:

* Expand the SearchController to... well search HCB! At the moment it's just a dummy API.
    * Would using Caleb's API make sense here instead of implementing a new search API?
* Map out how users select "what" they're searching for, how do we distinguish between searching for cards, donations, or transactions? Right now we... can't.
    * One option would be having people add modifiers to their search, eg. `@card`:
        Screenshot 2023-12-23 at 12 17 06 AM
        If we were to do this, we could add a bunch of additional modifiers! So many modifiers... we would just need to design a custom search input for `kbar` that adds fancy styles to these modifiers.
    
    * Another, less flexible but easier to implement solution, would be to have people select what they'd want to search before searching. If possible, I'd prefer to have a modifiers setup.
        Screenshot 2023-12-23 at 12 22 39 AM

* We'd need to make some custom styles for this, espc. if we got with the modifiers setup.

But yeah! Here's a potential structure for this sort of thing... was a funny little problem to try and figure out. 

There's probably a world where we could replace the whole command bar with search... this doesn't quite do that but it lays the foundations.


sampoder:
Currently I'm trying to think how we could do this in CSS:

Frame 1 (1)

(to do the modifiers thing)

sampoder:
So I was thinking a bit about syntax for the modifiers. Here's what I'm kinda landing on:

`@` is reserved for defining "types" (transactions, cards, organisations, donations etc.), adding a `:` after the type allows you to define a subtype (eg. `@donations:recurring` or `@txns:card`.

> question: should you be allowed more than one type? I think yes.

`[]` should be used for adding conditionals. We should figure out a way such that you can use these operators `>`, `<`, `=`, `>=`, and `<=`. And imo the syntax should be `[(property) (operator) (value)]`. Different types of objects will have different properties, we should use the properties to filter out the types that people are searching for. For example if someone adds `[date > dec 11]` we should then only search types which have dates for example transactions and donations but not cards.

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,164 @@
...
+    
+    authorised_results = []
+    
+    results.each do |result|
+      begin
+        Pundit.authorize(current_user, result, :show?)
+        authorised_results << result
+      rescue Pundit::NotAuthorizedError
+      end
+    end
```
sampoder: this is painfully slow!.... **painfully**.

sampoder:
bug to look into: at the moment, this is breaking events on the command bar

sampoder:
the events bug has been fixed. the big problem at the moment is performance, at the moment all the logic seems to working but every search is really slow. (ambiguous searches are the slowest). 

i think the main area where we could speed things up is the authorising stuff. all stuff to explore tomorrow...

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,307 @@
+# frozen_string_literal: true
+
+class SearchController < ApplicationController
+  include Rails::Pagination
+  include UsersHelper
+  skip_before_action :signed_in_user
```
garyhtou:
```suggestion
```

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,307 @@
+# frozen_string_literal: true
+
+class SearchController < ApplicationController
+  include Rails::Pagination
```
garyhtou: is this still used?

should we use `kaminari` instead?

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,307 @@
...
+
+    begin
+      shorthands = {
+        "org"          => "organization",
+        "organisation" => "organization",
+        "txn"          => "transaction"
+      }
+      sub_shorthands = {
+        "ach"      => "ach_transfer",
+        "transfer" => "account_transfer",
```
garyhtou: what is an `account_transfer`? should that be `disbursement`?

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,307 @@
...
+        "card"     => "card_charge",
+        "check"    => "check_deposit",
+        "fee"      => "fiscal_sponsorship_fee"
+      }
+
+      parsed = params[:query]&.scan(/(?:[^@]+|@[^@]+)/)&.map { |str|
+        type = str.match(/@(\w+)/)&.[](1)
+        conditions = str.scan(/\[([^\]]+)\]/)
+        subtype = str.match(/@#{type}:(\w+)/)[1] rescue nil
```
garyhtou:
```suggestion
        subtype = str.match(/@#{type}:(\w+)/)&.[](1)
```

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,307 @@
...
+              user: if x.instance_of?(CanonicalTransaction) && x.transaction_source_type == "RawStripeTransaction"
+                      x.stripe_cardholder.user.full_name
+                    else
+                      x.instance_of?(StripeCard) ? x.stripe_cardholder.user.full_name : ""
+                    end,
+              type: x.class,
+              balance: if x.instance_of?(Event)
+                         x.balance
+                       else
+                         0
```
garyhtou: `nil`?

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,307 @@
...
+        end
+      end
+
+      respond_to do |format|
+        format.json do
+
+          results = authorised_results.map { |x|
+            {
+              label: x["last4"] || x["custom_memo"] || x["memo"] || x["name"] || x["full_name"] || "Unknown",
+              object: x,
```
garyhtou: We're dumping attributes that not all users have access to. Is there a way to get rid of `object`?

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,307 @@
...
+      end
+    end
+  end
+
+  private
+
+  def convert_to_float(input)
+    match = input.match(/\d+(\.\d{1,2})?/)
+    if match
+      result = match[0].to_f.round(2)
```
garyhtou: double check that `[0]` returns the entire capture and not just the first capture group

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,307 @@
...
+          "conditions" => conditions.select { |condition| condition[0].split.length >= 3 }.map { |condition|
+            {
+              "property": condition[0].split[0],
+              "operator": condition[0].split[1],
+              "value": condition[0].split[2..].join(" ")
+            }
+          },
+          "query"      => subtype ? str.gsub(/@\w+:\w+\s*|\[.*?\]/, "").strip : str.gsub(/@\w+\s*|\[.*?\]/, "").strip
+        }
+      }
```
garyhtou: Create a `Search::Parser` that handles parsing and validation of query

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,307 @@
...
+          item["conditions"]&.each do |condition|
+            potential_types.each do |type|
+              potential_types.delete(type) unless types[type]["properties"].include?(condition)
+            end
+          end
+        else
+          potential_types = [item["type"]]
+        end
+
+        potential_types.each do |type|
```
garyhtou: Create a `Search::Engine` class for taking in a `Search::Query` object and returns a `Search::Results`

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,307 @@
...
+            cards = cards.where("event_id = #{event_id}") if event_id
+            item["conditions"]&.each do |condition|
+              case condition[:property]
+              when "date"
+                value = Chronic.parse(condition[:value], context: :past)
+                cards = cards.where("stripe_cards.created_at #{condition[:operator]} ?", value)
+              end
+            end
+            card_stripe_id = cards.first&.stripe_id unless potential_types.length > 1 && (event_id || user_id)
+            results.concat(cards)
```
garyhtou: `Search::Engine::StripeCard`

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,307 @@
...
+        rescue Pundit::NotAuthorizedError
+        end
+      end
+
+      respond_to do |format|
+        format.json do
+
+          results = authorised_results.map { |x|
+            {
+              label: x["last4"] || x["custom_memo"] || x["memo"] || x["name"] || x["full_name"] || "Unknown",
```
garyhtou: Create a `Search::Results::CanonicalTransaction` which is a [decorator/wrapper](https://refactoring.guru/design-patterns/decorator) for CT.

All decorators should have the same interface (the set of methods defined). You can do this by having them all inherit from an abstract base class with pre-define methods that `raise NotImplementedError`.

app/controllers/search_controller.rb
sampoder: we use this term in a couple of places, it's the public facing term for disbursement

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class SearchController < ApplicationController
+  # GET /search
+  def index
+    authorize Event
```
garyhtou: Make sure the controller requires user signed in

```suggestion
```

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+        SearchService::Engine.new(query, current_user).run,
+        current_user
+      ).run
+
+      respond_to do |format|
+        format.json do
+          response.content_type = "text/json"
+          render json: SearchService::Formatter.new(results).run
+        end
+      end
```
garyhtou:
```suggestion
      render json: SearchService::Formatter.new(results).run
```

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    rescue Errors::ValidationError => e
+      # Handle any validation errors
+      respond_to do |format|
+        format.json do
+          response.content_type = "text/json"
+          render json: { error: e.message }
+        end
+      end
```
garyhtou:
```suggestion
      render json: { error: e.message }
```

app/services/search_service/engine.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module SearchService
+  class Engine
+    def initialize(query, user)
+      @query = query
+      @user = user
+      @admin = false && user.admin?
```
garyhtou:
```suggestion
```

app/services/search_service/engine/cards.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module SearchService
+  class Engine
+    class Cards
+      def initialize(query, user, context)
+        @query = query
+        @user = user
+        @admin = false && user.admin?
```
garyhtou: Remove the `false` in every one

app/services/search_service/engine.rb
```diff
@@ -0,0 +1,47 @@
...
+      @context = {
+        event_id: nil,
+        user_id: nil,
+        card_stripe_id: nil
```
garyhtou: Is there a way to use our own ID rather than Stripe's id?

app/services/search_service/engine.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module SearchService
+  class Engine
+    def initialize(query, user)
```
garyhtou: Pluralize it

garyhtou (commented):
Since we're creating our own query language here, let's first write out a spec for this language, then we can verify our implementation follows it

app/services/search_service/specification.md
```diff
@@ -0,0 +1,94 @@
...
+* Organisations
+  * Filters: date
+  * Children: Transactions, Stripe Cards, Users
+  * Subtypes: none
+* Stripe Cards
+  * Filters: date
+  * Children: Transactions
+  * Subtypes: none
+* Users
+  * Filters: date
```
garyhtou: we prob want to remove filtering users by date (`created_at`) before public launch

app/services/search_service/specification.md
```diff
@@ -0,0 +1,94 @@
...
+  * Children: Transactions, Stripe Cards, Organisations
+  * Subtypes: none
+
+**Example Searches**
+Whilst not explicitly stated, all results will be items that the current users can access.
+Outernet
+
+This will search for all Stripe Cards, Users, Transactions, and Events similar to “Outernet”. For example, you’ll get “Outernet” events and transactions which have “Outernet” in their memo.
```
garyhtou:
```suggestion

**Example Searches**
> Whilst not explicitly stated, all results will be items that the current users can access.

```
Outernet
```

This will search for all Stripe Cards, Users, Transactions, and Events similar to “Outernet”. For example, you’ll get “Outernet” events and transactions which have “Outernet” in their memo.
```

app/services/search_service/specification.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+This will search for all organisations with Outernet in their name.
+
+```
+@org Outernet
+```
+
+This will search for all organisations with Outernet in their name.
+
```
garyhtou: duplicated

```suggestion
```

app/services/search_service/specification.md
```diff
@@ -0,0 +1,94 @@
...
+```
+Outernet @txn Stickers
+```
```
garyhtou:
```suggestion
```

This search will return transactions related to stickers on anything named Outner
```

File: app/services/search_service/specification.md [hunk 24]
```diff
@@ -0,0 +1,94 @@
...
+
+```
+@org Outernet @txn Stickers
+```
+
+These two searches will both return transactions related to stickers on the Outernet organisation.
```
garyhtou:
```suggestion
This search will return transactions related to stickers on the Outernet organisation. 
```

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class SearchController < ApplicationController
+  skip_after_action :verify_authorized # do not force pundit
```
garyhtou:
```suggestion
  skip_after_action :verify_authorized # do not force pundit
  before_action :signed_in_admin # This feature is only available to admins for the time being
```

garyhtou (commented):
🪄

sampoder:
This line is the reason for `stripe_card_id`:

```
transactions = transactions.stripe_transaction
                                     .where("raw_stripe_transactions.stripe_transaction->>'card' = ?", @context[:card_stripe_id])
```

package.json
```diff
@@ -34,6 +34,8 @@
     "react_ujs": "^2.7.1",
     "react-dom": "^18.2.0",
     "react-snowfall": "^1.1.1",
+    "react_ujs": "^2.7.1",
```
cjdenio: er, we already have `react_ujs` 3 lines above?

PR #4842, opened 2023-12-26
Redact account verification transaction amounts in Transparency Mode

cjdenio:
Part of #4836 
Screenshot 2023-12-26 at 5 00 45 PM


app/helpers/application_helper.rb
```diff
@@ -341,4 +341,10 @@ def fillout_form(id, params = {}, prefix: "")
     "https://forms.hackclub.com/t/#{id}?#{URI.encode_www_form(query)}"
   end
 
+  def redacted_amount(max: 100)
+    tag.span class: "tooltipped tooltipped--w", style: "cursor: default", "aria-label": "Hidden for security" do
+      tag.span(style: "filter: blur(5px)") { render_money rand(max) }
```
garyhtou: @cjdenio if we're blurring it, can we set the content to be "$0.00" rather than a random number? I've ran into some odd issues relying on using random numbers and would prefer if we avoid it

In the API, maybe we return `null` or `0`

cjdenio: That's fine— I only chose to use a random amount because it looked slightly nicer. What were the issues you ran into?

PR #4845, opened 2023-12-28
Allow admins to instantly charge a fee for disbursements

cjdenio:
Screenshot 2023-12-28 at 1 21 34 AM


db/migrate/20231227052150_add_should_charge_fee_to_disbursements.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddShouldChargeFeeToDisbursements < ActiveRecord::Migration[7.0]
+  def change
+    add_column :disbursements, :should_charge_fee, :boolean, default: false
```
garyhtou: Some disbursements in the past were charged fees. Would it be possible to have `should_charge_fee` reflect that for old disbursements so that we can have accurate revenue calculations for previous years?

cjdenio: Yep! I'm writing a one-time job to backfill `should_charge_fee` / `CanonicalPendingTransaction#fee_waived`

garyhtou: perfect!

garyhtou: https://github.com/hackclub/hcb/pull/4874

PR #4855, opened 2023-12-31
Fix uploaded profile picture deletion

polypixeldev:
## Summary of the problem

Makes profile picture deletion work.
Fixes #4854 

## Describe your changes

When I clicked the link to delete the profile picture in development mode, the specific error was that it could not find a user (at the beginning of the `delete_profile_picture` function in `users_controller.rb`. It also showed the params of the function when it was called, which included the user id as  `user_id` instead of `id`. I'm not sure why the key is different for this specific function in `users_controller.rb`, but it seemed to fix it when I changed it.


cjdenio (approved):
Thanks!

sampoder (approved):

PR #4856, opened 2023-12-31
Don't allow for renaming on homepage for demo orgs

sampoder:
## Summary of the problem

Ruby would freak out and not load the page

## Describe your changes

Does a check to make sure it isn't demo mode.


app/views/hcb_codes/_memo.html.erb
```diff
@@ -3,7 +3,7 @@
 <% form ||= false %>
 <% prepended_to_memo ||= "" %>
 
-<% if organizer_signed_in?(hcb_code.event) %>
+<% if hcb_code.event&.demo_mode? == false && organizer_signed_in?(hcb_code.event)  %>
```
sampoder: The "&" is needed because our fake hcb codes don't have events it seems

sarthaktexas (approved):
lgtm thanks sam

sarthaktexas (approved):
lgtm thanks sam

app/views/hcb_codes/_memo.html.erb
```diff
@@ -3,7 +3,7 @@
 <% form ||= false %>
 <% prepended_to_memo ||= "" %>
 
-<% if organizer_signed_in?(hcb_code.event) %>
+<% if hcb_code.event&.demo_mode? == false && organizer_signed_in?(hcb_code.event) %>
```
garyhtou:
```suggestion
<% if organizer_signed_in?(hcb_code.event) && !hcb_code.event&.demo_mode? %>
```

garyhtou (approved):

app/views/hcb_codes/_memo.html.erb
sampoder: I don't think this will work because when event is nil it'll be true

PR #4857, opened 2023-12-31
Disable Transaction Tags on Demo Mode Accounts

sampoder:
## Summary of the problem

Causing all sorts of issues... https://hackclub.slack.com/archives/C04LDFJ5X7S/p1704064171905179

## Describe your changes

Don't let people enable the flag and hide it for folks who have enabled it (plus once fully enabled this will prevent the issue)


sarthaktexas (approved):
lgtm thanks sam

cjdenio (approved):

PR #4859, opened 2024-01-01
Disable rename if event isn't defined

sampoder:
These transactions are the fake mock data ones. See: https://github.com/hackclub/hcb/pull/4856#discussion_r1438953113


garyhtou:
i'm getting this bug
image

sampoder:
oh that's another bug let me fix that

sampoder:
that other bug is fixed!

PR #4862, opened 2024-01-01
Add Tags Filter for Admins

sampoder:
## Summary of the problem

Requested by Liv! Closes #4834.

## Describe your changes

Screenshot 2023-12-31 at 10 39 40 PM


app/controllers/admin_controller.rb
```diff
@@ -170,6 +170,7 @@ def events
 
     @events = filtered_events
     @count = @events.count
+    @tags = EventTag::Tags.constants.map { |constant| EventTag::Tags.const_get(constant) }
```
garyhtou: unfortunately, there are tags that can exist in the database, but not be defined as a constant. I think we need to do something like

```ruby
EventTag.all
```

app/views/admin/_events_filter.html.erb
```diff
@@ -57,6 +57,11 @@
   <%= form.select :funded, [["Both", "both"], ["Not Funded", "not_funded"], ["Funded", "funded"]], selected: @funded %>
 <% end %>
 |
+<%= form.label :tagged_with do %>
+  Tagged with
+  <%= form.select :tagged_with, @tags, { include_blank: "anything", selected: @tagged_with } %>
```
garyhtou:
```ruby
# app/models/event_tag.rb
class EventTag
  def full_name
    return name unless purpose.present?

    "#{purpose}: #{name}"
  end
end
```

```suggestion
  <%= form.select :tagged_with, @tags.map { |t| [t.full_name, t.id] }, { include_blank: "anything", selected: @tagged_with } %>
```

app/controllers/admin_controller.rb
```diff
@@ -1148,6 +1150,7 @@ def filtered_events(events: Event.all)
     relation = relation.not_organized_by_teenagers if @organized_by == "adults"
     relation = relation.demo_mode if @demo_mode == "demo"
     relation = relation.not_demo_mode if @demo_mode == "full"
+    relation = relation.includes(:event_tags).where(event_tags: { name: @tagged_with }) unless @tagged_with == "anything"
```
garyhtou: Unfortunately, `name` is not a good enough identifier since it's not unique. `name` is only unique within a specific `purpose`. So, we should use the id

```suggestion
    relation = relation.includes(:event_tags).where(event_tags: { id: @tagged_with }) unless @tagged_with == "anything"
```

garyhtou (commented):
Two things:
1. Get list of possible tags from the DB, not the constants
2. Use `id` to identify a tag, not it's `name`. When a tag is displayed, it should be prefixed with it's `purpose` (if it exists)

PR #4863, opened 2024-01-01
Place pending monetary requests at the top

sampoder:
## Summary of the problem

Closes https://github.com/hackclub/hcb/issues/4832.

## Describe your changes

Now they're at the top! SQL is cool, hopefully this is correct?


app/controllers/admin_controller.rb
```diff
@@ -775,7 +784,10 @@ def disbursements
     relation = relation.processing if @processing
 
     @count = relation.count
-    @disbursements = relation.page(@page).per(@per).order("created_at desc")
+    @disbursements = relation.page(@page).per(@per).order(
+      Arel.sql("CASE WHEN aasm_state = 'pending' THEN 0 ELSE 1 END"),
```
```suggestion
      Arel.sql("aasm_state = 'reviewing' DESC"),
```

app/controllers/admin_controller.rb
```diff
@@ -490,7 +490,10 @@ def ach
     relation = relation.pending if @pending
 
     @count = relation.count
-    @ach_transfers = relation.page(@page).per(@per).order("created_at desc")
+    @ach_transfers = relation.page(@page).per(@per).order(
+      Arel.sql("CASE WHEN aasm_state = 'pending' THEN 0 ELSE 1 END"),
```
```suggestion
      Arel.sql("aasm_state = 'pending' DESC"),
```

also works, see
https://github.com/hackclub/hcb/blob/1e6d8f3a511735a4e456a408fc3fab4ff613c496/app/controllers/organizer_position_deletion_requests_controller.rb#L9

sampoder: oh that's so much cleaner, let me switch over to that

cjdenio (approved):

app/controllers/admin_controller.rb
garyhtou: smart!

PR #4866, opened 2024-01-01
Double click, not single click for renaming on ledger

sampoder:
## Summary of the problem

Single clicking made life hard for people, especially when they were trying to copy the memo.

## Describe your changes

Now you need to double click!


garyhtou (approved):
i don't understand how this works 😵‍💫, but it does sure work!

app/views/hcb_codes/_memo.html.erb
```diff
@@ -26,12 +26,14 @@
             data-action="dblclick->navigation#navigate"
             title="<%= hcb_code.memo %>"
             data-navigation-location-param="<%= edit_hcb_code_path(hcb_code, inline: true, prepend: prepended_to_memo) %>"
-            data-navigation-frame-param="<%= "#{ActionView::RecordIdentifier.dom_id(hcb_code)}:memo" %>">
-        <%= link_to "#{prepended_to_memo} #{hcb_code.memo}", edit_hcb_code_path(hcb_code, inline: true, prepended_to_memo: ),
-            class: "mr2 align-middle tooltipped tooltipped--s",
-            style: "color: inherit; text-decoration: none;",
-            "aria-label": "Click to rename this transaction",
-            data: { turbo: true } %>
+            data-navigation-frame-param="<%= "#{ActionView::RecordIdentifier.dom_id(hcb_code)}:memo" %>"
```
garyhtou: i think we don't need `ActionView::RecordIdentifier.`

app/views/hcb_codes/_memo.html.erb
```diff
@@ -26,12 +26,14 @@
...
+            style="cursor: pointer">
+          <span
+            class="mr2 align-middle tooltipped tooltipped--s"
+            style="color: inherit; text-decoration: none;"
+            aria-label="Click to rename this transaction">
```
garyhtou: I think this tooltip is not accessible because there is a HTML `title` attribute set above

garyhtou (commented):
good to merge as is. just left a couple thoughts

PR #4869, opened 2024-01-02
Add Receipts to ACHs & Checks

sampoder:
## Summary of the problem

See Liv's issue: https://github.com/hackclub/hcb/issues/4831

## Describe your changes

Add's a field:

Screenshot 2024-01-01 at 9 31 39 PM

Styling thoughts?


app/controllers/ach_transfers_controller.rb
```diff
@@ -45,7 +45,16 @@ def new
 
   # POST /ach_transfers
   def create
-    @ach_transfer = @event.ach_transfers.build(ach_transfer_params.merge(creator: current_user))
+    @ach_transfer = @event.ach_transfers.build(ach_transfer_params.except(:file).merge(creator: current_user))
+
+    if ach_transfer_params[:file]
```
cjdenio: This should be moved to after the `authorize` line

app/controllers/ach_transfers_controller.rb
```diff
@@ -45,7 +45,16 @@ def new
...
+      ::ReceiptService::Create.new(
+        uploader: current_user,
+        attachments: ach_transfer_params[:file],
+        upload_method: :transfer_create_page,
+        receiptable: @ach_transfer
```
cjdenio:
```suggestion
        receiptable: @ach_transfer.local_hcb_code
```

sampoder: Oh sweet, that's actually perfect

app/models/ach_transfer.rb
```diff
@@ -47,6 +48,7 @@ class AchTransfer < ApplicationRecord
 
   include AASM
   include Commentable
+  include Receiptable
```
cjdenio: I think we should continue attaching receipts to `HcbCode`s, to keep things consistent with the way they operate right now

sampoder: That makes sense. Do we immediately create a HcbCode for increase checks?

cjdenio: Yep! `check.local_hcb_code` will create one if one doesn't exist

sampoder: Awesome :D! I cleaned everything up, let me know what you think

app/views/ach_transfers/_form.html.erb
```diff
@@ -43,6 +43,17 @@
...
 
+  <div class="field">
+    <%= form.label :file, "Attach a receipt" %>
+    <%= form.file_field :file,
+        multiple: true,
+        include_hidden: false,
+        required: false,
+        accept: "image/*,image/heic,.pdf",
+        style: "margin: 8px 0px" %>
+    <span class="muted">Required for reimbursements / goods & services payments.</span>
```
garyhtou: This copy makes a bunch of sense for reimbursements.

The "goods & services payments" portion might be a bit confusing since we're asking for a receipt before payment is made. Should we be asking for an invoice or purchase order?

app/views/ach_transfers/_form.html.erb
```diff
@@ -43,6 +43,17 @@
     </div>
   </div>
 
+  <div class="field">
+    <%= form.label :file, "Attach a receipt" %>
```
sampoder:
```suggestion
    <%= form.label :file, "Attach a receipt / invoice" %>
```

Thoughts @garyhtou? @hackliv would invoices be fine?

garyhtou: that should do it!

garyhtou (approved):
It might be a good idea to render these receipts (`ach_transfer.receipts.transfer_create_page`) on the page where admins process ACHs/Checks

- `/admin/Q1h57n/ach_start_approval`
- `/admin/167/increase_check_process`

Adding the `transfer_create_page` will only show docs uploaded on creation; excluding documents (e.g. receipts) added later. i think this _might_ be what we want.

---

This is great to merge as is! If the admin stuff isn't handled in this PR, please create an issue for it

PR #4870, opened 2024-01-02
Make the styling consistent for different filter types in the filter menu & keep the button in the enlarged state

sampoder:
## Summary of the problem

Sarthak flagged this issue: https://github.com/hackclub/hcb/issues/4693. In general, it'd also just be good to make them consistent. Making them consistent didn't fix it but adding a class to the button so we can keep it larger does fix it.

## Describe your changes

Now there doesn't seem to be a shift: https://www.figma.com/file/1jRP9wWbjLGCe72dQdimou/weird-bug?type=design&node-id=0%3A1&mode=design&t=F13gLoprU2PZyd4f-1

Closes https://github.com/hackclub/hcb/issues/4693


sarthaktexas:
this pr does not fix it

sarthaktexas:
this in fact does not fix it -- it's not a size issue, it's the way the filter types are display-ed in the flexbox div. you can't use flexrow with implicit width divs, they need an explicit size or a different display setting.
Screenshot 2024-01-03 at 8 54 40 AM

sampoder:
can you share a screen recording of it shifting? i'm no longer getting any shift: https://www.figma.com/file/fnDn5gusdFA0UrfexonsSS/Untitled?type=design&node-id=0%3A1&mode=design&t=2YknjMD3NeLRM2Fj-1

sampoder:
they're also the exact same width:

Screenshot 2024-01-03 at 8 22 44 AM

Screenshot 2024-01-03 at 8 22 57 AM

app/assets/stylesheets/components/_buttons.scss
```diff
@@ -25,7 +25,8 @@ input[type='submit'],
   user-select: none;
 
   &:hover,
-  &:focus {
+  &:focus,
+  &.open {
```
cjdenio:
```suggestion
  &[aria-expanded=true]{
```

app/javascript/controllers/menu_controller.js
```diff
@@ -85,6 +86,7 @@ export default class extends Controller {
       this.content && this.content.remove()
     }
 
+    this.toggleTarget.classList.remove("open")
```
cjdenio:
```suggestion
```

app/javascript/controllers/menu_controller.js
```diff
@@ -36,6 +36,7 @@ export default class extends Controller {
   }
 
   open() {
+    this.toggleTarget.classList.add("open")
```
cjdenio:
```suggestion
```

cjdenio (approved):

PR #4873, opened 2024-01-02
Add HQ staff to shame dashboard

maxwofford:
## Summary of the problem

This adds HQ staff to the dashboard of missing receipts! Now that we're using this so often, maybe we should think about making this a user-facing feature

Screenshot 2024-01-02 at 12 12 42


garyhtou (approved):
shame!! 🤡

PR #4883, opened 2024-01-03
Refactor User and Event Feature Previews into it's own controller

sampoder:
## Summary of the problem

Closes https://github.com/hackclub/hcb/issues/4726

## Describe your changes

I combined the two different implementations into one implementation that can handle both types of actors. I've also had them both use a shared partial.


garyhtou:
@sampoder let's review this during our next call!

PR #4884, opened 2024-01-03
Add `copy_to_clipboard` helper and use on GSuite config page

sarthaktexas:
## Summary of the problem

long table values such as the dkim key were hard to copy

## Describe your changes

now they be copied to clipboard with a click. adds the clipboard utility + popover-n to table elements

image


sarthaktexas:
This has the clipboard helper on only a few `<td>` fields because that's what was requested by orgs. I am working on a separate refactor of this table in a different (local) branch. This is the fastest implementation I can do.

app/views/events/g_suite_overview.html.erb
```diff
@@ -128,7 +129,7 @@
                   <td>3600</td>
                   <td>MX</td>
                   <td>1</td>
-                  <td>ASPMX.L.GOOGLE.COM.</td>
+                  <td><%= copy_to_clipboard("ASPMX.L.GOOGLE.COM.", tooltip_direction: "w", if: false) do %>ASPMX.L.GOOGLE.COM.<% end %></td>
```
garyhtou: Are the `if` intentionally hardcoded to `false`?

app/helpers/application_helper.rb
```diff
@@ -347,4 +347,12 @@ def redacted_amount
...
   end
 
+  def copy_to_clipboard(clipboard_value, tooltip_direction: "n", **options, &block)
+    return yield if options.delete(:if) == false
+
+    tag.span "data-controller": "clipboard", "data-clipboard-text-value": clipboard_value do
+      tag.span class: "pointer tooltipped tooltipped--#{tooltip_direction}", "aria-label": "Click to copy", "data-action": "click->clipboard#copy", **options, &block
+    end
+  end
+
```
garyhtou: Let's improve this so that it's optional to pass in a block. Calling `copy_to_clipboard` without a block assumes that you just want to output the `clipboard_value`.

```
<%= copy_to_clipboard("ASPMX.L.GOOGLE.COM.") %>
```

I think this might do it??? (not too sure)

```suggestion
  def copy_to_clipboard(clipboard_value, tooltip_direction: "n", **options, &block)
    block ||= -> { clipboard_value }
    return yield if options.delete(:if) == false

    tag.span "data-controller": "clipboard", "data-clipboard-text-value": clipboard_value do
      tag.span class: "pointer tooltipped tooltipped--#{tooltip_direction}", "aria-label": "Click to copy", "data-action": "click->clipboard#copy", **options, &block
    end
  end

```

garyhtou (approved):
👍 so that we can merge this and @malted can use it. Can you address the feedback in another PR?

garyhtou:
In another PR, we'll also want to switch over existing usage to this new helper

app/helpers/application_helper.rb
garyhtou: [`1843a59` (#4884)](https://github.com/hackclub/hcb/pull/4884/commits/1843a596cbdcae77ac31097f135a4826653ae43d)

app/views/events/g_suite_overview.html.erb
garyhtou: [`1b50ac2` (#4884)](https://github.com/hackclub/hcb/pull/4884/commits/1b50ac2b8badb605bd20b94da08a559647c05185)

garyhtou:
_("google workspace" was too long to fit in the commit message)_

Issue #3426, opened 2023-01-14
Associate Event and HcbCode


garyhtou:
Not sure if @albertchae was already planning to do this, but it'd be very handy to be able to do:
```ruby
Event.find(1).hcb_codes
```

and vice versa. Currently `HcbCode.find(1).events` is implemented, but it isn't an association

---

Here's what I have so far, but it seems very messy

## Event -> HcbCode

```ruby
# event.rb

has_many :hcb_codes, ->(event) {
  unscope(where: :event_id)
    .left_joins(:canonical_transactions, :canonical_pending_transactions)
    .left_joins(canonical_transactions: :canonical_event_mapping, canonical_pending_transactions: :canonical_pending_event_mapping)
    .where('canonical_event_mappings.event_id = ? OR canonical_pending_event_mappings.event_id = ?', event.id, event.id)
}


# based on https://stackoverflow.com/questions/36749740/activerecord-has-many-through-through-multiple-models
```

## HcbCode -> Event

```ruby
# hcb_code.rb

has_many :events, ->(hcb_code) {
  unscope(where: :hcb_code_id)
    .left_joins(:canonical_transactions, :canonical_pending_transactions)
    .where("canonical_transactions.hcb_code = ? OR canonical_pending_transactions.hcb_code = ?", hcb_code.hcb_code, hcb_code.hcb_code)
}
```
This is incorrect since it seems like HcbCode relies on more than just CanonicalTransaction and CanonicalPendingTransaction to get the event.
https://github.com/hackclub/bank/blob/94a91292ec187a16d0f125c55f1e837cdf549144/app/models/hcb_code.rb#L117-L148

https://heroku-app97991095.airbrake.io/projects/288439/groups/3449473709812445527?tab=overview

albertchae:
Hey @garyhtou , this is definitely on my list of associations to untangle, but it's a pretty gnarly one. I have some ideas on how to approach it and I saw you added some telemetry to see if `try`ing on some of the associations makes sense. Do you want to pair on it sometime?

garyhtou:
Yes! @albertchae, I'd love to. let's find a time sometime next week or later

albertchae:
@garyhtou cool I'll ping you on slack to figure out a time

albertchae:
🥲 
<img width="370" height="50" alt="Image" src="https://github.com/user-attachments/assets/f287953b-8f8f-4adb-a923-275f91352e40" />

PR #4889, opened 2024-01-04
Improve performance for My Inbox page

garyhtou:
## Summary of the problem

The My Inbox page is very very slow. There are soooo many SQL queries being ran. This number only increases the more transactions missing receipts you have.

## Describe your changes

I've attempted to fix a portion of these performance issues by:
1. Preloading where possible
2. Performing computations in the database (instead of Ruby)
3. Decreasing the default number of transactions `per` page (pagination) from 20 to 15

The largest bottleneck for performance issues on this page is `HcbCode#event`. That method is NOT a proper association and thus can not be preloaded 😔. We have an issue open for this: https://github.com/hackclub/hcb/issues/3426

There are no visual changes besides decreasing the number of transactions `per` page (pagination)

### Results

I have been testing locally using Zach's account since he's the user who will experience the worst performance hits due to the number of missing receipts.

It appears the number of SQL queries have been halved 📉! yay!

**Before** (`main` branch)
image

**After** (this branch)
image

---

Also, I've required a higher confidence level for showing suggestions: https://github.com/hackclub/hcb/commit/b5ba0e278621b80eef47f2d01f34274d6775ec23


sampoder:
thank you! this would be incredible

PR #4898, opened 2024-01-04
Set queue priority for jobs

sampoder:
## Summary of the problem

Flagged by Gary in #4417.

## Describe your changes

I've gone through each job and checked the priority. I've also removed them from `schedule.yml`.


app/jobs/canonical_pending_transaction_job/send_twilio_declined_message.rb
```diff
@@ -2,7 +2,7 @@
 
 module CanonicalPendingTransactionJob
   class SendTwilioDeclinedMessage < ApplicationJob
-    queue_as :critical
```
garyhtou: is there a reason for switching from `critical` to `default`?

i think there's a high priority for these to go out ASAP; especially if someone's swiping a card in person and is wondering why it declined

sampoder: I'm cool with it either way, I just thought this was on equal standing with a mailer

app/jobs/canonical_pending_transaction_job/send_twilio_declined_message.rb
```diff
@@ -2,7 +2,7 @@
...
+    queue_as :default
```
garyhtou:
```suggestion
    queue_as :critical
```

app/jobs/canonical_pending_transaction_job/send_twilio_receipt_message.rb
```diff
@@ -2,7 +2,7 @@
 
 module CanonicalPendingTransactionJob
   class SendTwilioReceiptMessage < ApplicationJob
-    queue_as :critical
+    queue_as :default
```
garyhtou:
```suggestion
    queue_as :critical
```

garyhtou (approved):
great! ty for this Sam!

PR #4901, opened 2024-01-04
Fix the cards page in transparency mode

sampoder:
## Summary of the problem

We were referencing the birthday of the current user.

## Describe your changes

Only render form if organiser signed in.


cjdenio (approved):

albertchae:
Nice thanks for fixing this. FYI @KiKoS0 this was introduced by https://github.com/hackclub/hcb/pull/4471 because this partial was being loaded for logged out users before Sam's fix

PR #4910, opened 2024-01-04
Don't report user-induced errors to AppSignal

northeastprince:
## The Problem
Errors are getting reported that aren't actual issues on our end, like `RecordNotFound` or `InvalidCrossOriginRequest` errors.
## The Solution
Just don't report them in the first place!


garyhtou (approved):

config/appsignal.yml
```diff
@@ -3,6 +3,18 @@ default: &defaults
...
+    - ActionDispatch::Http::MimeNegotiation::InvalidType
+    - ActionDispatch::Http::Parameters::ParseError
+    - ActionController::BadRequest
+    - ActionController::UnknownHttpMethod
+    - ActionController::MethodNotAllowed
+    - ActionController::UnknownFormat
+    - ActionController::ParameterMissing
+    - ActionController::InvalidCrossOriginRequest
+    - ActionController::InvalidAuthenticityToken
+    - ActiveRecord::RecordNotFound
```
garyhtou: We may want to report `RecordNotFound` in certain cases. Such as `RecordNotFound` in jobs. What are your thoughts?

garyhtou (changes requested):

config/appsignal.yml
northeastprince: It's pretty standard to not report `RecordNotFound` errors, so if it means something went wrong if it occurs in a job, I think a different/more-specific exception should be raised.

garyhtou: Let's report `ActiveRecord::RecordNotFound`. It can be indicative of a bug on our end

garyhtou:
```suggestion
```

garyhtou (changes requested):

config/appsignal.yml
northeastprince: Ok,  but then we'll have to change the notification settings as they pop up for controllers so we don't get notified for them after every deployment.

garyhtou (approved):

garyhtou:
>Ok, but then we'll have to change the notification settings as they pop up for controllers so we don't get notified for them after every deployment.

Yes! Let's do that! Can you see if you're able to configure that in AppSignal?

PR #4911, opened 2024-01-04
Add confirmation & tooltip to accidental misuses

sampoder:
## Summary of the problem

Well... it was just a bit confusing.

## Describe your changes

Now it's less confusing because there's a tooltip and we get your confirmation.

This works towards https://github.com/hackclub/hcb/issues/4904.


app/views/hcb_codes/_accidental_misuse.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if local_assigns[:hcb_code].event&.hack_club_hq? %>
   <div class="flex muted justify-end items-end">
     <%= inline_icon "sad", size: 20 %>
-    <%= link_to invoice_as_personal_transaction_hcb_code_path(id: local_assigns[:hcb_code].hashid), method: :post do %>
+    <%= link_to invoice_as_personal_transaction_hcb_code_path(id: local_assigns[:hcb_code].hashid), method: :post, class: "tooltipped tooltipped--w", 'aria-label': "Invoice #{hcb_code.stripe_cardholder&.user&.name} for this transaction.", data: { confirm: "Are you sure you'd like to invoice #{hcb_code.stripe_cardholder&.user&.name} for this transaction? They'll be emailed an invoice." } do %>
```
garyhtou: Maybe tooltip should be on the north (`tooltipped--n`) or south (`tooltipped--s`)

image

garyhtou: Is there a case where `hcb_code.stripe_cardholder` will be `nil`? I'm wondering because I see we're using safe navigation here.

If `hcb_code.stripe_cardholder` can be `nil`, that likely means sending the invoice will fail.

sampoder: Not that I can think of but I did more because I thought it'd be better safe than sorry. I'll update the rest of the code is cleanly handles it.

sampoder: Sure

sampoder:
```suggestion
    <%= link_to invoice_as_personal_transaction_hcb_code_path(id: local_assigns[:hcb_code].hashid), method: :post, class: "tooltipped tooltipped--s", 'aria-label': "Invoice #{hcb_code.stripe_cardholder&.user&.name} for this transaction.", data: { confirm: "Are you sure you'd like to invoice #{hcb_code.stripe_cardholder&.user&.name} for this transaction? They'll be emailed an invoice." } do %>
```

sampoder: Alright, I just pushed something that has it revert to the current_user if the stripe cardholder is undefined (and it tells you that you'll be invoicing yourself)

garyhtou (approved):

PR #4916, opened 2024-01-05
Speed up general receipt upload box using Turbo Streams & Direct Uploads

sampoder:
## Summary of the problem

This page was slow. For every single receipt you uploaded using the big general receipt upload box, it would have to reload the page.

## Describe your changes

Ian & I switched the process over to Turbo Streams, check it out:

https://github.com/hackclub/hcb/assets/39828164/ae6fde03-d5ab-4f41-b95f-657c23624c71

Here's what we've made async:

1. Deleting receipts from the receipt bin
2. Uploading receipts to receipt bin
3. Uploading receipts to specific transactions
4. Add a receipt to a transaction from the bin


sarthaktexas (dismissed):
works for me tank u sam and ian

sampoder:
Want to flag something that I just remembered: we haven't added Turbo Streams for receipt linking from the suggested pairing window. Fun.

sampoder:
Another edge case I want to flag with suggested pairings:

```
<% if pairing.hcb_code.canonical_transactions.any? %>
                      <% pairing.hcb_code.canonical_transactions.each do |ct| %>
                        <%= render partial: "canonical_transactions/canonical_transaction", locals: { ct:, force_display_details: true, hide_tags: true, show_event_name: true } %>
                      <% end %>
                    <% else %>
                      <% pairing.hcb_code.canonical_pending_transactions.each do |pt| %>
                        <%= render partial: "canonical_pending_transactions/canonical_pending_transaction", locals: { pt:, force_display_details: true, hide_tags: true, show_event_name: true } %>
                      <% end %>
                    <% end %>
```
We need to check the case where a transaction HAS a suggested pairing but a person uploads / links a receipt from the `Transactions` list, we need to make sure that this code (in `generate_streams`):

```
    if @receiptable
      if @receiptable.canonical_transactions&.any?
        @receiptable.canonical_transactions.each do |ct|
          streams.append(turbo_stream.replace(
                           ct.local_hcb_code.hashid,
                           partial: "canonical_transactions/canonical_transaction",
                           locals: { ct:, force_display_details: true, receipt_upload_button: true, show_event_name: true, updated_via_turbo_stream: true }
                         ))
        end
      else
        @receiptable.canonical_pending_transactions&.each do |pt|
          streams.append(turbo_stream.replace(
                           pt.local_hcb_code.hashid,
                           partial: "canonical_pending_transactions/canonical_pending_transactio",
                           locals: { pt:, force_display_details: true, receipt_upload_button: true, show_event_name: true, updated_via_turbo_stream: true }
                         ))
        end
      end
    end
```
Is replacing the transactions in the Transactions list and not in the suggested pairing window. At the moment, it replaces the transaction in the suggested pairing window.

sampoder:
To fix both of these issues, we need to write some Turbo stream code that once a receipt is linked / uploaded through whatever method (incl. the suggested pairing window) we remove any suggested pairs from the carousel. 

A couple things to think about: how do we keep the users current position in the carousel? Say they've clicked through 3 pairings and then accept the 4th, we don't want to then show them the first 3 again.

Lot's of interesting things to consider... _sigh_.

app/models/user.rb
```diff
@@ -224,6 +224,24 @@ def active_mailbox_address
...
+      # Don't suggest receipts ignored more than twice
+      ineligible_receipt_ids = SuggestedPairing.ignored.group("receipt_id").having("COUNT(*) >= 2").pluck(:receipt_id)
+      pairings = SuggestedPairing
+                 .unreviewed
+                 .where(receipt_id: @receipts.ids - ineligible_receipt_ids)
+                 .where("distance <= ?", 1500) # With at least a certain confidence level
+                 # Only get the closest pairing for each receipt
+                 .order(:receipt_id, distance: :asc)
+                 .select("DISTINCT ON (receipt_id) suggested_pairings.*")
+                 .select { |pairing| pairing.hcb_code.needs_receipt? }
```
sampoder: added this; do we want to be showing suggested pairings for transactions that already have a receipt? i thought we wouldn't. this will be a lot faster after #4888.

app/models/user.rb
```diff
@@ -224,6 +224,24 @@ def active_mailbox_address
     self.mailbox_addresses.activated.first
   end
 
+  def suggested_receipt_pairings
```
garyhtou: here's a refactor:

```ruby
# app/models/receipt_bin.rb
class ReceiptBin
  def initialize(user)
    @user = user
  end

  def suggested_receipt_pairings
    # blah blah blah code here
  end
end
```

```ruby
# app/models/user.rb
def receipt_bin
  ReceiptBin.new(self)
end
```

**Usage example:**
```ruby
User.first.receipt_bin.suggested_receipt_pairings
```

---

This concepts somewhat falls under the idea of using POROs (Plain Old Ruby Object)

config/routes.rb
```diff
@@ -47,6 +47,7 @@
     get "settings/security", to: "users#edit_security"
     get "settings/admin", to: "users#edit_admin"
     get "inbox", to: "static_pages#my_inbox", as: :my_inbox
+    get "suggested_pairings", to: "static_pages#suggested_pairings", as: :suggested_pairings
```
garyhtou: let's make the path: `/receipt_bin/suggested_pairings`

You can do this using a route `scope`

db/schema.rb
garyhtou, line 1: unrelated

app/models/user.rb
garyhtou: ty

db/schema.rb
sampoder: oops!

app/models/receipt.rb
```diff
@@ -49,7 +49,9 @@ class Receipt < ApplicationRecord
 
   after_create_commit do
     # Queue async job to extract text from newly upload receipt
+    # and to suggest pairings
     ReceiptJob::ExtractTextualContent.perform_later(self)
+    ReceiptJob::SuggestPairings.perform_later(self)
```
garyhtou: pseudo code:
```suggestion
    if async_upload
      ReceiptJob::SuggestPairings.perform_later(self)
    else
      ReceiptJob::SuggestPairings.perform_now(self)
    end
```

app/views/receipts/_form_v3.html.erb
```diff
@@ -1,113 +1,136 @@
...
+      </div>
+      <% if defined?(enable_linking) && defined?(receiptable) && defined?(inline_linking) %>
+        <section class="card pt0 mt2" id="link_receipt_<%= instance %>_select" style="display: none;">
+          <header class="pb2" style="z-index: 1000;<% if local_assigns[:popover] %> padding-top: 1rem;<% end %>">
+            <%= pop_icon_to "view-close", nil, class: "modal__close muted m1", tabindex: 0, onclick: "event.preventDefault(); $('#link_receipt_#{instance.to_json}_select').hide(); $('#link_receipt_#{instance.to_json}_form').show();" %>
+            <h2 class="h1 mt0 mb0 pb0 border-none">Select receipt</h2>
+          </header>
+          <%= async_frame_to link_modal_receipts_path(receiptable_type: receiptable.class, receiptable_id: receiptable.id, popover: local_assigns[:popover]), as: :span do %>
+            <img style="display: none;" src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNkYAAAAAYAAjCB0C8AAAAASUVORK5CYII=" onload="let e = this.parentElement; $.get($(e).data('src'), data => $(e).replaceWith(data));">  <!-- This inline JS is needed to load an async frame that enters the DOM after the initial page load. -->
```
garyhtou: this is creative lol

sarthaktexas: this is so funny

app/models/user/receipt_bin.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class User
+  class ReceiptBin
+    def initialize(user)
+      @user = user
+    end
+
+    def suggested_receipt_pairings
```
cjdenio: Can this be a concern on `User`? I think `user.suggested_receipt_pairings` is simpler than `User::ReceiptBin.new(user).suggested_receipt_pairings`

cjdenio (changes requested):
https://github.com/hackclub/hcb/assets/34525547/8bafd4c5-934c-4acf-b80d-047678331d06

I can't get this to work at all

app/models/user/receipt_bin.rb
garyhtou: I believe user has a `receipt_bin` method. So usage looks like:

```ruby
user = User.find(1)
user.receipt_bin.suggested_receipt_pairings
```

It helps decouple receipt bin from user

garyhtou:
🪄 I just chatted with Caleb. Go ahead and test this really thoroughly and then merge it!

PR #4918, opened 2024-01-05
Org navigation: Fix icon for Transfers page

sarthaktexas:
## Summary of the problem

duplicate icons for transfer and transactions

## Describe your changes

changes transfer icon to payment-transfer icon

image


garyhtou (approved):
good catch!

PR #4921, opened 2024-01-05
Adds a copy email button

sarthaktexas:
## Summary of the problem

duh.

## Describe your changes

adds the button.

image


app/views/events/_organization_position_invite.html.erb
```diff
@@ -5,6 +5,11 @@
     <p class="flex-auto">
       <strong class="h3 block"><%= position.user.name || "Unnamed" %></strong>
       <%= organizer_signed_in? ? position.user.email : "Sign in to view email" %>
+      <% admin_tools("m0 p0 badge mb2 block w-fit", "span") do %>
+      <span class="m0 badge bg-muted pointer" onClick="navigator.clipboard.writeText(<%= position.user.email %>);">
```
cjdenio: this is [XSS](https://en.wikipedia.org/wiki/Cross-site_scripting) vulnerable, here's an implementation that isn't:

```suggestion
      <span class="m0 badge bg-muted pointer" data-controller="clipboard" data-clipboard-text-value="<%= position.user.email %>" data-action="click->clipboard#copy">
```

cjdenio (approved):

PR #4948, opened 2024-01-08
CardGrant: Make transaction list full width

sampoder:
## Summary of the problem

It looked a little weird. #4939.

## Describe your changes

Screenshot 2024-01-08 at 1 20 37 AM


garyhtou (approved):
thanks for redacting haha

PR #4949, opened 2024-01-08
Make canceled card grants less confusing visually

sampoder:
## Summary of the problem

https://github.com/hackclub/hcb/issues/4936

## Describe your changes

If the user has agreed to the terms:

Screenshot 2024-01-08 at 2 06 35 AM

If they haven't:

Screenshot 2024-01-08 at 2 07 04 AM

And on the ledger:

Screenshot 2024-01-08 at 2 06 56 AM


app/models/card_grant.rb
```diff
@@ -105,14 +105,20 @@ def pending_invite?
 
   def cancel!(canceled_by)
     if balance > 0
-      DisbursementService::Create.new(
+      custom_memo = "Cancelled: grant to #{user.name}"
```
garyhtou: can we make this copy edit for more clarity?

```ruby
"Return of funds from cancellation of grant to #{user.name}"
```

PR #4951, opened 2024-01-08
Process disbursements on Column

cjdenio:
Closes #4755


cjdenio:
Going to merge this as it doesn't affect the stability of the web app. I'll be awake at (_checks notes_) 7pm to verify this is working in production.

garyhtou:
yay yeah!

PR #4952, opened 2024-01-08
Add an admin-only audit log for events

cjdenio:
## Summary of the problem

We track a bunch of data via paper_trail, but it's not accessible anywhere in the UI.

## Describe your changes

audit log!

Screenshot 2024-01-08 at 3 07 49 PM


cjdenio:
Merging as a bit of an unfinished admin experiment. This is loaded asynchronously so it won't affect page load time.

PR #4957, opened 2024-01-08
Format JavaScript files with Prettier

cjdenio:
Installs [Prettier](https://prettier.io) and adds it to our CI. The only config change is enabling `trailingComma`, which makes Git diffs nicer when adding/removing items from a multi-line array


garyhtou:
Should we update `bin/lint`?

cjdenio:
Done!

PR #4961, opened 2024-01-09
Display all receipts on ACH/Check process page

cjdenio:
## Summary of the problem

When an admin goes to approve an ACH, they can only see the receipt if it was uploaded via the ACH form (not attached to the transaction later).


garyhtou:
I was [thinking](https://github.com/hackclub/hcb/pull/4869#pullrequestreview-1807336615) we might want that scope. although I don't feel strongly for one or the other.

PR #4978, opened 2024-01-09
Saved ACH recipients (`PaymentRecipient`)

cjdenio:
## Summary of feature
- When an ACH transfer is sent, HCB either creates an `AchRecipient`, or links against an existing `AchRecipient` with name provided on the ACH form.
- HCB using the recipient's name for lookup ensures that:
  - ACHs _aren't_ accidentally linked if the sender selects an existing recipient via the dropdown, then changes the name to someone else.
  - ACHs _are_ still linked if the sender somehow doesn't use the provided autofill
- ACH recipients are automatically updated if the sender changes the autofilled bank details.
- ACH recipients can be removed. Removing them doesn't affect past ACHs sent to that recipient.

https://github.com/hackclub/hcb/assets/34525547/71b42319-8748-4f18-82e8-47ddb9c4f7c2


app/models/ach_recipient.rb
```diff
@@ -0,0 +1,29 @@
...
+
+# == Schema Information
+#
+# Table name: ach_recipients
+#
+#  id                        :bigint           not null, primary key
+#  account_number_ciphertext :text
+#  bank_name                 :string
+#  name                      :string
+#  routing_number            :string
```
garyhtou: Even though routing numbers are public information, I'd like to suggest that we encrypt it. The only model that doesn't have it encrypted is `AchTransfer` (https://github.com/hackclub/hcb/issues/4981)

image

cjdenio: Hmm. Should `bank_name` be encrypted too? Many banks have just one routing number that can easily be looked up

garyhtou: yeah probably

garyhtou:
Would you like to handle https://github.com/hackclub/hcb/issues/5003 as a part of this PR? Or, do so separately?

cjdenio:
I think it would be good to handle them separately— it'll be easy to add an email field to this feature once #5003 is implemented.

app/models/ach_transfer.rb
```diff
@@ -226,4 +233,17 @@ def scheduled_on_must_be_in_the_future
     end
   end
 
+  def create_or_link_ach_recipient
+    self.ach_recipient = event.ach_recipients.find_by("name ILIKE ?", AchTransfer.sanitize_sql_like(recipient_name)) ||
+                         event.ach_recipients.build(name: recipient_name)
+
+    self.ach_recipient.name = self.recipient_name
```
garyhtou: Let's make AchRecipients immutable. If a user edits an AchRecipient, it will delete the old one and create a new one

app/models/ach_recipient.rb
```diff
@@ -0,0 +1,39 @@
...
+# Table name: ach_recipients
+#
+#  id                        :bigint           not null, primary key
+#  account_number_ciphertext :text
+#  bank_name_ciphertext      :string
+#  name                      :string
+#  routing_number_ciphertext :string
+#  created_at                :datetime         not null
+#  updated_at                :datetime         not null
+#  event_id                  :bigint           not null
```
garyhtou: On our call, we chatted about possibly making this polymorphic and/or associating to User. It would future-proof this model, however, it's not important at this time.

app/models/ach_recipient.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  belongs_to :event
+  has_many :ach_transfers
+
+  has_encrypted :account_number, :routing_number, :bank_name
+
+  validates :name, uniqueness: { scope: :event_id }
+
+  scope :order_by_last_used, -> { includes(:ach_transfers).order("ach_transfers.created_at DESC") }
+
```
garyhtou: validate the model is ready-only

garyhtou:
Caleb's also planning to add an `email` field in a separate PR. (related to https://github.com/hackclub/hcb/pull/5051)

cjdenio:
Merging after a call with Gary!

PR #4985, opened 2024-01-10
Changed some transactions to remove Bank

aramshiva:
Changed some transactions from "HACK CLUB BANK" to "HCB".

This is part of #4965.


aramshiva:
I am not familiar with some things like the `canonical_transaction.rb` file, so can someone make sure I didn't mess something up?

garyhtou (commented):
Unfortunately, all of these need to stay as "Hack Club Bank" because it is matching data stored in the production database. If we were to update these to "HCB", we won't be able to search for the old transactions.

cjdenio (changes requested):
Unfortunately all of these occurrences need to stay the same— many legacy parts of the transaction engine rely on the phrase "Hack Club Bank" for routing money to the correct organization.

However, a few of these instances (`HACK CLUB BANK DONATE` and `HACK CLUB BANK PAYOUT`) are no longer used and could be removed entirely in the future.

cjdenio:
aww gary beat me to it

garyhtou:
hahah

aramshiva:
kk, mb!

garyhtou:
> kk, mb!

All good! I'm very glad that you're making an effort for this! The transaction engine's one of the tougher one of update/work on.

I did a quick search of the codebase, it looks like we've got all the easily change able occurrences of "Hack Club Bank" and "Bank". The rest are a part of https://github.com/hackclub/hcb/issues/4963 which is a _very very very_ heavy lift.

PR #4996, opened 2024-01-10
Move Invoice States Over To AASM

### LINKED ISSUE #4992: Invoice: Fix state management

Here's what we need to do.

sampoder:
Open question: is the archived stuff the cleanest way of doing this?


sampoder:
Well based on RSpec, `mark_deposited!` is in the wrong place. Shucks. @garyhtou or @cjdenio any idea where it should be?

cjdenio:
`mark_deposited` should be called once the payout arrives in our bank account. For donations, this is implemented in `DonationJob::Nightly`. For invoices, it might look like:
```ruby
Invoice.in_transit.each do |invoice|
  if invoice.local_hcb_code.canonical_transactions.count == 2 # payout + fee reimbursement
	invoice.mark_deposited!
  end
end
```

app/models/invoice.rb
```diff
@@ -135,14 +136,40 @@ class NoAssociatedStripeCharge < StandardError; end
   aasm do
     state :open_v2, initial: true
```
garyhtou, line 138: at some point, it could be nice to get rid of the `_v2`

sampoder: Agreed, though I'd like to get this working in production before dropping the `_v2`

PR #5006, opened 2024-01-10
Admin Pending Ledger: Link to HCB code instead of Event

sampoder:
## Summary of the problem

Flagged by @sarthaktexas: https://github.com/hackclub/hcb/issues/4917. Wanted to get a quick fix.

## Describe your changes

Links to the HCB code instead.


app/views/admin/pending_ledger.html.erb
```diff
@@ -38,7 +38,7 @@
         <td><%= cpt.id %></td>
         <td><%= cpt.date %></td>
         <% if cpt.event %>
-          <td><%= link_to cpt.memo, event_path(cpt.event) %></td>
+          <td><%= link_to cpt.memo, hcb_code_path(cpt.local_hcb_code) %></td>
```
garyhtou: I _think_ `hcb_code_path` might not be necessary since Rails knows `local_hcb_code` is of type/class `HcbCode`. So, by default it'll link to it's show page
```suggestion
          <td><%= link_to cpt.memo, cpt.local_hcb_code %></td>
```

sampoder: cool

garyhtou (approved):

PR #5011, opened 2024-01-11
Improve Usage Breakdowns

sampoder:
## Summary of the problem

Usage breakdowns were slow. Very slow.

## Describe your changes

now they are fast! because we don't load hcb codes when we don't need to.

ALSO. I moved the button. It was too big.


app/views/hcb_codes/breakdown.pdf.erb
```diff
@@ -85,7 +85,11 @@
             <% @spent_on.each do |transaction| %>
               <tr>
                 <td class="bold" colspan="2">
-                  <%= link_to transaction[:memo].strip, transaction[:url] %>
+                  <% if transaction[:url] %>
```
garyhtou: are there transactions without a `url`?

app/models/hcb_code.rb
```diff
@@ -396,4 +396,44 @@ def not_admin_only_comments_count
...
+    available = 0
+    if self.canonical_transactions.any? { |ct| ct.amount_cents.positive? }
+
+      income = ::EventService::PairIncomeWithSpending.new(event: self.event).run
+
+      # PairIncomeWithSpending is done on a per CanonicalTransaction basis
+      # This compute it for this specific HcbCode
+
+      self.canonical_transactions.each do |ct|
+        if (ct[:amount_cents] > 0) && income[ct[:id].to_s]
```
garyhtou: Should this be `ct.amount_cents`?

PR #5022, opened 2024-01-12
Remove unnecessary search for user_session if there is no session_token

albertchae:
Return early with a nil `user_session` if `session_token` is blank

This triggers a sequential scan on `user_sessions` which could
potentially get expensive.

```
bank-hackclub::DATABASE=> explain select * from user_sessions where deleted_at is null and session_token_bidx is null limit 1;
                                QUERY PLAN
--------------------------------------------------------------------------
 Limit  (cost=0.00..0.80 rows=1 width=271)
   ->  Seq Scan on user_sessions  (cost=0.00..945.95 rows=1188 width=271)
         Filter: ((deleted_at IS NULL) AND (session_token_bidx IS NULL))
(3 rows)
```


cjdenio (approved):
Thanks for this!

PR #5025, opened 2024-01-12
improve ledger exports!

sampoder:
## Summary of the problem

https://github.com/hackclub/hcb/issues/4594

## Describe your changes

Now we include data about the merchant but we limit to only non-public exports.


sampoder:
(oh also we're using a new gem... the old one was wack!)

app/controllers/exports_controller.rb
```diff
@@ -142,7 +142,7 @@ def transactions_json
   end
 
   def transactions_ledger
-    ::ExportService::Ledger.new(event_id: @event.id).run
+    ::ExportService::Ledger.new(event_id: @event.id, public_only: current_user ? false : true).run
```
sampoder: do a better check

PR #5026, opened 2024-01-12
Add tests for classes in app/services/hcb_code_service

KiKoS0:
## Summary of the problem
Improves test coverage of the app/services/hcb_code_service/ classes.
Related to: https://github.com/hackclub/hcb/issues/4491

## Describe your changes
Add tests for the following classes:
- `HcbCodeService::Generate::ShortCode`
- `HcbCodeService::Comment::Create`
- `HcbCodeService::FindOrCreate`
- `HcbCodeService::CanDispute`
- `HcbCodeService::AiGenerateMemo`
- `HcbCodeService::SuggestedMemos`


spec/services/hcb_code_service/generate/short_code.rb
```diff
@@ -0,0 +1,27 @@
...
+      end
+    end
+  end
+
+  context "when short_code already exists" do
+    it "regenerates a new one" do
+      srand(42)
+      old_hcb_code = create(:hcb_code)
+
+      srand(42)
```
albertchae: nice use of `srand`. Might be worth leaving a comment about how this affects the RNG?

spec/services/hcb_code_service/comment/create_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+
+      expect(comment).to be_a(Comment)
+      expect(comment.persisted?).to eq(true)
+      expect(comment.commentable).to eq(hcb_code)
+      expect(comment.user).to eq(user)
+      expect(comment.content).to eq("Example comment")
+      expect(comment.admin_only).to eq(true)
+    end
+  end
```
albertchae: follow up: let's add a case for when `hcb_code_id` doesn't exist

spec/services/hcb_code_service/find_or_create_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  context "when hcb_code doesn't exist" do
+    let(:hcb_code) { "HCB-401-1" }
+
+    it "creates a new one and returns it" do
+      hcb_code_record = described_class.new(hcb_code:).run
+
+      expect(hcb_code_record).to be_a(HcbCode)
+      expect(hcb_code_record.persisted?).to eq(true)
+    end
```
albertchae: should we also check that `hcb_code_record.code` is eq to `hcb_code` (the string)?

spec/factories/hcb_code_factory.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :hcb_code do
+    sequence(:hcb_code) { |n| "HCB-#{::TransactionGroupingEngine::Calculate::HcbCode::INVOICE_CODE}-#{n}" }
```
albertchae: nice, we could probably consider setting up traits for the different types of `INVOICE_CODE`, `DONATION_CODE` etc. But this is fine for now

spec/services/hcb_code_service/can_dispute_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+
+describe HcbCodeService::CanDispute do
+
+  context "when it's an invalid transaction" do
+    let(:hcb_code) { build_stubbed(:hcb_code) }
+
+    it "returns false" do
+      result, = described_class.new(hcb_code:).run
+
+      expect(result).to eq(false)
```
albertchae: I think we should check the error message as well, even if it makes the test more brittle because we're looking for exact error text, it's more valuable to make sure we are getting false for the expected reason

spec/services/hcb_code_service/can_dispute_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+    end
+  end
+
+
+  context "when stripe transaction is older than 90 days" do
+    let(:hcb_code) { build_stubbed(:hcb_code) }
+    it "returns false" do
+      allow(hcb_code).to receive(:no_transactions?).and_return(false)
+      allow(hcb_code).to receive(:stripe_card?).and_return(true)
+      allow(hcb_code).to receive(:date).and_return(Date.today - 91.days)
```
albertchae: it would be more ideal if we set up the test data so that these checks pass rather than stubbing, but this is fine for now. I think this would be a valuable follow up to learn more about the hcb_code data model though

app/services/hcb_code_service/ai_generate_memo.rb
```diff
@@ -2,21 +2,21 @@
 
 module HcbCodeService
   class AiGenerateMemo
-    def initialize(hcb_code:)
+    def initialize(hcb_code:, conn: nil)
```
albertchae: nice job making `conn` an injectable dependency

spec/services/hcb_code_service/suggested_memos_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+describe HcbCodeService::SuggestedMemos do
+
+  context "when a bank fee transaction is passed" do
+    let(:event) { create(:event) }
+
+    it "returns all memo suggestions for similar bank fee transactions" do
+      3.times { create_bank_fee_transaction }
+
+      new_transaction = create_bank_fee_transaction(custom_memo: "HACK CLUB BANK FEE")
```
albertchae: should you refer to `OUTGOING_BANK_FEE_MEMO` here instead of the literal string?

albertchae: oh I see, you create 3 of them that have random numbers at the end and then check if a memo with a similar prefix gets those suggested 🤔 I think it still might make sense to use the constant here

spec/services/hcb_code_service/suggested_memos_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+OUTGOING_BANK_FEE_MEMO = ::TransactionEngine::SyntaxSugarService::Shared::OUTGOING_BANK_FEE_MEMO
+
+describe HcbCodeService::SuggestedMemos do
```
albertchae: this is a really interesting class that could probably have a ton of different tests, but this is a good foundation. Someone (not necessarily us) can add more later

albertchae (commented):
Great work. I think my biggest comment is regarding whether we could set up the data for `HcbCodeService::CanDispute` instead of mocking the checks, but that can be a follow up to this PR.

spec/factories/hcb_code_factory.rb
KiKoS0: yeah that would be nice i think.

spec/services/hcb_code_service/suggested_memos_spec.rb
KiKoS0: i 💯 agree, it's not intentional.

spec/services/hcb_code_service/suggested_memos_spec.rb
KiKoS0: Exactly, yeah that test already covers most of it i think. And given that i'm not knowledgeable about its exact use cases yet, i might end up with suboptimal and/or coupled tests.

spec/services/hcb_code_service/comment/create_spec.rb
KiKoS0: ~i think that would raise, it doesn't help right?~
oh because it's usage can rescue, good call 👍 
https://github.com/hackclub/hcb/blob/b50c69bd8975132c0cac19d69e6aa0a7c550d2a8/app/controllers/hcb_codes_controller.rb#L117

spec/services/hcb_code_service/can_dispute_spec.rb
albertchae: we discussed this in a call but this ends up pulling on a lot of dependent objects, so we'll tackle it in a follow up by identifying models on the "edge" of the graph and adding test coverage before working towards `hcb_code`

albertchae (approved):
looks good, thanks for addressing feedback

PR #5028, opened 2024-01-13
Allow faking donations in staging and development

polypixeldev:
## Summary of the problem
In staging and development, it's difficult to test code that deals with donations or money in an event. By allowing donations to be faked in staging and development, testing code that deals with donations is easy, and as much money as you want can be added to an event using donations.

Note that Stripe does not let you use real card data in testing - fake card data must be used, such as that described on their docs at https://stripe.com/docs/testing#testing-interactively.

Closes #4838 

## Describe your changes
In testing mode, Stripe will not validate card details or actually transfer any money, but it will respond with status `"requires_payment_method"`. This PR marks donations that have this status when not in prod as in transit, and upon finishing the donation adds the donation to the ledger.


app/controllers/donations_controller.rb
```diff
@@ -113,6 +113,13 @@ def finish_donation
   def finished
     @donation = Donation.find_by!(url_hash: params[:donation])
     @event = @donation.event
+
+    # Add the donation to the ledger automatically when not in prod, since the stripe webhook will never run
+    if !Rails.env.production?
```
cjdenio: Could this somehow be opt-in? It's often useful to test the entire end-to-end payment flow while in development

cjdenio:
```suggestion
    unless Rails.env.in?(%w(production test))
```

This should be disabled in `test` mode in addition to `production`

sampoder:
https://github.com/hackclub/hcb/issues/4838#issuecomment-2336511347

PR #5034, opened 2024-01-14
Make Styling Changes To Suggested Pairings Window

sampoder:
## Summary of the problem

Things were a bit weird on my account! https://github.com/hackclub/hcb/issues/4886

## Describe your changes

Going to send videos in Slack (`#hcb-engr-notifs`) (they have private info, sorry). Basically I just made things a bit stricter on the flexbox widths so you got a more consistent structure.


sampoder:
Something I'd like feedback on... the arrows. Idk what to do with them? They kind of look awkward wherever I put them. I brought them inside the card because that felt like the best option.

The videos are slowly uploading in this thread: https://hackclub.slack.com/archives/C04LDFJ5X7S/p1705199548069049

app/views/static_pages/_suggested_pairings.html.erb
garyhtou, line 1: I love that you broke this up into its own partial! Is this currently being used? The diff seems to suggest the old code is still there

sampoder: Oh these are changes for a different PR... let me revert that!

garyhtou (approved):

PR #5038, opened 2024-01-15
Use Turbo Frames on the Org/Event Edit Page

sampoder:
## Summary of the problem

Flagged by @arianna-m: https://github.com/hackclub/hcb/issues/5035.

## Describe your changes

This page.... was just one big form. I've split it up into multiple forms, enabled Turbo on them and then made them each a Turbo frame. This makes more sense to me considering the page's visual structure.

I also had to make a couple of edits in `ui.js` so that the thing that hides the additional transparency mode-related settings works even after the Turbo Frame is refreshed.


app/assets/javascripts/ui.js
```diff
@@ -403,6 +388,31 @@ document.addEventListener("turbo:load", () => {
   }
 })
 
+document.addEventListener("turbo:frame-load", () => {
```
cjdenio: A more reliable option might be an event listener on `document`. `turbo:frame-load` may not trigger / trigger twice under certain circumstances.

```diff
- $(transparencyToggle).on('change', e => {
+ $(document).on("change", "#event_is_public", e => {
```

Reference: https://turbo.hotwired.dev/handbook/building#:~:text=When%20possible%2C%20avoid%20using%20the%20turbo%3Aload%20event%20to%20add%20other%20event%20listeners%20directly%20to%20elements%20on%20the%20page%20body.%20Instead%2C%20consider%20using%20event%20delegation%20to%20register%20event%20listeners%20once%20on%20document%20or%20window.

sampoder: Sounds good, thanks Caleb!

sampoder:
This is going to have all sorts of merge conflicts with the work to switch over the forms to a tabbed system. Closing for that reason.

PR #5043, opened 2024-01-16
Personal Transaction: Fix edge case where amount < $1

sampoder:
## Summary of the problem

See https://hackclub.slack.com/archives/C04LDFJ5X7S/p1705430653245099. 

## Describe your changes

Basically doesn't let people do this on charges of less than a dollar because we can't invoice them (Stripe...).


app/controllers/hcb_codes_controller.rb
```diff
@@ -193,6 +193,11 @@ def invoice_as_personal_transaction
     event = hcb_code.event
 
     authorize hcb_code
+    
+    if hcb_code.amount_cents >= -100
+      flash[:error] = "Invoices can only be generated for charges of $1.00 or more."
+      return redirect_to @hcb_code
```
sampoder:
```suggestion
      return redirect_to hcb_code
```

app/controllers/hcb_codes_controller.rb
```diff
@@ -194,6 +194,11 @@ def invoice_as_personal_transaction
 
     authorize hcb_code
 
+    if hcb_code.amount_cents >= -100
+      flash[:error] = "Invoices can only be generated for charges of $1.00 or more."
+      return redirect_to hcb_code
+    end
+
     personal_tx = HcbCode::PersonalTransaction.create(hcb_code:, reporter: current_user)
```
garyhtou: What happens if we add a validation to `HcbCode::PersonalTransaction` and then call `create!` instead of `create`

app/views/hcb_codes/_accidental_misuse.html.erb
```diff
@@ -1,10 +1,17 @@
+<% @disabled = local_assigns[:hcb_code].amount_cents >= -100 %>
```
garyhtou: We don't want to use instance variables inside partials because they can mess with other partials. We can use normal variables though.

```suggestion
<% disabled = local_assigns[:hcb_code].amount_cents >= -100 %>
```

app/views/hcb_codes/_accidental_misuse.html.erb
garyhtou, line 1: I'm think we can remove the changes to this file and let the flash message communicate that it's not possible

sampoder: Sure!

app/controllers/hcb_codes_controller.rb
sampoder: What would that look like?

garyhtou:
```ruby
# In the model:
module HcbCode
  class PersonalTransaction < ApplicationRecord
    # https://guides.rubyonrails.org/active_record_validations.html
    validates :amount_cents, comparison: { less_than: -100 }
  end
end
```

```ruby
# In the controller
personal_tx = HcbCode::PersonalTransaction.create!(hcb_code:, reporter: current_user)
# Using the create! bang method will raise an error on a failure to save (such as a validation error)
```

garyhtou: I think this should work, but there's only one way to find out!

sampoder: I this would still cause issues because we create the invoice before validation

garyhtou: ah, right 😔

Okay, let's merge as is. In the future, we should look into trying to change that `before_validation` to `before_create`

app/controllers/hcb_codes_controller.rb
```diff
@@ -194,6 +194,11 @@ def invoice_as_personal_transaction
 
     authorize hcb_code
 
+    if hcb_code.amount_cents >= -100
```
```suggestion
    if hcb_code.amount_cents < -100
```

garyhtou: I think this is the opposite of what we want

sampoder: -99 cents transactions aren't supported, i think `>=` is the right way round because it's negative numbers

garyhtou: ahh yeah ur right 😵‍💫 i'm confusing myself now ahhhh

garyhtou (approved):

PR #5049, opened 2024-01-16
fix balances exports!

sampoder:
## Summary of the problem

https://github.com/hackclub/hcb/issues/5046

## Describe your changes

just disabled turbo


sarthaktexas:
this seems like a cheat

sampoder:
well it fixes the actual problem, turbo can't really handle csv/json

garyhtou (approved):

garyhtou:
After merging, test in prod and then notify Bence in the thread! thanks 💖

PR #5050, opened 2024-01-17
Use an `email` param to prefill grantee email when visiting grant from notification email

sampoder:
## Summary of the problem

https://github.com/hackclub/hcb/issues/5048

## Describe your changes

In the email, we make the link have an email param. Then we just use that email param, I felt like this was cleaner than doing signing


app/controllers/card_grants_controller.rb
```diff
@@ -43,7 +43,9 @@ def show
     @card_grant = CardGrant.find_by_hashid!(params[:id])
 
     if !signed_in?
-      return redirect_to auth_users_path(return_to: card_grant_path(@card_grant)), flash: { info: "To continue, please sign in with the email you received the grant." }
+      url_queries = { return_to: card_grant_path(@card_grant) }
+      url_queries[email] = params[:email] if params[:email]
```
garyhtou: please test this! I think there's an error on this line

sampoder: wait wack let me check

sampoder: alright fixed: 

Screenshot 2024-01-16 at 8 43 11 PM

don't know where that colon went

garyhtou (approved):
great! ty

PR #5061, opened 2024-01-18
Add staging credentials and service id for twilio verifications (used for login)

albertchae:
SMS part of https://github.com/hackclub/hcb/issues/4927

I created a twilio subaccount and added the Verify service under it. From there
I had to add a new service id for staging as well as api keys in credentials under twilio
for `account_sid_sms_verify` and `auth_token_sms_verify`. There are 3 other twilio keys but they aren't needed for login.

I think we could have created another instance of the Verify service under the main account as well, but this should be fine for additional separation between prod and staging.


app/services/twilio_verification_service.rb
```diff
@@ -9,7 +9,7 @@ class TwilioVerificationService
   )
 
   # This isn't private/sensitive so it's okay to keep here
-  VERIFY_SERVICE_ID = "VAa06a66dad4c1ca3c199a46334ff11945"
+  VERIFY_SERVICE_ID = Rails.env.production? ? "VAa06a66dad4c1ca3c199a46334ff11945" : "VAe30d49e92f634419aacdc8648948dc75"
```
garyhtou: Even though this isn't private/sensitive, I wonder if it should be moved into credentials so that we don't need the condition.

garyhtou (approved):
Great! good to merge as is

app/services/twilio_verification_service.rb
albertchae: will merge this PR and make this change in a follow up

PR #5064, opened 2024-01-19
Fix birthdate nameerror

manuthecoder:
Fix undefined behavior in #5063


sampoder (approved):
This fixed the issue, though I wanted to check with @farhat-ha if there isn't a larger issue here

albertchae:
> This fixed the issue, though I wanted to check with @farhat-ha if there isn't a larger issue here

Hm I think as Caleb said in https://github.com/hackclub/hcb/issues/5063#issuecomment-1899565287, this shouldn't be happening if you are up to date with latest `main` and have run all the migrations

PR #5068, opened 2024-01-20
Add a lost receipt form dialog that users must fill out when marking a transaction as having no/lost receipt.

rluodev:


cjdenio:
Closing for now as it doesn't seem like this is being worked on

PR #5069, opened 2024-01-20
Organizer Deletion Request: Calculate red flags when viewing

polypixeldev:
## Summary of the problem

Previously, the number of missing receipts from a user on an event and the number of active cards from the user on an event were reported by the user filing the deletion request, which can cause inaccuracies and things can slip through the cracks.

Closes #4915
Closes #4715 

## Describe your changes

This PR calculates these two numbers on-the-fly by adding methods to the organizer position deletion request model, which are then used by the view to display the counts. This PR also changes how red flags on the OPDR page are shown, to display the specific number as well as emojis that should be clearer to reviewers. It also removes the deletion request form options for missing receipts and active cards.

image

image


YodaLightsabr (approved):
Looks good! @garyhtou what are your thoughts on the `subject_has_outstanding_transactions_stripe` and `subject_has_active_cards` columns? We haven't remove them yet because we were wondering if we should keep them to preserve the old data.

app/models/organizer_position_deletion_request.rb
```diff
@@ -82,4 +82,20 @@ def open
...
   end
 
+  def organizer_missing_receipts
+    organizer_position.user.transactions_missing_receipt.select do |hcb_code|
+      hcb_code.event == organizer_position.event
+    end
+  end
+
+  def organizer_active_cards
+    organizer_stripe_cards.where(stripe_status: "active")
```
garyhtou:
```suggestion
    organizer_stripe_cards.active
```

garyhtou (approved):

PR #5093, opened 2024-01-23
Add test for `AchTransfersController`

cjdenio:


cjdenio:
@albertchae I'm trying to figure out the best way to mock authentication in a way that can be re-used across tests— I figured you might have thoughts about this

albertchae:
> @albertchae I'm trying to figure out the best way to mock authentication in a way that can be re-used across tests— I figured you might have thoughts about this

@cjdenio good question, I haven't done any controller tests yet in hcb so haven't run into this yet. For examples of popular authentication libraries
- devise: there's a `sign_in` helper you can call before each test https://www.rubydoc.info/github/plataformatec/devise/Devise%2FControllers%2FSignInOut:sign_in
- clearance: also includes `sign_in` helpers for tests https://github.com/thoughtbot/clearance#controller-test-helpers

So I think we should think of doing something similar, maybe it'd call the code in https://github.com/hackclub/hcb/blob/main/app/helpers/sessions_helper.rb and then set the cookie for future requests, or set `current_user` somehow so that future requests see a current_user and think the user is signed in.

I can tinker with this later tonight and see if I figure out something straightforward

albertchae:
@cjdenio here's a proof of concept that gets the 2 tests in here to pass https://github.com/hackclub/hcb/pull/5098. Maybe we can generalize it into a test helper and then refactor it to be less janky later

PR #5096, opened 2024-01-23
Admin: Organization summary export

sampoder:
## Summary of the problem

See #5066.

## Describe your changes

Added an additional export to `/admin/balances`

I would like feedback on addresses (we don't seem to be using this field anymore, do we even store addresses?) and the month by month revenue because I couldn't figure out a clean way to do that.


app/controllers/admin_controller.rb
```diff
@@ -985,16 +986,34 @@ def balances
...
+               else
+                 [
+                   [:organization_id, ->(e) { e.id }],
+                   [:organization_name, ->(e) { e.name }],
+                   [:net_balance, ->(e) { render_balance.call(e, :settled_balance_cents) }],
+                   [:expenses, ->(e) { render_balance.call(e, :settled_outgoing_balance_cents) }],
+                   [:revenue, ->(e) { render_balance.call(e, :settled_incoming_balance_cents) }],
+                   [:category, ->(e) { e.category }],
+                   [:address, ->(e) { e.owner_address }], # are we using this? where are addresses stored?
+                   [:tags, ->(e) { e.event_tags.pluck(:name) }],
```
```suggestion
                   [:tags, ->(e) { e.event_tags.pluck(:name).join(",") }],
```

app/controllers/admin_controller.rb
```diff
@@ -975,6 +975,7 @@ def bookkeeping
   def balances
     @start_date = params[:start_date].present? ? Date.parse(params[:start_date]) : nil
     @end_date = params[:end_date].present? ? Date.parse(params[:end_date]) : nil
+    @partnered_events = params[:partnered_events] || false
```
garyhtou: Remove scope. We want to export for all

garyhtou:
We'll also need to remove that CSV file from github

sampoder:
NEED TO ADD: by-month raised. use spread operator. use `settled_incoming_balance_cents`. use `min(event.created_at)`.

app/models/event.rb
```diff
@@ -458,6 +458,15 @@ def settled_incoming_balance_cents(start_date: nil, end_date: nil)
         ct.sum(:amount_cents)
       end
   end
+  
+  def monthly_revenue(month)
```
garyhtou: This seems fairly similar to

https://github.com/hackclub/bank/blob/48ebaaddfcf9b060a0a238bce5ac5a64276bd59b/app/models/event.rb#L451

Maybe we can re-use that

app/controllers/admin_controller.rb
```diff
@@ -975,6 +975,7 @@ def bookkeeping
   def balances
     @start_date = params[:start_date].present? ? Date.parse(params[:start_date]) : nil
     @end_date = params[:end_date].present? ? Date.parse(params[:end_date]) : nil
+    @partnership_info = params[:partnership_info] || false
```
```suggestion
    @monthly_breakdown = params[:monthly_breakdown] || false
```

app/controllers/admin_controller.rb
```diff
@@ -985,16 +986,57 @@ def balances
...
       [:organization_name, ->(e) { e.name }],
       [:net_balance, ->(e) { render_balance.call(e, :settled_balance_cents) }],
       [:expenses, ->(e) { render_balance.call(e, :settled_outgoing_balance_cents) }],
       [:revenue, ->(e) { render_balance.call(e, :settled_incoming_balance_cents) }],
-      [:start_date, ->(_) { @start_date }],
-      [:end_date, ->(_) { @end_date }]
     ]
+
+
+    if @partnership_info
```
```suggestion
    if @monthly_breakdown
```

app/controllers/admin_controller.rb
```diff
@@ -985,16 +986,57 @@ def balances
     render_balance = ->(event, type) {
       ApplicationController.helpers.render_money(event.send(type, start_date: @start_date, end_date: @end_date))
     }
+
+    render_monthly_revenue = ->(event, year, month) {
+      ApplicationController.helpers.render_money(
+        @monthly_revenue[event.id] ? @monthly_revenue[event.id][Date.new(year, month, 1).strftime("%Y-%m")] : 0
+      )
```
```suggestion
      key = Date.new(year, month, 1).strftime("%Y-%m")
      ApplicationController.helpers.render_money(
        @monthly_revenue.dig(event.id, key) || 0
      )
```

app/controllers/admin_controller.rb
```diff
@@ -985,16 +986,57 @@ def balances
...
-      [:start_date, ->(_) { @start_date }],
-      [:end_date, ->(_) { @end_date }]
     ]
+
+
+    if @partnership_info
+      template.concat(
+        [
+          [:category, ->(e) { e.category }],
+          [:address, ->(e) { e.owner_address }], # are we using this? where are addresses stored?
```
```suggestion
```

app/controllers/admin_controller.rb
```diff
@@ -985,16 +986,57 @@ def balances
...
+          [:tags, ->(e) { e.event_tags.pluck(:name).join(",") }],
+          [:joined, ->(e) { e.activated_at || e.created_at }],
+          [:start_date, ->(_) { @start_date }],
+          [:end_date, ->(_) { @end_date }]
```
garyhtou: move this to all

app/controllers/admin_controller.rb
```diff
@@ -985,16 +986,57 @@ def balances
...
+        ]
+      )
+      @monthly_revenue ||=
```
garyhtou: replace `||=` with `=`

app/controllers/admin_controller.rb
```diff
@@ -985,16 +986,57 @@ def balances
...
+                                          .where("canonical_transactions.amount_cents > ?", 0)
+                                          .group("canonical_event_mappings.event_id, date_trunc('month', canonical_transactions.created_at)")
+                                          .select("date_trunc('month', canonical_transactions.created_at) AS month,
+                               COALESCE(SUM(canonical_transactions.amount_cents), 0) AS revenue,
+                               canonical_event_mappings.event_id AS event_id")
+                                          .to_sql
+          ActiveRecord::Base.connection.exec_query(sql_query).each_with_object({}) do |item, result|
+            result[item["event_id"]] = {
+              item["month"].strftime("%Y-%m") => item["revenue"].to_i
+            }
```
```suggestion
            result[item["event_id"]] ||= {}
            result[item["event_id"]][item["month"]] = item["revenue"].to_i
```

garyhtou (commented):
Preload event tag association

On line 984
```diff
-@events = filtered_events
+@events = filtered_events.includes(:event_tags)
```

app/views/admin/balances.html.erb
```diff
@@ -25,6 +25,7 @@
         <span class='mr1'>EXPORT</span>
         <%= button_tag("CSV", value: "csv", name: "format", disable_with: false, formtarget: "_blank") %>
         <%= button_tag("JSON", value: "json", name: "format", disable_with: false, formtarget: "_blank") %>
+        <%= button_tag("CSV (partners)", value: "true", name: "partnership_info", type: "submit", formaction: "/admin/balances.csv", disable_with: false, formtarget: "_blank") %>
```
sampoder:
```suggestion
        <%= button_tag("CSV (partners)", value: "true", name: "monthly_report", type: "submit", formaction: "/admin/balances.csv", disable_with: false, formtarget: "_blank") %>
```

garyhtou (approved):

PR #5097, opened 2024-01-24
Make Emburse Cards Read Only

aramshiva:
Fixes #5012. Makes Emburse Cards read-only and removes functions like edit and cancel.


garyhtou (approved):
Great work here! This is good to merge.


We'll need to come back in the future to remove the API integration

aramshiva:
Yay!

YodaLightsabr (approved):

PR #5098, opened 2024-01-24
Proof of concept of signing in for tests

albertchae:
## Summary of the problem

We don't have a way to sign in for tests and hcb has a handrolled authentication system, so we can't use an off the shelf helper

https://github.com/hackclub/hcb/pull/5093#issuecomment-1907176834

## Describe your changes

Fake signing in by having the controller find a `current_user` in `sessions_helper`


spec/controllers/ach_transfers_controller_spec.rb
```diff
@@ -14,6 +14,12 @@
   end
 
   describe "validate_routing_number" do
+    before do
+      # janky way of signing in for tests, we basically override the call in sessions_helper that checks for current_user to return this test user
+      user = create(:user)
+      allow_any_instance_of(described_class).to receive(:current_user).and_return(user)
```
albertchae: Generally this way of stubbing is discouraged https://github.com/rubocop/rspec-style-guide#any_instance_of, but it does work for us 🤷

cjdenio:
```suggestion
      allow(controller).to receive(:current_user).and_return(user)
```
`controller` lets you access the underlying controller instance:
https://rspec.info/features/6-0/rspec-rails/controller-specs/controller-spec/#:~:text=A%20controller%20is%20exposed%20to%20global%20before%20hooks

albertchae: !! wow TIL. in that case I might even prefer something like this

`controller.instance_variable_set(:@current_user, user)`

since I like that it calls the real "current_user" method

PR #5105, opened 2024-01-25
Stripe Activation now does stuff

aramshiva:
Fixes #5090. Activating Cards now does anything.


app/controllers/stripe_cards_controller.rb
```diff
@@ -106,7 +106,7 @@ def create
     end
 
     return redirect_back fallback_location: event_cards_new_path(event), flash: { error: "Birthday is required" } if current_user.birthday.nil?
-    return redirect_back fallback_location: event_cards_new_path(event), flash: { error: "Event is in Playground Mode" } if event.demo_mode?
+    return redirect_back fallback_location: event_cards_new_path(event), flash: { error: "Org is in Playground Mode" } if event.demo_mode?
```
sampoder:
```suggestion
    return redirect_back fallback_location: event_cards_new_path(event), flash: { error: "Organization is in Playground Mode" } if event.demo_mode?
```

or ~~Organisation~~

config/routes.rb
```diff
@@ -282,11 +282,16 @@
...
-    post "update_name"
-    post "freeze"
-    post "defrost"
-    post "activate"
+    member do
+      get "edit"
+      post "update_name"
+      post "freeze"
+      post "defrost"
+    end
```
cjdenio: Heads up that `member` blocks change the format of path helpers (e.g. `stripe_card_freeze_path` becomes `freeze_stripe_card_path`)
Screenshot 2024-01-24 at 8 24 31 PM

aramshiva: Resolved!

YodaLightsabr:
- [x] - Default cards to inactive
- [x] - Build out activation page to enter last4
- [x] - Remove activate button from card show page and add information about shipping
- [x] - Add activate method to take that response and activate the cards

aramshiva:
Does anyone know where the card is being defaulted to activated when creating a card? I wasn't able to find it.

sampoder:
Please test this but I think it’s here: https://github.com/hackclub/hcb/blob/6dd0d6b78c1578a956c99b54d3599eb071377e82/app/services/stripe_card_service/create.rb#L72

aramshiva:
I wasnt able to find it for some reason, but it was right infront of me, I forgot to check card services! Thanks!

aramshiva:
http://hcb.hackclub.com/stripe_cards/activation
^ change `hack.af/activate` on merge to this

app/mailers/stripe_card_mailer.rb
```diff
@@ -8,12 +8,12 @@ def physical_card_ordered
...
 
     mail to: @recipient,
-         subject: "Your new HCB card (ending in #{@card.last4}) for #{@event.name} is on its way"
+         subject: "Your new HCB card for #{@event.name} is on its way"
   end
 
   def lost_in_shipping
     mail to: @recipient,
-         subject: "Your HCB card (ending in #{@card.last4}) for #{@event.name} was lost in shipping"
+         subject: "Your HCB card (ending in #{@card.last4}) for #{@event.name} was lost in shipping."
```
sampoder:
```suggestion
         subject: "Your HCB card for #{@event.name} was lost in shipping."
```

Giving them the last four probably isn't that helpful if we're hiding it at all other times

config/routes.rb
```diff
@@ -282,11 +282,17 @@
...
+
+    collection do
+      get "activation"
```
sampoder: Could we make the path `/stripe_cards/activate` and then differentiate between GET / POST requests?

I bit like how we do with `/auth`:

```
get "auth", to: "users#auth"
post "auth", to: "users#auth_submit"
```

It's a small thing but I feel like that'd be cleaner?

app/controllers/stripe_cards_controller.rb
```diff
@@ -146,6 +162,10 @@ def update_name
     redirect_to stripe_card_url(card), flash: updated ? { success: "Card's name has been successfully updated!" } : { error: "Card's name could not be updated" }
   end
 
+  def activation
+    authorize StripeCard
+  end
+
```
sampoder: We can skip doing Pundit all together if we're just making the policy `true` :D 

```
skip_after_action :verify_authorized, only: [:activation]
```

This can go at the top it will do that for you

app/mailers/stripe_card_mailer.rb
aramshiva: I talked to Ian about this and he thought it might be useful with HCB support iirc

app/controllers/stripe_cards_controller.rb
aramshiva: oh, this is just a placeholder for now

app/controllers/stripe_cards_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class StripeCardsController < ApplicationController
+  skip_after_action :verify_authorized, only: [:activation]
```
garyhtou: since we're using `skip_authorization`, this line likely is not needed

app/views/stripe_cards/shipping.html.erb
```diff
@@ -1,55 +1 @@
...
-        </footer>
-      <% end %>
-    </div>
-    <%= image_tag @map_url, style: "width: 100%; max-width: 512px; height: auto; object-fit: cover;", width: 256, height: 128 if @map_url %>
-  </div>
-  <% @map_url = nil %>
-<% end %>
-</section>
-<% end %>
+<%= render partial: "stripe_cards/shipping" %>
```
garyhtou: thank you!!

config/routes.rb
garyhtou: i like this idea!

I'd also suggest to make the `POST` a **member** path.

`GET`: /stripe_cards/activate
`POST`: /stripe_cards/:last4/active (/stripe_cards/1234/activate)

app/controllers/stripe_cards_controller.rb
sampoder: `activation` also doesn't seem to exist

app/controllers/stripe_cards_controller.rb
```diff
@@ -48,25 +50,58 @@ def defrost
...
+
+    unless @card.present?
+      flash[:error] = "Card not found"
+      skip_authorization
+      redirect_back fallback_location: activate_stripe_cards_path and return
+    end
+    
     authorize @card
 
-    # Does this card replace another card? If so, attempt to cancel the old card
```
sampoder: Can we keep this code comment? feels helpful

garyhtou (commented):
this is good to merge after making the form look more like a card, and testing it!

aramshiva:
### To-Dos

Before Merge
- [x] Upgrade UI to include realistic card
- [x] Handle current cards already in transit
- [ ] Change `hack.af`

After Merge
- [ ] Write a changelog post
- [ ] Ship in #ship and #hcb
- [x] Message @sarthaktexas for Twitter/X post.

app/controllers/stripe_cards/activation_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      if @card.nil?
+        flash[:error] = "Card not found"
+        skip_authorization
+        redirect_back fallback_location: new_stripe_cards_activation_path and return
+      end
+
+      authorize @card, :activate?
+
+      if @card.activated?
+        if @card.created_at < Date.new(2024, 2, 22)
```
garyhtou:
```suggestion
        # Handle cards issued before activation feature launch
        if @card.created_at < Date.new(2024, 2, 22)
```

garyhtou (approved):
yay!

PR #5107, opened 2024-01-25
Letter Opener is now available in staging.

aramshiva:
Fixes #5106. Letter Opener is now accessible in staging, before if trying to access a staging builds, you werent able to login due to admin@example.com being the only user and Letter Opener can't be used.


aramshiva:
Now to wait for @garyhtou's auto staging deployer to see if it actually works 😅

aramshiva:
Closing due to @maxwofford's comment. Mistake on my end

> Are we still pulling a full db backup into staging? If so, this should probably wait until we've switched to non-production data
https://github.com/hackclub/hcb/issues/5106#issuecomment-1909290081

PR #5120, opened 2024-01-26
Delete Tasks Page

devramsean0:
## Summary of the problem
This removes the Admin tasks page. This page is seemingly unreachable and is also duplicated on the homepage anyway when signed in as admin. See #5067 

## Describe your changes
Removes the method on the function and deletes the relevant ERB file


cjdenio (commented):
also remove this line!
https://github.com/hackclub/hcb/blob/2e317ad287d83fe70c2ce5ec84f35769915af2d1/config/routes.rb#L526

cjdenio (approved):

sarthaktexas (approved):

PR #5055, opened 2024-01-17
Switch Airtable from token to personal access token

maxwofford:
Fixes https://github.com/hackclub/hcb/issues/5042

This also reformats the rails encrypted credentials to the following shape:
```yaml
airtable:
    pat: SECRET_VALUE
    base: SECRET_VALUE
    key: DEPRECATED_WARNING
```

PR #5127, opened 2024-01-28
Replace Airbridge with Airtable API using PAT

malted:
cc @maxwofford 

Airtable are revoking all legacy API keys in 2 days; and upon investigation I realised Airbridge is only used in one place.

Blocking on @maxwofford's #5055 to be able to use `Rails.application.credentials.airtable[:pat]`.

Closes https://github.com/hackclub/hcb/issues/5042


malted:
Oh wait I should use the Airtable client

garyhtou:
Yes! We've got https://github.com/sirupsen/airrecord installed; it's a Ruby client for Airtable

garyhtou:
> Oh wait I should use the Airtable client

During my 1:1 with Malted, we decided to skip this for now. We'll continue using Faraday. If anyone like to take on switching Faraday to Airrecord, feel free to do so!

garyhtou:
@maxwofford please update your PAT to allow access for:
- Onboard Verifications: `app4Bs8Tjwvk5qcD4`
- Bank Applications: `apppALh5FEOKkhjLR`

The admin cards for these are erroring due to insufficient permissions.

garyhtou (approved):
Code is good! We're just blocked by https://github.com/hackclub/hcb/pull/5127#issuecomment-1918438148

Let's merge by no later than Feb 1st, even if we don't have the PAT access updated

PR #5128, opened 2024-01-29
Remove stubbing from can_dispute tests

KiKoS0:
Followup to this comment: https://github.com/hackclub/hcb/pull/5026#discussion_r1460683201

Turns out it's possible to test it if we just ignore the generated `hcb_code` generated by `after_create` in `canonical_transactions` and updating it with the under test `hcb_code`.


spec/services/hcb_code_service/can_dispute_spec.rb
```diff
@@ -2,10 +2,13 @@
 
 require "rails_helper"
 
+HCBCode = TransactionGroupingEngine::Calculate::HcbCode
```
albertchae: this is a shortcut for module scoping, right?

spec/services/hcb_code_service/can_dispute_spec.rb
```diff
@@ -16,11 +19,10 @@
   end
 
   context "when it's valid transaction but not a stripe transaction" do
-    let(:hcb_code) { build_stubbed(:hcb_code) }
+    let(:hcb_code) { build(:hcb_code) }
+    let!(:ct) { create(:canonical_transaction).update(hcb_code: hcb_code.hcb_code) }
```
albertchae: nit, but does `create(:canonical_transaction, hcb_code: hcb_code.hcb_code)` work?

spec/services/hcb_code_service/can_dispute_spec.rb
```diff
@@ -30,12 +32,10 @@
 
 
   context "when stripe transaction is older than 90 days" do
-    let(:hcb_code) { build_stubbed(:hcb_code) }
-    it "returns false" do
-      allow(hcb_code).to receive(:no_transactions?).and_return(false)
-      allow(hcb_code).to receive(:stripe_card?).and_return(true)
-      allow(hcb_code).to receive(:date).and_return(Date.today - 91.days)
+    let(:hcb_code) { build(:hcb_code, code_type: ::HCBCode::STRIPE_CARD_CODE) }
+    let!(:ct) { create(:canonical_transaction, date: Date.today - 91.days).update(hcb_code: hcb_code.hcb_code) }
```
albertchae: same question here, but I'm guessing there's possibly something weird with the callback on `hcb_code` for create?

albertchae (commented):
Looks good, mainly some questions on setting the hcb_code on the transaction during test setup. I only commented on 2 but I see the others all follow that pattern.

spec/services/hcb_code_service/can_dispute_spec.rb
KiKoS0: it should work but it gets overwritten by:
https://github.com/hackclub/hcb/blob/544982275efe7a82451ff152dd722af79a207047/app/models/canonical_transaction.rb#L101
https://github.com/hackclub/hcb/blob/544982275efe7a82451ff152dd722af79a207047/app/models/canonical_transaction.rb#L353-L361

albertchae: ok that's what I thought. this is fine for now then 🚀

albertchae (approved):

PR #5134, opened 2024-01-29
Add tests for canonical_transaction services

KiKoS0:
## Summary of the problem
Improves test coverage of the `app/services/canonical_pending_transaction_service` and `app/services/canonical_transaction_service` classes.

Related to: https://github.com/hackclub/hcb/issues/4491

## Describe your changes
Add tests for the following classes:
- `CanonicalPendingTransactionService::Unsettle`
- `CanonicalTransactionService::SetEvent`
- `CanonicalTransactionService::Stats::During`


spec/services/canonical_pending_transaction_service/unsettle_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+           canonical_pending_transaction:,
+           canonical_transaction:)
+
+    ach_transfer = create(:ach_transfer, event:, amount: amount_cents, aasm_state: "deposited")
+    create(:raw_pending_outgoing_ach_transaction, canonical_pending_transaction:, ach_transfer:, amount_cents:)
+
+    expect(ach_transfer.in_transit?).to eq(false)
+    service.run
+    ach_transfer.reload
+    expect(ach_transfer.in_transit?).to eq(true)
```
albertchae: I would check `expect(CanonicalPendingSettledMapping.exists?).to eq(false)` here too

spec/services/canonical_pending_transaction_service/unsettle_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    create(:canonical_pending_settled_mapping,
+           canonical_pending_transaction:,
+           canonical_transaction:)
+
+    ach_transfer = create(:ach_transfer, event:, amount: amount_cents, aasm_state: "deposited")
+    create(:raw_pending_outgoing_ach_transaction, canonical_pending_transaction:, ach_transfer:, amount_cents:)
+
+    expect(ach_transfer.in_transit?).to eq(false)
+    service.run
+    ach_transfer.reload
```
albertchae: these reloads are the worst part of rails unit tests lol

spec/services/canonical_transaction_service/stats/during_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+RSpec.describe CanonicalTransactionService::Stats::During do
+  let(:event_1) { create(:event, omit_stats: false) }
+  let(:event_2) { create(:event, omit_stats: false) }
+  let(:event_3) { create(:event, omit_stats: true) }
+
+  it "returns stats related to events transactions" do
+    # Ommited because event is ommiting stats
+    create(:canonical_transaction, event: event_3, amount_cents: 250)
+    # Ommited because outside of date range
```
albertchae: nit: spelling should be `omitted` and `omitting`

spec/services/canonical_transaction_service/stats/during_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    create(:canonical_transaction, event: event_2, amount_cents: -100)
+
+    start_time = Date.parse("2019-01-01")
+    end_time = Date.today
+
+    result = described_class.new(start_time:, end_time:).run
+
+    expected = { transactions_volume: 3600,
+                 expenses: -100,
+                 raised: 3500,
```
albertchae: Does this include the transactions on line 17, 18 and 21? because based on comments I would have thought this was only 17 and 18 and would total to 2500

spec/services/canonical_transaction_service/set_event_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      expect(CanonicalEventMapping.exists?(id: canonical_event_mapping.id)).to eq(false)
+    end
+
+    it "creates a new mapping" do
+      described_class.new(canonical_transaction_id: canonical_transaction.id,
+                          event_id: event.id,
+                          user:).run
+
+      canonical_transaction.reload
+      expect(canonical_transaction.canonical_event_mapping.event).to eq(event)
```
albertchae: Hm, I guess there are different schools of thought on whether tests like this should be combined or separated (this is related to my other comment too). I guess I'm fine either way for this particular test, so no change necessary.

albertchae (approved):
looks great overall

spec/services/canonical_transaction_service/stats/during_spec.rb
KiKoS0: thank you, well that's really weird that ruby isn't enabled by default in my spell checker extension...

image

spec/services/canonical_transaction_service/set_event_spec.rb
KiKoS0: Right yeah I usually only separate it if the behavior is too complex to test in one single test making it convoluted or if it's a side effect unrelated to the primary behavior. For this case I followed this test https://github.com/hackclub/hcb/blob/f0f7018889995b470d139a03f6355db5043d9806/spec/services/canonical_pending_transaction_service/settle_spec.rb.

spec/services/canonical_pending_transaction_service/unsettle_spec.rb
KiKoS0: i think in general they are usually avoided if the action (service class in this case) returned the updated model but in this case i'm not even sure what the `service.run` function returns.

spec/services/canonical_transaction_service/stats/during_spec.rb
KiKoS0: Yeah it's 17, 18 and 21. A similar confusion is in the tested code as well how do you think we can clarify it better?

https://github.com/hackclub/hcb/blob/890aac3796d107214483c3b8079219fe96bd6e28/app/services/canonical_transaction_service/stats/during.rb#L15-L16

spec/services/canonical_pending_transaction_service/unsettle_spec.rb
albertchae: yeah true, I have seen that this combined with `let` vs `let!` behavior can lead to some really hard to understand tests knowing what's persisted or not and when things get read from the db. Returning the updated object from the service is definitely one way to mitigate some of that

spec/services/canonical_transaction_service/stats/during_spec.rb
albertchae: hahaha that's funny.

PR #5135, opened 2024-01-29
Enables branch code coverage in simplecov

KiKoS0:
## Summary of the problem
Enables branch coverage to show it in the UI along with line coverage, for now the [primary one](https://github.com/simplecov-ruby/simplecov#primary-coverage) is still line coverage.

image


albertchae (approved):
LGTM, we can consider changing primary coverage later

PR #5141, opened 2024-01-30
Add anonymous donations

polypixeldev:
## Summary of the problem

Some donors would like to remain anonymous when donating to a transparent organization.
Closes #5112 

## Describe your changes

This PR adds a checkbox on the start donation page of an event that allows donors to choose to be anonymous. A tooltip explains the feature. If checked, the donation will have the new `anonymous` field on the donation schema set to true. Anonymity is reflected across the ledger (with a custom memo), the donations page, and the transparency API. This checkbox is also behind a feature preview for each event.

image


db/schema.rb
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 # This file is auto-generated from the current state of the database. Instead
```
```suggestion
# frozen_string_literal: true

# This file is auto-generated from the current state of the database. Instead
```

ah yes frozen literals. amazing work on this PR by the way! the graphic is epic. just check a couple other unrelated changes in `db/schema.rb` (I know it's annoying...)

app/views/donations/_donation_form.html.erb
```diff
@@ -59,6 +59,16 @@
       </div>
     </div>
 
+    <% if Flipper.enabled?(:anonymous_donations_2024_01_29, @event)%>
+      <div class="field field--checkbox mb2">
+        <%= form.check_box :anonymous, style: "margin-left: 0;" %>
+        <%= form.label :anonymous, "Donate anonymously?", style: "display: inline; flex-grow: 1;" %>
```
sampoder:
```suggestion
        <%= form.label :anonymous, "Keep my donation private", style: "display: inline; flex-grow: 1;" %>
```
When I hear "anonymous" I assumed that it would mean that organisers couldn't see my name, I think this makes it a bit clearer and is closer to the wording in https://github.com/hackclub/hcb/issues/5112 (just a bit shorter).

Thoughts?

polypixeldev: I agree, "Keep my donation private" is clearer

app/views/events/edit.html.erb
```diff
@@ -88,4 +88,30 @@
...
+    <% elsif Flipper.enabled?(:anonymous_donations_2024_01_29, @event) %>
+      <%# Feature is enabled, let the user opt-out %>
+      <%= link_to "Disable", disable_feature_event_path(feature: :anonymous_donations_2024_01_29), method: :post, class: "btn bg-accent" %>
+    <% else %>
+      <%# Feature is disabled, let the user opt-in %>
+      <%= link_to "Enable", enable_feature_event_path(feature: :anonymous_donations_2024_01_29), method: :post, class: "btn bg-info" %>
+    <% end %>
+  </div>
+  <p>
+    Allow donors to appear anonymously on public pages!
```
garyhtou:
```suggestion
    Allow donors to appear anonymously in Transparency Mode!
```

garyhtou (approved):
Very cool!

YodaLightsabr (approved):

garyhtou:
let's ship & change log this! @polypixeldev please work with @YodaLightsabr on writing a draft for both

PR #5145, opened 2024-01-30
Stripe Issuing Top-Up: Update buffer to $200k

garyhtou:
## Summary of the problem

We need to adjust our Stripe Issuing floating balance to match our expected spending. Stripe requires us to keep a balance with them to approve card authorizations. If our balance is too low, authorizations will automatically be declined.

## Describe your changes

This PR updates our buffer to $200k. 

As of Jan 2024, our maximum Stripe issuing spending in a given week is [$95k (week of Oct 1st, 2023)](https://hcb.hackclub.com/blazer/queries/487-stripe-issuing-spending-per-week). We want to have a buffer of double the expected max weekly spending. The week time-frame is important because it can take [5 business days](https://stripe.com/docs/issuing/funding/balance?locale=en-GB&push-pull-preference=pull#:~:text=take%20up%20to-,5%20business%20days,-to%20show%20up) for Top-ups to become available.

image

With our current spending habits (average $50k/wk) and a buffer of $200k, our average "age" of money will be 1 month and should not dip below 2 weeks. However, considering that it can take up to 5 business days for top-ups to transfer, the age is more like 3 weeks (avg) and 1 week (min).

---

I stole the concept of ["age of money" from YNAB](https://www.ynab.com/blog/what-is-the-ideal-age-of-money)

**Thought:** in the future, we might be able to following to set a dynamic buffer: `max(one_month_moving_average, one_quarter_moving_average) * 4`


garyhtou:
image
image

PR #5168, opened 2024-02-01
Drag-n-drop files into ACHs & Checks

sampoder:
## Summary of the problem

Selecting a file is boring, uncool, and slow. Drag-n-dropping is... the opposite: exciting, cool, and fast.

## Describe your changes

https://github.com/hackclub/hcb/assets/39828164/e3c15c3e-c585-4eb1-820e-647e2d491fdf

(featuring the [Macaque](https://en.wikipedia.org/wiki/Macaque), Sarthak's favourite animal)


sarthaktexas:
An animation about where the file goes would be helpful. See an example by drag and dropping a file into a GitHub PR comment, see how it puts a 2pt dotted gray outline inside the box it's dropping into?

YodaLightsabr:
You might be able to use a full page dropzone like we're using for HCB codes.

sampoder:
Oh I like that idea, let me go find that code

sampoder:
I've added the full page drop

PR #5174, opened 2024-02-02
Organizer Deletion Request: Add warning if user is last signee

polypixeldev:
## Summary of the problem

Orgs should not be left without a contract signee when an organizer is removed. A warning to HCB ops processing deletion requests would help with this.
Closes #4942 

## Describe your changes

This PR adds an additional ⚠️ red flag to the deletion request, as well as an extra final warning in the confirmation dialog if the delete organizer button is clicked.

image
image


sarthaktexas:
Thank you!! This helps tremendously. Can we make the line about last signee visually bolder? Maybe make it red? What if we change the orange text to something else when they are the last signee and make it red instead?

YodaLightsabr (approved):
Nice work!

polypixeldev:
@sarthaktexas Good idea, how about now?

app/models/organizer_position_deletion_request.rb
```diff
@@ -92,6 +92,14 @@ def organizer_active_cards
     organizer_stripe_cards.active
   end
 
+  def organizer_last_signee?
+    event_signee_count == 1 && organizer_position.signee?
+  end
+
+  def event_signee_count
+    organizer_position.event.organizer_positions.where(is_signee: true).size
+  end
```
sampoder: Should this be moved to the event model? `event.rb`

And then it would be `@opdr.event.signee_count`

sampoder (approved):

sarthaktexas (approved):

sarthaktexas:
In the future, make sure to update the top-most comment with the updated screenshots.

YodaLightsabr (approved):

garyhtou (approved):
good to go!

Shaun The Sheep

PR #5185, opened 2024-02-03
Make transactions neutral colored when zero on ledger

polypixeldev:
## Summary of the problem

To be able to distinguish between zero (refunded) transactions and positive transactions on the ledger.
Closes #5142 

## Describe your changes

A `.transaction--zero` class was added, which uses the neutral color, and applied to transactions with a value of zero

image
image


YodaLightsabr (approved):

garyhtou (approved):
lovely!

PR #5186, opened 2024-02-03
Improve Breakdown Exports

sampoder:
## Summary of the problem

Feedback from Zach, see https://github.com/hackclub/hcb/issues/5182

## Describe your changes

Does changes requested


config/initializers/inflections.rb
```diff
@@ -20,4 +20,5 @@
 ActiveSupport::Inflector.inflections(:en) do |inflect|
   inflect.irregular "has", "have"
   inflect.irregular "hasn't", "haven't"
+  inflect.acronym 'ID'
```
sampoder:
```suggestion
```

We can't take this approach, it messes up Rail's system

garyhtou (approved):
This will prevent titleize from converting `ID` to `Id`
```ruby
@headers = header_syms.map { |h| h.is_a?(String) ? h : h.to_s.titleize(keep_id_suffix: true) }
```

test for merging!

PR #5188, opened 2024-02-03
Admin: Move Event balances page's Export button for lower visual importance

aramshiva:
Fixes #5016
New:
CleanShot 2024-02-06 at 17 40 40@2x
Old:
CleanShot 2024-02-02 at 17 43 10@2x

> The Export button on the admin's organization search page looks like the search button.
> Maybe people unintentionally click it when trying to search.
> 
> Solution
> Move it next to search and make the button's color muted.
https://github.com/hackclub/hcb/issues/5016#issue-2077547502


garyhtou:
Can we move it to the right of the search button instead of the search bar?

image

garyhtou:
Please re-request for my review after the edit and after Ian has approved it! thanks 😄

app/views/admin/events.html.erb
```diff
@@ -5,13 +5,10 @@
 <%= paginate @events %>
 
 <%= form_with local: true, url: events_admin_index_path, method: :get do |form| %>
-  <%= form.text_field :q, value: params[:q], style: "vertical-align: top;", placeholder: "Search" %>
+  <%= form.text_field :q, value: params[:q], style: "vertical-align: top;" %>
```
YodaLightsabr:
```suggestion
  <%= form.text_field :q, value: params[:q], style: "vertical-align: top;", placeholder: "Search" %>
```

YodaLightsabr (approved):

garyhtou (commented):
Is the event_filter partial being used elsewhere? if it is, we likely don't wanna show the button elsewhere!

I'm not able to double check right now because i'm on my phone

garyhtou (changes requested):
the partial is being used [here](https://github.com/hackclub/bank/blob/e2d49a66f9544d0be87b49dd8b609efa17971a0a/app/views/admin/balances.html.erb#L8)

we likely don't want to render two types of "export" buttons on the admin event balances page

garyhtou (approved):

PR #5189, opened 2024-02-03
Stripe Card: Add warning about the default billing address

polypixeldev:
## Summary of the problem

Currently, users will sometimes get confused due to the billing address on cards defaulting to HCB's Santa Monica address. This causes transactions to get canceled when they enter the wrong billing address.
Closes #5155

## Describe your changes

This PR adds an info message to the cards detail page which lets the user know that the default billing address is being used. Users will be taken to this page after issuing a virtual card and physical card activation (#5105). 
This PR also adds a warning/info box to the billing address change page that lets the user know of the consequences of changing your billing address.

image

image

image


sampoder:
Does this clog things up too much? Personally, I use the West Hollywood address out of person preference and I wonder if there are more folks like me. Maybe we could make it more subtle?

Like we could potentially have just the `i` icon next to the address and then on hover we could show the full warning text.

app/views/stripe_cards/_details.html.erb
```diff
@@ -87,10 +87,17 @@
...
-    </p>
+      <div class="flex justify-between">
+        <%= render_address stripe_card.cardholder %>
+        <% if @is_default_billing_address %>
+          <p class="info tooltipped tooltipped--w" aria-label="You're using HCB's default billing address, not your home address. You can change it in your user settings if you'd like." style="display: flex;">
+            <%= inline_icon "info", size: 24, style: "position: unset;" %>
+          </p>
+        <% end %>
+      </div>
+    </div>
```
sampoder:
```erb
    <div>
      <strong class="inline-flex items-center" style="gap: 4px">
        Billing address
        <% if @is_default_billing_address %>
          <span class="info tooltipped tooltipped--e" aria-label="You're using HCB's default billing address, not your home address. You can change it in your user settings if you'd like." style="display: flex;">
            <%= inline_icon "info", size: 18, style: "position: unset;" %>
          </span>
        <% end %>
      </strong>
      <%= render_address stripe_card.cardholder %>
    </div>
```
Alternative design:
Screenshot 2024-02-05 at 1 47 45 PM

sampoder (approved):
Looks good to me! I mocked up an alternative design but I think either works honestly.

app/views/users/edit_address.html.erb
```diff
@@ -47,6 +47,21 @@
...
         </div>
 
+        <div class="card border b--warning pb0 mt2 mb2">
+          <div class="flex items-center warning">
+            <%= inline_icon "info", size: 32, class: "mr1" %>
+            <p class="bold mt0 mb0">Important info about changing your billing address</p>
+          </div>
+          <ul class="mt1 mb2">
+            <li>Changing your address applies immedietly to all of your cards.</li>
+            <li>Recurring card charges like subscriptions might break when your billing address changes.</li>
```
garyhtou:
```suggestion
            <li>Recurring card charges like subscriptions may decline if they aren't updated with your new billing address.</li>
            <li>Changing your address immediately applies to all of your cards.</li>
```

app/views/stripe_cards/_details.html.erb
garyhtou: [`fc35deb` (#5189)](https://github.com/hackclub/hcb/pull/5189/commits/fc35deb3f03f1f5c3927c8e5e0373773b22b8c96)

garyhtou (approved):
yay

YodaLightsabr (approved):

PR #5190, opened 2024-02-03
TX Engine: Refactor try_to to safely

manuthecoder:
## Summary of the problem
See #5070


sampoder (dismissed):
This looks good to me & seems to work locally! I'm not too familiar with this part of the codebase though so I'd appreciate a sanity check

garyhtou (commented):
Love it! two things:

```
safely do
  stuff
end
```
is the same as 
```
safely { stuff }
```

I'd prefer the curly brace syntax since it keeps everything on one line.


the other thing is that we need to make sure that `safely` will report errors to both Airbrake and AppSignal. i think we currently don't have that configured

PR #5192, opened 2024-02-03
Cards info panel now aligns card number and hidden dots

rluodev:
now fixed like sarthak showed

See https://github.com/hackclub/hcb/issues/5143 for screenshots


garyhtou (approved):
yay! so satisfying now.

good to merge after approval from @sarthaktexas

cjdenio (changes requested):
Screenshot 2024-02-05 at 5 47 45 PM

It's still not quite aligned— `items-end` pushes the dots towards the bottom of the container. `items-center` should do the trick!

sarthaktexas:
NO @cjdenio do items-center and then reveal the number… the icon causes it to not be centered even when you do that

cjdenio:
I did, and it was... can you share a screenshot of the text being misaligned with `items-center`, then correctly aligned with `items-end`?

sarthaktexas:
ykkk i can't actually recreate it anymore

app/views/stripe_cards/_details.html.erb
```diff
@@ -39,7 +39,7 @@
             <strong>
               Card number
             </strong>
-            <div class="flex justify-between mono">
+            <div class="flex justify-between mono items-end">
```
sarthaktexas:
```suggestion
            <div class="flex justify-between mono items-center">
```

sarthaktexas (changes requested):

cjdenio (approved):

sarthaktexas (approved):

PR #5194, opened 2024-02-04
Admin users list now shows card grants

rluodev:
title

(will add pics in like a day)


app/views/admin/users.html.erb
```diff
@@ -55,6 +56,20 @@
             <% end %>
           <% end %>
         </td>
+        <td>
+          <% if user.card_grants.size > 0 %>
```
garyhtou: add `includes(:card_grants)` to `@users` to prevent N+1s

garyhtou (changes requested):

app/views/admin/users.html.erb
rluodev: wait what does this mean

garyhtou: here's a bit of info! https://medium.com/doctolib/how-to-find-fix-and-prevent-n-1-queries-on-rails-6b30d9cfbbaf

Inside the controller, wherever we define the @users variable, we want to add `includes(:card_grants)` to the chain.

For example:
```ruby
@users = User.all.includes(:card_grants)
```

app/controllers/admin_controller.rb
```diff
@@ -289,7 +289,7 @@ def users
 
       relation = @event.users.includes(:events)
     else
-      relation = User.includes(:events)
+      relation = User.includes(:events).includes(:card_grants)
     end
```
garyhtou: I'd suggest writing it as like because both branches of the condition needs the includes for events and card grants
```ruby
      relation = @event.users
    else
      relation = User.all
    end
    relation = relation.includes(:events).includes(:card_grants)
```

garyhtou: [`89e8a80` (#5194)](https://github.com/hackclub/hcb/pull/5194/commits/89e8a80cc523cb98f40f2f74a823e088a83592d7)

garyhtou (approved):
great!

sarthaktexas (approved):

PR #5196, opened 2024-02-04
Email commands: now we have more!

sampoder:
## Summary of the problem

Email commands are cool but we only have one. Now we have `@missing` and `@tag` as well. 

We also had a couple of bugs, as shown by this HCB code: https://hcb.hackclub.com/hcb/GkHWr4g.

Closes https://github.com/hackclub/hcb/issues/5336

## Describe your changes

Screenshot 2024-02-04 at 11 45 25 AM

We also handle the bug where it would upload the email with the commands in it as a file. And where we'd just rename the transaction to "to rename the transaction."


garyhtou (commented):
Let's hold off on adding `@missing` for now. this is because we will soon require users to fill out a form in order to mark a transaction as missing a receipt. That will require them to be on HCB (since it can't/shouldn't be done via email)


Also, for adding tags:
1. We should not allow the creation of tags
2. We should fuzzy search for the closest tag

garyhtou:
Also, i think tags are currently feature flagged. if they are, we should check if it's enabled for that org

sampoder:
Coolio! @garyhtou: just removed `@missing` and tweaked `@tag`. What's the best way to do fuzzy search on the database? I'm just using `ILIKE` rn.

garyhtou:
Postgres full-text search!
https://pganalyze.com/blog/full-text-search-ruby-rails-postgres

sampoder:
Oh that’s sick! Just implemented

app/mailboxes/hcb_code_mailbox.rb
```diff
@@ -70,7 +70,7 @@ def process_command(command)
       return unless command["argument"] && Flipper.enabled?(:transaction_tags_2022_07_29, @hcb_code.event)
 
       event = @hcb_code.event
-      tag = event.tags.find_by("label ILIKE ?", Tag.sanitize_sql_like(command["argument"]))
+      tag = event.tags.search_label(Tag.sanitize_sql_like(command["argument"])).first
```
garyhtou: I have a feeling we might not need to `Tag.sanitize_sql_like`

app/mailboxes/hcb_code_mailbox.rb
```diff
@@ -0,0 +1,123 @@
...
+    end
+
+    return bounce_error if result&.empty? && @commands.none?
+
+    HcbCodeMailer.with(
+      mail: inbound_email,
+      reply_to: @hcb_code.receipt_upload_email,
+      renamed_to: @renamed_to,
+      tagged_with: @tagged_with,
+      missing: @missing,
```
sampoder:
```suggestion
```

app/views/hcb_code_mailer/bounce_success.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  <% end %>
+</p>
+
+<p>
+  <% if @tagged_with.any? %>
+    <%= @receipts_count > 0 && @renamed_to ? "Lastly, we've" : @receipts_count > 0 ? "We've also" : "We've" %>
+    tagged it with
+    <%= @tagged_with.to_sentence %>.
+  <% end %>
+</p>
```
garyhtou:
```suggestion
<% if @receipts_count > 0 %>
  <p>
    We got your <%= "receipt".pluralize(@receipts_count) %>, thanks!
  </p>
<% end %>

<% if @renamed_to %>
  <p>
    We've <%= "also" if @receipts_count > 0 %> renamed the transaction to <i>"<%= @renamed_to %>"</i>.
  </p>
<% end %>

<% if @tagged_with.any? %>
  <p>
    <%= @receipts_count > 0 && @renamed_to ? "Lastly, we've" : @receipts_count > 0 ? "We've also" : "We've" %>
    tagged it with
    <%= @tagged_with.to_sentence %>.
  </p>
<% end %>
```

garyhtou (approved):
good to merge after fixing merge conflicts and testing!

PR #5199, opened 2024-02-05
Hide unaccessible items on the sidebar when in transparency mode

devramsean0:
## Summary of the problem

This simplifies the UI when in transparency mode, This makes it easier for users to work out when they aren't logged in and hides buttons that could be confusing to people not registered under the org

See #5154 for additional context

## Describe your changes

Removes the disabled class from the items, and instead applies `display: none;` to the hidden elements

sidebar when in transparency mode
image


sarthaktexas (dismissed):

garyhtou (commented):
I think want to entirely hide these from the HTML by not rendering them.
`display: none` 100% works, but let's opt for the cleaner solution here!

### Here's my suggestion:

In [`event/_nav.html.erb`](https://github.com/hackclub/bank/blob/013fb20d304786ddc71235c384c0cd31b3cc58cf/app/views/events/_nav.html.erb#L1), when we call `dock_item`, add an `if`/`unless` to it.

Example:

```ruby
<%= dock_item "blah blah blah" if organizer_signed_in? %>
```

devramsean0:
> I think want to entirely hide these from the HTML by not rendering them. `display: none` 100% works, but let's opt for the cleaner solution here!
> ### Here's my suggestion:
> 
> In [`event/_nav.html.erb`](https://github.com/hackclub/bank/blob/013fb20d304786ddc71235c384c0cd31b3cc58cf/app/views/events/_nav.html.erb#L1), when we call `dock_item`, add an `if`/`unless` to it.
> 
> Example:
> 
> ```ruby
> <%= dock_item "blah blah blah" if organizer_signed_in? %>
> ```

I went for a slightly cleaner method (in my opinion) of stopping the render, I have wrapped the helper in an `unless disabled` block

app/helpers/events_helper.rb
```diff
@@ -4,18 +4,20 @@
 
 module EventsHelper
   def dock_item(name, url = nil, icon:, tooltip: nil, async_badge: nil, disabled: false, selected: false, **options)
```
garyhtou: Let's add a new parameter called `hidden`. If hidden is true, then we won't render the dock_item.

This is necessary because with user permissions, there are cases where we want to disable a dock_item, but still have it visible to the user. For example, an organizer doesn't have access to a user

garyhtou (dismissed):

PR #5201, opened 2024-02-05
Are you a manager?

malted:


app/models/organizer_position.rb
```diff
@@ -7,6 +7,7 @@
 #  id         :bigint           not null, primary key
 #  deleted_at :datetime
 #  first_time :boolean          default(TRUE)
+#  is_manager :boolean          default(FALSE), not null
```
cjdenio: Can this be an enum `role` field? It would give us more flexibility in the future if we wanted more roles

app/views/events/_form.html.erb
```diff
@@ -91,7 +91,7 @@
         </div>
       <% end %>
 
-      <%= form.submit "Update" %>
+      <%= form.submit "Update", class: "#{!policy(@event).update? && "disabled"}" %>
```
cjdenio:
```suggestion
      <%= form.submit "Update", class: ("disabled" unless policy(@event).update?) %>
```

malted: mmmm so clean

app/models/organizer_position.rb
malted: I've been going back and forth on this.
We might expand the perms feature in the future to use a concrete enum of them, but I think we should talk about it more before switching to it (what if we do a matrix of fine-grained feature perms and never end up using the enum? Or something)

northeastprince: I think customizable permissions as opposed to roles is overkill. Even so, the more flexible route is an enum rather than a boolean anyway.

malted: One of my main goals with this project has been to keep it as simple for the user as possible; having only two permission levels plays a big part in this. You're either a normal user, or you're not.
I disagree with the idea that the ability for a manager to choose _exactly_ how someone can spend an org's money is overkill, while having n arbitary, uncustomisable roles is not.
I would like to hear more of your rationale though, do you want to hop on a call this week?

northeastprince: I don't think a call is [necessary](https://dev.37signals.com/the-radiating-programmer) or [beneficial](https://37signals.com/how-we-communicate#:~:text=Internal%20communication%20based%20on%20long%2Dform%20writing) for these types of architectural decisions, but feel free to DM me if you ever wanna schedule one.

I don't have any context on what prompted this, so feel free to correct me, but I've never really seen a need for fine-grained permissions on HCB orgs. I think that, generally, if someone is trusted to have access to a group's money, I would assume there's not much they're _not_ trusted to do within that. That's not to say there isn't a need to restrict some things users can do, which is why I think there definitely should be a few basic roles based on the most common use cases, just that adding fine-grained permissions seems like a lot of complexity that isn't worth it. Again, there are way more people here who have interacted with customers more than me, so I could be way off on this.

Re: an enum vs. a boolean, I think even if we only have 2 roles at the moment, there are a few advantages. First off, its intent is clearer. If I see a `manager` column, I don't really know what it actually means and have to look for its uses to figure out what it does. Whereas if I see a `role` enum on an organizer mapping, I see all the roles and what someone is if they're not a manager, etc. Additionally, a scope is automatically added so you can query by role. Finally, if we do decide to add more roles in the future, a database migration is not necessary.

malted: Hasn't loaded yet but if these are 37signals links i'm

malted:
Note to self: determine if checks are needed on `account_number.html.erb`. (Don't think they are because of the policy)

app/models/organizer_position.rb
garyhtou: Run `annotate` to regenerate the comments

app/views/events/_form_invite.erb
```diff
@@ -8,7 +8,7 @@
 
     <div class="flex items-center">
       <%= form.email_field :email, class: "mr2" %>
-      <%= form.submit 'Send', disabled: !organizer_signed_in? %>
+      <%= form.submit 'Send', disabled: !admin_signed_in? && OrganizerPosition.find_by(user, event: record.event).manager? %>
```
garyhtou: Update the `organizer_signed_in?` helper method to support taking in a role. maybe the syntax should be:

```ruby
organizer_signed_in?(as: :manager)
```

app/views/events/_organization_position_invite.html.erb
```diff
@@ -17,10 +17,22 @@
...
     <%= pop_icon_to "controls", edit_user_path(position.user.slug), class: "ml2", data: { turbo_frame: "_top" } if position.user == current_user && !admin_signed_in? %>
   </section>
+
+  <div class="mb1">
+    <% if position.manager? %>
+      <%= inline_icon "flag-fill", size: 20, style: "vertical-align: middle; margin: -0.2rem 0.5rem 0 0;" %>
+      <span>Is a manager</span>
+    <% else %>
+      <%= inline_icon "flag", size: 20, style: "vertical-align: middle; margin: -0.2rem 0.5rem 0 0;" %>
+      <span>Is not a manager</span>
```
garyhtou: Let's use the term "member"

app/views/events/edit.html.erb
```diff
@@ -77,10 +76,10 @@
       </div>
     <% elsif Flipper.enabled?(:transaction_tags_2022_07_29, @event) %>
       <%# Feature is enabled, let the user opt-out %>
-      <%= link_to "Disable", disable_feature_event_path(feature: :transaction_tags_2022_07_29), method: :post, class: "btn bg-accent" %>
+      <%= link_to "Disable", disable_feature_event_path(feature: :transaction_tags_2022_07_29), method: :post, class: "btn bg-accent #{!policy(@event).disable_feature? && "disabled" } tooltipped tooltipped--w", "aria-label": "Only managers can disable feature previews." %>
```
garyhtou: `!policy(@event).disable_feature? && "disabled"` can be refactored to use `if`/`unless`. see Caleb's comment

app/views/events/team.html.erb
```diff
@@ -37,15 +37,15 @@
         <p class="line-height-3 flex-auto">
           <strong><%= organizer_signed_in? ? invite.user.email : "Sign in to view" %></strong>
         </p>
-        <% if invite.sender == current_user || admin_signed_in? %>
+        <% if admin_signed_in? || (invite.sender == current_user && policy(invite).cancel?) %>
```
garyhtou: We should the policy so that it accounts for `admin_signed_in?` and `invite.sender == current_user`. This way, this line can just be `policy(invite).cancel?`

app/views/invoice_mailer/notify_organizers_recv.html.erb
garyhtou, line 1: can we rename this to `notify_organizers_paid`?

db/schema.rb
```diff
@@ -1237,6 +1237,7 @@
     t.integer "sort_index"
     t.boolean "first_time", default: true
     t.boolean "is_signee"
+    t.boolean "is_manager", default: false, null: false
```
garyhtou: we'll need to run `db:migrate` to update this to `role`

spec/mailers/previews/invoice_mailer_preview.rb
```diff
@@ -1,9 +1,14 @@
...
 
 class InvoiceMailerPreview < ActionMailer::Preview
-  def notify_organizers
+  def notify_organizers_sent
     @invoice = Invoice.last
-    InvoiceMailer.with(invoice: @invoice).notify_organizers
+    InvoiceMailer.with(invoice: @invoice).notify_organizers_sent
+  end
+
+  def notify_organizers_recv
```
garyhtou: renamed to `notify_organizers_paid`

app/mailers/invoice_mailer.rb
```diff
@@ -1,7 +1,18 @@
...
+    @invoice = params[:invoice]
+    @emails = @invoice.sponsor.event.users
+                      .filter { |u| OrganizerPosition.find_by(user: u, event: @invoice.sponsor.event).manager? }
+                      .map { |u| u.email }
+
+    mail to: @emails, subject: "#{@invoice.creator.name} sent an invoice to #{@invoice.sponsor.name}"
+
+  end
+
+  def notify_organizers_recv
```
garyhtou:
```suggestion
  def notify_organizers_paid
```

app/services/invoice_service/open_to_paid.rb
```diff
@@ -29,7 +29,7 @@ def run
       end
 
       if invoice.event.users.any?
-        InvoiceMailer.with(invoice:).notify_organizers.deliver_later
+        InvoiceMailer.with(invoice:).notify_organizers_recv.deliver_later
```
garyhtou:
```suggestion
        InvoiceMailer.with(invoice:).notify_organizers_paid.deliver_later
```

app/models/organizer_position.rb
```diff
@@ -43,6 +44,8 @@ class OrganizerPosition < ApplicationRecord
 
   alias_attribute :signee, :is_signee
 
+  enum :role, { regular: 0, manager: 1 }
```
garyhtou: Let's rename `regular` to `member`

garyhtou:
```suggestion
  # The enum values will allow us to have a hierarchy of roles in the future
  # For example, Managers have access to everything below them
  enum :role, { regular: 25, manager: 100 }
```

db/migrate/20240205181044_add_role_to_organizer_positions.rb
```diff
@@ -0,0 +1,5 @@
+class AddRoleToOrganizerPositions < ActiveRecord::Migration[7.0]
+  def change
+    add_column :organizer_positions, :role, :integer, null: false, default: 0
```
garyhtou: We might want the default to be manager (`100`). Then, add another migration to change the default to `0`. This is because we want everyone to be a manager from the start.

app/models/organizer_position.rb
garyhtou: validate that there is at least one manager per org

app/models/organizer_position.rb
garyhtou, line 1: When a user is automatically invited via the application form, they should be invited to be a manager.

I think this also means we need to keep track of the role when inviting users

app/views/events/team.html.erb
malted: Done

app/views/events/_form_invite.erb
malted: Done

app/views/invoice_mailer/notify_organizers_recv.html.erb
malted: Done

app/models/organizer_position.rb
malted: Will this not be solved by setting the default value to 100?

app/models/organizer_position.rb
malted: Done

app/controllers/invoices_controller.rb
```diff
@@ -130,6 +130,8 @@ def create
 
     flash[:success] = "Invoice successfully created and emailed to #{@invoice.sponsor.contact_email}."
 
+    InvoiceMailer.with(invoice: @invoice).notify_organizers_sent.deliver_now
```
garyhtou: We prefer to deliver later unless there's an urgent need to make sure it gets delivered ASAP

```suggestion
    InvoiceMailer.with(invoice: @invoice).notify_organizers_sent.deliver_later
```

app/controllers/organizer_position_invites_controller.rb
garyhtou, line 1: I think we should pull this into a concern that is used by both `OrganizerPositionInvitesController` and `OrganizerPositionController`

app/controllers/organizer_position_invites_controller.rb
```diff
@@ -94,6 +94,21 @@ def toggle_signee_status
...
   end
 
+  def change_position_role
+    organizer_position_invite = OrganizerPositionInvite.find(params[:id])
+    authorize organizer_position_invite
+
+    to = Integer(params[:to])
+
+    if OrganizerPositionInvite.roles.value?(to)
+      organizer_position_invite.update_attribute(:role, to)
```
garyhtou: `update_attribute` will skip callbacks. We should use the normal `update` instead.

```suggestion
      organizer_position_invite.update(role: to)
```

app/controllers/organizer_position_invites_controller.rb
```diff
@@ -94,6 +94,21 @@ def toggle_signee_status
     redirect_back(fallback_location: event_team_path(@invite.event))
   end
 
+  def change_position_role
+    organizer_position_invite = OrganizerPositionInvite.find(params[:id])
+    authorize organizer_position_invite
+
+    to = Integer(params[:to])
```
garyhtou: I have a slight feeling that Rails will take `params[:to]` (string), and there's no need to cast to an Integer

app/helpers/sessions_helper.rb
```diff
@@ -56,11 +56,14 @@ def current_user=(user)
     @current_user = user
   end
 
-  def organizer_signed_in?(event = @event)
-    @organizer_signed_in ||= Hash.new do |h, event_key|
-      h[event_key] = (signed_in? && event_key&.users&.include?(current_user)) || admin_signed_in?
-    end
-    @organizer_signed_in[event]
+  def organizer_signed_in?(as = :member, event = @event)
```
garyhtou:
```suggestion
  def organizer_signed_in?(event = @event, as: :member)
```

Please update usages to use this new method signature. The reason I prefer a keyword hash is because it's easier to optionally leave out that argument.

app/helpers/sessions_helper.rb
```diff
@@ -56,11 +56,14 @@ def current_user=(user)
     @current_user = user
   end
 
-  def organizer_signed_in?(event = @event)
-    @organizer_signed_in ||= Hash.new do |h, event_key|
```
garyhtou: I think we want to keep this existing memorization. It was likely written this way for performance reason 🤷‍♂️

app/helpers/sessions_helper.rb
```diff
@@ -56,11 +56,14 @@ def current_user=(user)
...
-    end
-    @organizer_signed_in[event]
+  def organizer_signed_in?(as = :member, event = @event)
+    as = as[:as] if as.instance_of?(Hash)
+    position = OrganizerPosition.find_by(user: current_user, event:)
+
+    # We want to make sure the user position's role is >= `as`'s corresponding int.
+    valid_role = (position&.role_before_type_cast || -1) >= OrganizerPosition.roles[as]
+
+    (signed_in? && event&.users&.include?(current_user) && valid_role) || admin_signed_in?
```
garyhtou: this is really tricky one
```suggestion
    run = -> do
      next true if admin_signed_in?
      next false unless signed_in? && event.present?

      required_role_num = OrganizerPosition.roles[as]
      raise ArgumentError, "invalid role #{as}" unless required_role_num.present?

      valid_position = event.organizer_positions.find do |op|
        next false unless op.user == current_user

        role_num = OrganizerPosition.roles[op.role]
        next false unless role_num.present?

        # Allows higher roles to succeed when checking for lower role
        # For example, `organizer_signed_in?(as: :member)` returns true if you're a manager
        role_num >= required_role_num
      end

      valid_position.present?
    end

    # Memoize results based on method arguments
    @organizer_signed_in ||= Hash.new do |h, key|
      h[key] = run.call
    end
    key = [event, as]
    @organizer_signed_in[key]
```
please make sure this still works

app/mailers/invoice_mailer.rb
```diff
@@ -1,7 +1,17 @@
 # frozen_string_literal: true
 
 class InvoiceMailer < ApplicationMailer
-  def notify_organizers
+  def notify_organizers_sent
+    @invoice = params[:invoice]
+    @emails = @invoice.sponsor.event.users
+                      .filter { |u| OrganizerPosition.find_by(user: u, event: @invoice.sponsor.event)&.manager? }
+                      .map { |u| u.email }
```
garyhtou: At the moment, this is an N+1 query. Here's how it would run:
1. Query database for list of users
2. In Ruby, for each user...
	3. Query the database to try to find an OrganizerPosition
4. In Ruby, loop over each user...
	5. Get their email
In total, we make N+1 queries (where N is the number of users). In addition, we also loop over the list of users twice.

Here's my suggestion:

```suggestion
    @emails = @invoice.sponsor.event.users.where(organizer_position: {role: :manager}).pluck(:email)
```
_(please test to make sure it actually works)_

This code should just make one database query and does not require any looping in Ruby.

---

Side note: Both solutions still have 2 additional queries I didn't mention for getting the sponsor and event

app/mailers/invoice_mailer.rb
```diff
@@ -1,7 +1,17 @@
...
+
+    mail to: @emails, subject: "#{@invoice.creator.name} sent an invoice to #{@invoice.sponsor.name}"
+
```
garyhtou: It might be nice to include the name of the Event. especially since HQ peeps are in multiple events
```suggestion
    mail to: @emails, subject: "{@invoice.event.name}: #{@invoice.creator.name} sent an invoice to #{@invoice.sponsor.name}"
```

app/models/organizer_position.rb
```diff
@@ -51,4 +58,10 @@ def tourable_options
     }
   end
 
+  private
+
+  def at_least_one_manager
+    event.organizer_positions.where(role: :manager).any?
```
garyhtou: There is a chance this won't prevent the last user from being removed (i could be wrong. double check it)

This is why:
1. We say `organizer_position.destroy`
2. Rails will run validations on `organizer_position`
3. It queries the database for `event.organizer_positions.where(role: :manager).any?`
4. The query returns `true` because the `organizer_position` has not been deleted **yet** (because rails runs validations before making any updates)
5. Rails actually deleted organizer_position from the database.

app/models/organizer_position.rb
```diff
@@ -43,6 +44,12 @@ class OrganizerPosition < ApplicationRecord
 
   alias_attribute :signee, :is_signee
 
+  # The enum values will allow us to have a hierarchy of roles in the future.
+  # For example, managers have access to everything below them.
+  # TODO: this is replicated on OrganizerPositionInvite. Extract to a concern.
+  enum :role, { member: 25, manager: 100 }
+  validate :at_least_one_manager
```
garyhtou: Let's do as the TODO says and extract this to a concern. let me know if you need help with this

app/policies/hcb_code_policy.rb
garyhtou, line 1: I think we can undo all changes to this file.

We can all users on the org to be able to view HcbCode show pages.

app/policies/organizer_position_deletion_request_policy.rb
```diff
@@ -15,7 +15,7 @@ def create?
     target_is_in_event = event.organizer_positions.include?(record.organizer_position)
     target_has_no_pending_request = record.organizer_position.organizer_position_deletion_requests.under_review.none?
 
-    user.admin? || (user_in_event? && target_is_in_event && target_has_no_pending_request)
+    user.admin? || (user_in_event? && target_is_in_event && target_has_no_pending_request && current_user_is_manager?)
```
garyhtou: If we're checking for `current_user_is_manager?`, then there's no need to check for `user_in_event?`

app/policies/organizer_position_invite_policy.rb
```diff
@@ -26,11 +26,21 @@ def reject?
...
 
+  def change_position_role?
+    user.admin? || organizer_signed_in?(as: :manager)
+  end
+
+  private
+
+  def current_user_is_manager?
+    OrganizerPosition.find_by(user: record.user, event: record.event)&.manager?
+  end
```
garyhtou: This method name is misleading. It appears to be checking for `invitee_is_manager?`

app/policies/organizer_position_policy.rb
```diff
@@ -17,4 +17,14 @@ def toggle_signee_status?
...
   end
 
+  def change_position_role?
+    return true if user.admin?
+
+    user_position = OrganizerPosition.find_by(user:, event: record.event)
+
+    return false if record.is_signee? && !user_position&.is_signee?
+
+    record.manager?
```
garyhtou: I see the intention behind this, but let's keep it simple for now especially since the majority of `is_signee` data in production is wildly incorrect. (Ops hasn't gotten a chance to backfill data)
```suggestion
    current_user_position = OrganizerPosition.find_by(user:, event: record.event)
    current_user_position&.manager?
```

app/policies/organizer_position_invite_policy.rb
garyhtou, line 1: We need to also update the `new?` policy.

app/views/events/_form_invite.erb
```diff
@@ -20,7 +20,7 @@
         </div>
       <% end %>
 
-      <%= form.submit 'Send', disabled: !organizer_signed_in? %>
+      <%= form.submit 'Send', disabled: !organizer_signed_in?(as: :manager) %>
```
garyhtou: This there a way we can have this controlled by the policy?

Maybe something like
```ruby
policy(:organizer_position_invite).new?
```

psst. we also need to update the `new?` policy

app/views/events/_organization_position_invite.html.erb
```diff
@@ -28,7 +47,13 @@
             <%= pop_icon_to "clock", "#", class: "muted", disabled: true %>
           </div>
         <% else %>
-          <%= pop_icon_to "member-remove", new_event_organizer_remove_path(organizer_id: position.user.slug), class: "error tooltipped tooltipped--s", 'aria-label': "Remove user from organization", data: { turbo_frame: "_top" } %>
+          <%= pop_icon_to "member-remove",
+                          new_event_organizer_remove_path(organizer_id: position.user.slug),
+                          class: "error tooltipped tooltipped--s",
+                          'aria-label': "Remove user from organization",
+                          data: { turbo_frame: "_top" },
+                          disabled: !organizer_signed_in?(as: :manager)
```
garyhtou: We could likely use a policy here:

```ruby
policy(:organizer_position_deletion_request).new?
```

app/views/events/_organization_position_invite.html.erb
```diff
@@ -9,6 +9,25 @@
...
+
+        <div class="flex g0 line-height-0 items-center">
+          <p class="h4 info"><%= position.role.capitalize %></p>
+          <% if organizer_signed_in?(as: :manager) %>
+            <div data-controller="menu" data-menu-placement-value="bottom-start" data-action="click->menu#toggle click@document->menu#close keydown@document->menu#keydown" data-menu-target="toggle">
+              <%= inline_icon "down-caret", class: "info pointer" %>
+
+              <div data-menu-target="content" class="menu__content menu__content--2 menu__content--compact">
+                <% OrganizerPosition.roles.each do |(name, value)| %>
+                  <%= link_to change_position_role_organizer_path(:id => position.id, :to => value), data: { turbo_method: "post", turbo_frame: "_top" } do %>
```
garyhtou: It seems like this path (`change_position_role_organizer_path`) is specific to organizer positions. Does it also work for invites?

app/views/invoice_mailer/notify_organizers_sent.html.erb
```diff
@@ -0,0 +1,20 @@
+<p>
+  Hi there,
+</p>
+
+<p>
+  <strong><%= @invoice.creator.full_name %></strong> from your HCB org <em><%= @invoice.sponsor.event.name %></em>
+  just sent an invoice for <strong><%= render_money @invoice.item_amount %></strong>
+  to <em><%= @invoice.sponsor.name %></em> (<%= @invoice.sponsor.contact_email %>)
+  for the purpose of <em><%= @invoice.item_description %></em>
```
garyhtou: copy edit:
>Malted on Hack Club HQ just sent a **$100** [invoice to _Google_](#) for _Hackathon Sponsorship_.

Link the words `"invoice to sponsor.name"` to the invoice

app/views/invoice_mailer/notify_organizers_sent.html.erb
```diff
@@ -0,0 +1,20 @@
+<p>
+  Hi there,
+</p>
+
+<p>
+  <strong><%= @invoice.creator.full_name %></strong> from your HCB org <em><%= @invoice.sponsor.event.name %></em>
```
garyhtou:
```suggestion
  <strong><%= @invoice.creator.name %></strong> from your HCB org <em><%= @invoice.sponsor.event.name %></em>
```

`name` is preferred over `full_name`

db/migrate/20240205181044_add_role_to_organizer_positions.rb
```diff
@@ -0,0 +1,5 @@
+class AddRoleToOrganizerPositions < ActiveRecord::Migration[7.0]
+  def change
+    add_column :organizer_positions, :role, :integer, null: false, default: 100
```
garyhtou:
```suggestion
    add_column :organizer_positions, :role, :integer, null: false, default: 100
    change_column_default :organizer_positions, :role, 25
```
we should change the default back to 25 after creating the column. this way, it matches the invite model

garyhtou (commented):
Great progress here!

app/controllers/invoices_controller.rb
malted: Is this to reduce load on the mail servers? Does it do some sort of scheduling

app/controllers/organizer_position_invites_controller.rb
malted: Good to know! Thanks

app/controllers/organizer_position_invites_controller.rb
malted: It didn't, which is why I cast it.

app/helpers/sessions_helper.rb
malted: I know it was for performance but it - I really have no clue why - didn't work with the memoisation. I would like to find a way to add it back because I don't want to slow all usages of this down!

app/mailers/invoice_mailer.rb
malted: Ahh this is super informative thank you! I should remember to do things in the statement instead of Ruby more often.

malted: This didn't work exactly but pluralising `organizer_position` fixed it! Committed separately because I don't know how do edit this suggestion

app/mailers/invoice_mailer.rb
malted: Good idea!

app/views/invoice_mailer/notify_organizers_sent.html.erb
malted: Done!

db/migrate/20240205181044_add_role_to_organizer_positions.rb
malted: Done!

app/views/events/_organization_position_invite.html.erb
malted: No. I had already created a separate path but wasn't using it. Fixed, cheers!

app/views/events/_organization_position_invite.html.erb
malted: I would have, but no `OrganizerPositionDeletionRequest` exists yet. Do you see a way around this?

app/policies/organizer_position_deletion_request_policy.rb
malted: Done!

app/policies/organizer_position_invite_policy.rb
malted: Done!

app/policies/organizer_position_invite_policy.rb
malted: Oops! Good catch - cheers.

app/models/organizer_position.rb
malted: On agenda for next 1:1.

app/models/organizer_position.rb
malted: I understand this better now, thank you for the explanation.

app/views/events/_organization_position_invite.html.erb
garyhtou: Okay, let's hold off on that for now then

db/migrate/20240226202456_add_role_to_organizer_position_invite.rb
```diff
@@ -0,0 +1,5 @@
+class AddRoleToOrganizerPositionInvite < ActiveRecord::Migration[7.0]
+  def change
+    add_column :organizer_position_invites, :role, :integer, null: false, default: 25
```
garyhtou:
```suggestion
    add_column :organizer_position_invites, :role, :integer, null: false, default: 100
    change_column_default :organizer_position_invites, :role, 25
```

app/policies/ach_transfer_policy.rb
malted, line 1: Woah woah woah shouldn't some of these (`approve?`, `reject?`) only be for admins?

malted: (Talking to myself (not crazy))

app/controllers/organizer_position_invites_controller.rb
```diff
@@ -94,14 +95,34 @@ def toggle_signee_status
...
+    organizer_position_invite = OrganizerPositionInvite.find(params[:id])
+    authorize organizer_position_invite
+
+    was = organizer_position_invite.role
+    to = params[:to]
+
+    if was != to
+      organizer_position_invite.update(role: to)
+
+      flash[:success] = "Changed #{organizer_position_invite.user.name}'s role from #{was} to #{OrganizerPosition.roles.key(to)}."
```
cjdenio:
```suggestion
      flash[:success] = "Changed #{organizer_position_invite.user.name}'s role from #{organizer_position_invite.role_previously_was} to #{organizer_position_invite.role}."
```

app/controllers/static_pages_controller.rb
```diff
@@ -40,6 +40,21 @@ def branding
...
+      { name: "Original Dark", criteria: "For black or dark colored backgrounds.", background: "black" },
+      { name: "Outlined Black", criteria: "For white or light colored backgrounds.", background: "snow" },
+      { name: "Outlined White", criteria: "For black or dark colored backgrounds.", background: "black" }
+    ]
+    @icons = [
+      { name: "Icon Original", criteria: "The original HCB logo.", background: "smoke" },
+      { name: "Icon Dark", criteria: "HCB logo in dark mode.", background: "black" }
+    ]
+    @event_name = signed_in? && current_user.events.first&.name || "Hack Pennsylvania"
+    @event_slug = signed_in? && current_user.events.first&.slug || "hack-pennsylvania"
```
cjdenio:
```suggestion
```

app/views/events/transfers.html.erb
```diff
@@ -11,7 +11,7 @@
   <% if organizer_signed_in? %>
     <%= link_to "Reimbursement", event_reimbursements_path(@event), class: "btn" %>
 
-    <%= link_to event_transfers_new_path(event_id: @event.slug), class: "btn bg-success", data: { behavior: "modal_trigger", modal: "send_transfer" } do %>
+    <%= link_to event_transfers_new_path(event_id: @event.slug), class: "btn bg-success #{!policy(@event).new_transfer? && "disabled"}", data: { behavior: "modal_trigger", modal: "send_transfer" } do %>
```
cjdenio:
```suggestion
    <%= link_to event_transfers_new_path(event_id: @event.slug), class: "btn bg-success #{"disabled" if !policy(@event).new_transfer?}", data: { behavior: "modal_trigger", modal: "send_transfer" } do %>
```

malted:
Being merged with in-person approval from @maxwofford. We're at the team retreat and getting it merged now before we make changes at our 4pm team sync.

PR #5207, opened 2024-02-05
Org Account Statements: Handle no transaction before the start date in PDF exports

sampoder:
## Summary of the problem

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/424/samples/timestamp/2024-02-05T18:39:11Z

If someone generated a report for a month which was their first month of operating and had no transactions in that month, we tried to get a balance from previous months but the org didn't exist in those previous months.

This route was also broken: https://hcb.hackclub.com/exports/[slug].pdf

This is also fixed by appending `to_datetime` no matter whether the date was a param or the default

## Describe your changes

See above


garyhtou (approved):

garyhtou:
thanks for the quick fix on this one!

PR #5220, opened 2024-02-06
Link to the address setting when providing a warning

sampoder:
## Summary of the problem

I thought it would be nice for us to link here, maybe we should change the icon to some sort of edit icon?

## Describe your changes

Made it a link!


polypixeldev (approved):
great! I think the cursor and link hover is enough to signify it

garyhtou (approved):
great!

PR #5227, opened 2024-02-06
Improve criteria for marking a Stripe authorization as declined

cjdenio:
- Respect `pending_request.amount` when it's present
  - This field is present in test mode under certain circumstances (e.g. an authorization that was just created), and I assume live mode acts identically
- Don't decline $0 authorizations until a considerable amount of time has _actually_ passed
  - This is an edge case— only 3 $0 non-reversed authorizations exist, and they were all made on the same day in 2021 (Stripe downtime?)

related to #5208


app/services/pending_event_mapping_engine/decline/stripe.rb
```diff
@@ -17,7 +17,7 @@ def run
           end
 
           # 2. identify authed (0 amount and a considerable amount of time has passed)
-          if rpst.amount_cents == 0
+          if rpst.amount_cents == 0 && Time.at(rpst.stripe_transaction["created"]) < 3.days.ago
```
garyhtou: Should this be 30 days? Captures can happen up to 30 days after an authorization

https://stripe.com/docs/treasury/money-movement/timelines#:~:text=Card%20transaction%20are%20typically%20captured,to%2030%20days%20after%20authorization.

PR #5233, opened 2024-02-07
Remove unused Airtable ApplicationsHelper

aramshiva:
Fixes #5184. Removes Airtable ApplicationsHelper as it is not being used and airtable.base does not even exist!


YodaLightsabr (approved):

garyhtou (approved):
lgtm!

Dog Human Fashion

PR #5235, opened 2024-02-07
Prefix CountryEnumable boolean methods

polypixeldev:
## Summary of the problem

To make it more clear what CountryEnumable boolean methods do on models where it is defined.
Fixes #4933

## Describe your changes

This PR adds the prefix and updates all methods that check if the country enum value is a certain country to use the prefixed methods. It removes the `country_us` method in the event model to the prefixed method, and its references as well.


YodaLightsabr (approved):

garyhtou (approved):
works great! thanks

Michael Scott Thank You

PR #5238, opened 2024-02-08
Admin: Add Event nav item for Legal Documents

sarthaktexas:
## Summary of the problem
Makes a more efficient workflow for admins.

## Describe your changes
Releases a new feature flag `:admin_legal_documents_2024_02_05` for a feature (that allows them to go directly to legal documents) so that admins that don't need this feature don't get it.
image


app/views/events/_nav.html.erb
```diff
@@ -104,6 +104,15 @@
         tooltip: "Commonly asked questions and how-tos",
         icon: "info",
         selected: local_assigns[:selected] == :documentation %>
+      <% if Flipper.enabled?(:admin_legal_documents_2024_02_05, current_user) %>
+        <% admin_tools("m0 p0 overflow-visible w-fit") do %>
+          <%= dock_item "Legal Documents",
+            event_documents_path(@event),
+            tooltip: "Go directly to legal documents",
```
sampoder:
```suggestion
            tooltip: "Fiscal sponsorship contract etc.",
```

sarthaktexas: what

sampoder: well it felt kind of redundant to repeat legal documents so i thought maybe clarifying what sort of documents there are would be nice

sarthaktexas: oh! i see what you mean ✅

garyhtou:
Notes from when @sarthaktexas and I chatted a few weeks ago:
- Let's make this available when right-clicking the existing nav item
- Similar design to right-clicking the HCB icon

PR #5239, opened 2024-02-08
Mock Data: Fixes mock cards & fixes mock tx ledger

sarthaktexas:
## Describe your changes
Fixes mock data so there are only 2 mock cards with two different states. Screenshot 2024-02-05 at 12 26 47 PM

Also makes donations have recurring donations as transactions were listed as recurring donations before.


sampoder (approved):
LGTM!

PR #5265, opened 2024-02-15
`:receipt_required` scope: Remove Emburse transactions & add Increase Checks

sampoder:
We're simplifying this, at least a little, nice! Closes https://github.com/hackclub/hcb/issues/5114.

I've also added a little piece of documentation, which made me realise that this hasn't been considering Increase checks which are HCB-402 so I've added those.


cjdenio (dismissed):
`def type` should also be updated to include Increase checks:

```diff
  def type
	# ...
-   return :check if check?
+   return :check if check? || increase_check?
	# ...
  end
```

sampoder:
Good catch, thank you Caleb!

PR #5270, opened 2024-02-15
Use UTF-8 Encoding On Receipt Bin Emails to PDF

sampoder:
Closes https://github.com/hackclub/hcb/issues/5160.

If someone with wkhtmltopdf locally could test this, that'd be great.


garyhtou (approved):

garyhtou (commented):
i haven't tested this, but it makes sense

garyhtou (changes requested):
didn't work 😔

image

garyhtou (approved):
It doesn't fix emojis because an emoji font is required, but it does fix other encoding problems, so let's merge it

PR #5275, opened 2024-02-15
Remove `hackathons` airtable key in airtable_info

devramsean0:
## Summary of the problem

Remove the airtable key in the airtable_info helper.

See #5183

## Describe your changes

I just deleted the key and it's contents

closes #5183


garyhtou (approved):
thanks!

Cat Apple

PR #5278, opened 2024-02-15
Rename `admin_tools` and `admin_tools_if` to be singular

devramsean0:
## Summary of the problem

This renames the method to be singular instead of plural
See #5212

## Describe your changes

I just searched the project and renamed any occurances

n/a

closes #5212


app/models/hcb_code.rb
sarthaktexas, line 1: ?

sarthaktexas (dismissed):

sarthaktexas (approved):
verified every method name was changed

sarthaktexas (changes requested):
add alias

sarthaktexas (approved):

PR #5280, opened 2024-02-15
Bank Fee Nightly Job: Make transfers to Operating before to Main

sampoder:
## Summary of the problem

Operating normally floats at $0. We can't then take money out of it before putting money in.


garyhtou (approved):
🤞

garyhtou:
_(actually, imma merge this tmrw cause I need to go to sleep now 😴)_

PR #5285, opened 2024-02-16
`hcb_code_pins` -> `pinned_hcb_codes`

sampoder:
Still seems to work: 

Screenshot 2024-02-16 at 1 26 50 PM


app/views/events/_pinned_transactions.html.erb
```diff
@@ -1,6 +1,6 @@
-<% if @event.hcb_code_pins.any? && @event.hcb_code_pins.size > 0 %>
+<% if @event.pinned_hcb_codes.any? && @event.pinned_hcb_codes.size > 0 %>
```
garyhtou: i'm not too sure why this is here
```suggestion
<% if @event.pinned_hcb_codes.any? %>
```

garyhtou (approved):
great! ty

PR #5287, opened 2024-02-18
Improve Edit History View for Comments

sampoder:
## Summary of the problem

Just looked a bit bland.

## Describe your changes

Now it looks like:

Screenshot 2024-02-17 at 11 10 26 PM


app/views/comments/_item_v2.html.erb
```diff
@@ -43,26 +43,36 @@
 </div>
 
 <% if comment.edited? %>
-  <section class="modal" data-behavior="modal" role="dialog" id="comment-modal-<%= comment.hashid %>">
-    <%= modal_header "Edit history" %>
+  <section class="modal modal--scroll bg-snow" data-behavior="modal" role="dialog" id="comment-modal-<%= comment.hashid %>">
+    <%= modal_header "Edit History" %>
     <ul class="list-reset">
-    <% comment.versions.order("created_at desc").filter { |v| v.event == "update" }.each do |version| %>
+      <% comment.versions.reverse.filter { |v| v.event == "update" }.each do |version| %>
```
garyhtou: To improve on the existing code, I _think_ this works (and is more performant)

```suggestion
      <% comment.versions.where(event: "update").order(created_at: :desc).each do |version| %>
```

app/views/comments/_item_v2.html.erb
```diff
@@ -43,26 +43,36 @@
...
+        <li class="mb2">
+          <%= version.whodunnit.nil? ? "Automatically " : user_mention(User.find(version.whodunnit)) %>
+          updated <%= time_ago_in_words version.created_at %> ago
+          <p class="mb1">
+            <div class="card border b--muted">
+sanitize Diffy::Diff.new(
+                  Comment.new(content_ciphertext: version.changeset["content_ciphertext"][0]).content,
+                  Comment.new(content_ciphertext: version.changeset["content_ciphertext"][1]).content
```
garyhtou: ah that's an interesting problem that i didn't realize we'd encounter!

app/assets/stylesheets/components/_comments.scss
```diff
@@ -30,7 +30,7 @@
   margin-top: 0;
 }
 
-.comments .card.current-user {
+.comments .card.current-user .comment_edit_card {
```
garyhtou: Is this new `comment_edit_card` class being used? I'm not seeing references

app/assets/stylesheets/components/_comments.scss
```diff
@@ -53,4 +53,61 @@
...
+.diff ins strong {
+  font-weight: normal;
+  background: #9f9;
+}
+.diff li.diff-comment {
+  display: none;
+}
+.diff li.diff-block-info {
+  background: none repeat scroll 0 0 gray;
+}
```
garyhtou: I wonder if this should be move into it's own file and then imported into this comment file

app/assets/stylesheets/components/_comments.scss
sampoder:
```suggestion
.comments .card.current-user {
```

sampoder: Good catch; thanks!

app/views/comments/_item_v2.html.erb
sampoder: Yeah, it's weird... any better ideas?

garyhtou (approved):

PR #5291, opened 2024-02-18
Add running balance graph

polypixeldev:
## Summary of the problem

It would be nice to have a running balance graph so that users can visualize the balance of an organization over time.
Closes #4649  

## Describe your changes

Adds a running balance graph behind the balance text on the event show page, with the number in front updating with the date that the user is hovering on. Graph color changes with if the graph is trending up or down.


sampoder:
Woah!! This is really cool:

Screenshot 2024-02-18 at 10 58 43 PM

garyhtou:
small thought: i wonder if there's a way to smooth out the line to make it more visually appealing

image

app/models/event.rb
```diff
@@ -478,8 +478,8 @@ def fronted_incoming_balance_v2_cents(start_date: nil, end_date: nil)
       begin
         pts = canonical_pending_transactions.incoming.fronted.not_declined
 
-        pts = pts.where("date >= ?", @start_date) if @start_date
-        pts = pts.where("date <= ?", @end_date) if @end_date
+        pts = pts.where("date >= ?", start_date) if start_date
+        pts = pts.where("date <= ?", end_date) if end_date
```
polypixeldev: Although this is no longer being used in this PR, it's still a small bug that should be fixed - should it be brought out into a separate PR?

app/assets/stylesheets/application.scss
```diff
@@ -401,6 +401,37 @@ summary {
...
+  }
+}
+
+.sparkline--down {
+  stroke: $red;
+  &.sparkline--filled {
+    fill: rgba($red, 0.3);
+  }
+}
+
```
garyhtou: Can we move this to a separate file?

polypixeldev:
Latest code:
[Screencast from 2024-03-07 21-34-59.webm](https://github.com/hackclub/hcb/assets/79737178/517e560f-83da-4b5c-94d9-ce68f4457406)
It seems to work for my Expensicon test event, but for another (HCB fan club) it seems to glitch and calculate 0 for all days, and the frontend seems to also not like it.

If I use the previous method of calculating, the balance is calculated right, although width still seems to shift slightly, frontend needs fix
[Screencast from 2024-03-07 21-36-28.webm](https://github.com/hackclub/hcb/assets/79737178/9599039d-106c-4e5c-8628-a7dcf081271a)

app/models/event.rb
sampoder: Let's separate this into a separate PR

sampoder (approved):
This looks good to go! Thank you so much for the work you put into this @polypixeldev.

Would you like to ship this?

polypixeldev:
Sure! I can start drafting a post today

PR #5292, opened 2024-02-18
Use `@inbound_mail&.mail&.subject` and then have a fallback

sampoder:
This makes progress towards fixing https://github.com/hackclub/hcb/issues/5286. I think that one issue might be us adding a new subject (this shouldn't break it... the `In-Reply-To` header should just work! but no Gmail!) so I've made it so that we only use the custom subjects as fallbacks.

@cjdenio thoughts? Was there a reason you avoided doing this in https://github.com/hackclub/hcb/pull/4786. 

Also, does anyone know, is there a cleaner way to do this?


garyhtou:
Would the stuff we learned about `thread_id` and `references` headers help here?

sampoder:
We should certainly do that! Let's put it in a separate PR though

sampoder:
I'm going to try this out, if it doesn't fix the bug in Gmail I'm going to revert it

garyhtou:
Great! I do think we should still implement `thread_id` and `references` headers regardless

PR #5293, opened 2024-02-18
Subtle Admin Comments Redesign

sampoder:
## Summary of the problem

Closes https://github.com/hackclub/hcb/issues/5272.

## Describe your changes

Thought this looked nice, happy to make it more prominent. 

Screenshot 2024-02-18 at 2 25 33 PM

Screenshot 2024-02-18 at 2 26 13 PM


cjdenio (changes requested):
This causes admin comments to be shown to normal users!

app/assets/stylesheets/_dark.scss
```diff
@@ -193,6 +193,11 @@ html[data-dark='true'] {
   .comments .card.current-user {
     background-color: mix($darkless, map-get($palette, info), 90%);
   }
+  
+  .comments .card.comment__admin {
```
garyhtou: I wonder if `admin` should be a modifier instead of an element?

https://getbem.com/naming/

`comment--admin`

app/views/comments/_item_v2.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="mt2 mb2">
```
garyhtou: I _think_ this might fix the problem. please test it!

```suggestion
<% return if comment.admin_only? && !admin_signed_in? %>

<div class="mt2 mb2">
```

app/views/comments/_list_v2.html.erb
```diff
@@ -1,12 +1,6 @@
 <section class="comments">
   <% @hcb_code.comments.order(created_at: :asc).each do |comment| %>
```
garyhtou: rendering as a collection is more performant

```ruby
<%= render "comments/item_v2", collection: @hcb_code.comments.order(created_at: :asc), as: :comment %>
```

sampoder:
LOL! My bad...

sampoder:
Fixed!

garyhtou (commented):
oops, i left comments, but never published those comments

app/views/comments/_item_v2.html.erb
garyhtou: Instead of handling the authorization in the list partial, we should do it in the item partial for 3 reasons:

1. ~~i think the ternary will mess up preloading~~ (jk i lied)
2. adding the condition in the item partial will also "protect" other usages of the item partial
3. having this condition in the item partial makes it easy to do `policy(comment).show?` if we wanted to use Pundit

cjdenio: This might also be a good use-case for [Pundit scopes](https://github.com/varvet/pundit?tab=readme-ov-file#scopes)

sampoder: Ended up going with Pundit, but yep!

sampoder:
@cjdenio & @garyhtou pushed some code that uses Pundit, thoughts? Also added in Pundit scopes.

app/policies/comment_policy.rb
```diff
@@ -17,4 +37,8 @@ def update?
     user.admin? || (record.commentable.event.users.include?(user) && record.user == user)
   end
 
+  def show?
+    user.admin? || (record.commentable.event.users.include?(user) && !record.admin_only)
```
cjdenio: `commentable.event` will have weird behavior on disbursements (which have more than one Event)

app/views/comments/_list_v2.html.erb
```diff
@@ -1,15 +1,8 @@
...
-      <% end %>
-    <% else %>
-      <%= render "comments/item_v2", comment: %>
-    <% end %>
-  <% end %>
 
-  <% if @hcb_code.comments.empty? && local_assigns[:blankslate?] %>
+  <%= render partial: "comments/item_v2", collection: policy_scope(@hcb_code.comments).order(created_at: :asc), as: :comment %>
+
+  <% if policy_scope(@hcb_code.comments) && local_assigns[:blankslate?] %>
```
cjdenio:
```suggestion
  <% if policy_scope(@hcb_code.comments).empty? && local_assigns[:blankslate?] %>
```

app/views/comments/_item_v2.html.erb
```diff
@@ -1,5 +1,7 @@
+<% return if !CommentPolicy.new(current_user, comment).show? %>
```
cjdenio:
```suggestion
<% return if !policy(comment).show? %>
```

app/policies/comment_policy.rb
sampoder: Should be fixed now!

garyhtou:
I think this is good to go, but I'll leave the final review to @cjdenio!

sampoder:
Coolio

cjdenio (approved):

cjdenio:
This is great, thanks!

PR #5299, opened 2024-02-19
Use `EventPolicy` for exports

sampoder:
Closes https://github.com/hackclub/hcb/issues/5298


garyhtou (approved):

garyhtou:
thanks for the quick fix!

PR #5303, opened 2024-02-20
OAuth authorization: allow overriding color scheme via query param

cjdenio:
worthwhile feature


garyhtou (commented):
maybe move this into a controller concern?

cjdenio:
The thing is: the controller for this is defined by a gem and can't easily be modified

garyhtou:
it's a bit of work, but i think this might work.

https://doorkeeper.gitbook.io/guides/configuration/other-configurations#custom-controllers

```ruby
class RespectThemeParam
  extend ActiveSupport::Concern
  included do
    before_action do
      # code
    end
  end
end
```

```ruby
class DoorkeeperController < ActionController::Base
  include RespectThemeParam
end
```

```ruby
Doorkeeper.configure do
  base_controller 'DoorkeeperController'
end
```

you can also totally skip the concern, but i think it might be nice because we likely will want to reuse this logic elsewhere.

garyhtou:
Or, the easier, but janky (and not recommended solution) is to reopen the `Doorkeeper::ApplicationController` class.

```ruby
class Doorkeeper::ApplicationController
  include RespectThemeParam
end

PR #5305, opened 2024-02-20
Add email and phone number to team cards

malted:
Goes from this:
image
To this:
image

Click on the email and phone to copy.

(The manager dropdown ~~is hidden~~ has been removed for now; will be needed v soon)


app/views/events/_organization_position_invite.html.erb
```diff
@@ -1,26 +1,19 @@
...
-                <%= inline_icon("message-simple", size: 24, class: "warning") %>
-              <% end %>
-            <% end %>
+        <% if false %> <!-- if feature enabled -->
+          <div class="inline-flex flex-row items-center info">
+            <p class="m0">Manager</p>
+            <%= inline_icon("down-caret", size: 28) if true %> <!-- if admin or manager -->
           </div>
         <% end %>
       </div>
```
garyhtou: can we move this code over to the other PR? It shouldn't be a part of this PR if it's not used in this PR. this helps keep the git history clean.

thank you! i can help move the code if you'd like!

garyhtou (approved):
great! ty

PR #5308, opened 2024-02-21
Refactor Comments to use `list_v2` and `form_v2`

sampoder:
## Summary of the problem

Closes https://github.com/hackclub/hcb/issues/5294 and, as a result of that, also closes https://github.com/hackclub/hcb/issues/5297.

## Describe your changes

`list_v2` and `form_v2` now take `comments` and `commentable` local variables respectively.


garyhtou:
with these changes, are any of the V1s still being used?

if not, maybe we can remove them?

PR #5309, opened 2024-02-21
Add a `hcb internals` category and disable usage breakdowns for them

sampoder:
See https://hackclub.slack.com/archives/C026RKHLPNJ/p1708535816743239.


app/models/event.rb
```diff
@@ -369,6 +369,7 @@ class Event < ApplicationRecord
     'grant recipient': 9,
     salary: 10, # e.g. Sam's Shillings
     ai: 11,
+    'hcb operations': 12 # eg. https://hcb.hackclub.com/clearing
```
garyhtou: What are your thoughts on "hcb internals"?

```suggestion
    'hcb internals': 12 # eg. https://hcb.hackclub.com/clearing
```

we might end up using this category for other things such as a clearinghouse for reimbursements payouts

garyhtou (approved):
ty

PR #5310, opened 2024-02-21
Bank Fee: Fix book transfer direction

garyhtou:
and rename method to `book_transfer_accounts`


garyhtou:
this stuff's confusing 😵‍💫

PR #5318, opened 2024-02-22
Use Apple Maps for maps via `maps.hackclub.com`

sampoder:
Screenshot 2024-02-21 at 7 45 12 PM


garyhtou:
so 🆒❗

garyhtou:
What are some good things to know about that API?
- Does it have a rate limit?
- Does it use an API Key?
- Is there an associated repo?

it might be nice to comment some info above that url

sampoder:
Added!

garyhtou (approved):
we love 

PR #5319, opened 2024-02-22
Use `stripe_obj&.shipping&.tracking_number&.present?`

sampoder:
Should fix the new bug.


garyhtou (approved):
\*blindly approving prs\*

i _trust_ you and caleb

sampoder:
still testing before i merge

sampoder:
Screenshot 2024-02-21 at 8 01 34 PM

works on an erroring card

garyhtou:
It looks like this broke things in development. When the Stripe API doesn't work, `stripe_obj` returns a Hash, which we can't call methods on.
https://github.com/hackclub/bank/blob/f5ccd23350132916b9220075bf3f63eb65800eb5/app/models/stripe_card.rb#L226

I'd suggest converting that Hash into a [`Struct`](https://ruby-doc.org/core-2.5.1/Struct.html)

PR #5321, opened 2024-02-22
Use `&` syntax for ETA

sampoder:
https://hackclub.slack.com/archives/C04LDFJ5X7S/p1708575724646049


app/views/stripe_cards/_shipping.html.erb
```diff
@@ -33,7 +33,7 @@
       </footer>
     <% else %>
       <section class="pl2 muted pb2 mb1">
-        <% eta = stripe_card.stripe_obj[:shipping][:eta] %>
+        <% eta = stripe_card&.stripe_obj&.shipping&.eta %>
```
garyhtou:
```suggestion
        <% eta = stripe_card.stripe_obj&.shipping&.eta %>
```
stripe_card wouldn't be `nil`. if it were, we'd get many errors above

garyhtou (approved):

garyhtou (changes requested):
can we update the other `[:shipping]` above? (convert the `[]` syntax into method calls)

sampoder:
Done!

garyhtou (approved):

PR #5322, opened 2024-02-22
Create `HcbCode#pinnable?` and use it hide the pin option

sampoder:
Cleaner solution to what we had before.


garyhtou (commented):
what if we moved the condition into the partial?

garyhtou:
hahah the typo was my fault lolimage

garyhtou:
can you get rid of this method?
https://github.com/hackclub/bank/blob/b3619501bdd10134d7b0ce9d57317e55e6dc3450/app/models/hcb_code.rb#L353

sampoder:
I've instead gone with `pinnable?` and also integrate a check for `event` (this error happened last night)

garyhtou (approved):
makes sense to me

PR #5324, opened 2024-02-22
Add more support for BankFee credits

garyhtou:
- Previously, fee credits did not have pending transactions.
	- Now they have them!
	- They're also fronted (can be spent instantly)
	- image
	   As you can see, the fee credit under the balance is gone. Instead, it's replaced with a fronted pending transactions
- Previously, the Bank Fee icon tooltip said "Fiscal sponsorship payment".
	- Now it says "Fiscal sponsorship" and the memo provides more details on whether it's a fee or credit.
- We also now account for fee credits in an Org's total fees paid
	- Fee credits should decrease `Event#total_fee_payments_v2_cents`.
	- Previously, processing a fee credit would INCREASE the fee credit that an org is owed. This was caused by the `abs` instead of negating the value.


garyhtou:
please sanity check this 🙏

sampoder (approved):
This looks good to me!

garyhtou (commented):
got verbal approval from Caleb last week

PR #5328, opened 2024-02-22
Use `OpenStruct` for `stripe_obj` in dev

sampoder:
See https://hackclub.slack.com/archives/C04LDFJ5X7S/p1708573327727739


garyhtou:
I just came across the `Data` class in Ruby. maybe that's more suitable? (i haven't looked into this)

sampoder:
Is there an easy way to convert a Hash to a Data object? I kind of like how simple the current version is

PR #5337, opened 2024-02-23
Slight Redesign of Stripe Card Shipping

sampoder:
## Summary of the problem

Works towards but doesn't finish https://github.com/hackclub/hcb/issues/5333. 

## Describe your changes

I haven't done the text overlay, I'm not sure what we're looking for there?

Screenshot 2024-02-23 at 1 15 32 PM

(this card doesn't have a map yet)


garyhtou:
The thought behind the text overlay was that the map looks a bit bare by itself. it's almost like whitespace.

another solution could be making the map smaller

malted:
+1 for making the map smaller. Try making it wider than it is tall?

sampoder:
Thoughts on this?

Screenshot 2024-02-24 at 5 04 52 PM

garyhtou:
i love it! what if we reorganized it to be:

```
heading
estimated to arrive

(space)

address
```

PR #5341, opened 2024-02-26
Redesign the Organization Settings Page

devramsean0:
## Summary of the problem

The current org page is confusing and long, this PR splits it into multiple individual tabs

## Describe your changes

I wanted to make it fit inside the current pages routes. To do this I am selectively rendering a partial according to a `tab` query parameter.

I modelled the design after the User settings page

image

## TODO:
- [x] Decide on final categorization's for settings
- [x] Update now potentially broken links 
- [x] Ensure that all settings are accounted for

Closes #6003


malted:
Do you think it would make sense to move the donation settings to the donations page?

It's the only organisation setting that corresponds to a main tab (`/donations`).

Only concern with this is that it would lead to settings' locations becoming more fragmented.

lachlanjc:
@garyhtou, this is ready for review! I've dusted off this PR & updated the panes with the latest settings UI. I've adopted the new native browser `switch` type for checkboxes, which looks cleaner in Safari (& coming to more browsers). We can think about moving Donations/Reimbursements settings out later, but I think breaking Feature Previews onto its own tab is already a good start. This paves the way to move Google Workspace in here in a separate PR.

Here's the 3 tabs:

CleanShot 2024-05-17 at 14 03 23@2x
CleanShot 2024-05-17 at 14 03 42@2x
CleanShot 2024-05-17 at 14 05 19@2x

app/views/events/settings/_admin.html.erb
```diff
@@ -0,0 +1,144 @@
+<div class="flex flex-col gap-4">
+<% admin_tool do %>
+  <strong>Org slug history:</strong> <%= @event.slugs.pluck(:slug).join ", " %>
```
sampoder: Minor copy edit.

```suggestion
  <strong>Organization's slug history:</strong> <%= @event.slugs.pluck(:slug).join ", " %>
```

app/views/events/settings/_details.html.erb
```diff
@@ -1,9 +1,9 @@
 <%= form_with(model: event, local: true, class: "mb3") do |form| %>
   <%= form_errors event, "organization" %>
 
-  <h3 class="mb1">Organization profile</h3>
+  <h3 class="mb1">Organization details</h3>
```
sampoder: I think we can delete this header, it seems to be redundant: 
Screenshot 2024-05-18 at 3 31 49 AM

sampoder (commented):
In general, this PR looks good to me. A couple of quick comments.

1) Yes, I agree we should split off Reimbursements / Donations into it's own tab. Do I think this should block this PR? No, but it'd be nice to have.

2) I think the "Account Information" at the top is out of place:

Screenshot 2024-05-18 at 3 33 33 AM

It doesn't feel like a distinct element on the page and feels like it's been shoved in there.

Here's a quick mockup of putting that information into a card:

Screenshot 2024-05-18 at 3 35 31 AM

The only thing I don't like about the card is that it's very wide but our text isn't.

sampoder:
Thank you for helping get this PR over the line!

lachlanjc:
@sampoder Good call on the org info, I wasn't a fan of that either. I think making it one sentence is fine:
CleanShot 2024-05-21 at 16 42 35@2x

lachlanjc:
Thinking the same on pulling out more tabs, want to ship this then keep iterating but it's already an improvement!

sampoder:
Sentence looks perfect, thanks for that

sampoder (dismissed):
LGTM

garyhtou (approved):
Great! Thank you for the work @lachlanjc. This is approved by @MelSmith104 and myself. We'll merge this on Monday after notifying the ops team

PR #5353, opened 2024-02-28
Check if onboarded before rendering the feedback form

sampoder:
## Summary of the problem

Fixes https://github.com/hackclub/hcb/issues/5347.

## Describe your changes

Based on a user's events


app/models/user.rb
```diff
@@ -236,6 +236,10 @@ def transactions_missing_receipt
     end
   end
 
+  def onboarded?
```
garyhtou: i think this method name may be misleading because Users also have a separate onboarding flow for setting up their profile (name, phone, etc.)

i can't think of a good name. i think in this case, it might be okay to remove the method and all `events.not_demo_mode.any?` in both places

app/views/layouts/application.html.erb
```diff
@@ -179,7 +179,7 @@
       </main>
     <% end %>
 
-    <%= render "application/feedback_form" if signed_in? %>
+    <%= render "application/feedback_form" if current_user&.onboarded? %>
```
sampoder:
```suggestion
    <%= render "application/feedback_form" if current_user&.events&.not_demo_mode&.any? %>
```

sampoder:
Sure thing, done

garyhtou (approved):

PR #5363, opened 2024-02-29
Use `column_transactions`'s `available_amount` if `amount_cents == nil`

sampoder:
## Summary of the problem

Closes https://github.com/hackclub/hcb/issues/5362

## Describe your changes

I use the column transaction object's `available_amount` key which we use to eventually set the `amount_cents`: https://github.com/hackclub/hcb/blob/7eecf220f1048a154af2b64c4fb9c9ef6a0fc68a/app/services/transaction_engine/nightly.rb#L95


cjdenio (commented):
If `amount_cents` is nil, it stands to reason that `column_transaction` would also be nil (they're set in the same place). I'm wondering if there's a weird Ruby thing with this being in a `rescue` block

sampoder:
Oh yeah good point, I think it's because we set it after creating it

cjdenio:
I thought of that, though it looks like passing a block to `find_or_create_by` should work as expected: https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-find_or_create_by:~:text=This%20method%20accepts%20a%20block%2C%20which%20is%20passed%20down%20to%20create.

PR #5371, opened 2024-03-01
Cache Transactions Table for Public Viewers

sampoder:
Brought the SQL queries for HQ down to 68 from the hundreds on my local branch.


sampoder:
This introduced some.... unintended consequences.

sampoder:
This is breaking for non-pending transactions.

app/views/events/show.html.erb
```diff
@@ -129,13 +129,15 @@
...
             </tr>
           </thead>
-          <tbody data-behavior="transactions">
-            <%= render "pending_fee_transaction" if @event.fronted_fee_balance_v2_cents > 0 %>
-
-            <%= render partial: "canonical_pending_transactions/canonical_pending_transaction", collection: @pending_transactions, as: :pt, locals: { event: @event, show_amount: true } %>
-
-            <%= render partial: "canonical_transactions/canonical_transaction", collection: @transactions, as: :ct, locals: { event: @event, show_amount: true } %>
-        </tbody>
+          <% cache_if !organizer_signed_in? do %>
```
northeastprince: Why only for public viewers?

sampoder: It'd be great to be able to do this for organisers but I think that'll need more testing. This change is just so that we can handle surges in traffic.

northeastprince: So you didn't run into any stale data when testing it as an organizer?

sampoder: I haven't tested it as an organiser, my focus was on public viewers

northeastprince: Why?

sampoder: https://github.com/hackclub/hcb/pull/5371#discussion_r1509576347, goal of PR is to handle surges in traffic to transparent organisations. Just because some media stuff might be happening soon.

northeastprince: But if it works for organizers too then we shouldn't have that conditional. However, it looks like there a lot of organizer and admin checks littered throughout the partials, so it's not feasible at the moment.

app/views/events/show.html.erb
```diff
@@ -129,13 +129,15 @@
               <th></th>
             </tr>
           </thead>
-          <tbody data-behavior="transactions">
-            <%= render "pending_fee_transaction" if @event.fronted_fee_balance_v2_cents > 0 %>
+          <% cache_if !organizer_signed_in? do %>
```
garyhtou: believe this will cache forever since it doesn't have a key to work with.

See https://api.rubyonrails.org/v7.1.3/classes/ActionView/Helpers/CacheHelper.html#method-i-cache

garyhtou:
BTW, before we implement aggressive caching, we might need to handle https://github.com/hackclub/hcb/issues/3676 so that we don't take down Sidekiq.

At the moment, Rails cache and Sidekiq share the same Redis instance. This means that if Redis gets filled up (due to Rails caching), then Sidekiq will break.

PR #5372, opened 2024-03-01
Update Recipients Logic for HCB Comments

sampoder:
Closes https://github.com/hackclub/hcb/issues/5369


app/models/hcb_code.rb
```diff
@@ -420,11 +420,19 @@ def generate_and_set_short_code
   end
 
   def comment_recipients_for(comment)
+    users = []
+
+    users += self.comments.map { |c| c.user }
+
+    users += self.events.flat_map { |e| e.users }
+
+    users += self.events.map { |e| e.point_of_contact }
```
```suggestion
    users = []
    users += self.comments.map(&:user)
    users += self.events.flat_map(&:users)
    users += self.events.map(&:point_of_contact)
```

app/models/hcb_code.rb
```diff
@@ -420,11 +420,19 @@ def generate_and_set_short_code
...
+
     if comment.admin_only?
-      return User.admin.pluck(:email).excluding(comment.user.email)
+      return users.select { |u| u.admin? }.collect(&:email).excluding(comment.user.email)
```
```suggestion
      return users.select(&:admin?).collect(&:email).excluding(comment.user.email)
```

app/models/hcb_code.rb
```diff
@@ -420,11 +420,19 @@ def generate_and_set_short_code
...
     end
 
-    self.events.flat_map { |e| e.users.pluck(:email) }.excluding(comment.user.email)
+    return users.collect(&:email).excluding(comment.user.email)
```
```suggestion
    users.collect(&:email).excluding(comment.user.email)
```

garyhtou (commented):
Please test and merge!

I'm handing you the 🪄 merge wand for this one

garyhtou (changes requested):
hmm I just realized something. This will include point of contacts in the same email. Users may get freaked out when they see their point of contact was notified for an unimportant comment.

This problem also applies to users who left admin-only comments. From a user's point of view, they would be very confused as to why random people (admins) also received the notification email (as seen in the "to" field of the email).

garyhtou:
The solution might be to allow `comment_recipients_for` to return a nested array:

- To send 1 email that is addressed to three 3 users, the method should return
    ```ruby
    [user1, user2, user3]
    # or
    [[user1, user2, user3]]
    ```

- To send 2 emails that are each addressed to 2 users, the method should return
    ```ruby
    [[user1, user2], [user3, user4]]
    ```

- To send a separate email to each user, the method should return
    ```ruby
    [[user1], [user2], [user3], [user4]]
    ```

sampoder:
That makes sense, I'm going to move them into `bcc` and make the `reply-to` simply the comment creator. That feels pretty standard to me, that's what GitHub does for comments.

sampoder:
The `reply-to` is temporary. Long term, the should be a comments mailbox that let's you make comments by replying.

sampoder:
I'm going to merge this because I think this getting annoying for admins

PR #5374, opened 2024-03-02
[Stripe Card] Rename `activated` to `initially_activated` pt 1/3

rluodev:
This PR creates the `:initially_activated` column and modifies the activation controller to start writing activation state to it as well as the original `:activated` column. In addition, it also contains a migration to migrate data from `:activated` to `:initially_activated`.

The next PR will switch all code from using `:activated` column to `:initially_activated`, and the final PR will remove the `:activated` column after we've verified that data has been migrated properly.


app/controllers/stripe_cards/activation_controller.rb
```diff
@@ -38,6 +38,7 @@ def create
       end
 
       @card.update(activated: true)
+      @card.update(initially_activated: true)
```
garyhtou: we can combine this with the line above
```suggestion
      @card.update(activated: true, initially_activated: true)
```

garyhtou (commented):
I'll take a closer look at this after sarthak reviews

db/migrate/20240302012613_migrate_data_activated_to_initially_activated_stripe_cards.rb
```diff
@@ -0,0 +1,7 @@
+class MigrateDataActivatedToInitiallyActivatedStripeCards < ActiveRecord::Migration[7.0]
+  def change
+    StripeCard.find_each(batch_size: 1000) do |card|
```
sarthaktexas: find_each [automatically uses a batch_size of 1000](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each)!
```suggestion
    StripeCard.find_each do |card|
```

sarthaktexas (dismissed):
This looks good to me. I'm out of the loop as to why we need this, but assuming we do, good work! I have one tiny little thing, but Gary might prefer being more verbose about it.

db/migrate/20240302011059_add_initially_activated_column_stripe_cards.rb
```diff
@@ -0,0 +1,9 @@
+class AddInitiallyActivatedColumnStripeCards < ActiveRecord::Migration[7.0]
+  def up
+    add_column :stripe_cards, :initially_activated, :boolean, default: false
```
garyhtou: Do we want to add `null: false` to make the column require a value?

rluodev: it hasn't historically been null: false (the original column isn't)

garyhtou: good point! I just did a quick query in production
image


It looks like the only values are `false` and `true`. So we should be good to make it `null: false`. Please do so! thanks

rluodev: fixed!

db/migrate/20240302012613_migrate_data_activated_to_initially_activated_stripe_cards.rb
```diff
@@ -0,0 +1,7 @@
+class MigrateDataActivatedToInitiallyActivatedStripeCards < ActiveRecord::Migration[7.0]
+  def change
+    StripeCard.find_each do |card|
+      card.update(initially_activated: card.activated)
+    end
```
garyhtou: Try `StripeCard.update_all("initially_activated = activated")`

I think this will be faster. The current code took 76 secs for me to run

rluodev: fixed!

garyhtou (approved):

garyhtou:
@sampoder when @rluodev creates PRs for part two (switch usage to `initially_activated` and ignore `activated` column) and three (drop `activated` column), can you please help review and merge them?

garyhtou:
the `initially_activated` column has been backfilled!
https://hcb.hackclub.com/console/sessions/53

sampoder:
Yeah fs!

From Sam

PR #5376, opened 2024-03-02
Avoid Paginated CSV Exports on `/admin/events`

sampoder:
Fixes https://github.com/hackclub/hcb/issues/5365.


garyhtou (approved):
seems reasonable to me. lgtm!

PR #5377, opened 2024-03-02
Reduce SQL Queries Required To Load The Homepage (HQ from approx. 2000 -> 400)

sampoder:
HCB is slow! We make a lot of queries to load each page. Let's do a little less of that. This PR has a bunch of random things that bring down the queries.

The results are:

Screenshot 2024-03-01 at 8 01 42 PM

Screenshot 2024-03-01 at 8 02 29 PM


app/models/hcb_code.rb
```diff
@@ -108,7 +108,7 @@ def humanized_type
 
   def amount_cents
     @amount_cents ||= begin
-      return canonical_transactions.sum(:amount_cents) if canonical_transactions.exists?
+      return canonical_transactions.sum(:amount_cents) if canonical_transactions.any?
```
sampoder: Every time we do `exists?` it makes a database call, `any?` doesn't do that: https://stackoverflow.com/questions/48350500/differences-between-any-and-exists-in-ruby-on-rails

app/controllers/events_controller.rb
```diff
@@ -86,7 +86,7 @@ def show
     if !signed_in? && !@event.holiday_features
       @hide_holiday_features = true
     end
-
+
```
```suggestion

```
this whitespace does not speed up things

garyhtou: hahah

garyhtou: i wish

app/models/hcb_code.rb
```diff
@@ -178,15 +178,15 @@ def events
       end
   end
 
-  def pretty_title(show_event_name: true, show_amount: false)
+  def pretty_title(show_event_name: true, show_amount: false, event_name: nil, amount_cents: self.amount_cents)
```
sampoder: Previously `pretty_title` was making multiple database calls to get the event name and the amount cents. We already know that!

app/models/hcb_code.rb
garyhtou: Yeah, that's what I think too. I did also find this post that seems to contradict that
https://www.ombulabs.com/blog/benchmark/performance/rails/present-vs-any-vs-exists.html

Can you double-check that `any?` is better than `exists?`. Maybe I'm misunderstanding the post

sampoder: `exists?` is better when you don't have preloaded records, that seems to be my understanding

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -1,4 +1,5 @@
 <% instance = pt.__id__ %>
+<% tagged_with = pt.local_hcb_code.tags.filter { |tag| !@event || tag.event == @event } %>
```
sampoder: Unfortunately, I haven't figured out preloading this one.  However, this method prevents us from doing this query for each tag that an event has. On events like HQ, they have a lot of tags so that would cause a lot of queries.

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -136,9 +137,9 @@
   <% end %>
 </tr>
 
-<% if Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && pt.local_hcb_code.event.present? %>
+<% if Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (@event || pt.local_hcb_code.event).present? %>
```
sampoder: We already know the event! Let's not try to calculate it

app/views/hcb_codes/_memo.html.erb
```diff
@@ -4,7 +4,7 @@
 <% prepended_to_memo ||= "" %>
 <% renamed ||= false %>
 
-<% if organizer_signed_in?(hcb_code.event) && !hcb_code.event&.demo_mode? %>
+<% if organizer_signed_in?(@event || hcb_code.event) && !(@event || hcb_code.event)&.demo_mode? %>
```
garyhtou: ah clever

app/views/canonical_transactions/_canonical_transaction.html.erb
```diff
@@ -52,7 +53,7 @@
                 <% (@event || ct.local_hcb_code.event).tags.each do |tag| %>
                   <div class="flex items-center">
                     <%= button_to toggle_tag_hcb_code_path(id: ct.local_hcb_code.hashid, tag_id: tag.id), class: "menu__action #{tag_dom_id(ct.local_hcb_code, tag, "_toggle")}", form_class: "flex-auto", form: { "data-turbo" => "true" } do %>
-                      <%= "✓" if ct.local_hcb_code.tags.include?(tag) %>
+                      <%= "✓" if tagged_with.include?(tag) %>
```
garyhtou: I don't think this line functions the same as before. is this intentional?

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -53,7 +54,7 @@
                 <% (@event || pt.local_hcb_code.event).tags.each do |tag| %>
                   <div class="flex items-center">
                     <%= button_to toggle_tag_hcb_code_path(id: pt.local_hcb_code.hashid, tag_id: tag.id), class: "menu__action #{tag_dom_id(pt.local_hcb_code, tag, "_toggle")}", form_class: "flex-auto", form: { "data-turbo" => "true" } do %>
-                      <%= "✓" if pt.local_hcb_code.tags.include?(tag) %>
+                      <%= "✓" if tagged_with.include?(tag) %>
```
garyhtou: Same comment as `_canonical_transaction.html.erb`

app/views/canonical_transactions/_canonical_transaction.html.erb
sampoder: Yeah it's intentional, when on a specific event we should only show tags that are on that event

garyhtou:
This is a crazy improvement! I'll finish reviewing it later!

sampoder:
should be included: https://github.com/hackclub/hcb/pull/5529#discussion_r1525724198

check for other cases like this

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
albertchae: Hmm that's odd, I would have thought https://github.com/hackclub/hcb/blob/1f68fb08c66907b59d8792a889595fd7ca5d80c3/app/services/pending_transaction_engine/pending_transaction/all.rb#L30-L32 would handle this. I wonder if the `pt` here isn't coming from the same objects in memory as `@canonical_pending_transactions` 🤔

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
albertchae: in theory could we use `@event` by itself without the `||`? When would `@event` not be set here?

albertchae (commented):
One thing I'd recommend if you haven't done it already, try to curl `/hq` before and after these pages and use `diff` to see if anything is different in the HTML. It can help you identify if you accidentally changed behavior

app/models/hcb_code.rb
albertchae: This post covers it well https://www.speedshop.co/2019/01/10/three-activerecord-mistakes.html#any-exists-and-present, I think `any?` should cover most cases except if you really do need to check `exists?` live against the database instead of the in memory objects.

app/models/hcb_code.rb
```diff
@@ -317,7 +317,7 @@ def increase_check
   end
 
   def disbursement
-    @disbursement ||= Disbursement.find_by(id: hcb_i2) if disbursement?
+    @disbursement ||= Disbursement.includes(:card_grant).find_by(id: hcb_i2) if disbursement?
```
garyhtou: From my understanding, `find_by` always returns one record. So I'm unsure if using `includes` here makes a performance improvement. Can you double-check this?

If it's joining Disbursement and CardGrant, check for performance improvements

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -141,9 +142,9 @@
   <% end %>
 </tr>
 
-<% if Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && pt.local_hcb_code.event.present? %>
+<% if Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (@event || pt.local_hcb_code.event).present? %>
   <section class="modal modal--scroll modal--popover bg-snow" data-behavior="modal" role="dialog" id="transaction_details_<%= instance %>">
-    <%= modal_header(pt.local_hcb_code.pretty_title(show_event_name: defined?(show_event_name), show_amount: defined?(show_amount)), level: "h1", external_link: pt.local_hcb_code.url) %>
+    <%= modal_header(pt.local_hcb_code.pretty_title(show_event_name: defined?(show_event_name), show_amount: defined?(show_amount), event_name: @event.name, amount_cents: pt.amount), level: "h1", external_link: pt.local_hcb_code.url) %>
```
garyhtou: Adding a safe nav instead `@event` is not defined. I see that we're doing `@event || pt.local_hcb_code.event`. so there's a chance it's not defined

```suggestion
    <%= modal_header(pt.local_hcb_code.pretty_title(show_event_name: defined?(show_event_name), show_amount: defined?(show_amount), event_name: @event&.name, amount_cents: pt.amount), level: "h1", external_link: pt.local_hcb_code.url) %>
```

app/views/canonical_transactions/_canonical_transaction.html.erb
```diff
@@ -151,9 +152,9 @@
   <% end %>
 </tr>
 
-<% if Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && ct.local_hcb_code.event.present? %>
+<% if Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (@event || ct.local_hcb_code.event).present? %>
   <section class="modal modal--scroll modal--popover bg-snow" data-behavior="modal" role="dialog" id="transaction_details_<%= instance %>">
-    <%= modal_header(ct.local_hcb_code.pretty_title(show_event_name: defined?(show_event_name), show_amount: defined?(show_amount)), level: "h1", external_link: ct.local_hcb_code.url) %>
+    <%= modal_header(ct.local_hcb_code.pretty_title(show_event_name: defined?(show_event_name), show_amount: defined?(show_amount), event_name: @event.name, amount_cents: ct.amount), level: "h1", external_link: ct.local_hcb_code.url) %>
```
garyhtou:
```suggestion
    <%= modal_header(ct.local_hcb_code.pretty_title(show_event_name: defined?(show_event_name), show_amount: defined?(show_amount), event_name: @event&.name, amount_cents: ct.amount), level: "h1", external_link: ct.local_hcb_code.url) %>
```

app/models/hcb_code.rb
sampoder:
```suggestion
    @disbursement ||= Disbursement.find_by(id: hcb_i2) if disbursement?
```

garyhtou (approved):

garyhtou:
```
❯ curl "http://localhost:4000/hq" > hq-after.html
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  332k  100  332k    0     0  89599      0  0:00:03  0:00:03 --:--:-- 89594

❯ curl "http://localhost:4000/hq" > hq-before.html
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  332k  100  332k    0     0  66755      0  0:00:05  0:00:05 --:--:-- 87532
❯ diff hq-before.html hq-after.html
10c10
< <meta name="csrf-token" content="PmkxwuqcD5iYySmAZszLVVyAd4lw0DnCTgaizxTLahN8OFGlaIkJra8W8ikpQeFRhN7AzZ1lg0CwQDsU_OnQAg" />
---
> <meta name="csrf-token" content="nSRxALHiqKC5i1L49qKlHwtWtb0TaNRBQnMgQKSbX8o9slj8qf2IjuZ2k3cbdA0bfxrIgRCze_1g5icI3J_TYw" />
96c96
<     <a class="header-center text-decoration-none flex items-center tooltipped tooltipped--s" aria-label="You’re in development mode. Drop the tables for all I care." data-controller="analytics " data-analytics-event-name-value="Icon right-clicked" data-ar-usdz-href-value="/bank.usdz" data-action="contextmenu-&gt;menu#toggle contextmenu-&gt;analytics#track click@document-&gt;menu#close keydown@document-&gt;menu#keydown" data-menu-target="toggle" href="/">
---
>     <a class="header-center text-decoration-none flex items-center tooltipped tooltipped--s" aria-label="You’re in development mode. You pansy." data-controller="analytics " data-analytics-event-name-value="Icon right-clicked" data-ar-usdz-href-value="/bank.usdz" data-action="contextmenu-&gt;menu#toggle contextmenu-&gt;analytics#track click@document-&gt;menu#close keydown@document-&gt;menu#keydown" data-menu-target="toggle" href="/">
4877,4879c4877,4879
<         Build <em aria-label="ea73eec917b59366222ca0da5be20ba3c0c76a8e" class="tooltipped tooltipped--n tooltipped--lg">ea73eec-dirty</em>
<         from about 15 hours ago.
<       Loaded using 433 unique SQL queries (+ 155 cached).
---
>         Build <em aria-label="ada07a6436bbfc7189a70034c6e72ca92f4a9e5d" class="tooltipped tooltipped--n tooltipped--lg">ada07a6-dirty</em>
>         from 7 minutes ago.
>       Loaded using 136 unique SQL queries (+ 27 cached).
4889c4889
<   <script async nonce="" type="text/javascript" id="mini-profiler" src="/mini-profiler-resources/includes.js?v=90a68676a0c0d704b4438ca3f27d46c4" data-css-url="/mini-profiler-resources/includes.css?v=90a68676a0c0d704b4438ca3f27d46c4" data-version="90a68676a0c0d704b4438ca3f27d46c4" data-path="/mini-profiler-resources/" data-current-id="t3nnf4jzsx94ffqlf28t" data-ids="t3nnf4jzsx94ffqlf28t,uzeyft98m9dqbgq58206,z6dnn6ln4xnq5cn7aarq,ve82ghgt82lm7v34n52i,agx8x6mco56x9a4au2ww,bx1xw8dm2k1o6psqnd2c" data-horizontal-position="left" data-vertical-position="top" data-trivial="false" data-children="false" data-max-traces="20" data-controls="false" data-total-sql-count="false" data-authorized="true" data-toggle-shortcut="alt+p" data-start-hidden="false" data-collapse-results="true" data-html-container="body" data-hidden-custom-fields="" data-turbo-permanent="false"></script>
---
>   <script async nonce="" type="text/javascript" id="mini-profiler" src="/mini-profiler-resources/includes.js?v=90a68676a0c0d704b4438ca3f27d46c4" data-css-url="/mini-profiler-resources/includes.css?v=90a68676a0c0d704b4438ca3f27d46c4" data-version="90a68676a0c0d704b4438ca3f27d46c4" data-path="/mini-profiler-resources/" data-current-id="bx1xw8dm2k1o6psqnd2c" data-ids="bx1xw8dm2k1o6psqnd2c,uzeyft98m9dqbgq58206,z6dnn6ln4xnq5cn7aarq,ve82ghgt82lm7v34n52i,agx8x6mco56x9a4au2ww" data-horizontal-position="left" data-vertical-position="top" data-trivial="false" data-children="false" data-max-traces="20" data-controls="false" data-total-sql-count="false" data-authorized="true" data-toggle-shortcut="alt+p" data-start-hidden="false" data-collapse-results="true" data-html-container="body" data-hidden-custom-fields="" data-turbo-permanent="false"></script>
```

PR #5381, opened 2024-03-03
Admin: New organization onboarding settings page

rluodev:
In preparation for the upcoming redesign of the settings page by Sean to be paginated, we wanted to provide a way for admins to continue easily activating organizations without increasing the number of clicks. To this end, this PR adds an "Activate this Org" page in the navigation sidebar, which provides a page for editing basic organization admin settings, organization tags, and attaching organization formation documents (you can attach multiple files I just only showed one for simplicity). This page will only be accessible to admins, and only when an organization is in playground mode.

image

_^updated by Sarthak_

Screenshot 2024-03-02 at 8 32 40 PM

Screenshot 2024-03-02 at 8 32 49 PM

Screenshot 2024-03-02 at 8 33 23 PM

(ignore the first document i was testing)

~~**Note**: There is one known issue: Sometimes, the script I wrote for displaying the selected files does not work. I can't figure out how to fix this, but you can just reload the page and it'll (hopefully) work. if someone can tell me why this does not work I would very much appreciate it.~~

This is fixed - it turned out to be a turbo pages issue, so I just added a turbo:load event listener.


sarthaktexas:
great work! let's work on the ui next :)

rluodev:
sarthak requests:

- 3 drag and drop file fields
- rename a file if fiscal sponsorship contract in file name -> "Fiscal Sponsorship Contract"
- fs contract is mandatory, moa + mission adherance optional

app/policies/document_policy.rb
```diff
@@ -46,4 +46,12 @@ def fiscal_sponsorship_letter?
     !(record&.unapproved? || record&.pending?) && !record.demo_mode? && (record.users.include?(user) || user.admin?)
   end
 
+  def activation_flow?
+    user.admin?
```
sarthaktexas:
```suggestion
    user&.admin?
```

app/policies/document_policy.rb
```diff
@@ -46,4 +46,12 @@ def fiscal_sponsorship_letter?
...
+  end
+
+  def activate?
+    user.admin?
```
sarthaktexas:
```suggestion
    user&.admin?
```

app/policies/event_policy.rb
```diff
@@ -165,6 +165,14 @@ def validate_slug?
     user_or_admin
   end
 
+  def activation_flow?
+    user.admin? && record.demo_mode?
```
sarthaktexas:
```suggestion
    user&.admin? && record.demo_mode?
```

app/policies/event_policy.rb
```diff
@@ -165,6 +165,14 @@ def validate_slug?
...
+  end
+
+  def activate?
+    user.admin? && record.demo_mode?
```
sarthaktexas:
```suggestion
    user&.admin? && record.demo_mode?
```

sarthaktexas (dismissed):

app/views/events/_nav.html.erb
```diff
@@ -106,6 +106,15 @@
         tooltip: "View legal documents, financial statements, and FAQ",
         icon: "docs",
         selected: local_assigns[:selected] == :documentation %>
+      <% if current_user.admin? && @event.demo_mode? %>
```
garyhtou:
```suggestion
      <% if policy(@event).activation_flow? %>
```

app/views/events/activation_flow.html.erb
```diff
@@ -0,0 +1,127 @@
...
+          <% end %>
+          <%= form_with url: event_event_tags_path(@event.id), data: { turbo: true } do |form| %>
+            <%= form.hidden_field :event_id, value: @event.id %>
+            <%= form.text_field :name, data: { "behavior" => @event.event_tags.none? ? "autofocus" : "" }, class: "menu__input", placeholder: "+ Create tag", autocomplete: "off", required: true %>
+          <% end %>
+        </div>
+      </div>
+    </div>
+  </div>
+<% end %>
```
garyhtou: It looks like this was copied from the Event settings page. Can we make them both share a partial? We don't want to copy code/have duplicated code in multiple places

app/views/events/activation_flow.html.erb
```diff
@@ -0,0 +1,127 @@
...
+        <%= form.radio_button :sponsorship_fee, 0.00 %>
+        <%= form.label :sponsorship_fee, value: 0.00, class: "justify-center h-fit-content", style: "line-height: 1;" do %>
+          <strong>0% (waived)</strong>
+        <% end %>
+      </div>
+    </div>
+
+    <div class="field pr2">
+      <%= form.label :country %>
+      <%= form.collection_select :country, Event.countries.map{ |c| [c.first, get_country_name_from_alpha2(c.first)] }, :first, :last, { include_blank: "Select a country" } %>
```
garyhtou: We want to require that this a country is selected

```suggestion
      <%= form.collection_select :country, Event.countries.map{ |c| [c.first, get_country_name_from_alpha2(c.first)] }, :first, :last, { prompt: "Select a country" }, { required: true } %>
```

app/views/events/activation_flow.html.erb
```diff
@@ -0,0 +1,127 @@
...
+    </div>
+
+    <div class="field pr2">
+      <%= form.label :category %>
+      <%= form.collection_select :category, Event.categories, :first, :first, { include_blank: "Select a category" } %>
```
garyhtou: we want this required
```suggestion
      <%= form.collection_select :category, Event.categories, :first, :first, { include_blank: "Select a category" }, { required: true } %>
```

app/views/events/activation_flow.html.erb
```diff
@@ -0,0 +1,127 @@
...
+    </div>
+
+    <div class="field pr2">
+      <%= form.label :point_of_contact_id, "Point of contact" %>
+      <%= form.collection_select :point_of_contact_id, User.admin.all, :id, :email, selected: @current_user.email %>
```
garyhtou: Double check that this works. You might need to change `@current_user.email` to `current_user.id`

app/views/events/activation_flow.html.erb
```diff
@@ -0,0 +1,127 @@
...
+            for (var i = 0; i < fileInput.files.length; i++) {
+              var listItem = document.createElement("div");
+              listItem.innerText = fileInput.files[i].name;
+              listItem.classList.add("card", "mb1", "mt1", "h4", "pr6");
+              output.appendChild(listItem);
+            }
+          });
+        });
+      </script>
+    </div>
```
garyhtou: Look at https://github.com/hackclub/bank/blob/7bd88569c9424d34da61f7f5b49896cea0beb06f/app/views/receipts/_form_v3.html.erb#L4-L21 to see how to implement a cleaner upload system with dropzone

app/controllers/events_controller.rb
```diff
@@ -732,6 +732,36 @@ def validate_slug
...
+  def activation_flow
+    authorize @event
+  end
+
+  def activate
+    authorize @event
+    puts "abcdefg"
+    puts event_params[:files][1]
+
+    event_params[:files].drop(1).each do |file|
```
garyhtou: Why are we dropping the first file?

app/controllers/events_controller.rb
```diff
@@ -732,6 +732,36 @@ def validate_slug
...
+      @document = Document.new(user: current_user, event_id: @event.id, name: file.original_filename, file:)
```
garyhtou: It seems this doesn't need to be an instance variable. we can just use `document`

app/controllers/events_controller.rb
```diff
@@ -732,6 +732,36 @@ def validate_slug
...
+
+      authorize @document
+
+      unless @document.save
+        render :new, status: :unprocessable_entity
```
garyhtou: I think we want to render `:activation_flow` instead of `:new`. New will show the form for creating a org

app/controllers/events_controller.rb
```diff
@@ -732,6 +732,36 @@ def validate_slug
...
+      end
+    end
+
+    @event.update(event_params.except(:files))
+    @event.update(demo_mode: false)
+
+    if @event.save
```
garyhtou: Calling `update` already saves its. Also, we can merge the two update calls together

```suggestion
    if @event.update(event_params.except(:files), demo_mode: false)
```

app/controllers/events_controller.rb
```diff
@@ -771,11 +801,12 @@ def event_params
       :website,
       :background_image,
       :stripe_card_shipping_type,
+      files: [],
```
garyhtou: This may affect other actions using `event_params`. If the `files` is only being used for `activate`, then I suggest not adding it here, and directly accessing using `params[:files]`

app/controllers/events_controller.rb
```diff
@@ -732,6 +732,36 @@ def validate_slug
...
   end
 
+  def activation_flow
+    authorize @event
+  end
+
+  def activate
+    authorize @event
+    puts "abcdefg"
+    puts event_params[:files][1]
```
garyhtou: Let's drop these prints

app/controllers/events_controller.rb
```diff
@@ -732,6 +732,36 @@ def validate_slug
...
+
+      authorize @document
+
+      unless @document.save
+        render :new, status: :unprocessable_entity
+      end
+    end
+
+    @event.update(event_params.except(:files))
+    @event.update(demo_mode: false)
```
garyhtou: we can combine these two updates into one update. and we should likely be using the ! (bang) method

```suggestion
    @event.update!(event_params.except(:files), demo_mode: false)
```

app/views/events/activation_flow.html.erb
garyhtou: Double check that it requires you to select an option

PR #5382, opened 2024-03-03
Reply to Comments via Email

sampoder:
Life is hard... nowadays there are so many social media apps... Slack, iMessage, WhatsApp, Instagram, BONK, (New) Expensify, and (of course) HCB... responding to messages on all these different platforms is tough. We like our users, so we won't put them through the hassle of logging in if they want to quickly respond to a comment on their phone. Instead, we'll just let them email us their reply. And we'll handle the rest!

I based this off of GitHub's implementation, a couple of notes:

- There are no email responses to indicate success etc. GitHub does this and I feel like it keeps the thread clean / natural.
- Currently we rely on comment IDs to identify commentable. If a comment is deleted, your response won't be matched to a commentable. I'm alright with this don't currently support deleting comments in the UI. If anyone has any alternative approaches, let me know. I wanted something that was flexible for every commentable.


app/mailboxes/comment_mailbox.rb
```diff
@@ -0,0 +1,58 @@
...
+class CommentMailbox < ApplicationMailbox
+  # mail --> Mail object, this actual email
+  # inbound_email => ActionMailbox::InboundEmail record --> the active storage record
+
+  include Pundit::Authorization
+  include HasAttachments
+
+  before_processing :set_user
+  before_processing :set_comment
+  before_processing :set_commentable
```
garyhtou: this might work??

```suggestion
  before_processing do
    set_attachments(include_body: false)
  end
```

app/mailboxes/comment_mailbox.rb
```diff
@@ -0,0 +1,58 @@
...
+  def set_commentable
+    @commentable = @comment.commentable
+  end
+
+  def set_user
+    @user = User.find_by(email: mail.from[0].downcase)
+  end
+
+  def ensure_permissions?
+    return true if @comment.nil?
```
garyhtou:
```suggestion
    return false if @comment.nil?
```

app/mailboxes/comment_mailbox.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    authorize @comment, :new?
+
+  rescue Pundit::NotAuthorizedError
+    false
```
garyhtou:
```suggestion
    Pundit.policy(@user, @comment).new?
```

app/mailboxes/comment_mailbox.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def pundit_user
+    @user
+  end
+
```
garyhtou:
```suggestion
```

garyhtou (dismissed):
🪄

PR #5390, opened 2024-03-05
Use Loofah to parse text from HTML emails for email commands

sampoder:
Fixes https://github.com/hackclub/hcb/issues/5389, problem description in https://hackclub.slack.com/archives/C047Y01MHJQ/p1709574806372209


app/mailboxes/hcb_code_mailbox.rb
```diff
@@ -45,7 +45,7 @@ def process
   private
 
   def set_commands
-    content = text || body || html
+    content = text || Loofah.html5_fragment(html).to_text || body
```
garyhtou: what type of content is `body`? (HTML?, text?, anything?)

sampoder: It's the equivalent to the contents of an `.eml` file afaik. For plain text emails, it's just plain text

northeastprince:
If we're using the library directly now we should put it in the Gemfile. Also, is there no other simple way to do this without it?

sampoder:
This feels fairly simple to me; there was nothing built in that replicated JavaScript's innerText implementation

northeastprince:
I agree the API is great for our use case, but if there's a way to do it with comparable effort without relying on another dependency, I would [heavily prefer that](https://www.jorgemanrubia.com/2018/03/15/mind-your-dependencies). That said, it's a utility library which are pretty easy to switch out if needed, so it's not a huge deal.

garyhtou (commented):
All good with using Loofah!

The merge wand is yours 🪄! Please test & merge!

PR #5397, opened 2024-03-05
Don't Fallback for Amazon SES Emails

sampoder:
https://hackclub.slack.com/archives/C026RKHLPNJ/p1709661775585849


garyhtou:
@sampoder I haven't taken a close look at this, but I'm fairly certain the emails are due to Hackathons Backend and not HCB

sampoder:
Oh interesting, yeah you're right

sampoder:
Wait we actually probably still want this for HCB

garyhtou:
I'm a bit confused. are these SES emails being directed to hcb@hackclub.com by ActionMailbox? Or, are they being sent directly to hcb@hackclub.com (or hackathon@hackclub.com) by SES?

sampoder:
Not the current ones causing issues, but if one of our email bounces (that we sent from ActionMailer) it would bounce to `@hcb.hackclub.com`? Idk maybe I'm missing smth and Rails automatically handles those

PR #5396, opened 2024-03-05
Reserve `mobile`

sampoder:
https://hackclub.slack.com/archives/C047Y01MHJQ/p1709651260078989

PR #5398, opened 2024-03-05
Put reserved slugs on their own lines

malted:
Simplifies diffs.

Suggested by @garyhtou [here](https://github.com/hackclub/hcb/pull/5396#pullrequestreview-1917767173).


garyhtou (commented):
Thanks! Can you convert it into the percent string syntax?

```ruby
config.reserved_words = %w[
  new
  edit
  index
]
```

garyhtou:
(vscode's multi cursor or vim is a great way to make those edits!)

malted (commented):
Convert to percent string syntax

malted:
Done!

garyhtou (approved):
Great! ty

PR #5399, opened 2024-03-05
Admin: Alphabetizes countries with US and Canada on top

sarthaktexas:
@hackliv requested it.
Screenshot 2024-03-05 at 12 46 51 PM


garyhtou (changes requested):
oops, can we actually define the method in `CountryEnumable`?

https://github.com/hackclub/bank/blob/7d61a562d550febabaaecef2b994713053392450/app/models/concerns/country_enumable.rb#L3

garyhtou:
Please also link the issue!

garyhtou (dismissed):

garyhtou (approved):

PR #5400, opened 2024-03-06
Add `LetterOpener` quicklink to HCB logo menu

malted:
Also adds menu divider titles

image


sampoder (approved):
Thank you! This is really helpful

app/views/application/_header.html.erb
```diff
@@ -57,9 +57,16 @@
       <div class="menu__divider"></div>
       <a href="/" target="_blank">Open link in new tab</a>
       <% if current_user&.admin_override_pretend? %>
-        <div class="menu__divider"></div>
+        <div class="menu__divider" data-label="Admin"></div>
```
sarthaktexas: what does the value of `data-label` do here?

sarthaktexas (changes requested):

app/views/application/_header.html.erb
malted: tl;dr

`&::after { content: attr(data-label) }`

sarthaktexas (approved):

PR #5405, opened 2024-03-07
Handle Comments w/ Files in Mailers

sampoder:
Fixes https://github.com/hackclub/hcb/issues/5403. Implemented as described in the issue.


app/views/comment_mailer/hcb_code/notification.html.erb
```diff
@@ -1,7 +1,17 @@
 <% hcb_code = @commentable %>
 
-<%= @comment.user.name %> has left a comment on <i><%= hcb_code.memo %></i> in <i><%= hcb_code.event.name %></i>:
+<% if @comment.content? %>
 
-<blockquote><%== MarkdownService.instance.renderer.render @comment.content %></blockquote>
+  <%= @comment.user.name %> has left a comment on <%= "and attached a file to" if @comment.file %> <i><%= hcb_code.memo %></i> in <i><%= hcb_code.event.name %></i>:
```
garyhtou: There is a grammar issue

```diff
- Gary Tou has left a comment on and attached a file to _blah_
+ Gary Tou has left a comment and attached a file to _blah_
```

app/mailers/comment_mailer.rb
```diff
@@ -7,6 +7,12 @@ def notification
 
     return if @commentable.comment_recipients_for(@comment).empty?
 
+    return unless @comment.content || @comment.file
+
+    if @comment.file
+      attachments[@comment.file.filename.to_s] = @comment.file.download
+    end
```
garyhtou: I am very hesitant about attaching the files! It is common for people (especially operations) to attach files that include sensitive information.

For example, operations will commonly upload images of checks (which contain account/routing numbers). Also, sometimes W9s get uploaded (we discourage this). This means that those files are sent to EVERYONE's emails.

let's not attach files!

garyhtou: Handled here https://github.com/hackclub/hcb/pull/5667

PR #5407, opened 2024-03-07
Add a mailer for differing settled amount (eg. tips)

sampoder:
See https://github.com/hackclub/hcb/issues/5404. Now, I feel like I may have massively over complicated this?

What I'm doing is checking the `request_history` attribute on the Stripe Authorisation and seeing if at any point `amount_cents` changes. If that's the case, we send the mailer.

Is there an easier way to do that?


app/services/canonical_pending_transaction_service/settle.rb
```diff
@@ -18,6 +18,14 @@ def run!
           @canonical_transaction.custom_memo = @canonical_pending_transaction.custom_memo
           @canonical_transaction.save!
         end
+
+        if @canonical_pending_transaction.raw_pending_stripe_transaction
+          stripe_authorization_id = @canonical_pending_transaction.raw_pending_stripe_transaction.stripe_transaction_id
```
sampoder: Also, is this the right way to be pulling `stripe_authorization_id`s? I saw this code in another spot but Stripe transactions are different from Stripe Authorisations

app/views/canonical_pending_transaction_mailer/notify_settled.html.erb
```diff
@@ -0,0 +1,11 @@
...
+
+<% if @cpt.local_hcb_code.receipt_required? %>
+  <%= link_to "Upload your receipt for this charge (this link can used by anyone for two weeks)", @upload_url %>.
+
+  <% if @receipt_upload_feature %>
+    <p>
+      Alternatively, <em>reply to this email with an image or PDF receipt attached</em>. When responding with a receipt, include a line beginning with <em>@rename</em> to rename the transaction or with <em>@tag</em> to tag the transaction.
+    </p>
+  <% end %>
+<% end %>
```
sampoder: I wanted to keep this around just to encourage people to upload more receipts!

app/models/stripe_authorization.rb
```diff
@@ -159,6 +159,10 @@ def deleted_at
     nil
   end
 
+  def amount_cents_changed?
```
garyhtou: this method name override an existing one provided by rails
https://api.rubyonrails.org/classes/ActiveModel/Dirty.html#method-i-2A_changed-3F

app/views/canonical_pending_transaction_mailer/notify_settled.html.erb
garyhtou: Can we make this a shared partial with the other mailer that has this copy? It would probably need to take in two arguments: `hcb_code` and `user`

app/models/stripe_authorization.rb
```diff
@@ -159,6 +159,10 @@ def deleted_at
...
+    stripe_obj.merchant_data.request_history.chunk { |req| req["amount_cents"] }.any? { |_, group| group.size > 1 }
```
garyhtou: dang, `chunk` is such a cool method! 😎🆒

app/views/canonical_pending_transaction_mailer/notify_settled.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= @cpt.memo %> has settled at <%= render_money @cpt.amount_cents %>; this was a charge on your card ending in <%= @cpt.stripe_card.last4 %> that previously had a different amount.
```
sampoder: This won't work! @cpt.amount_cents is the old amount.

PR #5408, opened 2024-03-08
Make card details copy-on-click-able

malted:
does what it says on the tin

oh also refactoring


sarthaktexas:
# go to bed.

app/views/stripe_cards/_details.html.erb
```diff
@@ -32,61 +32,57 @@
...
-    <% if stripe_card.virtual? %>
 
-        <% if stripe_card.user == current_user %>
-          <div class="fs-mask">
-            <strong>
-              Card number
-            </strong>
-            <div class="flex justify-between mono items-center">
+    <%# Card number %>
+    <% if stripe_card.virtual? %>
```
sarthaktexas: you check if it's virtual here, but not if the stripe_card is theirs.

app/views/stripe_cards/_details.html.erb
```diff
@@ -32,61 +32,57 @@
...
+      <div class="fs-mask">
+        <strong>Card number</strong>
+        <div class="flex justify-between mono items-center">
+          <% if stripe_card.user == current_user %>
```
sarthaktexas: here you check if the stripe_card is theirs.

app/views/stripe_cards/_details.html.erb
```diff
@@ -32,61 +32,57 @@
...
+          <% end %>
+        </div>
+      </div>
+    <% end %>
 
-        <% if stripe_card.user == current_user %>
-        <div class="fs-mask mono">
-          <strong>CVC</strong>
+    <%# CVV %>
+    <% if stripe_card.user == current_user && stripe_card.virtual? %>
```
sarthaktexas: here you check both. why the inconsistency?

app/views/stripe_cards/_details.html.erb
```diff
@@ -32,61 +32,57 @@
...
-        </div>
       </div>
     <% end %>
+
+    <div class="fs-mask">
+      <strong>Expiration date</strong>
+      <%= render_exp_date stripe_card %>
+    </div>
+
+    <div>
```
sarthaktexas: can we fs-mask this too? or are we limiting the information that we mask to only sensitive content?

sarthaktexas (changes requested):
it appears there is inconsistent logic across blocks of code.

app/views/stripe_cards/_details.html.erb
sarthaktexas: WRONG try again 🤡🤡🤡

malted: stop being a meanie

sarthaktexas (approved):
code looks good. i see two tooltips on the card number with diff directions, i can't test if that causes a problem or not but i trust.

malted:
The different directions are intended!

app/views/stripe_cards/_details.html.erb
```diff
@@ -32,61 +32,60 @@
...
-                    <%= inline_icon "view", class: "align-middle" %>
-                  <% end %>
-                </div>
-              <% end %>
-            </div>
+    <div class="fs-mask">
+      <strong>Card number</strong>
+      <div class="flex justify-between mono items-center">
+        <% if stripe_card.user == current_user && stripe_card.virtual? %>
+          <%= copy_to_clipboard(stripe_card.full_card_number, class: "w-fit", tooltip_direction: "e") do %>
```
garyhtou: @malted important feedback:

We should not expose the full card number to the front end unless it was explicitly asked for.

This is because we're tracking when card numbers are accessed
https://github.com/hackclub/bank/blob/34aecab447b76773f0e91fc128350c8530cb55af/app/controllers/stripe_cards_controller.rb#L54-L56

Please make it so that this tooltip does not appear until `@show_card_details` is true

cc @sarthaktexas

app/views/stripe_cards/_details.html.erb
```diff
@@ -32,61 +32,60 @@
...
-          <strong>CVC</strong>
-          <% if @show_card_details %>
-            <%= stripe_card.cvc %>
-          <% else %>
-            <%= stripe_card.hidden_cvc %>
+    <div class="fs-mask">
+      <strong>CVC</strong>
+      <div class="flex justify-between mono items-center">
+        <% if stripe_card.user == current_user && stripe_card.virtual? %>
+          <%= copy_to_clipboard(stripe_card.cvc, class: "w-fit", tooltip_direction: "e") do %>
```
garyhtou: Same thing applies here

app/views/stripe_cards/_details.html.erb
garyhtou: A more UX-friendly solution could be to keep this functionality by making it fetch the card number via a request then copy it to the clipboard

PR #5413, opened 2024-03-09
Use `recurring_donation.message` if it exists

sampoder:
I believe this fixes https://github.com/hackclub/hcb/issues/5410


garyhtou (changes requested):
This fix would send the message for every single donation associated with that recurring donation.

We only want to send it for the first donation.

There exists code that handles this logic. Maybe it's not working correctly?
https://github.com/hackclub/bank/blob/808ef12b6760e6b21f396b46d59a3246cab89b39/app/services/recurring_donation_service/handle_invoice_paid.rb#L22-L24

PR #5414, opened 2024-03-09
Run `:send_payment_notification_if_needed` on create

sampoder:
And only run it if status == "succeeded". This is an edge case for recurring donations where we seem to create the `Donation` record when the Stripe status would already be "succeeded".  And therefore, we never send them a email: https://github.com/hackclub/hcb/issues/5410.

Fixes https://github.com/hackclub/hcb/issues/5410.


cjdenio (dismissed):
This will send the notification after the user initially puts in their name/email, _not_ after they submit their card details

sampoder:
Does it though? I just added a test for that

cjdenio:
Sorry! I was half asleep when I wrote that

garyhtou (approved):

PR #5419, opened 2024-03-09
Fix `organizer_signed_in?`

malted:


garyhtou (commented):
I think removing the memorization may slow down all page loads

malted:
@garyhtou memoised. Does this look better?

app/helpers/sessions_helper.rb
```diff
@@ -64,33 +64,17 @@ def current_user=(user)
...
 
   def organizer_signed_in?(event = @event, as: :member)
-    run = -> do
-      next true if admin_signed_in?
-      next false unless signed_in? && event.present?
+    memoization_key = [event, current_user, as]
+    @organizer_signed_in ||= {}
+    @organizer_signed_in[memoization_key] ||= begin
+      return true if admin_signed_in?
+      return false unless signed_in?
```
garyhtou: Returning inside a block will return from the entire method, not just the block. We likely want to use `next` here.

garyhtou: I'm not sure if it will store it inside the cache hash

app/helpers/sessions_helper.rb
```diff
@@ -64,33 +64,17 @@ def current_user=(user)
   end
 
   def organizer_signed_in?(event = @event, as: :member)
-    run = -> do
-      next true if admin_signed_in?
-      next false unless signed_in? && event.present?
+    memoization_key = [event, current_user, as]
+    @organizer_signed_in ||= {}
+    @organizer_signed_in[memoization_key] ||= begin
```
garyhtou: When the block returns a falsey value (such as `false`), it will likely ignore the memorization in future calls.

[HCB TRIVIA - Ruby, Memoization using Or-Equals Operator.pdf](https://github.com/hackclub/hcb/files/14641141/HCB.TRIVIA.-.Ruby.Memoization.using.Or-Equals.Operator.pdf)

app/helpers/sessions_helper.rb
```diff
@@ -64,33 +64,17 @@ def current_user=(user)
   end
 
   def organizer_signed_in?(event = @event, as: :member)
-    run = -> do
-      next true if admin_signed_in?
-      next false unless signed_in? && event.present?
+    memoization_key = [event, current_user, as]
```
garyhtou: I believe we don't need `current_user` in the memoization key, because `organizer_signed_in?` is always called with the perspective of the current user

garyhtou (changes requested):

PR #5422, opened 2024-03-10
Auto-attach receipts SMS-d in the last five minutes

sampoder:
Will save folks some time!


YodaLightsabr (approved):
LGTM! I haven't tested it though.

sampoder:
I'm going to merge this & test on the way home today

garyhtou:
Note from Mentors call:

We likely want to move this into the service (or model) so that it can also work for email uploads. When moving this code, make sure we filter by `upload_method`. For example, uploads via receipt bin or transaction pages likely shouldn't have this 5 minute feature

PR #5426, opened 2024-03-10
Fix copy for anon donations in emails

polypixeldev:
## Summary of the problem

Copy seemed a little odd when I got the email for an anonymous donation.

## Describe your changes

Added a check for if the donation is anonymous in the appropriate place - thanks @sarthaktexas for the suggestion for the copy!


sarthaktexas (approved):

garyhtou:
Thanks for catching this!

PR #5429, opened 2024-03-10
UI: Adds context menus to dock items partials

sarthaktexas:
Here's a context menu:
Screenshot 2024-03-10 at 1 23 33 AM


app/helpers/events_helper.rb
```diff
@@ -3,19 +3,25 @@
...
+          if async_badge
+            inline_icon(icon, size: 32) +
+            turbo_frame_tag(async_badge, src: async_badge, data: { controller: "cached-frame", action: "turbo:frame-render->cached-frame#cache" })
+          else
+            inline_icon(icon, size: 32)
+          end
+        end) + content_tag(:span, name.html_safe, class: "line-height-3") +
+          content_tag(:div, "", data: { "menu-target": "content" }, 'data-menu-target': "content", class: "menu__content menu__content--2 menu__content--compact") do
+            block.call if block
+          end
```
garyhtou: Maybe we don't render this `div` if there's no block

garyhtou (changes requested):
I've got an odd horizontal scroll
image

sampoder:
Is this still planned @sarthaktexas?

PR #5431, opened 2024-03-10
[User Roles] Don't let people change their own roles

malted:
image

Admins can still change their own roles because nowt really changes


sampoder:
Left Side is my favourite manager!

garyhtou:
What if we hide the dropdown and the `v` icon? that would remove the need for a tooltip

malted:
Done!

garyhtou (approved):
policy is good! just one comment about renaming

PR #5434, opened 2024-03-10
Reimbursements: Integrate with transaction engine

garyhtou:
## Summary of the problem

we used to print money out of thin air by creating `CanonicalTransaction`s without importing them from Column.

## Describe your changes

we now create `CanonicalPendingTransaction`s and import the `CanonicalTransaction`s when they arrive.

`Reimbursement::ExpensePayout`s have fronted `CanonicalPendingTransaction`. These are settled when the CanonicalTransaction from the Column book transaction arrives. Those CTs are mapped into the CPT/ExpensePayout's HcbCode using an HcbCode short code as the memo.

`Reimbursement::PayoutHolding` don't NOT have pending transactions


sampoder:
I'm going to merge this to prevent conflicts, Caleb if you have any thoughts, can you drop them in the main PR?

PR #5461, opened 2024-03-11
Adds blankslate for expense list + disables reimbursement button on pg mode

sarthaktexas:


app/views/reimbursement/reports/_create_form.html.erb
```diff
@@ -29,6 +29,8 @@
       <%= form.label :event_id, "Organization", class: "mt1 bold" %>
       <%= form.select :event_id, current_user.events.not_demo_mode.pluck(:name, :id) %>
     <% end %>
-    <%= form.submit "Submit", class: "mt2" %>
```
sampoder: We should do this at a policy level

app/views/reimbursement/reports/show.html.erb
sarthaktexas, line 1: most of this is fixing the formatting

PR #5475, opened 2024-03-12
Fix disbursement policy checks

malted:
Comments left in as I'm coming back to this after fixing


maxwofford (approved):
Sitting down with @malted to review. This is an approve with the understanding that this is getting refactored (and comments updated) in an upcoming PR.

PR #5481, opened 2024-03-12
Create has_activity_log concern

maxwofford:
This is a work in progress. Once this is ready, you can do this:

```ruby
# Report.rb
include HasActivityLog

# ReportsController.rb
@history = @report.audit_log(
  public_fields: %w[aasm_state name maximum_amount_cents event_id],
  include_comments: true
)
 

# reports/show.html.erb
<%= render partial: "activity_log/item", collection: @history, as: :item %>
```

And you'll get a github-activity like partial rendered on your page like this:

Screenshot 2024-03-11 at 22 48 52

Other perks:
- It'll also be an allow-list of changes, so you don't accidentally show fields to a user that you didn't mean to reveal.
- The heavy lifting is done in activerecord, and doesn't use `is_a?` so in theory this can handle long changelogs much faster.

Currently, it's missing a couple things to reach feature parity:
- Associations (showing an event name like "HQ" instead of just ID like "183"). I know this is possible because the live implementation supports this. Don't quite know how it works yet though
- Enriched icons and colors. The current setup looks really pretty because @sampoder and @YodaLightsabr wrote a really nice looking set of icons and colors for specific events.


garyhtou:
@sampoder when you work on the new Activity Log feature, can you consider this as a use case?

PR #5482, opened 2024-03-12
Fix 500 on disbursement new

maxwofford:
Fixes a typo preventing disbursements#new from loading for non-admin users


malted:
Can't believe I missed this, cheers!

PR #5483, opened 2024-03-12
Hide user roles in feature flag

maxwofford:
Before, even without the feature flag enabled, parts of the new manager feature were leaking:

Screenshot 2024-03-12 at 10 03 04

Now they're all gone:

Screenshot 2024-03-12 at 10 03 13


malted:
Thank you!

malted (approved):

PR #5485, opened 2024-03-12
Refactor disbursement create policy

malted:
Adds manager constraint to `create?`
Refactor per https://github.com/hackclub/hcb/pull/5475#pullrequestreview-1929882430


malted:
In another PR, I will disable the disbursement submit button for those who can not `EventPolicy::new_transfer?`

app/policies/disbursement_policy.rb
```diff
@@ -14,12 +14,14 @@ def new?
...
   end
 
   def create?
-    user&.admin? || (record.destination_event.users.include?(user) && record.source_event.users.include?(user))
-    # user.admin? || user_associated_with_events? && user_who_can_transfer?
+    user&.admin? || (
+      record.destination_event.users.include?(user) &&
+      record.source_event.users.include?(user) &&
+      EventPolicy.new(user, record.source_event).new_transfer?
+    )
```
garyhtou: I believe `record.source_event.users.include?(user)` and `EventPolicy.new(user, record.source_event).new_transfer?` check for the same thing.

I wonder if we can rewrite this as:

```suggestion
    user&.admin? || (
      record.destination_event.users.include?(user) &&
      Pundit.policy(user, record.source_event).new_transfer?
    )
```

In other words, a user can create a disbursement if that can initiate transfers from the source event, and they are a member of the destination event.

garyhtou (approved):

garyhtou:
This is just a refactor — it's functionally the same

PR #5486, opened 2024-03-12
[User Roles] Disable Disbursement form fields for members

malted:
image


app/views/disbursements/_form.html.erb
```diff
@@ -49,18 +53,20 @@
...
+    <%= form.text_field :name, placeholder: "Donating extra funds to another organization", maxlength: 60, required: true, disabled: %>
     <span class="muted">This is to help HCB keep record of our transactions. It will be used as the memo on your transaction list.</span>
   </div>
 
-  <div class="actions inline-block">
-    <%= form.submit "Make transfer", class: "btn", disabled: @source_event&.demo_mode? %>
-  </div>
+  <%= content_tag :div,
+    class: "actions inline-block #{"tooltipped tooltipped--e" if disabled}",
+    aria: { label: playground_mode ? "Disbursements are disabled in Playground mode" : !user_can_transfer ? "Only managers can make disbursements!" : nil } do %>
```
cjdenio:
```suggestion
    aria: { label: playground_mode ? "Transfers are disabled in Playground mode" : !user_can_transfer ? "Only managers can make transfers!" : nil } do %>
```

sarthaktexas:
i just KNOW gary is going to say move this logic into the controller

garyhtou:
> i just KNOW gary is going to say move this logic into the controller

LOL, you're starting to get the hang of it!

<br/><br/>

Close, i'd suggest moving the `@source_event&.demo_mode?` check into the policy, and then just using
```ruby
!policy(@source_event).new_transfer?
```
to control the disabled attribute.

However, I'm cool with how it currently is; I'm focusing on the priority of fixing bugs and getting this out to users ASAP

app/views/disbursements/_form.html.erb
```diff
@@ -49,18 +51,20 @@
...
   </div>
 
-  <div class="actions inline-block">
-    <%= form.submit "Make transfer", class: "btn", disabled: @source_event&.demo_mode? %>
-  </div>
+  <%= content_tag :div,
+    class: "actions inline-block #{"tooltipped tooltipped--e" if disabled}",
+    aria: { label: playground_mode ? "Transfers are disabled in Playground mode" : !user_can_transfer ? "Only managers can make transfers!" : nil } do %>
+    <%= form.submit "Make transfer", class: "btn", disabled: %>
+  <% end %>
```
garyhtou: i'm not sure if i understand why this needs to be a `content_tag`. i think w can use a normal `div`

garyhtou (approved):

PR #5492, opened 2024-03-13
Show the billing address change confirmation flash on submit instead of on reload

malted:
Previously nothing would happen, until you reloaded the page, where the flash would then appear. Now it appears immediately.


garyhtou (approved):
cool!

@cjdenio, is there a better way to go about this without using turbo streams for flashes?

PR #5494, opened 2024-03-13
[Reimbursements] Add the word reimbursement close to the start of the notif email

malted:
image

Not sure if this is the optimal change. Thoughts?


sarthaktexas (dismissed):
I like this way more

garyhtou:
Please add to email subject too!

garyhtou:
I'd also suggest moving the name of the report to the end of the subject in case it's really long

sarthaktexas (approved):

sarthaktexas (dismissed):
We need to change the format of all of them.

garyhtou:
Yes please!

sampoder (approved):

PR #5507, opened 2024-03-14
Add custom memos for reimbursement-related transactions

sampoder:
* Closes #5506
* Closes #5505


app/models/ach_transfer.rb
```diff
@@ -78,6 +78,7 @@ class AchTransfer < ApplicationRecord
   has_one :grant, required: false
   has_one :raw_pending_outgoing_ach_transaction, foreign_key: :ach_transaction_id
   has_one :canonical_pending_transaction, through: :raw_pending_outgoing_ach_transaction
+  has_one :reimbursement_expense_payout, class_name: "Reimbursement::ExpensePayout", inverse_of: :ach_transfer, required: false
```
garyhtou: i belive `optional: true` is now preferred instead of `required: false`
```suggestion
  has_one :reimbursement_expense_payout, class_name: "Reimbursement::ExpensePayout", inverse_of: :ach_transfer, optional: true
```

app/models/hcb_code.rb
```diff
@@ -348,6 +348,13 @@ def reimbursement_payout_holding?
     hcb_i1 == ::TransactionGroupingEngine::Calculate::HcbCode::PAYOUT_HOLDING_CODE
   end
 
+  def reimbursement_payout_transfer?
+    return true if increase_check? && increase_check.reimbursement_payout_holding.present?
+    return true if ach_transfer? && ach_transfer.reimbursement_payout_holding.present?
+
+    false
+  end
+
```
garyhtou: Can we define a `reimbursement_payout_holding` method?

```suggestion
  def reimbursement_payout_transfer?
    reimbursement_payout_holding.present?
  end

  def reimbursement_payout_holding
    if increase_check?
      increase_check.reimbursement_payout_holding
    elsif ach_transfer?
      ach_transfer.reimbursement_payout_holding
    else
      nil
    end
  end
```

Then later, in `reimbursement_payout_transfer_memo`, we can do the following to get the ID
```ruby
reimbursement_payout_transfer_memo.report.id
```

app/models/increase_check.rb
```diff
@@ -54,6 +54,7 @@ class IncreaseCheck < ApplicationRecord
 
   has_one :canonical_pending_transaction
   has_one :grant, required: false
+  has_one :reimbursement_expense_payout, class_name: "Reimbursement::ExpensePayout", inverse_of: :increase_check, required: false
```
```suggestion
  has_one :reimbursement_expense_payout, class_name: "Reimbursement::ExpensePayout", inverse_of: :increase_check, optional: true
```

app/models/ach_transfer.rb
sampoder: This will cause an error, see https://github.com/hackclub/hcb/pull/5507/commits/55fab0e788486a07e720c2b564e0953cd8dce5b5 's tests

sampoder:
Sorry I've bundled a couple of thing into this PR. I needed the `has_one`.

app/models/hcb_code.rb
```diff
@@ -348,6 +348,20 @@ def reimbursement_payout_holding?
...
+  def reimbursement_payout_holding
+    if increase_check?
+      increase_check.reimbursement_payout_holding
+    elsif ach_transfer?
+      ach_transfer.reimbursement_payout_holding
+    else
+      nil
+    end
+  end
```
YodaLightsabr:
```suggestion
    increase_check&.reimbursement_payout_holding || ach_transfer&.reimbursement_payout_holding || super
  end
```

It seems like we're redefining the `reimbursement_payout_holding` method. Would using save navigation work, and should we fallback to `super`?

cc @garyhtou

YodaLightsabr (dismissed):

app/models/ach_transfer.rb
garyhtou: This should be an association with reimbursement payout holding

app/models/hcb_code.rb
```diff
@@ -348,6 +348,20 @@ def reimbursement_payout_holding?
...
+
+  def reimbursement_payout_holding
+    if increase_check?
+      increase_check.reimbursement_payout_holding
+    elsif ach_transfer?
+      ach_transfer.reimbursement_payout_holding
+    else
+      nil
+    end
```
garyhtou: let's inline this code into `reimbursement_payout_transfer?`

app/models/hcb_code.rb
```diff
@@ -348,6 +348,20 @@ def reimbursement_payout_holding?
     hcb_i1 == ::TransactionGroupingEngine::Calculate::HcbCode::PAYOUT_HOLDING_CODE
   end
 
+  def reimbursement_payout_transfer?
+    reimbursement_payout_holding.present?
+  end
```
```suggestion
  def reimbursement_payout_transfer?
    reimbursement_payout_transfer&.reimbursement_payout_holding.present?
  end

  def reimbursement_payout_transfer
    if increase_check? && increase_check.reimbursement_payout_holding.present?
      increase_check
    elsif ach_transfer? && ach_transfer.reimbursement_payout_holding.present?
      ach_transfer
    else
      nil
    end
  end
```

app/models/hcb_code.rb
sampoder: Good catch, thanks! :D

app/models/hcb_code/memo.rb
```diff
@@ -22,6 +22,9 @@ def memo(event: nil)
         return ach_payment_memo if ach_payment?
         return grant_memo if grant?
         return outgoing_fee_reimbursement_memo if outgoing_fee_reimbursement?
+        return reimbursement_payout_holding_memo if reimbursement_payout_holding?
+        return reimbursement_expense_payout_memo if reimbursement_expense_payout?
+        return reimbursement_payout_transfer_memo if reimbursement_payout_transfer?
```
garyhtou: Move line 27 above ach and check. then add a comment explaining why

(payout transfer is also an ach, which would return early)

app/models/hcb_code/memo.rb
```diff
@@ -95,6 +98,18 @@ def outgoing_fee_reimbursement_memo
         "🗂️ Stripe fee reimbursements for week of #{ct.date.beginning_of_week.strftime("%-m/%-d")}"
       end
 
+      def reimbursement_payout_holding_memo
+        "Payout Holding for Reimbursement Report #{reimbursement_payout_holding.report.id}"
```
garyhtou:
```suggestion
        "Payout Holding for Reimbursement Report #{reimbursement_payout_holding.report.hashid}"
```

app/models/hcb_code/memo.rb
```diff
@@ -95,6 +98,18 @@ def outgoing_fee_reimbursement_memo
...
+      end
+
+      def reimbursement_expense_payout_memo
+        reimbursement_expense_payout.expense.memo
+      end
+
+      def reimbursement_payout_transfer_memo
+        return "Payout Transfer for Reimbursement Report #{reimbursement_payout_transfer_memo.report.id}"
```
garyhtou:
```suggestion
        return "Payout Transfer for Reimbursement Report #{reimbursement_payout_transfer_memo.report.hashid}"
```

app/models/ach_transfer.rb
garyhtou, line 1: 1. Let's add a `company_name` and `company_entry_description` column.
2. when creating an AchTransfer, optionally set those columns. If they're not set, fill in the columns with the existing logic.
3. Add a validation that newly created AchTransfer must have those columns filled
   - old ach transfer will have these columns as `NULL`
4. when calling `send_ach_transfer!`, use the columns

we can pull this into another PR

the reasoning behind this is to decouple AchTransfer from Reimbursements

PR #5508, opened 2024-03-14
Pundit: Simplify policies for admins

garyhtou:
## Summary of the problem

The Pundit files are kinda long. I found that there are patterns in our policies.
- With nearly all of our policy actions, we want admins to be able to perform them.
- There are a lot of actions that only admins can perform
- A lot of actions within the same policy have shared logic

## Describe your changes

This adds two helpers `permit_admins_to` and `only_admins_can` to simplify the policies. I'm not exactly sure if I love them and wanted to open this PR to get some thoughts before I continue updating the rest of the policies.

I'm basically building my own DSL here. See the [method definition for documentation](https://github.com/hackclub/hcb/blob/garyhtou/simplify-admin-policies/app/policies/application_policy.rb#L57-L111) on how they work.

In addition, we could also add methods such as:

```ruby
# This handles shared logic
policy_for :new?, :create?, :edit?, :update? do
  user&.admin? || record.event.users.include?(user)
end

# same as...

def new?
  user&.admin? || record.event.users.include?(user)
end

def create?
  user&.admin? || record.event.users.include?(user)
end
# ...
```

If we wanted to go with a more DSL-like syntax, we could do something like:
```ruby
# I renamed the `policy_for` method above to `action`
# defines `new?` method using the block
action :new? do
  user&.admin? || record.event.users.include?(user)
end

# defines `new?` and `create?` methods using the block
action :new?, :create? do
  user&.admin? || record.event.users.include?(user)
end

# using `action` allows us to chain methods together in an English-like manner.
# This defines `new?` and `create?` methods using `user&.admin? || block.call`
permit_admins action :new?, :create? do
  record.event.users.include?(user)
end
# this is the equivalent of defining methods `:new?` and `:create?` with the body `user&.admin? || record.event.users.include?(user)`.
```


garyhtou:
What do we think about this? Is this just some cursed Ruby that no one will understand?

malted:
I think the second example (the DSL-like one) is way too obfuscated.

I really like the first example though.

garyhtou:
With the introduction of https://github.com/hackclub/hcb/issues/5525, this experiment might not make such sense any more. The patterns become less defined

PR #5513, opened 2024-03-14
Remove scrollbar when there's no need to scroll

YodaLightsabr:
## Summary of the problem
There's an extra scrollbar

## Describe your changes
I've removed the scrollbar when it's not needed

[Before](https://hackclub.slack.com/archives/C06JEH8HJSK/p1710390558481269)
[After](https://hackclub.slack.com/archives/C06JEH8HJSK/p1710390558481269)


sampoder (approved):
LGTM! Tested on pairing call.

PR #5519, opened 2024-03-14
[Reimbursements] Add quick expense feature to reports

sampoder:
Turns the Add an Expense button into a file dropzone:

https://github.com/hackclub/hcb/assets/39828164/d024f3f9-260d-4553-afa5-29a33fad2d44


garyhtou:
so cool! i wonder if we could add a tooltip to the button to help people discover this feature

sampoder:
nice idea, yeah let me add that

garyhtou (commented):
🪄 i love it!

PR #5521, opened 2024-03-14
Fix invoices erroring when no country

cjdenio:


app/views/hcb_codes/_invoice.html.erb
```diff
@@ -110,7 +110,7 @@
           <%= @sponsor.address_line2 %><br>
         <% end %>
         <%= @sponsor.address_city %>, <%= @sponsor.address_state %> <%= @sponsor.address_postal_code %> <br>
-        <%= ISO3166::Country[@sponsor.address_country].common_name %>
+        <%= ISO3166::Country[@sponsor.address_country]&.common_name %>
```
garyhtou: I see that this is fixing the case where a Sponsor has an empty `address_country`.
image


However, one concern I have is that this fix will silence cases where unknown countries are looked up. For example,

```ruby
ISO3166::Country["foobar"]&.common_name
```
will simply return `nil`. However, we likely want to get an Airbrake error about this.

## Solution

```ruby
class Sponsor < ApplicationRecord
  def address_country_name
    return nil if address_country.blank?
    unless (country = ISO3166::Country[address_country])
      Airbrake.notify("#{address_country}" is an unknown country code. Invoice: #{self.inspect}")
    end
    country.common_name
  end
```
```erb
<%= @sponsor.address_country_name %>
```

PR #5523, opened 2024-03-14
Require a CPT for Receipt SMS notifications

sampoder:
Odd error: https://heroku-app97991095.airbrake.io/projects/288439/groups/3757233280289307220?tab=notice-detail


garyhtou (commented):
great fix! thanks

PR #5529, opened 2024-03-15
Use `local_hcb_code`

sampoder:


garyhtou (approved):
let's get this merged, but there might be one weird bug

sampoder:
interesting

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -46,7 +46,7 @@
             <% end %>
           <% end %>
 
-          <% if !defined?(receipt_upload_button) && !defined?(hide_tags) && organizer_signed_in? && pt&.hcb_code&.event && !pt&.hcb_code&.event&.demo_mode? && Flipper.enabled?(:transaction_tags_2022_07_29, @event || pt.local_hcb_code.event) %>
+          <% if !defined?(receipt_upload_button) && !defined?(hide_tags) && organizer_signed_in? && pt&.local_hcb_code&.event && !pt&.local_hcb_code&.event&.demo_mode? && Flipper.enabled?(:transaction_tags_2022_07_29, @event || pt.local_hcb_code.event) %>
```
garyhtou: In the condition, we're checking to make sure `pt.local_hcb_code.event` exists. However, the Flipper flag uses `@event || pt.local_hcb_code.event`. This could cause a weird bug for disbursements where `@event` might not always match `local_hcb_code.event`.

If the goal is to not render this for mock data transactions, then i'd do this:

```suggestion
          <% if !defined?(receipt_upload_button) && !defined?(hide_tags) && organizer_signed_in? && !pt.instance_of(OpenStruct) && Flipper.enabled?(:transaction_tags_2022_07_29, @event || pt.local_hcb_code.event) %>
```
(same change for the CT partial)

PR #5538, opened 2024-03-15
Climate Directory: allow `grant recipient` orgs

cjdenio:
We have a few orgs classified as "grant recipients" that should still be able to be listed on the Climate Directory (flagged by Paul)


app/api/api/directory.rb
```diff
@@ -20,7 +20,6 @@ class Directory < Grape::API
         disallowed_categories = [
           "hardware grant",
           "outernet guild",
-          "grant recipient",
           "salary"
         ]
         orgs = orgs.where.not(category: disallowed_categories)
```
garyhtou: I don't remember why this `disallowed_categories` code exists. I think filtering by `where({ event_tags: { name: EventTag::Tags::CLIMATE, purpose: :directory } })` should be good enough?

cjdenio: Honestly yeah, I can't remember why this exists

cjdenio:
Cool, the logic is now (user opted into directory OR has climate directory tag)

app/api/api/directory.rb
```diff
@@ -15,15 +15,8 @@ class Directory < Grape::API
...
+                      Event.includes(:event_tags).where({ event_tags: { name: EventTag::Tags::CLIMATE, purpose: :directory } }),
                     )
-        disallowed_categories = [
-          "hardware grant",
-          "outernet guild",
-          "grant recipient",
-          "salary"
-        ]
-        orgs = orgs.where.not(category: disallowed_categories)
         orgs = Event.where(id: orgs.select(:id))
```
garyhtou: I probably was the one who wrote this code, but I think reload using the ids is unnecessary. we should be able to just do

```ruby
orgs = orgs.includes(:event_tags).with_attached_logo.with_attached_background_image
```

we might even be able to drop the includes(:event_tags) because it's already included above?

PR #5540, opened 2024-03-15
Add easter eggs for "hack on hcb" form

garyhtou:
## Summary of the problem

We want more people to join us! we're _soft launching™️_ a form for people to express their interest in joining us

### HTML comment on all pages
image

### `title` attribute of dino image on 500 pages
image

### Console on 500 pages
image

### and a good old flavor text
image


sampoder (dismissed):
LGTM!

PR #5542, opened 2024-03-15
Add missing copy to meatball menu + conditionally render pin function

sampoder:
Fixes https://github.com/hackclub/hcb/issues/5531. I also realised we weren't doing a `pinnable?` check so I added that back.


app/views/hcb_codes/_meatballs.html.erb
```diff
@@ -1,24 +1,33 @@
...
-      <%= inline_icon "pin-fill", size: 24 %>
-      <%= hcb_code.pin.present? ? "Unpin from" : "Pin to" %> <%= hcb_code.event.name %>
-    <% end %>
-    <% if (hcb_code.stripe_card? || hcb_code.stripe_force_capture?) && hcb_code.stripe_card.present? && !hcb_code.personal_transaction.present? %>
-      <%= link_to invoice_as_personal_transaction_hcb_code_path(id: hcb_code.hashid), method: :post, class: "tooltipped tooltipped--s flex items-center", 'aria-label': "Invoice #{hcb_code.stripe_cardholder&.user&.name || "yourself"} for this transaction.", data: { confirm: "Are you sure you'd like to invoice #{hcb_code.stripe_cardholder&.user&.name || "yourself"} for this transaction? They'll be emailed an invoice." } do %>
-        <%= inline_icon "sad", size: 24 %>
-        Flag as personal transaction
+<% is_pinned = hcb_code.pin.present? %>
+<% pinnable = hcb_code.pinnable? %>
+<% flagable = hcb_code.pinnable? %>
```
garyhtou: Should flagable and pinnable share the same logic?

From my understanding, we can only `flag` card transactions, but we can `pin` nearly all transactions

sampoder: Nope, bad typo!

PR #5547, opened 2024-03-15
Meatball menus render on mobile

sarthaktexas:
## Summary of the problem

adds a z-index

## Describe your changes

it adds a z-index

well it works now


app/views/hcb_codes/_meatballs.html.erb
```diff
@@ -3,7 +3,7 @@
       "#",
       class: "align-middle", "aria-label": "More Options",
       data: { turbo: true, "menu-target": "toggle", action: "menu#toggle click@document->menu#close keydown@document->menu#keydown" } %>
-  <div class="menu__content menu__content--2 h5 menu__content--compact" data-menu-target="content">
+  <div class="menu__content menu__content--2 h5 menu__content--compact" data-menu-target="content" style="z-index: 999;">
```
garyhtou: Should this be a part of the `menu__content` class?

garyhtou (approved):

app/views/hcb_codes/_meatballs.html.erb
sarthaktexas: no.

PR #5565, opened 2024-03-16
Clarify that this button changes other people's information

sampoder:
Closes https://github.com/hackclub/hcb/issues/5545.


app/views/reimbursement/reports/_actions.html.erb
```diff
@@ -67,7 +67,7 @@
         <% elsif admin_signed_in? %>
           <%= link_to payouts_user_path(user), class: "btn bg-warning" do %>
             <%= inline_icon "profile" %>
-            Add Payout Information
+            Add <%= user.first_name %>'s Payout Information
           <% end %>
```
garyhtou: I think an `admin_tool` would help a lot here!

PR #5573, opened 2024-03-16
Reimbursement: Hide Conversation for Initial Drafts

sampoder:
If you've never submitted it, we won't show the conversation view.

To implement this I've added `initial_draft?` to `Reimbursement::Expense` and then conditionally rendered the partial.


app/views/reimbursement/reports/show.html.erb
```diff
@@ -104,7 +104,9 @@
 
 <%= render partial: "actions", locals: { report: @report, user: @user } %>
 
-<%= render partial: "conversation" %>
+<% unless @report.initial_draft? %>
```
garyhtou: As an organizer, I may want to comment on an old, `initial_draft?` report to ask if the user is ready to submit

sampoder: How would the user see your comment?

garyhtou: oh- good point 🤡

garyhtou: nevermind then. the logic condition for that is gonna get overly compmlicated

garyhtou (approved):

PR #5576, opened 2024-03-16
[Reimbursements] Add Report Deleting

sampoder:
Currently, I've made this admin only if the report isn't an initial draft.

Screenshot 2024-03-15 at 10 01 22 PM

https://github.com/hackclub/hcb/assets/39828164/fe15f9f5-c77c-4550-88b8-6beec912e7b2


app/views/reimbursement/reports/show.html.erb
```diff
@@ -19,14 +19,13 @@
     </span>
     <span class="ml0 badge bg-<%= @report.status_color %>"><%= @report.status_text %></span>
   </h3>
-  <h2 class="heading mt0 mb1 flex justify-start" style="gap: 4px;">
+  <h2 class="heading mt0 mb1 flex justify-start" style="gap: 0px;">
```
sarthaktexas:
```suggestion
  <h2 class="heading mt0 mb1 flex justify-start">
```

sarthaktexas (dismissed):

app/views/reimbursement/reports/_edit_form.html.erb
```diff
@@ -19,5 +19,11 @@
       <span class="muted">You can transfer a report to any organization you're a part of.</span>
     <% end %>
   </div>
-  <%= form.submit "Update Report", class: "mt2" %>
+  <div class="flex mt2" style="gap: 8px">
+    <%= form.submit "Update Report" %>
+  </div>
```
sarthaktexas: what is the flexbox here used for? there is only one element inside.

app/views/reimbursement/reports/_edit_form.html.erb
```diff
@@ -19,5 +19,11 @@
...
+<% end %>
+
+<% admin_tool_if !@report.initial_draft? do %>
+  <%= button_to "Delete Report", reimbursement_report_path(@report), class: "ml0 info bg-error btn", method: :delete, type: "button" %>
```
sarthaktexas: I think you should put this within the flexbox above.

app/views/reimbursement/reports/_edit_form.html.erb
garyhtou: ~~it's to keep things flexyyyyy lol~~

app/controllers/reimbursement/reports_controller.rb
```diff
@@ -194,6 +194,19 @@ def request_changes
       redirect_to @report
     end
 
+    def destroy
+
+      authorize @report
+
+      @report.destroy
+
+      if admin_signed_in? || organizer_signed_in?
```
garyhtou: I believe the `admin_signed_in?` is redundant because `organizer_signed_in?` will always return `true` if you're an admin

```suggestion
      if organizer_signed_in?
```

app/policies/reimbursement/report_policy.rb
```diff
@@ -46,6 +46,10 @@ def admin_approve?
       admin && open
     end
 
+    def destroy?
+      ((team_member || creator) && record.initial_draft?) || admin
+    end
```
garyhtou: Important: No one (including admins) should be able to delete a reimbursement that has been reimbursed

garyhtou (dismissed):
🪄 love it! Just one major bug/consideration

PR #5577, opened 2024-03-16
[Reimbursement::Report] Use `amount_cents` for displaying amounts

sampoder:
This will close https://github.com/hackclub/hcb/issues/5575.


garyhtou (commented):
I've got a slight concern that we have a separate method used just for displaying. It leaves the possibility where the button shows a different amount than what will actually be reimbursed.

Looking at the three methods, I think what we really want here is:

```ruby
# Used where possible
def reimbursable_amount_cents
  [expenses.approved.sum(:amount_cents), maximum_amount_cents].compact.min
end

# Used at the top of the reports page
def amount_cents_to_display
  return reimbursable_amount_cents if reimbursement_requested? || reimbursement_approved? || reimbursed?

  expenses.sum(:amount_cents)
end

# If we need the total expenses amount cents, then just use `expenses.sum(:amount_cents)`
```

sampoder:
I'm cool with that

sampoder:
Actually, just looked over the code. We can use `amount_cents` directly.

YodaLightsabr (approved):

sarthaktexas (approved):

PR #5579, opened 2024-03-16
[Reimbursement] Use a turbo stream for a report's blankslate

sampoder:
This will fix the bug where "Get started by adding an expense." would stick around when you created an expenses.

Closes https://github.com/hackclub/hcb/issues/5535


app/views/reimbursement/reports/_blankslate.html.erb
```diff
@@ -0,0 +1,5 @@
+<div id="blankslate">
+  <% unless report.expenses.any? %>
+    <%= blankslate "Get started by adding an expense." %>
+  <% end %>
+</div>
```
YodaLightsabr:
```suggestion
</div>

```

YodaLightsabr (approved):

PR #5580, opened 2024-03-16
[Reimbursements] Replace `expense_number` custom implementation with `sequenced`

sampoder:
Now using this gem: https://rubygems.org/gems/sequenced. Uses a custom migration file, I've ran this locally and it seemed fine (it's based on the docs https://rubygems.org/gems/sequenced) but I'd like someone to look over it closely for me.

Closes https://github.com/hackclub/hcb/issues/5534


db/migrate/20240316063623_add_sequential_id_to_reimbursement_reports.rb
```diff
@@ -0,0 +1,24 @@
...
+      
+      execute <<~SQL
+        UPDATE reimbursement_expenses
+        SET sequential_id = old_reimbursement_expenses.next_sequential_id
+        FROM (
+          SELECT id, ROW_NUMBER()
+          OVER(
+            PARTITION BY reimbursement_report_id
+            ORDER BY id
+          ) AS next_sequential_id
```
garyhtou: Links to old `expense_number`s will likely break, but that's okay!

garyhtou (commented):
Sequential numbers get reused when Expenses get deleted. From my understanding, we always want a new number.

Here's my suggestion:

In a separate PR, add `acts_as_paranoid` to Report and Expense. [Make sure deletion works properly](https://github.com/hackclub/hcb/pull/5583#issuecomment-2002606312)

Then, come back to this PR and test it to see if soft deletion helps fix the problem

garyhtou:
The migration itself looks good!

sampoder:
Closing for now, we should pick this back up at some point but we should do it in a new branch.

PR #5583, opened 2024-03-16
[Reimbursements] Add Papertrail to Expenses & display approved by / approval removed by

sampoder:
Screenshot 2024-03-16 at 9 31 21 AM
Screenshot 2024-03-16 at 9 31 31 AM

Closes https://github.com/hackclub/hcb/issues/5582


garyhtou:
Can we also add paper trail to report? and make sure everything works correctly when deleting

sampoder:
Papertrail is already in the report, it's being used to generate the conversation. Wdym by deleting?

sarthaktexas (approved):

garyhtou:
Sometimes validations fail unexpectedly when working with deleted records.

For example, a unique constraint might validate with all records, and not just the non-deleted records.

I'd suggest looking through our validations to see if any are affected

garyhtou:
Oh my bad, I just realized i was thinking of `acts_as_paranoid` instead of `has_paper_trail`

PR #5593, opened 2024-03-16
Statements: Add an end date parameter

sarthaktexas:
## Summary of the problem

Multiple users have asked for yearly statements

## Describe your changes
Adds an optional query parameter.


sampoder (approved):
I'm cool with this, let's now show this in the UI for now though.

PR #5599, opened 2024-03-16
Reimbursements: Add summary bar

YodaLightsabr:
This adds a summary bar with information and a hint/description for users based on the state of the button. It's also mobile responsive.

Here are a few screenshots of the summary bar in different states.

image
image
image
image


sampoder (approved):
This is incredible! I've made a couple of small copy edits and styling changes but with that, let's get this merged.

PR #5603, opened 2024-03-17
[Comments] Add User Mentions

sampoder:
Very basic implementation at the moment:

Screenshot 2024-03-17 at 1 00 08 PM

Currently:

* You can mention people using friendly ID
* Click a user's face on a comment to mention them
* It will custom render the user mentions
* Each comment has a `mentioned` method that returns an array of people who were mentioned

To improve this:

* People need to be able to set their username
* There should be something that prompts people to like the GitHub pop up:

Screenshot 2024-03-17 at 12 59 09 PM


app/helpers/users_helper.rb
```diff
@@ -59,11 +59,11 @@ def avatar_for(user, size = 24, options = {})
     src = profile_picture_for(user, size)
```
```suggestion
    # Handle case when this method is used outside the controller
    current_user = defined?(current_user) ? current_user : nil

    src = profile_picture_for(user, size)
```

garyhtou: The rest of the changes in this method, exclude for `!options[:disable_tooltip]`, and be removed

app/models/comment.rb
```diff
@@ -67,6 +67,17 @@ def action_text
     return "commented"
   end
 
+  def mentioned_users
+    mentioned_users = []
+    content.scan(/([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+\.[a-zA-Z0-9_-]+)/) do |match|
+      u = User.find_by(email: match) rescue nil
+      if u && (!commentable.event.present? || commentable.event.users.include?(u))
```
garyhtou: This won't work for disbursement because disbursements have two events

app/models/comment.rb
```diff
@@ -67,6 +67,17 @@ def action_text
...
+        mentioned_users << u
+      end
+    end
+    mentioned_users
+  end
```
garyhtou: Currently, I see that we're handling authorization in this method. From the name `mentioned_users`, I would expect to get all users. We should handle authorization elsewhere and use Pundit.

```suggestion
  def mentioned_users
    emails = content.scan(/@([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+\.[a-zA-Z0-9_-]+)/)
    User.where(email: emails)
  end
```
^ please test this

But there are two changes that should be here:
- Only email addresses prefixed with `@` should be mentioned. (make sure scan doesn't include the @)
- Use URI's email regex helper

app/services/markdown_service.rb
```diff
@@ -20,6 +27,15 @@ def preprocess(fulldoc)
       fulldoc
     end
 
+    def postprocess(fulldoc)
+
+      # this is used to mention users
+      # in comments, it's very basic.
```
garyhtou:
```suggestion
      # This is to strip the `@` in front of user mentions after generating their HTML
```

app/services/markdown_service.rb
```diff
@@ -5,8 +5,15 @@ class MarkdownService
...
   class MarkdownRenderer < Redcarpet::Render::HTML
     include ApplicationHelper # for render_money helper
+    include UsersHelper # for mention users helper
     include Rails.application.routes.url_helpers
 
+    def initialize(current_user: nil, record: nil, **args)
+      super(**args)
+      @current_user = current_user
+      @record = record
+    end
```
garyhtou:
```suggestion
    def context(current_user: nil, record: nil)
      @current_user = current_user
      @record = record
      self
    end
```

app/services/markdown_service.rb
```diff
@@ -111,8 +137,8 @@ def app_hosts
 
   end
 
-  def renderer
-    markdown_renderer = MarkdownRenderer.new(hard_wrap: true,
+  def renderer(current_user: nil, record: nil)
+    markdown_renderer = MarkdownRenderer.new(current_user:, record:, hard_wrap: true,
                                              filter_html: true)
```
garyhtou:
```suggestion
    markdown_renderer = MarkdownRenderer.new(hard_wrap: true,
                                             filter_html: true).context(current_user:, record:)
```

app/views/comments/_form.html.erb
```diff
@@ -8,7 +8,17 @@
     <% end %>
   </div>
 
-  <%= form.text_area :content, class: "card mb2 overflow-visible", placeholder: "Add a note…", "data-behavior": "ctrl_enter_submit", style: "min-height: 100px" %>
+  <text-expander keys="@">
+    <%= form.text_area :content, class: "card mb2 overflow-visible", "data-behavior": "ctrl_enter_submit comment", placeholder: "Add a note…", style: "min-height: 100px", id: "comment_textarea" %>
+  </text-expander>
+
+  <ul role="listbox" class="mention-suggestions card" hidden popover>
+    <% commentable.event.users.each do |u| %>
```
garyhtou: This breaks for disbursements

app/views/comments/_form.html.erb
```diff
@@ -8,7 +8,17 @@
     <% end %>
   </div>
 
-  <%= form.text_area :content, class: "card mb2 overflow-visible", placeholder: "Add a note…", "data-behavior": "ctrl_enter_submit", style: "min-height: 100px" %>
+  <text-expander keys="@">
+    <%= form.text_area :content, class: "card mb2 overflow-visible", "data-behavior": "ctrl_enter_submit comment", placeholder: "Add a note…", style: "min-height: 100px", id: "comment_textarea" %>
+  </text-expander>
+
+  <ul role="listbox" class="mention-suggestions card" hidden popover>
```
garyhtou: styling is broken in firefox

app/views/comments/_item.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="mt2 mb2">
     <div class="card border b--muted pb0 pt0 mb2 max-width-4<%= " current-user" if comment.user == current_user %>  <%= " comment--admin" if comment.admin_only? %>" id="comment-<%= comment.hashid %>">
       <div class="border-bottom card__banner secondary">
-        <%= user_mention comment.user %> <%= comment.action_text %>
+        <%= user_mention comment.user, { avatar: { data: { behavior: "mention", mention: "@#{comment.user.email}" } } } %> <%= comment.action_text %>
```
garyhtou: Let's rename `mention` to `mention-value`

garyhtou: Let's update `user_mention` and `avatar_for` to behave like:

```ruby
user_mention(user, click_to_mention: true)

# same thing for `avatar_for`
```

`click_to_mention` should be a keyword arg, and not as apart of the existing `options` hash. This is because the `options` hash is currently confusing and it's not clear whether something is a parameter to the method or being passed directly to HTML.

When you pass `click_to_mention` to `user_mention`, it should just pass that argument into `avatar_for` which then adds the necessary HTML attributes to make it happen

app/views/comments/_form.html.erb
garyhtou:
```suggestion
    <% (@event || commentable.event).users.each do |u| %>
```

garyhtou:
```ruby
class Mention
  def initialize(user, record, event: nil)
    @user, @record = user, record
    @event = event || @record.try(:event)
  end

  def mentionables
    record.try(:mentionables)
  end

  def mentionable_users
  end

end
```

```ruby
module HasMentionables
  extend ActiveSupport::Concern
  included do
    def mentionables
      # returns list of mentionables
    end
  end
end
```

app/models/comment.rb
sampoder: I tried doing this:

```
emails = content.scan(/@(#{URI::MailTo::EMAIL_REGEXP})/)
```

and:

```
emails = content.scan(/@(#{URI::MailTo::EMAIL_REGEXP})/)
```

Neither seems to work, anything you'd recommend @garyhtou?

app/models/hcb_code.rb
```diff
@@ -464,6 +464,17 @@ def comment_mailer_subject
     return "New comment on #{self.memo}."
   end
 
+
+  def comment_mentionable(current_user: nil)
+    users = []
+    users += self.comments.map(&:user)
+    users += self.comments.map(&:mentioned_users).flatten(1)
```
```suggestion
    users += self.comments.flat_map(&:mentioned_users)
```

app/models/hcb_code.rb
```diff
@@ -464,6 +464,17 @@ def comment_mailer_subject
...
+    users += self.events.select { |e| !current_user || Pundit.policy(current_user, e).team? }.flat_map(&:users)
```
```suggestion
    users += self.events.includes(:users).select { |e| !current_user || Pundit.policy(current_user, e).team? }.flat_map(&:users)
```

app/models/hcb_code.rb
```diff
@@ -464,6 +464,17 @@ def comment_mailer_subject
     return "New comment on #{self.memo}."
   end
 
+
+  def comment_mentionable(current_user: nil)
+    users = []
+    users += self.comments.map(&:user)
```
```suggestion
    users += self.comments.includes(:users).map(&:user)
```

app/models/hcb_code.rb
```diff
@@ -464,6 +464,17 @@ def comment_mailer_subject
...
+    users += self.comments.map(&:mentioned_users).flatten(1)
+    users += self.events.select { |e| !current_user || Pundit.policy(current_user, e).team? }.flat_map(&:users)
+    users += self.events.map(&:point_of_contact)
```
```suggestion
    users += self.events.includes(:point_of_contact).map(&:point_of_contact)
```

app/models/reimbursement/report.rb
```diff
@@ -186,6 +186,15 @@ def comment_recipients_for(comment)
       users.excluding(comment.user).collect(&:email_address_with_name)
     end
 
+    def comment_mentionable(current_user: nil)
+      users = []
+      users += self.comments.map(&:user)
```
garyhtou:
```suggestion
      users += self.comments.includes(:users).map(&:user)
```

app/models/reimbursement/report.rb
```diff
@@ -186,6 +186,15 @@ def comment_recipients_for(comment)
...
+      users += self.comments.map(&:mentioned_users).flatten(1)
```
garyhtou:
```suggestion
      users += self.comments.flat_map(&:mentioned_users)
```

app/views/comment_mailer/hcb_code/notification.html.erb
```diff
@@ -4,7 +4,7 @@
 
   <%= @comment.user.name %> has left a comment on <%= "and attached a file to" if @comment.file %> <i><%= hcb_code.memo %></i> in <i><%= hcb_code.event.name %></i>:
 
-  <blockquote><%== MarkdownService.instance.renderer.render @comment.content %></blockquote>
+  <blockquote><%== MarkdownService.instance.renderer(current_user: nil, record: @comment).render @comment.content %></blockquote>
```
garyhtou:
```suggestion
  <blockquote><%== MarkdownService.instance.renderer(record: @comment).render @comment.content %></blockquote>
```

app/models/comment.rb
```diff
@@ -67,6 +67,11 @@ def action_text
     return "commented"
   end
 
+  def mentioned_users
+    emails = content.scan(/@([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+\.[a-zA-Z0-9_-]+)/)
+    User.where(email: emails)
+  end
+
```
garyhtou:
```suggestion
  # This regex was stolen from URI::MailTo::EMAIL_REGEXP
  USER_MENTION_REGEX = /@([a-zA-Z0-9.!\#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*)/

  def mentioned_users
    emails = content.scan(USER_MENTION_REGEX)
    User.where(email: emails)
  end

```

garyhtou (commented):
🪄

Please test:
- Mailers (the one that renders markdown)  ✅
- new mentions javascript (dropdown and the click to mention) ✅
- reimbursement expense mentioning   ✅
- pages that use `user_mention` or `avatar_for`  ✅
- markdown rendered not related to comments (donation page or transparency message)   ✅

app/helpers/users_helper.rb
```diff
@@ -67,16 +68,19 @@ def avatar_for(user, size = 24, options = {})
     alt ||= user&.initials
     alt ||= "Brown dog grinning and gazing off into the distance"
 
+    options[:data] = (options[:data] || {}).merge(behavior: "mention", mention_value: "@#{user.email}") if click_to_mention
+
     image_tag(src, options.merge(loading: "lazy", alt:, width: size, height: size, class: klass))
   end
 
-  def user_mention(user, options = {}, default_name = "No User")
+  def user_mention(user, options = {}, default_name = "No User", click_to_mention: false)
```
garyhtou: For user mentions inside comments, we should put them in a pill/badge-like style.

sampoder:
Pills: 

Screenshot 2024-04-04 at 4 20 33 PM

PR #5608, opened 2024-03-18
[Reimbursements] Add `ctrl_enter_submit` to comments

sampoder:
This was on the launch document but we never got around to it.


garyhtou:
dang i did not realize we had that attribute

garyhtou:
oh i guess i wrote that code and forgot about it
image

PR #5611, opened 2024-03-18
[Reimbursements] Add an explainer of "Reports" for first-time users

sampoder:
This was on our launch document, the concept of "reports" might be a bit confusing for people.

Now for folks who've never created a report before, we'll have a handy helper message:

Screenshot 2024-03-18 at 12 12 22 AM

Screenshot 2024-03-18 at 12 12 29 AM


sampoder:
I'd love copy edits / suggestions! This took me way too long... and I'm still not quite happy with it.

app/views/reimbursement/reports/_create_form.html.erb
```diff
@@ -1,6 +1,12 @@
 <section data-behavior="modal" role="dialog" id="create" class="modal modal--scroll bg-snow">
   <%= form_with(model: Reimbursement::Report, class: "mb3", style: "max-width: 600px;", method: :post) do |form| %>
     <%= modal_header "Open a Report" %>
+    <% unless current_user.created_reimbursement_reports.any? %>
+      <div class="card border b--info mt2 mb2 container--sm mx-auto">
+        <p class="mt0">Reports are a collection of expenses compiled by a team member or volunteer and submitted to HCB for reimbursement.</p>
+        <p class="mb0">To be reimbursed through HCB, start by creating a report, giving it a title, and adding expenses. Subsequently, submit the report for review from your team and, then, the HCB team.</p>
```
garyhtou: here's my attempt. i'm not sure if I love it

```
Paid for something out of pocket? Create a report to get reimbursed.

Reimbursement reports are a collection of expenses paid by an individual who is now requesting to be reimbursed by an organization. To begin the process, start by creating a report, adding expenses, and then submitting the report for review.
```

---

- we want to make it clear that the organization is the one reimbursing, not HCB
- no need to say "give it a title" — it's a minor detail, and the form will force them to provide one
- I like the "Paid for something out of pocket?" intro because it might help guide those who don't know the term "reimbursement"

sampoder: Thanks! I like this, I'm going to be merge the two

sampoder:
This is simplified a fair bit but I think I'm going to go with this for now:

Screenshot 2024-03-18 at 4 53 24 PM

Would still love feedback and edits though :D!

garyhtou (approved):
sounds good to me

sampoder:
coolio, lgtm then

PR #5612, opened 2024-03-18
[Reimbursements] Use Hash IDs

sampoder:
We had `/reports/1`... `/reports/2` etc. which was great for testing, but sequential IDs when you may not have access to everything is a bit odd. @maxwofford flagged that we should move away from this last weekend, this PR switches `Reimbursement::Report` & `Reimbursement::Expense` to both use hash IDs


garyhtou (commented):
🪄 thanks for moving us to hashids!

PR #5618, opened 2024-03-18
Include users' names in mailers

cjdenio:
Closes #5520


app/models/user.rb
```diff
@@ -254,6 +254,10 @@ def build_payout_method(params)
     self.payout_method = payout_method_type.constantize.new(params)
   end
 
+  def email_address_with_name
+    ActionMailer::Base.email_address_with_name(email, name)
```
garyhtou: what happens if we don't have a name? either `nil` or `blank?`

cjdenio: This is handled:
```ruby
[1] pry(main)> ActionMailer::Base.email_address_with_name("caleb@hackclub.com", nil)
=> "caleb@hackclub.com"
[2] pry(main)> ActionMailer::Base.email_address_with_name("caleb@hackclub.com", "")
=> "caleb@hackclub.com"
[3] pry(main)> 
```

garyhtou: Amazing!

garyhtou (dismissed):
Great!

just one spot left
https://github.com/hackclub/bank/blob/fc8585391526c9f20b6a9a16a939ba94e4f1f0a0/app/views/events/team.html.erb#L12

PR #5620, opened 2024-03-18
Fix backwards ACH amount in API

cjdenio:


app/views/hcb_codes/_ach_transfer.html.erb
```diff
@@ -113,3 +113,5 @@
     </section>
   <% end %>
 </article>
+
+<% title "#{render_money(@hcb_code.amount_cents.abs)} transfer to #{@hcb_code.ach_transfer.recipient_name}" %>
```
cjdenio: Added because the default "Transaction for -$100.00" title didn't look very nice

PR #5627, opened 2024-03-18
[User roles] Change permission comparison on `/roles` to a table

malted:
Before

image

After

image


app/views/static_pages/roles.html.erb
```diff
@@ -4,7 +4,7 @@
   <h1>User Roles</h1>
   <h2>Overview</h2>
   <p>
-    <% reference_date = Date.new(2024, 3, 1) %>
+    <% reference_date = Date.new(2024, 3, 9) %>
     <% days_since = (Date.today - reference_date).to_i %>
 
     Hi there! 👋 On <%= reference_date.strftime("#{reference_date.day.ordinalize} %B %Y") %> (<%= days_since %> days ago),
```
sarthaktexas: What is this abnormal/non-US date format?

malted: get a grip Sarthak

garyhtou (approved):

PR #5639, opened 2024-03-19
[Reimbursements] Add a modal for payout information 

sampoder:
https://github.com/hackclub/hcb/assets/39828164/ce2ed10f-98ec-4acd-a9ee-94048681d614


app/views/reimbursement/reports/_actions.html.erb
```diff
@@ -80,15 +80,15 @@
       <div style="flex-grow: 1;" class="flex justify-end">
         <% if !user.payout_method.present? %>
           <% if user == current_user %>
-            <%= link_to settings_payouts_path, class: "btn bg-warning" do %>
+            <%= link_to "#", class: "btn bg-warning", data: { behavior: "modal_trigger", modal: "edit_payout" } do %>
```
garyhtou: i think we want to keep the link href to accessibility. it also allow people to middle click and open the page in a new tab.

(let's make sure it doesn't both open the modal and the tab)

```suggestion
            <%= link_to settings_payouts_path, class: "btn bg-warning", data: { behavior: "modal_trigger", modal: "edit_payout" } do %>
```

app/views/reimbursement/reports/_actions.html.erb
```diff
@@ -80,15 +80,15 @@
...
               <%= inline_icon "profile" %>
               Add Payout Information
               <% button_hint = capture do %>
                 <span class="bold">Your next step:</span> before you can be reimbursed, you need to provide HCB your payout information.
               <% end %>
             <% end %>
           <% elsif admin_signed_in? %>
-            <%= link_to payouts_user_path(user), class: "btn bg-warning" do %>
+            <%= link_to "#", class: "btn bg-warning", data: { behavior: "modal_trigger", modal: "edit_payout" }  do %>
```
garyhtou:
```suggestion
            <%= link_to payouts_user_path(user), class: "btn bg-warning", data: { behavior: "modal_trigger", modal: "edit_payout" }  do %>
```

app/views/reimbursement/reports/_actions.html.erb
```diff
@@ -198,3 +198,8 @@
     </div>
   <% end %>
 </div>
+
+<section data-behavior="modal" role="dialog" id="edit_payout" class="modal modal--scroll bg-snow">
+  <%= modal_header "Add Payout Information" %>
+  <%= render "/users/payout_form", user: report.user, states: ISO3166::Country.new("US").subdivisions.values.map { |s| [s.translations["en"], s.code] }, report: %>
```
garyhtou: Maybe we inline the `states` stuff where it's being used in the partial (I think there's only one place). this will prevent us from having this code both here and in the controller

garyhtou (commented):
🪄 you're so quick 🫨 great work, I love it

sampoder:
🫨

PR #5642, opened 2024-03-19
"account statement" -> "financial statement"

zachlatta:
## Summary of the problem

It's important that users understand that they are fiscally sponsored by us, not that they have a bank account by us. 

## Describe your changes

This is a small wording change to make it clear this statement is the ledger of your fund, not the ledger of bank account.


cjdenio:
I think this makes the copy more confusing honestly— "account statement" is a term most users will be familiar with, "fund statement" is not. We already make the fund/account distinction very clear during the application process

zachlatta:
The problem is that they technically have a "fund" within our foundation, not an "account" within our foundation. For other fiscal sponsors, they call the statements they generate "fund statements" too.

I know it's not perfect, but this is something that it's important for us to have the right wording around.

sampoder:
could we use wording like "monthly statement" or "financial statement" to avoid using the word "account"? i feel like those are more familiar terms. we also do have an explanation of fiscal sponsorship at the bottom of every statement. we can also just go with "Statements".

is the main concern here that users will think they have a bank account with us? or that a non-user of HCB who sees an org's statement will think that that org has a bank account with us?

zachlatta:
If we remove the word "account" and have a description of fiscal sponsorship on every statement, that'd be good for me.

If an external party ever comes to us and says: "Hey! You tricked your users into thinking they had bank accounts with you!", I want us to be able to respond with "No, we make it extremely clear, in every part of our onboarding and in the app itself, that this is a fiscally sponsored fund - not a bank account".

Right now they can object to that last statement and say "No, look - your generated statements are 'account statements'. Like a bank account statement!"

sampoder (approved):
i've gone with financial statement, because technically we can have non-monthly statements. @cjdenio is that cool with you?

garyhtou (approved):

PR #5644, opened 2024-03-20
[Reimbursements] Add an approve all expenses button

sampoder:
Fixes https://github.com/hackclub/hcb/issues/5632 in my opinion.

Adds this button (when no expenses have been approved):
 
Screenshot 2024-03-19 at 7 15 21 PM

Which does what you'd expect.


garyhtou:
should it maybe be approve all & reimburse?

sampoder:
I would them to still be separate stages, just so that its clear whats happening

garyhtou:
i haven't thought much about this flow, but i just wanted to suggest it as another option

sampoder:
Yeah, I'd like to get this in first, see how people feel about it. And if people would just prefer one button, switch over to that.

garyhtou:
sounds good to me!

PR #5647, opened 2024-03-20
[DO NOT MERGE] Sensitive database column discusssion

albertchae:
## Summary of the problem

@KiKoS0 and I generated a possible yaml file with database obfuscation decisions (currently targeting https://github.com/Qovery/Replibyte) in hackmd, but this representation was not conducive to discussion. So I'm moving that to annotations on `db/schema.rb`, so we can discuss via the Github PR UI. Currently this is only considering built in transformers https://www.replibyte.com/docs/transformers, but we can write custom ones if necessary

The [second commit here](https://github.com/hackclub/hcb/commit/c9657e46d59a3ff97384d0d3dcab0caa049bf191) enables parsing `db/schema.rb` to generate the yaml we need to run replibyte. The script is
```ruby
require 'skeema/ripper'
schema = Skeema::Ripper.parse('db/schema.rb')

tables = schema.keys

replibyte_tables = tables.filter do |table|
  schema[table].any? do |column_entry|
    hash = column_entry.values.first
    hash&.key?("replibyte:")
  end
end

replibyte_tables.each do |table|
  puts "table: #{table}"
  puts "columns:"
  schema[table].each do |column_entry|
    column_name = column_entry.keys.first
    hash = column_entry.values.first
    if hash&.key?("replibyte:")
      puts "  - name: #{column_name}"
      puts "    transformer_name: #{hash["replibyte:"]}"
    end
  end
  puts ""
end
```

Alternatively, we may consider using https://github.com/GreenmaskIO/greenmask or other tools, but either way the decision on which columns to redact/censor is tool agnostic.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -205,6 +205,8 @@ group :development do
   gem "foreman"
 end
 
+gem 'skeema', git: 'https://github.com/albertchae/skeema-replibyte', branch: 'skeema-replibyte'
+
 gem "jbuilder", "~> 2.11"
 
 gem "libledger"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,4 +1,11 @@
 GIT
+  remote: https://github.com/albertchae/skeema-replibyte
+  revision: c0becd4c6919735b79fd64f01861fe39041f3321
+  branch: skeema-replibyte
+  specs:
+    skeema (0.0.2)
+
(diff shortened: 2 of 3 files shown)


albertchae:
Closing this as providing a sanitized schema dump is not currently on the roadmap

PR #5652, opened 2024-03-20
Check Deposit: Let non-managers upload checks

malted:


app/policies/check_deposit_policy.rb
```diff
@@ -6,7 +6,7 @@ def index?
   end
 
   def create?
-    !record.event.demo_mode? && !record.event.outernet_guild? && user_who_can_transfer?
+    admin_or_user? && !record.event.demo_mode?
```
garyhtou: Unfortunately, we need to keep this outernet guild check!

```suggestion
    admin_or_user? && !record.event.demo_mode? && !record.event.outernet_guild?
```

garyhtou (approved):
Great! please test it and merge!

PR #5664, opened 2024-03-20
[Reimbursements] Add Soft Deletion To Reports

sampoder:
See recent Airbrake errors.


garyhtou:
For context, hard deleting reports but soft deletion expenses causes a foreign key validation error

PR #5667, opened 2024-03-21
[Comments] Don't attach files

sampoder:
Security concerns: https://github.com/hackclub/hcb/pull/5405/files#r1533199014


garyhtou (approved):
ty

PR #5670, opened 2024-03-21
[StripeCardPolicy] Refactor policy

malted:
Previously they were using the `admin_or_user?` helper method which is problematic and should be BANISHED.


garyhtou (commented):
These policies for the `StripeCardController` should be refactored so that they're using the `StripeCardPolicy`

app/policies/event_policy.rb
```diff
@@ -56,6 +56,14 @@ def card_overview?
     is_public || admin_or_user?
   end
 
+  def new_stripe_card?
+    create_stripe_card?
+  end
+
+  def create_stripe_card?
+    user&.admin? || (admin_or_user? && is_not_demo_mode?)
```
garyhtou:
```suggestion
    admin_or_user? && is_not_demo_mode?
```

garyhtou (approved):
Double check that `admin_or_user?` isn't used elsewhere now that it's private

garyhtou:
i just double checked. we're good

garyhtou:
No functional changes

PR #5681, opened 2024-03-22
[Receipt Bin] Filter Out Salary Account HCB Codes

sampoder:
## Summary of the problem

Slowing down HQ members from getting their receipts.

## Describe your changes

Filter out based on Stripe card -> Event association.

Long term, I'd like a better way of doing this but we've had this issue for too long


sampoder:
Screenshot 2024-03-21 at 5 28 45 PM

Works locally.

YodaLightsabr (approved):

garyhtou:
Yay! i'm glad that worked!

PR #5685, opened 2024-03-22
[User roles] Hide ACH confirmation letter link for non-managers

malted:
Before
image

After
image


app/views/hcb_codes/_ach_transfer.html.erb
```diff
@@ -96,9 +96,11 @@
         <span>
           <%= @hcb_code.ach_transfer.status_text_long %>
           <% if @hcb_code.ach_transfer.deposited? %>
-            <span class="muted">
-              <%= link_to "(#{ organizer_signed_in? ? 'view confirmation letter' : 'sign in to view' })", ach_transfer_confirmation_path(@hcb_code.ach_transfer, format: :pdf), target: "_blank" %>
-            </span>
+            <% if policy(@event).new_transfer? %>
```
garyhtou: combine with existing `if` above!

garyhtou: We should probably use a different policy for this

malted: Done!

malted: Done!

garyhtou (approved):
Great!

PR #5687, opened 2024-03-22
[User permissions] Hide disbursement confirmation letter link for non-managers

malted:


sampoder:
This feels like something non-managers should be able to see?

malted:
You're right I think it was 5am when I made this (I had just hidden it for ACHs)

PR #5688, opened 2024-03-22
[User permissions] Hide cancel ACH button for non-managers

malted:


garyhtou (commented):
good idea!

two things:
1. should we be using the cancel ach transfer policy instead of the Event's new transfer policy?

2. let's disable the button instead of hiding it. i think it's important for users to understand what actions are possible, even if they aren't able to perform them

malted:
Both done!

garyhtou (approved):

PR #5689, opened 2024-03-22
Vertically align items in ACH details toolbar

malted:
Alignment before:
image

Alignment after:
image

<details>
<summary>"alignment 🤓☝️":</summary>

</details>


garyhtou (approved):
great! FYI, i think this is an issue on other types of pages as well. any hcb code show page with a meatball menu likely has this issue

sampoder:
Yeah we should check invoices and donations as well

sampoder:
Donations and invoices seem good, same with cheques

garyhtou:
okay, thanks!

PR #5690, opened 2024-03-22
[User permissions] Don't let members create/edit gsuite

malted:


app/policies/g_suite_account_policy.rb
```diff
@@ -14,7 +14,7 @@ def show?
   end
 
   def reset_password?
-    admin_or_manager?
+    admin_or_manager? || (user == record.user && record.event.users.include?(user))
```
garyhtou: There's no connection between an HCB Users and the account owner. `g_suite_account.user` is the creator, but not necessarily the owner

```suggestion
    admin_or_manager?
```

garyhtou (approved):
great!

PR #5695, opened 2024-03-22
Fix postal code issue on org settings

cjdenio:
Fixes #5694


sampoder:
Oh perfect thanks Caleb

PR #5699, opened 2024-03-22
[Receipt Bin] Move Inline Linking Box Outside of Upload Form

sampoder:
Lol this was an interesting one! See https://hackclub.slack.com/archives/C026RKHLPNJ/p1711140097097149 for context.

The issue here was that, with Direct Uploads, would attempt to "submit" the form after it had finished directly uploading this file. When there's a form inside of a form, it would for some reason... submit that form. Which would cause https://github.com/hackclub/hcb/issues/5697.

Closes https://github.com/hackclub/hcb/issues/5697.


garyhtou (commented):
interesting! thanks 🪄

PR #5700, opened 2024-03-22
[Recurring Donations] Properly check if this is the first donation.

sampoder:
This should fix our issue with recurring donation messages.


garyhtou:
good catch!

PR #5701, opened 2024-03-23
console.log the hack with us message

YodaLightsabr:
## Summary of the problem
We have this (which is sick!)
image
but it's only in the Elements tab

## Describe your changes
so I added this
image


sampoder (approved):
LGTM!
Screenshot 2024-03-22 at 6 07 48 PM

PR #5708, opened 2024-03-23
[Reimbursements] "Request Changes" -> "Return to"

sampoder:
The request changes button could also be used as a general "Mark as draft" button, I think this is better copy:

Screenshot 2024-03-22 at 6 26 15 PM

Screenshot 2024-03-22 at 6 26 10 PM

Closes https://github.com/hackclub/hcb/issues/5497


garyhtou (commented):
i love this so much!

PR #5712, opened 2024-03-23
[Reimbursements] `approve` / `unapprove` + Turbo on buttons

sampoder:
Fixes https://github.com/hackclub/hcb/issues/5705 & makes the process of approving multiple expenses at once faster by adding Turbo.


app/controllers/reimbursement/expenses_controller.rb
```diff
@@ -63,20 +63,31 @@ def update
...
         format.html { redirect_to @expense.report }
       end
     end
+    
+    def unapprove
+      authorize @expense
+    
+      if @expense.approved?
+        @expense.mark_pending!
+      end
```
garyhtou: thoughts on this?
```suggestion
      @expense.mark_pending! if @expense.may_mark_pending?
```

save for the `mark_approve!` above

sampoder: I'm cool with that

PR #5714, opened 2024-03-23
[Reimbursements] Disable move button until a report is selected

sarthaktexas:
fixes #5703

Uses Alpine to not render the form until something is selected.

Export-1711176116229


sampoder (approved):
this is awesome thank you so much Sarthak!

PR #5715, opened 2024-03-23
[Reimbursements] fixes transition on the move button

sarthaktexas:
It appears after doing some in-depth reading, transitions on x-if were deprecated in v2. So, this is how to actually do it.


sampoder (approved):
lgtm!

PR #5719, opened 2024-03-23
[OPDR] Delete organizer button went to wrong path

sarthaktexas:
## Summary of the problem

We kept getting [404](https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/578/samples/timestamp/2024-03-23T08:38:36Z) errors because it was being redirected to the wrong path.


garyhtou (commented):
good catch! 🪄

image

sarthaktexas:
wtf

PR #5724, opened 2024-03-23
[Stripe Cards] Don't show PAN tooltip by default

sarthaktexas:
There was a tiny error in @malted's latest PR. Users could get the card number without us logging it. This fixes it by only showing the tooltip when they explicitly ask for it.


app/views/stripe_cards/_details.html.erb
```diff
@@ -37,12 +37,12 @@
       <strong>Card number</strong>
       <div class="flex justify-between mono items-center">
         <% if stripe_card.user == current_user && stripe_card.virtual? %>
-          <%= copy_to_clipboard(stripe_card.full_card_number, class: "w-fit", tooltip_direction: "e") do %>
-            <% if @show_card_details %>
+          <% if @show_card_details %>
+            <%= copy_to_clipboard(stripe_card.full_card_number, class: "w-fit", tooltip_direction: "e", if: @show_card_details) do %>
```
garyhtou:
```suggestion
            <%= copy_to_clipboard(stripe_card.full_card_number, class: "w-fit", tooltip_direction: "e") do %>
```

garyhtou (approved):
Thanks! (i tested this locally)

PR #5725, opened 2024-03-23
[Ledger Audits] Add a red accent to HQ-related transactions

sampoder:
https://hackclub.slack.com/archives/C06JEH8HJSK/p1711238244551379

Context: https://github.com/hackclub/hcb/issues/5674


garyhtou:
the red accent/border could be misleading — it feels like an error, or that the transaction is currently flagged.

sampoder:
any other colour you think we could use? i just went for red coz hack club

garyhtou:
Maybe a badge would be better

PR #5727, opened 2024-03-24
[Reimbursements] Allow for a custom `company_name` & `company_entry_description` on ACH Transfers

sampoder:
Closes https://github.com/hackclub/hcb/issues/5503 & https://github.com/hackclub/hcb/issues/5504.


app/models/ach_transfer.rb
```diff
@@ -157,7 +159,8 @@ def send_ach_transfer!
         account_number:,
         routing_number:,
       },
-      company_name: event.name[0...16],
+      company_name: (company_name || event.name)[0...16],
```
garyhtou: before_validation, set event.name if it's nil

app/models/ach_transfer.rb
```diff
@@ -10,6 +10,8 @@
 #  amount                    :integer
 #  approved_at               :datetime
 #  bank_name                 :string
+#  company_entry_description :string
+#  company_name              :string
```
garyhtou: validate length and allow blank

app/controllers/ach_transfers_controller.rb
```diff
@@ -45,7 +45,7 @@ def new
 
   # POST /ach_transfers
   def create
-    @ach_transfer = @event.ach_transfers.build(ach_transfer_params.except(:file).merge(creator: current_user))
+    @ach_transfer = @event.ach_transfers.build(ach_transfer_params.except(:file).merge(creator: current_user, company_name: @event.name, company_entry_description: "PAYMENT"))
```
garyhtou: we can omit the description here. when i'd prefer to pass nothing to Column and have them fill it in the default "PAYMENT".

make sure that passing a null to Column's API will cause them to fill in the field with "PAYMENT" and not set it as an empty string. if it is an empty string, then compact params before sending the API request

garyhtou (commented):
couple feedback (sorry, the comments are brief bc i'm typing this on my phone!)

app/controllers/ach_transfers_controller.rb
sampoder: We do `compact_blank` on the Column call already so yeah I think we can omit this

```suggestion
    @ach_transfer = @event.ach_transfers.build(ach_transfer_params.except(:file).merge(creator: current_user, company_name: @event.name))
```

app/models/ach_transfer.rb
sampoder: Would this be the best way to do this? 

* What about for ACHs that have already been created but not sent? We'd still want a fallback for them, right?

* Also, what if the event's name changes?

app/controllers/ach_transfers_controller.rb
```diff
@@ -45,7 +45,7 @@ def new
 
   # POST /ach_transfers
   def create
-    @ach_transfer = @event.ach_transfers.build(ach_transfer_params.except(:file).merge(creator: current_user))
+    @ach_transfer = @event.ach_transfers.build(ach_transfer_params.except(:file).merge(creator: current_user, company_name: @event.name))
```
cjdenio:
```suggestion
    @ach_transfer = @event.ach_transfers.build(ach_transfer_params.except(:file).merge(creator: current_user))
```

This will fail validation if the event's name is longer than 16 characters

sampoder: 164 in `app/models/ach_transfer.rb` should prevent this?

cjdenio: I mean, it fails the validation on line 79

Screenshot 2024-04-29 at 9 47 55 AM

sampoder: Gotcha!

app/models/ach_transfer.rb
garyhtou:
```ruby
before_validation do
  company_name = event.name if company_name.blank?
end
```

```ruby
AchTransfer.find_each do |ach|
  ach.update!(company_name: ach.event.name[0...16])
end
```

> Also, what if the event's name changes?

let's not worry about that

garyhtou (commented):
🪄

PR #5730, opened 2024-03-25
Implement spec for PendingTransactionEngine::CanonicalPendingTransactionService::Import::BankFee

albertchae:
- **Add RawPendingBankFeeTransaction factory and basic spec**
  Because a factory for this model doesn't exist, create the minimum record that will pass ActiveRecord validations
- **Implement spec for PendingTransactionEngine::CanonicalPendingTransactionService::Import::BankFee**

I noticed a couple potential improvements in the implementation here, but I am putting them off until a follow up. That way we know these tests are exercising the current behavior and that we can refactor without changing the tests.

TODOs:
- add not null constraint to RawPendingBankFeeTransaction date_posted?
- rename RawPendingBankFeeTransaction date_posted to date?
- refactor the `not in` query in Import::BankFee to a left join


spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
albertchae, line 1: I am going to overcomment on this file so that people know to tests can get a sense of the process I'm going through.

The first thing I did was read through https://github.com/hackclub/hcb/blob/main/app/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee.rb and try to identify the different cases to test. The things I noticed in the implementation were
- having previously processed raw pending bank fee transactions or not
- setting `fronted` based on positive amount_cents or not
- batching on 100 at a time

Most tests should follow this structure
- set up the conditions
- execute the code we're trying to test
- check that everything we expected to happen did happen

I think there are various catchy acronyms for this but I don't remember any off the top of my head. I did find [AAA (Arrange Act Assert)](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/) just now though , but there are others.

Then the first thing I did during setup was want to create a `RawPendingBankFeeTransaction`. We can of course do this with plain ActiveRecord like `RawPendingBankFeeTransaction.create!`, but it is common for rails projects to use https://github.com/thoughtbot/factory_bot. The main benefit is you can set up a lot of sane defaults for your models in the factory definition so that the test can highlight only the fields relevant to the current test.

Since we didn't have an existing factory for `RawPendingBankFeeTransaction`, I created one in `spec/factories/raw_pending_bank_fee_transaction_factory.rb` and added a small test to make sure the factory is creating a valid object in `spec/models/raw_pending_bank_fee_transaction_spec.rb`. The reason for adding this test is that we don't want to see other tests fail because the factory is invalid when we could have caught that with a test with much smaller scope.

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe PendingTransactionEngine::CanonicalPendingTransactionService::Import::BankFee do
```
albertchae: Nesting `describe` and `context` blocks in rspec tests is more an art than a science. My general heuristics are to avoid nesting more than 1-3 levels deep and to not rely on nesting for shared setup. It's generally better to repeat yourself in tests over being excessively DRY (Don't Repeat Yourself) to make them more readable. 

If you end up having a lot of shared test setup that gets overridden in certain nested blocks, it can feel like unpacking a matryoshka doll to figure out what variables are in scope and why. And often you can end up with data that isn't relevant to the test at all which just makes things confusing ([often called mystery guests](https://thoughtbot.com/blog/mystery-guest)).

There are a few acronyms for describing the no DRY in tests. One common one is DAMP https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html, but I honestly don't find DAMP's meaning helpful to me except just that it's a pun on the opposite of DRY 😆

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  context "when there is a raw pending bank fee transaction ready for processing" do
+    it "processes into a CanonicalPendingTransaction" do
+      expect(RawPendingBankFeeTransaction.count).to eq(0) # there are no previously processed raw pending bank fee transactions
```
albertchae: This is actually unnecessary because tests usually start with a blank database due to this setting https://github.com/hackclub/hcb/blob/35b7726a4f4e9ee63e55704768984de15f11dc7f/spec/rails_helper.rb#L49C10-L49C36, which runs each test in a database transaction and rolls it back after. See https://edgeguides.rubyonrails.org/testing.html#testing-parallel-transactions (side note: the setting was renamed in https://github.com/rails/rails/pull/19282, we should probably fix in HCB too)

So given that it's necessary and also this might seem like it goes against the setup/exercise/verify structure I mentioned, in this case I chose to do this because the code we're testing makes a difference between whether are any previously processed `raw_pending_bank_fee_transaction`s or no, so I like having this assertion to explicitly state we're checking the case where there are none.

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      create(:raw_pending_bank_fee_transaction, date_posted: Date.current)
+      # TODO: there is a not null on CanonicalPendingTransaction#date, but not on RawPendingBankFeeTransaction#date_posted,
+      # should we propagate the not null there so we never run into this?
```
albertchae: so it turns out the factory I added was sufficiently valid according to the `RawPendingBankFeeTransaction` model, but then it fails this use unless `date_posted` is populated since `CanonicalPendingTransaction`'s date has a not null. IMO we should add the not null to `RawPendingBankFeeTransaction` as well, but I am going to defer that to a follow up

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      expect do
+        described_class.new.run
+      end.to change { CanonicalPendingTransaction.count }.by(1)
```
albertchae: usually you want to check the return value of the code under test (i.e. `described_class.new.run` here) when you can, but this service doesn't return anything and the only significant behavior it has is creating a `CanonicalPendingTransaction`

It's debatable whether we should check that the created `CanonicalPendingTransaction` has the fields we expect, since that would couple this test very heavily to the implementation in https://github.com/hackclub/hcb/blob/35b7726a4f4e9ee63e55704768984de15f11dc7f/app/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee.rb#L12-L16 and you usually want to test behavior over implementation. Sometimes it's worth being extra pedantic and checking implementation as well. This might actually be one of the cases where it's worth doing it, since importing is literally copying fields from one to another, so I might end up adding it here as a follow up after all.

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+  end
+
+  context "when there are previously processed raw pending bank fee transactions" do
+    before do
+      raw_pending_bank_fee_transaction = create(:raw_pending_bank_fee_transaction)
+      _processed_bank_fee_canonical_pending_transaction = create(:canonical_pending_transaction, raw_pending_bank_fee_transaction:)
+    end
```
albertchae: so the "setup" or "arrange" phase of a test is typically handled in different ways in rspec. It's very common to see it done using `let`, but I'll comment on that later.

Here I chose to use a `before` block which is short for `before(:each)`, so it'll run for every test under this context block. That means it'll run for `ignores it when processing` and the subcontext `when there are also ready to process raw pending bank fee transactions`

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    it "ignores it when processing" do
+      expect do
+        described_class.new.run
+      end.to change { CanonicalPendingTransaction.count }.by(0)
+    end
```
albertchae: hopefully this is fairly self explanatory

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    context "when there are also ready to process raw pending bank fee transactions" do
+      it "processes into a CanonicalPendingTransaction" do
+        new_bank_fee_transaction = create(:raw_pending_bank_fee_transaction, date_posted: Date.current)
+        expect(RawPendingBankFeeTransaction.count).to eq(2) # there is a processed and non processed bank fee transaction
```
albertchae: The `before` block in the outer context ran for this test as well, so now we have 2 `RawPendingBankFeeTransaction`s

It might actually be better to turn the comment I left here into assertions that are self documenting, but when there are only a few database records at play, it can be an easier shortcut to do things by numbers.

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+
+        expect do
+          described_class.new.run
+        end.to change { CanonicalPendingTransaction.count }.by(1)
+
+        pending_transaction = CanonicalPendingTransaction.last
+        expect(pending_transaction.raw_pending_bank_fee_transaction_id).to eq(new_bank_fee_transaction.id)
```
albertchae: Here we actually do want to check that the most recently processed `CanonicalPendingTransaction` matches the raw source. It's a bit of a hack to use `CanonicalPendingTransaction.last`, but it's possible to do this since we just verified above that only one was created.

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+    end
+  end
+
+  context "fronted" do
+    let(:raw_pending_bank_fee_transaction) {
```
albertchae: This test uses `let` instead. We want to use `let` when we need to both share some setup AND introduce a variable into scope that the test can refer to. The other way of introducing variables into scope is instance variables, but that can leak across tests in messy ways. More reading on this in https://www.codewithjason.com/difference-let-let-instance-variables-rspec/

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      create(:raw_pending_bank_fee_transaction,
+             date_posted: Date.current,
+             amount_cents:)
```
albertchae: What can be really weird about `let`s is that they can refer to variables that aren't defined in the current `context` but are defined in sub `context`s. I generally don't like to abuse this, but the reason I did it here is that it means each of the test cases can pinpoint exactly the field I change to effect the test outcome.

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    before do
+      raw_pending_bank_fee_transaction
+    end
```
albertchae: So `let`'s are lazy by default, which means the `create(:raw_pending_bank_fee_transaction ...` actually doesn't execute unless the variable is referred to somewhere, which is why I refer to it in this before block to realize the object. However, if you read the link https://www.codewithjason.com/difference-let-let-instance-variables-rspec/, you'll see that you can use `let!` to do this as well. Jason recommends "I always use the let! version because I find the execution path to be more easily understandable." which is definitely true, but in larger tests can lead to instantiating objects/creating database records that weren't needed for that specific test, which is often the source of pain with slow test suites. It's common for suites to mix `let` and `let!` which then is a pain for the reader to parse for a single slender `!` to judge what's happening. There is a much spicier take on `let`s in https://thoughtbot.com/blog/lets-not which I mostly agree with but I still haven't completely dropped them and instead try to use this middle ground where I don't blindly instantiate every object and try to explicitly identify which ones do. 🤷

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      it "fronted is false" do
+        expect do
+          described_class.new.run
+        end.to change { CanonicalPendingTransaction.count }.by(1)
+
+        pending_transaction = CanonicalPendingTransaction.last
+        expect(pending_transaction.amount_cents).to eq(amount_cents)
+        expect(pending_transaction).to_not be_fronted
+      end
+    end
```
albertchae: Because of the way I put the shared set up in the outer context pointing at an inner `amount_cents`, I am able to succinctly change the one thing that matters in each of these tests and then see whether we got `fronted` or not

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      let(:amount_cents) { 100 }
+
+      it "fronted is true" do
+        expect do
+          described_class.new.run
+        end.to change { CanonicalPendingTransaction.count }.by(1)
+
+        pending_transaction = CanonicalPendingTransaction.last
+        expect(pending_transaction.amount_cents).to eq(amount_cents)
+        expect(pending_transaction).to be_fronted
```
albertchae: Technically this could be written as `expect(pending_transaction.fronted).to eq(true)`, which might be a good way to start. I believe rubocop might convert those for you automatically. I honestly don't know if the rspec DSL is "better" but I will say it's very commonly used and I'm sad I know it...

spec/models/raw_pending_bank_fee_transaction_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe "RawPendingBankFeeTransaction" do
```
garyhtou: is there a preference/difference between

```ruby
describe "RawPendingBankFeeTransaction" do
end
```

```ruby
describe RawPendingBankFeeTransaction do
end
```

```ruby
describe RawPendingBankFeeTransaction, type: :model do
end
```

e.g. https://github.com/hackclub/bank/blob/a94eea25119bb50fa626a02f15db31cafa8a65d8/spec/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb#L23

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
garyhtou: there's also WET (Write Everything Twice)!

i'm surprised that Testing on the Toilet is still running given that it started in 2007

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
garyhtou: good idea! Should this not null be applied to all `RawPending*Transaction`?

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
garyhtou: oo that's tricky, good to know!

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
garyhtou: hmmm yeah, i understand that `.to be_*` reads better, but it's less intuitive to write/understand. Also, it's more difficult when searching for usages of the `fronted` method.

garyhtou (commented):
ty for this!

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
KiKoS0: i think so, it seems the same improvements mentioned in this PR apply to most if not all `RawPending*Transaction` models/processing.

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
albertchae: oh yeah i forgot about WET, I think that's better than DAMP both as an acronym and as a pun

spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/bank_fee_spec.rb
albertchae: yeah I don't think it's a bad idea if you want to adopt a rubocop rule in HCB that prefers the more explicit version

spec/models/raw_pending_bank_fee_transaction_spec.rb
albertchae: I think there's no difference between doing `"RawPendingBankFeeTransaction"` or `RawPendingBankFeeTransaction`. I personally prefer the latter because I think it'd result in a class loading error sooner than if we use the string, but I think here I got the string from using a RubyMine test generator and didn't realize. I do think some people are against using references to classes instead of using their string literals for performance reasons, but I haven't really looked into that.

I'm actually not really sure what `type: :model` does. I know `:feature` or `:controller` give you more things and also it lets you add more default behavior from `rails_helper` using the spec type.

Rspec docs say this
https://rspec.info/features/6-0/rspec-rails/model-specs/
https://rspec.info/features/6-0/rspec-rails/controller-specs/
https://rspec.info/features/6-0/rspec-rails/request-specs/request-spec/

PR #5731, opened 2024-03-25
Add docker-compose for running postgres-15 without destroying postgres-11 data

albertchae:
- Add an overlay `docker-compose.postgres-12.yml` which allows us to run
  a postgres-12 docker container instead of postgres-11
- Add `--pg-12` to `docker_dev_setup.sh`, which lets devs set up a fresh
  database on postgres 12
- Add `--pg-12` to `docker_start.sh` which lets devs run their web
  container against the postgres 12 container
- Add `upgrade_docker_postgres_11_to_12.sh`, a script which should
  upgrade a dev's database non destructively. We have configured
  the postgres-12 container to use a volume with the version
  (pg-12-data), so we won't destroy the old volume and can roll back if
  necessary. This script dumps the data from the old volume and loads it
  into pg-12-data volume and adds it to the database.


docker-compose.postgres-12.yml
```diff
@@ -0,0 +1,11 @@
+version: '3'
+services:
+  db:
+    image: "postgres:12.16"
```
KiKoS0: Is production running `12.16`? Because the last `12.x` is [`12.18`](https://hub.docker.com/_/postgres).

docker_dev_setup.sh
```diff
@@ -3,6 +3,11 @@
 # installs dependencies with Docker Compose and seeds the development database
 # reach out to Max Wofford (max@hackclub.com) if you have any questions or issues
 
+if [[ $* == *--pg12* ]]
+then
+  export COMPOSE_FILE=docker-compose.yml:docker-compose.postgres-12.yml
```
KiKoS0: I think having it as an override is pretty clever 👍

KiKoS0 (commented):
The migration and start script ran fine for me 🎉 

Just a note to update the readme instructions and command to include `--pg12`.

docker-compose.postgres-12.yml
albertchae: looks like it is 12.17 currently. I'll update
image

albertchae:
note to self: double check dev containers https://github.com/hackclub/hcb/pull/4202#issuecomment-1933574265

garyhtou:
We decided to switch to 15 by nuking postgres 11 data in development and reseeding.
https://github.com/hackclub/hcb/pull/11032

PR #5733, opened 2024-03-25
[`StripeCardPolicy`] Restrict member's permissions

malted:


garyhtou (approved):

garyhtou:
This falls inline with https://hcb.hackclub.com/roles

PR #5738, opened 2024-03-25
`@hcb_code.stripe_card` not `@card`

sampoder:


app/views/hcb_codes/_stripe_card.html.erb
```diff
@@ -86,10 +86,26 @@
...
+          <% elsif reason == "card_inactive" %>
+            <em>This transaction was declined by HCB as this card is frozen.</em>
+          <% elsif reason == "suspected_fraud" || reason == "verification_failed" %>
+            <em>This transaction was declined by HCB due to failed suspected fraud / failed checks.</em>
+          <% elsif reason == "webhook_declined" && webhook_declined_reason == "merchant_not_allowed" %>
+            <em>This transaction was declined by HCB because this merchant is not allowed.</em>
+          <% elsif reason == "webhook_declined" %>
+            <em>This transaction was declined by HCB due to insufficient funds.</em>
+          <% else %>
+            <em>This transaction was declined by the merchant.</em>
```
cjdenio: This isn't right: transactions are always declined by us, not the merchant

sampoder: Oh, alrighty, I'll change it to "This transaction was declined."

PR #5739, opened 2024-03-25
Team Page: Auto overflow on email & phone parent

malted:
Solves a bug where the email and phone runs out of the card

Before:
image

After:
image
Now it horizontally scrolls.


garyhtou (approved):
In the PR description, can you add screenshot of what it looks after?

PR #5741, opened 2024-03-25
[Receipt uploads] Disable + shorten receipt upload button upon successful upload

malted:
image

Blind fix but will definitely work


sampoder (approved):
LGTM!

PR #5742, opened 2024-03-25
[User Roles] Let non-managers see org setting (but disable form)

malted:
Disables all form fields for members

image


app/policies/event_policy.rb
```diff
@@ -165,12 +165,12 @@ def validate_slug?
     admin_or_user?
   end
 
+  private
+
   def admin_or_user?
```
garyhtou: https://github.com/hackclub/bank/blob/34aecab447b76773f0e91fc128350c8530cb55af/app/controllers/stripe_cards_controller.rb#L69 and https://github.com/hackclub/bank/blob/34aecab447b76773f0e91fc128350c8530cb55af/app/controllers/stripe_cards_controller.rb#L74 will break if this is private!!

app/policies/event_policy.rb
```diff
@@ -33,7 +33,7 @@ def by_airtable_id?
   end
 
   def edit?
-    admin_or_manager?
+    admin_or_user?
```
garyhtou: This policy is good. However, on the page, a member can still type into some fields. Please double check that everything's disabled, including form fields

garyhtou (changes requested):

app/policies/event_policy.rb
garyhtou: Fixed by https://github.com/hackclub/hcb/pull/5670

app/policies/event_policy.rb
malted: Everything is disabled!

app/views/events/_form.html.erb
```diff
@@ -19,10 +19,13 @@
       <% if @event.logo.attached? %>
         <%= image_tag @event.logo, height: 72 %>
       <% end %>
-      <div class="field--fileupload">
-        <%= form.label :logo, "Choose File", class: "field--fileupload__label" %>
-        <%= form.file_field :logo, accept: "image/*,image/heic", class: "field--fileupload__field" %>
-      </div>
+
+      <% if policy(@event).remove_logo? %>
```
garyhtou:
```suggestion
      <% if policy(@event).update? %>
```

app/views/events/_form.html.erb
```diff
@@ -37,10 +40,13 @@
       <% if @event.background_image.attached? %>
         <%= image_tag @event.background_image, width: 256 %>
       <% end %>
-      <div class="field--fileupload">
-        <%= form.label :background_image, "Choose File", class: "field--fileupload__label" %>
-        <%= form.file_field :background_image, accept: "image/*,image/heic", class: "field--fileupload__field" %>
-      </div>
+
+      <% if policy(@event).remove_background_image? %>
```
garyhtou:
```suggestion
      <% if policy(@event).update? %>
```

garyhtou (approved):

PR #5743, opened 2024-03-25
[User Roles] Blur Check Deposit images for non-managers

malted:
image


garyhtou (changes requested):
The policies/permissions are good. However,

## Weird styling

The card has a weird border on the bottom.

image

## Leaking image

The image can still be seen by a member on the CheckDeposit index page on organization

garyhtou (dismissed):

garyhtou:
Check Deposit images contain sensitive information such as account & routing numbers. We hide this info members

garyhtou (approved):

PR #5745, opened 2024-03-26
Notification settings! (Early version)

YodaLightsabr:
## Summary of the problem
Notifications!


app/views/users/edit_notifications.html.erb
```diff
@@ -0,0 +1,129 @@
...
+          <%= form.label :donations do %>
+            <%= inline_icon "support", size: 28 %>
+            <strong>Donations</strong>
+          <% end %>
+           <%= form.check_box :reimbursements %>
+          <%= form.label :reimbursements do %>
+            <%= inline_icon "attachment", size: 28 %>
+            <strong>Reimbursements</strong>
+          <% end %>
+        </div>
```
garyhtou: These "radio boxes" are usually used for select-one. In this case, they act more like checkboxes, where you can select multiple.

db/migrate/20240319234849_add_notification_settings.rb
```diff
@@ -0,0 +1,6 @@
+class AddNotificationSettings < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :enabled_notifications, :integer,default: 0, null: false
+    add_column :organizer_positions, :enabled_notifications, :integer,default: 0, null: false
```
garyhtou:
```suggestion
    add_column :users, :enabled_notifications, :integer, default: 0, null: false
    add_column :organizer_positions, :enabled_notifications, :integer, default: 0, null: false
```

YodaLightsabr:
The current notification system isn't super comprehensive, but it works for most people. I'm closing this for now but we can re-evaluate the need for this if we have to.

PR #5749, opened 2024-03-26
[Reimbursements] Set `increase_check` and `ach_transfer`

sampoder:
We weren't doing this for some reason.


app/services/reimbursement/payout_holding_service/nightly.rb
```diff
@@ -24,6 +24,8 @@ def run
               )
               check.save!
               check.send_check!
+              payout_holding.increase_check = check
+              payout_holding.save!
```
garyhtou: use update! and maybe wrap check.save! and payoutholding.update! in a transaction? (probably not necessary)

PR #5760, opened 2024-03-26
User Nav: Make Settings nav item visible again on mobile

garyhtou:
## Summary of the problem

If a user has access to the Reimbursements beta, it adds another item to their user navigation. This additional nav item ends up being "too many" which hide the last item (Settings).

## Describe your changes

image

This commit updates the max number of User nav items on mobile to 5. And makes the nav item names smaller so we can fit more.


garyhtou:
I think a further improvement would be to evenly space the icons. Currently reimbursements takes up more space due to the larger name

garyhtou:
I'm going to merge this so the user settings is accessible again. If anyone is interested, feel free to follow up with another PR to improve this

PR #5766, opened 2024-03-26
Enable send transfer button all the time

malted:


garyhtou:
Was handled in a different PR. there was no diff after updating the branch with main

PR #5767, opened 2024-03-26
[Org Settings] Remove account & routing number

malted:


garyhtou (approved):
Great! It already has it's own dedicate page, so no need to show it here too

PR #5774, opened 2024-03-26
[Termination letters] add safe navigator to admin condition on policy

sarthaktexas:


garyhtou (approved):
you're good to merge small fixes like these without approval, as long as you test them

PR #5776, opened 2024-03-27
[`bin/lint`] Run linters concurrently

garyhtou:
This will save you roughly 16.968 seconds 🤷‍♂️

### Before
```
bin/lint  41.61s user 3.33s system 93% cpu 47.840 total
```

### After
```
bin/lint  43.91s user 3.69s system 154% cpu 30.872 total
```

It runs all the linters using `concurrently` with the `--group` flag so that the output is not interwoven.

---

Any thoughts on this?


garyhtou:
@cjdenio any thoughts on this?

bin/lint
```diff
@@ -1,15 +1,22 @@
...
 set -e
 
-echo "=== Running Rubocop ==="
-bundle exec rubocop -a
-# To run potentially unsafe cops, use: bundle exec rubocop -A
+# Check if "concurrently" command is available
+if [ -x "$(command -v concurrently)" ]; then
+  # Concurrently is installed
+  echo "=== Running Linters with Concurrently ==="
+  concurrently "bundle exec rubocop -a" "exec erblint app/views/ -a" "yarn lint --fix" "yarn format" --raw --group
```
cjdenio: Wouldn't this need to be `yarn run concurrently`?

garyhtou: oh yes, likely.

i had it installed globally

YodaLightsabr (approved):
@garyhtou I tried reviving this, and it seems to work fine on my end. Running Rubocop and ERB Lint simultaneously will be nice!

sampoder (approved):
LGTM - let's do it

PR #5784, opened 2024-03-27
[Reimbursements] Handle Failed ACH Transfers

sampoder:
I'd like an extra pair of eyes on this, I don't know if this is a valid way of doing this. Here's what we're doing:

1) When Column tells us that an ACH fails, we'll send an email to the report creator.
2) We mark their payout holding as `failed`
3) They'll need to update their payout settings, we'll then mark all of their failed payout holdings as "settled"
4) We then reattempt sending all "settled" payout holdings as usual


app/models/user/payout_method/ach_transfer.rb
```diff
@@ -18,6 +18,8 @@ class AchTransfer < ApplicationRecord
       has_encrypted :account_number, :routing_number
       validates :routing_number, format: { with: /\A\d{9}\z/, message: "must be 9 digits" }
       validates :account_number, format: { with: /\A\d+\z/, message: "must be only numbers" }
+      after_save-> { Reimbursement::PayoutHolding.where(report: user.reimbursement_reports).failed.mark_settled! }
```
sampoder: Is this valid?

app/mailers/reimbursement_mailer.rb
```diff
@@ -38,5 +38,13 @@ def expense_unapproved
 
     mail to: @report.user.email, subject: "An update on your reimbursement for #{@expense.memo}", from: hcb_email_with_name_of(@report.event)
   end
+  
+  def ach_failed
+    @payout_holding = params[:reimbursement_payout_holding]
+    @report = @payout_holding.report
+    @reason = params[:reason]
+      
+    mail subject: "[Reimbursements] ACH transfer for #{@report.name} failed to send", to: @report.user.email
```
garyhtou: use the email with name helper thingy

app/models/user/payout_method/ach_transfer.rb
garyhtou: seen valid to me. there's just some weird formatting (missing space in front of `->`)

app/models/user/payout_method/check.rb
```diff
@@ -24,6 +24,8 @@ class Check < ApplicationRecord
       validates :address_state, inclusion: { in: ISO3166::Country.new("US").subdivisions.keys, message: "This isn't a valid US state!", allow_blank: true }
       validates :address_postal_code, format: { with: /\A\d{5}(?:[-\s]\d{4})?\z/, message: "This isn't a valid ZIP code." }
       attribute :address_country, :text, default: "US"
+      after_save-> { Reimbursement::PayoutHolding.where(report: user.reimbursement_reports).failed.mark_settled! }
```
garyhtou: When we support tracking failed states for checks, we have to ensure they can't be cashed again to prevent sending two cash-able checks.

app/models/ach_transfer.rb
```diff
@@ -118,7 +118,12 @@ class AchTransfer < ApplicationRecord
 
     event :mark_failed do
       after do |reason: nil|
-        AchTransferMailer.with(ach_transfer: self, reason:).notify_failed.deliver_later
+        if reimbursement_payout_holding.present?
+          ReimbursementMailer.with(reimbursement_payout_holding:, reason:).ach_failed.deliver_later
```
garyhtou: i'd maybe move this line into PayoutHolding (as a callback for mark_settled)

garyhtou (commented):
🪄ty

PR #5786, opened 2024-03-27
Fix `notify_settled` emails

sampoder:
Sorry about this one @cjdenio! I think I've fixed all the issues but could you help double check for me? Also, do you know a good way to test this locally? Or a better place to put this line?


app/services/canonical_pending_transaction_service/settle.rb
```diff
@@ -18,6 +18,10 @@ def run!
           @canonical_transaction.custom_memo = @canonical_pending_transaction.custom_memo
           @canonical_transaction.save!
         end
+
+        if @canonical_pending_transaction.raw_pending_stripe_transaction && (@canonical_pending_transaction.amount_cents != @canonical_transaction.amount_cents)
```
cjdenio:
```suggestion
        if @canonical_transaction.amount_cents < 0 && @canonical_pending_transaction.raw_pending_stripe_transaction && (@canonical_pending_transaction.amount_cents != @canonical_transaction.amount_cents)
```

I don't think this should trigger for refunds (at least not with the same email copy)

sampoder: cool!

PR #5795, opened 2024-03-27
Add Tailwind + start to replace some Basscss helpers

cjdenio:
Related to #5623 

A lot of Basscss helpers are also Tailwind classes verbatim, or at least very close. For now, classes are manually safelisted to prevent Tailwind from overriding existing classes.


sarthaktexas (commented):
this looks good. can we add tailwind prettier to reorder classes automatically

cjdenio:
Gonna merge this to kick things off. Every change in this PR has been individually reviewed, and I can confirm this makes no unexpected changes to HCB's styling.

re: tailwind prettier— unfortunately we can't format ERB files with Prettier since the syntax breaks Prettier

PR #5802, opened 2024-03-27
[Column Statement] Fix missing last day of month

garyhtou:
## Summary of the problem

Column report for Feb was missing the last day of the month (29th). It wasn't a leap day bug, but because the report for the last day of the month wasn't generated yet.

## Describe your changes

- [x] Check to make sure timestamp stuff works
- [x] Fix `end_date` timestamp in prod for February 2024
- [x] Add check in `ColumnService#transaction` to raise an error if missing report spanning the requested dates

@sampoder can you handle this?


sampoder:
Production data has been fixed:

Screenshot 2024-03-27 at 4 52 00 PM

app/services/column_service.rb
```diff
@@ -49,6 +49,12 @@ def self.transactions(from_date: 1.week.ago, to_date: Date.today, bank_account:
 
     document_ids = reports.pluck "json_document_id"
 
+    dates = reports.pluck "from_date"
+
+    if (from_date.to_date..to_date.to_date).reject { |date| dates.include?(date.to_date.iso8601) }.any?
+      raise StandardError.new("Missing Column reports for #{from_date.to_date.iso8601} to #{to_date.to_date.iso8601}")
+    end
```
garyhtou: we need to double check that is works for the transaction engine import

garyhtou (commented):
🪄

PR #5804, opened 2024-03-27
Refactor `InvoicePolicy`

malted:


app/policies/invoice_policy.rb
```diff
@@ -13,27 +13,27 @@ def index?
   end
 
   def new?
-    !unapproved? && (is_public || admin_or_user)
+    (is_public? || admin_or_user?) && !unapproved?
   end
 
   def create?
-    !record.unapproved? && !record.outernet_guild? && (user&.admin? || record.users.include?(user))
+    admin_or_user? && !record.unapproved?
```
cjdenio: This will error since `admin_or_user?` expects `record` to be an `Invoice`.

one solution: add a common method to `ApplicationPolicy` like
```ruby
class ApplicationPolicy
  # ...

  def admin_or_user?
    event = record.is_a?(Event) ? event : record.try(:event)
    user&.admin? || (user && event && event&.users&.include?(user))
  end
end

**Review by cjdenio — CHANGES_REQUESTED** [review]

File: app/policies/invoice_policy.rb [hunk 2]
```diff
@@ -13,27 +13,27 @@ def index?
   end
 
   def new?
-    !unapproved? && (is_public || admin_or_user)
+    (is_public? || admin_or_user?) && !unapproved?
```
cjdenio: `is_public?` isn't defined

app/policies/invoice_policy.rb
garyhtou: `event = record.is_a?(Event) ? event : record.try(:event)` is slow. let's just define a method in InvoicePolicy — we'll make a better shared solution later

app/policies/invoice_policy.rb
malted: Fixed!

garyhtou:
after a review of this pr on call with @malted, we decided this is no longer needed

PR #5814, opened 2024-03-28
[Login Code Exchanges] Rate limit to 3 per-email per-minute

sampoder:
Fixes https://github.com/hackclub/hcb/issues/5809


garyhtou (changes requested):
Is the twilio rate limit for requesting codes? or exchanging codes? or both? please read their docs on this

this will would fail Availability in the CIA triad. in fact, i think it makes it worse. A possible solution is to rate limit with the key being IP (and optionally with email address too)

sampoder:
That was my understanding in the call though I may have been totally wrong. That said, I think I have a better idea for handling this so I'm going to close this and implement that tomorrow. Basically the idea is to fallback to emails.

PR #5816, opened 2024-03-28
Remove Unused files and helpers referencing SVB

aramshiva:
Fixes #4984.
Removes unused SVB files/helpers


aramshiva:
I'll do some extra checks once preview is deployed as don't want to make sure I broke something.

aramshiva:
ScreenShot 2024-03-27 at 22 25 09@2x
@garyhtou can you provide logs?

garyhtou:
@cjdenio when you get a chance, could you review this?

app/services/transaction_engine/transaction_csv_service/upload.rb
```diff
@@ -10,10 +10,6 @@ def initialize(transaction_csv_id:)
       def run
         transaction_csv.mark_processing!
 
-        transaction_csv.file.open do |f|
```
cjdenio: This service should be completely removed— deleting these three lines just makes it do nothing

cjdenio (changes requested):
I see the test for `Parsers::SiliconValleyBank` was removed, but the service itself was not— we should be removing both at the same time

cjdenio (approved):

PR #5821, opened 2024-03-28
Donation page: show custom error message on 3DS failure

sampoder:
Closes https://github.com/hackclub/hcb/issues/5820. There's probably better copy for this, thoughts?


app/javascript/controllers/stripe_controller.js
```diff
@@ -47,8 +47,11 @@ export default class extends Controller {
         if (this.hasErrorsTarget) {
           const flash = document.createElement('p')
           flash.classList.add('flash', 'error', 'fit', 'mt1', 'mb3')
-          flash.textContent = `Something went wrong: ${result.error.message}`
-
+          if (result.error.code == 'payment_intent_authentication_failure') {
```
cjdenio: We should also handle `card_declined`, which triggers if a bank's 3DS integration is down

cjdenio (dismissed):

app/javascript/controllers/stripe_controller.js
sampoder:
```suggestion
          if (result.error.code == 'payment_intent_authentication_failure' || result.error.code == 'card_declined') {
```
cool yeah

sampoder:
Going to merge this but if folks have more copy edits, please feel free to add them!

PR #5840, opened 2024-03-28
[Reimbursements] Assign Reviewers to Reports

sampoder:
Requested by Mel. Closes https://github.com/hackclub/hcb/issues/5827.

For example, if an events intern, wants reimbursement, they should assign Dev for review. If it's a fundraising intern, assign Christina etc.

Screenshot 2024-03-28 at 2 36 44 PM
Screenshot 2024-03-28 at 2 37 01 PM
Screenshot 2024-03-28 at 2 37 09 PM


app/views/reimbursement/reports/_edit_form.html.erb
```diff
@@ -19,6 +19,10 @@
       <span class="muted">You can transfer a report to any organization you're a part of.</span>
     <% end %>
   </div>
+  <div class="field">
+    <%= form.label :reviewer_id, "Assigned Reviewer", class: "mt2 bold" %>
+    <%= form.collection_select :reviewer_id, @report.event.users, :id, :name, include_blank: "Anyone" %>
+  </div>
```
garyhtou: When should this field be disabled?

garyhtou: Should this field be available to non-organizers? It could be used as an exploit to get a list of users for a non-transparent org.

garyhtou (commented):
Two things:
1. When should the reviewer field be disabled?
    - I likely shouldn't be able to update that after organizer approval.
3. The audit log looks wonky
    image
    you can fix this in another PR if you'd like

🪄 go ahead and merge after fixing and testing

app/views/reimbursement/reports/_edit_form.html.erb
sampoder: The field shouldn't be visible if the report is closed, so we should be good

sampoder: yeah good idea!

PR #5843, opened 2024-03-28
[User Settings] Only do saved changes check if they have a payout method

sampoder:


sampoder:
Screenshot 2024-03-28 at 3 35 57 PM

fixed ^

PR #5847, opened 2024-03-28
[User Roles] Disable card actions for non-managers

malted:
Also adds tooltips:
image

We also hide the activate button if you don't have permission to activate this card


app/views/stripe_cards/actions/_freeze.html.erb
```diff
@@ -1,22 +1,32 @@
...
+                disabled: !policy(stripe_card).defrost? %>
+  </div>
 
 <% elsif stripe_card.stripe_status == 'active' %>
-  <span class="tooltipped tooltipped--n"
-        aria-label="You can instantly defrost
-                    <%= stripe_card.user == current_user ? "your" : stripe_card.user.possessive_name %>
-                    card anytime">
-
+  <% freeze_tooltip = if policy(stripe_card).defrost?
```
garyhtou:
```suggestion
  <% freeze_tooltip = if policy(stripe_card).freeze?
```

garyhtou (approved):

PR #5851, opened 2024-03-29
[Column Statements] Add start / closing balances

sampoder:
Closes https://github.com/hackclub/hcb/issues/5849

@garyhtou can you sanity check this? Specifically that we're choosing the right `first_txn` and `last_txn`


app/jobs/column/statement_job.rb
```diff
@@ -65,6 +65,10 @@ def perform(date = Date.today.prev_month)
         column_statement.file.attach(io: File.open(file), filename: "column_statement_report_#{end_date.iso8601}.csv")
         column_statement.start_date = start_date
         column_statement.end_date = end_date
+        first_txn = transactions_by_report.last.transactions.last
+        last_txn = transactions_by_report.first.transactions.first
```
garyhtou: I believe Ruby hashes don't guarantee order. So `last` and `first` are non-deterministic.

sampoder:
```suggestion
        first_txn = transactions_by_report.last.transactions.first
        last_txn = transactions_by_report.first.transactions.last
```

garyhtou (commented):
🪄 let's merge this to get something in. however, we'll rewrite this job in a few weeks

PR #5853, opened 2024-03-29
[Notify Settled Emails] Refactor mailer to take a `canonical_transaction_id` parameter

sampoder:
Both methods work when testing locally, see tests on https://github.com/hackclub/hcb/pull/5818 and this test I did on this branch:

Screenshot 2024-03-29 at 9 12 46 AM

However, right after the job ran, `@ct` was undefined in the mailer: https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/603/samples/timestamp/2024-03-29T06:02:25Z.

We were previously getting `@ct` using `@cpt.local_hcb_code.ct`. My theory, and I may be wrong, is that for some transactions, they aren't on the hcb_code yet so `@cpt.local_hcb_code.ct` returns nil. So instead, I'm now having the `canonical_transaction_id` passed from `CanonicalPendingTransactionService::Settle` to the mailer.

Could someone sanity check the code in `CanonicalPendingTransactionService::Settle` for me?


sampoder:
I'm going to merge this so it doesn't error again tonight, I've double checked this and I'm pretty confident in it

sampoder:
I'm going to merge this so it doesn't error again tonight, I've double checked this and I'm pretty confident in it

PR #5859, opened 2024-03-29
Card activation button should be a link, not a <form>

cjdenio:


garyhtou:
@cjdenio  should we change the one on line 19?

cjdenio:
So I was going to change that instance, but that whole card replacement flow is broken anyway lol 🤷

PR #5860, opened 2024-03-30
[`notify_settled`] Switch to `after_create_commit` & remove from DB transaction

sampoder:
Gets around the problem of the CanonicalTransaction not being saved in the database when we generate the email.


app/services/canonical_pending_transaction_service/settle.rb
```diff
@@ -22,7 +22,7 @@ def run!
         if @canonical_transaction.amount_cents < 0 && @canonical_pending_transaction.raw_pending_stripe_transaction && (@canonical_pending_transaction.amount_cents != @canonical_transaction.amount_cents)
           CanonicalPendingTransactionMailer.with(
             canonical_pending_transaction_id: @canonical_pending_transaction.id,
-            canonical_transaction_id: @canonical_transaction.id,
+            settled_amount_cents: @canonical_transaction.amount_cents
           ).notify_settled.deliver_later
         end
       end
```
garyhtou: This here really should be outside of the transaction. there's no need to have this inside — especially since and error will cause it to not settle (because the transaction gets rolled back)

garyhtou (commented):
Interesting, i'd expect for the CT to have already been committed to the database. but I guess this would likely fix any problems we have.

One feedback. then, please test and 🪄

cjdenio (commented):
Change
https://github.com/hackclub/hcb/blob/79086d23c6c8d822bfe0b8e25e8de95d5501ada2/app/models/canonical_transaction.rb#L103

to `after_create_commit`– this should solve the problem in a more maintainable way

sampoder:
Changed things up, thanks you both!

PR #5863, opened 2024-03-30
[Receipts] Enable Text-to-Receipt for non-Receipt Bin uses

sampoder:
I realised that people who don't have the receipt bin feature preview can't use the new flow for uploading receipts via SMS even though the features doesn't rely on receipt bin. I've done a slight refactor to make this work, @YodaLightsabr thoughts?


YodaLightsabr (approved):
This looks good!

PR #5866, opened 2024-03-31
Implement spec for PendingTransactionEngine::CanonicalPendingTransactionService::Import::OutgoingCheck

KiKoS0:
- Creates a factory for `RawPendingOutgoingCheckTransaction` and a basic model spec for it.
- Implements specs for `PendingTransactionEngine::CanonicalPendingTransactionService::Import::OutgoingCheck` class.

The same improvements from https://github.com/hackclub/hcb/pull/5730 can be applied to the `Import::OutgoingCheck` class.

I have a question regarding the last commit which [refactors the check association to use belongs_to instead of find_by](https://github.com/hackclub/hcb/commit/0fc78d32fdab6f87fd9c48fb99a009ce8622207f). 

Given that the type of the foreign key (`check_transaction_id`) is a string (`varchar`) and not an integer (`bigint`) like the id of the check table and the fact that there is no foreign key constraint between the two tables, I'm not sure if the usage of `find_by` instead of an ActiveRecord association is intentional or not. Because if it's not then the refactor is valid and it might even be a good idea to eventually reconcile the column types and possibly tighten the association by adding the database constraint (although extremely more involved especially if this pattern is repeated in many other models)

Let me know if I'm missing something here or if you can describe the flow of what model is created first (`check` or `raw_pending_outgoing_check_transaction`) before getting processed. Thank you!


sampoder:
Hey @KiKoS0! I'm going to close this PR, we no longer actively use the `Check` model and have transitioned to the `IncreaseCheck` model / sending checks through Column.

PR #5867, opened 2024-03-31
🚀 🌕 HCB: To The Moon (April Fools 2024)

sampoder:
April Fools?

image

image

image

<details>
<summary>old images</summary>

Screenshot 2024-03-31 at 3 54 07 PM
Screenshot 2024-03-31 at 3 54 43 PM

</details>


app/controllers/users_controller.rb
```diff
@@ -290,6 +290,19 @@ def disable_feature
     redirect_back fallback_location: settings_previews_path
   end
 
+  def to_the_moon
```
garyhtou: LOL I LOVE THE NAME

app/helpers/application_helper.rb
```diff
@@ -23,6 +23,15 @@ def render_money(amount, opts = {})
     end
   end
 
+  def number_to_currency(num, opts = {})
+    if @current_user && Flipper.enabled?(:april_fools_2024_04_01, @current_user)
+      opts[:unit] = !opts[:unit] || opts[:unit] == "$" ? "Ɖ" : opts[:unit]
```
garyhtou: I'm having a hard time understanding this line.

This logic seems like it will _always_ set unit to `"Ɖ"`

app/assets/stylesheets/components/_stats.scss
```diff
@@ -8,7 +8,7 @@
   margin-bottom: 0.25rem;
 }
 .stat:not(.stat--plain) .stat__value:before {
-  content: '$';
+  content: 'Ɖ';
```
garyhtou: Should we feature flag this in some way?

Easiest way might be to conditionally render CSS that overrides this

garyhtou: Oops, looks like you just did exactly that!

app/helpers/application_helper.rb
cjdenio: It's a little confusing, basically this reads (IF unit is unset OR unit is "$", set unit to "Ɖ")

both sides of the OR could be parenthesized for clarity

cjdenio: but the Flipper check above should prevent this from happening in all cases

sampoder: yep! Caleb's right
```suggestion
      opts[:unit] = (!opts[:unit]) || (opts[:unit] == "$") ? "Ɖ" : opts[:unit]
```

garyhtou:
it would be cool if the `body` had a subtle background of moon craters

garyhtou (approved):
**🚀 TO THE MOON 🌕**

LET'S MERGE THIS!

PR #5868, opened 2024-03-31
[HcbCode] Add a rich metadata view for Reimbursements Payout Holding

sampoder:
It'd be nice to have a breakdown of each of these!


app/views/hcb_codes/_reimbursement_payout_holding.html.erb
```diff
@@ -0,0 +1,49 @@
...
+          <strong>ACH Transfer</strong>
+          <%= link_to "View Transfer", @hcb_code.reimbursement_payout_holding.ach_transfer %>
+        </p>
+      <% end %>
+      <% if @hcb_code.reimbursement_payout_holding.increase_check %>
+        <p>
+          <strong>Check</strong>
+          <%= link_to "View", @hcb_code.reimbursement_payout_holding.increase_check %>
+        </p>
+      <% end %>
```
garyhtou: Most people won't have access to these links, right?

garyhtou: Maybe we use `link_to_if`

sampoder: This will only be for admins / members of clearinghouse so everyone should have access :D

PR #5870, opened 2024-04-01
Implement specs for the rest of CanonicalPendingTransactionService classes

KiKoS0:
Implements specs for the rest of the classes in this scope `PendingTransactionEngine::CanonicalPendingTransactionService::*` following the example in https://github.com/hackclub/hcb/pull/5730. Again most of the improvements mentioned in that PR also apply to these classes so I'll follow up a refactor PR after this one.


spec/services/pending_transaction_engine/canonical_pending_transaction_service/import/outgoing_ach_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+  let(:ach_transfer) {
+    create(:ach_transfer,
+           event:,
+           amount: 1000,
+           aasm_state: "deposited",
+           recipient_email: "example@example.com",
+
+           # A workaround to avoid that the ach_transfer record creates its own raw_pending_outgoing_ach_transaction
+           # and canonical_pending_transaction record when running the `after_create` callback.
+           scheduled_on: Date.current + 1.day)
```
KiKoS0: The alternative is to extract the `after_create` behavior to a method and stub it which isn't necessarily better imo but let me know if you think it is.

https://github.com/hackclub/hcb/blob/adc09287f3ab19cf247f865fa40a501a4fd37f31/app/models/ach_transfer.rb#L134-L142

albertchae: I think your workaround is fine enough for now

albertchae (approved):

PR #5882, opened 2024-04-02
[April Fools 2024] 🌎 Back to Earth (Reverts #5867 and #5874)

garyhtou:
it's no longer april fools! it's time to return back to earth


garyhtou:
After merging, we'll also delete the flipper flag from prod

PR #5673, opened 2024-03-21
Create `callout` partial and use for unauthorized transfers

malted:
From:
image

To:
image

This brings them in line with the rest of HCB's design system.

This PR also adds the following callout for members on transfer forms:
image

PR #5886, opened 2024-04-02
[Reimbursements] Contact operations to migrate from Expensify

sampoder:
image


malted:
```erb
<%= render "callout", title: "Looking for Expensify?", footer: :questions do %>
	<p class="m0">We have our own reimbursement tool now, which we recommend using instead! #{link_to "Learn more here" foo_path}</p>
<% end %>
```

Depends on https://github.com/hackclub/hcb/pull/5673

garyhtou (approved):

garyhtou (changes requested):
oo one sec

garyhtou (approved):

PR #5893, opened 2024-04-03
[User Roles] Disable OPDR button as per policy

malted:
image

Also refactors `OrganizerPositionPolicy::change_position_role?`


app/policies/organizer_position_policy.rb
```diff
@@ -19,11 +19,19 @@ def toggle_signee_status?
...
+  end
+
+  def can_remove?
+    admin_or_manager?
+  end
+
+  private
+
+  def admin_or_manager?
+    user.admin? || OrganizerPosition.find_by(user:, event: record.event)&.manager?
```
garyhtou: not necessary, but we tend to call similar methods in transparency mode
```suggestion
    user&.admin? || OrganizerPosition.find_by(user:, event: record.event)&.manager?
```

garyhtou (approved):

PR #5902, opened 2024-04-03
[Reimbursements] Disable report creation if a user isn't a member of any events

sampoder:
Closes https://github.com/hackclub/hcb/issues/5899


garyhtou:
ah i see. what are your thoughts on hiding the whole nav dock item if there aren't any orgs

PR #5914, opened 2024-04-04
[Reimbursements] Provide Fallback for Email Invites

sampoder:
Edge case where organiser uses the public form.


sampoder:
just gotta check this works locally

sampoder:
Screenshot 2024-04-03 at 6 47 25 PM

fixes it!

garyhtou:
wasn't this fix for an invite, not an expense approval?

sampoder:
Sorry yep! Just retested, confirmed it did fix the bug:

Screenshot 2024-04-03 at 11 20 58 PM

PR #5921, opened 2024-04-04
[Stripe Card] Add max name length validation and flash errors

garyhtou:
## Summary of the problem

image

## Describe your changes

max length is 40. this number was chosen because all reasonable names in prod were shorter than that. It also doesn't overflow the card.

I'll be truncating names over 40 in pro.


garyhtou:
```ruby
irb(main):001:0> StripeCard.where("length(name) > 40").pluck(:id)
=> [1137, 1093, 741, 2837, 1108, 1817, 3096, 1652]
irb(main):002:0> StripeCard.where("length(name) > 40").update_all("name = left(n
ame, 40)")
=> 8
irb(main):003:0> StripeCard.where(id: [1137, 1093, 741, 2837, 1108, 1817, 3096,
1652]).pluck(:id, :name)
```

I initally tried truncating the name in Ruby using `name[...40]` — although that worked in development — it timed out/failed on Heroku. So I had to use the `left` SQL function instead.

Something that is important to mention that `length` and `left` in SQL likely behave different than Ruby when working with unicode characters

PR #5932, opened 2024-04-05
[MarkdownService] Only attempt a user-autolink if `@record` is defined

sampoder:


sampoder:
https://hackclub.slack.com/files/USNPNJXNX/F06T45NG6CU/screenshot_2024-04-05_at_10.49.32___am.png

garyhtou (changes requested):
I think we should rescue instead. Some policies may allow for a `nil` record

sampoder:
My understanding is that 

Pundit.policy(u, @record)

is looking for the policy related to @record & then checking if u has access. So the bug here was that if @record was nil, we don't have a policy for nil items

garyhtou:
ah shit, right

PR #5940, opened 2024-04-05
Don't say an expense is too much when it's equal to the limit

northeastprince:
A limit is the max... not over the max...


sampoder (approved):
Thank you! Sorry about that.

PR #5943, opened 2024-04-06
[`/admin/hq_receipts`] Sort Page By Number of Missing Receipts

sampoder:


sampoder:
I'm going to merge this. It's slow but not unusably slow. My thinking is that end of the day it's still faster than clicking through all of the pages. Most of the slow down is actually getting the memos of the thousands of transactions rendered on the first page.

PR #5944, opened 2024-04-06
[HQ Receipts] Remove Pagination

sampoder:
In production, this was a lot slower than in development. Unfortunately, I don't think this ordering is super feasible right now.


sampoder:
Screenshot 2024-04-05 at 11 05 36 PM

We got a Heroku timeout

PR #5947, opened 2024-04-06
[Mock Data] Disable popover when showing mock data

garyhtou:
Closes https://github.com/hackclub/hcb/issues/5946

## Summary of the problem

Mock Data transactions (hcb_codes) don't have `pretty_title` defined. It's used to render the popover content.

## Describe your changes

We don't need popovers for mock data, so i disabled the popover


garyhtou:
FYI @sampoder. I think this bug occurred from

```ruby
@event || ct.local_hcb_code.event
```

Previously it was just `ct.local_hcb_code.event`. The `local_hcb_code` came from mock data (is an OpenStruct) and did not have `event` defined on it. Thus, the popover was hidden. But after adding `@event` to it, it attempted to render the popover content, and thus called `pretty_title` on an OpenStruct

PR #5953, opened 2024-04-08
[Reimbursements] Run the payouts job every hour

sampoder:
Closes https://github.com/hackclub/hcb/issues/5952.


garyhtou (approved):
LGTM. we should maybe add a callback on the model to do it instantly

PR #5955, opened 2024-04-08
[Reimbursements] Support Mileage Reimbursements

sampoder:
Currently only the model implementation.


sampoder:
here's a screen recording! this is ready for review


https://github.com/hackclub/hcb/assets/39828164/f72d0e12-ee0e-4f9e-9ab8-2764dee42b50

app/controllers/reimbursement/expenses_controller.rb
```diff
@@ -121,7 +121,9 @@ def actions_turbo_stream
     end
 
     def replace_expense_turbo_stream
-      turbo_stream.replace(@expense, partial: "reimbursement/expenses/expense", locals: { expense: @expense })
+      turbo_stream.replace(@expense, partial: "reimbursement/expenses/expense", locals: {
+                             expense: @expense.becomes(@expense.type&.constantize || Reimbursement::Expense)
```
garyhtou: i think this works?
```suggestion
                             expense: @expense.becomes(@expense.class)
```

app/models/reimbursement/expense.rb
```diff
@@ -11,6 +11,8 @@
 #  description             :text
 #  expense_number          :integer          default(1), not null
 #  memo                    :text
+#  type                    :string
+#  value                   :integer          default(0), not null
```
garyhtou: can make this a decimal field?

app/models/reimbursement/expense.rb
```diff
@@ -86,6 +92,29 @@ def rejected?
...
+    def rate
+      1
+    end
+
+    def set_amount_cents
+      self.amount_cents = rate * value
+    end
+
+    def is_standard?
+      type.nil? || type == "Reimbursement::Expense"
```
garyhtou: If Rails auto-fills normal reimbursements with `"Reimbursement::Expense"` on create. Then, instead of having a check for `type.nil?`, we should:
1. Backfill the column
2. Make the column not-null
3. remove this `type.nil?` check

app/models/reimbursement/expense.rb
```diff
@@ -96,5 +125,11 @@ def status_color
       "success"
     end
 
+    def valid_expense_type
+      unless type.nil? || [Reimbursement::Expense.name, Reimbursement::Expense::Milage.name].include?(type)
```
garyhtou: hopefully we can get rid of the `type.nil?` because of the comment above

app/models/reimbursement/expense.rb
```diff
@@ -96,5 +125,11 @@ def status_color
...
+        errors.add(:type, "is an invalid expense type.")
```
garyhtou:
```suggestion
        errors.add(:type, "must be a valid expense type.")
```

app/models/reimbursement/expense.rb
```diff
@@ -86,6 +92,29 @@ def rejected?
       report.rejected? || pending? && report.closed?
     end
 
+    # multiplier for value -> amount_cents
+    def rate
+      1
+    end
+
+    def set_amount_cents
+      self.amount_cents = rate * value
```
garyhtou:
```suggestion
      self.amount_cents = (rate * value).round
```

app/views/reimbursement/expenses/_form.html.erb
```diff
@@ -2,19 +2,25 @@
...
+      <div>
+        <%= form.label :amount, "Amount", class: "bold" %>
+        <%= form.number_field :amount, placeholder: "500.00", step: 0.01, min: 0.01, max: [expense.report.maximum_amount_cents, 500_000].compact_blank.min, data: { controller: "truncate-decimal", action: "truncate-decimal#truncate", "expense-form-target": "field" }, onkeydown: "event.key == 'Enter' && event.target.form.requestSubmit()" %>
+      </div>
+    <% else %>
+      <div class="ml1">
+        <%= form.label :value, expense.value_label, class: "bold" %>
+        <%= form.number_field :value, placeholder: "500.00", step: 0.01, min: 0.01, max: 500_000 * expense.rate, data: { controller: "truncate-decimal", action: "truncate-decimal#truncate", "expense-form-target": "field" }, onkeydown: "event.key == 'Enter' && event.target.form.requestSubmit()" %>
+      </div>
+    <% end %>
```
garyhtou: Make it so that we don't need additional logic for a standard expense? Have it use the `value` field, with a rate of `100` (to convert dollars to cents)

app/views/reimbursement/expenses/_expense.html.erb
```diff
@@ -33,6 +33,9 @@
           <% end %>
           <div style="flex-grow: 1" data-expense-form-target="memo" class="bold <%= "muted" unless expense.memo.presence %>">
             <%= expense.memo.presence || "Untitled Expense" %>
+            <% unless expense.is_standard? %>
+              (<%= render_money expense.amount %>)
+            <% end %>
```
garyhtou:
```suggestion
            <%= expense.card_label.presence || "Untitled Expense" %>
```

garyhtou: Or consider making it say "loading..." since it immediately gets overridden by stimulus

app/models/reimbursement/expense/milage.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Reimbursement
+  class Expense
+    class Milage < ::Reimbursement::Expense
```
garyhtou:
```suggestion
    class Mileage < ::Reimbursement::Expense
```

app/models/reimbursement/expense/milage.rb
```diff
@@ -0,0 +1,17 @@
...
+      def rate
+        67
+      end
+
+      def value_label
+        "Miles"
```
garyhtou:
```suggestion
        "Miles (67¢/mile)"
```

app/controllers/reimbursement/expenses_controller.rb
garyhtou: Does rails have method that converts it to the `type` without needing to pass it in as a param

app/models/reimbursement/expense/milage.rb
sarthaktexas: LMFAO

app/models/reimbursement/expense.rb
sampoder: it doesn't sadly

app/controllers/reimbursement/expenses_controller.rb
sampoder: this unfortunately doesn't work :/

app/views/reimbursement/expenses/_expense.html.erb
```diff
@@ -68,13 +65,13 @@
                   data: { turbo: true, "menu-target": "toggle", action: "menu#toggle click@document->menu#close keydown@document->menu#keydown" } %>
               <div class="menu__content menu__content--2 h5 menu__content--compact" data-menu-target="content" style="z-index: 999;">
                 <% unless expense.is_standard? %>
-                  <%= link_to reimbursement_expense_path(expense.id, reimbursement_expense: { type: "Reimbursement::Expense" }), data: { turbo: true, "turbo-method": :patch, "action": "menu#close" } do %>
+                  <%= link_to reimbursement_expense_path(expense.id, reimbursement_expense: { type: "Reimbursement::Expense", value: 0, amount_cents: 0 }), data: { turbo: true, "turbo-method": :patch, "action": "menu#close" } do %>
```
garyhtou:
```suggestion
                  <%= link_to reimbursement_expense_path(expense.id, reimbursement_expense: { type: "Reimbursement::Expense", value: 0 }), data: { turbo: true, "turbo-method": :patch, "action": "menu#close" } do %>
```

app/views/reimbursement/expenses/_expense.html.erb
```diff
@@ -68,13 +65,13 @@
...
                     <%= inline_icon "payment-docs", size: 24 %>
                     Revert to a standard expense
                   <% end %>
                 <% end %>
-                <% unless expense.type == "Reimbursement::Expense::Milage" %>
-                  <%= link_to reimbursement_expense_path(expense.id, reimbursement_expense: { type: "Reimbursement::Expense::Milage" }), data: { turbo: true, "turbo-method": :patch, "action": "menu#close" } do %>
+                <% unless expense.type == "Reimbursement::Expense::Mileage" %>
+                  <%= link_to reimbursement_expense_path(expense.id, reimbursement_expense: { type: "Reimbursement::Expense::Mileage", value: 0, amount_cents: 0 }), data: { turbo: true, "turbo-method": :patch, "action": "menu#close" } do %>
```
garyhtou:
```suggestion
                  <%= link_to reimbursement_expense_path(expense.id, reimbursement_expense: { type: "Reimbursement::Expense::Mileage", value: 0 }), data: { turbo: true, "turbo-method": :patch, "action": "menu#close" } do %>
```

garyhtou: Also, let's remove `amount_cents` from the permitted params list

app/views/reimbursement/expenses/_form.html.erb
```diff
@@ -11,16 +11,11 @@
...
-      </div>
-    <% else %>
-      <div class="ml1">
-        <%= form.label :value, expense.value_label, class: "bold" %>
-        <%= form.number_field :value, placeholder: "500.00", step: 0.01, min: 0.01, max: 500_000 * expense.rate, data: { controller: "truncate-decimal", action: "truncate-decimal#truncate", "expense-form-target": "field" }, onkeydown: "event.key == 'Enter' && event.target.form.requestSubmit()" %>
-      </div>
     <% end %>
+    <div class="ml1">
+      <%= form.label :value, expense.value_label, class: "bold" %>
+      <%= form.number_field :value, placeholder: "500.00", step: 0.01, min: 0.01, max: 500_000 * expense.rate, value: number_with_precision(expense.value, precision: 2), data: { controller: "truncate-decimal", action: "truncate-decimal#truncate", "expense-form-target": "field" }, onkeydown: "event.key == 'Enter' && event.target.form.requestSubmit()" %>
```
garyhtou: no need for a max, and this formula doesn't make sense in the context of mileage
```suggestion
      <%= form.number_field :value, placeholder: "500.00", step: 0.01, min: 0.01, value: number_with_precision(expense.value, precision: 2), data: { controller: "truncate-decimal", action: "truncate-decimal#truncate", "expense-form-target": "field" }, onkeydown: "event.key == 'Enter' && event.target.form.requestSubmit()" %>
```

garyhtou: try using the `ctrl_enter_submit` data behavior

app/controllers/reimbursement/expenses_controller.rb
```diff
@@ -101,7 +101,7 @@ def destroy
     private
 
     def expense_params
-      params.require(:reimbursement_expense).permit(:amount, :memo, :description, :reimbursement_report_id, :event_id).compact_blank
+      params.require(:reimbursement_expense).permit(:amount, :value, :memo, :description, :reimbursement_report_id, :event_id, :type).compact_blank
```
garyhtou:
```suggestion
      params.require(:reimbursement_expense).permit(:value, :memo, :description, :reimbursement_report_id, :event_id, :type).compact_blank
```

garyhtou (commented):
🪄

PR #5960, opened 2024-04-10
Create new organizers with default `is_signee: false`

rluodev:
## Summary of the problem
We were creating new organizers and organizer position invites with `is_signee: null` because we didn't have a default value.

## Describe your changes
This will set the default value for newly created `organizer_position`s to `false` instead of `null`


rluodev:
@garyhtou why is the rspec failing? I thought I set the column default to false instead of null for new records

sampoder:
I just pushed a fix to RSpec for you, the problem was that the initialize function was defaulting to `nil`

app/services/event_service/create_demo_event.rb
```diff
@@ -30,7 +30,7 @@ def run
 
         @event.save!
 
-        OrganizerPositionInviteService::Create.new(event: @event, sender: @point_of_contact, user_email: @email, initial: true).run!
+        OrganizerPositionInviteService::Create.new(event: @event, sender: @point_of_contact, user_email: @email, initial: true, is_signee: true).run!
```
sampoder: When someone signs up on our website and we make a demo event for them do we want them to be a signee or do we want them to not be one? @sarthaktexas thoughts?

rluodev: @sampoder one thing I've noticed with organizations is that while one user might create the account it might sometimes be another user that they invite who actually signs the contract

rluodev: It might be best to keep this as false for that reason

rluodev:
I literally just realized this branch is named hella weirdly

mustve forgot to switch branches before committing

garyhtou:
I think we don't want to change null to false yet. @RhysMP and the team is working on backfilling data. We need to `null`s to know which ones haven't been backfilled yet

rluodev:
@garyhtou I'm helping with this too - the statuses of all orgs is in a spreadsheet; is there a particular reason we need to track this using nulls on HCB as well?

garyhtou:
> @garyhtou I'm helping with this too - the statuses of all orgs is in a spreadsheet; is there a particular reason we need to track this using nulls on HCB as well?

Ah, good to know. I just wanted to make sure that we're tracking this information, reliably, somewhere. i do think tracking this information on HCB would be more reliable. would you mind sharing the spreadsheet with me?

rluodev:
shared!

RhysMP:
imo I think tracking the data on HCB is better since it means we wouldn't miss anything. Though it is an add of a couple hours since the backfill button was not created until a bit into this project.

rluodev:
Ok I'm going to remove the migration but keep the part where we default false creating organizers

PR #5961, opened 2024-04-10
Remove `creator_bar` & add card network to card page

malted:
image


app/views/stripe_cards/_stripe_card.html.erb
```diff
@@ -1,7 +1,6 @@
 <a class="text-decoration-none" href="<%= !show_mock_data? ? (local_assigns[:href] || stripe_card_path(stripe_card)) : "javascript:;" %>">
 
-  <%= creator_bar stripe_card, prefix: "issued " unless defined? headless %>
```
cjdenio: I think the creator bar is still useful on pages like hcb.hackclub.com/hq/cards— could you bring this back and set `headless` in `stripe_cards/show` instead?

cjdenio (changes requested):

cjdenio (approved):

PR #5966, opened 2024-04-10
Add flex-wrap to cards shipping header

sampoder:
Fixes https://github.com/hackclub/hcb/issues/5964


sampoder:
Screenshot 2024-04-10 at 12 28 56 PM

PR #5969, opened 2024-04-11
Change org settings icon

malted:
This PR makes it the same as the user settings icon as per @dwahbe's feedback

image


sampoder (approved):

maxwofford:
tbh I think this is even more confusing now– I just looked down at the nav and clicked settings because I wasn't looking closely, but they take me to different pages. Is there a ticket we can discuss this in?

PR #5970, opened 2024-04-11
Change transactions icon to list

malted:
As per @dwahbe's feedback.

image


sampoder:
I feel like the old transaction logo makes more sense, no?

malted:
How so?

sampoder:
It communicates what type of items I'll be seeing on the page and it also looks nicer. A list icon could be used for anything so it isn't telling me much (other than that there will be a list of transactions, which is implied by it being plural imo, and is a bit bland.

sarthaktexas:
i'm w sam

malted:
hm the thing that doesn't make sense for me is the arrows around the dollar sign with the current one - circular arrows make absolutely no sense here.

dwahbe:
I think the tricky thing using that icon for the Transactions tab is that we also use it to convey that a donation is a recurring donation.

Screenshot 2024-04-29 at 12 57 25 PM

garyhtou:
Neither are the right icon. Let's design a new icon, if necessary

PR #5988, opened 2024-04-12
[Mailers] Use `email_address_with_name` everywhere

sampoder:
Closes https://github.com/hackclub/hcb/issues/5520. Uses it wherever feasible (ie. we're emailing a user).

I don't think I missed any instances but give me a ping if I did.

I tested this by looking at the mailer previews for each of these which all rendered as expected.

There are four cases where I couldn't look at the previews:

Login codes: I tested this by requesting a login code in dev. Worked as expected.

`notify_declined` on CPT mailer: this implementation mirrors what we did in the other two methods in this file

grant_mailer and check_deposit_mailer don't have previews so I'm going to make those in a separate PR and then test


sampoder:
Tested the remaining ones, all good!

YodaLightsabr (approved):
Looks good.

PR #5995, opened 2024-04-12
Create a weekly job to delete stale playground events

### LINKED ISSUE #5332: Organization: Delete old playground orgs
I'd love to delete old playground orgs that are just hanging around doing nothing. These orgs likely exist because they applied, but were rejected or decided to not onboard.

I'm thinking we delete playground orgs older than 1 year.

@hackliv, is this okay to do from an operations perspective?

---

For engineering, maybe we consider a soft delete (although that can be tricky with associations 🤔)

sampoder:
Closes https://github.com/hackclub/hcb/issues/5332, is this how we want to approach the problem? This implementation uses soft deletion.

### DIFF
--- a/app/jobs/delete_stale_playground_events.rb
+++ b/app/jobs/delete_stale_playground_events.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class DeleteStalePlaygroundEvents < ApplicationJob
+  queue_as :low
+
+  def perform
+    Event.demo_mode.where("created_at < ?", 6.months.ago).delete_all
+  end
+
+end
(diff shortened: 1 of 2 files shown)


app/jobs/delete_stale_playground_events.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class DeleteStalePlaygroundEvents < ApplicationJob
+  queue_as :low
+
+  def perform
+    Event.demo_mode.where("created_at < ?", 6.months.ago).delete_all
```
garyhtou, line 7: Explore using `destory_all` if it exists and soft deletes

(delete doesn't run validations/callbacks, but destroy does)

garyhtou (changes requested):
Let's make sure we have cascading soft deletion of these associations:
- Event
	- OrganizerPositionInvite
	- OrganizerPositions
		- OrganizerPositionDeletionRequest
		- Tour
	- EventTag

check all associations for Event and the models above

jaspermayone:
Will take a stab at patching this up when I have time!

PR #5999, opened 2024-04-12
Refactor homepage event card into partial

cjdenio:


app/views/events/_event_card.html.erb
garyhtou, line 1: I think this should be named `card.html.erb`

PR #6003, opened 2024-04-12
Redesign org settings page

malted:
Each setting category is now an individual card...
image

...which expands out into a modal
image


garyhtou:
Oops, did not mean to close this

sampoder:
Chatted with Ben

PR #6006, opened 2024-04-12
Invoice: Show that sponsor paid the charged amount, not the invoiced amount

sampoder:
Closes https://github.com/hackclub/hcb/issues/2282


garyhtou (commented):
Does this have the possibility of showing an inaccurate amount? e.g. when the fee reimbursement and payout arrive at different times

sampoder:
Hmm. I’d have to test a bit but if that was the case the amounts displayed
in the ledger would also be inaccurate, right?

garyhtou:
tbh, i think we're likely fine in this scenario. but please test it!

PR #6012, opened 2024-04-13
[Appsignal] Configure Frontend Error Reporting

sampoder:
Pretty simple, based on https://docs.appsignal.com/guides/configuration.html#front-end-javascript.

Closes https://github.com/hackclub/hcb/issues/4892.


app/javascript/appsignal.js
```diff
@@ -0,0 +1,13 @@
+import Appsignal from '@appsignal/javascript'
+
+const environment = process.env.NODE_ENV || 'development'
+
+const appsignal =
+  environment != 'development'
+    ? new Appsignal({
+        key: '<YOUR FRONTEND API KEY>',
```
sampoder: @garyhtou I can't access the page to get the API keys for Appsignal, can you get me one and help me put it in here?

app/javascript/controllers/index.js
```diff
@@ -3,10 +3,16 @@
...
 import { Application } from '@hotwired/stimulus'
 import { definitionsFromContext } from '@hotwired/stimulus-webpack-helpers'
+import { installErrorHandler } from '@appsignal/stimulus'
+import appsignal from '../appsignal'
 import airbrake from '../airbrake'
 
 const application = Application.start()
 
+if (appsignal) {
+  installErrorHandler(appsignal, application)
```
cjdenio: This gets overwritten by Airbrake since a Stimulus app can only have 1 error handler. See [the source for @appsignal/stimulus](https://unpkg.com/browse/@appsignal/stimulus@1.0.18/dist/esm/index.js)

cjdenio (changes requested):

app/javascript/controllers/index.js
sampoder: Oh interesting, do you know if there's an alternative way of doing this?

sampoder: Or would we have to write a custom handler that does both.

northeastprince: Do we still need Airbrake, at least for this?

PR #6018, opened 2024-04-14
[Donations] Add a `tax_deductible` boolean to donations.

sampoder:
## Summary of the problem

See https://github.com/hackclub/hcb/issues/5649 for context. Closes https://github.com/hackclub/hcb/issues/5649. 

## Describe your changes

Adds this checkbox:

Screenshot 2024-04-14 at 4 15 44 PM

With this tooltip:

Screenshot 2024-04-14 at 4 15 49 PM

Which is used to conditionally render the line:

```
Per IRS guidelines, you agree that no goods or services will be provided in return for this gift.
```


cjdenio:
Thanks for this! Can we somehow have this be set by the org instead of the donor? This seems like a confusing option to have considering most donations will be tax-deductible

sampoder:
One option would be having a route like `/payments/start/hq`, I just thought that might be a hassle for organisers

sampoder:
I've switched us over to having a separate `/payments/start/hq` route, @cjdenio any thoughts on how we could communicate this to organisers?

sarthaktexas:
I think this implementation is a bit weird.

> Scenario:
> I'm donating to this organization.
> Why is it asking me this? What is considered a good or service? I'm confused. I'm going to email them to ask this question and **delay my donation**.

That would be my thought process. I think the solution here is two-step:
1. Put a checkbox on the donation settings to allow for goods and services.
2. Change copy:
(checkbox) Is this donation in exchange for a good or service? Instead of the popover, if this is checked it pops up an alert below saying if it is for a good or service, it is not tax deductible (exactly like the shared email warning).

thoughts, @garyhtou @cjdenio ?

garyhtou (dismissed):
- Keep the checkbox
- Update checkbox copy to ask whether they're receiving goods or services
- Remove the payments page

PR #6019, opened 2024-04-14
Update poppler to 4.2.2

YodaLightsabr:
## Summary of the problem
For a while I had an error when running HCB in development. I tracked it down to an issue with the gio2 gem.

See https://hackclub.slack.com/archives/C047Y01MHJQ/p1712439020831569

## Describe your changes
I updated poppler to fix this error.


garyhtou:
@albertchae would you mind reviewing this?
- Ensure there aren't breaking changes/regressions
- And is the right fix for the problem we're running into?

Gemfile
```diff
@@ -75,7 +75,7 @@ gem "rounding"
 gem "business_time"
 
 
-gem "poppler" # PDF parsing
```
albertchae: Hm I guess hcb doesn't have a standard for this looking at the rest of the `Gemfile`, but usually pinning to an exact version isn't recommended unless you have to prevent upgrades due to bugs in newer versions. Alternatively, I think pessimistically pinning to the minor version (e.g. `gem "poppler", "~> 4.2"` is reasonable too since according to semver, minor versions can add new functionality in a backwards compatible way but major versions are backwards incompatible, so having some manual intervention in the `Gemfile` is OK.

You should be able to achieve the same upgrades you need in the `Gemfile.lock` by running something like `bundle update poppler` without any changes here or the pessimistic pin. In fact, I usually prefer to run `bundle update poppler --conservative` or `bundler update poppler --strict` to minimize the gems I upgrade in a single PR. (N.B. You'd have to revert the `Gemfile.lock` changes you have in this commit locally before running these otherwise they will already be locked to all these new versions)

Some more reading about this
https://thoughtbot.com/blog/a-healthy-bundle
https://thoughtbot.com/blog/rubys-pessimistic-operator

https://www.reddit.com/r/rails/comments/sujdqk/best_practice_should_i_specify_versions_in_gem/

https://stackoverflow.com/questions/9265213/should-i-specify-exact-versions-in-my-gemfile (I disagree with the second answer here though and most of the comments in the thread do too)

albertchae:
I skimmed the code changes in https://github.com/ruby-gnome/ruby-gnome/compare/4.1.2...4.2.2. There's a lot because every ruby gnome gem lives in that codebase even if we only use some subset of them.

>     * Ensure there aren't breaking changes/regressions
> 
I feel fairly confident about this because ruby-gnome is a stable project and they follow semver, and this is a minor version upgrade. Quoting https://semver.org/, `MINOR version when you add functionality in a backward compatible manner`. Skimming the diff there didn't look like anything major in poppler and the other libraries we're using.

>     * And is the right fix for the problem we're running into?

Yes looks like it, I missed in your thread @YodaLightsabr that the exact issue you ran into was https://github.com/ruby-gnome/ruby-gnome/issues/1610 which seems to be a homebrew/macOS problem and this was fixed in https://github.com/ruby-gnome/ruby-gnome/commit/8e7c5c4f5d33edff68c383399ae2c2a36c5cf451 and released in 4.2.2 https://github.com/ruby-gnome/ruby-gnome/blob/master/NEWS.md#ruby-gnome-422-2024-04-02

garyhtou:
Hey @YodaLightsabr, I know we had previously chatted through Albert's feedback and you were planning on addressing them.

While upgrading our Ruby version, I was blocked by this problem. So, I've gone ahead and upgraded poppler in this commit [`d4bc0d7` (#6117)](https://github.com/hackclub/hcb/pull/6117/commits/d4bc0d74a15de19746f00e974d70f956f71ca8e6)

PR #6025, opened 2024-04-16
Use modal on playground mode banner

malted:
image

image

image

Once you've requested activation:
image


app/policies/event_policy.rb
garyhtou, line 1: Did you intend to delete this file? 👀

~~we no longer need events~~

malted: yes i HATE event_policy.rb

malted: worst policy

PR #6029, opened 2024-04-16
Add auto margins

malted:


garyhtou (changes requested):
how about we use tailwind!

cc @cjdenio

PR #6047, opened 2024-04-18
Create an Admin Mailer to notify admins of events in Slack

sampoder:
Closes https://github.com/hackclub/hcb/issues/5929 and closes https://github.com/hackclub/hcb/issues/5930.

Screenshot 2024-04-17 at 7 36 00 PM


sampoder:
I've tested the after_create by making a request myself in dev

app/models/organizer_position_deletion_request.rb
```diff
@@ -42,6 +42,8 @@ class OrganizerPositionDeletionRequest < ApplicationRecord
 
   scope :under_review, -> { where(closed_at: nil) }
 
+  after_create { AdminMailer.with(opdr: self).opdr_notification.deliver_later }
```
garyhtou: @sampoder This should be an `after_create_commit`. otherwise, you can run into race conditions where you can get an error from Sidekiq saying the ID doesn't exist.

PR #6052, opened 2024-04-18
[Admin] Receipt Reports

sampoder:
Adds a CSV exports to `/admin/users` which includes the total missing receipts. I need some sanity checks on the changes to `transactions_missing_receipts`. Closes https://github.com/hackclub/hcb/issues/5931/


app/models/user.rb
```diff
@@ -246,8 +254,12 @@ def receipt_bin
 
   def transactions_missing_receipt
     @transactions_missing_receipt ||= begin
+      return [] unless stripe_cards.any? || emburse_cards.any?
+
       user_cards = stripe_cards.includes(:event).where.not(event: { category: :salary }) + emburse_cards.includes(:emburse_transactions)
```
```suggestion
      user_cards = stripe_cards.includes(:event).where.not(event: { category: :salary }) + emburse_cards.includes(:emburse_transactions)
      return [] unless user_cards.any?
```

garyhtou (commented):
🪄

app/models/user.rb
```diff
@@ -255,10 +255,10 @@ def receipt_bin
   def transactions_missing_receipt
     @transactions_missing_receipt ||= begin
       user_cards = stripe_cards.includes(:event).where.not(event: { category: :salary }) + emburse_cards.includes(:emburse_transactions)
-      return [] unless user_cards.any?
+      return HcbCode.none unless user_cards.any?
```
garyhtou: ah good catch

sampoder: yep!

sampoder:
@RhysMP let me know when you want to use this!

PR #6056, opened 2024-04-19
Admin: Internal Ops Team Reminders on User Requests

sampoder:
Closes https://github.com/hackclub/hcb/issues/5788. See https://hackclub.slack.com/files/USNPNJXNX/F06V707B5AP/screenshot_2024-04-19_at_11.33.00___am.png for preview.


config/schedule.yml
```diff
@@ -122,3 +122,7 @@ calculate_subjects_metrics:
 ledger_audit_weekly_job:
   cron: "0 0 * * MON" # run first day of each week
   class: "GenerateLedgerAuditJob"
+
+send_admin_reminders_job:
+  cron: "30 * * * *" # run every hour
```
garyhtou: do we want to be sending the same reminders every hour?

sampoder: Hmm, I'm not sure. That's how I read Liv's message but maybe @hackliv can clarify for us

hackliv: Every hour might be a bit much - I'm thinking the hour before the cutoff period (24 hour turnaround time on processing requests on weekdays). Is this possible?

PR #6076, opened 2024-04-22
Remove HCB Wrapped (for now)

cjdenio:
It takes up a lot of space in our JS bundle :peefest:


garyhtou:
Good catch! i'd suggest removing it from package.json too

PR #6078, opened 2024-04-22
Reimbursements: hide "Another Organizer" option when there are no other organizers

cjdenio:
It currently looks like this when you're the only member of an org:
Screenshot 2024-04-22 at 11 37 26 AM


sampoder (approved):
looks good to me!

PR #6079, opened 2024-04-22
Improve reimbursement experience on mobile

cjdenio:
Right: before, left: after
Screenshot 2024-04-22 at 1 16 19 PM


sampoder:
thank you so much for all of these

sampoder (approved):
nice!

PR #6103, opened 2024-04-25
Redact ACCTVERIFY amounts in transaction exports

cjdenio:


app/services/export_service/csv.rb
```diff
@@ -37,7 +37,7 @@ def row(ct)
         [
           ct.date,
           ct.local_hcb_code.memo,
-          ct.amount_cents,
+          @public_only && ct.likely_account_verification_related? ? 0 : ct.amount_cents,
```
garyhtou: would a NULL be prefered?

garyhtou:
good catch!

app/services/export_service/csv.rb
cjdenio: I think 0 is better so it doesn't mess up scripts that expect a number. This is what we do in the transparency API

garyhtou: cool!

PR #6104, opened 2024-04-25
Fix transactions erroring when event has no point of contact

cjdenio:
Technically not possible, but life finds a way

```sql
bank-hackclub::DATABASE=> SELECT COUNT(*) FROM events WHERE point_of_contact_id IS NULL;
 count
-------
    31
```


garyhtou (approved):
life does find a way!

garyhtou:
thanks

PR #6107, opened 2024-04-25
Fix background on <select> inputs

cjdenio:


app/assets/stylesheets/components/_forms.scss
```diff
@@ -196,7 +196,10 @@ label {
 
 $select: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" fill="%238492a6" viewbox="0 0 32 32"><path d="M0 6 L32 6 L16 28 z"/></svg>');
```
garyhtou: woah this is some interesting black magic

PR #6115, opened 2024-04-26
[Reimbursements] Pass in `bank_name` when creating ACHs

sampoder:
Screenshot 2024-04-25 at 6 31 40 PM

context ^^


app/services/reimbursement/payout_holding_service/nightly.rb
```diff
@@ -40,6 +40,7 @@ def run
                 send_email_notification: false,
                 routing_number: payout_holding.report.user.payout_method.routing_number,
                 account_number: payout_holding.report.user.payout_method.account_number,
+                bank_name: Increase::RoutingNumbers.list(routing_number: payout_holding.report.user.payout_method.routing_number).first&.dig("name")&.titleize || "Unknown",
```
garyhtou: In most cases, `bank_name` is used as a display name. So I think "Bank Account" is more fitting than "Unknown"

```suggestion
                bank_name: Increase::RoutingNumbers.list(routing_number: payout_holding.report.user.payout_method.routing_number).first&.dig("name")&.titleize || "Bank Account",
```

garyhtou (dismissed):
long term fix would be to have users supply the bank name

we will need to stop using that increase API very soon

sampoder:
Agreed!

PR #6122, opened 2024-04-26
Use Column's API for routing number lookups

cjdenio:
Resolves #5838


sampoder:
We probably need to switch out its usage in https://github.com/hackclub/hcb/pull/6115 as well

cjdenio:
Good call, just updated!

app/controllers/ach_transfers_controller.rb
```diff
@@ -82,14 +82,14 @@ def validate_routing_number
...
 
-    banks = Increase::RoutingNumbers.list(routing_number: params[:value])
-
-    valid = banks.size > 0
+    bank = ColumnService.get "/institutions/#{params[:value]}" # This is safe since params[:value] is validated to only contain digits above
 
     render json: {
-      valid:,
-      hint: valid ? banks.first&.dig("name")&.titleize : "Bank not found for this routing number.",
+      valid: true,
```
garyhtou: We need stronger filtering here. We may want to use the ach_eligible and routing_number_type keys from Column's response

PR #6130, opened 2024-04-28
Org service levels admin filter

YodaLightsabr:
Work-in-progress 

closes #6109


app/services/event_service/nightly.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module EventService
+  class Nightly
```
garyhtou: Rename the Nightly to UpdateServiceLevel

app/services/event_service/nightly.rb
```diff
@@ -0,0 +1,18 @@
...
+    def run
+      save_service_level
+    end
+
+    private
+
+    def save_service_level
+      Event.all.map do |event|
+        event.update!(service_level: event.computed_service_level)
+      end
```
garyhtou:
```suggestion
      Event.find_each(batch_size: 100) do |event|
        event.update!(service_level: event.computed_service_level)
      end
```

db/migrate/20240427055503_add_service_level_to_events.rb
```diff
@@ -0,0 +1,5 @@
+class AddServiceLevelToEvents < ActiveRecord::Migration[7.0]
+  def change
+    add_column :events, :service_level, :integer, default: 1
```
garyhtou: add `null: false`

app/models/event.rb
garyhtou, line 1:
```ruby
before_create { service_level = computed_service_level }
```

PR #6132, opened 2024-04-28
[Receipt Service] Switch to `user.transactions_missing_receipt`

sampoder:
This should be a lot faster!


YodaLightsabr (approved):

YodaLightsabr:
Reviewed and tested on call w/ Sam

PR #6146, opened 2024-04-29
[Reimbursements] Ensure emails are lower case

sampoder:
Closes https://github.com/hackclub/hcb/issues/6144


sampoder:
tested locally! BoB became bob

garyhtou (commented):
Please use `ActiveRecord::Base::Normalize` instead

https://blog.saeloun.com/2023/11/02/rails-7-1-introduces-active-record-base-normalization

PR #6151, opened 2024-04-30
[Reimbursements] Preload files

sampoder:
Closes https://github.com/hackclub/hcb/issues/6148


app/views/reimbursement/reports/_expenses.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if report.expenses.any? %>
-  <% report.expenses.order(created_at: :asc).each do |expense| %>
+  <% report.expenses.order(created_at: :asc).includes(receipts: [:file_attachment]).each do |expense| %>
```
garyhtou: you also want it's blobs. see https://stackoverflow.com/questions/51785703/how-do-you-solve-n1-for-activestorage-urls

garyhtou:
https://github.com/hackclub/hcb/pull/6153

PR #6160, opened 2024-05-01
[Personal Txns] Handle attempted recreations

sampoder:
Resolves error 666, incredible. https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/666


sampoder:
Screenshot 2024-04-30 at 7 07 10 PM

PR #6162, opened 2024-05-01
[Reimbursements] Handle ACHs Rejected by Column

sampoder:
## Summary of the problem

When we run `ach_transfer.approve(User.find_by(email: "bank@hackclub.com"))` this is the first time that we communicate our ACH to Column. Occasionally, Column will reject ACHs. For example, ACHs with incorrect routing numbers are rejected. 

At the moment, this leads to an unapproved ACH.

## Describe your changes

I've made it so that if ACH approval succeeds, we proceed as usual.

However, if it fails, we:

- Move to the failed state
- Send an email to the user
- Delete the ACH from our records


app/services/reimbursement/payout_holding_service/nightly.rb
```diff
@@ -44,10 +44,20 @@ def run
                 creator: User.find_by(email: "bank@hackclub.com")
               )
               ach_transfer.save!
-              ach_transfer.approve!(User.find_by(email: "bank@hackclub.com"))
-              payout_holding.ach_transfer = ach_transfer
-              payout_holding.save!
-              payout_holding.mark_sent!
+              begin
+                ach_transfer.approve!(User.find_by(email: "bank@hackclub.com"))
+              rescue
```
garyhtou: should this be written as a transaction? (if possible)

sampoder: Can we write it in a transaction? I'm worried about us making a call to Column and rolling back the changes after that. The rollback wouldn't happen on Column

sampoder:
Tested this locally, it worked

sampoder:
I've just made a small edit, I'm going to have us reject the ACH instead of destroying it. I'm going to do that so that we have a better log of everything but don't interfere with ops' workflows.

app/services/reimbursement/payout_holding_service/nightly.rb
```diff
@@ -44,10 +44,20 @@ def run
...
+                ach_transfer.mark_rejected!(User.find_by(email: "bank@hackclub.com"))
+                payout_holding.mark_failed!
+                ReimbursementMailer.with(
+                  reimbursement_payout_holding: payout_holding,
+                  reason: "Your routing number / account number was invalid."
+                ).ach_failed.deliver_later
+              else
+                payout_holding.ach_transfer = ach_transfer
```
garyhtou: Does this line need to be ran, even if there's an error?

app/services/reimbursement/payout_holding_service/nightly.rb
```diff
@@ -44,10 +44,20 @@ def run
...
+                payout_holding.save!
+                payout_holding.mark_sent!
```
garyhtou: the stuff in the `else` can likely just be placed inside the `begin`. if there an error with `approve!` it will skip the rest of the lines and go directly to `rescue`

sampoder: I didn't put it there because if the lines after `approve!` fail, I don't want us to then reject the ACH as it's already been sent

app/services/reimbursement/payout_holding_service/nightly.rb
sampoder: Not necessarily, though we could... I personally don't think we should but I don't have strong thoughts on it

PR #6163, opened 2024-05-01
[Reimbursements] Add a tool for monitoring the clearinghouse

sampoder:
## Summary of the problem

Errors happen on the clearinghouse! I've been working on fixing them but having a tool to monitor this issues would be incredible. This is the tool I've designed for us to get a quick overview of why the clearinghouse's balance isn't $0 at any given moment. 

## Describe your changes

See screenshots at https://hackclub.slack.com/archives/C06JEH8HJSK/p1714546049447689


app/models/reimbursement/payout_holding.rb
```diff
@@ -58,6 +58,16 @@ class PayoutHolding < ApplicationRecord
...
     end
 
+    def payout_transfer
+      if ach_transfer.present?
+        ach_transfer
+      elsif increase_check.present?
+        increase_check
+      else
+        nil
+      end
```
garyhtou:
```suggestion
      ach_transfer || increase_check
```

sampoder:
With cases like this: https://hackclub.slack.com/archives/C047Y01MHJQ/p1714591570301449 often happening, I'm going to merge this so we have a tool. Would still love feedback and suggestions, just I'll add those in a later PR.

I've tested this locally and it only affects admin pages.

PR #6179, opened 2024-05-02
[Reimbursements] Clarify that reimbursements are US only

sampoder:
Closes https://github.com/hackclub/hcb/issues/6178


garyhtou:
Let's send some copy when creating a report (for both yourself and someone else). We should also add this to the public reimbursements page.

PR #6196, opened 2024-05-05
[Reimbursements] Render receipts count on org's ledger

sampoder:
It's handy! Plus it makes everyone look like they've been good and have upload their receipts.


app/models/hcb_code.rb
```diff
@@ -439,6 +439,12 @@ def receipt_optional?
     !receipt_required?
   end
 
+  def receipts
+    return reimbursement_expense_payout.expense.receipts if reimbursement_expense_payout.present?
```
garyhtou: what happens is someone attaches receipts to the HcbCode? will they be hidden?

also, does this affect preloading?

i wonder if there a way to write this with a join/union

PR #6201, opened 2024-05-06
Return as invalid if bank doesn't support ACHs & add validation for payouts

sampoder:
Closes https://github.com/hackclub/hcb/issues/6185, though I'm not sure how much of this we need because of the regex filter we have


garyhtou (commented):
🪄

Do we want this same logic for validation ach payout setting?

maybe a shared service can be good

PR #6205, opened 2024-05-07
Create new organizers with default `is_signee: false`

rluodev:
## Summary of the problem
We were creating new organizers and organizer position invites with `is_signee: null` because we didn't have a default value.

## Describe your changes
This will set the default value for newly created `organizer_position`s to `false` instead of `null`


sarthaktexas (dismissed):
this looks good to me but await gary review

garyhtou:
@sampoder would you be interested in reviewing and running the migration on your computer?

garyhtou:
🪄 after Sam's review

app/services/event_service/create_demo_event.rb
```diff
@@ -31,7 +31,7 @@ def run
 
         @event.save!
 
-        OrganizerPositionInviteService::Create.new(event: @event, sender: @point_of_contact, user_email: @email, initial: true).run!
+        OrganizerPositionInviteService::Create.new(event: @event, sender: @point_of_contact, user_email: @email, initial: true, is_signee: true).run!
```
garyhtou: We can't guarantee that the creator of a demo mode ends up signing the contract. The ops person who onboards and approves an org should also update contract signee status.

```suggestion
        OrganizerPositionInviteService::Create.new(event: @event, sender: @point_of_contact, user_email: @email, initial: true).run!
```

sampoder (approved):
Migration ran on my computer, thanks Ruien!

PR #6235, opened 2024-05-10
Reimbursements: Default report name should be the current date, not the date the app booted

cjdenio:
Model "helpers" like `attribute` and `has_many` are evaluated when the model is first loaded, _not_ every time the model is instantiated


sampoder (approved):
lol incredible thank you

PR #6248, opened 2024-05-11
Use Tailwind for donations QR code generator

cjdenio:


sampoder:
lol that is so much cleaner than what we had before, thank you

PR #6253, opened 2024-05-12
Experiment: move all the top-level config files into `config/`

cjdenio:
The root of the repository is getting pretty cluttered— this is one way of fixing that


garyhtou (commented):
How do IDEs react to this change? Many of them look for config files in the root of the repo/workspace

cjdenio:
I've updated all the file locations for VS Code (see `.vscode/settings.json`) and it works. I assume RubyMine and other editors let you specify custom config file locations as well

garyhtou:
I'm in favor of keeping it in the default location to lessen the configuration required and prevent tools from unexpectedly breaking. It's a cool experiment though!

PR #6254, opened 2024-05-12
[Donations] Support a `background` param

sampoder:
I feel like this has been requested a fair bit! Here's HQ with a navy blue:

Screenshot 2024-05-11 at 6 30 53 PM

I didn't use cookies because, unlike dark mode, there's no toggle for background colour and I didn't want to cause weird cases where people visit a different donation page and still have a specific background colour. Instead, the param persists throughout the process.

Closes https://github.com/hackclub/hcb/issues/4604


app/controllers/donations_controller.rb
```diff
@@ -202,6 +203,12 @@ def check_dark_param
     end
   end
 
+  def check_background_param
+    # because we're going to be injecting this value into a stylesheet,
+    # we ensure that it's a hex code to prevent: https://css-tricks.com/css-security-vulnerabilities/
+    @background = params[:background] unless (params[:background] =~ /^[0-9a-fA-F]{6}$/).nil?
```
sarthaktexas: oh this is smart

sarthaktexas (dismissed):
this looks good -- the comments are helpful. i think gary might have words on any sec vulns but i can't think of any more than you've already covered

sampoder:
nice! :D @garyhtou what do you think?

app/controllers/donations_controller.rb
garyhtou: Ruby regex is tricky because `^` and `$` isn't actually start and end of string. Instead, you need to use `\A` and `\z`.

```suggestion
    @background = params[:background] unless (params[:background] =~ /\A[0-9a-fA-F]{6}\z/).nil?
```

app/views/donations/_donation_form.html.erb
```diff
@@ -71,6 +71,13 @@
       Your contribution is tax-deductible<% if !@event.country_US? %> in the US<% end %>.
     </p>
 
+    <% if @background.present? %>
+      <%# @background is defined in the donations controller's check_background_param method %>
+      <%# it's used for organisations to set a custom background colour when embedding the form %>
+      <%# we include it here to ensure that this colour is used at all stages %>
+      <%= form.hidden_field :background, value: @background, name: :background %>
```
garyhtou: is the `name:` necessary?

app/views/donations/finish_donation.html.erb
```diff
@@ -19,7 +19,7 @@
       data-controller="stripe"
       data-stripe-api-key-value="<%= StripeService.publishable_key %>"
       data-stripe-client-secret-value="<%= @donation.stripe_client_secret %>"
-      data-stripe-return-url-value="<%= finished_donation_donations_url(@event, @donation.url_hash) %>"
+      data-stripe-return-url-value="<%= finished_donation_donations_url(@event, @donation.url_hash, { background: @background }.compact_blank) %>"
```
garyhtou: Is there a reason we're `compact_blank` here, and not elsewhere?

app/controllers/donations_controller.rb
garyhtou: Alternatively, you can use [`\h` for hexadecimal](https://ruby-doc.org/3.2.2/Regexp.html#:~:text=/%5Ch/%20%2D%20A%20hexdigit%20character%20(%5B0%2D9a%2DfA%2DF%5D)).

```suggestion
    @background = params[:background] unless /\A(\h{3}){1,2}\z/.match?(params[:background])
```

garyhtou: This also supports 3 digit hex

app/views/donations/_donation_form.html.erb
sampoder: yeah, otherwise it'll be put inside the donation params

app/views/donations/finish_donation.html.erb
sampoder: So if `@background` is nil, it'll have no impact. That felt safer.

app/controllers/donations_controller.rb
sampoder: oh I like this!

app/controllers/donations_controller.rb
```diff
@@ -202,6 +203,12 @@ def check_dark_param
     end
   end
 
+  def check_background_param
+    # because we're going to be injecting this value into a stylesheet,
+    # we ensure that it's a hex code to prevent: https://css-tricks.com/css-security-vulnerabilities/
+    @background = params[:background] unless (params[:background] =~ /\A[0-9a-fA-F]{6}\z/).nil?
```
sampoder:
```suggestion
    @background = params[:background] unless /\A(\h{3}){1,2}\z/.match?(params[:background])
```

garyhtou (commented):
🪄

PR #6255, opened 2024-05-12
Preload `:canonical_pending_declined_mapping` for CPTs

sampoder:
Closes https://github.com/hackclub/hcb/issues/5411

Seemed to be causing an n+1 with `declined?`


sampoder:
This is incredibly minor :peefest:

PR #6263, opened 2024-05-12
Consolidate `events/_form_invite` and `organizer_position_invites/_form`

sampoder:
Now we just have one! Closes https://github.com/hackclub/hcb/issues/5257


app/views/organizer_position_invites/new.html.erb
```diff
@@ -4,5 +4,4 @@
 <%= link_to @invite.event.name, @invite.event, class: "breadcrumb" %>
 <h1>Invite Another Member</h1>
 
-<!-- Does this page only get rendered when validation fails? -->
```
sampoder: fun fact, it also gets rendered when you open the invite modal in a new tab!

PR #6276, opened 2024-05-13
[Reimbursements] Provide defaults for `new` local

sampoder:
https://hackclub.slack.com/archives/C04LDFJ5X7S/p1715614770304869

Looks like a side effect of https://github.com/hackclub/hcb/pull/4660. 

https://www.shakacode.com/blog/rails-7-1-allows-templates-to-define-accepted-locals/


app/views/reimbursement/expenses/_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%# locals: (expense:, new:) -%>
+<%# locals: (expense:, new: false) -%>
 
 <% new ||= false %>
```
garyhtou: We can drop this line now that we have strict locals

PR #6279, opened 2024-05-13
Normalise email on `User` using `ActiveRecord::Base::Normalization`

sampoder:
Closes https://github.com/hackclub/hcb/issues/6147


garyhtou (commented):
🪄 how do we handle existing non-conforming emails?

PR #6280, opened 2024-05-13
[Account Numbers] Create searchable admin page of Column account numbers

rluodev:
## Summary of the problem

Ops had to ask engineers anytime they wanted to figure out which org owned an account number.

## Describe your changes

There is now a page for ops to find the org that owns an account number

Screenshot 2024-05-13 at 2 50 13 PM

Screenshot 2024-05-13 at 2 50 27 PM

Screenshot 2024-05-13 at 2 50 44 PM

Screenshot 2024-05-13 at 2 50 54 PM

Screenshot 2024-05-13 at 2 51 07 PM

Screenshot 2024-05-13 at 2 51 37 PM


rluodev:
Note: due to `has_encrypted` I had to use blind index for this (which meant adding three migrations to add account_number_bidx, add index for said column, and backfill said column)

app/controllers/admin_controller.rb
```diff
@@ -1168,6 +1168,36 @@ def hq_receipts
...
+    if @account_number_type == "1"
+      relation = relation.where(deposit_only: true)
+    elsif @account_number_type == "2"
+      relation = relation.where(deposit_only: false)
+    end
+
+    relation = relation.where(account_number: @q) if @q
+
+    @count = relation.count
+    @account_numbers = relation.page(@page).per(@per).order("id asc")
```
```suggestion
    @account_numbers = relation.page(@page).per(@per).order("events.id desc")
```

this way it matches the sorting on `/admin/events`

app/views/admin/account_numbers.html.erb
```diff
@@ -0,0 +1,46 @@
+<% title "Organization Account Numbers" %>
+
+<%= form_with local: true, url: account_numbers_admin_index_path, method: :get do |form| %>
+  <%= form.label "Account #" %>
+  <%= form.text_field :q, value: params[:q], placeholder: "Search" %>
+  <%= form.label "Event" %>
+  <%= form.collection_select(:event_id, Event.reorder(Event::CUSTOM_SORT), :id, :admin_dropdown_description, { include_blank: "Select An Event", selected: @event_id }, { width: 250, style: "max-width: 250px" }) %>
```
cjdenio:
```suggestion
  <%= form.collection_select(:event_id, Event.reorder(Event::CUSTOM_SORT), :id, :admin_dropdown_description, { include_blank: "Select An Event", selected: @event_id }, { style: "max-width: 250px" }) %>
```

has no effect

app/views/admin/account_numbers.html.erb
```diff
@@ -0,0 +1,46 @@
+<% title "Organization Account Numbers" %>
+
+<%= form_with local: true, url: account_numbers_admin_index_path, method: :get do |form| %>
```
cjdenio:
```suggestion
<%= form_with method: :get do |form| %>
```

app/views/admin/account_numbers.html.erb
```diff
@@ -0,0 +1,46 @@
...
+  <%= form.label "Event" %>
+  <%= form.collection_select(:event_id, Event.reorder(Event::CUSTOM_SORT), :id, :admin_dropdown_description, { include_blank: "Select An Event", selected: @event_id }, { width: 250, style: "max-width: 250px" }) %>
+  <%= form.label "Account number type" %>
+  <%= form.select :account_number_type, [["Show all", "0"], ["Deposit only", "1"], ["Spend + Deposit", "2"]], selected: @account_number_type %>
+  <%= form.submit "Search" %>
+<% end %>
+
+<div class="flex items-center mb2">
+  <div class="flex-grow">
+    <%= page_entries_info @account_numbers, entry_name: @account_numbers.count > 1 ? "Account Numbers" : "Account Number" %>
```
cjdenio:
```suggestion
    <%= page_entries_info @account_numbers, entry_name: "Account Number" %>
```
`page_entries_info` automatically pluralizes the `entry_name`!

app/controllers/admin_controller.rb
```diff
@@ -1168,6 +1168,36 @@ def hq_receipts
...
+    @account_number_type = params[:account_number_type].present? ? params[:account_number_type] : nil # default/nil = show all, 1 = deposit only, 2 = spend + deposit
+
+    relation = Column::AccountNumber.includes(:event)
+
+    if @event_id
+      relation = relation.where(event_id: @event_id)
+    end
+
+    puts "aaaaaa"
+    puts @account_number_type
```
garyhtou: 👀

rluodev:
aaaaaa

(Ok but actually I forgot I left that in there)

app/views/static_pages/_admin.html.erb
```diff
@@ -87,6 +87,7 @@
       <%= render partial: "static_pages/card", locals: { task_path: common_documents_path, human_name: "Common docs" } %>
       <%= render partial: "static_pages/card", locals: { task_path: flipper_path, human_name: "Flipper" } %>
       <%= card_to "Pending Ledger", pending_ledger_admin_index_path, { badge: CanonicalPendingTransaction.unsettled.count } %>
+      <%= card_to "Org Account Numbers", account_numbers_admin_index_path, { badge: Column::AccountNumber.count, subtle_badge: true } %>
```
sampoder:
```suggestion
      <%= card_to "Account Numbers", account_numbers_admin_index_path, { badge: Column::AccountNumber.count, subtle_badge: true } %>
```
Any reason we can't just call it Account Numbers? I feel like that's cleaner

rluodev: fine by me!

app/controllers/admin_controller.rb
```diff
@@ -1168,6 +1168,33 @@ def hq_receipts
...
+      relation = relation.where(event_id: @event_id)
+    end
+
+    if @account_number_type == "1"
+      relation = relation.where(deposit_only: true)
+    elsif @account_number_type == "2"
+      relation = relation.where(deposit_only: false)
+    end
+
+    relation = relation.where(account_number: @q) if @q
```
sampoder: Do we want to fuzzy search?

sampoder (dismissed):
Looks good to me! I just added a bit of styling to the top:

Screenshot 2024-05-14 at 12 12 14 PM

app/controllers/admin_controller.rb
rluodev: would we want to use https://github.com/seamusabshere/fuzzy_match gem?

sampoder: It's not exactly fuzzy searching but we can use the `ILIKE` operator like this:

```
scope :search, ->(q) { joins(:user).where("users.full_name ILIKE :query OR reimbursement_reports.name ILIKE :query", query: "%#{User.sanitize_sql_like(q)}%") }
```

That way if someone only puts in part of an account number it will still show up

rluodev:
The issue with ilike is that with the has_encrypted attribute we’d have to
decrypt and load all account numbers into memory before searching them

app/controllers/admin_controller.rb
sampoder: Yeah you're right, that's a bummer!

sampoder (approved):

hackliv (approved):

garyhtou (commented):
🪄 good to merge after testing, and after another person has ran the migration on their computer!

sampoder (approved):
the migration worked on my computer

rluodev:
i think it's waiting on someone from hcb-engineers group to approve

garyhtou (approved):

PR #6290, opened 2024-05-14
Reduce visual loudness of typography

lachlanjc:
## Summary of the problem

We've long been using SF Rounded in Safari, but it never became available in other browsers. I don't want core typography to look so different across browsers, so dropping back to the system font stack. Additionally, there's a lot of unneeded shouting across typography:

- Our 700/800-weight bold looked fun with SF Rounded, but unnecessarily loud with standard SF. Dropping weight down to 600 to streamline
- Dropping down the size of our massive modal dialog titles
- No need for table headings to be in all caps

## Describe your changes

Smaller modal titles:
CleanShot 2024-05-14 at 16 23 48@2x

Table headings, before:
CleanShot 2024-05-14 at 16 28 54@2x

Table headings, after:
CleanShot 2024-05-14 at 16 29 06@2x

On the homepage, metadata doesn't need to be bold:
image

In the sidebar, the item font size is slightly smaller & lighterweight:
image


app/views/events/donation_overview.html.erb
```diff
@@ -204,10 +204,10 @@
...
     <thead>
       <tr>
-        <th>STATUS</th>
-        <th>DATE</th>
-        <th>FROM</th>
-        <th style="text-align: right;">AMOUNT</th>
+        <th>Status</th>
+        <th>Date</th>
+        <th>From</th>
+        <th style="text-align: right;">Amount</th>
```
sarthaktexas:
```suggestion
        <th class="text-right">Amount</th>
```

app/views/events/partner_donation_overview.html.erb
```diff
@@ -50,9 +50,9 @@
   <table>
     <thead>
       <tr>
-        <th>STATUS</th>
-        <th>DATE</th>
-        <th style="text-align: right;">AMOUNT</th>
+        <th>Status</th>
+        <th>Date</th>
+        <th style="text-align: right;">Amount</th>
```
sarthaktexas:
```suggestion
        <th class="text-right">Amount</th>
```

app/views/events/transfers.html.erb
```diff
@@ -56,10 +55,10 @@
...
       <tr>
         <th></th>
-        <th>STATUS</th>
-        <th>DATE</th>
-        <th>TO</th>
-        <th style="text-align: right;">AMOUNT</th>
+        <th>Status</th>
+        <th>Date</th>
+        <th>To</th>
+        <th style="text-align: right;">Amount</th>
```
sarthaktexas:
```suggestion
        <th class="text-right">Amount</th>
```

app/views/invoices/index.html.erb
```diff
@@ -80,10 +80,10 @@
...
   <thead>
     <tr>
-      <th>STATUS</th>
-      <th>DATE</th>
-      <th>TO</th>
-      <th style="text-align: right;">AMOUNT</th>
+      <th>Status</th>
+      <th>Date</th>
+      <th>To</th>
+      <th style="text-align: right;">Amount</th>
```
sarthaktexas:
```suggestion
      <th class="text-right">Amount</th>
```

sampoder:
Most of these changes look pretty good (especially the table headings!). Personally, though, I preferred the bolder event / organization name in the left navbar.

app/assets/stylesheets/components/_transactions.scss
```diff
@@ -156,5 +156,5 @@ html[data-dark='true'] {
 }
 
 .transaction__icon {
-  padding-left: 1rem; 
+  padding-left: 1rem;
```
sarthaktexas:
```suggestion
  @apply pl-4;
```

sarthaktexas: We're trying to move over to Tailwind, so most of my review is converting things that can be included with this PR.

app/assets/stylesheets/components/_transactions.scss
```diff
@@ -26,7 +26,7 @@
     #33d6a6 100px
   );
   background-attachment: fixed;
-  font-weight: bold;
+  font-weight: 600;
```
sarthaktexas: for example, not converting this because it involves converting more than just this.

app/assets/stylesheets/components/_dock.scss
```diff
@@ -109,9 +109,9 @@
 }
 
 .pwa__prompt__hide {
-  position: absolute; 
-  right: 5px; 
-  top: 5px; 
+  position: absolute;
+  right: 5px;
+  top: 5px;
   cursor: pointer;
```
sarthaktexas:
```suggestion
  @apply absolute right-[5px] top-[5px] cursor-pointer;
```

lachlanjc:
@sampoder That's valid! I'm about to redesign the whole sidebar, so the org name styling is temporary, but I think once you live with this change for a little while it'll grow on you.

sarthaktexas (changes requested):
yes!! I've been wanting to make these changes for a while but did not want to mess with your design. thanks! **can we have a longer discussion on the organization name?**

sarthaktexas:
Oh! Nevermind, looks like Sam commented the same thing!

app/views/reimbursement/reports/_create_form.html.erb
```diff
@@ -55,7 +55,7 @@
       <%= form.select :event_id, current_user.events.not_demo_mode.pluck(:name, :id) %>
     <% end %>
     <div class="tooltipped tooltipped--e inline-block" aria-label="<%= @event&.demo_mode? ? "You don't have access to this feature at the moment." : "This will start your reimbursement report." %>">
-      <%= form.submit "Start report", class: "mt2", disabled: @event && (!organizer_signed_in? || @event&.demo_mode?) %>
+      <%= form.submit "Start report", class: "bg-info mt2", disabled: @event && (!organizer_signed_in? || @event&.demo_mode?) %>
```
cjdenio: Does this have any effect? `bg-info` should be the default

app/assets/stylesheets/basscss.css
```diff
@@ -31,9 +31,7 @@
 .text-decoration-none {
   text-decoration: none;
 }
-.bold {
-  font-weight: 700;
-}
+.bold,
```
cjdenio: Ideally I'd rather not change this, since this file is meant to be static utility classes

app/assets/stylesheets/components/_dock.scss
cjdenio: could probably be `right-1` (4px) or `right-1.5` (6px) to avoid the arbitrary value

lachlanjc: I didn't edit this code, just running Prettier here. None of this is relevant to this PR so resolving this.

app/views/reimbursement/reports/_create_form.html.erb
lachlanjc: Not right now! I'm working on a buttons redesign on another branch & it was useful there, dropping it here.
```suggestion
      <%= form.submit "Start report", class: "mt2", disabled: @event && (!organizer_signed_in? || @event&.demo_mode?) %>
```

app/assets/stylesheets/basscss.css
lachlanjc: I'm dropping all usages of 700 weight across the UI, this doesn't seem worth re-architecting before switching to the Tailwind class. I've resolved this by updating the font weight Tailwind uses for bold classes as well. Thanks!

app/assets/stylesheets/components/_transactions.scss
lachlanjc: This isn't relevant to this PR since we're not updating non-typography code regardless, but `@apply` is an anti-pattern for Tailwind, in my experience. It's best to write a strong configuration theme with all the CSS variables stored in one place, let Tailwind use those tokens only (Tailwind in an app should not have its default color palette, for instance), then consume the tokens through 1) Tailwind classes in HTML/JSX, and 2) the same token variables in the component-based CSS.

Tailwind's strength is in keeping styling colocated with view code, such that classes like this aren't left behind in CSS files while unused, and maintaining a strong design system of tokens. Using `@apply` inside component-based CSS, especially with arbitrary values, has all the problems of component-based CSS having classes not tied to their real usage, doesn't enforce the design system. It uses Tailwind more as a DSL/shorthand for CSS, which adds unnecessary complexity over plain CSS.

app/assets/stylesheets/basscss.css
sarthaktexas: oh shoot i didn't notice this at all @cjdenio thanks for noticing

app/views/events/_filter.html.erb
```diff
@@ -11,7 +11,7 @@
     <div data-controller="menu" data-menu-append-to-value="turbo-frame#ledger" data-menu-placement-value="bottom-end">
       <button
         type="button"
-        class="btn btn-small bg-info menu__toggle menu__toggle--arrowless overflow-visible z1 mr1"
+        class="btn btn-small bg-info menu__toggle menu__toggle--arrowless overflow-visible z1"
```
sarthaktexas:
```suggestion
        class="btn btn-small bg-info menu__toggle menu__toggle--arrowless overflow-visible z-10"
```

app/views/events/_filter.html.erb
```diff
@@ -30,9 +30,9 @@
   <div data-controller="menu" data-menu-append-to-value="turbo-frame#ledger" data-menu-placement-value="bottom-end">
     <button
       type="button"
-      class="btn btn-small bg-info menu__toggle menu__toggle--arrowless overflow-visible z1 mr1"
+      class="btn btn-small bg-info menu__toggle menu__toggle--arrowless overflow-visible z1"
```
sarthaktexas:
```suggestion
      class="btn btn-small bg-info menu__toggle menu__toggle--arrowless overflow-visible z-10"
```

sarthaktexas (dismissed):

sarthaktexas (commented):
changing z values to use increments of 10 not 1.

tailwind.config.js
```diff
@@ -42,6 +42,9 @@ module.exports = {
       ],
       'check-handwriting': ['Damion', 'cursive'],
     },
+    fontWeight: {
```
cjdenio: This should be nested under `extend`, otherwise it overwrites Tailwind's existing font-weight utilities (`font-normal`, etc)

lachlanjc:
@sarthaktexas I genuinely appreciate your desire to clean up CSS across the board here—there's a ton of work left to do in that area. That said, this PR is about typographic hierarchy, not switching other parts of the app to Tailwind or changing our approach to z-indexes. If you'd like my eyes on a PR with the changes you'd like to make, or if you want to discuss CSS architecture, I'm super happy to contribute in those ways. But it's crucial that on a PR, the creator is responsible for the changes shipped, and that there's a clear scope so we can get it out the door.

tailwind.config.js
lachlanjc: Oh great catch! I've allowlisted only the font weights we want to use.

garyhtou (dismissed):
Overall, I like what's happening here! I do think that the organization name should be bolder, however, I'll leave the final decision up to you.

garyhtou (approved):

garyhtou (approved):

PR #6292, opened 2024-05-14
Rename Invoices to Payment Requests

lachlanjc:
## Summary of the problem

Since the very first launch of Bank, customers (including me!) have been confused about "Invoices." Most customers have never used the concept before, and the Invoices tab is not where you went to pay an invoice, which is when someone else would use that word to you. Right now, smaller projects are using prefilled donation links to collect sponsor payments, which don't have the weight of emails and clear tracking of payments (on both ends) our invoice feature already provides.

I workshopped with @dwahbe to come up with a new name: Payment Requests. Stripe calls them "payment links," which don't have quite the weight; Mercury also uses [payment requests](https://mercury.com/blog/inside-mercury/august-2022-product-updates). They describe exactly what you're doing—requesting a payment—instead of a resulting object. As we move out of the old design flows of lists with create actions (the Rails default, so an easy place to start) toward flows based on user intention (I want to send money/request money, let me do that quickly), I think it's a big step forward.

As part of rolling this change out, Dylan is writing a changelog entry (by EOD tomorrow, 05-15) so customers aren't confused where their invoices have gone.

## Describe your changes

I've updated the name everywhere in the copy, but not the underlying model, since I don't think there's any reason to. The URL is now `/requests` instead of `/invoices`, with a redirect from `/invoices`.

CleanShot 2024-05-14 at 17 59 23@2x
CleanShot 2024-05-14 at 17 59 09@2x
CleanShot 2024-05-14 at 17 59 47@2x


app/views/invoices/_form.html.erb
```diff
@@ -14,7 +14,7 @@
   <% end %>
 
   <%= form.fields_for :sponsor_attributes do |form| %>
-    <h2>Sponsor details</h2>
+    <h2 class="h2">Sponsor details</h2>
 
     <p data-behavior="sponsor_update_warning" style="display: none;" class="flex items-center info line-height-3">
       <%= inline_icon "important", size: 32, class: "mr1" %>
```
sampoder: Whilst we're at it, it might be worthwhile using a more general term than "Sponsor". Sender / supporter might work, though I'm not a massive fan of either. It's certainly confusing to be sending invoices / requests to a "Sponsor" if they're actually someone you've done contract work for, or someone who owes you money for something etc.

sampoder (commented):
Any reason we're not just going with "Requests"? 

Payment requests feels like a bit of a mouthful. Honestly, I think I prefer Invoices over Payment Requests. At the end of the day, creating an invoice / request, sends what Stripe refers to as an invoice in their email to the sponsor:

Screenshot 2024-05-14 at 6 19 46 PM

Is this people will expect if we change the wording on our end? We could probably look into changing the word invoice there too but you also get a PDF which looks like an invoice.

I'm neutral between Requests & Invoices. I think removing the Sponsor wording is probably more pressing.

lachlanjc:
Thanks for the thoughts! Invoice makes sense from the outside, it's the business standard terminology and companies should forward these to their accounts payable just like any other invoice. We definitely should not remove it from the PDF etc.

I don't think "Requests" is clear enough what they are; you could be requesting permissions etc.

Good feedback on Sponsors, I don't love that either. Supporter could be good, I'll give it a go, or we could use a more neutral "Recipient."

sarthaktexas:
i LOVE the term supporters

lachlanjc:
I talked to @dwahbe & I think "recipient" is clearest; one common use case Zach mentioned to me for Payment Requests is requesting payment from yourself if you accidentally use the wrong card for something, and "supporter" doesn't make sense there. We already use "recipient" on mailed checks, and no reason not to be consistent.

image

garyhtou:
@dwahbe, if you already have a change log post written, can you please post it here? I'd like full context before deciding whether this is a change we'd like to move forward with.

dwahbe:
@garyhtou

~~Invoices~~ - Now Payment requests
 
We have renamed Invoices to Payment requests. Since HCB was created, the name Invoices was confusing to some people. We figured the simpler, the better: Payment requests. All functionality remains the same.
 
Thanks to everyone for the great feedback!

garyhtou (changes requested):
As mentioned on our call earlier, we're taking a brief pause on this. I'll provide an update when available.

garyhtou:
I had a chat with Lachlan to share that we're focusing on financial literacy (e.g. teaching high schoolers about concepts such as "invoices"); and as a result, we'd like to keep the name

PR #6295, opened 2024-05-15
[Docker Compose] Add new file for DB and Redis-only setup

YodaLightsabr:
## Summary of the problem

image

This file has been sitting in my sidebar for months, and it would be nice to check it in to git!

## Describe your changes
Add the `docker-compose.dbredisonly.yml` file for running HCB outside of Docker but running postgres and redis inside of Docker. I'm open to naming suggestions.


sampoder (dismissed):
Docker-less, sigh.

garyhtou (dismissed):

garyhtou:
if you want to use this, i think you'll need to pass in a `-f` flag
`docker compose up -f docker-compose.dbonly.yml`

docker-compose.dbonly.yml
```diff
@@ -0,0 +1,20 @@
+version: '3'
```
garyhtou:
```suggestion
# To use this DB-only ("dockerless") setup, pass a `-f docker-compose.dbonly.yml` to docker compose.
# e.g. `docker compose up -f docker-compose.dbonly.yml`
version: '3'
```

garyhtou (approved):

garyhtou:
🐳

PR #6296, opened 2024-05-15
Event: Create a model for pricing plan

sampoder:
Resolves https://github.com/hackclub/hcb/issues/5092, or at least it will resolve https://github.com/hackclub/hcb/issues/5092 when ready.


sampoder:
Currently, this only has the models. I think that demo mode should still be handled separately, at least for now.

app/controllers/events_controller.rb
```diff
@@ -266,6 +266,12 @@ def update
     end
     fixed_user_event_params.delete(:hidden)
 
+    if fixed_event_params[:plan_type] && fixed_event_params[:plan_type] != @event.plan&.plan_type
+      @event.plan = Event::Plan.create(plan_type: fixed_event_params[:plan_type])
```
garyhtou: What happens to the existing plan?

garyhtou: It would be nice for old plans to still be associated to the Event for papertrail reasons.

This means the FK should be on Plan rather than Event. We may also want to look into doing a soft deletion here

app/controllers/events_controller.rb
```diff
@@ -266,6 +266,12 @@ def update
...
+      @event.save!
```
```suggestion
      @event.update!(plan: Event::Plan.create(plan_type: fixed_event_params[:plan_type]))
```

app/models/event.rb
```diff
@@ -54,13 +54,15 @@
 #  updated_at                                   :datetime         not null
 #  club_airtable_id                             :text
 #  emburse_department_id                        :string
+#  event_plan_id                                :bigint
```
garyhtou: move this FK to Plan

app/models/event.rb
```diff
@@ -574,7 +577,9 @@ def fee_balance_v2_cents
   alias fee_balance fee_balance_v2_cents
 
   def plan_name
-    if demo_mode?
+    if plan.present?
+      plan.label
+    elsif demo_mode?
```
garyhtou: Since we don't plan on supporting demo mode in this PR, we should put the `plan.label` after demo mode

app/models/event.rb
```diff
@@ -674,6 +679,14 @@ def dormant?
     !engaged?
   end
 
+  def sponsorship_fee
+    plan&.revenue_fee || super
+  end
+
+  def revenue_fee
+    plan&.revenue_fee || sponsorship_fee
+  end
```
```suggestion
  def sponsorship_fee
    # report to airbrake to deprecate
    revenue_fee
  end

  def revenue_fee
    plan&.revenue_fee || self[:sponsorship_fee] # to directly access the column and bypass the method
  end
```

app/models/event/plan.rb
```diff
@@ -0,0 +1,40 @@
...
+      if plan.respond_to?(method)
+        plan.public_send(method, *args, &block)
+      else
+        super
+      end
+    end
+
+    def respond_to_missing?(method, include_private = false)
+      plan.respond_to?(method) || super
+    end
```
garyhtou: something feels wrong here

I think using `.becomes` might be better here

app/models/event/plan/standard.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class Event
+  class Plan
+    class Standard
```
garyhtou: Maybe this should inherit from `Event::Plan` so that STI works

app/models/event/plan/standard.rb
```diff
@@ -0,0 +1,50 @@
...
+      def revenue_fee
+        0.07
+      end
+
+      def revenue_fee_label
+        ActionController::Base.helpers.number_to_percentage(revenue_fee * 100, precision: 0)
+      end
```
garyhtou: Move this into `Event::Plan`

app/models/event/plan/standard.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      def label
+        "Full Fiscal Sponsorship (#{revenue_fee_label})"
+      end
+
+      def self.available_features
+        # this must contain every HCB feature that we want enable / disable with plans.
+        %w[cards invoices donations account_number check_deposits transfers promotions google_workspace documentation settings reimbursements]
+      end
+
```
garyhtou: This should go into `Plan`

app/models/event/plan/standard.rb
```diff
@@ -0,0 +1,50 @@
...
+        Event::Plan::Standard.available_features.each do |feature|
+          self.class.define_method("#{feature}_enabled?") do
+            feature.in?(features)
+          end
+        end
+        Event::Plan::Standard.subclasses.each do |subclass|
+          self.class.define_method("#{subclass.name.underscore.split('/').last}?") do
+            self.is_a?(subclass)
+          end
+        end
```
garyhtou: Does rails already automatically define methods for checking for the subclass?

garyhtou: If we don't have a wrong reason for defining this, I would prefer that we don't. This is because in practice, we should be checking for features, not the plan type

app/models/event/plan/standard.rb
```diff
@@ -0,0 +1,50 @@
...
+      def features
+        Event::Plan::Standard.available_features
+      end
+
+      def initialize
+        Event::Plan::Standard.available_features.each do |feature|
+          self.class.define_method("#{feature}_enabled?") do
+            feature.in?(features)
+          end
+        end
```
garyhtou: I believe this can be defined on the class level when the class gets loaded.


```ruby
class Event
  class Plan
    self.available_features = %w[a b c]

    self.available_features.each do |feature|
      define_method(...)
    end
 end
end
```

app/models/event/plan/standard.rb
```diff
@@ -0,0 +1,50 @@
...
+          self.class.define_method("#{subclass.name.underscore.split('/').last}?") do
+            self.is_a?(subclass)
+          end
+        end
+        self.class.define_method("standard?") do
+          Event::Plan::Standard.subclasses.each do |subclass|
+            return false if self.is_a?(subclass)
+          end
+          true
+        end
```
garyhtou: I'm in support to having this `standard?` method. Define it directly on the class though.


```ruby
class Event
  class Plan
    def standard?
      plan_type == "Standard"
    end
 end
end
```

app/policies/event_policy.rb
```diff
@@ -33,35 +33,35 @@ def by_airtable_id?
   end
 
   def edit?
-    admin_or_user?
+    admin_or_user? && plan.settings_enabled?
```
garyhtou: a lot of this logic needs to be changed to `admin? || (user? && plan.settings_enabled?)`

app/policies/event_policy.rb
```diff
@@ -211,4 +211,8 @@ def is_public
     record.is_public?
   end
 
+  def plan
+    !user&.admin? && record.plan ? record.plan : Event::Plan::Standard.new
```
garyhtou: Let's remove this fallback.

The logic should just be `record.plan`. Then the fallback would like inside the `Event#plan` method

app/views/events/_form.html.erb
```diff
@@ -318,6 +318,16 @@
       <%= form.collection_select :point_of_contact_id, User.admin.all, :id, :email %>
     </div>
 
+    <div class="field">
+      <%= form.label :plan_type, capture { %>
+        Current Plan
+      <% } %>
```
garyhtou: i think we can just give it a string

app/views/events/_form.html.erb
```diff
@@ -318,6 +318,16 @@
...
+      <%= form.select :plan_type, [[Event::Plan::Standard.new.label, Event::Plan::Standard.name]] + Event::Plan::Standard.subclasses.map { |p| [p.new.label, p.name] }, selected: @event.plan.plan_type %>
```
garyhtou: define a method called `Event::Plan.available_plans`. It should return an array of classes (including Standard).

Then you can map that array here. Or, use Rail's collection form helper

garyhtou: for admin sake, it may be nice to have a more details on what each plan means. either include it in the dropdown (that may not be enough space). Or, below the dropdown, just give a dump of plan name, fee %, and enabled features

app/models/event/plan/standard.rb
```diff
@@ -0,0 +1,50 @@
...
+        0.07
+      end
+
+      def revenue_fee_label
+        ActionController::Base.helpers.number_to_percentage(revenue_fee * 100, precision: 0)
+      end
+
+      def label
+        "Full Fiscal Sponsorship (#{revenue_fee_label})"
+      end
```
garyhtou:
```suggestion
      end
      
      def description
        "The standard fiscal sponsorship plan"
      end
```

garyhtou (commented):
When you start utilizing plans to check features, make sure you preload is necessary

app/views/events/settings/_admin.html.erb
```diff
@@ -28,10 +28,10 @@
         <%= form.label :name %>
         <%= form.text_field :name %>
       </div>
-
+      
       <div class="field">
-        <%= form.label :sponsorship_fee, "Sponsorship fee (as fraction of revenue)" %>
-        <%= form.text_field :sponsorship_fee, value: event.revenue_fee || 0.07, placeholder: "0.07", required: true %>
+        <%= form.label :plan_type, "Current plan" %>
+        <%= form.select :plan_type, [[Event::Plan::Standard.new.label, Event::Plan::Standard.name]] + Event::Plan::Standard.subclasses.map { |p| [p.new.label, p.name] }, selected: @event.plan&.plan_type, include_blank: "Select a plan..." %>
```
sampoder: Clean up

app/models/event/plan.rb
```diff
@@ -0,0 +1,66 @@
...
+# Foreign Keys
+#
+#  fk_rails_...  (event_id => events.id)
+#
+class Event
+  class Plan < ApplicationRecord
+    belongs_to :event
+
+    include AASM
+
```
garyhtou: Validate one active plan per Event

app/models/event/plan/two_point_nine_percent.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class Event
+  class Plan
+    class TwoPointNinePercent < Standard
```
garyhtou: lol 🤣

db/migrate/20240515151522_create_event_plan.rb
```diff
@@ -0,0 +1,10 @@
+class CreateEventPlan < ActiveRecord::Migration[7.1]
+  def change
+    create_table :event_plans do |t|
+      t.string :plan_type
+      t.string :aasm_state
```
garyhtou: null: false

app/models/event/plan.rb
```diff
@@ -0,0 +1,66 @@
...
+        feature.in?(features)
+      end
+    end
+
+    def self.available_plans
+      plans = Event::Plan.subclasses
+      Event::Plan.subclasses.each do |subclass|
+        plans += subclass.subclasses
+      end
+      plans
```
garyhtou:
```ruby
Event::Plan.descendants
```

PR #6298, opened 2024-05-15
Reimbursements: Switch mileage icon to car

lachlanjc:
## Summary of the problem

The current mileage button uses the receipts icon, which is confusing.

## Describe your changes

I designed a car icon recently, so dropping it in here!

image


sampoder (approved):
THANK YOU! This has been long talked about.

PR #6304, opened 2024-05-16
Simplify check deposit page design

lachlanjc:
## Summary of the problem

We've got two orange elements competing for attention right now on the check deposit page:
CleanShot 2024-05-15 at 21 03 16@2x

## Describe your changes

Not a dramatic change, just making one element clearly more important than the other & reducing the loudness of the info panel:
CleanShot 2024-05-15 at 21 03 08@2x


sampoder (approved):
Looks good to me!

PR #6308, opened 2024-05-16
Streamline organization ledger design

lachlanjc:
## Summary of the problem

There's 4 key pieces to a transaction row: the type, the date, the memo, and the amount. I've got an update for each one today.

1. The type icons have needed some love.
    - I've updated the icon for recurring donations, so it's more closely connected to the heart visual we use for donations, and there's one icon in use in those rows instead of two.
    - We were using the `bank-account` icon for direct ACH/debits, which is visually complex for a common occurrence, and confuses our logo with transactions. A $ suffices.
    - We were using the `minus-fill` icon for our fees, which draws heavy visual attention (as the only filled icons in the app) to the transactions our customers like the least . I've replaced that with an outline of our logo for clarity. 
3. For the date, ISO is great for accounting, and we should keep using it on spreadsheets, but it's not the fastest to read. Using our existing date formatting is way quicker to understand.
4. Memos are complicated. All-caps labels are hard to read & feel fragile—anyone who's used Apple Card knows what a night-and-day difference seeing simple vendor names is. Soon we'll kick off a project to intelligently capitalize all transaction memos, but I've started by sentence-casing our own.
    - Additionally, the italicized memos for pending transactions aren't great. I've replaced with a Pending badge that's cleaner.
    - I've dropped the comments badge, which were a temp fix until the email notifications we've since built!
5. Amounts. The red/green diff colors for transaction rows were my very first design idea for HCB, to connect finances to the GitHub UI many of our customers use. They've done their time; seeing a wall of red on your home screen feels like you're doing something wrong when spending money is exactly what you should be doing in your org. Inspired by Mercury, Apple Card, et al, I've switched us to showing the debit as an absolute sum ("$512" is a $512 charge), and additions with a plus sign ("+$512 donation").

## Describe your changes

Before:
CleanShot 2024-05-16 at 03 07 45@2x

After:
CleanShot 2024-05-16 at 03 03 14@2x

CleanShot 2024-05-16 at 02 48 29@2x

Here's the recurring donations table, before:
CleanShot 2024-05-16 at 02 37 07@2x

After:
image


sampoder (commented):
Love the new icon for recurring donations! Thoughts on making the `bank-icon` icon a little smaller, it feels a bit large compared to the others (this may be largely an optical illusion):
Screenshot 2024-05-18 at 3 28 47 PM

sampoder:
I'm leaning towards keeping the comments indicator. Even though we send emails, not everyone has them enabled and the indicator is helpful when you're looking back on transactions (at least it is for me personally).

I love the new pending badge, feels a lot more integrated.

I think that removing `-` is going to be a little confusing for users, at least at first. I haven't used Mercury but I think that one reason it works for the Apple Card is that cards are seen as a tool for spending money, when there's a positive amount, it's the exception. I don't think that's the same for us. 

I'd prefer to keep the red and green backgrounds. That's a personal preference and I think everyone's going to have different opinions it. I do agree that on dark mode, it is very red. On light mode is feels way nicer.

app/views/events/donation_overview.html.erb
```diff
@@ -153,8 +151,8 @@
           <th></th>
           <th>Started on</th>
           <th>Name</th>
-          <th>Amount</th>
-          <th>Total donated</th>
```
sampoder: Would it be reasonable to make this total? Or would that be too ambiguous?

app/views/events/donation_overview.html.erb
```diff
@@ -172,10 +170,10 @@
...
               <% end %>
             </td>
-            <td>
-              <%= render_money recurring_donation.amount %><span class="muted">/month</span>
+            <td class="text-right">
+              <%= render_money recurring_donation.amount %><span class="muted">/mo</span>
             </td>
-            <td>
+            <td class="text-right">
               <strong><%= render_money recurring_donation.total_donated %></strong>
```
sampoder: Thoughts on making the monthly amount bolded and the total non-bolded?

app/views/events/donation_overview.html.erb
```diff
@@ -140,11 +140,9 @@
...
 
 <% if @recurring_donations.size > 0 %>
-  <h2>
-    Recurring Donations
-    <span style="font-weight: 400; font-size: 0.7em">
-      <%= render_money @recurring_donations_monthly_sum %><span class="muted">/month</span>
-    </span>
+  <h2 class="heading pb-0 border-b-0">
+    Recurring donations
+    <%= badge_for "#{render_money @recurring_donations_monthly_sum}/month" %>
```
sampoder: I think instead of putting this in a badge we should add a word in between "Recurring Donations" and the amount. Maybe

Recurring donations of $275/month or 
Recurring donations worth $275/month

YodaLightsabr:
+1 to what Sam said. I really like the new icons, and I think they fit well. The pending badges look great as well; how do you think those would look at the front of the memo?

dwahbe:
I'm loving these changes and think there is also some great feedback here. I agree with @sampoder on keeping comment badges. I also think that we can add a "+" or "-" with each transaction for clarity.

Beyond those two items, I like the changes in the initial PR. I think the pending badge is really helpful, especially when you want to quickly glance at your latest transactions. I'm a fan of removing the color, it feels very loud and as @lachlanjc said, seeing an overwhelming amount of red makes you feel like you're doing something wrong.

app/views/events/donation_overview.html.erb
lachlanjc: I don't think this is taking up too much space & total is a little more ambiguous

app/views/events/donation_overview.html.erb
lachlanjc: We don't include numbers/totals in headings anywhere else in the app—is there a reason this is worth an exception for?

lachlanjc:
I've brought back the comments badge, moved the pending/declined badges to the front (great suggestion!), and made _all_ transactions show either a preceding + or - for better clarity.

image

I'm standing by dropping the backgrounds; I think this design is plenty clear without their ugliness.

app/views/events/donation_overview.html.erb
malted: +1 for not in a pill - but agreed it would look weird in a heading. Is there something else we can use?

malted:
Cool changes! Some thoughts;

I really think the coloured backgrounds of the rows give a clear indication of the recent state of money movements.
Whether a transaction is (in|out)going is the most critical piece of information and being able to determine this while reading any part of the row is necessary.
I think if this part of the PR specifically gets merged it would make Bank harder to use.
Can we explore alternatives? What if we made the backgrounds a bit more muted than they currently are?

Also, the new fiscal sponsorship fee icon looks too much like the HCB transfer icon - it doesn't made immediate sense to me what it means and they don't look like distinct transaction types - even though they are completely different.

jaspermayone:
I also think the colors are super useful, perhaps we just tone them down a bit?

dwahbe:
I really like this new design. On @malted's point that the colors make it easy to know if a transaction is adding or subtracting from your balance, I agree. But I think that is also communicated clearly with the new design, which has a green color!

When someone looks at their transactions, they don't need to instantly know the total number of incoming or outgoing transfers. They want to look at their latest transactions and make sure they are correct, money is being spent on items that make sense, and they want to see who is spending or giving them that money.

Because of this, removing the fill color makes it easier to use. In both light and dark mode its much easier to read and less intense. It also doesn't feel like you are doing something wrong if you just make a lot purchases (where previously you would see a wall of red).

I think the fill colors would make sense if we wanted every org to continue to raise money and not spend it, but that's not the case. One of the best things about orgs on HCB is that they are in the world doing real stuff. We shouldn't be giving them an interface that looks like they are doing something wrong when they spend money, particularly if they make a lot of small transactions. 

We want to communicate a clear recap of their latest transactions, still using color to display if a transaction added or subtracted from their balance. But HCB's interface shouldn't have an opinion about which transaction was good or not, and the current colors have a strong opinion.

sampoder:
I've been getting feedback in the office, people seem to largely prefer keeping the backgrounds. Can we split that off into a different PR have a discussion about that there?

jaspermayone:
I do also like how my bank does it, where the transaction rows aren't colored, but the amount is.

CleanShot 2024-05-24 at 11 58 37@2x

PR #6310, opened 2024-05-16
Return ID from `/api/v1/events/create_demo`

sampoder:
This will help with https://github.com/hackclub/hcb/issues/6305.


sampoder:
https://github.com/hackclub/hcb/issues/6305

PR #6313, opened 2024-05-16
Add `airtable_record` method to `Event`

sampoder:
Resolves https://github.com/hackclub/hcb/issues/6305. I think this is the best way of approaching this issue.


garyhtou (commented):
🪄

can you come up with a reliable way to backfill data? something that accounts for slug changes

sampoder:
i've backfilled about 1800 of them using slug history:

Screenshot 2024-05-16 at 9 03 13 PM

garyhtou:
can you write a query to see the number of orgs that have shared a slug?

PR #6319, opened 2024-05-17
Add a fancy popover for the bin!

sampoder:
Rummage! We'll need to enable the Flipper flag for everyone in a FIRST team, after that they'll see it once.

https://github.com/hackclub/hcb/assets/39828164/c200a233-c3ff-4ea3-8b1c-7bd84305f1fd


app/views/layouts/application.html.erb
```diff
@@ -145,7 +145,40 @@
             </div>
           </div>
         <% end %>
+        <% if @event && @the_bin %>
```
garyhtou: plz move to partial

app/controllers/concerns/set_event.rb
```diff
@@ -14,6 +14,11 @@ def set_event
       @organizer_position = @event.organizer_positions.find_by(user: current_user) if signed_in?
       @first_time = params[:first_time] || @organizer_position&.first_time?
 
+      if @the_bin && !@event.robotics_team?
+        @the_bin = false
+        Flipper.enable_actor(:the_bin_popup_2024_05_17, current_user)
+      end
```
garyhtou: can you help explain the intentions behind this section?

garyhtou (commented):
is it possible to get both the usual welcome overlay AND the bin overlay at the same time?

app/controllers/concerns/set_event.rb
sampoder: on my phone but basically this was so that we wouldn't render it in contexts where it doesn't make sense.

say I'm both organising a hackathon and on a first team, it'd be weird if this showed up when I directly went to my hackathons finances

app/views/layouts/application.html.erb
```diff
@@ -145,7 +145,40 @@
...
+          <div class="welcome" data-controller="welcome" data-welcome-organizer-position-value="<%= @organizer_position&.id %>">
```
cjdenio:
```suggestion
          <div class="welcome" data-controller="welcome">
```

Otherwise this gets reported as an onboarding tour in our analytics

cjdenio (dismissed):

app/controllers/application_controller.rb
```diff
@@ -23,6 +23,8 @@ class ApplicationController < ActionController::Base
 
     @first_visit = cookies[:first_visit] != "1"
     cookies.permanent[:first_visit] = 1
+
+    @the_bin = Flipper.enabled?(:the_bin_popup_2024_05_17, current_user) && Flipper.disable_actor(:the_bin_popup_2024_05_17, current_user)
```
cjdenio: Can this just be run on `events#show`? I'd rather not be running this Flipper check on every request

YodaLightsabr: I haven't tested it yet but https://github.com/hackclub/hcb/pull/6319/commits/f16137beb02e384657012c01b005104885c9c196 should fix this.

garyhtou: `set_event` might be used in multiple places. i'd suggest

```ruby
# EventsController
before_action only: :show do
  # code
end
```

app/controllers/concerns/set_event.rb
```diff
@@ -14,6 +14,11 @@ def set_event
       @organizer_position = @event.organizer_positions.find_by(user: current_user) if signed_in?
       @first_time = params[:first_time] || @organizer_position&.first_time?
 
+      if Flipper.enabled?(:the_bin_popup_2024_05_17, current_user) && @event.robotics_team?
```
sampoder:
```suggestion
      if Flipper.enabled?(:the_bin_popup_2024_05_17, current_user) && @event.robotics_team? && !@first_time
```

app/controllers/concerns/set_event.rb
```diff
@@ -13,7 +13,6 @@ def set_event
       @event = Event.friendly.find(id)
       @organizer_position = @event.organizer_positions.find_by(user: current_user) if signed_in?
       @first_time = params[:first_time] || @organizer_position&.first_time?
```
sampoder:
```suggestion
      @first_time = params[:first_time] || @organizer_position&.first_time?

```

garyhtou (commented):
🪄

PR #6326, opened 2024-05-20
Enable comments for external report authors

YodaLightsabr:
## Summary of the problem
https://hackclub.slack.com/archives/CN523HLKW/p1716228114662479

## Describe your changes
Previously, the "Add note" button would be disabled if an organizer wasn't signed in. However, this doesn't factor in users who aren't organizers but are still involved in the report.


sampoder (approved):
LGTM

jaspermayone:
Thank you both!

PR #1428, opened 2021-04-06
Edits to generated PDFs (fiscal sponsorship confirmation + ach transfer confirmation)

garyhtou:
ACH Transfer | Fiscal Sponsorship Confirmation
-|-
Screenshot 2021-04-08 234957|Screenshot 2021-04-08 180157

PR #183, opened 2019-07-10
Account balance in transaction list

tmb:
image

A few issues:
* Balances are screwed up on events that have errors with some fee reimbursements (ones i discovered: Windy City Hacks, Spectra)
* They all go sumo if you expand one that it doesn't like (see: 
image)

@MaxWofford will work with you on this tomorrow along with the other PR

PR #6327, opened 2024-05-20
[Increase][Account Numbers] Disable all account numbers

rluodev:
We're no longer using increase so we should disable these.


garyhtou (changes requested):
I'd suggest writing this as a one time job (there's a folder of them). Also, don't worry about the "down".

We usually reserve migrations for changes to the database schema. Sometimes we'll also make "data migrations" in there, but it's not best practice.

garyhtou (approved):

garyhtou:
I ran the job in prod!

## Before running the job

All account numbers were `active`. None were `disabled` or `canceled`
image

## After running the job

1470 `IncreaseAccountNumber`s on HCB are now `disabled`. The only exception is Hack Club HQ.

Looking at Increase's dashboard, there are now 13 account numbers that remain active.
- `ECM Check Deposit (Max testing)`
- `Expensify account`
- `another caleb test`
- `caleb testing`
- `#183 (Hack Club HQ)`  — the one remaining active `IncreaseAccountNumber`
- `Wise`
- `Credit (receive) only`
- `Donations for fiscally sponsored orgs`
- `Payroll (Gusto)`
- `<REDACTED> #1428 (testing payouts Max)` (redacted name of fiscally sponsored org)
- `Venmo`
- `Stripe`
- `Main account number (deposit/withdrawal)`

## Manual disabling

I manually went through the list above to determine whether to disable each one. Here are the results.

|Name|Status|Note/Reason|
|--|--|--|
| `ECM Check Deposit (Max testing)` | Disabled | It has no transactions
| `Expensify account` | Disabled | Last transaction was Feb 27th |
| `another caleb test` | Disabled | no transactions |
| `caleb testing` | Disabled | no transactions |
| `#183 (Hack Club HQ)` | 🟢 **Active**, for now | this is the one remaining active `IncreaseAccountNumber`. Last transaction as of writing was May 3rd |
| `Wise` | Disabled | no transactions |
| `Credit (receive) only` | Disabled | Last transaction as of writing was May 4th, **_2023_** |
| `Donations for fiscally sponsored orgs` | Disabled | no transactions |
| `Payroll (Gusto)` | Disabled | Last transaction was Jan 1st |
| `<REDACTED> #1428 (testing payouts Max)` (redacted name of fiscally sponsored org) | Disabled | no transaction|
| `Venmo` | Disabled | no transactions |
| `Stripe` | Disabled | last transaction was Dec 14th |
| `Main account number (deposit/withdrawal)` | 🟢 **Active**, for now | Last transaction as of writing was May 15th! arugh |

## Final Result

There remain 2 account numbers that have not been disabled.
- Hack Club HQ
- "Main account number (deposit/withdrawal)"

I'm going to work with @hackliv /Ops to shut down these remaining numbers — there are still some large sums moving through some of them as of May 15th.

PR #6328, opened 2024-05-20
Fix external report comments at policy level

YodaLightsabr:
## Summary of the problem
https://hackclub.slack.com/archives/CN523HLKW/p1716229687034189?thread_ts=1716228114.662479&cid=CN523HLKW


sampoder (approved):
This works; thanks Ian!

app/policies/comment_policy.rb
```diff
@@ -35,7 +35,13 @@ def show?
   private
 
   def users
-    record.commentable.respond_to?(:events) ? record.commentable.events.collect(&:users).flatten : record.commentable.event.users
+    if record.commentable.respond_to?(:events)
+      record.commentable.events.collect(&:users).flatten
```
garyhtou: fun fact! `collect(&:users).flatten` can be written as `flat_map(&:users)`

PR #6330, opened 2024-05-21
Check for current user before flipper check & reverse Flipper purpose

sampoder:


sampoder:
Tested this locally, works!

PR #6347, opened 2024-05-22
Prominently display major new features

malted:
Closes https://github.com/hackclub/hcb/issues/6331

image


app/views/layouts/application.html.erb
```diff
@@ -147,7 +147,10 @@
             </div>
           </div>
         <% end %>
+
+        <%= render "application/native_changelog" if @latest_changelog_post %>
```
cjdenio: Looks like this can only show up on `events#show`— it would be great to move this line into there and out of the global layout

app/models/changelog_post.rb
```diff
@@ -0,0 +1,18 @@
...
+#  id           :bigint           not null, primary key
+#  markdown     :string
+#  published_at :datetime
+#  title        :string
+#  created_at   :datetime         not null
+#  updated_at   :datetime         not null
+#  headway_id   :integer
+#
+class ChangelogPost < ApplicationRecord
+  scope :latest, -> { order(published_at: :desc).first }
```
cjdenio: this should be

```suggestion
  def self.latest
    order(published_at: :desc).first
  end
```
, a scope technically works but they're not meant to return a single record

app/jobs/headway_job.rb
```diff
@@ -0,0 +1,46 @@
...
+  def perform
+    connection = Faraday.new(url: "https://headwayapp.co") do |c|
+      c.request :url_encoded
+      c.response :json
+    end
+
+    response = connection.post do |req|
+      req.url "/_/graphql/changelogs"
+      req.headers["Content-Type"] = "application/json"
+      req.headers["Cookie"] = Rails.application.credentials.headway.cookie
```
cjdenio: Whose account is this cookie tied to? There should be a comment in the credentials file for posterity

app/views/layouts/application.html.erb
malted: This wouldn't work because otherwise it's not on top of elements such as the navbar.

Incorrect (show):
image

Correct (application):
image

app/jobs/headway_job.rb
malted: Done! It's tied to my (`malted+headway@hackclub.com`) account.

cjdenio (approved):

PR #6348, opened 2024-05-22
Create "Recently on HCB..." / Activity Log

sampoder:
Screenshot 2024-05-21 at 9 38 18 PM

Plus it has infinite scroll!

https://github.com/hackclub/hcb/assets/39828164/dfc552e5-ee03-424d-96c3-6ab210926cbf


jaspermayone:
Personally my favorite user to impersonate 😆

sampoder:
Oh yes, thank you for your service!

config/initializers/public_activity.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+PublicActivity::Activity.class_eval do
+  scope :for_user, ->(user) {
+    where("user_id = ?", user.id).or(where(event_id: user.events.pluck(:id)))
+  }
+end
```
garyhtou: do we need to use `class_eval`? or can we also just reopen the class?

```suggestion
class PublicActivity::Activity
  scope :for_user, ->(user) {
    where("user_id = ?", user.id).or(where(event_id: user.events.pluck(:id)))
  }
end
```

config/initializers/public_activity.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+PublicActivity::Activity.class_eval do
+  scope :for_user, ->(user) {
+    where("user_id = ?", user.id).or(where(event_id: user.events.pluck(:id)))
```
garyhtou: Is there a case where conceptionally, there are multiple users? How do we decide who's the owner and who's the `event_id`?

app/views/public_activity/_common.html.erb
garyhtou, line 1: very clean! I love this ahhhh. great job here

app/models/user.rb
```diff
@@ -44,6 +44,9 @@ class User < ApplicationRecord
 
   has_paper_trail only: [:access_level]
 
+  include PublicActivity::Model
+  tracked owner: proc{ |controller, model| controller&.current_user || User.find_by(email: "bank@hackclub.com") }, user_id: proc { |controller, model| model.id }, only: [:create, :update]
```
garyhtou: from the usage, `model` feels like the wrong word here. maybe `record`?

I'm not sure what class/type that argument holds. but usually, modal refers to the class

app/models/donation.rb
```diff
@@ -90,6 +93,9 @@ class Donation < ApplicationRecord
 
     event :mark_in_transit do
       transitions from: :pending, to: :in_transit
+      after do
+        create_activity(key: "donation.paid", owner: User.find_or_create_by!(email:))
```
garyhtou: with the current code, If i paid a donation using `gary@hackclub.com`, i'm assume it would not show up in my log because `user_id` is not set. what are our thoughts on this?

app/views/static_pages/index.html.erb
```diff
@@ -102,4 +102,5 @@
     <% end %>
   <% end %>
   <%= render partial: "static_pages/admin" %>
+  <%= render template: "static_pages/my_activities" %>
```
garyhtou: is curious why this is rendered as a template

db/migrate/20240521170127_create_activities.rb
```diff
@@ -0,0 +1,25 @@
...
+
+# Migration responsible for creating a table with activities
+class CreateActivities < ActiveRecord::Migration[6.1]
+  def self.up
+    create_table :activities do |t|
+      t.belongs_to :trackable, polymorphic: true
+      t.belongs_to :owner, polymorphic: true
+      t.string :key
+      t.text :parameters
+      t.belongs_to :recipient, polymorphic: true
```
garyhtou: do we know the gem's intended purpose for this field? it appears we're leaving it unused

db/migrate/20240521175040_add_user_id_event_id_to_activity.rb
```diff
@@ -0,0 +1,6 @@
+class AddUserIdEventIdToActivity < ActiveRecord::Migration[7.1]
+  def change
+    add_column :activities, :user_id, :integer
+    add_column :activities, :event_id, :integer
```
garyhtou: I think we want to `add_reference` so that they're foreign keys in the DB with indices

garyhtou (commented):
you're speedy quick!! 🏎️💨

config/initializers/public_activity.rb
sampoder: Oh sweet! Their docs' examples are a bit old

app/models/user.rb
sampoder: Yes... this confused me as well. Their docs used `model`, let's go with `record`

config/initializers/public_activity.rb
garyhtou: i just did some research on it! it looks like both works. class_eval might be preferred — I wasn't aware of that!

https://stackoverflow.com/questions/22291235/class-eval-and-open-classes

db/migrate/20240521170127_create_activities.rb
sampoder: yes, I just looked into this. looks like it's similar to our `user_id` use case. I'm going to replace that with `recipient`. it's polymorphic so if in the future we want to do https://github.com/hackclub/hcb/pull/6348#discussion_r1609163902 we can.

sampoder: https://rubydoc.info/gems/public_activity/PublicActivity/Common#activity_recipient-instance_method

config/initializers/public_activity.rb
sampoder: https://github.com/hackclub/hcb/pull/6348#discussion_r1609184646

db/migrate/20240521170127_create_activities.rb
sampoder: if we want multiple recipients, we can create a table such as `RecipientGroup` to send activities to multiple recipients

app/models/donation.rb
sampoder: that's correct, for now, I think that's alright. as a donor, I don't think I ever make the connection between me making a donation and my HCB account

sampoder:
Here's my thoughts about moving forward this. I don't think the UI is perfect yet, I'd like that to be hidden behind a Flipper flag but I would love starting to collect these activities so we have data to test with and also so that when we launch the feed isn't blank.

jaspermayone:
Would love to early test via flipper :)

app/controllers/static_pages_controller.rb
```diff
@@ -14,6 +14,7 @@ def index
       @events = @service.events
       @organizer_positions = @service.organizer_positions.not_hidden
       @invites = @service.invites
+      @activities = PublicActivity::Activity.for_user(current_user).page(1).per(1)
```
```suggestion
      @activities = PublicActivity::Activity.for_user(current_user).order(created_at: :desc).page(1).per(25)
```

app/controllers/static_pages_controller.rb
```diff
@@ -25,6 +26,11 @@ def index
     end
   end
 
+  def my_activities
+    @activities = PublicActivity::Activity.for_user(current_user).page(params[:page]).per(params[:per] || 1)
```
```suggestion
    @activities = PublicActivity::Activity.for_user(current_user).order(created_at: :desc).page(params[:page]).per(25)
```

app/views/public_activity/reimbursement_expense/_approved.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= render layout: "/public_activity/common", locals: { activity:, icon: "checkmark", color: "success" } do %>
+  <span class="align-middle">
+    approved <i><%= activity.trackable.memo %></i> on your report, <%= link_to activity.trackable.report.name, activity.trackable.report %>
```
garyhtou:
```suggestion
    approved <i><%= activity.trackable.memo %></i> on your reimbursement report, <%= link_to activity.trackable.report.name, activity.trackable.report %>
```

app/views/public_activity/organizer_position_invite/_create.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= render layout: "/public_activity/common", locals: { activity:, icon: "member-add", color: "info" } do %>
+  <span class="align-middle">
+    invited
+  </span>
+  <span class="muted" style="margin-right: -0.25em;">
+    <%= user_mention User.find(activity.user_id), { avatar: { style: "vertical-align: sub; transform: translateY(3px); margin-left: 2px;" } }, "Unknown User" %>
```
garyhtou: should this be recipient? my guess is that we can just use `activity.recipient` instead of finding the id

app/views/public_activity/reimbursement_report/_create.html.erb
```diff
@@ -0,0 +1,16 @@
+<%= render layout: "/public_activity/common", locals: { activity:, icon: "attachment", color: "info" } do %>
+  <span class="align-middle">
+    opened <i><%= activity.trackable.name %></i>
+  </span>
+  <% unless activity.trackable.user == activity.owner %>
```
garyhtou:
```suggestion
  <% unless activity.trackable.recipient == activity.owner %>
```

app/views/public_activity/reimbursement_report/_create.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    <span class="align-middle">
+      for
+    </span>
+    <span class="muted" style="margin-right: -0.25em;">
+      <%= user_mention activity.trackable.user, { avatar: { style: "vertical-align: sub; transform: translateY(3px); margin-left: 2px;" } }, "Unknown User" %>
+    </span>
+  <% end %>
+  <span class="align-middle">
+    on <%= link_to activity.trackable.event.name, activity.trackable.event %>
```
garyhtou:
```suggestion
    on <%= link_to activity.trackable.event.name, activity.trackable.event %>
    for reimbursement
```

app/views/public_activity/user_session/_create.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= render layout: "/public_activity/common", locals: { activity:, icon: "door-enter", color: "primary" } do %>
+  <span class="align-middle">
+    logged into HCB<%= activity.trackable&.device_info.present? ? " on #{activity.trackable.os_info}" : "" %>
```
garyhtou:
```suggestion
    logged into HCB<%= " on #{activity.trackable.os_info}" if activity.trackable&.os_info.present? %>
```

app/views/static_pages/index.html.erb
```diff
@@ -102,4 +102,5 @@
     <% end %>
   <% end %>
   <%= render partial: "static_pages/admin" %>
+  <%= render partial: "static_pages/my_activities" %>
```
garyhtou: Feature flag

app/models/donation.rb
garyhtou: If we end up creating a user, can we set it's name since we have that info

garyhtou (commented):
🪄

app/views/public_activity/organizer_position_invite/_create.html.erb
sampoder:
```suggestion
    <%= user_mention activity.recipient, { avatar: { style: "vertical-align: sub; transform: translateY(3px); margin-left: 2px;" } }, "Unknown User" %>
```

app/models/donation.rb
sampoder:
```suggestion
        create_activity(key: "donation.paid", owner: User.create_with(full_name: name).find_or_create_by!(email:))
```

app/views/static_pages/index.html.erb
sampoder:
```suggestion
  <%= render partial: "static_pages/my_activities" if Flipper.enabled?(:recently_on_hcb_2024_05_23, current_user) %>
```

PR #6350, opened 2024-05-22
Add "Review Required" to `/my/reimbursements`

sampoder:
Screenshot 2024-05-22 at 12 38 23 AM

I believe this closes the underlying issue in https://github.com/hackclub/hcb/issues/6338.


app/views/static_pages/my_reimbursements.html.erb
```diff
@@ -33,37 +33,38 @@
...
     <%= form.text_field :q, placeholder: "Search…", class: "flex-auto md-mr2", style: "width auto; max-width: none;", value: params[:q] %>
-
+    <%= link_to "?filter=my", class: "filterbar__item", "aria-selected": params[:filter] != "review_required", role: "tab" do %>
+      My Reports
+      <span class="m0 badge <%= params[:filter] == "review_required" ? "bg-primary" : "bg-snow text-black" %>" style="margin-left: 2px; transform: translateY(-1px)">
+        <%= current_user.reimbursement_reports.draft.count %>
+      </span>
+    <% end %>
+    <%= link_to "?filter=review_required", class: "filterbar__item", "aria-selected": params[:filter] == "review_required", role: "tab" do %>
+      Review Required
```
garyhtou: I'm not sure if I love "Review Required". What are your thoughts on "Review Requested" or "Awaiting Review"

sampoder: Cool with either, not too attached to "review required"

app/controllers/static_pages_controller.rb
```diff
@@ -162,16 +162,15 @@ def my_inbox
...
-    @reports = @reports.rejected if params[:filter] == "rejected"
+    @reports = current_user.reimbursement_reports unless params[:filter] == "review_required"
+    @reports = current_user.reimbursement_reports_to_review if params[:filter] == "review_required"
     @reports = @reports.search(params[:q]) if params[:q].present?
     @payout_method = current_user.payout_method
   end
 
   def my_draft_reimbursements_icon
     @draft_reimbursements_count = current_user.reimbursement_reports.draft.count
+    @review_required_reimbursements_count = current_user.reimbursement_reports_to_review.submitted.count
```
garyhtou: What are your thoughts on updating the `reimbursement_reports_to_review` to include a `-> { submitted }` by default? having "to_review" in the association name makes it like it's a list that actively requires the user's review, however, that's not the case.

garyhtou (commented):
🪄 thanks!

let's keep https://github.com/hackclub/hcb/issues/6338 open for now. i think this addresses a good portion of the needs. however, i'd like to fully hear out Dylan's idea before we close it

app/views/static_pages/my_reimbursements.html.erb
sampoder:
```suggestion
      Review Requested
```

dwahbe:
This looks great! A small thing but the capitalization on "My Reports" and " Review Required" should be sentence case, "My reports" and "Review required".

dwahbe:
Two other things. 

When I switch between "My reports" and "Review requested", it adds a scroll bar that shifts everything over a bit. Is there a way to change this so it doesn't shift?

For "My reports", it shows 0 next to them. I'm confused what this means. I have lots of reports created with different statuses. Is this a bug? Is that number meant to be the total number of reports or total draft reports? 

Brief screen recording below. 

https://github.com/hackclub/hcb/assets/153225786/92ad38a0-74a5-4898-b4e3-1f2b5794e9cb

dwahbe:
@sampoder just following up on these.

sampoder:
The scrollbar shift is by design, the page is now longer and needs a scrollbar. It's a browser related thing.

I'll quickly remove the 0 badges. I think we should have a conversation about sentence casing, I don't think it looks visually appealing in this use case, let's discuss in `#hcb-engr`.

dwahbe:
Sounds good for scrollbar. 

For the 0 badges, what was the initial thought there? I think it's a good thing to have, especially for "Reimbursements requested". 

Happy to discuss sentence case, but we should be consistent. Right now, almost all of HCB is using sentence case with some exceptions (Lachlan had it all on sentence case and as HCB grew, we got some inconsistencies). I created #6355 recently on this as well. So the discussion is more about whether we want to continue using sentence case or switch.

sampoder:
My plan is to hide them currently, just made a PR that did that.

I think we should do what looks best on a case by case basis

PR #6351, opened 2024-05-22
Add `last_seen_at` to `UserSession`

sampoder:
Closes https://github.com/hackclub/hcb/issues/6343


app/models/user.rb
```diff
@@ -287,6 +287,10 @@ def hack_clubber?
     return events.organized_by_hack_clubbers.any?
   end
 
+  def last_active_at
+    user_sessions.where.not(last_active_at: nil).order(last_active_at: :desc).first&.last_active_at
```
garyhtou: i'm guessing the `user_sessions.max(:last_active_at)` didn't work?

would something like `user_sessions.select("max(last_active_at)")` work too?

garyhtou (commented):
@albertchae do you have any thoughts on this implementation's performance and if/how it could be improved?

more context: https://github.com/hackclub/hcb/issues/6343

app/models/user_session.rb
```diff
@@ -104,6 +105,12 @@ def set_as_peacefully_expired
     self
   end
 
+  def touch_last_active_at
+    return if last_active_at&.after? 15.minutes.ago # prevent spamming writes
+
+    UserSessionJob::UpdateLastActiveAt.perform_later(user_session: self, last_active_at: Time.now)
```
northeastprince: Why is this offloaded? There's no notable database load or third party calls.

garyhtou: hitting redis is quicker than postgres

app/models/user_session.rb
```diff
@@ -104,6 +105,12 @@ def set_as_peacefully_expired
...
+  end
```
garyhtou: Actually, 15 minutes might be unnecessarily long. Maybe we make this 5 minutes. If we decide to show this in the UI (such as an admin tool), we can say "Last seen now" if `last_active_at` was within the past 10 mins.


"Now" feels inaccurate when it's longer than 10-15 mins.

```suggestion
  LAST_ACTIVE_AT_COOLDOWN = 5.minutes
  def touch_last_active_at
    return if last_active_at&.after? LAST_ACTIVE_AT_COOLDOWN.ago # prevent spamming writes

    UserSessionJob::UpdateLastActiveAt.perform_later(user_session: self, last_active_at: Time.now)
  end
```

garyhtou (commented):
I gave the naming some thought. Is it cool if we rename this to `last_seen_at`? I think that's less ambiguous than "active/activity". It's also what a lot of apps call it

app/models/user_session.rb
```diff
@@ -104,6 +105,13 @@ def set_as_peacefully_expired
     self
   end
 
+  LAST_SEEN_AT_COOLDOWN = 5.minutes
+  def touch_last_seen_at
+    return if last_seen_at&.after? LAST_SEEN_AT_COOLDOWN.ago # prevent spamming writes
+
+    UserSessionJob::UpdateLastSeenAt.perform_later(user_session: self, last_seen_at: Time.now)
```
albertchae: one issue with this approach is that even if you have a cooldown for 5 minutes, you might get a bunch of writes come in after that period (especially with offloading writes to sidekiq, but I think it could happen even if you inlined `user_session.update!(last_seen_at:)` here)

That would happen when a page makes multiple API requests to the rails server. Each of these requests in parallel would hit this check and see that cooldown has passed and queue up jobs to update the timestamp

app/jobs/user_session_job/update_last_seen_at.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module UserSessionJob
+  class UpdateLastSeenAt < ApplicationJob
+    queue_as :low
+
+    def perform(user_session:, last_seen_at:)
+      user_session.update!(last_seen_at:)
```
albertchae: @garyhtou would we want to update this column without also touching `updated_at`?  Because we could use https://api.rubyonrails.org/v7.1.3.3/classes/ActiveRecord/Persistence.html#method-i-touch

app/models/user.rb
```diff
@@ -287,6 +287,10 @@ def hack_clubber?
     return events.organized_by_hack_clubbers.any?
   end
 
+  def last_seen_at
+    user_sessions.maximum(:last_seen_at)
```
albertchae: We only keep one `user_session` record per device a user is using right? Because `maximum` could be slow with a lot of records, but I'm guessing we wouldn't have too many users over ~5 sessions

app/models/user_session.rb
```diff
@@ -10,6 +10,7 @@
 #  expiration_at            :datetime         not null
 #  fingerprint              :string
 #  ip                       :string
+#  last_seen_at           :datetime
```
albertchae: minor but weird this alignment is off. The annotate gem should re-align these comments when necessary 🤔

app/models/user_session.rb
albertchae: Discussed on slack with @garyhtou but I'm not sure how much we gain with the redis offloading if we aren't also batching up writes or deduping writes for the same user after. I think we should probably go with simplest implementation first and identify whether there is a noticeable increase in DB CPU usage before implementing more complicated solutions

app/models/user.rb
garyhtou: median is two sessions. average is 5. max is 448

I think these numbers will grow over time since when sessions expire, users are required to login again, which creates a new session.

Any ideas on how to make this more performant?

app/jobs/user_session_job/update_last_seen_at.rb
garyhtou: That's probably a good idea! Let's update `last_seen_at` without also touching `updated_at`

app/models/user_session.rb
garyhtou: Yeah, let's inline and monitor to see if it causes an issue

garyhtou:
@sampoder for this PR,
- let's drop the job
- update `last_seen_at` without touching `updated_at`. I think `update_column` may help with this

garyhtou (dismissed):
🪄 ty

sampoder:
can someone run this migration locally and test that it works for them as well?

sampoder:
ben tested it!

PR #6356, opened 2024-05-22
Create a 2FA Authentication System

sampoder:
Closes https://github.com/hackclub/hcb/issues/5811. Here's how this system works.

Sessions are now created "unauthenticated", they are only considered "authenticated" when we set the cookie in the user's browser. We create them early so that we can track which methods it has been authorised using. For example, after a user successfully enters a login code received in the email `user_session.authenticated_with_email?` will return true. By default, users just need one of `user_session.authenticated_with_email?`, `user_session.authenticated_with_sms?` or, `user_session.authenticated_with_webauthn?` to return true for their session to transition to "authenticated". If they have two factor authentication enabled and set up (currently there's a `user.using_2fa?` helper), they need two.

Nothing changes for users on 1FA. But for users on 2FA, they'll have to authenticate themselves twice and the UI will walk them through that.

This is after authenticating with my fingerprint:

Screenshot 2024-05-22 at 3 36 58 PM


app/views/user_sessions/choose_login_preference.html.erb
```diff
@@ -27,7 +27,7 @@
 
   <p>How would you like to sign in?</p>
 
-  <%= form_with url: set_login_preference_users_path, method: :post, data: { "controller" => "webauthn-auth form-disable", "webauthn-auth-target" => "authForm", "action" => "webauthn-auth#submit", webauthn_auth_return_to_value: @return_to } do |form| %>
+  <%= form_with url: set_login_preference_user_session_path(@session), method: :post, data: { "controller" => "webauthn-auth form-disable", "webauthn-auth-target" => "authForm", "action" => "webauthn-auth#submit", webauthn_auth_return_to_value: @return_to, turbo: false, session_id_value: @session.id } do |form| %>
```
sampoder:
```suggestion
  <%= form_with url: set_login_preference_user_session_path(@session), method: :post, data: { "controller" => "webauthn-auth form-disable", "webauthn-auth-target" => "authForm", "action" => "webauthn-auth#submit", webauthn_auth_return_to_value: @return_to, session_id_value: @session.id } do |form| %>
```

Gemfile
```diff
@@ -210,3 +210,5 @@ gem "doorkeeper", "~> 5.6"
 gem "cssbundling-rails", "~> 1.4"
 
 gem "sprockets-rails", "~> 3.4"
+
+gem "enummer"
```
garyhtou: I'd suggest using https://api.rubyonrails.org/classes/ActiveRecord/Store.html

PR #6360, opened 2024-05-23
[Reimbursement] Change HCB review step button from `mark as draft` to `return to [user]`

rluodev:
## Summary of the problem

Confusing wording/inability to request changes with a comment attached to the mark as draft

## Describe your changes

Admins now see `return to [user]` unless they're the report owner, in which case they see `mark as draft` still.

No visual changes aside from logic flow, which is difficult to show with pictures.


rluodev:
Please sanity check this logic fix (I've tested and I think it should work, but my brain also may be dead)

garyhtou:
I'll let @sampoder review this one!

app/views/reimbursement/reports/_actions.html.erb
```diff
@@ -159,7 +159,7 @@
               <span class="bold">Status:</span> the HCB team is reviewing this report. To make further changes, mark it as a draft.
             <% end %>
           <% end %>
-        <% elsif (!report.event.users.include?(current_user) || current_user == report.user) && report.unlockable? && !admin_signed_in? %>
+        <% elsif current_user == report.user && (report.unlockable? || admin_signed_in) %>
```
sampoder:
```suggestion
        <% elsif current_user == report.user && (report.unlockable? || admin_signed_in?) %>
```

PR #6370, opened 2024-05-23
Allow overriding port with `bin/dev`

cjdenio:
Now, `bin/dev` defaults to port 3000 while `bin/dev -p 4000` can be used to run HCB on a different port.


bin/dev
garyhtou, line 1: clever!

garyhtou (commented):
ty

PR #6381, opened 2024-05-24
[Migrations] Comment out problematic data migrations

garyhtou:
## Summary of the problem

The application code has changed since these migrations were written. These assumptions have not held up, thus causing errors.

## Describe your changes

Since we don't really need these migrations anymore, I'm commenting out the data migration portion. We should restrict migrations to only schema changes (such as by using one time jobs)


garyhtou:
I've confirmed this fixes the issue by running the migrations from scratch.

```sh
bin/rails db:drop db:create db:migrate
```

PR #6382, opened 2024-05-24
[Event] Update annotation

garyhtou:
I get this annotation often, and it seems to be valid based on [this migration code](https://github.com/hackclub/bank/blob/6da8311f7fa54883bc03100b03e834caf635d5b7/db/migrate/20210213222744_change_default_for_pending_transaction_engine_at.rb#L7).


rluodev:
yes please fix this it annoys my ocd self so much

PR #6387, opened 2024-05-24
Rescue User Creation when making donation activities

sampoder:
This was causing webhooks to fail if they didn't have their legal name. After merging I'll resend the webhooks.


app/models/donation.rb
```diff
@@ -94,7 +94,11 @@ class Donation < ApplicationRecord
     event :mark_in_transit do
       transitions from: :pending, to: :in_transit
       after do
-        create_activity(key: "donation.paid", owner: User.create_with(full_name: name).find_or_create_by!(email:))
+        begin
+          create_activity(key: "donation.paid", owner: User.create_with(full_name: name).find_or_create_by!(email:))
```
garyhtou: It's preferred to pass a block to `find_or_create_by`

https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_create_by

PR #6390, opened 2024-05-24
Integrate with Loops

rluodev:
## Summary of the problem

@zachlatta requested that we add HCB teenager users to Loops.

## Describe your changes

Whenever a user gets updated or created, we'll send their info to loops to create a new Loops `Contact`.

We'll also run a one-time job to create `Contact`s on Loops for all HCB teenager users.

They'll be added to loops default subscribed, and can opt out of receiving promo emails by unsubscribing through a loops email

image


app/services/user_service/sync_with_loops.rb
```diff
@@ -0,0 +1,82 @@
...
+    def teenager
+      @user.birthday&.after?(19.years.ago) || @user.includes(organizer_positions: :event).where(organizer_positions: { events: { category: "high school hackathon" } })
+    end
+
+    def using_default_address
+      @user.stripe_cardholder.stripe_billing_address_line1.downcase.strip == "8605 santa monica blvd #86294" &&
+        @user.stripe_cardholder.stripe_billing_address_city.downcase.strip == "west hollywood" &&
+        @user.stripe_cardholder.stripe_billing_address_state.downcase.strip == "ca" &&
+        @user.stripe_cardholder.stripe_billing_address_postal_code.downcase.strip == "90069" &&
+        @user.stripe_cardholder.stripe_billing_address_country.downcase.strip == "us"
```
garyhtou: i'd suggest doing a case insensitive check against https://github.com/hackclub/bank/blob/d5d2b851d37efc80413b34c82b498053814bae6a/app/models/stripe_cardholder.rb#L94

app/services/user_service/sync_with_loops.rb
```diff
@@ -0,0 +1,82 @@
...
+          h[:addressZipCode] = !using_default_address ? @user.stripe_cardholder.stripe_billing_address_postal_code : ""
+          h[:addressCountry] = !using_default_address ? @user.stripe_cardholder.stripe_billing_address_country : ""
+          h[:addressLastUpdatedAt] = @user.stripe_cardholder.updated_at.to_time
+        end
+      end.merge({
+                  hcbSignedUpAt: @user.created_at.to_time,
+                  hcbLastLoginAt: @user.user_sessions.order(created_at: :desc).last.created_at.to_time
+                })
+
+      puts @user.stripe_cardholder.updated_at.to_time
```
sampoder:
```suggestion
```

rluodev:
Did not mean to mark as ready, still need to write job

rluodev:
@garyhtou should we also add a consent checkbox on the account creation/settings page for subscribing to emails?

I'm thinking the wording should be something like:

`I'd like to receive updates and promotional emails from Hack Club`

If we don't have this, I think this might not be GDPR or other privacy law compliant.

rluodev:
I'm having it set to subscribed by default; lmk if we should have it be unsubscribed by default

zachlatta:
Yes, subscribed by default.

The opt out in settings is redundant because Loops manages subscription status and they opt out from the emails. Please cut that section from the page.

Please do not add an opt-in checkbox. Every teenager needs to be added by default to Loops. We can have a minimalist and small disclaimer (do not use the word "promotional" in it) if required.

app/services/user_service/sync_with_loops.rb
```diff
@@ -0,0 +1,78 @@
...
+        if @user.stripe_cardholder # we only add their address if they are a stripe cardholder; doesn't make sense otherwise
+          h[:addressLine1] = !using_default_address ? @user.stripe_cardholder.stripe_billing_address_line1 : ""
+          h[:addressLine2] = @user.stripe_cardholder.stripe_billing_address_line2 unless @user.stripe_cardholder.stripe_billing_address_line2.nil? || @user.stripe_cardholder.stripe_billing_address_line2 == ""
+          h[:addressCity] = !using_default_address ? @user.stripe_cardholder.stripe_billing_address_city : ""
+          h[:addressState] = !using_default_address ? @user.stripe_cardholder.stripe_billing_address_state : ""
+          h[:addressZipCode] = !using_default_address ? @user.stripe_cardholder.stripe_billing_address_postal_code : ""
+          h[:addressCountry] = !using_default_address ? @user.stripe_cardholder.stripe_billing_address_country : ""
+          h[:addressLastUpdatedAt] = @user.stripe_cardholder.updated_at.to_time
+          h[:hcbLastLoginAt] = @user.user_sessions.order(created_at: :desc).last.created_at.to_time unless @user.user_sessions.order(created_at: :desc).last.nil?
```
sampoder: This should be pulled out of the `if @user.stripe_cardholder`. We should also be using `@user.last_seen_at`,

app/services/user_service/sync_with_loops.rb
sampoder: `StripeCardholder` has `default_billing_address?` we should be using that, though maybe we need to make that method more robust.

app/models/user.rb
```diff
@@ -292,6 +295,10 @@ def hack_clubber?
     return events.organized_by_hack_clubbers.any?
   end
 
+  def teenager?
+    birthday&.after?(19.years.ago) || organizer_positions&.joins(:event)&.where(events: { category: "high school hackathon" })&.exists? || organizer_positions&.map(&:event)&.map(&:organized_by_teenagers?)&.include?(true)
+  end
```
sampoder: Are all these safe navigators necessary? I think this can be simplified to:

```suggestion
  def teenager?
    birthday&.after?(19.years.ago) || events.high_school_hackathon.any? || events.organized_by_teenagers.any?
  end
```

sampoder (commented):
Couple of quick suggestions!

app/services/user_service/sync_with_loops.rb
rluodev: so the thing about that is if there's no stripe_cardholder it'll error when I try to pull the address

rluodev:
It's default subscribed, but only if the contact is newly added from HCB and does not previously exist, so we don't resubscribe them every time they update their profile.

zachlatta:
Sounds good. Please re-subscribe them if they are new to HCB, but not if
they update their profile.

app/services/user_service/sync_with_loops.rb
sampoder: I'm referring to:

```
h[:hcbLastLoginAt] = @user.user_sessions.order(created_at: :desc).last.created_at.to_time unless @user.user_sessions.order(created_at: :desc).last.nil?
```

sampoder: Sorry, GitHub doesn't make that super clear

rluodev: oop i see what you mean now

app/services/user_service/sync_with_loops.rb
```diff
@@ -0,0 +1,75 @@
...
+        h[:userGroup] = "Hack Clubber" if @user.teenager?
+        h[:source] = "HCB" unless contact_exists?
+        if @user.stripe_cardholder # we only add their address if they are a stripe cardholder; doesn't make sense otherwise
+          h[:addressLine1] = !using_default_address ? @user.stripe_cardholder.stripe_billing_address_line1 : ""
+          h[:addressLine2] = @user.stripe_cardholder.stripe_billing_address_line2 unless @user.stripe_cardholder.stripe_billing_address_line2.nil? || @user.stripe_cardholder.stripe_billing_address_line2 == ""
+          h[:addressCity] = !using_default_address ? @user.stripe_cardholder.stripe_billing_address_city : ""
+          h[:addressState] = !using_default_address ? @user.stripe_cardholder.stripe_billing_address_state : ""
+          h[:addressZipCode] = !using_default_address ? @user.stripe_cardholder.stripe_billing_address_postal_code : ""
+          h[:addressCountry] = !using_default_address ? @user.stripe_cardholder.stripe_billing_address_country : ""
+          h[:addressLastUpdatedAt] = @user.stripe_cardholder.updated_at.to_time
```
zachlatta: Please only override the address if the address in Loops is empty or if `addressLastUpdatedAt` is older than `@user.stripe_cardholder.updated_at`

zachlatta: So basically - set the address if it's empty, or if HCB has a more recent address on file.

app/models/user.rb
```diff
@@ -303,6 +305,10 @@ def hack_clubber?
     return events.organized_by_hack_clubbers.any?
   end
 
+  def teenager?
+    birthday&.after?(19.years.ago) || events.high_school_hackathon.any? || events.organized_by_teenagers.any?
```
```suggestion
    birthday&.after?(19.years.ago) || events.high_school_hackathon.or(events.organized_by_teenagers).any?
```

app/services/user_service/sync_with_loops.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+
+module UserService
+  class SyncWithLoops
+    def initialize(user_id:)
+      @user = User.find(user_id)
+      @queue = Limiter::RateQueue.new(10, interval: 1)
```
garyhtou: We're creating a new queue for every sync. Instead, we want to share one globally

app/services/user_service/sync_with_loops.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    def contact_exists?
+      @queue.shift
+      conn = Faraday.new(url: "https://app.loops.so/")
+
+      resp = conn.send(:get) do |req|
+        req.url("api/v1/contacts/find")
+        req.headers["Authorization"] = "Bearer #{Rails.application.credentials.loops_key}"
+        req.params[:email] = @user.email
+      end
```
garyhtou: See if you can configure Faraday to use 529 instead of rate limiting ourselves

app/services/user_service/sync_with_loops.rb
```diff
@@ -0,0 +1,84 @@
...
+        addressState: cardholder.address_state,
+        addressZipCode: cardholder.address_postal_code,
+        addressCountry: cardholder.address_country,
+        addressLastUpdatedAt: format_unix(cardholder.updated_at)
+      }
+    end
+
+    def contact_exists?
+      @queue.shift
+      conn = Faraday.new(url: "https://app.loops.so/")
```
garyhtou: See if you can share the connection setup since they're so similar.

https://github.com/hackclub/bank/blob/78bebb0446d17f8dbfd61a4ec19e598189520c53/app/services/api_service/v2/deliver_webhook.rb#L38-L44

zachlatta:
Can I get an update on this, and timeline to have this live?

garyhtou:
Hi! @zachlatta, Ruien is finishing up school this week. I'm meeting with him tomorrow to help unblock this PR. I estimate we'll have this live next week.

Let me know if you need this sooner and I can pull in other engineers, if necessary.

app/services/user_service/sync_with_loops.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    def loops_has_address?
+      @contact_details["addressLine1"].present? &&
+        @contact_details["addressLine2"].present? &&
+        @contact_details["addressCity"].present? &&
+        @contact_details["addressState"].present? &&
+        @contact_details["addressZipCode"].present? &&
+        @contact_details["addressCountry"].present?
+    end
```
garyhtou:
```suggestion
      # To consider a contact as having an address, it must have a line 1, city, and country present.
      # Some international addresses don't have the concept for states or zip codes.
      @contact_details["addressLine1"].present? &&
        @contact_details["addressCity"].present? &&
        @contact_details["addressCountry"].present?
```

app/models/user.rb
```diff
@@ -365,4 +375,8 @@ def valid_payout_method
     end
   end
 
+  def sync_with_loops
+    UserService::SyncWithLoops.new(user_id: id, new_user: onboarding?).run if teenager?
```
```suggestion
    new_user = full_name_before_last_change.blank? && !onboarding?
    UserService::SyncWithLoops.new(user_id: id, new_user:).run if teenager?
```

garyhtou (commented):
Last couple feedback (written during on call with @rluodev)

rluodev:
@zachlatta the wording I've gone with on the user creation page is this statement "By signing up for HCB, you consent to receiving emails from Hack Club."

app/services/user_service/sync_with_loops.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    def format_unix(timestamp)
+      timestamp&.to_datetime&.strftime("%Q")&.to_i # as milliseconds
+    end
+
+      # To consider a contact as having an address, it must have a line 1, city, and country present.
+      # Some international addresses don't have the concept for states or zip codes.
+      @contact_details["addressLine1"].present? &&
+        @contact_details["addressCity"].present? &&
+        @contact_details["addressCountry"].present?
```
rluodev:
```suggestion
    def loops_has_address?
      # To consider a contact as having an address, it must have a line 1, city, and country present.
      # Some international addresses don't have the concept for states or zip codes.
      @contact_details["addressLine1"].present? &&
        @contact_details["addressCity"].present? &&
        @contact_details["addressCountry"].present?
    end
```

restore accidentally deleted method def

rluodev:
Screenshot 2024-06-05 at 6 01 31 PM

(wording at bottom)

zachlatta:
I think it probably makes more sense for the notice to be next to the email field. Is having a notice required? If not, let's please not have a notice. We will set it up so they quickly get a welcome email after signing up, which they can unsubscribe from if they don't want it.

app/services/user_service/sync_with_loops.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    def update(body:)
+      @queue.shift
+      conn = Faraday.new(url: "https://app.loops.so/")
+
+      conn.send(:post) do |req|
+        req.url("api/v1/contacts/update")
+        req.body = body.to_json
+        req.headers["Content-Type"] = "application/json"
+        req.headers["Authorization"] = "Bearer #{Rails.application.credentials.loops_key}"
```
sampoder: Let's make sure we edit creds before merging

app/views/users/edit.html.erb
```diff
@@ -188,8 +188,10 @@
         <%= form.submit @onboarding ? "Start using HCB" : "Save profile" %>
       </div>
     <% end %>
-    <% unless @onboarding %>
-
+    <% if @onboarding %>
+      <p class="h5 muted mt1 mb2">
+        By signing up for HCB, you consent to receiving emails from Hack Club.
+      </p>
     <% end %>
```
sampoder: This looks a little awkward to me, +1 on Zach's comment

sampoder (dismissed):
Good to go, thanks Ruien for all the great work on this!

app/views/users/edit.html.erb
garyhtou: i've removed it. we'll handle it in a separate PR!

garyhtou (approved):

PR #6395, opened 2024-05-25
Add public activity to `RawPendingStripeTransaction`

sampoder:


garyhtou:
I'm not sure if we want this at the "Raw" level. i think it might belong higher in the transaction engine

sampoder:
I initially thought that as well and was trying to find a better place, however, this felt like the only place where on create we could reliably get a event and a user. Any alternatives you'd prefer?

app/views/public_activity/raw_pending_stripe_transaction/_create.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= render layout: "/public_activity/common", locals: { activity:, icon: "checkmark", color: "success" } do %>
+  spent <%= render_money activity.trackable.canonical_pending_transaction.local_hcb_code.amount_cents.abs %> on <i><%= render_money activity.trackable.canonical_pending_transaction.local_hcb_code.memo %></i>
```
garyhtou: let's make sure this handles refunds

garyhtou (commented):
🪄

PR #6396, opened 2024-05-25
Fix unrelated "owners" of new users

sampoder:


garyhtou (commented):
can you help me understand in what case we use this?

i'm assuming for `user.create`. maybe the solution here should be to ensure the
```
activity.owner == the_new_user
```

sampoder:
Yes, sorry, that's the other fix that needs to be made.

PR #6400, opened 2024-05-25
Cache `/stats`

sampoder:
Cache it. Temporary fix. Tested locally, seems to work fine.


YodaLightsabr (dismissed):

YodaLightsabr (approved):

sampoder:
Did a bit more testing, the caching works (use `rails dev:cache`).

PR #6401, opened 2024-05-25
Fix maintenance mode logo

YodaLightsabr:
## Summary of the problem
Maintenance mode logo is hosted on hcb.hackclub.com which doesn't work when maintenance mode is enabled.

## Describe your changes
Host the image on CDN


sampoder (approved):

sampoder:
nice one

garyhtou:
lol good catch

PR #6403, opened 2024-05-26
Add search support for Increase checks

24c02:
## Summary of the problem

Increase checks didn't have a search scope/weren't filtered, thus they were returned as a result for any search on Transfers pages.

## Describe your changes

I added a pg_search scope to IncreaseCheck and a test for it in EventsController following the pattern from the Check model.

I'm not familiar with pg_search, so I don't know if there's anything that needs to happen on the DB side for this – that said, tested and working on my machine.


sampoder (approved):
LGTM! Tested locally and worked like a charm

garyhtou:
Amazing job with finding the bug and then fixing it!

24c02:
ty! sam deserves most of the credit though, he told me right where to look

PR #6407, opened 2024-05-27
Fix Stripe Cards on Mock Data when not signed in

sampoder:
Transparency mode + mock data was breaking this

Closes https://github.com/hackclub/hcb/issues/6374


app/controllers/events_controller.rb
```diff
@@ -379,6 +379,8 @@ def card_overview
 
     display_cards = if helpers.show_mock_data? && organizer_signed_in?
```
sarthaktexas: why keep this line then?

sampoder: oh because this handles the case when a organiser is signed in

PR #6410, opened 2024-05-27
Prevent impersonation while already impersonating

rluodev:
## Summary of the problem

Admins would be able to impersonate while already impersonating. This would cause the `impersonated_by` field to be updated to the user they were impersonating as, which would result in them only being able to return to the previous user they were impersonating.

## Describe your changes

The impersonation method no longer allows impersonating while impersonating already, and the buttons/links to do so are removed while impersonating.

Screenshot 2024-05-27 at 4 52 49 PM
Screenshot 2024-05-27 at 4 53 20 PM


sampoder:
An alternative fix would be:

```
if impersonate
  user_session.impersonated_by = current_session.impersonated_by || current_user
end
```

But I'm cool with either. Are there any folks impersonating after already impersonating? Who knows..

sampoder (approved):
Code looks good to me! See above comment for an alternative if we want to take a different approach.

RhysMP:
> An alternative fix would be:
> 
> ```
> if impersonate
>   user_session.impersonated_by = current_session.impersonated_by || current_user
> end
> ```
> 
> But I'm cool with either. Are there any folks impersonating after already impersonating? Who knows..

I don't think any one impersonates after impersonating but it does allow privilege escalation.

PR #6411, opened 2024-05-28
Fix the annoying schema change in `activities` table (`t.integer` to `t.bigint`)

rluodev:
## Summary of the problem

It's annoying. (event_id was an integer instead of bigint for some reason even though everywhere else it's bigint)

## Describe your changes

It is now no longer annoying.


sampoder (approved):

garyhtou:
I think we need to revert this PR. The proper way to make this change is by creating a migration — we shouldn't touch the schema file since that's generated

rluodev:
It builds like this in codespaces always, and it’s reflected in the annotations that this field is a bigint

garyhtou:
Ah is the issue here that the schema's outdated? (or not true to the migration?)

rluodev:
I think, yeah. In `event.rb`, the `id` is a `bigint` so it makes no sense that this table is referring to `event_id` as an `integer`

garyhtou:
Got it, that makes sense. Sorry for jumping to conclusions!

PR #6412, opened 2024-05-28
[Google Workspace] Fix verification error mailer error

rluodev:
## Summary of the problem

I think the change of `event_g_suite_overview_url(event_id: @g_suite.event.slug)` to `event_g_suite_overview_url(event: @g_suite.event)` screwed things over since it's looking for a specific field.

## Describe your changes

Reverted the change


garyhtou (approved):
Thanks for fixing!

---


Can you add a Mailer Preview for this?
See https://github.com/hackclub/bank/blob/f86dc88d6a214f2f87c85164d5086d1c0398385d/spec/mailers/previews/g_suite_mailer_preview.rb#L3

PR #6422, opened 2024-05-28
[Google Workspace] DNS record checklist

rluodev:
## Summary of the problem

It's kinda difficult for users to figure out what they're missing - this leads to additional user friction and takes up HCB ops capacity to answer questions.

## Describe your changes

When a verification error occurs, a new `Found` column will appear at the start of the table that has `✅` or `❌` based on if HCB can find the record. This will help users figure out wtf is going on with their dns

Screenshot 2024-06-12 at 1 39 46 PM

Screenshot 2024-06-12 at 1 39 11 PM


sarthaktexas (dismissed):

app/controllers/events_controller.rb
```diff
@@ -431,6 +431,40 @@ def g_suite_overview
 
     @g_suite = @event.g_suites.first
     @waitlist_form_submitted = GWaitlistTable.all(filter: "{OrgID} = '#{@event.id}'").any? unless Flipper.enabled?(:google_workspace, @event)
+
+    @result = [false, false, false, false, false, false, false]
```
sampoder: Instead of an array, could we use a hash?

rluodev:
# PLEASE DO NOT MERGE

gary wants this to look more like a checklist

app/controllers/events_controller.rb
rluodev: i kinda just did it like this for simplicity cause I can just take the index of the array

sampoder: To someone who didn't write the code, it's really confusing when it's presented like this

rluodev: hmm true, I'll change this to a hash

rluodev: oh wait no nvm i just remembered i also did it this way to make it simpler for the mx record checking, cause this way I can just take the mx record index and add two to it to change the result array index to true if found

sampoder: This either needs to be commented and explained clearly or implemented in a cleaner way. I'd strongly prefer for it to be a hash.

app/controllers/events_controller.rb
```diff
@@ -432,6 +432,40 @@ def g_suite_overview
...
```
```suggestion
    # this is janky and should be fixed at some point!
    @result = [false, false, false, false, false, false, false]
```

app/views/events/g_suite_overview.html.erb
```diff
@@ -69,6 +69,9 @@
               <table summary="Server addresses for use in DNS records configured for Google Workspace">
                 <thead>
                 <tr class="bold">
+                  <% if @g_suite.verification_error? %>
+                    <th>Found</th>
```
```suggestion
                    <th></th>
```

garyhtou (commented):
🪄 good to merge after adding a stikethrough on completed rows

rluodev:
@garyhtou done!

app/assets/stylesheets/_utilities.scss
```diff
@@ -207,3 +207,20 @@ $shades: (
...
+}
+
+tr.strikethrough td:before {
+  content: " ";
+  position: absolute;
+  left: 0;
+  top: 50%;
+  width: 100%;
+  border-bottom: 1px solid #8492a6;
+}
```
garyhtou:
```suggestion
}

```

PR #6442, opened 2024-05-29
Updates "Event" to "Organization" on /my/reimbursement

dwahbe:
Quick change to be consistent about how we refer to organizations.

Fixes #6441.


sampoder (approved):
Awesome, LGTM

PR #6443, opened 2024-05-29
Install console1984 and audit1984

manuthecoder:
See #6352


sampoder (dismissed):
PR seems perfect to me; thanks for taking this on!

config/routes.rb
```diff
@@ -7,6 +7,7 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see https://guides.rubyonrails.org/routing.html
 
+  mount Audits1984::Engine => "/console"
```
northeastprince: A base controller needs to be specified or this route needs to be constrained unless we want anyone to be able to access this.

northeastprince (changes requested):

config/routes.rb
garyhtou: Thanks for pointing that out! Let's go with adding a constraint on this mount

Gemfile
```diff
@@ -212,3 +212,6 @@ gem "cssbundling-rails", "~> 1.4"
 gem "sprockets-rails", "~> 3.4"
 
 gem "public_activity"
+
+gem "console1984"
+gem "audits1984"
```
garyhtou:
```suggestion
gem "audits1984"

```

db/schema.rb
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
```
garyhtou: Let's keep this at the top of the file

manu-gurudath:
the last commit is broken!

manu-gurudath:
finished adding admin constraints

garyhtou:
@sampoder if you get a chance, would you mind running migrations and testing locally? 🪄

db/migrate/20240529041318_create_console1984_tables.console1984.rb
```diff
@@ -0,0 +1,36 @@
...
+      t.text :reason
+      t.references :user, null: false, index: false
+      t.timestamps
+
+      t.index :created_at
+      t.index [ :user_id, :created_at ]
+    end
+
+    create_table :console1984_users do |t|
+      t.string :username, null: false
```
garyhtou: IIRC, it will ask for your name when you enter the rails console. console1984 does provide a way to programically set that value (we use it for https://github.com/hackclub/hackathons-backend).

Is there a way to get the name/email of the person who ran `heroku run bin/rails c`? This way we can autofill that

sampoder: I'm going to merge this with the manual name ask for now; @manuthecoder feel free to look into a programatic way if you're interested

PR #6444, opened 2024-05-29
[`Reimbursement::Expense`] Use `approved_by`

sampoder:
This doesn't fully close the issue but it's progress.


garyhtou:
amazing job on making this seamless 👏

PR #6452, opened 2024-05-29
Have `avatar_for`, `user_mention`, `profile_picture` to take a `default_image` prop

sampoder:
Related to https://github.com/hackclub/hcb/issues/6417


app/helpers/users_helper.rb
```diff
@@ -13,11 +13,11 @@ def gravatar_url(email, name, id, size)
     "https://gravatar.com/avatar/#{hex}?s=#{size}&d=https%3A%2F%2Fui-avatars.com%2Fapi%2F/#{CGI.escape(name)}/#{size}/#{get_user_color(id)}/fff"
   end
 
-  def profile_picture_for(user, size = 24)
+  def profile_picture_for(user, size = 24, default_image: "https://cloud-80pd8aqua-hack-club-bot.vercel.app/0image-23.png")
     # profile_picture_for works with OpenStructs (used on the front end when a user isn't registered),
     # so this method shows Gravatars/intials for non-registered and allows showing of uploaded profile pictures for registered users.
     if user.nil?
```
cjdenio: This could be rewritten as
```ruby
  def profile_picture_for
    if user.nil?
      default_image
    elsif # user.is_a user bleh
      Rails.application.routes...
    else
      gravatar_url(..
    end
  end
```
without the `src` variable

cjdenio (dismissed):
Looks great!

app/helpers/users_helper.rb
sampoder: Oh yeah, I wonder why we'd been using the `src` variable

PR #6459, opened 2024-05-29
Redirect back to previous page after session logout

cjdenio:
Closes #6437


app/controllers/users_controller.rb
```diff
@@ -230,17 +230,15 @@ def logout_session
     begin
       session = UserSession.find(params[:id])
       if session.user.id != current_user.id
-        Rail.logger.error "User id: #{user.id} tried to delete session #{session.id}"
```
cjdenio: deleting this line because there are TWO TYPOS

garyhtou: HAHAH

PR #6460, opened 2024-05-29
Add links to the activity log

sampoder:
For now, these are just items with a HCB code. But it will be expanded!


app/views/public_activity/_common.html.erb
```diff
@@ -17,11 +17,17 @@
           <% end %>
           <%= yield %>
         </span>
-        <% unless defined?(hide_timestamp) && hide_timestamp %>
-          <span class="tooltipped tooltipped--n align-middle timestamp" style="cursor: default" aria-label="<%= local_assigns[:activity].created_at.strftime("%b %e, %Y") %>">
+        <div class="flex items-center">
+          <% url = hcb_code_path(activity.trackable.hcb_code) if activity.trackable.try(:hcb_code) %>
```
garyhtou: you might be able to use `polymorphic_path`

garyhtou: https://api.rubyonrails.org/classes/ActionDispatch/Routing/PolymorphicRoutes.html

PR #6469, opened 2024-05-30
Reject `activity.trackable.nil?`

sampoder:
So we can unrevert https://github.com/hackclub/hcb/pull/6466


garyhtou:
Can we make this a database query rather than doing it in Ruby?

PR #6473, opened 2024-05-30
Update `LICENSE` to AGPL

sampoder:
Currently working idea is to go with AGPL. Let's not merge this until we have a legal 👍 and a second check with everyone on the team.


sampoder:
(If anyone has any suggestions on licenses, that'd be awesome. This was inspired by what Supabase / Cal.com use and also was Zach's preferred license. I'm not very opinionated about this.)

northeastprince:
Shouldn't we wait for #6345?

sampoder (commented):
I believe this is read to go.

garyhtou:
file looks good. Let's send out the courtesy emails first

garyhtou:
For posterity, we've got confirmation on AGPL
image

PR #6474, opened 2024-05-30
Subscription detection

manuthecoder:
Closes #3716


app/models/tag.rb
sampoder, line 1: These changes seem to be unrelated

config/routes.rb
```diff
@@ -622,6 +622,7 @@
   get "/event_by_airtable_id/:airtable_id" => "events#by_airtable_id"
   resources :events, except: [:new, :create], path_names: { edit: "settings" }, path: "/" do
     get "edit", to: redirect("/%{event_id}/settings")
+    get "subscriptions"
```
sampoder: I think we should do this on a per-card basis

app/controllers/events_controller.rb
```diff
@@ -215,6 +215,26 @@ def show
...
+
+    in_range = ->(t) { t.date.to_date >= 1.year.ago.to_date && t.amount_cents < 0 }
+    @all_transactions = @all_transactions.select(&in_range)
+    
+    @subscriptions = JSON.pretty_generate(HcbCode::Subscription.new.predict(
+      @all_transactions.map { 
+        |t| { hcb_code_id: t.hcb_code, amount_cents: t.amount_cents, date: t.date.to_date, merchant_name: t.stripe_merchant.name }
+      }
+    ))
+
```
sampoder: I would try to pull transactions from the same card

sampoder (commented):
Just a couple of quick comments!

manu-gurudath:
thank you for your feedback @sampoder!

I've made some changes, and you can now see a card's subscriptions by navigating to a card's details and clicking "See card subscriptions"


image


Not completely sure how to select merchant information since I do not have any data to experiment with, I used `t.stripe_merchant.id`. Feel free to edit it and push a commit to this branch. The merchant ID is crucial for grouping subscriptions

`app/controllers/events_controller.rb`
```ruby
    @subscriptions = JSON.pretty_generate(HcbCode::Subscription.new.predict(
      @all_transactions.map { 
        |t| { hcb_code_id: t.hcb_code, amount_cents: t.amount_cents, date: t.date.to_date, merchant_name: t.stripe_merchant.id }
      }
    ))
```

sampoder (changes requested):
Hey - thanks so much for making this PR! I'm struggling to get it to work locally but I wanted to provide some general implementation feedback:

1) I think for the initial version, it should only support subscriptions with the exact same amount. This should significantly reduce the complexity of the code.

2) It seems like we aren't saving these to the database at the moment? I don't think we need a database table for these at the moment, though I like how you'd name spaced this. Instead, I think we should focus on getting a pretty solid query for detecting potential subscriptions.

I wrote this up in SQL:

```sql
-- This first query finds raw stripe transactions where there
-- are multiple with the same merchant and amount on one card.
-- 
-- This gets us most of the way there but doesn't filter out
-- same item being purchased over and over again. For example,
-- if I pay $2.85 for a can of Celsius at the petrol station
-- every few days. That isn't a "subscription".

WITH potential_subscriptions AS (
    SELECT 
        raw_stripe_transactions.stripe_transaction->'merchant_data'->>'name' as merchant, 
        raw_stripe_transactions.stripe_transaction->>'card' as card,
        canonical_transactions.amount_cents as amount_cents,
        COUNT(*) as transaction_count,
        array_agg(raw_stripe_transactions.date_posted ORDER BY raw_stripe_transactions.date_posted DESC) as dates
    FROM canonical_transactions
    LEFT JOIN raw_stripe_transactions 
        ON raw_stripe_transactions.id = canonical_transactions.transaction_source_id 
    WHERE transaction_source_type = 'RawStripeTransaction'
    AND raw_stripe_transactions.date_posted >= NOW() - INTERVAL '6 months'
    GROUP BY merchant, card, canonical_transactions.amount_cents
),

-- This second query adds a column with an array of distances (in days) 
-- between the dates when transactions occured. My previous example,
-- would have a date_differences a bit like [2, 5, 2, 1, 3, 9].
-- That's a lot of Celsius!

potential_subscriptions_with_date_differences AS (
    SELECT 
        merchant, 
        card, 
        amount_cents, 
        transaction_count,
        dates,
        array(
            SELECT 
                (dates[i] - dates[i + 1])
            FROM generate_series(1, array_length(dates, 1) - 1) AS i
        ) as date_differences
    FROM potential_subscriptions
)

-- This last little bit calculates the standard deviation of the 
-- gaps between transactions and then filters things unlikely to be
-- subscriptions.

SELECT 
    merchant, 
    card, 
    amount_cents, 
    transaction_count,
    dates,
    stddev(unnested_date_differences)
FROM potential_subscriptions_with_date_differences,
LATERAL unnest(date_differences) as unnested_date_differences
WHERE transaction_count > 3
GROUP BY merchant, card, amount_cents, transaction_count, dates
-- having the 5 day gap requirement ensures that if someone
-- buys one thing a bunch on one day (eg. hotel rooms)
-- we don't call it a subscription
HAVING stddev(unnested_date_differences) < 5 AND avg(unnested_date_differences) > 5
ORDER BY stddev(unnested_date_differences) DESC
```

And it works pretty well imo! I can DM you the results if you'd like.

Last thing, let's move this out of the EventsController and into the StripeCardsController. That feels like the right place for it.

sampoder (changes requested):
Hey - thanks so much for making this PR! I'm struggling to get it to work locally but I wanted to provide some general implementation feedback:

1) I think for the initial version, it should only support subscriptions with the exact same amount. This should significantly reduce the complexity of the code.

2) It seems like we aren't saving these to the database at the moment? I don't think we need a database table for these at the moment, though I like how you'd name spaced this. Instead, I think we should focus on getting a pretty solid query for detecting potential subscriptions.

I wrote this up in SQL:

```sql
-- This first query finds raw stripe transactions where there
-- are multiple with the same merchant and amount on one card.
-- 
-- This gets us most of the way there but doesn't filter out
-- same item being purchased over and over again. For example,
-- if I pay $2.85 for a can of Celsius at the petrol station
-- every few days. That isn't a "subscription".

WITH potential_subscriptions AS (
    SELECT 
        raw_stripe_transactions.stripe_transaction->'merchant_data'->>'name' as merchant, 
        raw_stripe_transactions.stripe_transaction->>'card' as card,
        canonical_transactions.amount_cents as amount_cents,
        COUNT(*) as transaction_count,
        array_agg(raw_stripe_transactions.date_posted ORDER BY raw_stripe_transactions.date_posted DESC) as dates
    FROM canonical_transactions
    LEFT JOIN raw_stripe_transactions 
        ON raw_stripe_transactions.id = canonical_transactions.transaction_source_id 
    WHERE transaction_source_type = 'RawStripeTransaction'
    AND raw_stripe_transactions.date_posted >= NOW() - INTERVAL '6 months'
    GROUP BY merchant, card, canonical_transactions.amount_cents
),

-- This second query adds a column with an array of distances (in days) 
-- between the dates when transactions occured. My previous example,
-- would have a date_differences a bit like [2, 5, 2, 1, 3, 9].
-- That's a lot of Celsius!

potential_subscriptions_with_date_differences AS (
    SELECT 
        merchant, 
        card, 
        amount_cents, 
        transaction_count,
        dates,
        array(
            SELECT 
                (dates[i] - dates[i + 1])
            FROM generate_series(1, array_length(dates, 1) - 1) AS i
        ) as date_differences
    FROM potential_subscriptions
)

-- This last little bit calculates the standard deviation of the 
-- gaps between transactions and then filters things unlikely to be
-- subscriptions.

SELECT 
    merchant, 
    card, 
    amount_cents, 
    transaction_count,
    dates,
    stddev(unnested_date_differences)
FROM potential_subscriptions_with_date_differences,
LATERAL unnest(date_differences) as unnested_date_differences
WHERE transaction_count > 3
GROUP BY merchant, card, amount_cents, transaction_count, dates
-- having the 5 day gap requirement ensures that if someone
-- buys one thing a bunch on one day (eg. hotel rooms)
-- we don't call it a subscription
HAVING stddev(unnested_date_differences) < 5 AND avg(unnested_date_differences) > 5
ORDER BY stddev(unnested_date_differences) DESC
```

And it works pretty well imo! I can DM you the results if you'd like. We can adapt it or use it directly if we like.

Last thing, let's move this out of the EventsController and into the StripeCardsController. That feels like the right place for it.

sampoder:
Can we drop the model and build out a UI for this?

manu-gurudath:
I've created a UI and dropped the model for the subscriptions for now. However, we'll need this in the future if we want to allow users to override subscriptions and flag them (for example, the scenario where you buy Celcius every day). This could be handy with the UI I'm building now

sampoder:
Just as an fyi for anyone, the first version of this will be a badge on transactions on the card page:

Screenshot 2024-08-06 at 12 17 59 PM

sampoder (approved):
Thanks for the great work on this @manuthecoder! LGTM.

PR #6483, opened 2024-05-31
Add a users online flavour text

sampoder:
We now know this due to `last_seen_at`


app/services/flavor_text_service.rb
```diff
@@ -495,6 +495,7 @@ def flavor_texts
       "BOOOOOOOOOONNNNNNKKKKKKKKKKKKK",
       "Wanna&nbsp;<a href='#{Rails.configuration.constants.hack_on_hcb_form_url}' target='_blank' style='color: inherit'>hack on hcb</a>?".html_safe,
       "everyone's favorite money thing!",
+      "#{UserSession.where("last_seen_at > ?", 15.minutes.ago).count} online"
```
garyhtou: To prevent making this sql query run even when this line is not picked, i'd suggest the following

```ruby
flavor_texts =  [
  "something funny",
  "this is funny",
  -> { "#{UserSession.where("last_seen_at > ?", 15.minutes.ago).count("DISTINCT(user_id)")} online" },
  "also hilarious"
]

the_chosen_one = flavor_texts.sample
the_chosen_one = the_chosen_one.call if the_chosen_one.respond_to? :call
```

garyhtou: Use `count("DISTINCT(user_id)")` since a user may theoretically use multiple sessions in the past 15 mins

PR #6486, opened 2024-05-31
[`profile_picture_for`] Provide stronger default value

garyhtou:
## Summary of the problem

A parameter with a default value behaves differently than an `||=` (or-equals). Here's a quick demo

```ruby
def foo(default: "hello")
  puts default
end

my_site
# it prints "hello"

my_site(default: "bar")
# it prints "bar"

my_site(default: nil)
# it prints an empty line (nil)
```

In other methods, such as `avatar_for`, we explicitly pass `nil` to `profile_picture_for`. This causes `avatar_for` to run `image_tag(nil)`; which raises an ArgumentError.

## Describe your changes

The solution here is to use an `||=` which will override `nil` (and `false`) values.


garyhtou:
Source of bug: https://github.com/hackclub/hcb/pull/6452

PR #6502, opened 2024-06-01
Add `max-width` to usage data settings

manuthecoder:
See #6480


sampoder (commented):
I think we should set this on `field--options`, thoughts @garyhtou?

garyhtou:
> I think we should set this on `field--options`, thoughts @garyhtou?

I'll leave that decision up to you and Manu!

app/views/users/edit.html.erb
```diff
@@ -171,7 +171,7 @@
             Off by default.
           <% end %>
         </p>
-        <div class="field field--options">
+        <div class="field field--options" style="max-width: 420px">
```
sarthaktexas: why are we using pixels here?

sarthaktexas: we can keep it consistent by using more responsive viewport units

sarthaktexas: also, i agree with sam that this should be added to `field--options` instead, that would be a cleaner solution

sarthaktexas (dismissed):

app/views/users/edit.html.erb
manu-gurudath: I kept it the same as the previous section - I could change both of the section's widths (see line 144)

sampoder (approved):
Cool! Lets do it.

PR #6516, opened 2024-06-03
Fix "User was declined" in activity log

cjdenio:
Screenshot 2024-06-03 at 2 29 45 PM
Screenshot 2024-06-03 at 2 29 52 PM


sampoder (approved):
Thank you!

garyhtou:
i just wanna drop this gem here
image

PR #6520, opened 2024-06-03
Use seed database for staging instead of prod dump

albertchae:
## Summary of the problem

Per discussion with Gary, try using seed data for review apps instead of the prod dump.

## Describe your changes

Script now runs db:seed instead of pg_restore from prod dump


app.json
```diff
@@ -1,12 +1,9 @@
 {
   "name": "bank",
   "scripts": {
-    "postdeploy": "pg_dump --data-only -Fc $HEROKU_POSTGRESQL_JADE_URL | pg_restore --clean --if-exists --no-acl --no-owner -d $DATABASE_URL; bundle exec rails db:migrate"
+    "postdeploy": "bundle exec rake db:migrate db:seed"
```
garyhtou: is rake preferred over rails?

albertchae: force of habit from when it actually mattered (<= Rails 4?), but now it doesn't https://github.com/rails/rails/issues/18878

Updated it to `rails`

garyhtou (approved):

PR #6522, opened 2024-06-03
Make the activity log "live" and real-time

sampoder:
Untitled drawing


app/javascript/controllers/timestamp_controller.js
```diff
@@ -0,0 +1,45 @@
+import { Controller } from '@hotwired/stimulus'
```
sampoder: this controller is so that our timestamps update! that way you don't have a bunch of things "less than a minute ago".

app/views/public_activity/_activity.html.erb
```diff
@@ -0,0 +1 @@
+<%= render_activity(activity, locals: { current_user: }) %>
```
sampoder: this is the partial that is prepended to the list of activities

app/views/public_activity/_common.html.erb
```diff
@@ -28,7 +28,7 @@
             <%= inline_icon "link", width: 20, class: "muted ml-[-4px] mr-[2px]" %>
           <% end %>
         <% end %>
-        <span class="align-middle timestamp" style="cursor: default">
+        <span class="align-middle timestamp" style="cursor: default" data-controller="timestamp" data-timestamp-target="label" data-time="<%= activity.created_at.to_i * 1000 %>">
```
sampoder: see comment on javascript timestamp controller

app/views/public_activity/ach_transfer/_create.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= render layout: "/public_activity/common", locals: { activity: } do %>
+<%= render layout: "/public_activity/common", locals: { activity:, current_user: } do %>
```
sampoder: `current_user` isn't set when we render the stream via a broadcast. this system allows us to pass a `current_user` local variable to `_activity.html.erb` which we render in the broadcast.

config/initializers/public_activity.rb
```diff
@@ -1,12 +1,50 @@
...
+        stream,
+        action: :prepend,
+        target: "activities-1",
+        partial: "public_activity/activity",
+        locals: { activity: self, current_user: streams.first } 
+      )
+
+    end
+  }
+
```
sampoder: this is the code that does the broadcasting!

sampoder:
So this PR actually doesn't do broadcasting anymore, I want to separate that out. This does the scaffold though.

PR #6524, opened 2024-06-04
Log events accessed by a non-admin by ID

YodaLightsabr:
## Summary of the problem
Events can be accessed by IDs which could have unfavorable side effects including
- Being able to count how many orgs there are
- Index every transparent org (even if directory indexing is off)

See #6523

## Describe your changes
This PR logs instances of organizations being accessed with their ID by non-admins.


garyhtou (approved):

garyhtou:
Check for usages of `Event.friendly.find` and switch them over to use `set_event`. Otherwise, we don't get logs for those endpoints!

PR #6529, opened 2024-06-04
Use a secondary Redis instance for caching

sampoder:
Before merging this we need to rename the attachment from `HEROKU_REDIS_OLIVE` to `REDIS_CACHE`: https://dashboard.heroku.com/apps/bank-hackclub/resources.

I've reached out to Heroku because I can't figure it out for the life of me https://help.heroku.com/sharing/97a4002d-feb6-4fd0-a4ff-779b08a3c3d5

Closes https://github.com/hackclub/hcb/issues/3676


sampoder:
Figured it out!

garyhtou:
i'm curious. what is the process for renaming?

config/initializers/geocoder.rb
```diff
@@ -8,7 +8,7 @@
     api_key: Rails.application.credentials.ipinfo_io[:api_key],
   },
 
-  cache: Redis.new(url: ENV["REDIS_URL"], ssl_params: { verify_mode: OpenSSL::SSL::VERIFY_NONE }),
+  cache: Redis.new(url: ENV["REDIS_CACHE_URL"] || ENV["REDIS_URL"], ssl_params: { verify_mode: OpenSSL::SSL::VERIFY_NONE }),
```
garyhtou: Should we drop the `ENV["REDIS_URL"]` here? I don't want cache data filling up the redis we're using for jobs

garyhtou (commented):
Do we need to update docker-compose?

config/initializers/geocoder.rb
sampoder: I did this so that it still worked in development

sampoder:
I feel like we're good to run one Redis instance in development? Do we have Sidekiq on in development?

config/initializers/geocoder.rb
garyhtou: Please add a conditional then!

`Rails.env.development?`

garyhtou (commented):
🪄

PR #6559, opened 2024-06-06
[Reimbursements] Inputs take up full width of container

sarthaktexas:
## Summary of the problem

look at #6553

## Describe your changes
This makes inputs take up the full width on the modal to create new reimbursements. **This should be a discussion.** Should we do this for all modals? Should we even do this?

Screenshot 2024-06-05 at 7 08 54 PM


sampoder (approved):
LGTM!

PR #6576, opened 2024-06-06
Add validation for owner on `PublicActivity::Activity`

sampoder:
No more half validations! Closes https://github.com/hackclub/hcb/issues/6484


garyhtou:
Great job doing that half validation!

PR #6581, opened 2024-06-06
Hide admin-only comments from non-admins in the log

sampoder:
Otherwise it's kind of awkward!


app/views/public_activity/comment/_create.html.erb
```diff
@@ -1,3 +1,5 @@
-<%= render layout: "/public_activity/common", locals: { activity:, url: polymorphic_url(activity.trackable.commentable, anchor: "comment-#{activity.trackable.hashid}") } do %>
-  <%= activity.trackable.action_text %> on <%= link_to activity.trackable.commentable.try(:name) || activity.trackable.commentable.try(:memo) || "an item", activity.trackable.commentable %>
+<% unless activity.trackable.admin_only? && !current_user.admin? %>
```
garyhtou: Is it possible to replace this with a pundit policy check?

sampoder: good idea! yeah totally

PR #6585, opened 2024-06-06
Fix reimbursement layout shift

manuthecoder:
See #5595


sampoder (approved):
LGTM; thanks for improving this!

sampoder:
I don't think this closes the initial issue, maybe @sarthaktexas can help clarify but it is an improvement.

PR #6594, opened 2024-06-07
Change select height

manuthecoder:
See #6593


sampoder (approved):
Thanks for catching this!

PR #6596, opened 2024-06-07
Show view label

manuthecoder:
See #6595


sampoder:
Alternative idea, what if we made it so it was just a button to the other view? So if you're on List it's a button saying "Grid view" and if you are on Grid is a button saying "List view"

manu-gurudath:
changed!

sampoder (dismissed):
This is great; thank you so much!

PR #6601, opened 2024-06-07
When deleting expenses, change the payout holding's amount & destroy the associated expense payout

sampoder:
Also ensures we only delete pending expense payouts with a pending payout holding


app/models/reimbursement/expense.rb
```diff
@@ -36,7 +36,7 @@ class Expense < ApplicationRecord
     monetize :amount_cents
     validates :amount_cents, numericality: { greater_than_or_equal_to: 0 }
     attribute :expense_number, :integer
-    has_one :expense_payout
+    has_one :expense_payout, dependent: :destory
```
sampoder:
```suggestion
    has_one :expense_payout, dependent: :destroy
```

app/models/reimbursement/expense_payout.rb
```diff
@@ -53,7 +53,12 @@ class ExpensePayout < ApplicationRecord
     end
 
     before_destroy do
+      unless payout_holding.pending? && expense.pending?
+        self.errors.add(:base, "non-pending expenses or expenses non-pending payout holdings can not be destroyed")
```
sampoder:
```suggestion
        self.errors.add(:base, "non-pending expenses or expenses with non-pending payout holdings can not be destroyed")
```

PR #6608, opened 2024-06-07
[Pundit] Invert skip_authorizations

malted:
Closes https://github.com/hackclub/hcb/issues/6314


app/controllers/donations_controller.rb
```diff
@@ -6,7 +6,7 @@ class DonationsController < ApplicationController
   include SetEvent
   include Rails::Pagination
 
-  skip_after_action :verify_authorized, except: [:start_donation, :make_donation]
+  skip_after_action :verify_authorized, only: [:export, :show, :refund, :qr_code, :finish_donation, :finished]
```
sampoder: Looking at this list, we probably don't want to skip verifying that we've authorized the donation for some of these. Certainly export, show, and refund. Ik we used to but we probs want to stop doing that.

app/controllers/users_controller.rb
```diff
@@ -3,7 +3,30 @@
...
+                                               :exchange_login_code,
+                                               :logout,
+                                               :unimpersonate,
+                                               :receipt_report,
+                                               :edit_featurepreviews,
+                                               :edit_security,
+                                               :edit_admin,
+                                               :toggle_sms_auth,
+                                               :complete_sms_auth_verification,
+                                               :start_sms_auth_verification]
```
sampoder: Similar comment to above, we probably want to start verifying that we authorise some of these.

app/controllers/donations_controller.rb
malted: Agreed. How about we merge this in to stop the problem from getting worse (this PR changes no current behaviour) and then do an audit pass of all the auth skipped methods after?

sampoder: I'd be cool with that

garyhtou (approved):

PR #6611, opened 2024-06-07
[StatsD] Measure online user count

malted:
We use `StatsD` for reporting metrics like how long certain methods take to execute and how many times certain webhooks time out.

This PR closes https://github.com/hackclub/hcb/issues/6532 and reports the number of currently active users (`User.currently_active`) every five minutes so we can see it on Grafana.


sampoder (approved):
LGTM, nice

PR #6626, opened 2024-06-10
[Reimbursements] Add a tooltip to describe what the statuses mean for a reimbursement report

sampoder:
Screenshot 2024-06-10 at 9 13 35 AM

I only did the two review stages but we can add more if there are other confusing ones!

Closes https://github.com/hackclub/hcb/issues/6372


app/models/reimbursement/report.rb
```diff
@@ -164,6 +164,13 @@ def status_color
       return "primary"
     end
 
+    def status_description
+      return "#{event.name} must review this report" if submitted?
```
sampoder:
```suggestion
      return "Review requested from #{event.name}" if submitted?
```

PR #6633, opened 2024-06-10
[Ruby] upgrade to 3.2.0

garyhtou:
[Ruby 3.2.0](https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/) brings us YJIT for improved performance and WebAssembly (which we'll totally use 😉).

After merging, we'll need to keep a close eye on memory usage in production!


northeastprince:
Why would memory usage be affected if YJIT isn't enabled?

cjdenio (approved):

cjdenio:
To clarify: YJIT isn't enabled yet and we're planning to enable it separately

PR #6635, opened 2024-06-10
Render `event.balance_available_v2_cents` in flash error

sampoder:


app/models/increase_check.rb
```diff
@@ -101,7 +101,7 @@ class IncreaseCheck < ApplicationRecord
 
   validate on: :create do
     if amount > event.balance_available_v2_cents
-      errors.add(:base, "You don't have enough money to send this transfer! Your balance is #{ApplicationController.helpers.render_money(amount_cents)}.")
+      errors.add(:base, "You don't have enough money to send this transfer! Your balance is #{ApplicationController.helpers.render_money(event.balance_available_v2_cents)}.")
```
cjdenio:
```suggestion
      errors.add(:amount, "You don't have enough money to send this transfer! Your balance is #{ApplicationController.helpers.render_money(event.balance_available_v2_cents)}.")
```

needs to be changed back otherwise the error doesn't show up

cjdenio (changes requested):

PR #6640, opened 2024-06-11
Use production `VERIFY_SERVICE_ID` when `USE_PROD_CREDS` is true

sampoder:


app/services/twilio_verification_service.rb
```diff
@@ -9,7 +9,7 @@ class TwilioVerificationService
   )
 
   # This isn't private/sensitive so it's okay to keep here
-  VERIFY_SERVICE_ID = Rails.env.production? ? "VAa06a66dad4c1ca3c199a46334ff11945" : "VAe30d49e92f634419aacdc8648948dc75"
+  VERIFY_SERVICE_ID = Rails.env.production? || ENV["USE_PROD_CREDENTIALS"]&.downcase == "true" ? "VAa06a66dad4c1ca3c199a46334ff11945" : "VAe30d49e92f634419aacdc8648948dc75"
```
garyhtou: I'd recommend moving this into credentials!

PR #6653, opened 2024-06-11
Replace `docs` icons

malted:
From `docs` to `thumbsdown`
image

From `docs` to `plus`
image

From `docs` to `plus`
image


sampoder (commented):
Lets do it!

sampoder (approved):

PR #6663, opened 2024-06-11
Update Codespaces to Ruby 3.2.4 and Node 22

rluodev:
## Summary of the problem

Codespaces were broken

## Describe your changes

Codespaces not broken


garyhtou (commented):
oh, that's a good idea

sampoder (approved):
Thanks Ruien!

PR #6690, opened 2024-06-12
Add tag colors

manuthecoder:
See #6638 

> [!IMPORTANT]  
> Run `yarn install` to re-install the packages as it includes on new dependenceis

Now, you can add tags & colors to emojis through a new modal.
image
image
image
image


sampoder (changes requested):
A couple of points of feedback. I think we should drop the emojis, people can already add them in text (see hcb.hackclub.com/hq) and it will reduce the complexity. Plus if they're in plain text, it reduces the amount of images we need to load. Secondly, I think the colours are too sharp at the moment and they don't fit in well with the rest of the colours on the ledger. I'm not sure what to do about that at the moment, let me think about it over the weekend. Lastly we should probably use a blue button on the modal and allow for no colour or some sort of option that is no colour.

manu-gurudath:
I've removed the emojis feature and reduced the intensity of the colors. Let me know what you think!

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -54,6 +54,34 @@
               <% end %>
             </span>
           <% end %>
+
+          <% if !defined?(receipt_upload_button) && !defined?(hide_tags) && organizer_signed_in? && !pt.instance_of?(OpenStruct) && Flipper.enabled?(:transaction_tags_2022_07_29, @event || pt.local_hcb_code.event) %>
+            <div class="overflow-visible relative" style="margin-left: 0.5rem;" data-controller="menu" data-menu-append-to-value="turbo-frame#ledger">
+              <button class="list-badge add-tag-badge ml0 menu__toggle menu__toggle--arrowless" data-menu-target="toggle" data-action="menu#toggle click@document->menu#close keydown@document->menu#keydown">+ Add tag</button>
+              <%= render "hcb_codes/create_tag", code: pt.local_hcb_code, instance: instance %>
```
cjdenio: i'd like to avoid rendering this modal once for every transaction on the page

.ruby-version
sampoder, line 1: Can we discard these changes?

app/assets/stylesheets/application.scss
```diff
@@ -459,4 +460,4 @@ summary {
       0 1px 2px rgba(0, 0, 0, 0.25),
       0 2px 4px rgba(0, 0, 0, 0.125);
   }
-}
+}
```
sampoder:
```suggestion
}

```

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -215,4 +216,4 @@
       </div>
     <% end %>
   </section>
-<% end %>
+<% end %>
```
sampoder: Can we keep the new lines at the bottom of these files?

app/controllers/tags_controller.rb
```diff
@@ -8,7 +8,7 @@ class TagsController < ApplicationController
   def create
     authorize @event, policy_class: TagPolicy
 
-    tag = @event.tags.find_or_create_by(label: params[:label].strip)
+    tag = @event.tags.find_or_create_by(label: params[:label].strip, color: params[:color])
```
sampoder: Look into using `create_with`: https://apidock.com/rails/v5.2.3/ActiveRecord/QueryMethods/create_with

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -54,6 +54,34 @@
               <% end %>
             </span>
           <% end %>
+
+          <% if !defined?(receipt_upload_button) && !defined?(hide_tags) && organizer_signed_in? && !pt.instance_of?(OpenStruct) && Flipper.enabled?(:transaction_tags_2022_07_29, @event || pt.local_hcb_code.event) %>
```
sampoder: Any reason these were moved?

sampoder (dismissed):
Here's some more general design feedback:

It's hard to see what I've selected:

Screenshot 2024-06-24 at 1 12 45 PM

Let's make the text black here:

Screenshot 2024-06-24 at 1 17 03 PM

Is there a way to design this so that we only render the circle for tags with colours without making the alignment look weird? Maybe we could put them on the right?

Screenshot 2024-06-24 at 1 20 07 PM

manu-gurudath:
Thanks for the feedback! Please let me know if you have any more suggestions

garyhtou:
I'm cool with just doing colors for now. I will say that having icons as a first-class config is nice because it would allow us to collapse tags down to just their emoji and color. On hover, the badge could expand to show the label.

sampoder:
Just took another look, I appreciate the clearer toggles. How hard would it be to make the circle a checkmark? When clicking "Create" on the modal, it reloads the whole page. Would it be possible to change that? I think enabling Turbo on the form should make that possible.

manu-gurudath:
fixed!

app/views/hcb_codes/_create_tag.html.erb
```diff
@@ -0,0 +1,28 @@
...
+          colors.each do |color|
+        %>
+        <label class="tags__radio">
+          <%= form.radio_button :color, color, checked: color === "muted" %>
+          <div class="radio__control tag-darker tag-<%= color %>"></div>
+        </label>
+        <% end %>
+      </div>
+    </div>
+    <%= form.submit "Create", class: "btn bg-primary mt-2 float-right" %>
```
garyhtou: we tend to use blue for these types of buttons

```suggestion
    <%= form.submit "Create", class: "btn bg-info mt-2 float-right" %>
```

app/views/hcb_codes/_create_tag.html.erb
```diff
@@ -0,0 +1,28 @@
...
+<% else %>
+<section class="modal modal--scroll max-w-md bg-snow" data-behavior="modal" role="dialog" id="create_tag">
+  <%= modal_header("Create tag") %>
+  <%= form_with url: event_tags_path(@event), id: "create_tag_form", data: { turbo: true } do |form| %>
+    <%= form.hidden_field :hcb_code_id, value: "", id: "create_tag_hcb_code_id" %>
+    <div class="flex flex-col gap-3 mb-2">
+      <%= form.text_field :label, autofocus: true, placeholder: "Tag name", style: "max-width: 100%", autocomplete: "off", data: { turbo: true }, required: true %>
+      <div class="flex gap-3">
+        <% 
+          colors = ["muted", "red", "orange", "yellow", "green", "cyan", "blue", "purple"]
```
garyhtou: Please store this as constant in the model.

Then, add a validation in the model so that the `color` column can only contain these values

app/views/hcb_codes/_create_tag.html.erb
```diff
@@ -0,0 +1,28 @@
...
+          colors.each do |color|
+        %>
+        <label class="tags__radio">
+          <%= form.radio_button :color, color, checked: color === "muted" %>
```
garyhtou: I'm a bit confused by the `checked: color === "muted"`.

I'm not sure what the intention is, and it's uncommon to use `===` in ruby

app/assets/stylesheets/components/_tags.scss
```diff
@@ -0,0 +1,81 @@
...
+      &.tag-darker {
+        background-color: saturate(darken($color, 5%), 50%);
+
+        &.tag-muted {
+          background-color: darken($color, 5%)
+        }
+      }
+    }
+  }
+}
```
garyhtou: new line at EOF please

garyhtou (commented):
i love the new modal for creating a tag!

app/views/hcb_codes/_create_tag.html.erb
manu-gurudath: This sets the default color to `muted`

app/assets/stylesheets/components/_transactions.scss
```diff
@@ -183,4 +193,4 @@ html[data-dark='true'] {
 
 .pending_transactions_tags:has(> div) {
   margin-top: 2px;
-}
+}
```
garyhtou:
```suggestion
}

```

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
garyhtou, line 1: Instead of doing a `|| "muted"` fallback in ERB. Let's set the default value for the `color` column to be "muted".

If we're setting a default value, let's also make it `null: false`

garyhtou (dismissed):

garyhtou (dismissed):

app/views/hcb_codes/_create_tag.html.erb
```diff
@@ -0,0 +1,28 @@
+<% if button %>
+   <%= link_to "#", class: "menu__input", data: { behavior: "modal_trigger", modal: "create_tag" }, onclick: "document.getElementById('create_tag_hcb_code_id').value = '#{button}'" do %>
+      <%= inline_icon "plus", size: 15 %>
+      Create tag
+    <% end %>
```
garyhtou:
```suggestion
  <%= link_to "#", class: "menu__input", data: { behavior: "modal_trigger", modal: "create_tag" }, onclick: "document.getElementById('create_tag_hcb_code_id').value = '#{button}'" do %>
    <%= inline_icon "plus", size: 15 %>
      Create tag
  <% end %>
```

garyhtou (dismissed):

garyhtou (approved):

PR #6704, opened 2024-06-13
Fix SimpleCov not recognizing `HandleIssuingAuthorizationRequest` spec

cjdenio:
For some reason, our test coverage for this file was 0% until I added `_spec` to the spec file


garyhtou:
FYI @albertchae

PR #6706, opened 2024-06-13
[Spending controls] Allowances must not cause negative control balances

malted:
image

image


app/models/organizer_position/spending/control/allowance.rb
```diff
@@ -33,6 +33,14 @@ class Allowance < ApplicationRecord
         has_one :organizer_position, through: :control
         has_one :event, through: :organizer_position
 
+        validate :balance_positive
+
+        private
+
+        def balance_positive
+          errors.add(:control, "balance must be positive") if control.balance_cents + amount_cents < 0
+        end
```
sampoder:
```suggestion
        validate :balance_is_positive

        private

        def balance_is_positive
          errors.add(:control, "balance must be positive") if control.balance_cents + amount_cents < 0
        end
```
Also validate on create?

sampoder (approved):

PR #6714, opened 2024-06-13
Add rich session data

manuthecoder:
See #6712


sampoder (changes requested):
Can we use `maps.hackclub.com`? Search the codebase for examples of where we've used it.

sampoder (changes requested):
Thanks for that! A couple of things, geolocating for city is going to burn through all of our Geolocated API credits. Every time we create a user session, we save its longitude and latitude, so you can do `session.longitude` and `session.latitude`.

Secondly, for sessions that don't have an IP it makes the page look very imbalanced as there is a massive map on some sessions and nothing on others. Any thoughts on how we could fix that? Maybe there's a way we can design it in a way that the height doesn't change whether or not you have a map? Or we can render the entire world if we don't know? I'm not sure.

manu-gurudath:
>...geolocating for city is going to burn through all of our Geolocated API credits...

What if HCB used [findip.net's](findip.net) API? It's free and unlimited. 

>  ... for sessions that don't have an IP it makes the page look very imbalanced ... Any thoughts on how we could fix that?

What if we showed an empty square with "No preview available," like how my app currently does:
image

I'm also formatting the map & placeholder with`aspect-ratio: 1 / 1` to ensure it remains the correct dimensions.

app/views/users/_user_session.html.erb
```diff
@@ -22,6 +22,13 @@
       <% end %>
     <% end %>
   </span>
+  <% if session.latitude %>
+    <img class="w-full rounded mb-3 aspect-square" src="https://maps.hackclub.com/api/shipping?latitude=<%= session.latitude %>&longitude=<%= session.latitude %>" />
```
sampoder:
```suggestion
    <img class="w-full rounded mb-3 aspect-square" src="https://maps.hackclub.com/api/shipping?latitude=<%= session.latitude %>&longitude=<%= session.longitude %>" />
```

sampoder (approved):
LGTM!

PR #6745, opened 2024-06-15
Add card user filter

manuthecoder:
# Objectives
This PR essentially does two things:  
- Closes #6637 
   - This **removes** the ability for users to search cards by user via text input
- Creates a **_reusable_ stimulus tab component** which revamps the filter UI on the homepage
   - Adds an option for currently selected menu items
   - Adds transition for buttons
 
https://github.com/hackclub/hcb/assets/77016441/04a04ea4-351a-4a62-bea5-a6d9a66cd251

New UI for homepage:
image

# Reusable tab component

The main container (`<div>`) is marked with `data-controller="tabs"` and specifies the default tab via `data-tabs-default-tab-value`. Each tab button (`<button>`) has a unique ID, is marked as a target with `data-tabs-target="btn"`, and has a click event bound to the select method of the "tabs" controller.

Corresponding tab content sections (`<div>`) are also marked as targets with `data-tabs-target="tab"` and have IDs **_matching_** their respective buttons. 

```html
<div data-controller="tabs" data-tabs-default-tab-value="[ID default tab goes here]">
    <button id="[ID of tab]" data-tabs-target="btn" data-action="click->tabs#select">...</button>
    <button id="[ID of tab]" data-tabs-target="btn" data-action="click->tabs#select">...</button>
    ...

    <div data-tabs-target="tab" id="[ID of tab]">...</div>
</div>
```

# UI improvement: Currently active tab
Add the `.menu__options--active` class to an element to show a check mark next to it


sampoder (changes requested):
This is really nice! A couple of quick things:

* we're starting to get a lot of buttons up there, could we make the two blue ones just have icons and no text?
* can you check that the styles for the new filter menu are the same? i see that there's a new blue background, I think I personally prefer the blue text. also the menu divider below the options seems to be missing.

manu-gurudath:
I've gone ahead and turned the text buttons into icon buttons, and ensured that the menu buttons are consistent with the original UI.

sampoder:
The two designs aren't identical yet:

Screenshot 2024-06-23 at 11 31 41 AM

Screenshot 2024-06-23 at 11 33 42 AM

See font weight and divider

sampoder (approved):
Let's get this merged

PR #6764, opened 2024-06-17
Render `not_found` for unused HCB codes

sampoder:
Closes https://github.com/hackclub/hcb/issues/2297. @garyhtou can you sanity check, is this what I meant by "unused"?


garyhtou:
great!

PR #6782, opened 2024-06-18
[Spending controls] Deactivate controls when disabling related feature previews

malted:


app/controllers/features_controller.rb
```diff
@@ -39,8 +39,15 @@ def disable_feature
       if Flipper.disable_actor(feature, actor)
         # If it's the user permissions feature, make all the users & invites in the org managers.
         if feature == "user_permissions_2024_03_09" && actor.is_a?(Event)
```
garyhtou: In a separate PR, rip out this stuff for `user_permissions_2024_03_09` since it was already fully shipped!

garyhtou (approved):

garyhtou:
please test and merge!

app/controllers/features_controller.rb
malted: real

PR #6790, opened 2024-06-18
[Expired sessions] Fix query

malted:


sampoder (dismissed):
lgtm!

app/controllers/users_controller.rb
```diff
@@ -369,10 +369,8 @@ def edit_security
                         .user_sessions
                         .only_deleted
```
garyhtou: I think the `only_deleted` will hide sessions that expired because of time. Maybe we want `with_deleted`

malted: wait this is a great catch

garyhtou (approved):
There's one small bug left: Expired sessions older than 1 week may show up

PR #6798, opened 2024-06-18
Make receipt bin dropzone pastable

YodaLightsabr:
## Summary of the problem
You can't paste into receipt file upload

## Describe your changes
Add pasting to the receipt bin file upload box


garyhtou (dismissed):

app/javascript/controllers/file_drop_controller.js
```diff
@@ -54,6 +55,13 @@ export default class extends Controller {
     this.counter = 0
 
     this.submitting = false
+
+    if (this.pastableValue) {
+      document.body.addEventListener('paste', e => {
```
garyhtou: what happens if there are multiple file drop zones on a page?

YodaLightsabr: This is currently only enabled for the primary receipt bin dropzone. In the future, I'll change it to support multiple.

garyhtou (approved):

PR #6805, opened 2024-06-19
Add command palette trigger

manuthecoder:
See #6804

This PR adds a button to the sidebar which triggers the command bar. This way, it let users know this feature is available and encourage them to use it. 
image


sampoder:
Thanks for making this PR! I really appreciate the effort that's gone into and I think the design is pretty clean. At the moment, though, I feel like the command bar isn't quite ready to have a "Jump to..." button and I don't want to overcomplicate the UI. I'm going to close this as not planned for now, but thank you.

sampoder:
Actually... I have an idea

sampoder (approved):
I've added it so that it's only on user navigation menu. I think it looks good to me!

PR #6826, opened 2024-06-20
Putting dates first while displays transactions + moving column status

manuthecoder:
See #6794 

This PR reorders tables across HCB to ensure a more consistent navigation. 

>  In the Transactions, Donations, Invoices, Check deposit, Transfers, and Reimbursements pages, transactions are ordered by date. However, on all of these pages except Transactions, the date isn't the first column shown.
> 
> I think reordering these so the date is first will make transaction lists consistent and help users quickly navigate across HCB.
> 
> On these pages, the status is the first column, which I think could go last. The order could look like this:
> 
> Donations, recurring donations - Name, Started on, amount, total donated, status
> Donations - Date, From, Amount, Status
> Invoices - Date, To, Amount, Status
> Check deposit - Date, Amount, Status
> Transfers - Date, To, Amount, Status
> Reimbursements - Created, Report, From, Amount, Status
> Curious what y'all think!


sampoder:
I had a look at this running locally, honestly, I think I prefer the old version. I appreciate the effort put in here though. Could you create a post in `#hcb-engr-help` to collect others' thoughts?

dwahbe:
Another thought I had was putting the recipient / donor / receiver first. This could look like:

Donations, recurring donations - Name, Started on, amount, total donated, status
Donations - From, Amount, Status, Date
Invoices - To, Amount, Status, Date
Check deposit - Amount, Status, Date
Transfers - To, Amount, Status, Date
Reimbursements - Report, From, Amount, Status, Created

I think status first isn't ideal for these because typically it's a wall of the same status. I like date because it is consistent with Transactions and it is how we sort them but I also like receiver / donor / submitter. What do you think?

manu-gurudath:
What if we had a feature in HCB which allowed users to sort transactions by date, name, or status. This would be done by clicking on the table's header.

sampoder:
See https://github.com/hackclub/hcb/issues/6794#issuecomment-2237318121

sampoder:
Going to close this out as it looks like we aren't going to go with this, see https://github.com/hackclub/hcb/issues/6794#issuecomment-2237318121. Thanks Manu for working on this!

PR #6827, opened 2024-06-20
[Team] Add table view

manuthecoder:
See #6825

This PR allows users to view their team members in a table format.

image


sampoder (changes requested):
A couple of pieces of feedback:

* Let's make it "Invited at"
* The spacing between the avatar and the name is a little too much and the admin badge doesn't look good:
Screenshot 2024-06-20 at 10 40 26 AM

* the page should be `page_sm` not `page_md` for the grid view. unless we make it two columns.
* everything should be vertically centered
* the pagination should be at the bottom, not the top.

sampoder:
Just going to lint this and merge; let's have it production for a bit and experiment. If this is handy for folks I'd like to keep it. It does add additional complexity to the page though so if it becomes unused we might drop it. But yeah let's merge and see.

sampoder (approved):

PR #6848, opened 2024-06-21
Beautify scrollbars

manuthecoder:
This PR beautifies scrollbars throughout the HCB app. It works on dark mode, and intentionally left unstyled on mobile.  This way, it ensures scrollbars are consistent through different browsers & platforms. 

| Before | After |
|--------|--------|
| image| image | 

I would love to hear your feedback!


sampoder:
This is pretty pretty! That said, I'm not feeling great about introducing more custom CSS. I personally feel like browsers should handle scrollbars but I'd appreciate others' opinions...

cjdenio:
What Sam said— this looks nice, but personally I have scrollbars hidden OS-wide and it would be weird to have a non-native scrollbar show up just on HCB

PR #5295, opened 2024-02-19
Improve suggested pairings and add memo suggestions using GPT-4

YodaLightsabr:
## Summary of the problem
Suggested pairings currently rely on janky regular expressions to extract features from receipts. People assume that suggested pairings are accurate even if there are errors.

## Describe your changes
To fix this, suggested pairings will now use GPT-4 to extract features. The extracted features will also be used for suggesting memos to the user.

image
<small>The pairing suggestion screen inside of Receipt Bin</small>

image
<small>A suggested memo on the HCB code page</small>

PR #6868, opened 2024-06-23
Suggest Tags for Transactions With ✨ AI ✨

sampoder:
Cal Hacks is running a hackathon this weekend and it's AI-themed! I thought I'd jump in on the fun and try building something... as I watched Shrek 2 with everyone. So inspired by https://github.com/hackclub/hcb/pull/5295, I made a system for using GPT-4 to suggest tags for card transactions:

Screenshot 2024-06-23 at 12 12 05 AM

Screenshot 2024-06-23 at 12 12 34 AM

(the suggestions are in purple!)

Here's the current "accepting" the suggestion:

https://github.com/hackclub/hcb/assets/39828164/cc43e405-c827-4d91-886d-f749eed6f98f

And rejecting it by clicking the :x::

https://github.com/hackclub/hcb/assets/39828164/aca3e6f4-f055-4457-ba0d-d1424974fef1

The suggestions show up on hover.

Tooltip is a work in progress:

Screenshot 2024-06-23 at 1 00 09 AM

I ran it on the Outernet account, here were the results (see the purple tags): [Outernet – HCB.pdf](https://github.com/user-attachments/files/15943316/Outernet.HCB.pdf). It's accurate but also pretty conservative at the moment, but we could probably edit the prompt a bit more to change that.

Some open questions at the moment:

1) What does a good UI for this look like? I half like what I've got but it'd be nice if it was clearer that this was a suggestion from HCB.
2) Preloading the associations, at the moment, this slows down the page a little. I need to figure out how to preload the association.
3) When do we run the job? I'm not sure at this point, maybe every time the memo gets updated and on creation?

So yeah! That's that. Thoughts everyone?


app/models/hcb_code/tag/suggestion.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      aasm do
+        state :suggested, initial: true
+        state :accepted
+        state :rejected
+
+        event :mark_accepted do
+          transitions from: :suggested, to: :accepted
+          after do
+            ::HcbCodeTag.create!(tag_id:, hcb_code_id:)
```
garyhtou: Is this idempotent?

app/services/hcb_code_service/generate/suggested_tags.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        id = response.body.dig("choices", 0, "message", "content")&.delete("^0-9")
+
+        if tag = Tag.find_by_id(id)
+          if tag.event == @event
+            HcbCode::Tag::Suggestion.create(tag:, hcb_code: @hcb_code)
+          end
+        else
+          puts nil
+        end
```
garyhtou:
```suggestion
        if tag = Tag.find_by(id: id, event: @event)
          HcbCode::Tag::Suggestion.create(tag:, hcb_code: @hcb_code)
        end
```

garyhtou (commented):
I haven't taken a close look, but this is very cool!

At the moment, the purple tag doesn't make it clear that it's suggested.

app/models/hcb_code/tag/suggestion.rb
sampoder: It doesn't allow duplicates yeah, this will error though if we suggest, someone adds manually, and then someone accepts our suggestion. Good catch!

sampoder:
updated:

Screenshot 2024-06-26 at 3 10 42 PM

sampoder:
oh shoot just realised: what if someone deletes a tag?

sampoder:
the answer: they now get destroyed

app/models/tag.rb
```diff
@@ -23,6 +23,7 @@ class Tag < ApplicationRecord
 
   belongs_to :event
   has_many :hcb_code_tags
+  has_many :hcb_code_tag_suggestions, dependent: :destory
```
sampoder:
```suggestion
  has_many :hcb_code_tag_suggestions, dependent: :destroy
```

sampoder:
I'm going to merge these, unless 30%-ish of these get interacted with, I'm going to revert this and delete the tables / models.

PR #6872, opened 2024-06-24
[Spending controls] Invite users with a provisional allowance

malted:


garyhtou (changes requested):
I would suggest creating a new column on `OrganizerPositionInvite` called `initial_control_allowance_amount_cents`.
- validate it's positive
- allow null. when null, don't create a control.
- create the control and allowance here: https://github.com/hackclub/bank/blob/22bde5ae1192db229691c8f93913e83eb25daa18/app/models/organizer_position_invite.rb#L110

PR #6875, opened 2024-06-24
Include event slug in transaction URL

cjdenio:


garyhtou:
I'm curious about the intentions behind this. If it's related to performance and getting `@event`, then I wonder if the new transaction engine redesign (along with splitting Disbursements into two models) would help solve this

PR #6877, opened 2024-06-24
Stream extracted receipts data via Action Cable

sampoder:
Previously, you'd have to refresh the page to see this extracted data. Now, when it's available, we'll stream it to the user's receipt bin. Just like we do with suggested pairings!


app/models/receipt.rb
```diff
@@ -38,6 +38,8 @@ class Receipt < ApplicationRecord
   include PublicIdentifiable
   set_public_id_prefix :rct
 
+  include Turbo::Broadcastable
```
cjdenio: I'm 99% sure this is included automatically

sampoder: Oh sweet!

app/services/receipt_service/suggest.rb
```diff
@@ -31,7 +31,7 @@ def run!(include_details: false)
 
         if @receipt.receiptable.nil?
           content = turbo_stream_action_tag(:refresh_suggested_pairings)
-          ActionCable.server.broadcast(stream_name_from([@receipt.user, :suggested_pairings]), content)
+          ActionCable.server.broadcast(stream_name_from([@receipt.user, :receipt_bin]), content)
```
cjdenio:
```suggestion
          Turbo::StreamsChannel.broadcast_action_to([@receipt.user, :receipt_bin], action: :refresh_suggested_pairings)
```
This is a little better since it doesn't rely on importing `turbo-rails`'s internal logic

app/models/receipt.rb
sampoder: Can confirm that's the case

app/services/receipt_service/extract.rb
```diff
@@ -92,6 +92,17 @@ def run!
         )
       end
 
+      unless @receipt.receiptable
+        @receipt.broadcast_action_to(
+          [@receipt.user, :receipt_bin],
+          action: :replace,
+          target: nil,
```
cjdenio:
```suggestion
        @receipt.broadcast_replace_to(
          [@receipt.user, :receipt_bin],
```
super nitpicky

app/services/receipt_service/suggest.rb
sampoder: Oh sweet! I remember earlier a similar line wasn't working for us but this works

app/services/receipt_service/extract.rb
sampoder: I think we want to keep `target: nil` otherwise this will target the receipt:

```suggestion
        @receipt. broadcast_replace_to(
          [@receipt.user, :receipt_bin],
          target: nil,
```

app/services/receipt_service/extract.rb
```diff
@@ -92,6 +92,16 @@ def run!
         )
       end
 
+      unless @receipt.receiptable
+        @receipt. broadcast_replace_to(
```
sampoder:
```suggestion
        @receipt.broadcast_replace_to(
```

PR #6884, opened 2024-06-24
[Spending controls] Create invite with initial allowance amount

malted:
image


garyhtou:
@sampoder can you review this one?

app/models/organizer_position_invite.rb
```diff
@@ -111,7 +112,17 @@ def accept(show_onboarding: true)
 
     self.accepted_at = Time.current
 
-    self.save
+    ActiveRecord::Base.transaction do
+      self.save
+
+      unless initial_control_allowance_amount_cents.nil?
```
sampoder:
```suggestion
      if initial_control_allowance_amount_cents.present?
```

app/models/organizer_position_invite.rb
```diff
@@ -111,7 +112,17 @@ def accept(show_onboarding: true)
...
+        # Create control
+        organizer_position.spending_controls.create!
+
+        # Create allowance
+        organizer_position.active_spending_control.allowances.create!(authorized_by_id: sender_id, amount_cents: initial_control_allowance_amount_cents, memo: "Initial control")
```
sampoder:
```suggestion
        organizer_position.active_spending_control.allowances.create!(authorized_by_id: sender_id, amount_cents: initial_control_allowance_amount_cents, memo: "Initial allowance")
```

app/controllers/organizer_position_invites_controller.rb
```diff
@@ -21,7 +21,10 @@ def create
     role = invite_params[:role]
     is_signee = invite_params[:is_signee] || false
 
-    service = OrganizerPositionInviteService::Create.new(event: @event, sender: current_user, user_email:, is_signee:, role:)
+    enable_spending_controls = (invite_params[:enable_controls] == "true") && (role != "manager")
```
sampoder: Would it make sense to error if someone attempts to enable spending controls for a manager?

app/services/organizer_position_invite_service/create.rb
```diff
@@ -2,20 +2,23 @@
...
+      @enable_spending_controls = enable_spending_controls
+      @initial_control_allowance_amount = initial_control_allowance_amount
 
       args = {}
       args[:event] = @event
       args[:sender] = @sender
       args[:initial] = @initial
       args[:is_signee] = @is_signee
       args[:role] = @role if role
+      args[:initial_control_allowance_amount_cents] = @enable_spending_controls ? (@initial_control_allowance_amount.to_f * 100).to_i : nil
```
sampoder: can we use `Monetize.parse`?

sampoder:
Would it make sense to include this in the mailer on invite?

app/controllers/organizer_position_invites_controller.rb
malted: Added validation on opi

PR #6886, opened 2024-06-25
Remove unrelated invoices tooltip from `/breakdown`

sampoder:
It just doesn't make sense?


garyhtou (approved):
go to sleep sam!

sampoder:
(we're in the same timezone :p)

PR #6888, opened 2024-06-25
Add caching to `/balance_by_date`

sampoder:
This is to help cool down the amount of SQL queries needed for this feature. So it only gets hit once every 5 minutes or so. We're always overriding the final balance so that that is realtime updating.


app/controllers/events_controller.rb
```diff
@@ -232,16 +232,15 @@ def breakdown
...
-    rescue Pundit::NotAuthorizedError
-      render json: { error: "We couldn’t find that organization!" }
-      return
-    end
+    authorize @event
 
     max = [365, (Date.today - @event.created_at.to_date).to_i + 5].min
 
-    balance_by_date = ::TransactionGroupingEngine::Transaction::All.new(event_id: @event.id).running_balance_by_date
+    balance_by_date = Rails.cache.fetch("balance_by_date_#{@event.id}", expires_in: 5.minutes) do
```
garyhtou: is there a convention for the cache key that we should be following?

sampoder: Not that I know of? If there is one, I'd be happy to adopt it

PR #6889, opened 2024-06-25
Bump Ruby version to 3.2.4

garyhtou:
there's an issue with docker that I haven't been able to figure out!


garyhtou:
If anyone has some time, feel free to debug this

cjdenio:
@garyhtou can you share the specific issue you're running into?

cjdenio:
```
docker volume rm bank_node_modules bank_tmp bank_gems

docker build web --no-cache
```

might fix it if it's the same issue I ran into

garyhtou:
ah interesting, okay!

garyhtou:
@rluodev would you mind testing this? I remember you ran into the same issue

garyhtou:
In the meantime, @cjdenio can you sanity check this pr and merge so that we can get the security patches in prod?

cjdenio (approved):

PR #6891, opened 2024-06-25
Rollout running balances

sampoder:
To be merged after testing the new query in the production console, this should bring back the graphs!


sampoder:
Near instant on HQ. LGTM.

PR #6895, opened 2024-06-25
Small redesign of the Google Workspace waitlist page

sampoder:
Screenshot 2024-06-25 at 9 19 21 AM


sampoder:
(the <hr /> is very faint in the screenshot, i feel like it looks good locally but if anyone has different preferences, let me know)

PR #5817, opened 2024-03-28
💳 Add personalization design functionality to Stripe cards

YodaLightsabr:
## Summary of the problem
All of HCB's Stripe cards are black with the HCB logo on it. That's cool and all, but it would be nice if organizations could have personalized card designs so that they wouldn't confuse their cards.

## Describe your changes
This PR creates the `StripeCard::PersonalizationDesign` model and syncs it with Stripe. It automatically creates personalization designs for events whenever their name changes, and there's a one time job allowing us to backfill designs. Admins also have the ability to create and manage designs.

This has been tested with Test Data in Stripe.

Some screenshots of the UI:

| image | image |
|:--------:|:--------:|
| Design Selection Page | Admin Designs Page |

PR #6896, opened 2024-06-25
Upgrade Stripe to `2019-12-03`

sampoder:
Should help with https://github.com/hackclub/hcb/pull/5817


app/models/invoice.rb
```diff
@@ -261,7 +261,7 @@ def set_fields_from_stripe_invoice(inv = remote_invoice)
     self.auto_advance = inv.auto_advance
     self.due_date = Time.at(inv.due_date).to_datetime # convert from unixtime
     self.ending_balance = inv.ending_balance
-    self.finalized_at = inv.finalized_at
+    self.finalized_at = inv.respond_to?(:status_transitions) ? inv.status_transitions.finalized_at : (inv.try(:finalized_at) || self.finalized_at)
```
```suggestion
    self.finalized_at = inv.respond_to?(:status_transitions) ? inv.status_transitions.finalized_at : inv.try(:finalized_at)
```

PR #6898, opened 2024-06-25
Switch to `polynomially_longer`

sampoder:
DEPRECATION WARNING: `wait: :exponentially_longer` will actually wait polynomially longer and is therefore deprecated. Prefer `wait: :polynomially_longer` to avoid confusion and keep the same behavior. (called from <class:DeliverWebhook> at /home/runner/work/hcb/hcb/app/jobs/partnered_signup_job/deliver_webhook.rb:8)


garyhtou:
lol polynomially_longer is a funny name

PR #6902, opened 2024-06-25
Attempt to silence `fatal: not a git repository` error in logs

cjdenio:


garyhtou (commented):
these annoying logs have finally met their fatal ends

PR #6905, opened 2024-06-25
Allow donors to cover HCB's fees

sampoder:
On both one-time and recurring donations! Closes https://github.com/hackclub/hcb/issues/6903

Screenshot 2024-06-25 at 2 06 49 PM


dwahbe:
This looks great! 

For the copy next to the check box, I'm wondering if people won't know what HCB is. 

It could be something like "Add $ to cover all fees". For the tooltip, it could be "By clicking this box, you are covering fiscal sponsorship and credit card processing fees."

garyhtou (commented):
Can we add a boolean column for this? I'd like to be able to track how many are fee-covered

PR #6908, opened 2024-06-25
Upgrade ruby and node version for codespaces

manuthecoder:
GitHub Codespaces fails to build because the ruby version is outdated. This PR Updates the Ruby version so it builds the container successfully. 

It also upgrades the node version to 22


sampoder:
Thanks Manu for this PR! Ruien made a similar PR: https://github.com/hackclub/hcb/pull/6663. I just merged it. Let me know if that doesn't fix it.

PR #6909, opened 2024-06-25
Card popover for list view

manuthecoder:
Currently, card popovers only show up on grid view. This PR adds card popovers for list view. 

image


app/views/events/_event_cards_table.html.erb
```diff
@@ -43,9 +43,19 @@
           </td>
 
           <td>
-            <%= link_to "Details", stripe_card_path(d) if organizer_signed_in? && !@event.demo_mode %>
+            <%= link_to "Details", stripe_card_path(d), data: { behavior: "modal_trigger", modal: "card_popover_#{d.id}" } if organizer_signed_in? && !@event.demo_mode %>
```
sampoder: Shouldn't this only be if we're showing card popovers?

manu-gurudath: thanks for catching that!

sampoder (dismissed):
LGTM, thanks Manu!

PR #6910, opened 2024-06-25
Redesign see all your cards

manuthecoder:
The current UI for "See all your cards" feels visually out-of-place. This PR makes the UI simpler for the cards page. 

Before:
image

After: 

image


sampoder (approved):
Looks great to me!

PR #6930, opened 2024-06-26
Allow users to delete comments

cjdenio:
https://github.com/hackclub/hcb/assets/34525547/474de297-a1a3-4b91-b35d-86149b6fbf1c


app/views/reimbursement/reports/_conversation.html.erb
```diff
@@ -7,9 +7,7 @@
     <% comments = policy_scope(@report.comments) %>
     <% (versions + comments).sort_by(&:created_at).each do |item| %>
       <% if item.instance_of?(Comment) %>
-        <div class="conversation_item">
-          <%= render "comments/item", comment: item %>
-        </div>
+        <%= render "comments/item", comment: item, class: "conversation_item" %>
```
cjdenio: This is to make sure the whole Turbo Streams thing works properly!

sampoder:
This is sweet! At some point we should also track these in the activity log, I can help with that

cjdenio:
Sounds great, I'll try to add activity log support in a separate PR and get your review!

PR #6934, opened 2024-06-26
Refactor `Comment` routes to use a routing concern

cjdenio:
## Summary of the problem

Writing `resources :comments` under every commentable resource is a pain! Plus, it generates a bunch of unnecessary routes.

## Describe your changes

Now, you can just add `concerns: :commentable` to add comments to a resource.


garyhtou:
lovely!

PR #6935, opened 2024-06-26
Add a `fee_covered` boolean

sampoder:
So we can keep track of this! Gary asked for this.


.ruby-version
```diff
@@ -1 +1 @@
-3.2.4
+3.2.0
```
sampoder:
```suggestion
3.2.4
```

PR #6938, opened 2024-06-26
Receipt Extraction: Handle ChatGPT returning an array of JSON

sampoder:
Looks like ChatGPT sometimes returns an array of JSON: https://heroku-app97991095.airbrake.io/projects/288439/groups/3557109453095140611?tab=overview

Which the JSON parser thinks is valid but we can't use... so this fixes that.


app/services/receipt_service/extract.rb
```diff
@@ -57,7 +57,7 @@ def run!
       end
 
       extracted = begin
-        JSON.parse(ai_response).with_indifferent_access # JSON given by ChatGPT, may fail
+        JSON.parse(ai_response).yield_self { |r| r.is_a?(Array) ? r.first : r }.with_indifferent_access # JSON given by ChatGPT, may fail
```
sampoder:
```suggestion
        JSON.parse(ai_response).yield_self { |r| r.is_a?(Array) ? r.first : r }.with_indifferent_access # JSON given by ChatGPT, may fail. The `yield_self` handles ChatGPT returning an array.
```

sampoder:
Screenshot 2024-06-26 at 3 24 39 PM

PR #6950, opened 2024-06-27
More StatsD

malted:
Closes https://github.com/hackclub/hcb/issues/6620


garyhtou:
During our call, we found a few things that need to be fixed up.

- double check stat naming (especially with gauges/counts since they may already be namespaced by statsd/grafana)
- double check the data you're passing is accurate. for example, not all donations are paid. Also, for Invoice, use the existing unpaid scope.

app/models/donation.rb
```diff
@@ -97,6 +97,8 @@ class Donation < ApplicationRecord
       transitions from: :pending, to: :in_transit
       after do
         create_activity(key: "donation.paid", owner: nil)
+        StatsD.increment("Donation.paid")
```
sampoder: How are we going include pre-existing paid donations?

app/models/stripe_card.rb
```diff
@@ -185,33 +187,41 @@ def state
...
+      save!
+    end
   end
 
   def cancel!
-    StripeService::Issuing::Card.update(self.stripe_id, status: :canceled)
-    sync_from_stripe!
-    save!
-    card_grant.cancel! if card_grant&.active?
+    StatsD.measure("StripeCard.cancel") do
```
sampoder: What's the thought behind measuring these?

sampoder: I guess why specifically `StripeCard`. Sorry I'm just a little out of the loop

sampoder:
(oh btw, this is good to merge when you're ready.)

PR #6977, opened 2024-06-28
Upgrade settings inputs

manuthecoder:
Essentially, this PR has two main purposes: 
* Create a reusable **toggle switch** component which can be used throughout the application
* Redesign specific settings to enhance user experience. In summary, important settings are now displayed as sliders instead of checkboxes. Furthermore, more visual weight is added on those settings as well. 
    * Transparency mode settings: Convert main checkbox to slider, and hide "message to show visitors" section when not needed
    * Donations page
    * Reimbursements page

Screenshots: 
image
image
image


sampoder:
In a similar vein to https://github.com/hackclub/hcb/pull/6848, I'd prefer for these to be handled by the browser. Hiding the message is a good improvement though, we should handle that case.

Re browsers, check out: https://alvaromontoro.com/blog/68049/new-toggle-switch-lands-in-safari and https://webkit.org/blog/15054/an-html-switch-control/.

manu-gurudath:
Thank you for your input. The HTML switch element is a nice addition, but it's currently only available on Safari. To ensure broader compatibility and accessibility, the custom slider input I created might be a better fit. It works across all browsers and maintains accessibility

Also I feel like the sliders are easier to click on

garyhtou:
I love that Safari is adding a switch! I think it looks amazing and I wish other browsers would follow suit.

The slider makes sense in the screenshots you have above; it feels very Apple-like.

As long as the toggle switch component is robust, it may make sense to use it even though it's not a browser default. thoughts @sampoder?

sampoder:
I'm on the fence about it. If we want to do it, I think that we should make it closer to Apple's implementation. Maybe we could copy their styles?

manu-gurudath:
✅

https://github.com/hackclub/hcb/assets/77016441/80643f5c-959d-4a34-9146-ed663c8e0888

sampoder:
Wow that's sweet!

sampoder (changes requested):
This is incredible! One ask: before we merge this could we hide the donations settings if the page is disabled? And could we also use the slider for requiring peer review?

Thank you so much for all the work on this!

manu-gurudath:
Done! I also relocated the markdown icon in the Transparency Mode settings to prevent layout shifts when toggling.

sampoder (approved):
LGTM!

sampoder:
I've taken out the reimbursements changes because they didn't feel quite ready yet; I think it's because they didn't need an accordion so it felt a bit out of place.

PR #6978, opened 2024-06-28
[Stripe Card] Rename `activated` to `initially_activated` pt 2/3

rluodev:
Part 2: changes code referencing `activated` field to `initially_activated`

@sampoder: please run `StripeCard.update_all("initially_activated = activated")` in prod console just before merge, I think a couple slipped thru the cracks because i forgot to add additional writes in places where activated was being written to (oversight on my part).


sampoder (approved):
Thanks for this PR; tested locally & appears to be working

PR #6980, opened 2024-06-29
`docker-compose.yml`: run `yarn` before starting web container

ajsirota:
## Summary of the problem

Running the dev setup script didn't run `yarn` in the web container, so it never successfully started.

## Describe your changes

Now we run yarn before starting the server, so it has all the dependencies it needs. This has been tested locally.


sampoder:
Shouldn't we do this during setup?

ajsirota:
That would probably be better

ajsirota:
Actually it looks like yarn is already being run in the Dockerfile, so it's weird that that doesn't seem to be happening:
https://github.com/hackclub/hcb/blob/c76861e303610b03428f075a18deffd28e0568ae/Dockerfile#L34

ajsirota:
I'm gonna assume that this is One Of Those Things That Only Happens In My Environment (tm) and close this

PR #6987, opened 2024-07-01
Handle `charge_updated` event & use it to set payout fields

sampoder:
Donations were not being processed because when the `payment_intent_succeeded` event is sent the balance_transaction is null. See https://dashboard.stripe.com/payments/pi_3PXlpGFSaumjmb9r05xiI7uB as an example.

This PR has handle the event that comes in three seconds later to set these fields.


app/controllers/stripe_controller.rb
```diff
@@ -234,6 +234,21 @@ def handle_payment_intent_succeeded(event)
...
+  def handle_charge_updated(event)
+    # only proceed if charge is related to a donation and not an invoice
+    return unless event.data.metadata[:donation].present?
+  
+    # get donation to process
+    donation = Donation.find_by_stripe_payment_intent_id(event.data[:payment_intent])
+  
+    pi = StripeService::PaymentIntent.retrieve(
+      id: donation.stripe_payment_intent_id,
+      expand: ["charges.data.balance_transaction", "latest_charge.balance_transaction"]
```
```suggestion
      expand: ["latest_charge.balance_transaction"]
```

PR #6995, opened 2024-07-01
Improve transaction design

manuthecoder:
This pull request addresses issue #6962 by enhancing the application's design. It removes the distracting background color from entire transactions and applies it to a badge instead. This solution works in dark mode and supports special transactions, ensuring a consistent and clear user experience.

Screenshots: 
image
image
image


sampoder:
We've had this discussion a couple of times, see https://github.com/hackclub/hcb/pull/6308. We already have a version of this (no transaction backgrounds) implemented in production (though it's behind a Flipper flag). The Flipper flag is `:transactions_background_2024_06_05`.

So far, we've largely had people tell us that they'd prefer the backgrounds. https://hackclub.slack.com/archives/CN523HLKW/p1718051042879809

I'm going to close this, I'd prefer for us to iterate over that version rather than re-implementing it.

PR #6998, opened 2024-07-01
Use `::UserSessionMailer` not `UserSessionMailer`

sampoder:
I think this should fix https://hackclub.slack.com/archives/C04LDFJ5X7S/p1719869973705159


sampoder:
honestly this might not be the bug, I need to go test properly

malted (approved):

malted:
oops wait

cjdenio (changes requested):
look _real_ closely (3 Ss instead of 2)

PR #7013, opened 2024-07-03
Add check icon

manuthecoder:
This PR adds a (solo) check icon and replaces the icon for renaming a transaction - might be useful in other areas of the app as well!


sampoder (approved):
LGTM! I updated the check so that it is based on our current checkbox

PR #7014, opened 2024-07-03
Transfer popovers

manuthecoder:
Popovers, popovers, everywhere! Now you can check out transactions on the transfer page with a nifty popover...

image


sampoder (approved):
LGTM!

PR #7024, opened 2024-07-03
Switch to a production key for Plaid

sampoder:
We'll need to replace the SVB account because the access token for Plaid needs to be the production one


app/services/partners/plaid/shared/client.rb
```diff
@@ -27,6 +27,8 @@ def plaid_secret
             Rails.application.credentials.plaid[:development_secret]
           when "sandbox"
             Rails.application.credentials.plaid[:sandbox_secret]
+          when "production"
+            Rails.application.credentials.plaid[:production_secret]
```
cjdenio: I think this will break existing Plaid connections that are still using development mode?

app/services/partners/plaid/shared/client.rb
```diff
@@ -38,7 +40,7 @@ def plaid_env
           # Since we're only using one account & Plaid's development plan supports up
           # to 100 accounts, we're just going to stick with the development key in
           # production for now. Plaid is essentially read-only, so this also works in development.
```
cjdenio:
```suggestion
```

cjdenio (approved):

PR #7045, opened 2024-07-05
Disable changelog popovers only if there was a new post

sampoder:


sampoder:
Logic seems to be reversed

PR #7058, opened 2024-07-08
[Google Workspace] Regularly check DNS after verified

rluodev:
## Summary of the problem

people cant send emails because their domain expired or they removed dns records and get confused

## Describe your changes

we tell them that they're missing dns by scanning all domains nightly


app/views/g_suite_mailer/notify_of_error_after_verified.html.erb
```diff
@@ -0,0 +1,14 @@
+<p>Hi there,</p>
+
+<p>
+  During a regular scan of your domain, <strong><%= @g_suite.domain %></strong>, we detected that at least one critical DNS record that is required for Google Workspace to function properly is missing. If this is not remedied, all users on your Google Workspace may be unable to send or receive emails.
```
garyhtou:
```suggestion
  During a regular scan of your domain, <strong><%= @g_suite.domain %></strong>, we detected that at least one critical DNS record required by Google Workspace is missing. If this is not remedied, all users on your Google Workspace may be unable to send or receive emails.
```

garyhtou (approved):

PR #7059, opened 2024-07-08
[Stripe Card] Rename activated to initially_activated pt 3/4

rluodev:
In this PR, we ignore the `activated` column so writes to the model don't error/crash the app.

This is done using

```rb
self.ignored_columns += ["activated"]
```

this means there are actually 4 steps; the next one is where we actually drop the activated column.


app/models/stripe_card.rb
```diff
@@ -48,6 +48,7 @@
 #  fk_rails_...  (stripe_cardholder_id => stripe_cardholders.id)
 #
 class StripeCard < ApplicationRecord
+  self.ignored_columns += ["activated"]
```
sampoder:
```suggestion
  self.ignored_columns = ["activated"]
```

sampoder (approved):

sampoder:
Nice!

PR #7074, opened 2024-07-10
Rename transactions to financials

manuthecoder:
This PR renames the transactions button in the sidebar/bottom nav to "Financials" to ensure consistent naming with the title

image


sampoder:
Can we do this the other way around?

PR #7088, opened 2024-07-11
FedNow! 💨

malted:


jaspermayone:
Fednowwwwwwwwwwww

garyhtou:
For context, we're putting a pausing on this PR until we know it's feasible to provide this to users. At the moment, Column is quoting us 1.5% on FedNow transfers which is too much for us to eat.

PR #7090, opened 2024-07-11
Improve homepage layout

manuthecoder:
Iteration over @sampoder's feedback on #7074


sampoder:
Sorry, I don't think I worded by feedback too well. Let's keep "Transactions" in the navigation bar.

Something like this prehaps?

Screenshot 2024-07-11 at 11 59 07 PM

PR #7097, opened 2024-07-12
Wire 🌍

malted:


garyhtou:
Similar to the FedNow PR (see https://github.com/hackclub/hcb/pull/7088#issuecomment-2229371252), I need to first ensure we can reasonably provide Column wires to users before we continue here.

garyhtou:
> Similar to the FedNow PR (see [#7088 (comment)](https://github.com/hackclub/hcb/pull/7088#issuecomment-2229371252)), I need to first ensure we can reasonably provide Column wires to users before we continue here.

To provide an update, I'm waiting to hear back from Mel/SVB on our charges/pricing for wires at the moment.

PR #7110, opened 2024-07-14
Restructure `default_stripe_card_personalization_design`

sampoder:


app/models/event.rb
```diff
@@ -717,7 +717,7 @@ def airtable_record
   end
 
   def default_stripe_card_personalization_design
-    stripe_card_personalization_designs.where("stripe_name like '#{name} Black Card%'").order(created_at: :desc).first
+    stripe_card_personalization_designs.where("stripe_name like ?", "#{e.name} Black Card%").order(created_at: :desc).first
```
garyhtou: woah! we have a variable `e` defined here?

PR #7116, opened 2024-07-15
[Login] Create Backup Codes

### LINKED ISSUE #6707: [Login] Create back up codes
Similar to what github has. hash them when storing in the DB

When a user wants to restore their account.
1. Successful complete email login code
2. Enter a valid back up code

rluodev:


app/models/user/backup_codes_list.rb
garyhtou, line 1: - Back up codes should be hashed, not encrypted
- Instead of doing `BackupCodesList`, i'd recommend doing `BackupCode` and then doing a `has_many` with User.

app/views/users/generate_backup_codes.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= turbo_frame_tag :backup_codes_card do %>
+  <div class="card border b--info mt1 mb3 items-center">
+    <p>
+      Write down or save the following codes and keep them in a safe place. You can use these codes to log in if you lose access to your other 2FA methods. <span class="font-bold">You will not be able to see these codes again.</span>
```
garyhtou, line 4:
```suggestion
      Save the following codes in a safe place. You can use these codes to log in if you lose access to your other 2FA methods. <span class="font-bold">You will not be able to see these codes again.</span>
```

app/views/users/generate_backup_codes.html.erb
```diff
@@ -0,0 +1,14 @@
...
+    </p>
+    <ol class="grid grid-cols-1 md:grid-cols-2 gap-2 list-disc text-center items-center align-center m-auto">
+      <% @user.backup_codes_list.codes.each do |code| %>
+        <li class="md:w-[50%] p-0 m-0">
+          <pre class="text-lg m-1 p-2 text-center"><%= code %></pre>
+        </li>
+      <% end %>
+    </ol>
```
garyhtou, line 12: I would love a copy button

app/views/users/generate_backup_codes.html.erb
```diff
@@ -0,0 +1,14 @@
...
+  </div>
+<% end %>
```
garyhtou, line 14:
```suggestion
<% end %>

```

app/models/user/backup_codes_list.rb
garyhtou, line 1: Add a `belongs_to :user_session`

PR #7127, opened 2024-07-15
Make Stripe Card Designs Publicly Available

sampoder:
@garyhtou @YodaLightsabr when'd be a good time to do this?


garyhtou:
LMK once we have all the technical side finished up:
- generating designs for all existing events
- backfilling the default stripe personalization design into HCB
- etc. (the stuff we talked about when reviewing the main pr)

sampoder:
backfilling the default stripe personalization design into HCB is done!

PR #7128, opened 2024-07-16
[Stripe Cardholder] DRY up name-related methods in services

manuthecoder:
See #7124


sampoder (commented):
I seem to be getting:

```
undefined method `id' for nil:NilClass
```

sampoder (approved):

PR #7129, opened 2024-07-16
[Stripe Card] Rename activated to initially_activated pt 4/4

rluodev:
This PR is the last of 4 PRs and drops the `:activated` column, and fully moves over to `:initially_activated`


db/schema.rb
sampoder, line 1: Can you remove the unrelated changes?

sampoder (approved):

PR #7130, opened 2024-07-16
Disable card if user isn't an organizer

manuthecoder:
When an unauthorized user does not have access to an event, this bug appears when clicking on a card: 
image


sampoder:
I think `@show_card_popovers` should be written so that it is only true when `organizer_signed_in?`

sampoder (approved):

PR #7136, opened 2024-07-16
Fix team page

malted:


app/views/organizer_position_invites/_organizer_position_invite.html.erb
```diff
@@ -53,6 +53,9 @@
           </div>
         <% end %>
       </div>
+      <% if organizer_position_invite.initial_control_allowance_amount_cents.present? %>
+        (<%= render_money organizer_position_invite.initial_control_allowance_amount_cents %>)
+      <% end %>
```
sampoder: What if we moved this up so it read "Invited with a $2,000 allowance"?

PR #7140, opened 2024-07-16
[Spending controls] Add allowance validation and add invite copy

malted:
`Yes` is the default now
image

Better copy
image

Also fixes alpine bug


app/views/organizer_position_invites/_organizer_position_invite.html.erb
```diff
@@ -53,8 +53,10 @@
           </div>
         <% end %>
       </div>
-      <% if organizer_position_invite.initial_control_allowance_amount_cents.present? %>
-        (<%= render_money organizer_position_invite.initial_control_allowance_amount_cents %>)
+      <% if organizer_position_invite.initial_control_allowance_amount_cents.present? && !organizer_position_invite.initial_control_allowance_amount_cents.zero? %>
+        with a <%= render_money organizer_position_invite.initial_control_allowance_amount_cents %> allowance
+      <% else %>
+        with no allowance
```
sampoder: I think we can drop this

garyhtou: yeah, we should drop the "with no allowance". the "with a $1 allowance" is nice tho

db/schema.rb
garyhtou, line 1: drop the unrelated changes pls!

app/models/organizer_position/spending/control/allowance.rb
```diff
@@ -41,6 +41,10 @@ def balance_is_positive
           errors.add(:control, "balance must be positive") if control.balance_cents + amount_cents < 0
         end
 
+        def balance_is_nonzero
```
sampoder:
```suggestion
        def balance_is_non_zero
```
Should this have a matching `validate`?

sampoder (approved):
LGTM!

PR #7142, opened 2024-07-16
Revert "[Stripe Card Form] Disable country field (we're US-only)"

sampoder:
Reverts hackclub/hcb#7139
This seems to be causing errors:

Screenshot 2024-07-16 at 3 32 42 PM


YodaLightsabr (approved):

garyhtou:
oops my bad

PR #7144, opened 2024-07-16
Update filter condition for number

YodaLightsabr:
Update the counter to only show the number of fulfillable orders


sampoder (approved):
Worked!

PR #7146, opened 2024-07-17
Add pronouns

manuthecoder:
This PR allows users to add pronouns to their profile from their settings. These pronouns appear in event team pages


sampoder:
I feel like we don't really need to implement this, we're using it purely for display purposes? I don't know, I want to keep HCB functional.

garyhtou:
I wonder if this would help operations

northeastprince:
It's a common convention to put pronouns in parentheses after a name, and I don't see any reason users can't do that if they want to. [It's a hook; let's save the towel racks for when they're worth the tradeoff.](https://world.hey.com/jason/hooks-towel-bars-and-software-14c66c8c)

app/views/users/edit.html.erb
```diff
@@ -38,6 +38,10 @@
         <p class="h5 muted mt0 mb1">If you go by a different name.</p>
         <%= form.text_field :preferred_name, placeholder: @user.full_name.presence || "FiFi Hacks", maxlength: 30 %>
       </div>
+      <div class="field">
+        <%= form.label :pronouns %>
+        <%= form.select :pronouns, ["Prefer not to say", "She/Her", "He/Him", "They/Them", "Other"] %>
+      </div>
```
sampoder: This should probably be a string entry not a dropdown

app/views/users/edit.html.erb
```diff
@@ -38,6 +38,10 @@
         <p class="h5 muted mt0 mb1">If you go by a different name.</p>
         <%= form.text_field :preferred_name, placeholder: @user.full_name.presence || "FiFi Hacks", maxlength: 30 %>
       </div>
+      <div class="field">
+        <%= form.label :pronouns %>
+        <%= form.text_field :pronouns, placeholder: "Pronouns" %>
```
northeastprince: What's the point of the placeholder if there's a label above it?

manu-gurudath: Just wanted to fill in some awkward space - I changed the text a little bit so it isn't the exact same as the label above

sampoder: Can you make it `they/them/theirs`?

app/models/user.rb
```diff
@@ -17,6 +17,7 @@
 #  phone_number_verified    :boolean          default(FALSE)
 #  preferred_name           :string
 #  pretend_is_not_admin     :boolean          default(FALSE), not null
+#  pronouns                 :string
```
garyhtou: validate length of `pronouns` is less than 20-ish characters

app/views/organizer_positions/_organizer_position.html.erb
```diff
@@ -9,6 +9,7 @@
       </div>
       <div class="flex flex-col justify-center">
         <p class="h2 bold m0"><%= organizer_position.user.name %></p>
+        <p class="h5 bold m0"><%= organizer_position.user.pronouns == "Prefer not to say" ? "" : organizer_position.user.pronouns %></p>
```
garyhtou: I'm not sure if we need this `organizer_position.user.pronouns == "Prefer not to say"` check.


```suggestion
        <% if organizer_position.user.pronouns.present? %>
          <p class="h5 bold m0"><%= organizer_position.user.pronouns %></p>
        <% end %>
```

app/views/users/edit.html.erb
```diff
@@ -38,6 +38,10 @@
         <p class="h5 muted mt0 mb1">If you go by a different name.</p>
         <%= form.text_field :preferred_name, placeholder: @user.full_name.presence || "FiFi Hacks", maxlength: 30 %>
       </div>
+      <div class="field">
+        <%= form.label :pronouns %>
```
```suggestion
        <%= form.label :pronouns, "Pronouns (optional)" %>
```

app/views/users/edit.html.erb
garyhtou: I wonder if some users may mistake the placeholder as their pronouns being set to "they/them/theirs".

I think I like the placeholder that Manu has at the moment

PR #7160, opened 2024-07-17
Add Emailbutler to keep track of emails

sampoder:


db/migrate/20240717180813_create_emailbutler_tables.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.string :mailer, null: false
+      t.string :action, null: false
+      t.jsonb :params, null: false, default: {}
+      t.string :send_to, array: true
+      t.integer :status, null: false, default: 0
+      t.datetime :timestamp
+      t.integer :lock_version
+      t.timestamps
+    end
+    add_index :emailbutler_messages, :uuid, unique: true
```
sampoder:
```suggestion
    add_index :emailbutler_messages, :uuid, unique: true

    disable_extension 'uuid-ossp' unless extensions.include?('uuid-ossp')
```

sampoder: Should we do this?

Gemfile.lock
```diff
@@ -772,7 +786,7 @@ GEM
     statsd-ruby (1.5.0)
     stringio (3.1.1)
     stripe (11.7.0)
-    strong_migrations (2.0.0)
+    strong_migrations (1.8.0)
```
sampoder:
```suggestion
    strong_migrations (2.0.0)
```

db/migrate/20240717180813_create_emailbutler_tables.rb
garyhtou: hmmmm, do we know why they chose UUID instead of the usual autoincrementing IDs?

Gemfile
```diff
@@ -222,3 +222,6 @@ gem "audits1984"
 gem "rotp"
 
 gem "ruby-limiter"
+
+gem 'emailbutler'
+gem 'pagy'
```
garyhtou: Is pagy a gem requirement? We have kaminari

garyhtou:
Can you check out `ahoy_email`?

https://github.com/hackclub/hcb/issues/7054#issuecomment-2234848720

PR #7166, opened 2024-07-18
Include AI memo in suggestions dropdown

faisalsayed10:
fixes #7159


app/services/hcb_code_service/suggested_memos.rb
```diff
@@ -16,7 +16,11 @@ def initialize(hcb_code:, event:, confidence: 0.4)
     def run
       return [] unless @event
 
-      ranked_similar_transactions.pluck(:transaction).pluck(:custom_memo).uniq
+      results = ranked_similar_transactions.pluck(:transaction).pluck(:custom_memo).uniq
+
+      results.unshift(@hcb_code.suggested_memos.last) if @hcb_code.suggested_memos.any?
+
+      results
```
sampoder:
```suggestion
      @hcb_code.suggested_memos + ranked_similar_transactions.pluck(:transaction).pluck(:custom_memo).uniq
```

Any reason we aren't get all of the suggested_memos? For transactions with multiple receipts , we may generate multiple suggested memos

PR #7169, opened 2024-07-18
Improve reimbursement keybinds

manuthecoder:
This PR adds `ctrl+enter` and `escape` keybinds to close the lightbox when editing fields in reimbursement expense inputs.


sampoder (approved):
Two things:

1) We currently have ctrl-enter for each of the fields except the description field. Could you add `data-behavior="ctrl_enter_submit"` to that field?

2) I like having the escape key, however, I think that it should not save the form and instead cancel any edits made. Thoughts?

sampoder (changes requested):
Sorry, meant to request changes.

app/javascript/controllers/expense_form_controller.js
```diff
@@ -22,6 +22,13 @@ export default class extends Controller {
       field.readOnly = !this.enabledValue
       field.addEventListener('dblclick', () => this.edit())
       this.#addTooltip(field, 'Double-click to edit...')
+
+      field.addEventListener('keydown', e => {
+        if (e.key === 'Escape') {
+          this.formTarget.reset()
+          this.formTarget.requestSubmit()
```
garyhtou: Let's see if we can close the lightbox without submitting the POST request. however, this fine as is for now

app/views/reimbursement/expenses/_form.html.erb
garyhtou, line 1: Should we change memo to also use ctrl+enter? At the moment it's just enter to submit the form

app/javascript/controllers/expense_form_controller.js
```diff
@@ -136,7 +166,7 @@ export default class extends Controller {
       document.querySelector('.app__sidebar').style.zIndex = '1'
       this.lightboxTarget.addEventListener('click', e => {
         e.preventDefault()
-        this.formTarget.requestSubmit()
+        this.close()
```
sampoder: What's the thinking behind this change?

garyhtou: Please double-check this, but here's my understanding.

Previously, we were using `requestSubmit` to close the lightbox. Now, we have the `close` method.

manu-gurudath: Update: This function is called when the user clicks on the overlay. In my updated code, I've used a combination of requestSubmit() and close(). The reason for this is to make the UI feel more responsive when the user clicks outside the reimbursement input.

garyhtou (commented):
LGTM to me. @sampoder, would you like to do another review pass on this?

app/javascript/controllers/expense_form_controller.js
sampoder: I'm going to remove this because I want there to be feedback for when something is "saved"

```suggestion
```

app/javascript/controllers/expense_form_controller.js
```diff
@@ -22,6 +22,13 @@ export default class extends Controller {
       field.readOnly = !this.enabledValue
       field.addEventListener('dblclick', () => this.edit())
       this.#addTooltip(field, 'Double-click to edit...')
+
+      field.addEventListener('keydown', e => {
+        if (e.key === 'Escape') {
+          this.formTarget.reset()
+          this.close()
+        }
+      })
```
sampoder:
```suggestion
      document.addEventListener('keydown', e => {
        if (e.key === 'Escape' && this.enabledValue) {
          this.formTarget.reset()
          this.close()
        }
      })
```

sampoder (approved):

PR #7170, opened 2024-07-18
[Reimbursement] Decrease contrast on lightbox in light mode

manuthecoder:
This PR reduces the amount of contrast the lightbox currently has for the reimbursement expense form. 
image


sampoder:
Personally, I prefer the old design. I recognise this is one of those ones where different people have different styles but we went with the original design because it felt more in line with the rest of HCB.

manu-gurudath:
Would removing the blur make it better?

garyhtou:
I think i might like this more. i'd need to pull up a side-by-side to be sure

garyhtou (approved):

PR #7174, opened 2024-07-18
Install `ahoy_email` to keep track of emails sent from HCB

sampoder:


app/models/ahoy/message.rb
```diff
@@ -0,0 +1,29 @@
...
+#  index_ahoy_messages_on_to    (to)
+#  index_ahoy_messages_on_user  (user_type,user_id)
+#
+module Ahoy
+  class Message < ApplicationRecord
+    self.table_name = "ahoy_messages"
+
+    belongs_to :user, polymorphic: true, optional: true
+
+    encrypts :to, deterministic: true
```
sampoder:
```suggestion
    belongs_to :user, polymorphic: true, optional: true
```

PR #7185, opened 2024-07-18
Experimenting pinning NPM version instead of Node

garyhtou:
See https://github.com/hackclub/hcb/pull/7184 for context on the problem!

---

Before problem:
```
node: v22.4.1
npm: 10.8.1
yarn: 1.22.22
```

Problem:
```
node: v22.5.0
npm: 10.8.2
yarn: 1.22.22
```

https://github.com/hackclub/hcb/pull/7184 fixed the issue by pinning Node to v22.4.1.

In this PR, I'm going to try allowing Node v22.5.0 and pinning NPM to v10.8.1 to see if that's the problem.


garyhtou:
Nope, it seems to be the node version, not NPM

PR #7187, opened 2024-07-19
Add pagination for reimbursements

manuthecoder:
This PR addresses #7181 and paginates reimbursements

image


app/controllers/events_controller.rb
```diff
@@ -765,6 +765,7 @@ def reimbursements
     @reports = @reports.where(aasm_state: ["reimbursement_approved", "reimbursed"]) if params[:filter] == "reimbursed"
     @reports = @reports.rejected if params[:filter] == "rejected"
     @reports = @reports.search(params[:q]) if params[:q].present?
+    @reports = @reports.order(created_at: :desc).page(params[:page]).per(25)
```
```suggestion
    @reports = @reports.order(created_at: :desc).page(params[:page] || 1).per(params[:per] || 25)
```
Just something we do everywhere else!

sampoder (approved):

PR #7193, opened 2024-07-19
Add merchant icons to Stripe transactions on the ledger 

malted:
image


garyhtou (commented):
~~Let's make as few changes as possible here so we can get this PR over the line.~~

~~1. Add the merchant icon in the left column~~
~~2. Move the user avatars to a column that is immediately to the left of amount.~~

See updated comment below

garyhtou:
After a call with @malted @sampoder @dwahbe, we've decided to do the following:
- Move the date back to the 2nd column (after the merchant/category/type icon)
- Remove the icon for the user avatar column (keep it empty)

Stretch goal:
- Fill in the user avatar column as much as possible (e.g. creator of ACH transfer, etc.)
- To fix performance, consider creating a `/:hcb_code/user.png` route to lazy compute/load the transaction user's avatar.

app/helpers/events_helper.rb
```diff
@@ -46,7 +46,9 @@ def paypal_transfers_airtable_form_url(embed: false, event: nil, user: nil)
   end
 
   def transaction_memo(tx) # needed to handle mock data in playground mode
-    if tx.local_hcb_code.method(:memo).parameters.size == 0
+    if merchant_name = YellowPages::Merchant.lookup(network_id: tx.raw_stripe_transaction&.stripe_transaction&.[]("merchant_data")&.[]("network_id"))&.[]("name")
+      merchant_name
+    elsif tx.local_hcb_code.method(:memo).parameters.size == 0
```
garyhtou: I think this is no longer needed. This functionality was already added in a separate PR

https://github.com/hackclub/bank/blob/e2fdc14afa4777d5c01191cd0d75b5bfdb5531fc/app/models/hcb_code/memo.rb#L30

malted: removed

garyhtou (approved):
heck yeah!

PR #7198, opened 2024-07-20
Upgrade PWA button

manuthecoder:
This PR allows users to add HCB to the home screen by clicking a button in-app! It removes the old instructions which explain how to create a shortcut. It also uses `localStorage` rather than cookies to help make the code slightly more organized

Preview: 
image

Mobile preview:

image


sampoder:
I think that I prefer the old design, it was less in your face. I'm also not sure if we should be prompting to install the PWA on desktop?

garyhtou:
> I think that I prefer the old design, it was less in your face. I'm also not sure if we should be prompting to install the PWA on desktop?

Same here. I had a chat with @manuthecoder about this. Let's keep the old design (and/or look at Reddit's PWA/App prompt). +1 on not prompting for desktop

PR #7201, opened 2024-07-21
Accurately report ACH `estimated_arrival` for transfers sent on the weekday

sampoder:
I noticed that these send on Monday at 3:00am in the Column dashboard, so I've updated this method to reflect that


app/models/ach_transfer.rb
```diff
@@ -288,6 +288,8 @@ def estimated_arrival
       return now.change(hour: 15, minute: 0, second: 0) if now < now.change(hour: 13, min: 30, sec: 0)
     end
 
+    return 0.business_day.after(now).change(hour: 5, min: 30, sec: 0) unless now.workday?
```
sampoder:
```suggestion
    return 0.business_days.after(now).change(hour: 5, min: 30, sec: 0) unless now.workday?
```
imo this reads better lol

PR #7205, opened 2024-07-22
Keep a copy of every email HCB sends

malted:
Closes https://github.com/hackclub/hcb/issues/7054

image


northeastprince:
Bit confused on why we need another dependency here... doesn't seem like that many more lines of code to record it ourselves

garyhtou:
oops. @sampoder should have assigned himself to the ticket.
https://github.com/hackclub/hcb/pull/7174

Let's keep both open for now

malted:
Let's close https://github.com/hackclub/hcb/pull/7174.

db/schema.rb
```diff
@@ -12,8 +12,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_07_16_023501) do
+ActiveRecord::Schema[7.1].define(version: 2024_07_22_213027) do
   # These are extensions that must be enabled in order to support this database
+  enable_extension "amcheck"
```
```suggestion
```

app/views/static_pages/_admin.html.erb
```diff
@@ -66,6 +66,7 @@
       <%= card_to "Sponsors", sponsors_admin_index_path %>
       <%= card_to "Cards", stripe_cards_admin_index_path %>
       <%= card_to "Card designs", stripe_card_personalization_designs_admin_index_path %>
+      <%= card_to "Emails", emails_admin_index_path %>
```
sampoder: We should probably also add it to the nav bar

sampoder (dismissed):

app/views/static_pages/_admin.html.erb
garyhtou: Yes please! (add it to the admin nav bar)

https://github.com/hackclub/bank/blob/6b4b7955f3adf3699aff0bfecf5b08303a0cfabc/app/views/layouts/admin.html.erb#L48-L54

app/controllers/admin_controller.rb
```diff
@@ -1295,6 +1295,44 @@ def account_numbers
     render layout: "admin"
   end
 
+  def email
+    @message_id = params[:message_id]
+
+    @text_content = Ahoy::Message.find(@message_id).text_content
+    # rubocop:disable Rails/OutputSafety
+    @html_content = Ahoy::Message.find(@message_id).html_content.html_safe
+    # rubocop:enable Rails/OutputSafety
```
garyhtou: I think if you move the `html_safe` into ERB, you don't need to disable rubocop

app/controllers/admin_controller.rb
```diff
@@ -1295,6 +1295,44 @@ def account_numbers
...
+
+    render layout: "mailer"
```
garyhtou: Seeing that we need to render the `mailer` layout, I'm assuming that @html_content only contains the body of the mailer and not the full HTML that gets generated.

Is it possible for Ahoy Message to store the full text that is sent (including the layout?)

I would like this so that we have a more accurate representation of the email; especially if we update the layout in the future

app/controllers/admin_controller.rb
```diff
@@ -1295,6 +1295,44 @@ def account_numbers
     render layout: "admin"
   end
 
+  def email
+    @message_id = params[:message_id]
+
+    @text_content = Ahoy::Message.find(@message_id).text_content
```
garyhtou: This variable doesn't appear to be used

app/controllers/admin_controller.rb
```diff
@@ -1295,6 +1295,44 @@ def account_numbers
...
+    @html_content = Ahoy::Message.find(@message_id).html_content.html_safe
+    # rubocop:enable Rails/OutputSafety
+
+    render layout: "mailer"
+  end
+
+  def emails
+    @page = params[:page] || 1
+    @per = params[:per] || 100
+    @q = params[:q].present? ? params[:q] : nil
```
```suggestion
    @q = params[:q].presence
```

app/controllers/admin_controller.rb
```diff
@@ -1295,6 +1295,44 @@ def account_numbers
...
+    @user_id = params[:user_id] || nil
```
garyhtou: I belive the `nil` doesn't do anything!

```suggestion
    @user_id = params[:user_id]
```

app/controllers/admin_controller.rb
```diff
@@ -1295,6 +1295,44 @@ def account_numbers
...
+    @params = params.permit(:page, :per, :q, :user_id)
```
garyhtou: This line is unused since we aren't using the `@params` variable anywhere.

app/controllers/admin_controller.rb
```diff
@@ -1295,6 +1295,44 @@ def account_numbers
...
+
+    @count = messages.count
+
+    @messages = messages.page(@page).per(@per).order(sent_at: :desc)
+
+    respond_to do |format|
+      format.html do
+        render layout: "admin"
+      end
+    end
```
garyhtou: i think the respond_to block is unnecessary here

we can just do `render layout: "admin"`

app/views/admin/emails.html.erb
```diff
@@ -0,0 +1,58 @@
...
+      <th class="w-32">Sent at</th>
+      <th>Content</th>
+      <th>Recipient</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @messages.each do |msg| %>
+      <section class="modal bg-snow" style="height: 60vh;" data-behavior="modal" role="dialog" id="message_<%= msg.id %>">
+        <%= modal_header "Email to #{msg.user.name}" %>
+        <iframe src="<%= url_for(controller: "admin", action: "email", message_id: msg.id) %>" class="w-full h-full absolute border-none -mt-8 -ml-8 pb-24">iframes are not supported</iframe>
```
garyhtou: I think with iframes we'd be loading them unnecessarily even when a user doesn't click on the modal. Maybe our async frame is better?

However, since this is an admin page with low traffic, it's not a big deal

app/views/admin/emails.html.erb
```diff
@@ -0,0 +1,58 @@
...
+            <%= msg.subject %>
+          <% end %>
+        </td>
+        <td>
+          <% if full_name = msg.user.full_name %>
+            <%= link_to admin_user_path(msg.user) do %>
+              <%= full_name %> (<%= msg.user.email %>)
+            <% end %>
+          <% else %>
+            <%= msg.user.email %>
```
garyhtou: Should we also link the user here?

garyhtou (commented):
I have some concerns about making all emails visible to admins. Although we trust admins, it would allow admins to view login codes and access people's accounts without a proper papertrail via impersonation.

sampoder:
I think if we add 

```ruby
has_history only: []
```

To the LoginCodeMailer we can not include login code emails in our tracking, which would probably be a good thing.

garyhtou:
> I think if we add
> 
> ```ruby
> has_history only: []
> ```
> 
> To the LoginCodeMailer we can not include login code emails in our tracking, which would probably be a good thing.


I do think we should still track ALL emails we send. As a temp solution, let's only make the email admin pages available to super admins

app/models/ahoy/message.rb
```diff
@@ -0,0 +1,47 @@
...
+module Ahoy
+  class Message < ApplicationRecord
+    self.table_name = "ahoy_messages"
+
+    belongs_to :user, polymorphic: true, optional: true
+
+    include PgSearch::Model
+    pg_search_scope :search_subject, against: :subject
+
+    encrypts :to, deterministic: true
```
garyhtou: let's remove encryption for `to`

sampoder:
Most engineers aren't superadmins, this will make this feature less helpful if one of the main uses is debugging

sampoder:
To sign in with a login code, the admin would still need the browser token which they can only access through the production console which we do have a paper trail of

app/controllers/admin_controller.rb
```diff
@@ -1295,6 +1295,44 @@ def account_numbers
...
+    @q = params[:q].present? ? params[:q] : nil
+    @user_id = params[:user_id] || nil
+    @params = params.permit(:page, :per, :q, :user_id)
+
+    if @user_id
+      @user =  User.find(@user_id)
+      messages = @user.messages
+    else
+      messages = Ahoy::Message.all
+    end
```
garyhtou: double check this works
```suggestion
    messages = Ahoy::Message.all
    messages = messages.where(user: User.find(@user_id)) if @user_id
```

app/controllers/admin_controller.rb
malted: Rubocop still complains
```
erb interpolation with '<%= (...).html_safe %>' in this context is never safe (not autocorrected)
In file: app/views/admin/email.html.erb:1
```

app/controllers/admin_controller.rb
malted: The full content of the email (text and html) is stored

malted: There must be some sort of normalisation going on because the rendered output is the same, regardless whether it's rendered with the mailer layout or a blank one.

app/controllers/admin_controller.rb
garyhtou: hm, how about `<%= raw @html_content %>`?

app/controllers/admin_controller.rb
garyhtou: Got it, let's drop the `layout: "mailer"` then.

garyhtou:
> To sign in with a login code, the admin would still need the browser token which they can only access through the production console which we do have a paper trail of

that's a good point

garyhtou:
Given @sampoder's message, i'm indifferent to how we restrict access. Maybe we leave it unrestricted for all admins for now

malted:
Addressed all concerns & removed superadmin requirement.

app/views/layouts/admin.html.erb
```diff
@@ -50,7 +50,8 @@
            "HCB Fees": [bank_fees_admin_index_path, BankFee.in_transit_or_pending.count, %i[counter]],
            "Column Statements": [column_statements_admin_index_url, 0, %i[]],
            "Users": [users_admin_index_path, User.count, %i[counter]],
-           "Card Designs": [stripe_card_personalization_designs_admin_index_path, StripeCard::PersonalizationDesign.count, %i[]]
+           "Card Designs": [stripe_card_personalization_designs_admin_index_path, StripeCard::PersonalizationDesign.count, %i[]],
+           "Misc": [emails_admin_index_path, Ahoy::Message.count, %i[counter]]
```
sampoder:
```suggestion
           "Emails": [emails_admin_index_path, Ahoy::Message.count, %i[counter]]
```
Let's use `Emails`

PR #7231, opened 2024-07-25
Optimise transaction renaming

malted:
Skip if was not renamed


sampoder:
Is there any other time we hit the update method besides changing the memo?

garyhtou:
Ideally, this skipping would be handled on the frontend, not the backend

garyhtou:
> Is there any other time we hit the update method besides changing the memo?

I'm thinking this handles the case where you click to rename, then click off the input without changing it

PR #7233, opened 2024-07-25
Redesign account details

malted:
image


jaspermayone:
@malted is that HCB's prod act number??

jaspermayone:
one should probably remove that if so 👀

sampoder:
I don't think it makes sense to remove them from the main page. This page's primary purpose is for people to see their account and routing number which we seem to have removed. I do like the redesigned modal though!

garyhtou:
> one should probably remove that if so 👀

Thanks Jasper! I've redacted it

sampoder:
I've added them back for now so we can get this in because it looks great, we can put that in another PR if everyone else wants to go for removing them. Not sure what others think

sampoder (approved):

PR #7236, opened 2024-07-26
Improve money inputs

manuthecoder:
This PR addresses #7235 and improves money input functionality, including reimbursement amount inputs. It fixes this by selecting all the text inside an input field if the value matches "0.00"

Closes #7235


sampoder:
Would it be possible for us to go with the focus-ing implementation?

manu-gurudath:
done!

sampoder:
Let's just do this for the reimbursement expense form, I think that's the only that has issues

manu-gurudath:
> Let's just do this for the reimbursement expense form, I think that's the only that has issues

fixed!

sampoder (approved):

PR #7238, opened 2024-07-27
Revert "Bump flipper-active_record from 0.28.0 to 1.3.0"

sampoder:
Reverts hackclub/hcb#7219


garyhtou:
Did we run into an issue with this new flipper version?

sampoder:
We haven't upgraded Flipper yet because of this update: https://github.com/flippercloud/flipper/pull/847

PR #7244, opened 2024-07-29
[`organizer_signed_in?`] Fix accidental memoization for role in lambda

garyhtou:
replaces https://github.com/hackclub/hcb/pull/5419


garyhtou:
@malted would you mind fixing up rubocop, sanity checking, and merging?

malted (approved):

app/helpers/sessions_helper.rb
```diff
@@ -87,9 +87,9 @@ def organizer_signed_in?(event = @event, as: :member)
 
     # Memoize results based on method arguments
     @organizer_signed_in ||= Hash.new do |h, key|
-      h[key] = run.call
+      h[key] = run.call(**key)
     end
-    key = [event, as]
+    key = { event:, as: }
```
garyhtou: @malted, You had renamed the lamba keyword params to be `inner_event` and `inner_as`. However, we're still only passing `event` and `as` here!!!

garyhtou: Fix: https://github.com/hackclub/hcb/commit/6d2ceabc2cc0e6f59c3547cfc4890ac9bde85a92

PR #7245, opened 2024-07-29
Reverse order of birthdays (newest year first)

sampoder:


sampoder:
Shout out to this person: https://apidock.com/rails/ActionView/Helpers/DateHelper/date_select#190-Get-year-to-show-in-descending-order-Today-to-192-for-example-

PR #7248, opened 2024-07-30
Fix tag bugs

manuthecoder:
This PR does a couple of enhancements and bug fixes:

1. Decrease label contrast on dark mode for muted tags
2. Adds `flex-shrink: 0` to tags to prevent circle previews from shrinking when other children in a parent container are larger (thanks @sampoder for reporting this bug!)
3. Creates a `.tag__preview` class to reduce redundancy for tag previews


sampoder (approved):
LGTM!

PR #7251, opened 2024-07-30
[User Home page] Limit organizer avatars on Event cards

manuthecoder:
This PR limits the amount of profile pictures displayed on event cards. It's limited to 8, keeping hover styles in mind. 

| Before | After |
|--------|--------|
| image |image |


app/views/events/_event_card.html.erb
```diff
@@ -30,9 +30,14 @@
       </span>
     </div>
     <div class="avatar-row">
-      <% event.organizer_positions.each do |position| %>
+      <% event.organizer_positions.take(8).each do |position| %>
```
garyhtou:
```suggestion
      <% event.organizer_positions.first(8).each do |position| %>
```

garyhtou (approved):

PR #7293, opened 2024-07-31
Improve input focus styles

manuthecoder:
This PR improves the accessibility for inputs when focused by adding a subtle box shadow when focused. It makes the input's border appear thicker, making it easier to see where which input is focused.

| Before | After |
|--------|--------|
| image| image|


sampoder:
I think I preferred the subtler one we used to have

PR #7295, opened 2024-08-01
Create `my` controller

malted:
Closes https://github.com/hackclub/hcb/issues/5427


app/helpers/my_helper.rb
garyhtou, line 1: we can drop this file

config/routes.rb
garyhtou, line 1: please group the my stuff together

garyhtou: you might be able to use singular `resource` to define the paths

```
# i think this works??
resource :my do
  get :cards
end
```

app/controllers/my_controller.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+class MyController < ApplicationController
+  skip_after_action :verify_authorized # do not force pundit
```
garyhtou: can you add an `only:`?

i want all of these to be explicitly defined so that if someone else adds an action here in the future, they're forced to either authorize it, or add it to this list

garyhtou:
@sampoder if this is good to go, can you merge it?

config/routes.rb
sampoder: this would produce routes like `cards_my_path` which don't exactly make sense; I choose to stick with the current system and bunch them up within the scope so it's easy to locate things.

sampoder (approved):

PR #7303, opened 2024-08-01
Log out after inactivity

malted:
Closes https://github.com/hackclub/hcb/issues/7258


db/migrate/20240801213950_remove_session_duration_seconds_from_users.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveSessionDurationSecondsFromUsers < ActiveRecord::Migration[7.1]
+  def change
+    safety_assured do
+      remove_column :users, :session_duration_seconds, :integer
+    end
+  end
+
+end
```
sampoder: Before we drop columns normally we add them to the model's ignored columns in an earlier PR to make sure that everything is alright

app/models/user_session.rb
```diff
@@ -116,6 +116,14 @@ def set_as_peacefully_expired
...
   LAST_SEEN_AT_COOLDOWN = 5.minutes
   def touch_last_seen_at
+    if last_seen_at
+      diff_mins = (Time.now - last_seen_at) / 60
+      if diff_mins > User.session_validity_preference_minutes[user.session_validity_preference_minutes]
+        set_as_peacefully_expired
+        destroy
+      end
+    end
+
```
sampoder: Maybe we do this as a cron job?

app/helpers/sessions_helper.rb
```diff
@@ -24,7 +24,7 @@ def unimpersonate_user
   # DEPRECATED - begin to start deprecating and ultimately replace with sign_in_and_set_cookie
   def sign_in(user:, fingerprint_info: {}, impersonate: false, webauthn_credential: nil)
     session_token = SecureRandom.urlsafe_base64
-    expiration_at = Time.now + user.session_duration_seconds
+    expiration_at = Time.now + 2.weeks
```
garyhtou: SESSION_DURATION.from_now

app/models/user.rb
```diff
@@ -4,30 +4,30 @@
...
+#  locked_at                           :datetime
+#  payout_method_type                  :string
+#  phone_number                        :text
+#  phone_number_verified               :boolean          default(FALSE)
+#  preferred_name                      :string
+#  pretend_is_not_admin                :boolean          default(FALSE), not null
+#  receipt_report_option               :integer          default("none"), not null
+#  running_balance_enabled             :boolean          default(FALSE), not null
+#  seasonal_themes_enabled             :boolean          default(TRUE), not null
+#  session_validity_preference_minutes :integer          default("6_hours")
```
garyhtou: session_activity_timer

app/models/user.rb
```diff
@@ -68,6 +68,15 @@ class User < ApplicationRecord
...
   ], scopes: false, default: :user
 
+  # Number of hours
+  enum :session_validity_preference_minutes, {
+    "15 minutes": 15,
+    "1 hour": 60,
+    "6 hours": 60 * 6,
+    "1 day": 60 * 24,
+    "3 days": 60 * 24 * 3
+  }
```
```suggestion
  }, prefix: true
```

app/models/user.rb
```diff
@@ -68,6 +68,15 @@ class User < ApplicationRecord
     :superadmin,
   ], scopes: false, default: :user
 
+  # Number of hours
+  enum :session_validity_preference_minutes, {
+    "15 minutes": 15,
+    "1 hour": 60,
+    "6 hours": 60 * 6,
+    "1 day": 60 * 24,
+    "3 days": 60 * 24 * 3
```
garyhtou: add 1 week and 2 weeks

app/views/users/edit_security.html.erb
```diff
@@ -91,6 +91,17 @@
       Sign out everywhere else
     <% end %>
   </div>
+  <%= form_with(model: @user, local: true, class: "pb2", html: { onsubmit: "onSubmit()" }, data: { turbo_frame: "_top" }) do |form| %>
+    <div class="field">
+      <%= form.label :session_validity_preference_minutes, "Sign me out after" %>
+      <%= form.select :session_validity_preference_minutes, [["Never", nil]] + User.session_validity_preference_minutes.to_a %>
```
garyhtou: get rid of the never option

db/migrate/20240801210007_add_session_validity_preference_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddSessionValidityPreferenceToUsers < ActiveRecord::Migration[7.1]
+  def change
+    add_column :users, :session_validity_preference_minutes, :integer, default: 60 * 6
```
garyhtou: add null: false

app/views/users/edit_security.html.erb
```diff
@@ -91,6 +91,17 @@
       Sign out everywhere else
     <% end %>
   </div>
+  <%= form_with(model: @user, local: true, class: "pb2", html: { onsubmit: "onSubmit()" }, data: { turbo_frame: "_top" }) do |form| %>
```
garyhtou: what does `html: { onsubmit: "onSubmit()" }` do?

db/migrate/20240801213950_remove_session_duration_seconds_from_users.rb
garyhtou: it takes 3 prs:
1. remove all usages and ignore columns
2. actually drop the column
3. remove the ignore column code

app/models/user.rb
garyhtou: make 3 days the default

db/migrate/20240801213950_remove_session_duration_seconds_from_users.rb
malted: https://github.com/hackclub/hcb/pull/7408

app/jobs/session_expiry_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class SessionExpiryJob < ApplicationJob
+  queue_as :low
+  def perform
+    UserSession.joins(:user).where("user_sessions.last_seen_at + (users.session_validity_preference * interval '1 minute') < NOW()").find_each { |session|
+      session.set_as_peacefully_expired
+      session.destroy
+    }
```
garyhtou: please convert this into a `do` block.

`do` blocks are preferred for multi-line while `{}` is preferred for singe lines

app/models/user.rb
```diff
@@ -4,30 +4,30 @@
...
-#  locked_at                :datetime
-#  payout_method_type       :string
-#  phone_number             :text
-#  phone_number_verified    :boolean          default(FALSE)
-#  preferred_name           :string
-#  pretend_is_not_admin     :boolean          default(FALSE), not null
-#  receipt_report_option    :integer          default("none"), not null
-#  running_balance_enabled  :boolean          default(FALSE), not null
-#  seasonal_themes_enabled  :boolean          default(TRUE), not null
-#  session_duration_seconds :integer          default(2592000), not null
```
garyhtou: I think this annotation needs to be updated. I belive we aren't dropping this column in this PR

app/views/users/edit_security.html.erb
```diff
@@ -91,6 +91,17 @@
...
+    <div class="field">
+      <%= form.label :session_validity_preference, "Sign me out after" %>
+      <%= form.select :session_validity_preference, [["Never", nil]] + User.session_validity_preference.to_a %>
```
garyhtou: never shouldn't be an option

app/controllers/users_controller.rb
```diff
@@ -346,7 +346,6 @@ def user_params
       :profile_picture,
       :pretend_is_not_admin,
       :sessions_reported,
-      :session_duration_seconds,
       :receipt_report_option,
       :birthday,
```
sampoder, line 409: Do we need to add the new parameter?

sampoder (commented):
Quick thing, this is looking pretty good

app/controllers/users_controller.rb
malted: Yes! Done.

PR #7310, opened 2024-08-02
Require users to verify their TOTP config before saving

sampoder:
Closes https://github.com/hackclub/hcb/issues/7180


app/controllers/users_controller.rb
```diff
@@ -156,11 +156,24 @@ def edit_security
...
     @user.totp&.destroy!
-    @user.create_totp!
+    @totp = @user.build_totp
+    @totp.validate
+  end
+
+  def enable_totp
+    @user = params[:id] ? User.friendly.find(params[:id]) : current_user
+    authorize @user
+    @totp = @user.build_totp(secret: params[:secret])
```
garyhtou: We shouldn't receive this secret from the front end. Instead, I think we should create the TOTP record with an aasm state

garyhtou: Receiving this from the front-end would allow people to set the secret to anything they'd like!

PR #7311, opened 2024-08-02
Allow card grants to be topped up

sampoder:
Closes https://github.com/hackclub/hcb/issues/7241


app/views/card_grants/actions/_topup.html.erb
```diff
@@ -0,0 +1,21 @@
...
+
+  <%= link_to "#",
+              data: { behavior: "modal_trigger", modal: "topup" },
+              class: "btn" do %>
+    <%= inline_icon "plus", size: 20 %> Topup grant
+  <% end %>
+
+  <div class="modal modal--scroll bg-snow" data-behavior="modal" role="dialog" id="topup">
+    <%= modal_header "Topup grant" %>
+    <%= form_with(url: topup_event_card_grant_path(id: card_grant.hashid, event_id: @event.slug), class: "w-full") do |form| %>
```
garyhtou: is event_id used?

app/models/card_grant.rb
```diff
@@ -104,6 +104,21 @@ def pending_invite?
...
+    DisbursementService::Create.new(
+      source_event_id: event_id,
+      destination_event_id: event_id,
+      name: "Topup of funds for grant to #{user.name}",
+      amount: amount_cents / 100,
+      destination_subledger_id: subledger_id,
+      requested_by_id: topped_up_by.id,
+    ).run
+
+    update(amount_cents: self.amount_cents + amount_cents)
```
garyhtou: probably should be in a transaction. double check that the contents of the service work within a transaction

garyhtou: bang this!

garyhtou (commented):
🪄

PR #7331, opened 2024-08-05
Add Email to `PaymentRecipient`

sampoder:
```ruby
PaymentRecipient.all.each do |pr|
  ach_transfer = pr.ach_transfers.last || AchTransfer.where(recipient_name: pr.name, event_id: pr.event_id).last
  pr.update(email: ach_transfer.recipient_email)
end
```

```ruby
PaymentRecipient.where(email: nil).destroy_all!
```

Closes https://github.com/hackclub/hcb/issues/5792


sampoder:
Haven't done the deleting part yet. I'm on the fence about it, there's 55.

garyhtou:
> Haven't done the deleting part yet. I'm on the fence about it, there's 55.

As long as we're sure weren't used, then I think it's okay to delete. I'd assume most of them were accidentally created to people testing.

Just double check that these 55 PaymentRecipients aren't actually supposed to be linked to some AchTransfers (I saw you did a `where` with name and email. maybe a more fuzzy and manual search will result in more links)

PR #7355, opened 2024-08-06
Remove table header from ledger

sampoder:
I'm getting a sense of what people would prefer rn

Screenshot 2024-08-06 at 5 10 28 PM


sampoder:
wasn't too popular... maybe some other time

garyhtou:
I think we can 100% do this once we redesign the transaction into cards

garyhtou:
Once they're cards, it wouldn't make sense to have the table headings

PR #7357, opened 2024-08-06
Switch off of `Event.friendly.find` for all remaining POST requests

sampoder:


garyhtou:
~~that's very unfriendly of you :)~~

PR #7381, opened 2024-08-07
Convert images to PNGs before running OCR

sampoder:
Closes https://github.com/hackclub/hcb/issues/7360


garyhtou:
I would have somewhat expected for the file to have already been converted by Active Storage.
https://github.com/hackclub/bank/blob/bea8f3b7579d11a3b3ed8fa18cad75acb160706d/config/initializers/active_storage.rb#L3-L8

PR #7389, opened 2024-08-07
Add indicator to comment reaction bar

manuthecoder:
This PR addresses #7364 and shows an indicator in the comment reaction bar when you've reacted with an emoji!

image

Closes #7364


sampoder (dismissed):
This is sweet; thanks for this!

PR #7390, opened 2024-08-07
Add cash withdrawals to public activity

sampoder:


app/views/public_activity/raw_pending_stripe_transaction/_create.html.erb
```diff
@@ -6,6 +6,10 @@
     <%= render layout: "/public_activity/common", locals: { activity:, current_user:, url: hcb_code_path(hcb_code) } do %>
       was refunded <%= render_money hcb_code.amount_cents.abs %> from <i><%= hcb_code.memo %></i>
     <% end %>
+  <% elsif hcb_code.stripe_cash_withdrawal? %>
+    <%= render layout: "/public_activity/common", locals: { activity:, current_user:, url: hcb_code_path(hcb_code) } do %>
+      withdrew <%= render_money hcb_code.amount_cents.abs %> from <i><%= humanized_merchant_name(hcb_code.stripe_merchant) %></i>
+    <% end %>
```
garyhtou: Should this happen after the `declined?` condition? I'm assuming cash withdrawals can be declined; when they are, we should render a declined for activity (not a withdrawal)

PR #7396, opened 2024-08-07
Block cash withdrawals unless they're enabled on the card

sampoder:
Using Flipper is a temporary measure, just want to have something in place just in case


app/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb
```diff
@@ -50,6 +50,14 @@ def decline_with_reason!(reason)
         false
       end
 
+      def cash_withdrawal?
+        auth[:merchant_data][:category_code] == "6011"
+      end
+
+      def cash_withdrawals_not_allowed?
+        Flipper.enabled?(:cash_withdrawals_2024_08_07, card)
+      end
```
garyhtou: I think it would be best if the flipper flag held cards that were _allowed_. Essentially, make it a white list instead of a black list.

PR #7402, opened 2024-08-07
Fix `set_event` for transactions exports

sampoder:


app/controllers/concerns/set_event.rb
```diff
@@ -10,6 +10,7 @@ module SetEvent
 
     def set_event
       id = params[:event_name] || params[:event_id] || params[:id]
+      id = params[:event] if params[:event].is_a?(String) && id.nil? # sometimes params[:event] is a hash with nested attributes
```
garyhtou: you can ||= and drop the nil check

sampoder:
```suggestion
      id ||= params[:event] if params[:event].is_a?(String) # sometimes params[:event] is a hash with nested attributes
```

PR #7410, opened 2024-08-08
Make reimbursement payout holdings reversible

sampoder:
Closes https://github.com/hackclub/hcb/issues/7404.


app/models/reimbursement/payout_holding.rb
```diff
@@ -59,12 +60,42 @@ class PayoutHolding < ApplicationRecord
...
+        transitions from: :failed, to: :reversed
+      end
     end
 
     def payout_transfer
       ach_transfer || increase_check || paypal_transfer
     end
 
+    def reverse!
+      raise ArgumentError, "must be a failed payout holding" unless failed?
```
garyhtou: Should we confirm that the ACH/Check/PayPal money has been returned?

app/models/reimbursement/expense_payout.rb
```diff
@@ -79,6 +84,30 @@ def state_text
...
     end
+    
+    def reverse!
+      raise ArgumentError, "must be a settled expense payout" unless settled?
+
+      ActiveRecord::Base.transaction do
+
+        mark_reversed!
+
+        canonical_pending_transaction.decline!
```
garyhtou: we don't want to decline after it's been settled

app/models/reimbursement/payout_holding.rb
```diff
@@ -59,12 +60,42 @@ class PayoutHolding < ApplicationRecord
...
+        # these are reversed because this is reverse!
+        sender_bank_account_id = ColumnService::Accounts.id_of(book_transfer_receiving_account)
+        receiver_bank_account_id = ColumnService::Accounts.id_of(book_transfer_originating_account)
+      
+        ColumnService.post "/transfers/book",
+                           amount: amount_cents.abs,
+                           currency_code: "USD",
+                           sender_bank_account_id:,
+                           receiver_bank_account_id:,
+                           description: "HCB-#{local_hcb_code.short_code}"
```
garyhtou: flip the transfer order so that we aren't pulling out of a $0 account

PR #7455, opened 2024-08-12
Use `render_money` instead of `number_to_currency`

sampoder:
`render_money` supports cents!


sampoder:
Fixing this bug:

Screenshot 2024-08-12 at 11 44 28 AM

PR #7478, opened 2024-08-13
[Navbars] Make the dock item icons smaller

malted:
cc @dwahbe 

image


garyhtou (changes requested):
We should also update the size of the user sidebar icons

image

malted:
this change already does this @garyhtou

garyhtou (approved):
Great!

I didn't realize the user nav was using EventsHelper's dock_item!!

PR #7479, opened 2024-08-13
Hide author column on receipt bin tx rows

malted:
Fixes #7465


garyhtou:
See https://github.com/hackclub/hcb/issues/7465#issuecomment-2285117567

garyhtou:
We want the partials to not render the author column unless explicitly requested to

garyhtou:
This is because I think the only location where we want the author is on the org ledger, but these partials are used in many other places!

malted:
Just did that!

sampoder (approved):
LGTM!

PR #7483, opened 2024-08-13
Fix support button positioning on mobile & combine three dots into the event's navigation bar

sampoder:
Screenshot 2024-08-13 at 2 54 48 PM


garyhtou:
thanks!

PR #7484, opened 2024-08-13
Add more FRC flavor texts

ajsirota:
## Summary of the problem
There aren't enough robotics-specific flavor texts.

## Describe your changes
Now there are more. Some of these may be bad jokes, but that's fine, right?


app/services/flavor_text_service.rb
```diff
@@ -124,7 +124,12 @@ def frc_flavor_texts
       "Duct tape, ductape, duck tape",
       "🦆 📼",
       "Build season? HCB season!",
-      "Build season already?"
+      "Build season already?",
+      "Graciously accepting your team's donations",
+      "Your team may graciously decline, but your cards won't",
+      "<s>Not</s> held together with zip ties".html_safe,
+      "3, 2, 1, HCB!",
+      "If you have transparency mode on, is that money-pit-scouting?"
```
garyhtou: i love it!

PR #7487, opened 2024-08-13
Remove deprecated plural `admin_tools` helper

garyhtou:
## Summary of the problem

it's been deprecated!

## Describe your changes

I think there are no longer any usages, so we could we good to remove!


garyhtou:
someone please sanity check and merge for me! ty

PR #7491, opened 2024-08-13
Only create CPTs for the amount paid, not the amount due

sampoder:
We can't operate under the assumption that "paid" invoices are paid in full.

See the later messages in https://hackclub.slack.com/archives/C047Y01MHJQ/p1723568226574489


garyhtou (dismissed):
the code you have should be valid, however, it may be preferred to not change the parameters of the service.

Instead, use @invoice.remote_invoice (if that exists), or if we sync amount paid to a column, then do

sync_from_stripe! and then directly access the column

PR #7498, opened 2024-08-14
Show faces on top of the "+X" icon

sampoder:
Screenshot 2024-08-14 at 9 59 46 AM
Screenshot 2024-08-14 at 9 59 52 AM

Faces are fun!


sampoder:
lmao that looks like 6+ it's just an upside down +9

PR #7513, opened 2024-08-14
Add Tasks

sampoder:
This PR also creates the `HasTasks` and `IsTaskable` concerns as well as an example task. It doesn't have a UI at the moment.

Screenshot 2024-08-16 at 12 56 32 AM


app/models/task.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+# Indexes
+#
+#  index_tasks_on_assignee  (assignee_type,assignee_id)
+#  index_tasks_on_taskable  (taskable_type,taskable_id)
+#
+
+class Task < ApplicationRecord
+  belongs_to :taskable, polymorphic: true
+  belongs_to :assignee, polymorphic: true
```
garyhtou: validate presence

app/models/task.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  after_initialize do
+    raise "Cannot directly instantiate a Task" if self.instance_of? Task
+  end
```
garyhtou: maybe try

self.abstract_class = true

app/models/task.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def update_complete!
+    nil
+  end
```
garyhtou: great job using the Template Method design pattern!

app/models/task/receiptable/upload.rb
```diff
@@ -0,0 +1,17 @@
...
+class Task
+  module Receiptable
+    class Upload < Task
+      def update_complete!
+        if taskable.missing_receipt?
+          update(complete: false)
+        else
+          update(complete: true)
+        end
+      end
```
garyhtou: THIS IS AMAZING

garyhtou: once it's complete, can it be transitioned back to incomplete?

it looks like this code supports it, and i think it makes sense; at least with receipts

app/models/task/receiptable/upload.rb
```diff
@@ -0,0 +1,17 @@
...
+
+class Task
+  module Receiptable
+    class Upload < Task
+      def update_complete!
+        if taskable.missing_receipt?
+          update(complete: false)
+        else
+          update(complete: true)
+        end
```
garyhtou:
```
update(complete: !taskable.missing_receipt?)
```

app/models/concerns/receiptable.rb
```diff
@@ -38,5 +40,11 @@ def no_or_lost_receipt!
 
       raise e
     end
+
+    after_create_commit do
+      if missing_receipt? && (try(:author) || try(:user) || try(:event))
+        Task::Receiptable::Upload.create(taskable: self, assignee: try(:author) || try(:user) || try(:event))
+      end
```
garyhtou: this works! another way to write it is:

asignee = try ...
if missing receipt && user
  tasks.create(assignee:)
end

app/models/concerns/receiptable.rb
```diff
@@ -38,5 +40,11 @@ def no_or_lost_receipt!
 
       raise e
     end
+
+    after_create_commit do
+      if missing_receipt? && (try(:author) || try(:user) || try(:event))
+        Task::Receiptable::Upload.create(taskable: self, assignee: try(:author) || try(:user) || try(:event))
```
garyhtou: we should bang this and wrap it with a `safely`

app/models/concerns/is_taskable.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module IsTaskable
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :tasks, as: :taskable, dependent: :destroy
+
+    def update_task_completion!
```
garyhtou: i think we can drop the bang. the name makes it clear we're gonna write to the db.

also, maybe we should wrap this in `safely` to prevent it erroring out

app/models/task.rb
sampoder: Screenshot 2024-08-15 at 10 53 16 AM

Looks like that freaks ActiveRecord out :/

app/models/task/receiptable/upload.rb
sampoder: yeah, at the moment it can. it's probably only going to be some kinds of tasks that behave like this

app/models/task/receiptable/upload.rb
sampoder: oh thats much cleaner

app/models/task.rb
```diff
@@ -0,0 +1,49 @@
...
+#  index_tasks_on_taskable  (taskable_type,taskable_id)
+#
+
+class Task < ApplicationRecord
+  belongs_to :taskable, polymorphic: true
+  belongs_to :assignee, polymorphic: true
+  validates_presence_of :taskable, :assignee
+  broadcasts_refreshes_to ->(task) { [task.assignee, :tasks] }
+
+  scope :complete, -> { where(complete: true) }
```
garyhtou: I think this is already created for you

garyhtou:
```ruby
before_save do
  # for the actual method name, check the ActiveRecord Dirty module
  completed_at = complete_changes_to(true) ?  Time.now : nil
end
```

Remove the `update(completed_at: ...)` from the `update_complete!` method

app/views/my/tasks.html.erb
```diff
@@ -0,0 +1,40 @@
...
+</h1>
+
+<% @tasks.incomplete.each do |task| %>
+  <%= link_to task.url, class: "text-decoration-none", target: "_blank"  do %>
+    <div class="card bg-none flex items-center g2 mt2">
+      <input type="checkbox" style="height: 24px; width: 24px; pointer-events: none" />
+      <div class="flex flex-col">
+        <div class="bold"><%= task.text %></div>
+        <div>
+          Created <%= distance_of_time_in_words(Time.now, task.created_at) %> ago
```
garyhtou: Attach the Stimulus that auto updates the distance.

See if you can money patch the `distance_of_time_in_words` to automatically include the stimulus controller

app/views/my/tasks.html.erb
```diff
@@ -0,0 +1,40 @@
...
+        <div class="flex flex-col">
+          <div class="bold"><%= task.text %></div>
+          <div>
+            Completed <%= distance_of_time_in_words(Time.now, task.completed_at) %> ago
+          </div>
+        </div>
+      </div>
+    <% end %>
+  <% end %>
+<% end %>
```
garyhtou:
```suggestion
<% end %>

```

app/views/my/tasks.html.erb
```diff
@@ -0,0 +1,40 @@
...
+        </div>
+      </div>
+    </div>
+  <% end %>
+<% end %>
+
+<% if @tasks.complete.any? %>
+  <h3>Recently completed</h3>
+  
+  <% @tasks.complete.order(completed_at: :desc).first(5).each do |task| %>
```
garyhtou:
```suggestion
  <% @tasks.complete.where("completed_at > ?", 24.hours.ago).order(completed_at: :desc).first(5).each do |task| %>
```

garyhtou (dismissed):
🪄

PR #7515, opened 2024-08-14
Add `receipt_required?` to `Transaction` & `CanonicalTransaction`

sampoder:


sampoder:
Receipts are required for HCB codes, however, these two models still implement Receiptable. Not having a `receipt_required?` method can cause some funny issues, see tests in: https://github.com/hackclub/hcb/pull/7513/.

Receipts aren't required for either so I've gone with `false` for both.

PR #7524, opened 2024-08-15
Modify date logic in `ReceiptService::Suggest`

sampoder:
This PR does two things:

1) Switch over to using absolute values, see this for context: https://hackclub.slack.com/archives/GBXBFEED9/p1723674862006769?thread_ts=1723663750.056969&cid=GBXBFEED9

2) Add an in-between value so that if the dates don't perfectly match we still make the distance a little less

@YodaLightsabr thoughts?


app/services/receipt_service/suggest.rb
```diff
@@ -51,12 +51,19 @@ def distance(hcb_code)
...
+
+            distance = ((hcb_code.pt&.raw_pending_stripe_transaction&.created_at || hcb_code.date).to_date - @extracted.extracted_date.to_date).abs
+
+            if distance <= 1
+              0
+            elsif distance <= 5
+              0.2 + 0.8 * (distance / 5)
+            else
+              1
+            end
```
garyhtou: this reminds me of piece-wise functions lol

YodaLightsabr (approved):

app/services/receipt_service/suggest.rb
sampoder: lol yes

PR #7531, opened 2024-08-15
Strip email signatures on comment replies

sampoder:
I've chosen to directly set `content` for the time being because I want to see if we can just do that and avoid bloating the `Comment` model.


app/mailboxes/comment_mailbox.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "email_reply_parser"
```
garyhtou: It shouldn't be necessary to `require` since Rails autoloads

garyhtou (commented):
I do think that the full context is important in case the parser accidentally removes too much.

PR #7534, opened 2024-08-15
Use purple instead of red inside "choose from Receipt Bin modal"

sampoder:
Can someone sanity check this is the right thing to change? Closes https://github.com/hackclub/hcb/issues/7520


garyhtou (commented):
Should we define an $ai variable?

we somewhat did that here:
https://github.com/hackclub/bank/blob/234eeb0845db607843ae7c896b45f8aa23350436/app/assets/stylesheets/_variables.scss#L28

YodaLightsabr (approved):

PR #7542, opened 2024-08-15
Filter out verification transactions on amount filters

malted:


app/controllers/events_controller.rb
```diff
@@ -105,6 +105,10 @@ def show
       end_date: @end_date
     ).run
 
+    if @minimum_amount || @maximum_amount
+      @all_transactions = @all_transactions.filter { |t| !t.likely_account_verification_related? }
+    end
```
garyhtou: please test this!
```suggestion
    if @minimum_amount || @maximum_amount
      @all_transactions.reject!(&:likely_account_verification_related?)
    end
```

garyhtou: Ideally, this would be a search modifier in [here](https://github.com/hackclub/bank/blob/6a98b7b0251ed94e8a87df47f183dd2b2714f476/app/services/transaction_grouping_engine/transaction/all.rb#L3-L2) that follows the same logic as https://github.com/hackclub/bank/blob/6718f4e2a0a5aff20df46965ed3599504bcce75b/app/models/canonical_transaction.rb#L323-L325

garyhtou (approved):

PR #7544, opened 2024-08-15
[Org Ledger] Fix FS fee row

garyhtou:
image

It was missing a `<td>` and needed to be re-ordered

--

now fixed!


garyhtou:
FYI @malted since this came from adding the avatar row

PR #7573, opened 2024-08-17
Remove the "details" link from the transaction table

sampoder:
Removes the much loved "Details" link from the transaction table:

Screenshot 2024-08-17 at 6 25 10 PM

Also adds shift click to rename!

Double click to rename? Never heard of it. Never existed.

image

Thoughts? Not going to merge until I get some thoughts.


malted:
big fan of shift click to rename. who would do such a thing as to create double click to rename?? the streets call that foolish

also the memo being the details link is super intuitive

sampoder:
Chatted with Ian + Ben, largely positive feedback. In the future, it'd be nice to make the whole transaction clickable and get rid of the underline.

garyhtou:
+1 on getting rid of the underline

sampoder:
The fix to getting rid of the underline... is surprisingly complicated. I want to keep this as a native <a />. Which to make the whole transaction clickable, means making it no longer a table. One option might be to only show the underline on hover but I'm keeping this for right now so users get used to it.

PR #7584, opened 2024-08-19
[User] Redesign user settings onboarding

manuthecoder:
This PR redesign's HCB's onboarding experience. Below is a screenshot showing the changes made

image

As always, feel free to give additional feedback!


sampoder:
Damm! My only worry is that it might look a little out of place when next to the rest of our UI. This reminds me a lot of `hackclub/site`'s styling which is ever so slightly different.

We should probably use a standard button and we should make the columns equal width imo, it's a bit difficult to read at the moment. Also what if the data toggle was a slider?

garyhtou (commented):
I like the general direction of this.

Work with @sampoder to fix up the styling and get this merged. Here are a few things we talked about:
- Fields are misaligned
- Add log out button
- Remove email field
- Blank slate seems to be effect
- Make it obvious that preferred name is optional
	- Either add required `*` or "(optional)"
- Reword the right text (make it feel less serious)
- Check to make sure merge conflicts were resolved correct. I see that data sharing was added back to onboarding (we want to remove that). And @sampoder 's change to the birthday year selector is not in this PR.

sampoder (dismissed):
Thanks Manu for the great work on this!

Couple of comments:

1) The HCB logo is two big in comparision to the logout logo, I think it's alright to make it smaller.

2) Is there a reason for the circular border around the logout icon, I think it's alright to just do we have in the sidebar currently with the tooltip.

3) The bottom button still looks out of place, can we use a standard button?

Lastly, whilst we're making improvements to this page, we should probably improve the profile picture input to make it letter cluttered. Any ideas?

I'll work on getting an updated list of orgs for the right hand.

manu-gurudath:
image

changes applied! (see my message on slack regarding the button color ✨)

app/assets/stylesheets/application.scss
```diff
@@ -477,3 +477,7 @@ summary {
 .suggested_tag::after {
   transform: translateY(0px) translateX(-4px);
 }
+
+.iti {
```
garyhtou: Just leave a comment about what ITI is; or preferably, see if you can add the CSS inline

app/views/users/edit.html.erb
```diff
@@ -1,147 +1,277 @@
...
+        <p>
+        <h3 class="-mt-2.5 opacity-50">Let's get your profile set up in less than two minutes.</h3>
+        <%= form_with(model: @user, local: true, html: { onsubmit: "onSubmit()" }, data: { turbo_frame: "_top" }) do |form| %>
+          <%= form_errors @user, "profile", "We couldn't update your" %>
+          <% if current_season(override_preference: true) %>
+            <div class="field field--checkbox">
+              <%= form.check_box :seasonal_themes_enabled %>
+              <%= form.label :seasonal_themes_enabled, "#{by_season "✨", override_preference: true, fall: "🎃", winter: "🎄"} Enable #{current_season(override_preference: true)} theme", class: "h3 bold" %>
+            </div>
+          <% end %>
```
garyhtou: Hide this during onboarding

garyhtou (commented):
Right photo gallery has a weird bottom margin; and the top images/icon is clipped.

image

garyhtou:
this LGTM after the CDN links are added. @sampoder any thoughts?

app/views/users/edit.html.erb
```diff
@@ -1,147 +1,258 @@
...
         <% end %>
       </div>
-      <div class="field">
-        <%= form.label :preferred_name %>
-        <p class="h5 muted mt0 mb1">If you go by a different name.</p>
-        <%= form.text_field :preferred_name, placeholder: @user.full_name.presence || "FiFi Hacks", maxlength: 30 %>
-      </div>
+      <h1 class="border-0 mt-10 mb-2">Create your account</h1>
+      <turbo-frame id="settings" autoscroll data-autoscroll-behavior="smooth">
+        <p>
```
```suggestion
```

app/views/users/edit.html.erb
```diff
@@ -1,147 +1,258 @@
...
+        <h3 class="-mt-2.5 opacity-50">Let's get your profile set up in less than two minutes.</h3>
```
```suggestion
        <p class="muted">Let's get your profile set up in less than two minutes.</p>
```

garyhtou (approved):

PR #7592, opened 2024-08-20
Prefill previous shipping address

sampoder:
For folks who still use the default billing address! Or move a lot..


app/helpers/stripe_cards_helper.rb
```diff
@@ -28,30 +28,36 @@ def suggested(field)
     when :name
       current_user.full_name
     when :line1
-      current_user&.stripe_cardholder&.stripe_billing_address_line1 ||
+      current_user&.stripe_cards&.physical&.last&.stripe_shipping_address_line1 ||
```
garyhtou: we should set `last_physical_card` and `cardholder` as variables above to DRY this up a bit

PR #7595, opened 2024-08-20
Fix menu placement

manuthecoder:
Fixes #7594 by changing the menu position


app/views/application/_support_button.html.erb
```diff
@@ -1,4 +1,4 @@
-<div data-controller="menu" data-menu-placement-value="<%= content_for?(:nav) ? "bottom-start" : "bottom-center" %>">
+<div data-controller="menu" data-menu-placement-value="bottom-start">
```
garyhtou: should this be `top-end`? I think `bottom-start` would technically put the menu below `?` button, and thus below the page's viewport

sampoder: `top-end` seems like the right thing to do, at least based on https://floating-ui.com/docs/tutorial#placements

manu-gurudath:
Should we merge this?

sampoder:
Does `top-end` fix it as well?

PR #7634, opened 2024-08-21
Add in-house support for wires!

sampoder:
Screenshot 2024-08-21 at 2 57 16 PM

It's a wire! This PR is currently missing:

* Mailers
* Payment recipient support
* Author icons
* Activity log

They also have to be manually processed and mapped which is ugh.


app/models/wire.rb
```diff
@@ -0,0 +1,124 @@
...
+#  account_number_bidx       :string           not null
+#  account_number_ciphertext :string           not null
+#  amount_cents              :integer          not null
+#  approved_at               :datetime
+#  bic_code_bidx             :string           not null
+#  bic_code_ciphertext       :string           not null
+#  memo                      :string           not null
+#  payment_for               :string           not null
+#  recipient_email           :string           not null
+#  recipient_name            :string           not null
```
garyhtou: we'll need to add a handful more fields to this. TBD

app/models/hcb_code.rb
```diff
@@ -337,6 +342,10 @@ def paypal_transfer
     @paypal_transfer ||= PaypalTransfer.find_by(id: hcb_i2) if paypal_transfer?
   end
 
+  def wire
+    @wire ||= Wire.find_by(id: hcb_i2) if wire?
+  end
+
```
garyhtou: should we make this an association?

app/views/hcb_codes/_icon.html.erb
```diff
@@ -14,6 +14,9 @@
   <% elsif tx.local_hcb_code.paypal_transfer? %>
     <% label = "Paypal transfer" %>
     <% icon = inline_icon "paypal" %>
+  <% elsif tx.local_hcb_code.wire? %>
+    <% label = "Wire" %>
+    <% icon = inline_icon "bolt" %>
```
garyhtou: low priority: find a better icon. it seems like ppl generally use a global with arrows

app/models/wire.rb
```diff
@@ -0,0 +1,124 @@
...
+
+# == Schema Information
+#
+# Table name: wires
+#
+#  id                        :bigint           not null, primary key
+#  aasm_state                :string           not null
+#  account_number_bidx       :string           not null
+#  account_number_ciphertext :string           not null
+#  amount_cents              :integer          not null
```
garyhtou: we'll need to be able to handle various currencies

app/models/wire.rb
```diff
@@ -0,0 +1,124 @@
...
+
+  has_one :canonical_pending_transaction
+
+  monetize :amount_cents, as: "amount"
+
+  include PublicActivity::Model
+  tracked owner: proc{ |controller, record| controller&.current_user }, event_id: proc { |controller, record| record.event.id }, only: [:create]
+
+  after_create do
+    create_canonical_pending_transaction!(event:, amount_cents: -amount_cents, memo: "OUTGOING WIRE", date: created_at)
```
garyhtou: With currency conversion, it's impossible to accurately predict the amount cents. This means that the CT may not equal the CPT. I think this should be fine since it's outgoing and not fronted??? need to double check

sampoder: Yeah, that's fine. We don't front these so it's all good.

garyhtou (commented):
please also design this in a way where if we need to switch over to [Column](https://column.com/docs/wire), we don't need to overhaul everything.

(maybe use Column's terminology)

app/views/hcb_codes/_icon.html.erb
sampoder: yeah, maybe? I don't mind the bolt but acknowledge it isn't exactly a transaction type nor does it communicate that it's an international transfer

app/models/wire.rb
sampoder: interesting... that's going to make this a little tough

sampoder:
yeah, I had a look at Column's API. I'm pretty confident that we could convert this over. The only hard part might be mapping the final transfer but that should be fixable. 

the only thing right now is that Column has some optional address fields which this doesn't currently have but it'd be pretty easy for us to add those.

app/models/hcb_code.rb
sampoder: Could do! We only do that the reimbursement payouts at the moment, the rest of our transfers use this system

sampoder:
The quotes system described in https://column.com/docs/international-wires/outgoing is interesting.

garyhtou:
> The quotes system described in [column.com/docs/international-wires/outgoing](https://column.com/docs/international-wires/outgoing) is interesting.

ah that's interesting! It is somewhat similar to how SVB works. Except their quotes are only valid for a few minutes

db/schema.rb
```diff
@@ -12,7 +12,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2024_09_20_170139) do
+ActiveRecord::Schema[7.1].define(version: 2024_09_22_202224) do
```
```suggestion
ActiveRecord::Schema[7.2].define(version: 2024_09_22_202224) do
```

PR #7641, opened 2024-08-22
⚡ Make every ACH same-day. Same-day, every ACH.

sampoder:


garyhtou (dismissed):
Can you move this into the model?

garyhtou:
Instead of passing `same_day: true`, we can do on_initialize same_day = true

garyhtou:
We can remove this too: https://github.com/hackclub/bank/blob/f48176397466be42eb4a096a502d508caf849a66/app/services/reimbursement/payout_holding_service/nightly.rb#L48

PR #7645, opened 2024-08-22
Capitalization edits from @dwahbe

garyhtou:


app/views/users/_settings_nav.html.erb
```diff
@@ -14,7 +14,7 @@
   <% end %>
   <%= settings_tab active: active == "previews" do %>
     <%= link_to @user == current_user ? settings_previews_path : previews_user_path(@user), data: { turbo: true, turbo_action: "advance" } do %>
-      <span class="xs-hide">Feature&nbsp;</span>Previews
```
sampoder: This has been on my list to fix for a while but on mobile this will show up as "previews"

garyhtou: Ah good point

garyhtou: Maybe we can do some css magic

garyhtou: Or, just show the full "Feature previews" on mobile. I'm not sure how strong the need is to truncate

PR #7650, opened 2024-08-22
[Google Workspace] Fix table when verification errors

rluodev:
## Describe your changes
stuff


app/controllers/events_controller.rb
```diff
@@ -510,6 +512,20 @@ def g_suite_overview
             @result[1] = true
           end
         end
+        unless @g_suite.dkim_key.blank?
```
```suggestion
        if @g_suite.dkim_key.present?
```

app/views/events/g_suite_overview.html.erb
```diff
@@ -132,13 +132,19 @@
                 </tr>
                 <% unless @g_suite.dkim_key.blank? %>
```
```suggestion
                <% if @g_suite.dkim_key.present? %>
```

garyhtou (approved):

PR #7656, opened 2024-08-22
[UI] Stats: Fix large stat sizing

lachlanjc:
## Summary of the problem

Before, clicking between pages like Invoices & Donations, the top stats moved horizontally, due to CSS workarounds on each page. Now, they remain in place.

## Describe your changes

| Before | After |
| ------ | ------ |
| CleanShot 2024-08-22 at 17 02 14@2x | CleanShot 2024-08-22 at 17 01 14@2x |
| CleanShot 2024-08-22 at 17 06 55@2x | CleanShot 2024-08-22 at 17 07 22@2x |


sampoder (approved):
Thanks for this!

PR #7658, opened 2024-08-22
`sponsorship_fee` -> `revenue_fee`

sampoder:
Part of https://github.com/hackclub/hcb/pull/6296. Splitting this off to make https://github.com/hackclub/hcb/pull/6296 more reviewable.


garyhtou:
ty

garyhtou (commented):
Double check assignment works. I'm not seeing a `revenue_fee=` method; maybe it's not needed

sampoder:
I'm going to continue using sponsorship_fee for assignment. We don't do `sponsorship_fee = X` anywhere we just do `{sponsorship_fee: X}` (or some variation of that)

PR #7662, opened 2024-08-22
[UI] Receipts: Centralize import methods

lachlanjc:
## Summary of the problem

Before, the Receipt Center page had a bunch of grey text with instructions on how to import receipts. Copying the links required using context menus. 
CleanShot 2024-08-22 at 18 27 23@2x

For email, it sent you to your account settings to find or change your forwarding address, which then linked back to the Receipt Bin page (we refer to this page as Receipts, Receipt Center, Receipt Bin, and the URL is My Inbox).
CleanShot 2024-08-22 at 18 27 35@2x

## Describe your changes

I've centralized these instructions and configuration as parallel options at the top of the Receipts page:

CleanShot 2024-08-22 at 18 28 57@2x


app/views/events/g_suite_overview.html.erb
sampoder, line 1: Fixed in https://github.com/hackclub/hcb/pull/7661 as well!

app/views/my/inbox.html.erb
```diff
@@ -6,23 +6,55 @@
...
-  <p class="muted h3">
-    Alternatively, text your receipts to <%= sms_to "+1-864-548-4225", class: "break-none" %>
-    <% if @mailbox_address %>
-      or email them to <%= mail_to @mailbox_address.address, class: "break-none" %>.
-    <% else %>
-      or configure an email for receipt uploads in <%= link_to "your settings", my_settings_path %>.
-    <% end %>
-  </p>
+  <details class="my-4" <% @mailbox_address ? "open" : "" %>>
+    <summary>
```
sampoder: Thoughts on always showing them?

lachlanjc: It's not ideal to give up the vertical space at the top of the screen, but moving the section down doesn't make sense. If we drop the heading & card styling we won't use as much vertical space though!

app/views/mailbox_addresses/_mailbox_address.html.erb
garyhtou, line 1: BTW, the tooltips on the right of the card gets clipped
image

image

app/views/my/inbox.html.erb
lachlanjc: image

sampoder: I personally prefer this style but don't mind either way, we can merge something in and then improve on top of that

sampoder (approved):
LGTM!

app/views/mailbox_addresses/_mailbox_address.html.erb
lachlanjc: It's true, but we've had that bug for 5 years…

garyhtou: No longer relevant since the card was dropped!

garyhtou (approved):
Good to go!

garyhtou:
Thank you for these PRs @lachlanjc!

PR #7663, opened 2024-08-22
[UI] Streamline account settings

lachlanjc:
## Summary of the problem

- In the onboarding state, we had an overwhelming number of fields. It's pretty rare that users want to change their session duration or analytics settings, and they're not critical to getting started, so saved those for later.
- The notification options all had different copy and icons to refer to the same concepts, and had an unrelated setting for receipt importing in between. Grouping these as notifications and using consistent icons & copy makes them easier to parse.
- For our single-select tiles, having different widths & overflow onto two rows made the options look more overwhelming than they are.

(The receipt email generation is moving to Receipt Bin in #7662.)

## Describe your changes

| Before | After |
| ------ | ------ |
| Screen Shot 2024-08-22 at 18 52 17 | Screen Shot 2024-08-22 at 18 51 30 |
| image | image |


sampoder (dismissed):
These look good to me! Let's get https://github.com/hackclub/hcb/pull/7662 merged and then this as well. Thanks for cleaning this up.

app/views/users/edit.html.erb
```diff
@@ -82,89 +83,91 @@
...
+        <fieldset class="field">
+          <%= form.label :receipt_report_option, "Receipt reminders" %>
+          <p class="h5 muted mt0 mb1">
+            Get a regular email roundup of your transactions awaiting receipts.
+          </p>
+          <%= form.select :receipt_report_option,
+                          (User.receipt_report_options.map { |k, v| [k.humanize, k] }) %>
+        </fieldset>
+        <fieldset class="field">
+          <legend class="font-medium">Transaction comments</legend>
```
garyhtou: just as an FYI, this setting also affects comments on other models such as Reimbursement Reports

app/views/users/edit.html.erb
```diff
@@ -82,89 +83,91 @@
...
             <% end %>
           </div>
         </fieldset>
         <fieldset>
-          <legend class="heading h2 pt2">Charge notifications</legend>
-          <p class="h5 muted mt1 mb2">
-            How would you like to receive charge notifications?
+          <legend class="font-medium">Card use</legend>
+          <p class="h5 muted mt-0 mb-3">
+            Get a notification when there’s a charge on your debit card.
```
```suggestion
            Get a notification when there’s a charge on your HCB card.
```

garyhtou (commented):
Thank you for this!

app/views/users/edit.html.erb
sampoder: On yes, good catch, hence the threads wording

lachlanjc: Nice, added that information. "My Threads" I think wasn't clear enough, this should be clearer

sampoder: This is a minute detail but "my threads" also included threads you had previously commented on

lachlanjc:
Glad you like! Can one of you hit approve?

sampoder (approved):
Let's merge in #7662 first so that people don't lose access to their mailbox address

PR #7664, opened 2024-08-22
Create a Hack Club affiliate plan + annotate all models

sampoder:


app/models/event/plan/hack_club_affiliate.rb
garyhtou, line 1: Just a thought:

At the moment, when we want to enable something for HQ-projects only, we'd use the HQ category find them.

Now, can look for a `hack_club_affilated` feature (if we add one). It is also possible to look for this specific plan, but directly accessing plans should be discouraged.

sampoder: True. I guess going forward we could put that feature in the features array, like how we were talking about with card grants.

garyhtou: Yup!

PR #7668, opened 2024-08-22
Rename "Receipt centre" to "My receipts"

sampoder:
This also moves the badge to transactions, which gives it a bit more context. And removes the increasingly awkward "Receipt Bin" header:

Screenshot 2024-08-22 at 4 56 47 PM

The "Send in" features are technically part of "Receipt Bin". It was starting to get confusing having this title there.


sampoder:
My receipts was the title of the page, we had two names for it and it was getting confusing.

PR #7671, opened 2024-08-23
Document quick expense via SMS & email

sampoder:
Inspired by @lachlanjc's work on the receipts page.

Screenshot 2024-08-22 at 5 36 15 PM

Next thing is finding a way to remove the Review requested / My reports pill because I'm finding it distracting.


sampoder:
These have existed for a while, just they were never documented so a few people used them.

PR #7701, opened 2024-08-23
Replace the transactions table with a list of flexbox rows

sampoder:
Screenshot 2024-08-23 at 7 57 24 AM

Same old transaction table, now with flexbox. I've only switched over the ledger for now. Not sure how urgent removing the underline is so...

We also now use a combined partial (Manu had a similar PR to also do that but looks like it never got merged).

Can someone review and sanity check this? It's a fairly risky merge.


malted:
May you explain why you want to do this?

sampoder:
Sorry yeah. So Gary asked me to remove the underline from the memos because
those are now links. I thought the best way to do that would be to make the
whole row clickable but unfortunately you can’t make a table row a link
cleanly, hence this.

sampoder:
If there’s a better way to do it, I’m down… I’d prefer to keep it a table
if possible

lachlanjc:
I just solved this problem on another project! We can keep the a tag where it is, but add an onclick to the rows with JS that funnels the click event to the a tag, then add CSS for a row hover effect. This:

- Maintains the a11y of the table, which is crucial for screenreaders (this PR would otherwise significantly worsen navigation of transactions)
- Maintains the keyboard navigation behavior we have today, where only the transaction name is focusable
- Improves the experience for pointer users with an expanded clickable area
- Improves the aesthetics by removing the underline

sampoder:
Good idea! Let's do that.

PR #7706, opened 2024-08-23
[UI] Transactions: Use tabular numbers

lachlanjc:
## Summary of the problem

It's visually faster to parse the magnitude of numbers when each digit is the same width, so commas align vertically.

## Describe your changes

| Before | After |
| ------ | ------ |
| image | image |


sampoder (approved):

sampoder:
We should use this in more places imo, it’s a nice touch!

garyhtou:
That's amazing!

PR #7708, opened 2024-08-25
[UI] Use upload icon for receipts

lachlanjc:
## Summary of the problem

The previous icon meant share, and was being used inconsistently.

## Describe your changes

Before:
CleanShot 2024-08-25 at 15 01 31@2x

After:
image


sampoder (approved):
Thanks for this! I've also added them to the reimbursements receipt upload.

PR #7712, opened 2024-08-25
[UI] Use tabular numbers consistently

lachlanjc:
## Summary of the problem

I began rolling out tabular nums in #7706, and Sam suggested thinking through more thoroughly where else they make sense. Turns out, a bunch of places!

## Describe your changes

- We often have several badges near one another, which can be consistent widths
- Setting all numbers in tables fixes every numeric column app-wide
- Stats make a huge difference, by removing the visual jitter of hovering over the account balance graph:

Before:
CleanShot 2024-08-25 at 15 57 44

After:
CleanShot 2024-08-25 at 15 58 03


sampoder (approved):
Looking good!

PR #7713, opened 2024-08-25
[UI] Admin: Fix overlapping badges

lachlanjc:
## Summary of the problem

Though it's admin-only UI, the overlapping chevrons on badges have been bothersome.

## Describe your changes

Before:
CleanShot 2024-08-25 at 15 59 31@2x

After:
CleanShot 2024-08-25 at 16 00 02@2x


sampoder (approved):

sampoder:
Thanks!

PR #7716, opened 2024-08-25
[UI] Donations/Reimb: Simplify settings dialogs

lachlanjc:
## Summary of the problem

Our "popover" modals come from the side to indicate you're drilling in hierarchically to nested items. Settings are not a sub-item of a root view, they're an accessory, so they don't make sense as popovers; standard modal dialogs are more appropriate. At the same time, because the dialog is already a container, we can take the settings out of cards since they don't need visual grouping.

## Describe your changes

Before:
CleanShot 2024-08-25 at 19 16 43@2x

After:
image

Review with whitespace off.


sampoder (approved):
Amazing!

PR #7512, opened 2024-08-14
Improve dark mode

manuthecoder:
# Summary

## Dark mode improvements
-   Relocates the dark mode toggle to user settings (refer to the screenshot below).
-   Storing the dark mode preference as a cookie instead of using `localStorage`. Previously, the user's theme preference was stored via `localStorage`, which, despite preloading the script, couldn't entirely prevent a Flash of Unstyled Content (FOUC). The new approach uses cookies, allowing Ruby to automatically set `data-dark="true"` on the HTML element when the user loads the page, effectively avoiding a FOUC.
-   Introducing a `system` theme option. **Important:** The new implementation uses a `system_preference` cookie in `dark.js` to further prevent a FOUC on subsequent reloads.

## Login page
- Removes unnecessary buttons next to logo

image

PR #7717, opened 2024-08-26
Improve user sidebar

manuthecoder:
This PR essentially takes the user sidebar changes out of #7512. Essentially, it improves the UI of the `user` sidebar page by merging the settings button and name typography into one button. Below is an example of what it looks like:

image

Feel free to let me know what you think!


sampoder:
I've gone back and forward on this. Personally, I think I prefer the old design. I guess my thought process behind that was that the email is helpful information for users (as compared to Stripe or Column, where you might have a personal account and a business account). Additionally, I don't think that the "Settings" dock item is taking up too much space in the sidebar.

sampoder:
I do want to say that that this looks pretty nice and I appreciate the effort that you put in here!

PR #7729, opened 2024-08-26
Prefill goods or services checkbox

sampoder:
Screenshot 2024-08-26 at 3 21 50 PM


garyhtou:
Can we rename the modal header to "Prefilled link"

PR #7740, opened 2024-08-27
Improve Stripe card logos

sampoder:


app/models/event.rb
```diff
@@ -726,10 +726,10 @@ def revenue_fee
   def generate_stripe_card_designs
     ActiveRecord::Base.transaction do
       stripe_card_personalization_designs.update(stale: true)
-      stripe_card_logo.blob.open do |file|
-        ::StripeCardService::PersonalizationDesign::Create.new(file:, color: :black, event: self).run
-        ::StripeCardService::PersonalizationDesign::Create.new(file:, color: :white, event: self).run
-      end
+      file = attachment_changes['stripe_card_logo'].attachable
```
garyhtou: just as an FYI, if you later decide to call `generate_stripe_card_designs` from the console, this line would fail

sampoder: Yep! Adding a fallback

PR #7741, opened 2024-08-27
[UI] Visual refresh across core components

lachlanjc:
## Summary of the problem

We've had the same plastic-y looking buttons that grab focus with slightly garish colors & shadows for 4 years. They've been fun, but now that we have more pages & more features, elements are competing for attention, and it makes the app mentally more burdensome to use than necessary. We can tone down some of these colors & font sizes & shadows and improve peace & clarity across the site. It's a new era for HCB!

## Describe your changes

I've refreshed styling across:

- Card shadows
- Buttons
- Menus (dropping the loud hover state)
- Inputs (including now having two sizes, one smaller one for search bars so they fit perfectly)
- Dock
- Filterbars

CleanShot 2024-08-27 at 17 51 10@2x
CleanShot 2024-08-27 at 17 51 31@2x
CleanShot 2024-08-27 at 17 51 00@2x

The transaction filter menu shows the input/menu changes the best. I've removed the visual jiggling on hover, and fixed the different input heights across the tabs, the tabs adjusting width as they're selected, the size of the tabs, the width of the dividers:

https://github.com/user-attachments/assets/cc8b6e04-f90f-4a43-99e7-6e20966bb5fa


sampoder:
Moving the Export button up to the top is a sweet move; it's been awkwardly hovering there for a while :D

garyhtou (commented):
Observation: I noticed that this PR adds a subset of the icon changes from https://github.com/hackclub/hcb/pull/7773; I wonder if that was intentional.

Since I didn't get the chance to look through the majority for this PR and I'm OOO after today, I'm asking @sampoder to take point on reviewing this PR.

app/assets/images/icons/full-screen.svg
sampoder, line 1: Is there a reason for deleting this icon?

app/views/users/_nav.html.erb
```diff
@@ -19,7 +19,7 @@
     </button>
   </div>
 
-  <nav class="dock font-medium mt1">
+  <nav class="dock font-medium mt2">
     <%= dock_item "Home",
```
sampoder: Screenshot 2024-08-31 at 12 23 19 PM

To me the alignment here looks off, with the box around "Home" sticking out.

app/assets/stylesheets/components/_buttons.scss
sampoder, line 1: Cmd+clicking on dock items seems to cause this:

Screenshot 2024-08-31 at 12 31 59 PM

app/assets/stylesheets/components/_forms.scss
sampoder, line 1: Screenshot 2024-08-31 at 12 34 22 PM

On `/my/inbox` the buttons inside the inputs feel like they need padding around them

sampoder, line 1: FYI:

Screenshot 2024-08-31 at 12 36 13 PM

sampoder, line 1: Similar to an above comment, but pinned transactions used to rely on some sort of margin from the form below:

Screenshot 2024-08-31 at 12 37 23 PM

app/assets/stylesheets/components/_buttons.scss
sampoder, line 1: Screenshot 2024-08-31 at 12 38 49 PM

These two don't seem lined up (bottom of `/hq`)

app/assets/stylesheets/components/_filters.scss
```diff
@@ -1,15 +1,11 @@
...
-.pagination,
-.filterbar {
   padding: 0.75rem 0;
   align-self: center;
 }
 
-.filterbar__filters {
-  overflow-x: auto;
-  -webkit-overflow-scrolling: touch;
+.filterbar {
```
sampoder, line 1: I love the new filter bar on `/invoices`!

app/assets/stylesheets/components/_forms.scss
```diff
@@ -10,13 +10,13 @@ input[type='url'],
...
 textarea {
   -webkit-appearance: none;
-  background-color: map-get($palette, snow);
+  background-color: transparent;
   border-radius: 6px;
   border: 1px solid map-get($palette, smoke);
   color: inherit;
   display: block;
   font-size: 1rem;
-  min-height: 2.5rem;
```
sampoder, line 19: Personally I preferred the subtler border we had, especially in dark mode, this one really stands out

sampoder: This largely applies to dark mode.

app/assets/stylesheets/components/_forms.scss
sampoder, line 1: Screenshot 2024-08-31 at 1 15 16 PM

The input on the command bar is a little off

app/assets/images/icons/full-screen.svg
lachlanjc: It doesn't match the visual style of the rest of the set. I replaced it with a new version in the icon PR!

sampoder (changes requested):
I've left a couple more comments about things I found clicking around the interface. I haven't been able to click through everything so I might come back with a bit more in a bit!

These PRs are always hard to review because at the end of the day it comes down to personal preference. There are parts of this PR I love and parts of our current design system I love.

I guess I can run through different parts of the PR.

I'm a fan of the less-rounded buttons, they'll be a lot easier to use late down the line! I also really like the edited filter dropdown without the gradient on hover and the reduced border radius. 

One little thing on colours, I think that I prefer the brighter colours on the radio buttons in dark mode (in light mode, what this PR brings in looks good)

The monospaced tooltips aren't really working in my opinion; I think it's because we don't use a monospace font off so it doesn't fit in with the others.

There also seems to be a top border on tables, what was thinking behind that?

I think the last comment I'll make is that I think the new selected indicator on dock items is a bit distracting and doesn't line up well.

sampoder:
cc: @malted & @YodaLightsabr what do you all think? want to make sure everyone gets a chance to join in

app/assets/images/icons/full-screen.svg
sampoder: Sounds good!

lachlanjc:
Thank you for all your QA, @sampoder! Ready for another round whenever you have a moment. I don't have almost any sample data locally to test against, so I can't see all the potential states.

lachlanjc:
I would love to get this merged ASAP to reduce merge conflicts & unblock some other PRs! Do you want to do more QA @sampoder? Anything I can help with?

sampoder:
Sorry I missed this comment, please feel free to contact me via Slack (I get a lot of GitHub notifs!). I'm waiting to hear about others thoughts about this at the moment

app/views/events/donation_overview.html.erb
```diff
@@ -15,34 +15,45 @@
...
+      </button>
+      <div class="menu__content menu__content--2 h5" data-menu-target="content">
+        <%= link_to start_donation_donations_path(@event), target: "_blank" do %>
+          <%= inline_icon "external" %>
+          Open page
+        <% end %>
+        <button class="menu__action" type="button" tabIndex="0" data-controller="clipboard" data-clipboard-text-value="<%= start_donation_donations_url @event %>" data-action="click->clipboard#copy click->menu#close">
+          <%= inline_icon "link" %>
+          Copy link
+        </button>
```
garyhtou, line 39: This "Copy link" button is missing a hover effect.


https://github.com/user-attachments/assets/352b4c0b-28c2-4d89-93ed-711f0a947102


It would also be beneficial to have a confirmation/indication that the link was copied.

app/assets/stylesheets/components/_buttons.scss
```diff
@@ -3,34 +3,39 @@ input[type='submit'],
...
+
+  &:not([class*='bg-']):not([type='submit']) {
+    box-shadow: var(--shadow-border);
+
+    &:hover,
+    &:focus,
+    &[aria-expanded='true'] {
+      box-shadow: var(--shadow-border-md);
+    }
   }
```
garyhtou: I agree that our previous hover effect was too prominent. However, this new hover effect is not visible on my displays. I've heard similar feedback from the team (e.g. @YodaLightsabr).

https://github.com/user-attachments/assets/8ad22419-2870-449c-b0ea-faef28aca5e1

A visible hover state (alongside a `pointer` cursor) is important for indicating that a button/element is clickable. Our buttons have lost that visual cue.

As a solution, my first intuition may be to use a larger/darker `box-shadow` and slightly darken the background color. However, I'm sure you already have a couple of ideas here. I'm happy to do some brainstorming/research with you, if you'd like!

app/assets/stylesheets/components/_forms.scss
```diff
@@ -27,18 +27,31 @@ textarea {
...
     cursor: no-drop;
     color: map-get($palette, slate);
   }
+
+  &::placeholder {
+    color: map-get($palette, divider);
+  }
+  &::-moz-placeholder {
+    color: map-get($palette, divider);
+  }
```
garyhtou, line 44: In dark mode, this placeholder color appears as if it is text. I think this may confuse some users.

|Before (contrast ratio 3.87:1)|After (10.95:1)|
|-|-|
| image | image |

In light mode, this placeholder color is too light. For example, here's what the ACH Transfer form looks like:

|Before (4.41:1)|After (1.55:1)|
|-|-|
| image | image |

Seeing that the new placeholder color is too dark in dark mode and too light in light mode, I'd recommend making it darker.

In dark mode, the contrast ratio is up at 10.95:1. Given that the [recommended minimum is 4.5:1](https://accessibleweb.com/question-answer/does-placeholder-text-need-to-meet-color-contrast-requirements/), I think we have some room to make the placeholder color darker when in dark mode; which will also help solve light mode.

app/views/events/settings/_details.html.erb
```diff
@@ -105,7 +105,7 @@
         <% end %>
       </div>
     </div>
-    <ul class="mt0 pl2 mb2">
+    <ul class="mt0 pl4 mb2 slate">
```
garyhtou: Adding `slate` results in low contrast in dark mode (ratio of 1.65:1).

image

I would recommend `muted` instead since that seems to play better when considering both light and dark mode. Here's a preview:

|Dark mode (4.87:1)|Light mode (3.16:1)|
|-|-|
| image | image |

In addition, `muted` will also match the other body-like copy that we have on the page.
image

However, this is still not perfect since the light mode contrast is a tad low (3.16:1).

app/views/events/_title.html.erb
```diff
@@ -1,21 +1,30 @@
...
   <% if show_org_switcher? %>
     <div class="menu__content menu__content--switcher sm:w-[275px]" data-menu-target="content">
-      <p class="muted font-[800] text-sm mb-0 uppercase m-1">Organizations</p>
-      <% current_user.events.not_hidden.excluding(@event).with_attached_logo.map do |event| %>
+      <% admin_tool("p0 mb1") do %>
+        <button class="menu__action" data-controller="clipboard" data-clipboard-text-value="<%= @event.name %>" data-action="click->clipboard#copy">
+          <%= inline_icon "copy", size: 24 %>
+          Copy org name
+        </button>
+      <% end %>
```
garyhtou, line 15: This is much cleaner! Thank you

image

app/views/events/g_suite_overview.html.erb
garyhtou, line 1: image

I see that the `.info` class was renamed to `.bg-info` (in  `_badges.scss:41`). I believe line 26 should have its usage updated to `bg-info`.

https://github.com/hackclub/bank/blob/6cf7be91d513bed30a107ae6477ec5eaadba4dac/app/views/events/g_suite_overview.html.erb#L26

app/views/application/_logo.html.erb
```diff
@@ -26,16 +26,16 @@
...
+      Download dark PNG
     </a>
     <div class="menu__divider"></div>
     <a href="/branding">Brand guidelines</a>
     <div class="menu__divider"></div>
-    <a href="/" target="_blank">Open link in new tab</a>
+    <a href="/" target="_blank">Open home in new tab</a>
     <% if current_user&.admin_override_pretend? %>
       <div class="menu__divider" data-label="Admin"></div>
       <%= button_to "#{"✔ " if current_user.pretend_is_not_admin}Pretend not to be an admin", user_path(current_user), method: :patch, params: { pretend_is_not_admin: !current_user.pretend_is_not_admin }, class: "menu__action admin-tools", data: { turbo: false } %>
```
garyhtou, line 43: Similar to my comment in `app/views/events/donation_overview.html.erb`, this button is also missing a hover effect
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/0add054b-a090-4fcf-9504-0f20be73f5b0"/>|<video src="https://github.com/user-attachments/assets/8e7e65b0-1e45-4f7c-822c-f70761b961a4"/>|

app/views/organizer_position/spending/controls/index.html.erb
```diff
@@ -103,7 +103,7 @@
 <% end %>
 
 <% if @active_control %>
-  <div class="m-auto text-center mb-5">
+  <div class="m-auto text-center mb-5 filterbar">
```
garyhtou, line 106: I like the addition of the new filterbar UI. This particular usage of the filterbar is a bit tricky because it fills the full width.

image

Other usages of filterbar typically have a search bar to fill the remaining width.

image

If we take a close look at that first screenshot, we can see that the border from filterbar spans across the entire width. I would recommend using `width: fit-content` to have it wrap tightly around the filter buttons. Here's a preview:

image

app/views/application/_logo.html.erb
garyhtou: It appears all javascript-based menu items are missing their hover effect.

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/348aa4eb-3218-4fef-b334-ed4ad3f3ad1e"/>|<video src="https://github.com/user-attachments/assets/34608d5f-c067-4789-a245-98952c9503f3"/>|

This specific recording was taken on an HcbCode Show page. Make sure you have transaction tags enabled for the organization.

A similar menu can be accessed on an Organization's admin settings page (http://localhost:3000/devhacks/settings/admin). This menu is also lacking its hover effect.

image

app/assets/stylesheets/components/_buttons.scss
```diff
@@ -3,34 +3,39 @@ input[type='submit'],
...
-    box-shadow 0.25s ease-in-out;
+  transition: box-shadow 0.125s ease;
   user-select: none;
 
   &:hover,
   &:focus,
   &[aria-expanded='true'] {
-    color: $white;
-    box-shadow: 0 8px 16px rgba(0, 0, 0, 0.1875);
-    transform: scale(1.0625);
```
garyhtou, line 33: I see that we've removed the `transform: scale` on hover for buttons. I'd recommend also removing the effect from file upload buttons for consistency.

https://github.com/hackclub/bank/blob/57c5478221c1f726778090d8299996e3a6c1eebe/app/assets/stylesheets/components/_forms.scss#L110-L112

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/6ddd3617-9c36-40e7-90b5-5933e0083be8"/>|<video src="https://github.com/user-attachments/assets/6766e7f5-9935-4b0d-bb7a-527153989647"/>|

app/assets/stylesheets/components/_cards.scss
garyhtou, line 1: Here are a few things I've noticed about cards.

image

- The card's header and footer both have box shadows. But the body seems to be missing it.
- In the second card, the bottom horizontal line feels a bit loud when serving as a divider for the footer

<br/>

Here is another example of inconsistent box shadows. This specific screenshot is from the [Reimbursement Report show page](https://github.com/hackclub/bank/blob/b6eedc2fdfefaa01625955182d98d221c49b1ea0/app/views/reimbursement/reports/show.html.erb#L124-L129).
image

app/assets/stylesheets/components/_buttons.scss
```diff
@@ -63,70 +80,48 @@ input[type='submit'],
...
 
   &.bg-purple {
-    @include gradient(purple);
-  }
-
-  &.bg-ai {
-    @include gradient(ai);
+    background-color: map-get($palette, purple);
+    color: $white;
   }
```
garyhtou, line 103: `bg-purple` and `bg-muted` in dark mode are broken. They appear to be the same as a normal button without any `bg-*` class set.

image

As a reference, here's what they previously looked like:
image

app/assets/stylesheets/components/_buttons.scss
```diff
@@ -63,70 +80,48 @@ input[type='submit'],
...
+    background-color: map-get($palette, success);
+    color: darken(map-get($palette, success), 40%) !important;
   }
 
   &.bg-accent {
-    @include gradient(accent);
-    color: darken(map-get($palette, accent), 33.3%) !important;
+    background-color: map-get($palette, accent);
+    color: darken(map-get($palette, accent), 40%) !important;
   }
```
garyhtou: Looking at the set of button background colors, `success` and `accent` visually appear to be outliners.

image
image

_If you're noticing that purple isn't purple, see the comment above about broken color buttons in dark mode!_

The `bg-accent` also almost looks like it's disabled, especially in dark mode. I think it needs a stronger contrast against it's background.

<br/>

As a solution, I'd suggest making the `bg-success` and `bg-accent` button text match the others (`$white` in light mode, and `$smoke` in dark mode).

image

This change does not address the contrast issue that a couple of the button colors have. We might need to play around with the palette itself to address this. Let me know if you have any thoughts.

app/assets/stylesheets/components/_buttons.scss
```diff
@@ -44,12 +49,24 @@ input[type='submit'],
...
+      margin-left: 6px;
+      margin-right: -4px;
+    }
   }
 
   &.bg-primary {
-    @include gradient(primary);
+    background-color: map-get($palette, primary);
+    color: $white;
   }
```
garyhtou, line 74: The new `primary` button looks amazing! It draws attention, as a primary button should, and looks very "crisp".
image

app/helpers/donations_helper.rb
```diff
@@ -132,4 +132,8 @@ def donation_card_check_badge(check, donation = @donation)
     tag = inline_icon icon_name, size: 24
     content_tag(:span, class: "pr1 #{background} line-height-0 tooltipped tooltipped--w", 'aria-label': text) { tag }
   end
+
+  def donations_embed_html_code(event = @event)
+    html_escape "<iframe src='#{start_donation_donations_url event}' style='border:none;' name='donateFrame' scrolling='yes' frameborder='0' marginheight='0px' marginwidth='0px' height='512px' width='640px' allowfullscreen></iframe>".gsub("'", '"')
```
garyhtou, line 137: I'd recommend using a heredoc instead of gsub'ing.

https://www.rubyguides.com/2018/11/ruby-heredoc/

app/views/application/_logo.html.erb
```diff
@@ -26,16 +26,16 @@
...
     <a href="/brand/hcb-icon-icon-dark.png" download class="flex">
       <%= image_tag "logo-production-dark.svg", class: "mr1", size: 24 %>
-      Download PNG icon
+      Download dark PNG
     </a>
     <div class="menu__divider"></div>
     <a href="/branding">Brand guidelines</a>
     <div class="menu__divider"></div>
-    <a href="/" target="_blank">Open link in new tab</a>
+    <a href="/" target="_blank">Open home in new tab</a>
```
garyhtou: Good edits!

app/views/events/_title.html.erb
```diff
@@ -1,21 +1,30 @@
...
+      <p class="muted font-medium h5 pl-2 p-1 m0">Switch organizations</p>
+      <% events = current_user.events.not_hidden.excluding(@event).with_attached_logo %>
+      <% events.map do |event| %>
         <%= link_to event, title: event.name, class: "gap-2" do %>
           <% if event.logo.attached? %>
             <%= image_tag event.logo, height: 24, width: 24, class: "rounded" %>
+          <% elsif events.all? { |e| !e.logo.attached? } %>
+            <%# Show no placeholder %>
```
garyhtou, line 23: Amazing!

app/views/events/account_number.html.erb
```diff
@@ -135,10 +135,10 @@
     <% end %>
 
     <section class="modal modal--scroll bg-snow" data-behavior="modal" role="dialog" id="ach_form">
-      <%= modal_header "Account details" %>
+      <%= modal_header "Bank account details" %>
```
garyhtou, line 138: We avoid the word "bank" to prevent confusion around the legal standing of HCB.

```suggestion
      <%= modal_header "Account details" %>
```

app/views/events/donation_overview.html.erb
```diff
@@ -15,34 +15,45 @@
...
+      class: "tooltipped tooltipped--s", "aria-label": "Donations settings" %>
+  <% if @event.donation_page_enabled %>
+    <div data-controller="menu" data-menu-placement-value="bottom-end" class="line-height-0">
+      <button
+        type="button"
+        class="btn menu__toggle overflow-visible"
+        data-action="menu#toggle click@document->menu#close keydown@document->menu#keydown"
+        data-menu-target="toggle">
+        Donate page
+      </button>
```
garyhtou, line 30: This new button dropdown is amazing! I love it

app/helpers/donations_helper.rb
garyhtou: I'm now seeing that this is existing code and you had only moved it. Feel free to ignore this feedback if you'd like.

app/views/events/donation_overview.html.erb
```diff
@@ -108,10 +119,21 @@
...
+        type="button"
+        class="pointer pop mr2 align-middle tooltipped tooltipped--n"
+        aria-label="Copy HTML code"
+        data-controller="clipboard"
+        data-clipboard-text-value="<%= donations_embed_html_code %>"
+        data-action="click->clipboard#copy">
+        <%= inline_icon "copy", size: 28 %>
+      </button>
+    </div>
     <pre><%= html_escape "<iframe src='#{start_donation_donations_url @event}' style='border:none;' name='donateFrame' scrolling='yes' frameborder='0' marginheight='0px' marginwidth='0px' height='512px' width='512px' allowfullscreen></iframe>".gsub("'", '"') %></pre>
```
garyhtou: Please use the `donations_embed_html_code` helper you created to prevent the duplication of code. Thanks!

app/views/events/settings/_details.html.erb
```diff
@@ -170,7 +170,7 @@
       <%= inline_icon "important", size: 32, class: "mr1" %>
       <p class="bold mt0 mb0">If you change your URL, beware:</p>
     </div>
-    <ul class="mt1 mb2">
+    <ul class="mt1 mb2 slate">
```
garyhtou: See the comment above about `slate` (low contrast in dark mode).

app/views/events/show.html.erb
```diff
@@ -49,6 +50,26 @@
...
+          Export
+      </button>
+      <div class="menu__content menu__content--2" data-menu-target="content">
+        <%= link_to "All transactions (CSV)", transactions_exports_path(event: @event, format: :csv), target: :_blank %>
+        <%= link_to "All transactions (JSON)", transactions_exports_path(event: @event, format: :json), target: :_blank %>
+        <%= link_to "All transactions (Ledger)", transactions_exports_path(event: @event, format: :ledger), target: :_blank %>
+        <%= link_to "Monthly statements", event_statements_path(@event) %>
+      </div>
+    </div>
+  <% end %>
```
garyhtou: Great placement! White buttons can be very handy for drawing less attention!

Mel has voiced concerns about having white buttons on a white background. She's said that users already have a hard time finding buttons. Having buttons that fade into the background would make them more difficult for users to locate.

It would be best to limit the use of white buttons on white backgrounds.

This screenshot below is an example of an excellent use of a white button!
image

app/views/organizer_position_invites/_form.html.erb
```diff
@@ -75,5 +76,5 @@
     </div>
   <% end %>
 
-  <%= form.submit "Submit" %>
+  <%= form.submit invite.persisted? ? "Invite" : "Save" %>
```
garyhtou:
```suggestion
  <%= form.submit "Invite" %>
```
From my understanding, `invite.persisted?` will always be false here.

Also, I believe the two values are flipped because if `invite.persisted? == true`, this would indicate that it's already in the database and we're updating the record. Thus, we'd want it to say "Save". If `invite.persisted? == false`, then we'd be creating a new record and "Invite" makes sense.

app/assets/stylesheets/components/_dock.scss
garyhtou, line 1: I love this hover effect for the badge

image

garyhtou (changes requested):
Hi Lachlan, I love the direction that we're headed here. The small typography and card shadows feel very refined. I also love the new filterbar; I know @sampoder has also mentioned that too!

There's _a lot_ of good stuff here. I appreciated your patience as I worked through all of them and collected feedback from the team.

In my review, you'll find feedback from myself and the team. I've tried my best to colocate them with their respective lines of code and provide in-depth descriptions, screenshots, and possible solutions/alternatives. As always, I'd love to hear your thoughts.

Below are two general items that aren't tied to a specific line of code.

## Dark Mode borders

The team (@sampoder & @YodaLightsabr) has brought up that borders in dark mode can be a bit harsh. I also noticed this too especially on the My Reimbursements page.

| Before | image|
|-|-|
|After|image|

As for a solution, the first thing that comes to mind is darkening the borders for lower contrast. Another solution could be redesigning the layout of this page to use fewer bordered elements. However, pages with multiple inputs and tables are not uncommon. I'd love to hear any ideas you have here.

## Hack Club's roots
Sam has voiced that he misses the "fun" of the old interface. To summarize feedback from the entire team, this PR brings a professional, corporate-esque look to HCB; similar to [Column](https://column.com/)'s dashboard. The previous button styling tied HCB to Hack Club. The button gradients reflects the gradients we often use on sites like [hackclub.com](https://hackclub.com/). It's a core part of our and Hack Club's brand identity.

I personally see the benefits of the "professional" looking UI that this PR brings. On the flip side, I also understand the viewpoint of keeping our Hack Club roots. I was torn and met with Mel for her thoughts and a decision. She said:

>love the new button size and shape but prefer the old gradient style coloring - i think the new style is too harsh and i like that the colors we currently have are in line with the rest of hack club

I think Mel ultimately made this decision because over the past half year, we've emphasized focusing on our roots. Our priority is serving Hack Club and teenagers.

With Mel's decision in mind, the actionable steps here are:
1. Keep the new button size and shape
2. Bring back the button background gradients

---

I know this may be a lot of feedback. I can tell you had put a lot of thought into this PR! As always, I'd love to hear any input you have.

app/views/events/donation_overview.html.erb
lachlanjc: It would indeed! I'll let that be a follow-up polish item

app/views/events/g_suite_overview.html.erb
lachlanjc: I'm not able to run this page in local dev, but definitely a bug

app/views/organizer_position/spending/controls/index.html.erb
lachlanjc: Definitely unintentional, that's a perfect fix

app/assets/stylesheets/components/_buttons.scss
lachlanjc: Eek, we should delete those `bg-smoke` buttons; the darker text lowering contrast looks like a visual bug, and it's not communicating something distinct from our normal/muted buttons but looks disabled. I don't see `bg-purple` buttons used anywhere, let's go ahead & drop that variant entirely

app/assets/stylesheets/components/_dock.scss
lachlanjc: Oops yeah, that's a bug. Shouldn't have a ring on hover.

app/assets/stylesheets/components/_forms.scss
lachlanjc: Placeholder text doesn't need to meet contrast requirements, for exactly the reason you mentioned, it looks too prominent. Agreed on the dark mode though, let's reduce its visibility. Setting the color manually was to fix the visual bug of our greyscale using blue-tinted greys, and the browser using neutral greys, which looks bad visually side by side.

app/views/events/account_number.html.erb
lachlanjc: Checked with Dylan & we are in the clear legally on this—we do provide a bank account, even if we're not a bank.

app/views/events/donation_overview.html.erb
lachlanjc:
```suggestion
    <pre><%= donations_embed_html_code %></pre>
```

app/assets/stylesheets/components/_buttons.scss
lachlanjc: Oof that is visually loud! Ideally we'd drop those red buttons, the red communicates something being wrong or needing attention & this is a super neutral/chill action.

app/assets/stylesheets/components/_buttons.scss
lachlanjc: Great catch! Done

app/assets/stylesheets/components/_buttons.scss
garyhtou: +1 on deleting `bg-smoke` buttons!

I prefer to keep `bg-purple` around. We currently use purple to communicate AI-based features.

image
image

I can see a world where we use that purple button!

app/views/events/donation_overview.html.erb
garyhtou: re: copy indication as a follow-up item

Sounds good to me!

app/views/events/account_number.html.erb
garyhtou: Thanks for double-checking with Dylan. However, I just had a chat with Dylan yesterday to clear up confusion around this topic. I want to make sure we're on the same page.

I'm sure you're already aware of a good portion of this. Here's a quick rundown on the legal standing of HCB:

- HCB is legally not a bank; we don’t meet the regulatory or legal requirements. As a result, we legally can not call ourselves a bank. It's a fairly restricted word! This is the same with many fintech products including Chime, Cash App, and Betterment. It's the reason why both Betterment and Wealthfront call their products a "Cash Account".
- HCB partners with Member FDIC banks that provide banking services. We use Column, Increase, SVB, etc.
- The Hack Foundation/Zach is the legal owner of all the bank accounts with these banks. Fiscally sponsored organizations' money flows through Hack Club's bank accounts, but the bank accounts are fully owned by Hack Club. The Fiscal Sponsorship contract says we'll set aside a "fund" for each organization.
- We generate one bank **_account number_** designated for each organization. Under the hood, everyone’s money is sitting in the same bank accounts (owned by Hack Club). At the bank level, there is no separation of funds. That separation happens at the software level.
- From a legal perspective, we can likely say that "we provide banking-like services through our partner banks" (e.g. Column). However, here is what Zach has said on this topic:
	- “It’s important that users understand that they are fiscally sponsored by us, not that they have a bank account by us.”
	- "If an external party ever comes to us and says: 'Hey! You tricked your users into thinking they had bank accounts with you!', I want us to be able to respond with 'No, we make it extremely clear, in every part of our onboarding and in the app itself, that this is a fiscally sponsored fund - not a bank account'."

Because Zach is adamant about this, I need us to keep the modal header as "Account details" rather than "Bank account details".

Thanks for your understanding!

garyhtou:
Thanks for the reply and fixes, @lachlanjc. Feel free to ping me when this is ready for another review!

I know you're busy, so if you'd like us to handle the fixes, I'm also happy to do that. If you are interested, please give me a thumbs-up so that we're on the same page for the feedback above. We'll then edit this branch to address the feedback and get it ready to merge. Same thing for https://github.com/hackclub/hcb/pull/7773!

lachlanjc:
I've merged the branches & implemented some changes based on feedback! https://www.loom.com/share/c3bcc398ce40471f80047f7702b528f1

Please run with this from here—I'm sure there's some assorted bugs to catch etc. I haven't worked on the new mobile nav yet at all, only the desktop one.

dwahbe:
@lachlanjc so happy with this, thank you!! For some context, I met with @garyhtou and Mel and I'm going to start adding items here to finish this PR. @lachlanjc the team will finish this :)

- [ ] On mobile, quick actions should wrap and horizontally scroll, right now vertical
- [ ] When using a “card with a footer” the borders are not consistent, don’t show up on the bottom
- [ ] Break points are likely inconsistent, which causes Command K bar to show in mobile view. Happens when screen is 960px wide
Screenshot 2025-01-17 at 11 45 19 AM

- [ ] Mobile nav breaks, another breakpoint issue. 1024px is high range, low range is 768px
Screenshot 2025-01-17 at 11 45 41 AM
Screenshot 2025-01-17 at 11 45 38 AM

- [ ] Rename internal hcb transfers to HCB Book Transfer. This needs to be updated on the Transfers page and the quick action on the Overview page.
- [ ] Add “International wires” to transfer dropdown on Overview

PR #7743, opened 2024-08-27
Handle rejected Stripe card logos

sampoder:
Send an email when Stripe rejects an event's card


app/mailers/stripe_card_mailer.rb
```diff
@@ -21,6 +21,15 @@ def virtual_card_ordered
          subject: "New virtual HCB card (ending in #{@card.last4}) for #{@event.name}"
   end
 
+  def design_rejected
```
garyhtou: This should be in a StripeCard::PersonalizationDesignMailer

PR #7748, opened 2024-08-28
[Stripe Card Personalization Design] Copy edit to design rejection email

garyhtou:
Makes it a bit clearer which cards will use the standard HCB card design.


sampoder (approved):
LGTM!

PR #7749, opened 2024-08-28
Add Organization Picker

manuthecoder:
This PR is a split of the original #7512. It brings back @cjdenio's organization switcher feature flag 🎉

turns out, with some git trickery, splitting the two PR wasn't that hard...
cc @sampoder


sampoder (approved):
love it! i've made it red on hover & enabled it for everyone, this is good to go imo

PR #7766, opened 2024-08-28
[Google Workspace] Allow users to create Google Workspace email aliases through HCB

rluodev:
## Summary of the problem

Users could not manage aliases easily

## Describe your changes

Users can now manage aliases on HCB.

P.S. Fancy animations included :P

## images

Screenshot 2024-09-26 at 12 06 58 AM

Screenshot 2024-09-26 at 12 16 51 AM

Screenshot 2024-09-26 at 12 14 57 AM


app/models/g_suite_alias.rb
```diff
@@ -0,0 +1,41 @@
...
+class GSuiteAlias < ApplicationRecord
+  belongs_to :g_suite_account
+
+  before_create do |alias|
+    create_alias_with_google(alias.address)
+  end
+
+  before_destroy do |alias|
+    delete_alias_with_google(alias.address)
+  end
```
sampoder:
```suggestion
  before_create do
    create_alias_with_google(address)
  end

  before_destroy do
    delete_alias_with_google(address)
  end
```

rluodev:
not sure what happened here
```
In Gemfile:
  poppler was resolved to 4.2.2, which depends on
    cairo-gobject was resolved to 4.2.2, which depends on
      cairo
```

app/assets/stylesheets/application.scss
```diff
@@ -477,3 +477,7 @@ summary {
 .suggested_tag::after {
   transform: translateY(0px) translateX(-4px);
 }
+
+.rotated {
+  transform: rotate(180deg);
+}
```
garyhtou: Would tailwind's `rotate-180` class work?

app/controllers/g_suite_accounts_controller.rb
```diff
@@ -62,6 +62,10 @@ def destroy
 
     authorize @g_suite_account
 
+    @g_suite_account.g_suite_aliases.each do |al|
+      al.destroy
+    end
```
garyhtou:
```suggestion
    @g_suite_account.g_suite_aliases.destroy_all
```

app/controllers/g_suite_aliases_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    if @g_suite.accounts.map(&:address).include?(@address)
+      flash[:error] = "An account already exists with this email address."
+      return redirect_to event_g_suite_overview_path(event_id: @g_suite.event.slug)
+    end
+
+    if @g_suite.accounts.flat_map(&:g_suite_aliases).map(&:address).include?(@address)
+      flash[:error] = "An alias already exists with this email address."
+      return redirect_to event_g_suite_overview_path(event_id: @g_suite.event.slug)
+    end
```
garyhtou: Can we also add these checks as a validation in the alias model?

Even better would be to only use the model validations.

app/services/g_suite_alias_service/create.rb
garyhtou, line 1: Because these services are so tiny, I think we can go without them.

Let's drop the create and destroy services that we've created here, and op to inline the code where these services were being called.

garyhtou: We generally use services for large complex business logic.

app/services/partners/google/g_suite/create_user_alias.rb
```diff
@@ -0,0 +1,46 @@
...
+          unless Rails.env.production?
+            puts "☣️ In production, we would currently be creating an alias for a user on GW ☣️"
+            return
+          end
+
+          begin
+            Partners::Google::GSuite::User.new(email: @primary_email).run
+          rescue Google::Apis::ClientError
+            return false
+          end
```
garyhtou: I'm not familiar with GSuite's API. Are we re-creating the primary email address here? From my understanding, the user was already created, and now we're adding an alias to it. Also, how does this differ from `create_alias_with_google` in the model?

Maybe a comment here could help explain the process and the interactions with GSuite's API.

app/assets/stylesheets/application.scss
rluodev: I tried that but for some reason toggling that class wasn't working - I'll try it again tho

app/services/partners/google/g_suite/create_user_alias.rb
rluodev: Will do!

app/services/g_suite_alias_service/create.rb
rluodev: ok!

app/assets/stylesheets/application.scss
rluodev: yeah just tried it again and for some reason rotate-180 doesn't work

app/controllers/g_suite_aliases_controller.rb
rluodev: added as a validation but I also want to have the flash messages

app/models/g_suite_alias.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  private
+
+  def create_alias_with_google(alias_email)
+    Partners::Google::GSuite::CreateUserAlias.new(primary_email: g_suite_account.address, alias_email: address).run
+  end
+
+  def delete_alias_with_google(alias_email)
+    Partners::Google::GSuite::DeleteUserAlias.new(primary_email: g_suite_account.address, alias_email: address).run
+  end
```
garyhtou: the `alias_email` params are unused

app/services/partners/google/g_suite/create_user_alias.rb
rluodev: comments added!

app/models/g_suite_account.rb
```diff
@@ -78,7 +79,7 @@ def at_domain
   def reset_password!
     unless Rails.env.production?
       puts "☣️ In production, we would currently be syncing the GSuite account password reset ☣️"
-      return
+      # return
```
garyhtou: you'll want to discard these changes

app/models/g_suite_alias.rb
```diff
@@ -0,0 +1,53 @@
...
+#  index_g_suite_aliases_on_g_suite_account_id  (g_suite_account_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (g_suite_account_id => g_suite_accounts.id)
+#
+class GSuiteAlias < ApplicationRecord
+  belongs_to :g_suite_account
+
+  validates :address, uniqueness: { case_sensitive: false }, presence: true, if: -> { g_suite_account.g_suite_aliases.count < 30 }
```
garyhtou:
```suggestion
  validates :address, uniqueness: { case_sensitive: false }, presence: true
  validate :less_than_max_aliases

  def less_than_max_aliases
    if g_suite_account.g_suite_aliases.count > 30
      errors.add(:g_suite_account, "has too many aliases")
    end
  end
```

app/models/g_suite_alias.rb
```diff
@@ -0,0 +1,53 @@
...
+
```
garyhtou:
```suggestion
validate :address_unique_against_accounts

def address_unique_against_accounts
  if g_suite.accounts.where(address:).exists?
    errors.add(:address, "already exists")
  end
end
```

app/models/g_suite_alias.rb
```diff
@@ -0,0 +1,53 @@
...
+# Indexes
+#
+#  index_g_suite_aliases_on_g_suite_account_id  (g_suite_account_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (g_suite_account_id => g_suite_accounts.id)
+#
+class GSuiteAlias < ApplicationRecord
+  belongs_to :g_suite_account
```
garyhtou:
```suggestion
  belongs_to :g_suite_account
  has_one :g_suite, through: :g_suite_account
```

app/controllers/g_suite_aliases_controller.rb
```diff
@@ -0,0 +1,62 @@
...
```
garyhtou: Instead of rewriting error messages, we can pull them from the validation errors.

example: https://github.com/hackclub/bank/blob/2eb8dcb1c7d1daf49c6ea8f67436219737fba677/app/controllers/hcb_codes_controller.rb#L102-L107

app/controllers/g_suite_accounts_controller.rb
```diff
@@ -62,6 +62,8 @@ def destroy
 
     authorize @g_suite_account
 
+    @g_suite_account.g_suite_aliases.destroy_all
+
```
garyhtou:
```suggestion
```

I believe this line is not necessary since we have `depenent: :destory` configured inside the GSuiteAccount model for the `g_suite_aliases` association.

app/controllers/g_suite_aliases_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+class GSuiteAliasesController < ApplicationController
+  def create
+    @g_suite_account = GSuiteAccount.find(params[:g_suite_account_id])
+    @g_suite = @g_suite_account.g_suite
+    @g_suite_domain = @g_suite.domain
+    @address = "#{params[:address]}@#{@g_suite_domain}"
+
+    authorize @g_suite_account, policy_class: GSuiteAliasPolicy
+
+    @g_suite_alias = @g_suite_account.g_suite_aliases.new(address: @address)
```
garyhtou:
```suggestion
    @g_suite_alias = @g_suite_account.g_suite_aliases.new(address: @address)
    authorize @g_suite_alias
```

app/models/g_suite_alias.rb
```diff
@@ -0,0 +1,67 @@
...
+  has_one :g_suite, through: :g_suite_account
```
garyhtou:
```suggestion
  has_one :g_suite, through: :g_suite_account
  has_one :event, through: :g_suite
```

We are adding this after updating the GSuiteAliasController to pass a GSuiteAlias to the GSuiteAliasPolicy.

app/controllers/g_suite_aliases_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    if @g_suite_alias.save!
```
garyhtou:
```suggestion
    if @g_suite_alias.save
```

run non-bang version since it's inside a `if/else`.

app/controllers/g_suite_aliases_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    redirect_to event_g_suite_overview_path(event_id: @g_suite.event.slug)
+  end
+
+  def destroy
+    @g_suite_alias = GSuiteAlias.find(params[:g_suite_alias_id])
+    @g_suite_account = @g_suite_alias.g_suite_account
+    @g_suite = @g_suite_account.g_suite
+
+    authorize @g_suite_account, policy_class: GSuiteAliasPolicy
```
garyhtou:
```suggestion
    authorize @g_suite_alias
```

app/controllers/g_suite_aliases_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      flash[:error] = @g_suite_alias.errors.full_messages.to_sentence
+    end
+
+    redirect_to event_g_suite_overview_path(event_id: @g_suite.event.slug)
+  end
+
+  def destroy
+    @g_suite_alias = GSuiteAlias.find(params[:g_suite_alias_id])
+    @g_suite_account = @g_suite_alias.g_suite_account
+    @g_suite = @g_suite_account.g_suite
```
garyhtou:
```suggestion
```

app/controllers/g_suite_aliases_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+
+      flash[:success] = "Google Workspace alias deleted."
+    rescue => e
+      notify_airbrake(e)
+
+      flash[:error] = e.message
+    end
+
+    redirect_to event_g_suite_overview_path(event_id: @g_suite.event.slug)
```
garyhtou:
```suggestion
    redirect_to event_g_suite_overview_path(event_id: @g_suite_alias.event.slug)
```

app/controllers/g_suite_aliases_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    @g_suite_alias = GSuiteAlias.find(params[:g_suite_alias_id])
+    @g_suite_account = @g_suite_alias.g_suite_account
+    @g_suite = @g_suite_account.g_suite
+
+    authorize @g_suite_account, policy_class: GSuiteAliasPolicy
+
+    begin
+      ActiveRecord::Base.transaction do
+        @g_suite_alias.destroy
+      end
```
garyhtou:
```suggestion
      @g_suite_alias.destroy!
```

we can drop the transaction since we're only making one commit.

garyhtou: I'm also adding a bang so that it raises and error and get caught

app/models/g_suite_alias.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  validates :address, uniqueness: { case_sensitive: false }, presence: true
+  validate :fewer_than_max_aliases
```
garyhtou:
```suggestion
  validate :fewer_than_max_aliases, on: :create
```

app/services/partners/google/g_suite/create_user_alias.rb
```diff
@@ -0,0 +1,48 @@
...
+          unless Rails.env.production?
+            puts "☣️ In production, we would currently be creating an alias for a user on GW ☣️"
+            return
+          end
+
+          begin
+            # Checks that a user exists with the provided primary email
+            Partners::Google::GSuite::User.new(email: @primary_email).run
+          rescue ::Google::Apis::ClientError
+            throw :abort
```
garyhtou:
```suggestion
            # Raises if user was not found
            throw :abort
```

app/services/partners/google/g_suite/delete_user_alias.rb
```diff
@@ -0,0 +1,40 @@
...
+        def run
+          unless Rails.env.production?
+            puts "☣️ In production, we would currently be deleting an alias for a user on GW ☣️"
+            return
+          end
+
+          begin
+            # Checks that a user exists with the provided primary email
+            Partners::Google::GSuite::User.new(email: @primary_email).run
+          rescue ::Google::Apis::ClientError
```
garyhtou:
```suggestion
          # Raises if users is not found
          rescue ::Google::Apis::ClientError
```

config/routes.rb
```diff
@@ -306,6 +306,8 @@
     put "toggle_suspension"
   end
 
+  resources :g_suite_aliases, only: [:create, :destroy]
```
garyhtou: we probably want to nest this inside the `g_suites` resource

garyhtou: I'd recommend looking into "shallow routing"

garyhtou:
^ comments from review call with @rluodev!

config/routes.rb
rluodev: done! (i think)

app/services/partners/google/g_suite/delete_user_alias.rb
```diff
@@ -21,6 +21,7 @@ def run
           begin
             # Checks that a user exists with the provided primary email
             Partners::Google::GSuite::User.new(email: @primary_email).run
+          # Raises if users is not found
```
garyhtou:
```suggestion
            # Raises if users is not found
```

PR #7767, opened 2024-08-29
De-front CPTs when expense payouts are reversed

sampoder:
At least I'm not crazy this time.


app/models/reimbursement/expense_payout.rb
```diff
@@ -92,6 +92,8 @@ def reverse!
 
         mark_reversed!
 
+        canonical_pending_transaction.update(fronted: false)
```
garyhtou: Should we do a declined mapping here?

https://github.com/hackclub/bank/blob/fe682c6d2f80b2dd2b04a53f3496769fa1a7ec9c/app/models/canonical_pending_declined_mapping.rb#L20-L19

It's usually used for stripe cards, I'm not sure if it makes sense here

PR #7105, opened 2024-07-13
Redesign home page

manuthecoder:
# See #7092 
> ### Describe your goals
> 
> # Motivation
> Upon reorganizing the current events homepage in #7090, I identified an opportunity for HCB Events to benefit from a dedicated events homepage.
> 
> # What's this about!?
> The redesigned homepage will feature an "**at a glance**" view of events, replacing the current list of recent transactions. This new page will include information about team members, cards, and a predefined number of recent transactions. Widgets will offer links to transactions, cards, or entire pages. 
> 
> Future enhancements could allow users to reorder widgets or create custom "filter" widgets.
> 
> # Key Changes:
> - The transparency banner will be moved from the transactions page to the homepage.
> - The transaction history chart from the transactions page will become a separate widget on the homepage.
> 
> # Proposed Widgets:
> - Balance over time chart
> - Recent transactions table
> - Team members
> - Recent event activity (based on @sampoder's activity feature flag)
> - Cards (sorted by most recent transaction)
> - Recent donations
> - Overview of recent reimbursements
> 
> > [!IMPORTANT]  
> > **It is crucial to distinguish this from the breakdown page, which provides a detailed event report. The homepage will be minimalist, offering a concise "at a glance" overview.**
> 
> # Preview
> Frame 1
> 
> 
> **See it on Figma! https://www.figma.com/design/YTFyYUMwu3MbTZtVwxZp2y/Untitled?node-id=0-1&t=48xUtMQAaXL30ubf-1**
>

PR #7773, opened 2024-08-29
[UI] Add new icons

lachlanjc:
## Summary of the problem

I've designed a bunch of new icons I've added to the codebase!

## Describe your changes

On the nav, a few new icons:

- Updated the arrows on the Financials & Transfers to be more elegant
- New perks icon that's more abstract since we don't give you merch

- New team icon that communicates multiple people instead of one special person
- New settings icon that's distinct from the user settings page
- New reimbursements icon (vaguely in the shape of an R) that communicates the flow of money back & forth
- Since receipts are physically organized with paperclips, and they're also generic "attachments," using the paperclip icon for attachments
- Using the "money document" icon for invoices, since they are documents and no briefcases are involved

image


app/views/events/_nav.html.erb
```diff
@@ -12,7 +12,7 @@
     <% end %>
   </div>
   <nav class="dock dock--desktop mt2 font-medium">
-    <%= dock_item "Transactions",
```
sampoder: Let's keep Transactions imo, it's more direct. I know that page's header is Financials but that's going to change to "Transactions" after we replace the home page.

app/views/events/_nav.html.erb
```diff
@@ -83,12 +83,12 @@
     <%= dock_item "Team",
       event_team_path(event_id: @event.slug),
       tooltip: "Manage your team",
-      icon: "leader",
+      icon: "people-2",
```
sampoder: Looks sweet!

app/views/events/_nav.html.erb
```diff
@@ -83,12 +83,12 @@
...
       selected: local_assigns[:selected] == :team %>
     <%= dock_item "Perks",
       event_promotions_path(event_id: @event.slug),
       tooltip: (!policy(@event).promotions? ? "Your account isn't eligble for receive promos & discounts" : "Receive promos & discounts"),
-      icon: "shirt",
+      icon: "perks",
```
sampoder: Nice change

app/views/events/_nav.html.erb
```diff
@@ -36,14 +36,14 @@
...
         event_donation_overview_path(event_id: @event.slug),
-      tooltip: "Receive donations online",
+      tooltip: "Receive online donations",
       icon: "support",
       data: { tour_step: "donations" },
       selected: local_assigns[:selected] == :donations if @event.approved? %>
     <%= dock_item "Invoices",
       event_invoices_path(event_id: @event.slug),
-      tooltip: "Collect sponsor payments",
-      icon: "briefcase",
```
sampoder: We've chatted about this before in https://github.com/hackclub/hcb/issues/7324, the consensus seemed to be keeping briefcase even if it wasn't ideal with the goal of finding something better

sampoder (commented):
Sorry looks like these comments never posted!

app/views/events/_nav.html.erb
lachlanjc: I'd love to have a more exact icon, but I've struggled to design one that doesn't resemble Documents or Receipts. The verb with an invoice is to send—you send the invoice as a request, the recipient sends money back—so I think it's a far more appropriate choice than my previous selection of the briefcase icon, which doesn't correspond to any part of invoices. I've clarified the tooltip to make sure folks understand what these are.

sampoder: +1 on struggling to think of a more exact icon. There are a fair few places where you can send something on HCB though, for example, sending an invite to join a team. I think we should avoid that ambiguity, at least the briefcase icon was unique to invoices. That made it work for places like the activity log or the command bar.

If we want to continue this discussion, is it cool if we pull it into a separate PR that way it doesn't block on getting everything else merged?

garyhtou (commented):
+1 on @sampoder's feedback. Feel free to merge after the feedback is addressed!

Thanks for reviewing @sampoder; we're aligned here.

garyhtou:
@lachlanjc I love all these new icons you have. They're going to come in very handy!

app/views/events/_nav.html.erb
lachlanjc: I worked on a wide range of icons for invoices, most of them I didn't love. Here's the set, in various states of completion:
CleanShot 2024-08-31 at 13 54 10@2x

Adding wind to the send icon I think clarifies that you're doing the sending, so I'm pretty happy with this new set!
CleanShot 2024-08-31 at 13 53 08@2x

sampoder: Screenshot 2024-08-31 at 1 25 36 PM

What is we used this icon but with a dollar sign on the document?

lachlanjc: Unfortunately it's too visually busy at small sizes. I've dropped the envelope part, updated screenshot above!

app/views/events/_nav.html.erb
sampoder: Hey, uhh, I was working on getting this merged into production but this is one of a couple things blocking, can you please reapply my commit?

app/views/events/_nav.html.erb
sampoder: This is just the old receipts icon isn't it? I'm on the fence, @YodaLightsabr @malted what do you all think? My hesitation is that we're switching around things we've been using for a while but I do think the changes make sense.

lachlanjc: Yep, it's not ideal to be shuffling icons around, but I do think the new set is a lot clearer overall. I'm not over the moon with this invoices icon, but it does communicate "money document", and luckily there's a clear legend of which icons are which with the labeled sidebar items. My guess is after a day with the new set it'll be pretty intuitive for customers.

app/controllers/events_controller.rb
garyhtou, line 1: Just as an FYI, I found that these `icon` keys are not referenced anywhere. I've created a ticket to remove them: https://github.com/hackclub/hcb/issues/8069

For the time being, we can keep your changes here; your call.

app/javascript/components/command_bar/actions.js
```diff
@@ -446,7 +447,7 @@ export const adminActions = adminUrls => [
     id: 'admin_tool_31',
     section: 'Admin Tools',
     priority: Priority.HIGH,
-    name: 'Check Deposits',
+    name: 'Deposit checks',
```
garyhtou: I like that we're updating the capitalization here.

I will mention that across the codebase, we use the term "Check deposit(s)" (as a noun). This would be the first instance where we're calling it "Deposit checks".

For consistency, I would like for us to keep this as "Check deposits". This change can happen in this PR or as a follow-up PR.

app/views/hcb_codes/_icon.html.erb
garyhtou, line 1: Can we please update the icon for Reimbursement on line 28 from `attachment` to `reimbursement`?

https://github.com/hackclub/bank/blob/5556509eb9ac642b76b03fa58f3ccc31b0516763/app/views/hcb_codes/_icon.html.erb#L28

app/views/events/_nav.html.erb
```diff
@@ -12,10 +12,10 @@
     <% end %>
   </div>
   <nav class="dock dock--desktop mt2 font-medium">
-    <%= dock_item "Transactions",
+    <%= dock_item "Financials",
```
garyhtou: With the current content on the Transactions page, I can see why we'd want to rename this page to "Financials".

**I want to take a step back and provide some context on how this PR fits into our roadmap.**

In the review comment above, I mentioned how we'd like to ship this PR with our **UI3** launch. This **UI3** launch includes a new [**Home** page](https://github.com/hackclub/hcb/pull/7105) for organizations. Here's a sneak peek.

| | Home | Transactions |
|-|-|-|
| **Before** | nonexistent (was Transactions 👉) |  |
| **After** (https://github.com/hackclub/hcb/pull/7105) |  The new **Home** page provides glanceable, widget-based information on your organization.<br/><br/>Designed to be suitable for both Transparency Mode and organizers, it offers a one-stop shop for understanding the current state of an organization's finances. Chips have been added to provide quick access for frequent actions (the final version will draw from https://github.com/hackclub/hcb/pull/7839). |  The **Transactions** page has been slimed down to provide you no-frills access to an organization's ledger.<br/><br/>Search and filter to your heart's desire. The balance graph and team members have been moved to the **Home** page. <br/><br/>**This page now only contains an index of transactions.** |

<br/>

With that in mind, given that this PR will be shipped alongside the new **Home** page, I believe it's best to keep the Transactions page named "Transactions".

app/views/events/_nav.html.erb
```diff
@@ -12,10 +12,10 @@
...
       event_path(id: @event.slug),
       tooltip: "View your transactions",
-      icon: "transactions",
+      icon: "bank-account",
```
garyhtou: I want to mention that with this icon change, this page will be sharing the same `bank-account` icon as the **Account numbers** page.

I think that's not a big detail, however, I do want to make sure that we're aware of this similarity. I'm happy for us to use the `bank-account` icon for this page.

There is additional discussion about removing the **Account numbers** page entirely. That may help solve the duplicate icon.

app/assets/images/icons/plus.svg
garyhtou, line 1: This is a very much-needed icon! Thank you!

This PR renames the previous image `plus` icon to image `plus-box`.

I like how you did not update the existing usages to use `plus-box`. `plus-box` was a bit heavy for our typical use cases (buttons for adding/creating something). The (new) `plus` does a better job of communicating the same purpose without the added weight/busyness.

However, I did find one location where I prefer `plus-box` over `plus`. We have a banner that provides instructions for installing the PWA app. In the screenshots below, let's take a look at the plus in the red banner.

| Before | After|
|-|-|
|  |  |

The goal of the `+` plus is to replicate Apple's icon for "Add to Home Screen". Looking at the screenshot below, we can see Apple's icon is most similar to `plus-box`.

image

I would recommend that we switch this usage from `plus` to `plus-box`.
https://github.com/hackclub/bank/blob/89a7d64e4367f83265efb8b16dba5e3a4b7ff6c2/app/views/users/_nav.html.erb#L58

Other than that, this is great!

garyhtou (commented):
I've said this before and I'll say it again, I ABSOLUTELY LOVE all the new icons you have! They're super handy 🫶

Overall, there are some amazing changes here. I especially love the icons you picked for perks and teams. Those do a much better job of representing their respective pages. Also, thanks for the detailed PR description explaining the motivation behind each change.

I've left some very minor feedback. If you'd like to take a minute or two to address them, I'd appreciate it. As always, I'm happy to hear your thoughts on any feedback I left.

Also, I know you're in school and are very busy. So, if you'd prefer, I'd be happy to handle the feedback on your behalf. Let me know what would work best for you.

---

As for getting this shipped, Dylan may have already mentioned to you that we plan to include this in a large **UI3** launch. This PR, along with a couple of others, will be announced in a changelog post and shipped at the same time. The goal of **UI3** is to provide users a visual refresh and to make "breaking changes" to core concepts. (It also has the nice benefit of generating excitement and demonstrating forward progress to our users.)

Since this PR repurposes heavily used icons, we thought it was very fitting for the **UI3** launch.

| | Before | After|
|-|-|-|
|Reimbursement| image | image |
|Receipt| image | image |
|Invoice| image | image |

_The current Reimbursement icon becomes the Receipt icon. And the current Receipt icon becomes the Invoice icon._

With the large **UI3** launch/announcement, users will be mentally prepared for a new look. This will hopefully minimize any confusion from us switching the icons around!

We're headed in a great direction and I appreciate the work you're doing. As always, I'd be happy to hear any thoughts you have.

app/assets/images/icons/plus.svg
lachlanjc: Great catch, reverted!

app/views/events/_nav.html.erb
lachlanjc: Yep, that's fine; #7839 redesigns the account numbers section & removes that page!

garyhtou:
Hi @lachlanjc! I just wanted to check in to see if I may make the above-mentioned edits to this branch. I'd love to get this merged — same goes for [#7741](https://github.com/hackclub/hcb/pull/7741#issuecomment-2427655001)!

app/javascript/components/command_bar/actions.js
lachlanjc:
```suggestion
    name: 'Check deposits',
```

lachlanjc:
I think we're good to go with this now @garyhtou! Sorry for the delay, HCB has been at the bottom of my to-do list during a very busy midterms season.

garyhtou:
Thanks @lachlanjc! I totally get it— midterms are important! I'll do a final review pass on this and get it merged into our UI3 branch. From there, we'll test it with the rest of the UI3 PRs and prepare the changelog post.

Feel free to ping me once [#7741](https://github.com/hackclub/hcb/pull/7741#issuecomment-2427655001) is ready as well!

garyhtou (approved):
Great! Thank you. I'm now merging this into the UI3 branch

PR #7782, opened 2024-08-30
add skull emoji reaction option

24c02:
i'm mostly kidding but on the off chance that y'all would be crazy enough to merge this:


app/models/comment/reaction.rb
```diff
@@ -30,7 +30,7 @@ class Reaction < ApplicationRecord
     validates :emoji, presence: true
     validates :emoji, uniqueness: { scope: [:reactor_id, :comment_id], message: "has already been used for this comment" }
 
-    EMOJIS = %w[👍 👎 😄 🎉 😔 ❤️ 🚀 👀].freeze
+    EMOJIS = %w[👍 👎 😄 🎉 😔 ❤️ 🚀 👀 💀].freeze
```
garyhtou: half the deploys have a 💀 /j

```suggestion
    EMOJIS = (%w[👍 👎 😄 🎉 😔 ❤️ 🚀 👀] + (rand > 0.5 ? ["💀"] : [])).freeze
```

24c02: Screenshot 2024-08-29 at 10 38 04 PM

github refuses to let me respond appropriately

garyhtou: LOL

sampoder (approved):
lol let's do it... at least for a bit and see if it's used

PR #7788, opened 2024-08-30
Switch to `fixture_paths`

sampoder:
Closes https://github.com/hackclub/hcb/issues/7783


spec/rails_helper.rb
```diff
@@ -45,7 +45,7 @@
   config.include ActiveJob::TestHelper
 
   # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
-  config.fixture_path = Rails.root.join("spec/fixtures").to_s
+  config.fixture_paths << Rails.root.join("spec/fixtures").to_s
```
sampoder:
```suggestion
  config.fixture_paths = [Rails.root.join("spec/fixtures").to_s]
```

PR #7790, opened 2024-08-30
Decline CPTs associated with reversed `ExpensePayout`s

sampoder:


app/models/reimbursement/expense_payout.rb
```diff
@@ -93,6 +93,7 @@ def reverse!
         mark_reversed!
 
         canonical_pending_transaction.update(fronted: false)
```
garyhtou: I think because we're declining it, there might no longer be a need to change fronted to false.

this is how reject disbursement currently function

```sql
SELECT * FROM "canonical_pending_transactions" cpt
join "canonical_pending_declined_mappings" cpdm on cpdm.canonical_pending_transaction_id = cpt.id
where cpt.fronted = true
LIMIT 100
```

https://github.com/hackclub/bank/blob/459a0fb99077cb383cb36761b9c5167c8e3997d2/app/models/disbursement.rb#L151

PR #7809, opened 2024-09-01
Add preview for docx files

Ved235:


Ved235:
closes #7733

sampoder:
Hey, I tried this in staging and it doesn't seem to be working:

Screenshot 2024-09-01 at 10 07 20 AM

I've tried adding `libreoffice` to our `Aptfile` to fix that.

Ved235:
Could we talk in slack about this, writing comments here seems kinda slow!

sampoder:
Heroku is refusing to deploy this now that we're installing `libreoffice` (our Heroku doesn't use Docker to deploy):

```
 !     Compiled slug size: 897.8M is too large (max is 500M).
 !     See: http://devcenter.heroku.com/articles/slug-size
 !     Push failed
```

I think we're going to have to find another way or skip out on doing this sadly.

Ved235:
Ok, I am going to try to look for another way.

sampoder:
Yeah fs! Sounds good

PR #7812, opened 2024-09-01
Add creating donations to the API (WIP)

sampoder:


app/controllers/api/v4/donations_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+          name: "Stripey McStripeface",
+          amount: params[:amount_cents],
+          event_id: params[:event_id]
+        })
+
+        authorize @donation
+        
+        @donation.save
+
+        render "show"
```
sampoder: Using Stripe? Set the metadata's "donation" key to this

PR #7813, opened 2024-09-02
Fix command bar guide colors

Ved235:
## The problem
Initially the guide text colors in the command bar didn't switch correctly with change in theme causing difficulty in reading text.

## Changes

**Initial dark theme view**
image

**Final dark theme view**
image

_Working for light theme view as well_
image
image


app/javascript/components/command_bar/search/guide.js
```diff
@@ -1,11 +1,12 @@
-/* eslint react/prop-types:0 */
```
sampoder: Let's not remove this if possible

app/javascript/components/command_bar/search/guide.js
```diff
@@ -1,11 +1,12 @@
...
-
 import React from 'react'
 
 export function Guide() {
```
sampoder: Would it be possible to wrap this in one span? Does that work?

Ved235:
Hi, I made a new commit to resolve these.

sampoder (dismissed):
LGTM!

sampoder:
Thanks helping fix this, I really appreciate that!

PR #7824, opened 2024-09-04
Block public reimbursements for events on plans with reimbursements disabled

sampoder:


app/models/event.rb
```diff
@@ -748,6 +748,10 @@ def config
   def donation_page_available?
     donation_page_enabled && plan.donations_enabled?
   end
+
```
```suggestion

```

PR #7840, opened 2024-09-05
Hide extra `<hr />` in transparency mode settings

manuthecoder:
This PR addresses #7659 and removes the extra divider element in the transparency mode settings. 

| Before | After |
|--------|--------|
| image| image|


app/views/events/settings/_details.html.erb
```diff
@@ -118,7 +118,6 @@
         </ul>
       </li>
     </ul>
-    <div class="menu__divider mb1"></div>
     <div data-accordion-target="content">
```
sampoder: Shouldn't we move it to be inside of the content div?

manu-gurudath:
done! I also removed the margin from the divider since it was causing weird layout shifts because of jquery

sampoder (approved):
LGTM!

PR #7844, opened 2024-09-06
Paginate document download history

manuthecoder:
This PR addresses #7765 and paginates the document download history. It will show 20 items per page, with the most recent download first.


app/controllers/documents_controller.rb
```diff
@@ -131,7 +131,12 @@ def document_params
   end
 
   def set_document
+    @downloads = []
```
sampoder: We can drop this line
```suggestion
```

sampoder (approved):
LGTM! Thanks!

PR #7851, opened 2024-09-08
[Organization Ledger] Filter UI

manuthecoder:
This PR addresses @dwahbe's #6688 and adds a filter UI. It so adds a helper function called `upsert_query_params` which eliminates hard-to-read string interpolation which is currently used.

| | | 
|--------|--------|
| image | image |

### Below are some previews of what the final result looks like!
|  |  |
|--------|--------|
| image | image| 
| image | image |

#### Mobile
Tested on mobile!


sampoder:
The new helper is neat, thanks for adding that! (And I guess this might also be a question for @dwahbe) What's the motivation behind pulling this out a menu? I'm worried that this new design takes up too much space, especially because of the large buttons.

dwahbe:
This is great, awesome stuff @manuthecoder! What I like about this is you can now easily stack filters. For example, you might want to see a specific type of transaction mixed in with a date range.

For spacing, I agree that the list of transactions would be a bit low however, once the Overview page ships there won't be the money graph or team section. For the other pages, it doesn't look to be taking too much space to me.

sampoder:
Can we find a way of not using the large buttons? Especially for clearing.

manu-gurudath:
> I'm worried that this new design takes up too much space, especially because of the large buttons.

Should I make the buttons icon-only on mobile?

sampoder:
Yeah that's a good idea, maybe icon only on desktop as well

sampoder:
And we can use the circular icons (like we have for filter on cards in production right now) instead of the big ones

app/helpers/application_helper.rb
```diff
@@ -3,6 +3,14 @@
 module ApplicationHelper
   include ActionView::Helpers
 
+  def upsert_query_params(new_params)
+    params = request.query_parameters || {}
+    {
+      **params,
+      **new_params
+    }.compact
+  end
```
garyhtou:
```suggestion
  def upsert_query_params(**new_params)
    params = request.query_parameters || {}
    {
      **params,
      **new_params
    }.compact
  end
```

this means that usage switches from `upsert_query_params({a: 1})` to `upsert_query_params(a: 1)`

manu-gurudath:
image
done! ✅

manu-gurudath:
I just found out you can actually request reviews without clicking on the "🔁" icon

app/helpers/application_helper.rb
```diff
@@ -3,6 +3,14 @@
 module ApplicationHelper
   include ActionView::Helpers
 
+  def upsert_query_params(**new_params)
+    params = request.query_parameters || {}
+    {
+      **params,
+      **new_params
+    }.compact
```
garyhtou:
```suggestion
    params.merge(new_params).compact
```

app/helpers/application_helper.rb
```diff
@@ -3,6 +3,14 @@
 module ApplicationHelper
   include ActionView::Helpers
 
+  def upsert_query_params(**new_params)
```
garyhtou: Because of `compact`, it will remove keys with a `nil` value. Sometimes code will look for just the key and does not care about the value. As a result, we may see code that sets a query param with a `nil` value. To prevent code like that from breaking, we should:

Only remove the `nil` values if the `nil` came from `new_params`.

garyhtou (commented):
image
Misaligned right end

---

image

When you have a filter selected, then click on the filter button. None of the tabs are selected by default which results in an empty menu.

---

image

Add "Clear filter" tooltip

app/views/events/_filter.html.erb
```diff
@@ -127,6 +127,14 @@
             </div>
           <% end %>
         </div>
+        <div data-tabs-target="tab" id="receipts">
+          <div class="flex items-center">
+            <%= link_to("Missing receipts", upsert_query_params(user: params[:user], q: params[:q], missing_receipts: true), class: "flex-auto menu__action #{"menu__action--active" if @missing_receipts}") %>
+          </div>
+          <div>
+            <%= link_to("All transactions", upsert_query_params(user: params[:user], q: params[:q]), class: "flex-auto menu__action #{"menu__action--active" unless @missing_receipts}") %>
```
garyhtou: If we're upserting params, I think we don't need to pass `user` and `q` again

garyhtou:
Move filter button to the top when no filters are applied
- Transaction page: right of search bar
- Cards page: Top right next to other buttons.

garyhtou (approved):

PR #7852, opened 2024-09-08
update invoice max card amount text

mrhappyma:
## Summary of the problem
stripe invoices show up ach-only for >= $10,000 (at least in test mode), not just > (like the current text says).

## Describe your changes
updates the warning message to say this, and hides the card payment icon if the amount is $10,000 or more


sampoder (approved):
Good catch, thank you!

PR #7887, opened 2024-09-12
[YellowPages] Instrument missing network ID events

malted:
Replaces https://github.com/hackclub/hcb/pull/7886

Since the ahoy event has been removed we need some way to track this!


sampoder (approved):
Thanks!

PR #7888, opened 2024-09-12
Send donation notification on every donation

MatthewStanciu:
## Summary of the problem

I got a couple big donations, but since they weren't my org's first donations and didn't include a message, I didn't get notified. But they were big donations, >$200 each! And I want to be notified of them.

Slack thread: https://hackclub.slack.com/archives/CN523HLKW/p1726010191784519

## Describe your changes

This PR:

- Adds a new view for a generic donation email
- Sends that email if it's not the first donation or doesn't include a message
- Renames the method from `send_payment_notification_if_needed` to `send_donation_notification`
- Adds a test that ensures the correct emails are sent

This change would make it so that every org receives an email for every donation. I want this, but I can imagine larger orgs that get lots of donations would want to opt out. It may be good to pair this change with a toggle in the settings, either/or:

- Only send donation emails if the donation is larger than X
- Only send donation emails if it's my org's first donation or if the donor includes a message
- Opt out of donation emails


MatthewStanciu:
Test failure: https://github.com/hackclub/hcb/actions/runs/10823904016/job/30030197009?pr=7888#step:13:109

Test definition: https://github.com/hackclub/hcb/blob/ca63077bae859ffc280475890f6039a554f12343/spec/models/donation_spec.rb#L23

Question:

> \# only runs when status becomes succeeded, should not run on delete.

Does this mean the third email should not send, since it's being deleted at the end? My guess is no because we want to send when a donation is marked succeeded, and donation 3 is marked succeeded. But not 100% sure based on the wording of the comment.

If the answer is no, I will update the test definition. Just wanted to double-check first.

sampoder:
We can drop that test now @MatthewStanciu

MatthewStanciu:
@sampoder I removed that test fixture and replaced the one before it with one that tests that the correct emails are being sent

sampoder (approved):
LGTM; thank you so much Matthew!

PR #7922, opened 2024-09-19
Use preferred names on Stripe cards instead of full names

sampoder:
Closes https://github.com/hackclub/hcb/issues/7898


sampoder:
I don't know how your legal name is less characters than your preferred name but I'll keep it in there..

PR #7983, opened 2024-09-24
Update `estimated_fee_cents_usd` to $25 & a constant

sampoder:
It's $25!


app/views/wires/new.html.erb
```diff
@@ -13,7 +13,7 @@
 <%= render "callout",
     type: "info",
     title: "Important information about wire transfers" do %>
-    <span class="font-bold">Each wire transfer incurs a $20 flat fee</span>, this will be added on to the amount entered below and is passed on from our partner bank. Additionally, <span class="font-bold">currency exchange rates may fluctuate</span> between a wire being requested and a wire being sent.
+    <span class="font-bold">Each wire transfer incurs a <%= render_money Wire.new.estimated_fee_cents_usd %> flat fee</span>, this will be added on to the amount entered below and is passed on from our partner bank. Additionally, <span class="font-bold">currency exchange rates may fluctuate</span> between a wire being requested and a wire being sent.
```
garyhtou: if this is static, should it be a class method or a constant?

```ruby
Wire::ESTIMATED_FEE_CENTS_USD
# or
Wire.estimated_fee_cents_usd
```

calling `new` just to get that value feels weird

PR #8009, opened 2024-09-26
Fix flash styling for donation page

manuthecoder:
This PR addresses #7683 and fixes the flash styling for the donation page. It also refactors the donation page layout.

Final screenshots
image
image


app/views/layouts/application.html.erb
```diff
@@ -164,8 +164,9 @@
       </main>
     <% else %>
       <%= render "application/header" unless @no_app_shell %>
+      <%= yield(:header) %>
       <main class="container <%= content_for :container_class %>" style="<%= content_for :container_style %>">
-        <%= render "application/flash" %>
+      <%= render "application/flash" unless controller_name == "donations" || controller_name == "reports" %>
```
sampoder: We probably don't want to do this as there are some flashes still intended to be displayed here with these controllers

manu-gurudath: I've updated the flash logic so it only hides it on `start_donation_donations_path`

sampoder (approved):
Tested and this LGTM!

PR #8010, opened 2024-09-26
[AchTransfer] Encrypt Routing Number Part 1/3

rluodev:
This is part 1 of 3.

In this step, we add a `:routing_number_ciphertext` field to the `:ach_transfers` table and mark the field as migrating for lockbox.

Before step 2, someone with prod access will need to run `Lockbox.migrate(AchTransfer)` to add unencrypted values to the ciphertext field. They'll also need to run `BlindIndex.backfill(AchTransfer)` to add blind index values.

Step 2 will consist of ignoring the `:routing_number` field in ruby and papertrail.

Step 3 will consist of dropping the `:routing_number` column from the DB.


sampoder:
Can we make it searchable? https://github.com/ankane/blind_index

rluodev:
Wait but routing numbers are all the same cause they’re from the same bank tho

rluodev:
If you searched a routing number it’d just return all the ach transfers

sampoder:
Hmm fair enough, maybe we can do that separately if we ever add it to account number

sampoder:
I just generally find having blind indexes helpful but if you think it's too much of a hassle, I'm chill with going without it

rluodev:
Oh wait no actually that’s a brain fart on my end; I’ll add blind index

rluodev:
(I ended up confusing ach transfer and the account numbers we have for each org)

sampoder (approved):

PR #8019, opened 2024-09-29
Update flavor_text_service.rb to change bank to (not) bank

leowilkin:
## Summary of the problem
HCB isn't technically a bank, so the statement "The only bank brave enough to say [...]" is misleading.

## Describe your changes
i put a (not) in front of the bank message in the text service file


sampoder:
Uhh... honestly, I kind of like the current copy. Additionally, the context behind why we had to stop using the word "bank" has changed. Thanks for making the PR but I think we can stick with we got

PR #8050, opened 2024-10-02
Show merchant names on card grant pages

24c02:
## Summary of the problem
people receiving card grants can't see what merchants their grants can be spent on, which will cause confusion for those with multiple grants for different purposes.

## Describe your changes
i added an "allowed merchants" section to the spending instructions view, which shows merchant names from yellow_pages, or failing that just the raw MIDs.

while i was in there, i also made the organizer view of card grants show friendly names instead of raw MIDs :-)

organizer view:
Screenshot 2024-10-02 at 12 31 39 PM
grantee view:
Screenshot 2024-10-02 at 12 32 05 PM


app/controllers/card_grants_controller.rb
```diff
@@ -60,6 +60,7 @@ def show
 
     @frame = params[:frame].present?
     @force_no_popover = @frame
+    @allowed_merchants = @card_grant.allowed_merchants.map {|merchant_id| YellowPages::Merchant.lookup(network_id: merchant_id).name || merchant_id}.to_set
```
garyhtou: I'd recommend moving this logic into the model under a method called `allowed_merchant_names`. Then, you can call that method directly from the view!

garyhtou:
Thank you! I love how you're helping improve the card grant experience for everyone

app/controllers/card_grants_controller.rb
24c02: done!

24c02:
## newest visual changes:
organizer/admin:
Screenshot 2024-10-02 at 4 45 40 PM
grantee:
Screenshot 2024-10-02 at 4 46 45 PM

garyhtou (dismissed):
Thank you for the screenshots! I'm happy with this. Can you run `bin/lint` to fix up rubocop/erblint?

sampoder (approved):
LGTM; thanks so much Nora!

PR #8063, opened 2024-10-03
Fix card grant info display

24c02:
Screenshot 2024-10-03 at 7 42 28 AM

yesterday's PR makes it do this on account of i'm a goof.
this should make it not do that?


sampoder:
Try `.any?`

24c02:
.any? seems to work!

PR #8073, opened 2024-10-06
Transition `OPENAI_API_KEY` to Doppler and introduce new `Credentials` module

sampoder:
For context, https://github.com/hackclub/hcb/issues/7042. Here's how the new system will work!

Two things, for now, staging and development (without `USE_PROD_CREDENTIALS` set to `true`) will remain the same. We will likely end up moving these systems in the future.

But what will change? Firstly, how credentials work on Heroku and how credentials work for developers with production access.

### Heroku

Variables in Doppler (`hcb/production`) are synced into the environment variables in Heroku. The new `Credentials` module will fetch those environment variables from `ENV`.

### Development w/ production access

Anyone developing with production access, should set `DOPPLER_TOKEN` in their `.env` file. This should be a service token generated here: https://dashboard.doppler.com/workplace/2818669764d639172564/projects/hcb/configs/production/access.

Then, on starting their development server, all environment variables defined in Doppler will be added to `ENV` which the  new `Credentials` module will fetch from.

If a developer wants to change an environment variable locally or add a new variable local without changing the variables in production, they should define it in their `.env` file. These will override anything in Doppler.

### Key format

The key for an environment variable should follow this format:

```
Credentials.fetch(:COLUMN, :API_KEY) # COLUMN__API_KEY
Credentials.fetch(:COLUMN, :FS_MAIN_ACCOUNT_ID) # #COLUMN__FS_MAIN_ACCOUNT_ID
Credentials.fetch(:LOOPS_API_KEY) # LOOPS_API_KEY
```

### Access to Doppler

I've sent invites to @garyhtou and @YodaLightsabr at the moment, let me know if any one else would like access.

### Transitioning other variables

I'm starting with `OPENAI_API_KEY` because it's low stakes. Other variables will be progressively transitioned over.


garyhtou:
And we're rotating secrets before adding them to Doppler. This way if the master key is leaked, we know that secrets in Doppler are safe.

PR #8079, opened 2024-10-06
Fix white borders on org ledger

sampoder:
See https://stackoverflow.com/questions/7021085/td-positionrelative-overflows-into-the-table-border for context

Closes https://github.com/hackclub/hcb/issues/8065


garyhtou:
oo interesting!

PR #8080, opened 2024-10-06
Create and use `fallback_avatar_tooltip`

sampoder:
Closes https://github.com/hackclub/hcb/issues/8070


garyhtou (commented):
feedback here: https://github.com/hackclub/hcb/issues/8089

PR #8082, opened 2024-10-06
[HcbCode Show Page] Hide zip and state when blank

sampoder:
Closes #7657


app/views/hcb_codes/_stripe_card.html.erb
```diff
@@ -190,14 +190,18 @@
...
+                    <strong>Zip</strong>
+                    <span><%= @hcb_code.stripe_merchant["postal_code"] %></span>
+                  </p>
+                <% end %>
+                <% if @hcb_code.stripe_merchant["state"] %>
+                  <p>
+                    <strong>State</strong>
+                    <span><%= @hcb_code.stripe_merchant["state"]&.upcase %></span>
+                  </p>
+                <% end %>
```
garyhtou: As a tip, when the `if` condition and the rendered variable are the same and are getting long, I like to do:

```erb
<% if merchant_zip = @hcb_code.stripe_merchant["postal_code"] %>
  <p>
  	<%= merchant_zip %>
  </p>
<% end %>
```

PR #8087, opened 2024-10-07
Better merchant category names

24c02:
now that yellow_pages can turn MCCs into reasonable names, let's have them everywhere!


app/models/card_grant.rb
```diff
@@ -182,7 +182,7 @@ def allowed_categories
   end
 
   def allowed_category_names
-    allowed_categories.map(&:humanize)
+    allowed_categories.map { |category| YellowPages::Category.lookup(key: category).name || "#{category}*" }.uniq
```
garyhtou: out of curiosity, what's the purpose of the `*`?

app/views/hcb_codes/_stripe_card.html.erb
```diff
@@ -295,7 +295,8 @@
           </p>
           <p>
             <strong>Category</strong>
-            <%= @hcb_code.stripe_merchant["category"]&.humanize %>
+            <%# we actually don't want humanized here – it's useful to see exactly what a merchant codes as in stripe_format for the purposes of setting up card grant allowlists %>
+            <%= @hcb_code.stripe_merchant["category"] %>
```
garyhtou: Thoughts on doing "Fast food restaurants (`fast_food_restaurants`)"?

garyhtou (dismissed):
tested. this functions well!

app/views/hcb_codes/_stripe_card.html.erb
24c02: yeah, i think i like that better!

app/models/card_grant.rb
24c02: we don't enforce that card grant category restrictions are actual stripe categories, i'm using this as shorthand for "something might not be right here"

garyhtou (approved):

PR #8092, opened 2024-10-07
Temporarily disabling importing transactions from Column for UTC transition

sampoder:


sampoder:
Covering by bases here before I actually make the switch.

PR #8093, opened 2024-10-07
Column (EST to UTC): prevent importing transactions before UTC 3AM October 7th from reports generated in UTC

sampoder:
These transactions were imported from this report: https://dashboard.column.com/app/documents/reporting/srpt_2n6ISzGK9ibhB6Okxqmd752jwOW.

They'll also show up in our October 7th report as we're now using UTC. This line of code will prevent them from being re-imported.

This PR also prevents duplicate transactions in the monthly statements and re-enables importing Column transactions.

Closes https://github.com/hackclub/hcb/issues/5209


app/services/transaction_engine/nightly.rb
```diff
@@ -91,6 +91,21 @@ def import_raw_column_transactions!
 
       transactions_by_report.each do |report_id, transactions|
         transactions.each_with_index do |transaction, transaction_index|
+          next if transaction["effective_at_utc"] == transaction["effective_at_utc"] && transaction["effective_at_utc"] < "2024-10-07T03:00:00Z"
+          # transactions that meet this condition would have been imported in the EST October 6th Report
```
sampoder:
```suggestion
          # transactions that meet this condition would have been imported in a report using EST
```

app/services/transaction_engine/nightly.rb
```diff
@@ -91,6 +91,21 @@ def import_raw_column_transactions!
...
       transactions_by_report.each do |report_id, transactions|
         transactions.each_with_index do |transaction, transaction_index|
+          next if transaction["effective_at_utc"] == transaction["effective_at_utc"] && transaction["effective_at_utc"] < "2024-10-07T03:00:00Z"
+          # transactions that meet this condition would have been imported in a report using EST
+          # they should be skipped when importing from a UTC report.
+          # this is related to the transition from reporting in EST to UTC.
+          #
+          # explanation of each condition
+          #
+          # transaction["effective_at_utc"] && transaction["effective_at_utc"]
```
sampoder:
```suggestion
          # transaction["effective_at"] == transaction["effective_at_utc"]
```

app/services/transaction_engine/nightly.rb
```diff
@@ -91,6 +91,21 @@ def import_raw_column_transactions!
 
       transactions_by_report.each do |report_id, transactions|
         transactions.each_with_index do |transaction, transaction_index|
+          next if transaction["effective_at_utc"] == transaction["effective_at_utc"] && transaction["effective_at_utc"] < "2024-10-07T03:00:00Z"
```
sampoder:
```suggestion
          next if transaction["effective_at"] == transaction["effective_at_utc"] && transaction["effective_at_utc"] < "2024-10-07T03:00:00Z"
```

app/services/transaction_engine/nightly.rb
```diff
@@ -91,6 +91,21 @@ def import_raw_column_transactions!
 
       transactions_by_report.each do |report_id, transactions|
         transactions.each_with_index do |transaction, transaction_index|
+          next if transaction["effective_at"] == transaction["effective_at_utc"] && transaction["effective_at_utc"] < "2024-10-07T03:00:00Z"
```
sampoder:
```suggestion
          next if transaction["effective_at"] == transaction["effective_at_utc"] && transaction["effective_at_utc"] < "2024-10-07T03:00:00Z"

```

app/services/transaction_engine/nightly.rb
```diff
@@ -94,6 +91,22 @@ def import_raw_column_transactions!
...
+          # they should be skipped when importing from a UTC report.
+          # this is related to the transition from reporting in EST to UTC.
+          #
+          # explanation of each condition
+          #
+          # transaction["effective_at"] == transaction["effective_at_utc"]
+          #
+          # if this condition is true, this report was generated in UTC
+          #
+          # transaction["effective_at_utc"] < "2024-10-07T03:00:00Z"
```
garyhtou: should this be 4AM? We're currently in EDT timezone which is -4 from UTC.

app/services/transaction_engine/nightly.rb
```diff
@@ -94,6 +91,22 @@ def import_raw_column_transactions!
 
       transactions_by_report.each do |report_id, transactions|
         transactions.each_with_index do |transaction, transaction_index|
+          next if transaction["effective_at"] == transaction["effective_at_utc"] && transaction["effective_at_utc"] < "2024-10-07T03:00:00Z"
```
garyhtou: Please have it record which ones we are skipping. at needs at least the IDs. I'd recommend both printing and airbreak'ing. When it runs, have the heroku logs pulled up on your terminal (heroku logs -n 5000 -t -a bank-hack club) since Heroku only keeps a few hundred/thousands lines.

app/jobs/column/statement_job.rb
```diff
@@ -54,6 +54,8 @@ def perform(date = DateTime.current.prev_month)
 
       transactions_by_report.each_value do |transactions|
         transactions.reverse.each_with_index do |transaction, transaction_index|
+          next if transaction["effective_at"] == transaction["effective_at_utc"] && transaction["effective_at_utc"] < "2024-10-07T03:00:00Z" # see TransactionEngine::Nightly#import_raw_column_transactions! for a description of this logic
```
garyhtou: good job using < (less than), and not <= since report's end dates are exclusive. it doesn't include midnight.

app/services/transaction_engine/nightly.rb
sampoder: Good catch! Yes.

app/services/transaction_engine/nightly.rb
sampoder: Done! We don't have exact IDs unfortunately so I'm going to print the whole transaction object and then check it against my list that should be skipped

sampoder: There'll be 232 duplicates

PR #8095, opened 2024-10-07
[Google Workspace] Send state update mailers to all org managers

rluodev:
## Summary of the problem

Some Google Workspaces were created by admins or users who are no longer in orgs. this means that emails were going to these people instead of the proper org managers.

## Describe your changes

Google Workspace emails go to org managers now.


spec/mailers/previews/g_suite_mailer_preview.rb
```diff
@@ -5,7 +5,7 @@ def notify_of_configuring
     g_suite = GSuite.configuring.last
     g_suite ||= GSuite.last
     @g_suite_id = g_suite.id
-    @recipient = g_suite.created_by.email
+    @recipient = g_suite.event.organizer_positions.where(role: :manager).includes(:user).map(&:user).map(&:email_address_with_name)
     GSuiteMailer.with(
       recipient: @recipient,
```
garyhtou: Do we still need to pass `recipient` here? (same with line 21 below)

garyhtou (commented):
Instead of having a `recipient` param in GSuiteMailer, let's drop it and get the list of recipients from `@g_suite.event....`

garyhtou (commented):
It looks like we're having issues with RSpec because the Event factory does not have any associated Organization Positions. I'd recommend looking into adding it to the Event factory.

https://github.com/hackclub/bank/blob/d4894bc7f1f123f45430b04e287fac4cf26e0c2c/spec/factories/event_factory.rb#L4

garyhtou (approved):

PR #8097, opened 2024-10-07
Flag reimbursement reports where an admin's approval would make the event negative

sampoder:
There are still some (limited) circumstances where we might want this to go through.

This situation happens when a report is submitted but then there are more transactions on the event before an admin reviews the report.


sampoder:
I'd also be down to prevent this entirely. Just Lucy seems to be manually check this rn

sampoder:
So we actually do prevent this, just it wasn't showing up in the UI.

PR #8100, opened 2024-10-08
Disable PVSA Perk

RhysMP:
## Summary of the problem

Disabling PVSA perk as they have suspended our account. Looking into it. Fillout form has also been disabled.

## Describe your changes

image


sampoder (approved):
LGTM!

PR #8101, opened 2024-10-08
[Org Team] Fix blankslate wording

rluodev:
## Summary of the problem

The blankslate said "no other team members" if you were just filtering in a way that hides all members

## Describe your changes

Now it says "your filter is hiding all members" if you have a filter going because an activated org should never have a situation where there are no people. Admins will see "no team members" for no team orgs

Screenshot 2024-10-08 at 10 37 46 AM
Screenshot 2024-10-08 at 10 38 01 AM


app/views/events/team.html.erb
```diff
@@ -71,7 +71,7 @@
   </table>
   <%= paginate @positions %>
   <% else %>
-    <%= blankslate "No #{params[:filter] == "all" ? "others" : params[:filter] || "others"} on your team yet" %>
+    <%= blankslate (@event.organizer_positions.none? ? "No team members" : "Your filter is hiding all team members").to_s %>
```
garyhtou:
```suggestion
    <%= blankslate @event.organizer_positions.none? ? "No team members" : "No team members found" %>
```

app/views/events/team.html.erb
```diff
@@ -83,7 +83,7 @@
       <%= paginate @positions %>
     <% end %>
   <% else %>
-    <%= blankslate "No others on your team yet" %>
+    <%= blankslate (@event.organizer_positions.none? ? "No team members" : "Your filter is hiding all team members").to_s %>
```
garyhtou:
```suggestion
    <%= blankslate @event.organizer_positions.none? ? "No team members" : "No team members found" %>
```

PR #8136, opened 2024-10-11
[Disbursements] Improve organization selector

### LINKED ISSUE #8052: Organization dropdown for disbursements and admin pages
At present, admins have every single org listed in the org select dropdown menu. We could probably make this a better experience to allow for typing the org name and showing most relevant organizations first.

> hey! I'm requesting a feature that I think would really help me– dropdowns for orgs (like on the disbursement form) are getting really really long, and are unsearchable in some browsers now. i'd really love to have a text field I can type into that shows me suggestions, or some other alternative input for selecting the org I'm making a transfer to

https://hackclub.slack.com/archives/CN523HLKW/p1727919042807369

YodaLightsabr:
## Summary of the problem

It can be difficult to select organizations as an admin (especially on mobile) because the native dropdown lists thousands of organizations.

## Describe your changes

I've built a custom dropdown component we can use on the disbursements page (and in the future, elsewhere, too).

For admins, it shows the organization ID and moves the balance into a tooltip.

Here's a video:

https://github.com/user-attachments/assets/0e893cc8-62db-462b-be5e-9ae4086c60f1

And here's a screenshot on dark mode:

image


tailwind.config.js
```diff
@@ -61,6 +61,7 @@ module.exports = {
         snow: '#f9fafc',
         darkless: '#252429',
         dark: '#17171d',
+        black: '#1f2d3d',
```
garyhtou: Maybe we should find another name for this since it's not actually black

YodaLightsabr: 🙃 I thought the same thing as I pasted it in. I pulled it from https://github.com/hackclub/bank/blob/8d80155764b293676b52f6ed0e47b9ac043666e9/app/assets/stylesheets/_variables.scss#L1 though, and it's called black on the Hack Club [theme website](https://theme.hackclub.com) as well. Do you think we should keep the name for consistency?

app/assets/stylesheets/components/_forms.scss
```diff
@@ -283,7 +283,8 @@ label {
 
 $select: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" fill="%238492a6" viewbox="0 0 32 32"><path d="M0 6 L32 6 L16 28 z"/></svg>');
 
-select {
+select,
```
sampoder: Thoughts on hiding the scrollbar? I feel like it sticks out a little bit and that even without it, it would look like you could scroll

app/javascript/controllers/organization_select_controller.js
```diff
@@ -0,0 +1,165 @@
...
+          organizations[oldId].button.children[1].style.color = ''
+        }
+        this.dropdownTarget.children[1].innerText = name
+        this.dropdownTarget.children[1].value = id
+        this.dropdownTarget.children[1].dataset.fee = fee
+        this.dropdownTarget.value = id
+        this.dropdownTarget.dispatchEvent(new CustomEvent('feechange'))
+        button.style.backgroundColor = 'var(--info)'
+        button.style.color = 'white'
+        button.children[1].style.color = 'white'
```
sampoder: Might make sense to do this using a class instead of directly setting the styles.

app/javascript/controllers/organization_select_controller.js
```diff
@@ -0,0 +1,165 @@
...
+          threshold: -500000,
+        }
+      )
+
+      firstOrganization = result[0]?.obj
+
+      const shown = result.map(r => r.obj.organization)
+      const hidden = this.organizationTargets.filter(el => !shown.includes(el))
+
+      if (!ensure(() => text == this.searchTarget.value)) return
```
sampoder: What's the goal of this line? Is it because fuzzy searching is slow, so this ensures that the value of input hasn't changed in between the event being triggered and this line. Sorry I'm just trying to wrap my head around the ensure stuff.

app/javascript/controllers/organization_select_controller.js
```diff
@@ -0,0 +1,165 @@
...
+    const organizations = {}
+
+    const ensure = async callback => {
+      try {
+        await scheduler.yield()
+        return callback()
+      } catch {
+        return callback()
+      }
+    }
```
sampoder: I assume it's try / catched because not all browsers support `scheduler.yield()`? https://developer.mozilla.org/en-US/docs/Web/API/Scheduler/yield#browser_compatibility

app/views/disbursements/_form.html.erb
```diff
@@ -1,37 +1,84 @@
...
+  </div>
+
+    <%# form.label :event_id, "To" %>
+    <%# form.select(:event_id,
                     @allowed_destination_events.map do |event|
                       [
                         current_user.admin? ? event.admin_dropdown_description : event.disbursement_dropdown_description,
                         event.id,
-                        { "data-fee" => event.revenue_fee.to_s },
+                        { "data-fee" => 0 || event.revenue_fee.to_s },
```
sampoder: Screenshot 2024-10-16 at 9 24 37 PM

This will always return 0, right?

sampoder (dismissed):
This is really nicely done! And so much cleaner than what we had before....

sampoder:
Screenshot 2024-10-16 at 9 30 10 PM

When users only have a few organisations, thoughts on reducing the height?

app/javascript/controllers/organization_select_controller.js
YodaLightsabr: This could probably be cleaned up & put inline since it's only called once, but there's a cool new feature in some browsers that lets you put the current block back at the top of the stack. It can be useful if the input changes so that you can render with the latest information instead of wasting time rendering old data and delaying rendering the newest data. https://developer.mozilla.org/en-US/docs/Web/API/Scheduler/yield

app/javascript/controllers/organization_select_controller.js
YodaLightsabr: Yeah.

app/views/disbursements/_form.html.erb
YodaLightsabr: What an astute observation. This was left in from debugging, good catch.

YodaLightsabr: This was also commented out and isn't necessary to keep around, so I'll remove the whole section.

app/javascript/controllers/organization_select_controller.js
sampoder: Sweet! That's cool

app/views/disbursements/_form.html.erb
sampoder: Astute... lol nice, we can remove it then

garyhtou:
Feedback from call:
- Keep the scrollbar since it's helpful for seeing where you are in a long list. and important for accessibility
- See if you can get the dropdown to be on top of the page and not displace the page content. Maybe `position: absolute`?
- Add a tooltip for disabled options with a reason (e.g. Insufficient balance)
- We would height: min-content as sam mentioned

app/javascript/controllers/organization_select_controller.js
garyhtou: You could put the `callback()` in a finally block if you want to :)

sampoder:
I hear you on the scrollbar thing, hadn't thought about the benefit of it in a long list. I guess we keep it yeah.

app/assets/stylesheets/components/_forms.scss
YodaLightsabr: I think it's pretty helpful to be able to know where you are in the list, and it's nice to be able to drag it to the top or bottom.

app/javascript/controllers/organization_select_controller.js
YodaLightsabr: I've refactored the code a bit to make it more readable [eb73472](https://github.com/hackclub/hcb/pull/8136/commits/eb73472351dbd227e7caef38e68acd340d016e5f). A class might be a little bit cleaner, but I think it would fragment the component a lot more because we're already using JS and Tailwind. What do you think?

app/javascript/controllers/organization_select_controller.js
YodaLightsabr: Eslint is complaining about an empty catch block :/ Not sure how I could get around it without it being messier.

garyhtou:
Thanks!

PR #8137, opened 2024-10-11
Remove memoization from methods in `Event` that take `start_date` and `end_date` as parameters

sampoder:
We still memo-ize the balance available method, so hopefully we keep most of the benefits of memoization

Closes https://github.com/hackclub/hcb/issues/8128


garyhtou:
Can you see how this affects loading HQ's transactions and the stripe card approval flow?

I think Rails may chance these repeated queries, but if there's significant degrade in performance, then i'd suggest keeping the memorization:

```ruby
def balance(start_date: nil, end_date: nil)
  @balance ||= {}
  @balance[[start_date, end_date]] ||= begin
    # code
  end
end
```

sampoder:
Do you mean the authorisation webhook?

sampoder:
I'll test them both. The Stripe authorisation flow relies on `balance_available_v2_cents` which is still memoised.

garyhtou:
Got it, okay thanks!

sampoder:
Our SQL queries went from 368 to 369, I think this is good to merge.

sampoder:
I've walked through the Stripe authorisation flow and am confident that we shouldn't see any hits to performance. I'll keep an eye on `#hcb-engr-notifs`.

PR #8138, opened 2024-10-11
Validate email on donation

sampoder:
Currently only done on the frontend.

Closes https://github.com/hackclub/hcb/issues/8126.


app/models/donation.rb
```diff
@@ -79,6 +79,7 @@ class Donation < ApplicationRecord
   after_commit :send_donation_notification
 
   validates :name, :email, presence: true, unless: -> { recurring? || in_person? } # recurring donations have a name/email in their `RecurringDonation` object
+  validates :email, on: :create, format: { with: URI::MailTo::EMAIL_REGEXP, message: "must be a valid email address" }, allow_nil: true
```
garyhtou: I'm assuming we added `allow_nil: true` because recurring and in-person donations don't have an email on this model. If so, instead of doing allow nil, i'd suggest `unless: -> { recurring? || in_person? }`; like above

PR #8142, opened 2024-10-14
Replace `Loading...` text with SVG loading indicator

manuthecoder:
This PR addresses #8008, replacing `Loading...` text with a spinner icon. 

> | Before | After |
> |--------|--------|
> | static "Loading..." text|  | 
> 
> The above image is an SVG file [(source)](https://github.com/nickbruun/svg-loaders), and it's a simpler alternative to a skeleton screen

Home page and popover components are affected


app/views/events/home/_money_movement.html.erb
```diff
@@ -31,8 +31,8 @@
         <% if Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (@event || transaction.local_hcb_code.event).present? %>
           <section class="modal modal--scroll modal--popover bg-snow" data-behavior="modal" role="dialog" id="money_movement_tx_details_<%= transaction.local_hcb_code.__id__ %>">
             <%= modal_header(transaction.local_hcb_code.pretty_title(show_event_name: defined?(show_event_name), show_amount: defined?(show_amount), event_name: @event.name, amount_cents: transaction.amount), external_link: transaction.local_hcb_code.url) %>
-            <%= async_frame_to transaction.local_hcb_code.popover_url, as: :span, lazy: true do %>
-              Loading...
+            <%= async_frame_to transaction.local_hcb_code.popover_url, as: :div, class: "loading-container", lazy: true do %>
```
garyhtou: When the async frame finishes loaded, does it still have the `loading-container` class on it?

If so, maybe we move the class into a div inside the block.

garyhtou (commented):
Consider creating a partial called `application/_loading_indicator`.

It would have a `div` that contains the `inline_icon "loading"`. The div can then have inline styles with the stuff that's currently applied by the `loading-container` class.

garyhtou (approved):

PR #8144, opened 2024-10-14
Remove organizers from `transactions.html.erb`

manuthecoder:
Since we have a new event home page, we can remove the profile pictures from the transactions page. Pictures will still show if the flag isn't enabled

Preview
image

cc @garyhtou


app/views/events/transactions.html.erb
```diff
@@ -52,62 +52,82 @@
...
-      <% else %>
-        <span class="stat__value" data-balance-graph-target="balance">
+  <% if Flipper.enabled?(:event_home_page_redesign_2024_09_21, @event) %>
+    <div class="flex items-center details-horiz details-horiz--lg" style="gap: 8px; pointer-events: none;">
+      <div class="stat statset__wide" data-tour-step="balance">
+        <span class="stat__label">Account balance</span>
+        <% if @event.fee_balance_v2_cents < 0
+             # this event has a fee credit (owes negative fees) %>
+          <span class="stat__value"><%= render_money_amount @event.balance_v2_cents %></span>
+          <span class="self-end" style="font-size: 10px;">+$<%= render_money_amount @event.fee_balance_v2_cents.abs %>
```
malted:
```suggestion
          <span class="self-end" style="font-size: 10px;">+$<%= render_money_amount @event.fee_balance_v2_cents * -1 %>
```

sampoder (approved):

app/views/events/transactions.html.erb
sampoder: I think that we currently have makes more sense, at least inside the conditional

sampoder:
something for the future: we should work out how to make the balance smaller so it is differentiated from the homepage

PR #8157, opened 2024-10-15
[Card Grant] Send expiration warning emails at 30 days and 3 days to expiry

rluodev:
We now send two emails, one at 30 days and one at 3, to warn user of expiring


app/models/card_grant.rb
```diff
@@ -72,6 +72,7 @@ class CardGrant < ApplicationRecord
   scope :not_activated, -> { active.where(stripe_card_id: nil) }
   scope :activated, -> { active.where.not(stripe_card_id: nil) }
   scope :search_recipient, ->(q) { joins(:user).where("users.full_name ILIKE :query OR card_grants.email ILIKE :query", query: "%#{User.sanitize_sql_like(q)}%") }
+  scope :expires_on, ->(date) { joins(:card_grant_settings).active.where("card_grants.created_at + (card_grant_settings.expiration_preference * interval '1 day') < #{date}") }
```
garyhtou: Sorry, I know that I gave you this code, here are a couple of edits:
- Let's remove `active` from it and apply that wherever we call `expires_on`. This is because:
  ```ruby
  CardGrant.expires_on(1.day.from_now).count # let's say it's 4 cards
  CardGrant.expires_on(1.day.from_now).each(&:expire!) # expire them all
  CardGrant.expires_on(1.day.from_now).count
  #=> 0
  # It's kinda confusing when expires_on will change depending on whether cards
  # have already been expired or not. So, I'd recommend removing `active` from
  # inside the `expires_on` scope, and just add it when we use the scope.

  # For example, usage will become:
  CardGrant.active.expires_on(1.day.from_now).each(&:expire!)
  ```
- rename `expires_on` to `expires_after`. I think it's more descriptive since our query uses `<` and not an exact date.

garyhtou:
```suggestion
  scope :expires_on, ->(date) { joins(:card_grant_settings).active.where("card_grants.created_at + (card_grant_settings.expiration_preference * interval '1 day') < ?", date) }
```

By using `?` instead of directly interpolating the value into the string, Rails will sanitize the value for us (and do any formatting necessary).

app/jobs/card_grant_expiration_job.rb
```diff
@@ -3,9 +3,18 @@
...
       card_grant.expire!
     end
+
+    CardGrant.expires_on(1.month.from_now).find_each do |card_grant|
+      CardGrantMailer.with(card_grant:).card_grant_month_expire_notification.deliver_later
+    end
+
+    CardGrant.expires_on(3.days.from_now).find_each do |card_grant|
+      CardGrantMailer.with(card_grant:).card_grant_days_expire_notification.deliver_later
+    end
```
garyhtou: Is there a chance that someone will receive both emails at the same time?

app/mailers/card_grant_mailer.rb
```diff
@@ -8,4 +8,16 @@ def card_grant_notification
...
+    @card_grant = params[:card_grant]
+
+    mail to: @card_grant.user.email_address_with_name, subject: "[#{@card_grant.event.name}] Your #{@card_grant.amount.format} grant expires in one month"
+  end
+
+  def card_grant_days_expire_notification
+    @card_grant = params[:card_grant]
+
+    mail to: @card_grant.user.email_address_with_name, subject: "[#{@card_grant.event.name}] Your #{@card_grant.amount.format} grant expires in three days"
+  end
```
garyhtou: Maybe we just create one mailer action

```ruby
def expiration_notification
  mail to: sdflksjdflksdjf
end
```

In the view, use `distance_of_time_in_words`

app/models/card_grant.rb
rluodev: @garyhtou wait i just realized these have the same names - would this be an issue?

https://github.com/hackclub/hcb/blob/2faf70ef11f4f9bd8db136642c3fb6be6a0776d7/app/models/card_grant.rb#L189-L195

rluodev: oh wait nvm i got confused and forgot scope ≠ instance method

app/jobs/card_grant_expiration_job.rb
rluodev: wait no ignore me

rluodev: hmm yeah this would happen I feel like

app/mailers/card_grant_mailer.rb
rluodev: Screenshot 2024-10-17 at 7 48 07 PM

this would say "about one month" though which seems syntactically odd

rluodev: wait nvm just passed in a text month

app/views/card_grant_mailer/card_grant_expiry_notification.html.erb
```diff
@@ -0,0 +1,9 @@
+Your <%= @card_grant.amount.format %> grant from <%= @card_grant.event.name %> will expire in <%= @expiry_time %>.
+
+<br>
+<br>
+<br>
+
+<%= link_to "View your grant", card_grant_url(@card_grant, email: @card_grant.user.email), style: "background-color: #ec3750; color: white; padding: 1rem; border-radius: 0.75rem; margin: 0 auto 2rem 0; text-decoration: none" %>
+
+<br>
```
sampoder:
```suggestion
<p>
  Your <%= @card_grant.amount.format %> grant from <%= @card_grant.event.name %> will expire in <%= @expiry_time %>.
</p>

<%= link_to "View your grant", card_grant_url(@card_grant, email: @card_grant.user.email), style: "background-color: #ec3750; color: white; padding: 1rem; border-radius: 0.75rem; margin: 0 auto 2rem 0; text-decoration: none" %>
```

app/models/card_grant.rb
```diff
@@ -72,6 +72,7 @@ class CardGrant < ApplicationRecord
   scope :not_activated, -> { active.where(stripe_card_id: nil) }
   scope :activated, -> { active.where.not(stripe_card_id: nil) }
   scope :search_recipient, ->(q) { joins(:user).where("users.full_name ILIKE :query OR card_grants.email ILIKE :query", query: "%#{User.sanitize_sql_like(q)}%") }
+  scope :expires_after, ->(date) { joins(:card_grant_settings).where("card_grants.created_at + (card_grant_settings.expiration_preference * interval '1 day') < ?", date) }
```
sampoder: We should probably make a scope like `expires_on`

```suggestion
  scope :expired_before, ->(date) { joins(:card_grant_settings).where("card_grants.created_at + (card_grant_settings.expiration_preference * interval '1 day') < ?", date) }
  scope :expires_on, ->(date) { joins(:card_grant_settings).where("card_grants.created_at + (card_grant_settings.expiration_preference * interval '1 day') = ?", date) }
```

sampoder: That way we don't send the emails again and again.

rluodev: @sampoder would we want to switch to date.today if we made this change?

sampoder: Yes

app/jobs/card_grant_expiration_job.rb
```diff
@@ -3,9 +3,18 @@
 class CardGrantExpirationJob < ApplicationJob
   queue_as :low
   def perform
-    CardGrant.joins(:card_grant_setting).active.where("card_grants.created_at + (card_grant_settings.expiration_preference * interval '1 day') < NOW()").find_each do |card_grant|
+    CardGrant.active.expires_after(Time.now).find_each do |card_grant|
```
sampoder:
```suggestion
    CardGrant.active.expires_before(Time.now).find_each do |card_grant|
```

app/jobs/card_grant_expiration_job.rb
```diff
@@ -3,9 +3,18 @@
...
-    CardGrant.joins(:card_grant_setting).active.where("card_grants.created_at + (card_grant_settings.expiration_preference * interval '1 day') < NOW()").find_each do |card_grant|
+    CardGrant.active.expires_before(Time.now).find_each do |card_grant|
       card_grant.expire!
     end
+
+    CardGrant.active.expires_after(3.days.from_now).find_each do |card_grant|
+      CardGrantMailer.with(card_grant:, expiry_time: "3 days").card_grant_days_expire_notification.deliver_later
+    end
+
+    CardGrant.active.expires_after(1.month.from_now).find_each do |card_grant|
```
sampoder:
```suggestion
    CardGrant.active.expires_on(3.days.from_now).find_each do |card_grant|
      CardGrantMailer.with(card_grant:, expiry_time: "3 days").card_grant_days_expire_notification.deliver_later
    end

    CardGrant.active.expires_on(1.month.from_now).find_each do |card_grant|
```

PR #8161, opened 2024-10-16
[ChangelogPost] Migrate from Flipper to join table

KiKoS0:
Closes https://github.com/hackclub/hcb/issues/7728
## Summary of the problem
Migrates marking changelog posts as seen by users from Flipper to the database.

## Describe your changes
- Create a new association table `changelog_posts_users` to mark posts as seen by users.
- Cleanup the Flipper call to the `native_changelog_2024_07_03` feature flag which shouldn't be needed anymore.
- Brings back the [disabled](https://github.com/hackclub/hcb/pull/7704/files) updated logic, @garyhtou let me know if you meant that we should keep it disabled for now.


app/models/changelog_post.rb
```diff
@@ -17,4 +17,15 @@ def self.latest
...
 
+  has_and_belongs_to_many :viewers, class_name: "User"
+
+
+  def seen_by?(user)
+    viewers.include?(user)
+  end
+
+  def mark_seen_by!(user)
+    viewers << user
```
albertchae: I forget, does this persist to the DB right away or just the in memory array?

app/models/changelog_post.rb
```diff
@@ -17,4 +17,15 @@ def self.latest
     order(published_at: :desc).first
   end
 
+  has_and_belongs_to_many :viewers, class_name: "User"
+
+
+  def seen_by?(user)
+    viewers.include?(user)
```
albertchae: I think this might load all the users into memory before doing `include?` in ruby. Can you double check whether it does? If it does, we should probably do the AR equivalent of joining on `user_id` and checking for `exists?`

app/controllers/events_controller.rb
```diff
@@ -253,6 +253,16 @@ def transactions
...
     end
 
+
+    if current_user
+      latest_changelog_post = ChangelogPost.latest
+
+      if latest_changelog_post && !latest_changelog_post.seen_by?(current_user)
+        latest_changelog_post.mark_seen_by!(current_user)
+        @latest_changelog_post = latest_changelog_post
+      end
```
albertchae: @garyhtou should this feature dismiss the latest changelog post on any refresh?

garyhtou: This is how the feature functions at the moment. In the future we could improve it by having an `X` button mark the changelog as seen. But for now, this is good!

garyhtou:
Thanks for putting this PR together!

> Brings back the [disabled](https://github.com/hackclub/hcb/pull/7704/files) updated logic, @garyhtou let me know if you meant that we should keep it disabled for now.

It was disabled because of some UI bugs. I would suggest:
1. In this PR, leave it disabled so we can merge it without negatively affecting the UI
2. Fixing the UI bugs (https://github.com/hackclub/hcb/issues/7727)
3. After the bugs have been fixed, we can enable it!

If you'd like to explore fixing the UI bugs, feel free to do so; chat with @albertchae first.

app/models/changelog_post.rb
KiKoS0: Yeah I already checked that, it's smart enough to run an exists sql query on the new table. But let me check again and I'll link some logs.

KiKoS0: As i thought it does a: 
```sql
SELECT 1 AS one FROM "users" INNER JOIN "changelog_posts_users" ON "users"."id" = "changelog_posts_users"."user_id" WHERE "changelog_posts_users"."changelog_post_id" = $1 AND "users"."id" = $2 LIMIT $3  [["changelog_post_id", 4], ["id", 1], ["LIMIT", 1]]
```
###### Full log
```ruby
23:46:26 web.1    |   ChangelogPost Load (1.8ms)  SELECT "changelog_posts".* FROM "changelog_posts" ORDER BY "changelog_posts"."published_at" DESC LIMIT $1  [["LIMIT", 1]]
23:46:26 web.1    |   ↳ app/models/changelog_post.rb:17:in `latest'
23:46:26 web.1    |   User Exists? (2.9ms)  SELECT 1 AS one FROM "users" INNER JOIN "changelog_posts_users" ON "users"."id" = "changelog_posts_users"."user_id" WHERE "changelog_posts_users"."changelog_post_id" = $1 AND "users"."id" = $2 LIMIT $3  [["changelog_post_id", 4], ["id", 1], ["LIMIT", 1]]
23:46:26 web.1    |   ↳ app/models/changelog_post.rb:24:in `seen_by?'
23:46:26 web.1    |   TRANSACTION (1.6ms)  BEGIN
23:46:26 web.1    |   ↳ app/models/changelog_post.rb:28:in `mark_seen_by!'
23:46:26 web.1    |   ChangelogPost::HABTM_Viewers Create (29.4ms)  INSERT INTO "changelog_posts_users" ("changelog_post_id", "user_id", "created_at", "updated_at") VALUES ($1, $2, $3, $4) RETURNING "id"  [["changelog_post_id", 4], ["user_id", 1], ["created_at", "2024-10-16 21:46:26.614537"], ["updated_at", "2024-10-16 21:46:26.614537"]]
23:46:26 web.1    |   ↳ app/models/changelog_post.rb:28:in `mark_seen_by!'
23:46:26 web.1    |   TRANSACTION (3.5ms)  COMMIT
23:46:26 web.1    |   ↳ app/models/changelog_post.rb:28:in `mark_seen_by!'
```

app/models/changelog_post.rb
KiKoS0: It persists in the DB, I linked some logs in the other comment that includes the insert operation.

albertchae: Cool thanks, I don't use shovel operator enough so I always forget 😆

app/models/changelog_post.rb
albertchae: Awesome, thanks for confirming

spec/models/changelog_post_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  describe "mark_seen_by!" do
+    it "adds a user to the viewers association" do
+      expect(changelog_post.viewers).to be_empty
+      changelog_post.mark_seen_by!(user)
+      expect(changelog_post.viewers).to include(user)
+    end
+
+    it "does not add a user to the viewers association more than once" do
+      changelog_post.mark_seen_by!(user)
+      expect { changelog_post.mark_seen_by!(user) }.to raise_error(ActiveRecord::RecordNotUnique)
```
garyhtou: This line makes sense since we have a unique constraint on the database. However, I wonder if it would be beneficial to rescue `ActiveRecord::RecordNotUnique` inside `ChangelogPost#mark_seen_by!`.

When using `ChangelogPost#mark_seen_by!`, I think we generally don't care if it's already been marked as seen. So, having it silently fail on a "duplicate seen" may be beneficial. What are your thoughts?

KiKoS0: I agree, crashing the page because of a rare race condition seems unnecessary 👍

app/models/changelog_post.rb
```diff
@@ -17,4 +17,17 @@ def self.latest
...
+  has_and_belongs_to_many :viewers, class_name: "User"
+
+  def seen_by?(user)
+    viewers.include?(user)
+  end
+
+  def mark_seen_by!(user)
+    begin
+      viewers << user
+    rescue ActiveRecord::RecordNotUnique
```
albertchae: nit: maybe add a comment why we're swallowing this? although I guess you have a test case covering this behavior

albertchae (dismissed):
tests look good

garyhtou (dismissed):
Looks good to me! One small comment:

I wonder if we should name the methods `viewed_by?` and `mark_viewed_by!` since the association is called `viewers`. It feels more consistent this way.

I'm happy with the code as-is, so @albertchae feel free to merge if you'd like.

albertchae (approved):
LGTM, thanks for addressing latest comments. going to merge

PR #8176, opened 2024-10-17
[Plaid] Fix SVB Sync

garyhtou:
## Summary of the problem

SVB's switch from Online Banking to SVB Go, broke our Plaid transaction syncing. On Plaid's end, SVB Online Banking and SVB Go are two completely separate institutions. This meant that re-linking our bank account with Plaid resulted in a new Plaid Item ID, Plaid Access Token, Plaid Account ID, and also **new Plaid Transaction IDs**. Having new Plaid Transaction IDs breaks our idempotent import process.

We were able to sync all of October 8th, 2024's transactions before Plaid broke. There were no transactions on October 9th.

So, the goal is to start syncing back up from October 10th's transactions.

More context: I believe SVB Online Banking user access was shut down on the night of October 9th in prep for the migration to SVB Go which was live on October 10th (at least of our "company" on SVB).

## Describe your changes

I initially explored reauthenticating our existing `BankAccount` record (`BankAccount.find(4)`)`. However, this would not work because Plaid expects the bank account to still be with the same institution; duh. So, we were left with create a new Item on Plaid. There are two ways to do this on HCB.
1. Create a brand new `BankAccount` record.
2. Replace an existing `BankAccount` record by [keeping its ID](https://github.com/hackclub/bank/blob/868ed4d4e8b47cfe96352d609d9aade355a9bc8e/app/controllers/bank_accounts_controller.rb#L19-L20).

The second option sounded favorable at first. However, I discovered that when replacing an existing BankAccount record, the following fields are updated:
- `plaid_access_token`
- `plaid_item_id`
- `plaid_account_id`

The `plaid_account_id` is problematic because `RawPlaidTransaction` [uses it as a foreign key](https://github.com/hackclub/bank/blob/2cbab10a884cf89be398af5db1706159e60562a3/app/models/raw_plaid_transaction.rb#L22) for `BankAccount`.

This meant if we replaced the existing `BankAccount` record, all of the existing `RawPlaidTransactions` would lose their associations to their `BankAccount`. I'm not sure what the full extent of this impact would have been (it seemed minor). But either way, that didn't seem like the best idea.

So, I am opting to connect our new SVB Go Plaid account as a brand new `BankAccount` record.

Here are the steps that I'm going to perform in production:
1. Create a new `BankAccount` at https://hcb.hackclub.com/bank_accounts
	1. Choose the use as guest option. (Don't set up your phone number)
	2. Choose SVB GO
	3. Sign in using the dedicated SVB user that we have for plaid syncing
	4. Choose to connect FS Main
	5. Finish and skip the phone number thing again
	6. This created `BankAccount.find(50)`.
2. Merge this PR
	- In `app/services/transaction_engine/shared.rb`, have the new ID share the same `unique_bank_identifier` as the old FS Main. See [this comment](https://github.com/hackclub/hcb/pull/8176#discussion_r1804058046).
3. Disable syncing for the old FS Main `BankAccount.find(4)`.
4. Enable syncing for the new FS Main `BankAccount.find(50)`.
8. Run `TransactionEngine::Nightly.new.run`
9. Verify 7 new unmapped FS Main transactions show up on the admin ledger (https://hcb.hackclub.com/admin/ledger)
	- We only have SVB since we no longer really use SVB. We've switched to Column.


app/services/transaction_engine/shared.rb
```diff
@@ -11,7 +11,7 @@ def unique_bank_identifier
       case @bank_account_id
       when 1
         "EVENTFISCALSPONSORSHIP"
-      when 4
+      when 4, 50
```
garyhtou: The new `BankAccount` record should still use the existing `unique_bank_identifier` of `FSMAIN`.

app/services/transaction_engine/raw_plaid_transaction_service/plaid/import.rb
```diff
@@ -16,6 +16,28 @@ def run
           plaid_transactions.each do |plaid_transaction|
             next unless plaid_transaction.pending == false
 
+            if plaid_transaction.date <= "2024-10-09".to_date
```
garyhtou: eh, this was supposed to be "2024-10-08", but that's fine as "09".

Syncing for SVB technically broke on the night of the 9th I think. But there are no transaction on the 9th so this doesn't matter.

garyhtou:
Success!
image

PR #8182, opened 2024-10-17
Don't email if the card was cancelled a while ago.

sampoder:
Because emails are spammy.

Closes https://github.com/hackclub/hcb/issues/8141


app/mailers/canonical_pending_transaction_mailer.rb
```diff
@@ -45,6 +45,7 @@ def notify_declined
     @user = @card.user
 
     return unless @user.email_charge_notifications_enabled?
+    return if @card.canceled? && @card.updated_at < 1.month.ago
```
garyhtou: Can we try something more accurate than `updated_at`? If a card gets updated in any way after cancellation, then it would break this logic.

I think we could use papertrail instead (look for changes to `stripe_status`)

sampoder: Yeah, I know. Using Papertrail felt just as jank though.

garyhtou: The better solution here is:
1. Create a canceled_at column
2. Have our cancel card method fill that it (make sure webhook from stripe also fill it in)
	- Best bet may be adding it into the sync with stripe method?? (if stripe provides a cancel timestamp?)
3. backfill it using papertrail (or from stripe)

sampoder:
```suggestion
    return if @card.canceled? && @card.canceled_at < 1.month.ago
```

We now have a `canceled_at`! https://github.com/hackclub/hcb/pull/8530

sampoder:
FYI @MelSmith104 & @garyhtou, I know a fair few people were asking for this!

PR #8191, opened 2024-10-18
[Sendy] Remove unused user syncing code

garyhtou:
It looks like this portion is not used anywhere besides for the one time job.

I'm going to double check with ops that we're good to remove this


garyhtou:
Eh, i'm actually not going to check with ops since we're actively using Loops.so, and this code likely hasn't run in a very long time.

Easy to revert if never needed.

PR #8193, opened 2024-10-18
Make comments more readable on the admin page

sampoder:
Screenshot 2024-10-17 at 11 21 42 PM

We don't need to trick this up, and this fits in more with the rest of the page.

Closes https://github.com/hackclub/hcb/issues/8135


garyhtou (changes requested):
I'd prefer that we fix the CSS rather than duplicate code

garyhtou:
I do agree that this page doesn't need all the comment features. However, I think duplicating the code will lead to it being difficult to maintain

sampoder:
This leads on to run on effects, if we include the CSS for comments, suddenly you have to include the CSS for cards, then you need to include the CSS for hover effect for reactions. And at this point you start to see other parts of interface not looking as intended.

:/ we've dug ourselves into a hole by having too entirely separate style systems.

garyhtou:
Thoughts on using a param to disable [reactions](https://github.com/hackclub/bank/blob/5c8d280ed1e4504ed2139a5d5b44ea739123d094/app/views/comments/_item.html.erb#L33)?

I think with reactions hidden, the rest of the comment will look fine without css

sampoder:
I mean they're better with out the reactions... but they still look like imperfect? The borders for example are off. Idk, I'm fine with either but I prefer stripping away all the styles because it made it like every other element on the page. We're duplicating a small subsection of the original code, about 10 lines from a 84 line partial.

garyhtou (approved):
True, this is fairly minimal duplication

PR #8194, opened 2024-10-18
Default to weekly receipt reports for new users

sampoder:


sampoder:
Unfortunately there's no real way of finding people who haven't updated this field before. We don't have Papertrail on the field in the User model. We can do it jankily by changing everyone with `0` to `1` but that doesn't feel great to me.

garyhtou:
Cool, that's all good with me.

garyhtou:
@dwahbe thoughts on including this feature in the next newsletter to remind people it exists?

dwahbe:
I think that makes sense for the newsletter. I also think we could start linking Help Center articles on HCB. For example, we could link to a Help Center article about receipt uploads by responding to SMS / emails here. Could do the same for this as well. 

Screenshot 2024-10-18 at 11 32 40 AM

PR #8196, opened 2024-10-18
[User] Add Paper Trail

garyhtou:
## Summary of the problem

We really should be tracking changes to the user model. Example of a helpful use case: https://github.com/hackclub/hcb/pull/8194#issuecomment-2421559419

## Describe your changes

add `has_paper_trail` and skip the encrypted attributes (e.g. birthday)


sampoder (approved):
Yes please

PR #8204, opened 2024-10-19
[Wires] Create a globe icon & use it for wires

manuthecoder:
### Changes
This PR creates a globe icon and changes the "International Wire" button accordingly.

### Preview
image

See #8162


dwahbe:
Love this! Could we include this in UI3 with the other icon changes?

garyhtou (approved):
@manuthecoder, this looks good!

Hey @dwahbe, international wires, and this icon update will be separate from UI3. Wires is currently in closed beta, so we're still actively making improvements to polish it up. I'm going to merge this so that we can continue iterating and improving the feature.

I'm happy to discuss whether we want to release the International Wires feature as a part of UI3, however, I want to keep it separate since UI3 is primarily a UI refresh. Keeping this new financial feature separate will give us more granular control over its rollout and prevent adding to an already large UI3 project scope.

PR #8206, opened 2024-10-21
Typo and grammar fixes in nav

ajsirota:
## Summary of the problem
This hurts my eyes /j
Your organization isn't eligble for Google Workspace.

## Describe your changes
I fixed the grammar here. 

Note these changes are untested, but I don't see why it would break anything.


app/views/events/_nav.html.erb
```diff
@@ -92,14 +92,14 @@
       selected: local_assigns[:selected] == :team %>
     <%= dock_item "Perks",
       event_promotions_path(event_id: @event.slug),
-      tooltip: (!policy(@event).promotions? ? "Your account isn't eligble for receive promos & discounts" : "Receive promos & discounts"),
+      tooltip: (!policy(@event).promotions? ? "Your org isn't eligible for perks" : "Receive promos & discounts"),
```
garyhtou: Can we keep the promos/promotions & discounts? It's a bit more descriptive than "perks".

garyhtou (changes requested):
Hi @ajs256, thanks for the PR! Great catch on the typo.

In addition to the comment I've left below, can we also fully spell out "organization"? We generally don't abbreviate that term. Thanks!

garyhtou (approved):
ty!

PR #8218, opened 2024-10-21
[Stripe Card] Remove unused `ignored_columns` for `activated`

garyhtou:
Related to https://github.com/hackclub/hcb/issues/5313


garyhtou:
FYI @rluodev, after dropping the column from the DB, we should follow up with removing this line.

PR #8220, opened 2024-10-22
Fix color scheme for root element

manuthecoder:
Very small fix, but this PR moves the `color-scheme` attribute to the `<html />` element so it ensures that all scrollbars are consistent in dark mode

| Before | After |
|--------|--------|
| image| image|


sampoder (approved):

sampoder:
Good catch; thanks Manu!

garyhtou:
very good work here @manuthecoder!

PR #8225, opened 2024-10-22
[WIP] Add `extracted_total_amount_currency` to `Receipt`

### LINKED ISSUE #8207: [Receipt] Add `extracted_total_amount_currency` field
`extracted_total_amount_currency` will be for `extracted_total_amount_cents`

image

sampoder:
This is a WIP right now! @YodaLightsabr & @garyhtou I'd appreciate feedback on what is used here. Especially about the prompt, I'm not really sure how to do ChatGPT too well.


app/services/receipt_service/suggest.rb
```diff
@@ -43,7 +43,7 @@ def distance(hcb_code)
 
       distances = {
         amount_cents: {
-          value: @extracted.extracted_total_amount_cents && (hcb_code.amount_cents.abs - @extracted.extracted_total_amount_cents.abs).abs == 0 ? 0 : 1,
```
sampoder, line 46: @YodaLightsabr @garyhtou: I'm not sure what to do here, I doubt any currency conversion would be a perfect match. Should we make it somewhat fuzzy?

app/controllers/reimbursement/expenses_controller.rb
```diff
@@ -17,7 +17,10 @@ def create
           attachments: params[:file],
           upload_method: :quick_expense
         ).run!
-        @expense.update(memo: receipt.first.suggested_memo, amount_cents: receipt.first.extracted_total_amount_cents) if receipt.first.suggested_memo
+        @expense.update(
+          memo: receipt.first.suggested_memo,
+          amount_cents: Money.from_cents(receipt.first.extracted_total_amount_cents, receipt.first.extracted_total_amount_currency).cents.to_f / 100
```
garyhtou: I believe if the currency is not USD, we don't want to set the `amount_cents`

sampoder: Ah yes, this is a WIP piece of code. I need to add currency conversion, if we need that?

app/services/receipt_service/extract.rb
```diff
@@ -25,6 +25,7 @@ def run!
 
         subtotal_amount_cents
         total_amount_cents // the amount likely to be charged to a credit card
+        total_amount_currency // the currency that total_amount_cents, if unclear, assume USD
```
garyhtou, line 28: We should clarify which format. It should be [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217)

sampoder: Good idea! Something like?
```suggestion
        total_amount_currency // the currency in ISO 4217 that total_amount_cents, if unclear, assume USD
```
Honestly not sure what works best with GPT. It seemed to be assuming 4217 when testing.

app/controllers/reimbursement/expenses_controller.rb
garyhtou: If we add currency conversion, the conversion rate should be of the date that the purchase as made.

However, I wonder if that may confuse people if the rate has changed a lot since the purchase.

sampoder: Fair point, and this was my worry with https://github.com/hackclub/hcb/pull/8225#discussion_r1810010969. We could require users to enter the amount if the currency isn't USD.

YodaLightsabr:
Closing in favor of #11327

PR #8229, opened 2024-10-22
[Admin][Column Statement] New controller + generate report url on demand

garyhtou:
## Summary of the problem

Kris was running into expired S3 download URLs since the URLs were generated on page load.

## Describe your changes

Now when a user wants to download a report, we generate the S3 URL on the fly and redirect them to it.

I also moved all the Column Statement endpoints to it's own controller.


sampoder:
thank you for this!

PR #8235, opened 2024-10-22
[Event] Drop has_fiscal_sponsorship_document column part 1/3

rluodev:
This part: ignored column

Next part: drop column

Third part: unignore column


sampoder (dismissed):
LGTM, just make sure only defined `self.ignored_columns` once. rn we're doing it twice. same feedback for your other PR

sampoder (approved):

PR #8238, opened 2024-10-22
[Event] Drop `organization_identifier` column part 1/3

rluodev:
This part: removes usages of column and ignores it, and also creates migration to set column to nullable so we don't have issues with creating new `event`s.

Next part: drops the column

Third part: unignores column


db/schema.rb
sampoder, line 1: Check this for unrelated changes!

rluodev: fixed!

garyhtou (approved):

PR #8241, opened 2024-10-22
[Event] Drop `transaction_engine_v2_at` and `pending_transaction_engine_at` columns part 1/3

rluodev:
This part: remove usages and ignore columns

next part: drop columns

third part: unignore columns

Since this touches transaction engine cc @garyhtou


app/controllers/events_controller.rb
sampoder, line 1: We should probably drop `using_transaction_engine_v2?` from `ApplicationHelper`

rluodev: dropped them from applicationcontroller if that's what you meant

app/views/transactions/show.html.erb
```diff
@@ -142,13 +142,13 @@
...
-      </p>
-  <% end %>
+  <% unless @transaction.is_a?(CanonicalTransaction) %>
+    <% admin_tool("py0") do %>
+        <p>
+          Bank Account:
+          <%= link_to "##{@transaction.bank_account.id} (#{@transaction.bank_account.name})", @transaction.bank_account %>
+        </p>
+    <% end %>
   <% end %>
```
garyhtou: Using the existing logic, this section of code would never show because all orgs use transaction engine v2.

```suggestion
```

```ruby
unless using_transaction_engine_v2? || @transaction.is_a?(CanonicalTransaction)
if !(using_transaction_engine_v2? || @transaction.is_a?(CanonicalTransaction))
if !using_transaction_engine_v2? && !@transaction.is_a?(CanonicalTransaction)
if false && !@transaction.is_a?(CanonicalTransaction)
if false
```

app/views/transactions/show_deprecated.html.erb
```diff
@@ -3,11 +3,9 @@
 
 <%= render "events/nav" if @event %>
 
-<% unless using_transaction_engine_v2? %>
 <% admin_tool("mt3 mb0") do %>
   <%= link_to "All transactions", transactions_path, class: "breadcrumb" %>
 <% end %>
-<% end %>
```
garyhtou: Please drop this since it wouldn't show in production in can cases at the moment.

```ruby
unless using_transaction_engine_v2?
if !using_transaction_engine_v2?
if !true
if false
```

PR #8245, opened 2024-10-23
[Event] Remove usages of `sponsorship_fee` column and drop it part 1/3

rluodev:
this part: remove usages and ignore column

next part: drop column

third part: unignore column


sampoder (approved):

app/helpers/events_helper.rb
```diff
@@ -54,9 +54,6 @@ def transaction_memo(tx) # needed to handle mock data in playground mode
   end
 
   def humanize_audit_log_value(field, value)
-    if field == "sponsorship_fee"
-      return number_to_percentage(value.to_f * 100, significant: true, strip_insignificant_zeros: true)
-    end
```
garyhtou: I think there's a chance we may want to keep this. This data is from paper trail which would still have fields named "sponsorship_fee" in its records.

However, at the same time, this is a rarely used admin feature so I'm okay with the decimal not being formated as a percentage. Thoughts?

PR #8262, opened 2024-10-23
Handle race condition within `HcbCode::Tag::Suggestion`

sampoder:
https://heroku-app97991095.airbrake.io/projects/288439/groups/3909406092562234686?tab=notice-detail


app/controllers/hcb_code/tag/suggestions_controller.rb
```diff
@@ -6,7 +6,7 @@ class SuggestionsController < ApplicationController
       def accept
         suggestion = HcbCode::Tag::Suggestion.find(params[:suggestion_id])
         authorize suggestion
-        suggestion.mark_accepted!
+        suggestion.mark_accepted! unless suggestion.accepted?
```
garyhtou:
```suggestion
        suggestion.mark_accepted! if suggestion.may_mark_accepted?
```

app/controllers/hcb_code/tag/suggestions_controller.rb
```diff
@@ -18,7 +18,7 @@ def accept
       def reject
         suggestion = HcbCode::Tag::Suggestion.find(params[:suggestion_id])
         authorize suggestion
-        suggestion.mark_rejected!
+        suggestion.mark_rejected! unless suggestion.rejected?
```
garyhtou:
```suggestion
        suggestion.mark_rejected! if suggestion.may_mark_rejected?
```

sampoder:
That's what I was thinking of! Just couldn't think of it this morning

PR #8273, opened 2024-10-24
[Stripe Card] Validate unlisted designs are only issued by admins

garyhtou:
## Summary of the problem

Non-admins were able to issue unlisted designs. These are non-common designs that don't belong to a specific organization.

Thanks @24c02 for reporting.

## Describe your changes

Validate that only admins can issue cards with unlisted personalization designs


app/models/stripe_card.rb
```diff
@@ -440,9 +440,22 @@ def only_physical_cards_can_have_personalization_design
...
+    # use the chosen design.
+
     if personalization_design&.event.present? && personalization_design.event != event
       errors.add(:personalization_design, "must be of the same event")
     end
+
+    if personalization_design.unlisted? && !user.admin?
+      # Only admins can issue unlisted designs
+      errors.add(:personalization_design, "is invalid")
+    end
```
sampoder: It'd be nice to issue unlisted designs for others via the console, that's why this was never implemented AFAIK

24c02:
idk how much of an actual issue this is? like it'd be nice to fix but you have to actually guess the perso design ID
(inb4 somebody orders 8000 stripe cards ig)

app/models/stripe_card.rb
garyhtou: We could move this check into the controller to continue allowing console. I also don't think it's a big issue if people issue unlisted designs lol — so maybe we just close this pr 🤷‍♂️

sampoder: Separate to this... I'd love an option for admins to be able to issue cards for others from the UI

garyhtou:
@sampoder @YodaLightsabr I'll leave the final call on this up to yall. You guys have the best understanding of the unlisted feature since yall built it!

sampoder:
I think we close this

garyhtou:
sounds good to me!

PR #8276, opened 2024-10-24
[Event] Remove usages of `sponsorship_fee` column and drop it

rluodev:
this step: drop column

next step: unignore column


db/schema.rb
sampoder, line 1: Could you drop the unrelated changes?

rluodev: oop yes

rluodev: All done!

sampoder (approved):

garyhtou:
🪄

No need for review from me on this!

PR #8277, opened 2024-10-24
Search by name, not email for comment mentions

sampoder:
Closes https://github.com/hackclub/hcb/issues/8264


app/views/comments/_form.html.erb
```diff
@@ -15,7 +15,7 @@
 
   <ul role="listbox" class="mention-suggestions card" hidden popover>
     <% commentable.comment_mentionable(current_user:).each do |u| %>
-      <li role="option" class="cursor-pointer mention-suggestion" data-value="@<%= u.email %>" data-search="<%= u.email %>">
+      <li role="option" class="cursor-pointer mention-suggestion" data-value="@<%= u.email %>" data-search="<%= u.name %>">
```
garyhtou: i wonder what if we did both? would it work properly if we had:

```suggestion
      <li role="option" class="cursor-pointer mention-suggestion" data-value="@<%= u.email %>" data-search="<%= "#{u.name} #{u.email}" %>">
```

sampoder: Potentially? But, frankly, the email search was jank and didn't really ever make sense because you couldn't see the emails so you'd get seemingly unrelated results

PR #8282, opened 2024-10-24
Uninstall `lob`

garyhtou:
so long lob! 👋

Usage was removed in https://github.com/hackclub/hcb/pull/6070 https://github.com/hackclub/hcb/issues/3601


garyhtou:
We used Lob for printing and mailing checks before March 2023. 🪦

PR #8284, opened 2024-10-24
Uninstall `scientist` and related gems

garyhtou:
No longer used. I'm keeping the database tables though — mainly out of laziness


garyhtou:
Here's a ticket to drop the tables: https://github.com/hackclub/hcb/issues/8285

PR #8295, opened 2024-10-25
Remove `Commentable` from `Donation`

sampoder:
https://github.com/hackclub/hcb/issues/8280

All comments were moved over and backups are available in the shared drive.


garyhtou:
good call on the backup

PR #8296, opened 2024-10-25
Drop HCB Wrapped 2023 and `hcb-wrapped` NPM package

garyhtou:
## Summary of the problem

we shouldn't have this npm package installed (and code) if we're not using it at the moment

related to https://github.com/hackclub/hcb/issues/8289, but not closing since we still need to do a manual pass on all of our packages

## Describe your changes

_gone with the wind_ 💨


garyhtou:
We can bring this back for 2024 if needed!

PR #8312, opened 2024-10-26
Drop v1 Transaction's Money Received Mailer

garyhtou:
they aren't used since v1 transactions are dead


garyhtou:
^ me messing with branch protection/rules

Issue #8317, opened 2024-10-26
Fix incorrect job namespacing


garyhtou:
We have a lot of job classes namespaced incorrectly.

For example, given a model called `Cardboard`, we currently might have a `CardboardJob::Recycle`. However, it should really be called `Cardboard::RecycleJob`.

Rename: `CardboardJob::Recycle` → `Cardboard::RecycleJob`.

More examples (from codebase):
- `AchTransferJob::Daily` → `AchTransfer::DailyJob`
- `DonationJob::Nightly` → `Donation::NightlyJob`

When renaming these, be careful if the job is currently in Redis (e.g. scheduled/retrying jobs). When a job is stored in Redis and later run, Sidekiq will attempt to call its old name; which will fall if it no longer exists. As a result, you should do: https://github.com/sidekiq/sidekiq/wiki/FAQ#how-do-i-safely-rename-a-worker

Also, when renaming, you may need to change modules to classes. For example, with
`AchTransferJob::Daily`, `AchTransferJob` is currently a module. But when you switch it to `AchTransfer`, it will need to become a class (since it already defined as a class by the model).

garyhtou:
Here is a list of incorrectly namespaced jobs

- [x] `UserJob::SyncToLoops` https://github.com/hackclub/hcb/pull/8318
- [ ] `TransactionGroupingEngineJob::Nightly`
- [ ] `TransactionEngineJob::Nightly`
- [ ] `StripeCardJob::Nightly`
- [ ] `StripeAuthorizationJob::CreateFromWebhook`
- [ ] `StatsdJob::GaugeOnlineUserCount`
- [ ] `ReimbursementJob::Nightly`
- [ ] `ReceiptReportJob::Weekly`
- [ ] `ReceiptReportJob::Send`
- [ ] `ReceiptReportJob::Monthly`
- [ ] `ReceiptJob::SuggestPairings`
- [ ] `ReceiptJob::ExtractTextualContent`
- [ ] `PendingTransactionEngineJob::Nightly`
- [ ] `PendingEventMappingEngineJob::Nightly`
- [ ] `PayoutJob::Nightly`
- [ ] `PayoutJob::Invoice`
- [ ] `PayoutJob::Donation`
- [ ] `MetricJobs::CalculateSubjects`
- [ ] `MetricJobs::CalculateSingle`
- [ ] `MetricJobs::CalculateAppWide`
- [ ] `InvoiceJob::OpensToPaids`
[...]

PR #8318, opened 2024-10-26
`UserJob::SyncToLoops` -> `User::SyncToLoopsJob`

garyhtou:
https://github.com/hackclub/hcb/issues/8317

I'm not keeping the old constant alive because this is a cron job that has already ran. There won't be anything in Sidekiq's Redis upon merging/deploying.


garyhtou:
(i'm testing merge queue using this pr)

PR #8321, opened 2024-10-26
Add caching to event balance frame

YodaLightsabr:
## Summary of the problem
Sometimes users are left with a "—" on the homepage for quite a while while we compute organization balances.
image

## Describe your changes
This PR uses the `cached-frame` stimulus controller to load a cached value while turbo fetches the latest value


sampoder (approved):
LGTM!

PR #8332, opened 2024-10-27
Remove async frames from the codebase

sampoder:
Going forward: use Turbo Frames! They can be lazy loaded.

https://turbo.hotwired.dev/handbook/frames


garyhtou:
great job here!

PR #8333, opened 2024-10-28
Waive wire minimum for HQ projects

sampoder:


sampoder:
Plan is to switch to a `wire_minimum_exempt?` method on `Event::Plan`

PR #8336, opened 2024-10-28
Create `send_wire!` on `Wire`

sampoder:
This is the first step towards automating all wires sent through HCB


app/models/wire.rb
```diff
@@ -448,4 +448,42 @@ def self.recipient_information_accessors
 
   store :recipient_information, accessors: self.recipient_information_accessors
 
+  def send_wire!
+    return unless may_mark_approved?
+
+    account_number_id = event.column_account_number&.column_id ||
+                        Rails.application.credentials.dig(:column, ColumnService::ENVIRONMENT, :default_account_number)
+
+    column_wire_transfer = ColumnService.post("/transfers/international-wire", {
```
garyhtou: Can we add a column_id and backfill any wires in prod that have been sent via column?

sampoder: That's on my list to do already! We haven't sent any yet.

PR #8337, opened 2024-10-28
Move wire minimum amount to `Event::Plan`

garyhtou:
## Summary of the problem

`Event::Plan` essentially serves as authorization for Events. `minimum_wire_amount_cents` is kinda in that category. As a result, I think it should belong in `Event::Plan` rather than `Event`.

## Describe your changes

Closes https://github.com/hackclub/hcb/issues/8335

Also handles https://github.com/hackclub/hcb/pull/8333 by overriding the inherited `minimum_wire_amount_cents` for `Event::Plan::HackClubAffiliate`. It's a bit cleaner this way.


garyhtou:
https://github.com/hackclub/hcb/issues/8335#issuecomment-2440627284

PR #8340, opened 2024-10-28
Separate creation of a counterparty from the POST to `/transfers/international-wire`

sampoder:
So Column let's you do that for ACHs but not wires.... ugh.


app/models/wire.rb
```diff
@@ -454,26 +454,30 @@ def send_wire!
...
+          postal_code: address_postal_code,
+          country_code: recipient_country
+        }
+      }.merge(recipient_information.compact_blank)
+    }.compact_blank)
+
     column_wire_transfer = ColumnService.post("/transfers/international-wire", {
-      idempotency_key: self.id.to_s,
-      amount:,
+      idempotency_key: id.to_s,
```
```suggestion
      idempotency_key: self.id.to_s,
```

app/models/wire.rb
```diff
@@ -454,26 +454,30 @@ def send_wire!
     account_number_id = event.column_account_number&.column_id ||
                         Rails.application.credentials.dig(:column, ColumnService::ENVIRONMENT, :default_account_number)
 
+    column_counterparty = ColumnService.post("/counterparties", {
+      idempotency_key: id.to_s,
```
```suggestion
      idempotency_key: self.id.to_s,
```

app/models/wire.rb
```diff
@@ -454,26 +454,30 @@ def send_wire!
     account_number_id = event.column_account_number&.column_id ||
                         Rails.application.credentials.dig(:column, ColumnService::ENVIRONMENT, :default_account_number)
 
+    column_counterparty = ColumnService.post("/counterparties", {
+      idempotency_key: self.id.to_s,
+      routing_number_type: "bic",
+      routing_number: bic_code,
+      account_number: account_number,
```
```suggestion
      account_number:,
```

PR #8350, opened 2024-10-28
Add `column_id` to `Wire`

sampoder:


sampoder:
There's one to backfill in production

garyhtou:
thanks!

garyhtou:
Next step will be adding an HCB code and all the transaction engine stuff! fun!

PR #8368, opened 2024-10-30
Replace calls to `hack_club_hq?`

sampoder:
Needs testing.


garyhtou (commented):
thank you for this!

I wonder if we should define a `hack_club_afiliated?` method on `Event::Plan` to prevent directly accessing the model and it looks a bit cleaner. it could also give us flexibility (a hack club affiliated plan with a 2% fee; not saying that we'd ever do that lol)

PR #8371, opened 2024-10-30
Create `Event::Plan::SalaryAccount`

sampoder:


sampoder:
I volunteer as tribute!

PR #8372, opened 2024-10-30
[`Event::Plan`] Add papertrail and `inactive_at` column

garyhtou:
Adding papertrail for whodunnit.

And also an `inactive_at` column for AASM to fill in. This allows us to get a concrete "timeline" of which orgs switched from one plan to another — I guess it's currently possible without this column, but a bit more difficult.


sampoder (approved):
LGTM!

PR #8385, opened 2024-10-30
[`CountryEnumable`] Add back Tunisia (`TN`)

garyhtou:
## Summary of the problem

It looks like it was accidentally removed in https://github.com/hackclub/hcb/pull/2018/files. Thanks to @rluodev for reporting that it was missing.

## Describe your changes

I've verified that Tunisia was the only country missing.

```ruby
ISO3166::Country.all.map(&:alpha2) - Event.send(:country_enum_list).keys.map(&:to_s)
#=> ["TN"]
```

```ruby
Event.send(:country_enum_list).sort_by {|c| c.last}.each_cons(2) {|cs| ap cs unless cs.first.last + 1 == cs.last.last }
```

```
[
    [0] [
        [0] :TT,
        [1] 227
    ],
    [1] [
        [0] :TR,
        [1] 229
    ]
]
```


garyhtou:
To prevent stuff like this from happening and to clean up the code, I've created https://github.com/hackclub/hcb/issues/8386

PR #8402, opened 2024-11-03
Use `find_by(transaction_source_type: "RawStripeTransaction")` on `raw_stripe_transaction` for `HcbCode`

sampoder:


app/models/hcb_code.rb
```diff
@@ -207,7 +207,7 @@ def pretty_title(show_event_name: true, show_amount: false, event_name: event.na
   end
 
   def raw_stripe_transaction
-    ct&.raw_stripe_transaction
+    ct&.raw_stripe_transaction || ct2&.raw_stripe_transaction
```
```suggestion
    canonical_transactions.find_by(transaction_source_type: "RawStripeTransaction")&.raw_stripe_transaction
```

PR #8409, opened 2024-11-03
Let people know when their transactions have waived fees!

sampoder:
Because who doesn't like money.


app/views/hcb_codes/_donation.html.erb
```diff
@@ -139,10 +139,21 @@
...
     <section class="card__banner card__darker secondary border-top">
       <p class="my0 italic">
+        <% if @hcb_code.fee_waived? %>
+          Fees were waived on this transaction — enjoy!
+        <% end %>
         Incoming donations are immediately reflected in your account balance.
       </p>
     </section>
   <% end %>
+
```
sampoder:
```suggestion

```

PR #8433, opened 2024-11-05
Expand `HcbCode#fee_waived?` to take into account CT's fees being waived

sampoder:
Closes https://github.com/hackclub/hcb/issues/8432


app/models/hcb_code.rb
```diff
@@ -530,7 +530,7 @@ def pinnable?
   end
 
   def fee_waived?
-    pt&.fee_waived? || false
+    pt&.fee_waived? || ct&.fee&.waived? ||  false
```
```suggestion
    pt&.fee_waived? || ct&.fee&.waived? || false
```

PR #8434, opened 2024-11-05
Emphasise the submit action in the reimbursements instructions box

sampoder:
Closes https://github.com/hackclub/hcb/issues/8423


app/views/reimbursement/reports/show.html.erb
```diff
@@ -105,15 +105,15 @@
         Paid for something out of pocket? Get reimbursed by uploading your receipts.
       </p>
       <p>
-        Begin by adding an expense to this report below; for each expense, you'll need to:
+        Begin by adding an expense for each purchase below; for each expense, you'll need to:
```
sampoder:
```suggestion
        Begin by adding an expense to this report below. Each purchase needs to be a separate expense and for each expense, you'll need to:
```

PR #8438, opened 2024-11-05
[Event] Drop `transaction_engine_v2_at` and `pending_transaction_engine_at` columns part 2/3

rluodev:
this part: drop columns

next part: unignore columns


app/models/event.rb
```diff
@@ -20,7 +20,6 @@
 #  donation_reply_to_email                      :text
 #  donation_thank_you_message                   :text
 #  expected_budget                              :integer
-#  has_fiscal_sponsorship_document              :boolean
```
garyhtou: Please drop unrelated changes. ty!

rluodev: fixed!

garyhtou (approved):
amazing

PR #8439, opened 2024-11-05
Validate combined length of line 1 & 2 in `IncreaseCheck`

sampoder:
Closes https://github.com/hackclub/hcb/issues/8345


app/models/increase_check.rb
```diff
@@ -121,6 +121,13 @@ class IncreaseCheck < ApplicationRecord
     end
   end
 
+  validate do
+    if (address_line1.length + address_line2.length) > 50
+      errors.add(:address_line1, "Address line one and line two's combined length can not exceed 50 characters.")
+      errors.add(:address_line2, "Address line one and line two's combined length can not exceed 50 characters.")
```
```suggestion
      errors.add(:base, "Address line one and line two's combined length can not exceed 50 characters.")
```

I think with the current code, it would show the error twice

sampoder: So we actually don't show any errors at the moment, which is holding this up from being merged

PR #8457, opened 2024-11-06
Fix blankslate turbo stream for reimbursement reports

sampoder:
Closes https://github.com/hackclub/hcb/issues/8390


app/controllers/reimbursement/expenses_controller.rb
```diff
@@ -114,7 +114,7 @@ def total_turbo_stream
     end
 
     def blankslate_turbo_stream
-      turbo_stream.replace(:blankslate, partial: "reimbursement/reports/blankslate", locals: { report: @expense.report })
+      turbo_stream.replace(:blankslate, partial: "reimbursement/reports/blankslate", locals: { report: Reimbursement::Report.find(@expense.reimbursement_report_id) })
```
garyhtou: Would `@expense.report.reload` also work?

PR #8464, opened 2024-11-07
[Reimbursement] Skip Report broadcast on touches

garyhtou:
Here's a chance this might address https://github.com/hackclub/hcb/issues/8389. I'm not sure I completely understand the root of the issue, so this is kinda a shot in the dark.

@sampoder can you review and take ownership of this PR? Feel free to close if it's not helpful.


app/models/reimbursement/expense.rb
```diff
@@ -54,12 +54,6 @@ class Expense < ApplicationRecord
     validates :expense_number, uniqueness: { scope: :reimbursement_report_id }
     validate :valid_expense_type
 
-    # I'm not using broadcasts_refreshes_to here. That's because "after_commit" is called on a touch, "after_save" isn't.
-    # View https://github.com/hackclub/hcb/issues/8389 for more context - @sampoder
-
-    after_save -> { broadcast_refresh_later_to(self.report) }
-    after_destroy -> { broadcast_refresh_later_to(self.report) }
-
```
sampoder: Is there a reason why this was removed? If an expense touches a report we still want to refresh the report. The only time we don't want to is when the receipt touches the expense

garyhtou: From my understanding, the `touch: true on` `expense.rb:35` will already cause Report to broadcast refresh.

https://rubydoc.info/github/hotwired/turbo-rails/main/Turbo/Broadcastable#:~:text=This%20works%20great%20in%20hierarchical%20structures%2C%20where%20the%20child%20record%20touches%20parent%20records%20automatically%20to%20invalidate%20the%20cache%3A

[`f688e5a` (#8464)](https://github.com/hackclub/hcb/pull/8464/commits/f688e5a5f964055e04d55c74fb5bd9dba616b34f)
> Since Expense touches Report, and Report has it's own broadcast refresh, we can drop these broadcasts.

garyhtou: >The only time we don't want to is when the receipt touches the expense

I'm a bit confused by this. From my understanding, the Expense touches the Report.
```ruby
class Expense < ApplicationRecord
  belongs_to :report, inverse_of: :expenses, foreign_key: "reimbursement_report_id", touch: true
end
# https://github.com/hackclub/bank/blob/f688e5a5f964055e04d55c74fb5bd9dba616b34f/app/models/reimbursement/expense.rb#L35
```

I'm not seeing that Report touches Expenses
https://github.com/hackclub/bank/blob/a0001cdded36f996d26d0e2523f112d0d58096e6/app/models/reimbursement/report.rb#L61

sampoder: Sorry, so we still want refreshes when the expense is edited. However, we don't want them when a receipt is uploaded. Reading this PR, I think that we wouldn't get refreshes when the expense is edited because that's also a touch.

sampoder (approved):
LGTM! Just added back refreshes when the expense is edited (non touched)

garyhtou:
Got it, thanks. Instead of having Expense broadcast refresh to Report, should we broadcast updates to specific expenses?

PR #8480, opened 2024-11-08
Validate plan types on `Event::Plan`

sampoder:
Because this takes down seemingly everything if it's not valid...


app/models/event/plan.rb
```diff
@@ -72,6 +72,12 @@ def self.available_plans
       end
     end
 
+    validate do
+      unless plan_type.in?(Event::Plan.descendants.collect{|d| d.name})
```
garyhtou:
```suggestion
      unless plan_type.in?(Event::Plan.descendants.map(&:name))
```

app/models/event/plan.rb
```diff
@@ -72,6 +72,12 @@ def self.available_plans
...
+        errors.add(:base, "Invalid plan type.")
```
garyhtou:
```suggestion
        errors.add(:plan_type, "is invalid")
```

PR #8485, opened 2024-11-11
Waive fees on bank fee related CPTs

sampoder:
This is for fee credits, they are positive so HCB assumes we should charge a fee on them


garyhtou (approved):
LGTM

PR #8497, opened 2024-11-12
Render active Stripe cards first on `/my/cards`

sampoder:
Closes https://github.com/hackclub/hcb/issues/8441


garyhtou:
thank you for this!

PR #8500, opened 2024-11-13
Create an action bar on the ledger

sampoder:
Screenshot 2024-11-12 at 9 00 40 PM

I love the new filter design; I added it to both the exports button & the tag toggle!


garyhtou:
Could we move the filter button to be immediately left of the search? It'll keep all the filter-related elements close together.

My preferred order is (left to right):
1. Search bar
2. Filter button
3. Tag toggle
4. Export

PR #8517, opened 2024-11-13
[AdminController] Switch default layout to "admin"

KiKoS0:
Closes: https://github.com/hackclub/hcb/issues/8230
## Summary of the problem
Sets the default layout of `AdminController` and all its subclasses to "admin".

## Describe your changes
I tested all the pages individually other than the reimbursements one since it expects the clearing house event:
https://github.com/hackclub/hcb/blob/dbade2ffb8de93157be84fc8201f453e64a65ac1/app/controllers/admin_controller.rb#L397

This also fixes a minor bug with the active tab styling not being applied correctly when the `Wires` tab is active:

image


sampoder (approved):
Thank you so much for taking this on! And good catch on the wires thing.

albertchae (commented):
cool, glad none of the endpoints were actually depending on the application layout 😆

garyhtou (commented):
Amazing! Thank you!

PR #8534, opened 2024-11-14
Switch `robotics_team?` over to `EventTag`

sampoder:
We edge closer and closer.


sampoder:
Production has been backfilled!

PR #8556, opened 2024-11-16
[Transaction Engine] Fix SQL injection bug

rluodev:
## Summary of the problem

We were using string interpolation instead of question mark interpolation which left us open to an sql injection bug.

## Describe your changes

Change string interpolation to question mark interpolation


app/services/pending_transaction_engine/pending_transaction/all.rb
```diff
@@ -57,23 +57,23 @@ def canonical_pending_transactions
             if @user
               cpts =
                 cpts.joins("LEFT JOIN raw_pending_stripe_transactions on raw_pending_stripe_transactions.id = canonical_pending_transactions.raw_pending_stripe_transaction_id")
-                    .where("raw_pending_stripe_transactions.stripe_transaction->>'cardholder' = '#{@user&.stripe_cardholder&.stripe_id}'")
+                    .where("raw_pending_stripe_transactions.stripe_transaction->>'cardholder' = '?'", @user&.stripe_cardholder&.stripe_id)
```
garyhtou: I think the quotes wrapping the ? isn't necessary

garyhtou: same for lines 72 and 76

sampoder (approved):

garyhtou:
Great! Thanks for testing @sampoder. I'm glad the quotes didn't cause an issue!

PR #8561, opened 2024-11-16
[Transaction Engine] Use Question mark interpolation

rluodev:
Fix instances where we were using string interpolation


app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -87,16 +87,19 @@ def search_modifier_for(type)
...
         type = type.to_s
+        search_query = "%#{@search}%"
 
-        return "and (#{type}.memo ilike '%#{@search}%' or #{type}.friendly_memo ilike '%#{@search}%' or #{type}.custom_memo ilike '%#{@search}%')" if type == "ct"
+        if type == "ct"
+          return ActiveRecord::Base.sanitize_sql_array(["and (ct.memo ilike ? or ct.friendly_memo ilike ? or ct.custom_memo ilike ?)", search_query, search_query, search_query])
+        end
 
-        "and (#{type}.memo ilike '%#{@search}%' or #{type}.custom_memo ilike '%#{@search}%')"
+        ActiveRecord::Base.sanitize_sql_array(["and (#{type}.memo ilike ? or #{type}.custom_memo ilike ?)", search_query, search_query])
```
garyhtou: I'd suggest checking out [the docs](https://api.rubyonrails.org/v7.1/classes/ActiveRecord/Sanitization/ClassMethods.html#method-i-sanitize_sql_array). Maybe using keyword arguments will prevent us from having to duplicate arguments multiple times

rluodev: Ah I see what you mean I'll make that change today

garyhtou (commented):
`@search` was already escaped on line 8. Since we escaping on interpolation, we can remove the `quote_string` call on line 8.

rluodev:
@garyhtou ready for review!

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -88,15 +88,17 @@ def search_modifier_for(type)
 
         type = type.to_s
 
-        return "and (#{type}.memo ilike '%#{@search}%' or #{type}.friendly_memo ilike '%#{@search}%' or #{type}.custom_memo ilike '%#{@search}%')" if type == "ct"
+        if type == "ct"
+          return ActiveRecord::Base.sanitize_sql_array(["and (ct.memo ilike %:query% or ct.friendly_memo ilike %:query% or ct.custom_memo ilike %:query%)", { query: @search }])
```
garyhtou: I believe this line doesn't work as expected because it generates SQL where the `%` is outside of the quotes.

```ruby
ActiveRecord::Base.sanitize_sql_array(["and (ct.memo ilike %:query% or ct.custom_memo ilike %:query%)", { query: "this is my search" }])
#=> "and (ct.memo ilike %'this is my search'% or ct.custom_memo ilike %'this is my search'%)"
```

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -88,15 +88,17 @@ def search_modifier_for(type)
...
+        end
 
-        "and (#{type}.memo ilike '%#{@search}%' or #{type}.custom_memo ilike '%#{@search}%')"
+        ActiveRecord::Base.sanitize_sql_array(["and (#{type}.memo ilike %:query% or #{type}.custom_memo ilike %:query%)", { query: @search }])
```
garyhtou: Escaping statements for LIKE/ILIKE is tricky since not only does it need to escape quotes, but it also needs to escape `%` since otherwise, it'd act like a wild card.

```ruby
ActiveRecord::Base.sanitize_sql_array(["and (ct.memo ilike %:query% or ct.friendly_memo ilike %:query% or ct.custom_memo ilike %:query%)", { query: "what is 10 % 3?" }])
#=> "and (ct.memo ilike %'what is 10 % 3?'% or ct.friendly_memo ilike %'what is 10 % 3?'% or ct.custom_memo ilike %'what is 10 % 3?'%)"
```

Given the string "what is 10 % 3?", it doesn't escape the `%` which means that strings such as "what is 10 hello 3" would match.

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -88,15 +88,17 @@ def search_modifier_for(type)
...
       end
 
       def user_modifier
         return "" unless @user&.stripe_cardholder&.stripe_id.present?
 
-        "and raw_stripe_transactions.stripe_transaction->>'cardholder' = '#{@user.stripe_cardholder.stripe_id}'"
+        ActiveRecord::Base.sanitize_sql_array(["and raw_stripe_transactions.stripe_transaction->>'cardholder' = ?", @user.stripe_cardholder.stripe_id])
```
garyhtou: good!

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -112,13 +114,15 @@ def user_joins_for(type)
       def modifiers
         joins = []
         conditions = []
+        query_params = []
```
garyhtou: I believe it's more robust if we use the hash syntax here. With the array syntax, if we accidentally miss a `query_params`, variables will be interpolated into the wrong slots.

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -294,6 +315,8 @@ def canonical_transactions_grouped_sql
           #{modifiers}
           order by date desc, pt_ids[1] desc, ct_ids[1] desc
         SQL
+
+        ActiveRecord::Base.sanitize_sql_array([q, { event_id: @event_id }])
```
garyhtou: amazing; this is good

garyhtou (commented):
Two categories of comments:
- Escaping LIKE/ILIKE is hard. I left two comments about this
- Instead of using the array syntax for escaping `query_params`, I suggest using the hash syntax. See comment for a deeper explanation

app/services/transaction_grouping_engine/transaction/all.rb
rluodev: I tried to fix it, can you lmk if my change works?

app/services/transaction_grouping_engine/transaction/all.rb
rluodev: I tried escaping all % using `gsub("%", '\\%')` in the original assignment for @search variable in my most recent change

app/services/transaction_grouping_engine/transaction/all.rb
rluodev: done! not sure if completely correct tho

rluodev:
hi @garyhtou ready for rereview

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -112,13 +114,15 @@ def user_joins_for(type)
...
+        query_params = []
 
         if @tag_id
           joins << <<~SQL
             left join hcb_codes on hcb_codes.hcb_code = q1.hcb_code
             left join hcb_codes_tags on hcb_codes_tags.hcb_code_id = hcb_codes.id
           SQL
-          conditions << "hcb_codes_tags.tag_id = #{@tag_id}"
+          conditions << "hcb_codes_tags.tag_id = ?"
+          query_params << @tag_id
```
garyhtou: Should this be a hash too?

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -136,14 +140,31 @@ def modifiers
...
+        end
+
+        if @start_date
+          conditions << "#{date_select} >= cast(:start_date as date)"
+          query_params << { start_date: @start_date }
+        end
+
+        if @end_date
+          conditions << "#{date_select} <= cast(:end_date as date)"
+          query_params << { end_date: @end_date }
```
garyhtou: It looks like we're creating an array of hashes. Should we merge the hashes into one large one instead?

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -5,7 +5,7 @@ module Transaction
     class All
       def initialize(event_id:, search: nil, tag_id: nil, expenses: false, revenue: false, minimum_amount: nil, maximum_amount: nil, start_date: nil, end_date: nil, user: nil, missing_receipts: false)
         @event_id = event_id
-        @search = ActiveRecord::Base.connection.quote_string(search || "")
+        @search = search.gsub("%", '\\%') || ""
```
garyhtou: Would something like https://api.rubyonrails.org/v7.1/classes/ActiveRecord/Sanitization/ClassMethods.html#method-i-sanitize_sql_like be better?

garyhtou (approved):

PR #8570, opened 2024-11-18
Remove ignored columns statement from AchTransfer

sampoder:
It's been removed now


sampoder:
(the column has been removed)

PR #8573, opened 2024-11-18
Add click-to-copy in a couple places

24c02:
1. card grants didn't let you click to copy their stripe card details as a grantee. 
 they definitely should, as their primary purpose is being pasted into order forms :-P
2. regular stripe cards didn't let you click to copy ZIP/exp date, 2 other things i often find myself wanting to paste
3. as someone who needs to grab network IDs/category codes for issuing grants, being able to copy them from the tx would be nice ^_^


sampoder (approved):
LGTM!

PR #8584, opened 2024-11-19
[v4 API][Card Grant] Add topup endpoint and show disbursements

24c02:
i want these for high seas:
1. adds topup route! this will let me make subsequent orders increase a single grant rather than creating a bunch of cards i'll get asked to merge later
2. adds information on created disbursement transactions to create and topup – that way i can rename grants on the ledger to reflect their purpose when they go out


app/controllers/api/v4/card_grants_controller.rb
```diff
@@ -31,6 +31,19 @@ def show
         authorize @card_grant
       end
 
+      def topup
+        @card_grant = CardGrant.find_by_public_id(params[:id])
```
```suggestion
        @card_grant = CardGrant.find_by_public_id!(params[:id])
```

Technically, all the `find_by_public_id` calls in this controller should have a bang.

app/models/card_grant.rb
```diff
@@ -114,8 +114,9 @@ def pending_invite?
   def topup!(amount_cents:, topped_up_by: User.find(sent_by_id))
     raise ArgumentError.new("Topups must be positive.") unless amount_cents.positive?
 
+    new_disbursement = nil
     ActiveRecord::Base.transaction do
-      DisbursementService::Create.new(
+      new_disbursement = DisbursementService::Create.new(
```
garyhtou: What if instead of creating this new local variable, we switched the order of the statements?

`update!` first then call `DisbursementService::Create`. Since that's the last one it'll be what's returned by the method.

config/routes.rb
```diff
@@ -567,7 +567,11 @@
           end
         end
 
-        resources :card_grants, only: [:show]
+        resources :card_grants, only: [:show] do
+          member do
+            post "topup"
+          end
```
garyhtou: amazing! great job using a member route for this

app/controllers/api/v4/card_grants_controller.rb
```diff
@@ -31,6 +31,19 @@ def show
...
+
+        authorize @card_grant
+        begin
+          new_disbursement = @card_grant.topup!(amount_cents: params["amount_cents"], topped_up_by: current_user)
+        rescue ArgumentError => e
+          return render json: { error: "invalid_operation", messages: [e.message] }, status: :bad_request
+        end
```
garyhtou: Great error handling here

app/controllers/api/v4/card_grants_controller.rb
```diff
@@ -31,6 +31,19 @@ def show
...
+
+        render :topup, locals: { new_disbursement: }
```
garyhtou: "new" disbursement is semi-misleading since it implies that it hasn't been saved to the database. I'd just drop the word "new" and call it `disbursement`.

```suggestion
          disbursement = @card_grant.topup!(amount_cents: params["amount_cents"], topped_up_by: current_user)
        rescue ArgumentError => e
          return render json: { error: "invalid_operation", messages: [e.message] }, status: :bad_request
        end

        render :topup, locals: { disbursement: }
```

app/views/api/v4/card_grants/_card_grant.jbuilder
```diff
@@ -4,3 +4,4 @@ json.organization card_grant.event, partial: "api/v4/events/event", as: :event i
 json.amount_cents card_grant.amount_cents
 json.merchant_lock card_grant.merchant_lock
 json.category_lock card_grant.category_lock
+json.disbursement (local_assigns[:new_disbursement] || card_grant.disbursement), partial: "api/v4/transactions/disbursement", as: :disbursement if local_assigns[:expand]&.include?(:disbursement)
```
garyhtou: With this code, `disbursement` can contain different data depending on the context:
- During a topup, it's the topup's disbursement — or in other words, the most recent disbursement.
- If you later show a card grant, `disbursement` is the first disbursement.

I think this behavior can get confusing. Instead, I'd suggest returning an array for disbursement sorted by date asc. This will also omit the need for the `new_disbursement` param.

To get disbursements that belong to this card grant, filter the event's disbursement by `subledger_id`.

garyhtou (commented):
I love that you're putting this together! Here are a few comments

app/views/api/v4/card_grants/_card_grant.jbuilder
24c02: oh i actually like that a lot better

app/controllers/api/v4/card_grants_controller.rb
24c02: i don't think i agree with that – this causes the Event.find_by_public_id!(params[:event_id]) || Event.friendly.find(params[:event_id]) to 404, breaking using org slugs

app/controllers/api/v4/card_grants_controller.rb
```diff
@@ -22,15 +22,28 @@ def create
 
         @card_grant.save!
 
-        render "show"
+        render "create"
```
garyhtou: I believe by default, Rails will implicitly render the view with the same name as the action. So for the create action, I'll look for a create view if you don't explicitly call render in the action.

I'm totally cool with keep this line here — just wanted to share that info.

app/models/card_grant.rb
```diff
@@ -115,19 +115,22 @@ def topup!(amount_cents:, topped_up_by: User.find(sent_by_id))
...
       ).run
-
-      update!(amount_cents: self.amount_cents + amount_cents)
     end
   end
 
+  def disbursements
+    self.event.disbursements.where(destination_subledger_id: subledger_id).order(created_at: :desc)
+  end
+
```
garyhtou: Instead of defining this as a method, let's use the power of Rail's associations (it some with handy stuff like caching).

```
has_many :disbursements, ->(record) { where(destination_subledger_id: record.subledger_id) }, through: :event
```

You can put this at the top of the file near all the other associations.

We usually don't include a default order on associations, so we'll call `order(created_at: :desc)` wherever we use it.

app/views/api/v4/card_grants/_card_grant.jbuilder
```diff
@@ -4,3 +4,6 @@ json.organization card_grant.event, partial: "api/v4/events/event", as: :event i
 json.amount_cents card_grant.amount_cents
 json.merchant_lock card_grant.merchant_lock
 json.category_lock card_grant.category_lock
+json.disbursements card_grant.disbursements do |disbursement|
```
garyhtou: See the comment about creating a `disbursements` association

```suggestion
json.disbursements card_grant.disbursements.order(created_at: :desc) do |disbursement|
```

app/views/api/v4/card_grants/_card_grant.jbuilder
```diff
@@ -4,3 +4,6 @@ json.organization card_grant.event, partial: "api/v4/events/event", as: :event i
...
+  json.partial! "api/v4/transactions/disbursement", disbursement:
+end
```
garyhtou: new line — sorry; this is really nit picky

```suggestion
end

```

app/models/card_grant.rb
24c02: ah, thanks! i was wondering if there was a railsier way to do this, forgive me, i'm noob

app/views/api/v4/card_grants/_card_grant.jbuilder
```diff
@@ -4,3 +4,6 @@ json.organization card_grant.event, partial: "api/v4/events/event", as: :event i
 json.amount_cents card_grant.amount_cents
 json.merchant_lock card_grant.merchant_lock
 json.category_lock card_grant.category_lock
+json.disbursements card_grant.disbursements.order(created_at: :desc) do |disbursement|
+  json.partial! "api/v4/transactions/disbursement", disbursement:
+end
```
garyhtou:
```suggestion
if local_assigns[:expand]&.include?(:disbursements)
  json.disbursements card_grant.disbursements.order(created_at: :desc) do |disbursement|
    json.partial! "api/v4/transactions/disbursement", disbursement:
  end
end
```

app/views/api/v4/card_grants/create.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.partial! @card_grant, expand: [:user, :organization, :disbursement]
```
garyhtou:
```suggestion
json.partial! @card_grant, expand: [:user, :organization, :disbursements]
```

app/views/api/v4/card_grants/topup.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.partial! @card_grant, expand: [:user, :organization, :disbursement]
```
garyhtou:
```suggestion
json.partial! @card_grant, expand: [:user, :organization, :disbursements]
```

garyhtou (approved):

PR #8591, opened 2024-11-19
[Disbursements] Improve organization selector

YodaLightsabr:
## Summary of the problem

It can be difficult to select organizations as an admin (especially on mobile) because the native dropdown lists thousands of organizations.

## Describe your changes

I've built a custom dropdown component we can use on the disbursements page (and in the future, elsewhere, too).

For admins, it shows the organization ID and moves the balance into a tooltip.

Here's a video:

https://github.com/user-attachments/assets/0e893cc8-62db-462b-be5e-9ae4086c60f1

And here's a screenshot on dark mode:

image

Reverts hackclub/hcb#8590


app/javascript/controllers/organization_select_controller.js
```diff
@@ -0,0 +1,184 @@
...
+
+      const shown =
+        result.length > 0
+          ? result.map(r => r.obj.organization)
+          : orgValues
+              .sort((a, b) => a.index - b.index)
+              .map(o => o.organization)
+              .slice(0, 50)
+      const hidden = this.organizationTargets.filter(el => !shown.includes(el))
+      console.log('Search took', end - start, 'ms')
```
sampoder:
```suggestion
```

app/controllers/disbursements_controller.rb
```diff
@@ -45,22 +45,28 @@ def new
       name: params[:message]
     )
 
+    # @event = current_user.events.select("id").not_hidden.filter_demo_mode(false)
+
```
sampoder:
```suggestion
```

app/views/disbursements/_form.html.erb
```diff
@@ -1,57 +1,10 @@
 <% disabled = @source_event.present? && !policy(@source_event).create_transfer? %>
 
-<%= form_with(model: disbursement, local: true, url: disbursements_path, html: { "x-data" => "{ fee: null, event_id: null }" }) do |form| %>
+<%= form_with(model: disbursement, local: true, url: disbursements_path, html: { "x-data" => "{fee: null, event_id: null}", "data-turbo" => "false" }) do |form| %>
```
sampoder: Any reason for removing Turbo?

app/views/disbursements/_form.html.erb
```diff
@@ -1,57 +1,10 @@
...
-            <%= form.text_field :event_id, placeholder: "hack_the_seas", required: true, disabled:, data: { behavior: "extract_slug" } %>
-          </div>
-        </template>
-      <% else %>
-        <span class="muted">You can transfer to any organization you're a part of.</span>
-      <% end %>
-    <% end %>
-  </div>
+  <%= render partial: "select_organization", locals: { form: form, field_name: "source_event_id", disabled: disabled, events: @allowed_source_events, default_event: @source_event, sending: true } %>
+  <%= render partial: "select_organization", locals: { form: form, field_name: "event_id", disabled: disabled, events: @allowed_destination_events, default_event: nil, hide_balances: @source_event&.plan&.unrestricted_disbursements_enabled?, receiving: true } %>
```
sampoder: Imo the partial should be `events/_select.html.erb`

app/javascript/controllers/organization_select_controller.js
```diff
@@ -0,0 +1,184 @@
...
+        const oldFieldValue =
+          organizations[this.dropdownTarget.children[1].value]
+        if (oldFieldValue) {
+          Object.assign(oldFieldValue.button.style, {
+            backgroundColor: 'unset',
+            color: 'unset',
+          })
+          oldFieldValue.button.children[1].style.color = ''
+        }
+
```
sampoder:
```suggestion
        const previouslySelected =
          organizations[this.dropdownTarget.children[1].value]
        if (previouslySelected) {
          Object.assign(previouslySelected.button.style, {
            backgroundColor: 'unset',
            color: 'unset',
          })
          previouslySelected.button.children[1].style.color = ''
        }
```

app/javascript/controllers/organization_select_controller.js
```diff
@@ -0,0 +1,184 @@
...
+        Object.assign(button.style, {
+          backgroundColor: 'var(--info)',
+          color: 'white',
+        })
+        button.children[1].style.color = 'white'
```
sampoder, line 120: Is there a reason why we aren't defining these styles in a class and then toggling the class?

app/javascript/controllers/organization_select_controller.js
```diff
@@ -0,0 +1,184 @@
...
+
+        const fieldValue = this.dropdownTarget.children[1]
+        fieldValue.innerText = name
+        fieldValue.value = id
+        fieldValue.dataset.fee = fee
+
+        this.dropdownTarget.value = id
+        this.dropdownTarget.dispatchEvent(new CustomEvent('feechange'))
```
sampoder: Are we using `this.dropdownTarget.children[1]` essentially as a state variable that holds the currently selected event?

sampoder:
Tried testing this in production on my admin account - I seem to only be able to transfer to Hack Club HQ?


https://github.com/user-attachments/assets/c9e36e23-ca65-4a6e-82d8-863df58e1fa9

app/views/disbursements/_select_organization.html.erb
```diff
@@ -0,0 +1,42 @@
+<%# locals: (form:, field_name:, disabled:, events:, default_event:, hide_balances: false, sending: false, receiving: false) %>
+
+<div class="field" data-controller="organization-select">
+    <%= form.label field_name, sending ? "From" : "To" %>
+    <select name="disbursement[<%= field_name %>]" id="disbursement_<%= field_name %>" <%= "disabled" if disabled %> class="bg-[transparent] dark:bg-darkless input input--select flex items-center select-none cursor-pointer disabled:cursor-default" data-organization-select-target="dropdown">
```
sampoder, line 5: Is our intention to be able to use this for things that aren't disbursements?

app/views/disbursements/_form.html.erb
sampoder: Oh I guess we aren't planning to make this a general purpose organisations selector. In that case just `select_event`

app/views/disbursements/_select_organization.html.erb
```diff
@@ -0,0 +1,42 @@
...
+                        <button <%= "disabled" if disabled_message %> aria-label="<%= disabled_message %>" class="<%= "tooltipped tooltipped--i" if disabled_message %> text-[length:inherit] border-none bg-[transparent] w-full flex justify-between p-2 cursor-pointer disabled:cursor-default hover:bg-smoke hover:dark:bg-darkless transition-colors duration-150">
+                            <div class="text-left"><%= event.name %></div>
+                            <div class="text-muted">
+                                <% if current_user.admin? %>
+                                    <%= event.id %>
+                                <% elsif hide_balances %>
+                                    <%= turbo_frame_tag "event_balance_#{event.public_id}", src: event_async_balance_path(event, symbol: true), data: { turbo_permanent: true, controller: "cached-frame", action: "turbo:frame-render->cached-frame#cache" }, loading: :lazy do %>
+                                        <strong>-</strong>
+                                    <% end %>
```
sampoder: What are we trying to achieve here?

app/views/disbursements/_select_organization.html.erb
```diff
@@ -0,0 +1,42 @@
...
+
+    <div data-organization-select-target="wrapper" style="max-width: 384px;">
+        <input data-organization-select-target="search" style="display: none; border-bottom-left-radius: 0px; border-bottom-right-radius: 0px;" type="text" placeholder="Search">
+
+        <div data-organization-select-target="menu" style="display: none; border-width: 1px; border-style: solid; border-top: none; max-width: 384px; border-top-left-radius: 0px; border-top-right-radius: 0px;" class="border-smoke rounded-md dark:border-black overflow-hidden h-40">
+            <div class="w-full h-full rounded-md overflow-y-scroll" style="border-top-left-radius: 0px; border-top-right-radius: 0px; content-visibility: auto;">
+                <% events.each_with_index.map do |event, i| %>
+                    <% disabled_message = nil %>
+                    <% disabled_message = "Insufficient balance" if sending && !current_user.admin? && event.balance_available <= 0 %>
+                    <% disabled_message = "HCB transfers disabled" if sending && event != default_event && !policy(event).create_transfer? %>
```
sampoder: How come we don't disable it if `event == default_event`? I think I might be missing something - is it for members to be able to preview the form?

app/views/disbursements/_select_organization.html.erb
YodaLightsabr: For now, let's leave it as is. It would be cool to expand later but I don't want to DRY it too much preemptively.

app/views/disbursements/_select_organization.html.erb
YodaLightsabr: We don't want to load 5000 balances for non-admins all at once.

app/views/disbursements/_select_organization.html.erb
YodaLightsabr: People can change the to event, it doesn't always match default_event

app/javascript/controllers/organization_select_controller.js
YodaLightsabr: Yes

app/javascript/controllers/organization_select_controller.js
YodaLightsabr: It's 2 styles, and I think we're trying to shift towards tailwind instead of .scss files.

app/views/disbursements/_form.html.erb
YodaLightsabr: This is how it was previously (before the slug change) and it worked fine. Is there any reasoning for adding it back in?

app/views/disbursements/_select_organization.html.erb
sampoder: Yes - but how come if they don't have permission to create the transfer to that event, we don't disable it

sampoder: I'm asking why this line:

```ruby
sending && event != default_event && !policy(event).create_transfer?
```

is not:

```ruby
sending && !policy(event).create_transfer?
```

app/views/disbursements/_select_organization.html.erb
sampoder: Ah - maybe we make the variable `lazy_load_balances` or something, `hide_balances` is confusing

app/javascript/controllers/organization_select_controller.js
sampoder: Imo they should be in classes - I feel like that's cleaner and will make it easier to change down the line. Plus then you can just add and remove the class, not use unset or empty strings

app/views/disbursements/_form.html.erb
sampoder: It was removed in https://github.com/hackclub/hcb/commit/5d357fa269a96e679c9aaa3416f7c706801d9bd2#diff-1c5acab3fa3dfad661adbde557e22b311bde4e7fb1dadafaf85dd5d2f0861edaR3-L9 (the previous time we merged https://github.com/hackclub/hcb/pull/8136), before then we'd been using Turbo on this form. I think we should keep Turbo for consistency across all forms.

YodaLightsabr: Ah got it. I think there was a reason I had removed it, but I'll try testing it with Turbo to see if we can keep it.

sampoder (approved):
Tested and LGTM! Please just keep Turbo on

app/controllers/disbursements_controller.rb
```diff
@@ -45,22 +45,26 @@ def new
       name: params[:message]
     )
 
+    user_event_ids = current_user.events.reorder("organizer_positions.sort_index ASC").includes(organizer_positions: :user).select("events.id").map(&:id)
```
```suggestion
    user_event_ids = current_user.organizer_positions.reorder(sort_index: :asc).pluck(:event_id) 
```

app/controllers/disbursements_controller.rb
```diff
@@ -45,22 +45,26 @@ def new
...
+
     @allowed_source_events = if current_user.admin?
-                               Event.all.reorder(Event::CUSTOM_SORT).includes(:plan)
+                               Event.select("name, id, demo_mode, slug").all.reorder(Event::CUSTOM_SORT).includes(:plan)
```
```suggestion
                               Event.select(:name, :id, :demo_mode, :slug).all.reorder(Event::CUSTOM_SORT).includes(:plan)
```

app/controllers/disbursements_controller.rb
```diff
@@ -45,22 +45,26 @@ def new
...
                              else
                                current_user.events.not_hidden.filter_demo_mode(false)
-                             end
+                             end.sort_by { |e| user_event_ids.index(e.id) || Float::INFINITY }
```
garyhtou: This preserves the sort from custom sort
```suggestion
                             end.to_enum.with_index.sort_by { |e, i| [user_event_ids.index(e.id) || Float::INFINITY, i] }
```

app/controllers/disbursements_controller.rb
```diff
@@ -45,22 +45,26 @@ def new
...
     @allowed_destination_events = if current_user.admin?
-                                    Event.all.reorder(Event::CUSTOM_SORT).includes(:plan)
+                                    Event.select("name, id, demo_mode, can_front_balance, slug").all.reorder(Event::CUSTOM_SORT).includes(:plan)
```
```suggestion
                                    Event.select(:name, :id, :demo_mode, :can_front_balance, :slug).all.reorder(Event::CUSTOM_SORT).includes(:plan)
```

app/controllers/disbursements_controller.rb
```diff
@@ -45,22 +45,26 @@ def new
...
+                                  elsif @source_event&.plan&.unrestricted_disbursements_enabled?
+                                    Event.select("name, id, demo_mode, can_front_balance, slug").transparent.indexable.includes(:plan)
```
```suggestion
                                    Event.select(:name, :id, :demo_mode, :can_front_balance, :slug).indexable.includes(:plan)
```

app/controllers/disbursements_controller.rb
```diff
@@ -45,22 +45,26 @@ def new
...
                                   else
                                     current_user.events.not_hidden.without(@source_event).filter_demo_mode(false)
-                                  end
+                                  end.sort_by { |e| user_event_ids.index(e.id) || Float::INFINITY }
```
```suggestion
                                  end.to_enum.with_index.sort_by { |e, i| [user_event_ids.index(e.id) || Float::INFINITY, i] }
```

app/controllers/disbursements_controller.rb
```diff
@@ -45,22 +45,26 @@ def new
...
                                current_user.events.not_hidden.filter_demo_mode(false)
-                             end
+                             end.sort_by { |e| user_event_ids.index(e.id) || Float::INFINITY }
     @allowed_destination_events = if current_user.admin?
-                                    Event.all.reorder(Event::CUSTOM_SORT).includes(:plan)
+                                    Event.select("name, id, demo_mode, can_front_balance, slug").all.reorder(Event::CUSTOM_SORT).includes(:plan)
+                                  elsif @source_event&.plan&.unrestricted_disbursements_enabled?
+                                    Event.select("name, id, demo_mode, can_front_balance, slug").transparent.indexable.includes(:plan)
                                   else
                                     current_user.events.not_hidden.without(@source_event).filter_demo_mode(false)
```
```suggestion
                                    current_user.events.not_hidden.filter_demo_mode(false)
```

garyhtou: But double check that we validate you can't disburse to the same event

PR #8599, opened 2024-11-20
Stop `Missing target element "button" for "expense-form" controller`

sampoder:
It annoys me.

No seriously... we conditionally render the button and don't render it if the report is reimbursed


app/javascript/controllers/expense_form_controller.js
```diff
@@ -31,14 +31,16 @@ export default class extends Controller {
...
 
-    this.buttonTarget.addEventListener('click', e => {
-      e.preventDefault()
-      if (this.enabledValue) {
-        this.formTarget.requestSubmit()
-      } else {
-        this.edit(e)
-      }
-    })
+    if (this.hasButtonTarget) {
```
sampoder:
```suggestion
    // we don't render the button if the report is reimbursed
    if (this.hasButtonTarget) {
```

garyhtou:
Discarded changes that came from https://github.com/hackclub/hcb/pull/8598

PR #8606, opened 2024-11-20
Default `missing_receipts` to `nil` in ` PendingTransactionEngine::PendingTransaction::All`

sampoder:
We were only returning transactions with missing receipts


sampoder:
lies

PR #8607, opened 2024-11-20
Only run `.to_i` on `tag_id` when `tag_id` isn't nil

sampoder:
Bring back pending transactions!


app/services/pending_transaction_engine/pending_transaction/all.rb
```diff
@@ -41,6 +41,7 @@ def canonical_pending_transactions
                                               .order("canonical_pending_transactions.date desc, canonical_pending_transactions.id desc")
 
             if @tag_id
+              puts @tag_id
```
garyhtou:
```suggestion
```

garyhtou:
Good catch, thank you Sam!

PR #8625, opened 2024-11-21
[Org Team] Fix list view in transparency mode

garyhtou:
## Summary of the problem

`user.admin?` failed when in transparency mode
https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/1188/samples/timestamp/2024-11-21T05:10:42Z

## Describe your changes

guard clause to return false when user isn't signed in


sampoder:
Good catch - thanks!

PR #8626, opened 2024-11-21
Disable Floating UI's autoUpdate on elementResize

sampoder:
Closes https://github.com/hackclub/hcb/issues/8588

Let's see how this feels for a couple of days. Maybe it's the lesser of two evils? Not sure tbh.


app/javascript/controllers/menu_controller.js
```diff
@@ -58,7 +58,10 @@ export default class extends Controller {
     this.cleanup = autoUpdate(
       this.toggleTarget,
       this.content,
-      this.computePosition.bind(this, false)
+      this.computePosition.bind(this, false),
+      {
+        elementResize: false
```
sampoder:
```suggestion
        elementResize: false // See https://github.com/hackclub/hcb/issues/8588
```

sampoder:
Tried it out in dev; feeling really good!

PR #8630, opened 2024-11-21
decline based on grant rules before insufficient funds

24c02:
before, if you charged a grant card at the wrong merchant/category for more than its subledger had available, it would send the decline email for insufficient balance, which gave the vibe that the charge might've gone through if you had the money – it wouldn't :-P


sampoder (approved):

garyhtou:
Amazing, checking grant rules is also much quicker than the card's balance. so this is a double win! 🎊

PR #8631, opened 2024-11-21
Actually add papertrail to `User`

sampoder:


garyhtou:
Oh! We had it twice- thanks

PR #8643, opened 2024-11-22
[HCB Wrapped] Add reimbursement and card grant metrics

YodaLightsabr:
## Summary of the problem
HCB Wrapped needs additional metrics for 2024

## Describe your changes
Adds the following metrics (with maybe more to come):
- Reimbursement count
- Reimbursement amount
- Card grant count
- Card grant amount


app/models/metric/user/card_grant_amount.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+#  index_metrics_on_subject  (subject_type,subject_id)
+#
+class Metric
+  module User
+    class CardGrantAmount < Metric
+      include Subject
+
+      def calculate
+        user.card_grants.sum(&:amount_cents)
```
garyhtou:
```suggestion
        user.card_grants.sum(:amount_cents)
```

when sum takes a symbol, it sum the column using SQL. when it takes a block, it sum in Ruby by calling the block (method).

In our case here, we can just use SQL to sum

app/models/metric/user/reimbursement_amount.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+#  index_metrics_on_subject  (subject_type,subject_id)
+#
+class Metric
+  module User
+    class ReimbursementAmount < Metric
+      include Subject
+
+      def calculate
+        user.reimbursement_reports.sum(&:amount_cents)
```
garyhtou: This can stay as a block (`&:amount_cents`) since we don't have a column for it

garyhtou (approved):
🪄

app/models/metric/user/card_grant_amount.rb
YodaLightsabr: Oh that's right. Thanks for catching this.

PR #8648, opened 2024-11-23
[HQ Subordinate Summary] Run cron at 7am eastern

garyhtou:
switched it from sunday night to monday morning


sampoder:
Ah we should really think of a better name for this... subordinate has a certain implicit meaning

PR #8681, opened 2024-11-26
Fix empty scheduled disbursement tooltip for admins

transcental:
## Summary of the problem
Fixes #8217, admin tooltip on scheduled disbursements is incorrectly placed 

## Describe your changes
Moves the tooltip inside a conditional to only show on scheduled disbursements.

image
image


sampoder (approved):
LGTM! Thank you

PR #8686, opened 2024-11-26
Card title patch

Luke-Oldenburg:
## Summary of the problem
The cards aren't technically debit cards but we also don't want to mislead users into thinking that HCB offers lines of credit.

## Describe your changes
I renamed debit card to HCB card in two different places.


sampoder (approved):
Thank you!

PR #8695, opened 2024-11-26
Add error logging to Webauthn JS controller

sampoder:
Impossible to debug production-only bugs without this.


app/javascript/controllers/webauthn_auth_controller.js
```diff
@@ -86,6 +86,9 @@ export default class extends Controller {
         this.continueButtonTarget.value = 'Continue'
 
         this.errorTarget.classList.remove('display-none')
+
+        console.error(e)
+        airbrake?.notify(err)
```
sampoder:
```suggestion
        airbrake?.notify(e)
```

PR #8706, opened 2024-11-27
Fix whitespace causing emails to be unreadable

Luke-Oldenburg:
## Summary of the problem
If an email has leading or trailing whitespace, it causes it to be unreadable and HCB presents the error message "Please enter an email address."

## Describe your changes
I used `.strip` to strip the whitespace from the email address and process it as normal. Didn't test to see if this worked as I'm still struggling to login on the dev mode but it was the only file I saw those error messages in.


Luke-Oldenburg:
Closing because I was able to test now and it didn't work. Will reopen when fixed.

garyhtou:
I tried adding a leading/trailing whitespace and could not reproduce.

I believe we already handle it in the model: https://github.com/hackclub/bank/blob/cff8294dcba85dd1f2df16432a75adfb899d50d6/app/models/user.rb#L145

Feel free to continue testing and working on a fix! It would be amazing if you could also share a step-by-step guide on getting the error.

Luke-Oldenburg:
It happens in these two locations that I know of. There may be more.
Screenshot from 2024-11-27 11-45-34
Screenshot from 2024-11-27 11-44-32

PR #8717, opened 2024-11-27
[v4 API] [card grant] add balance & spending to card grants

24c02:
- adds `/api/v4/card_grants/:id/spending` to get card grant transactions
  - `?only_missing_receipts=true` should do what it says on the tin
- adds `balance_cents` to card grant v4 show
- `?include_disbursements=true` on card grant v4 show should also do what it says on the tin

i think this should be enough to programmatically stop high seas-ers from buying more card grants if they don't receipt their stuff in a timely manner >:3


app/views/api/v4/card_grants/show.json.jbuilder
```diff
@@ -1 +1 @@
-json.partial! @card_grant, expand: [:user, :organization]
+json.partial! @card_grant, expand: [:user, :organization, params.key?(:include_disbursements) && :disbursements]
```
garyhtou: I'd prefer if we could define expanded objects via and `expand` param.

For example, `expand=disbursements,user` would expand both disbursements and the user. In a way, it's similar to how the v3 API worked: https://hcb.hackclub.com/docs/api/v3/operations/get-a-single-organization

Are there other common ways that people define expandable keys?

app/controllers/api/v4/card_grants_controller.rb
```diff
@@ -40,6 +40,15 @@ def topup
         end
       end
 
+      def spending
+        @card_grant = CardGrant.find_by_public_id!(params[:id])
+        authorize @card_grant
+
+        @card = @card_grant&.stripe_card
+        @hcb_codes = @card&.hcb_codes
+        @hcb_codes = @hcb_codes.select(&:missing_receipt?) if params.key?(:only_missing_receipts)
```
garyhtou: Should this be paginated? It'd be easier to add now so that we don't incur a breaking change in the future

```suggestion
        @hcb_codes = @hcb_codes.select(&:missing_receipt?) if params[:missing_receipts] == true
```

In the future, this would allow `missing_receipts` to also take in `false`.


It may also be better to name this as `has_receipts` to `receipts_attached`. I generally prefer the positive phasing over the negative one ("missing").

garyhtou (commented):
I'll also mention that I'm cool with these changes as-is if high seas needs something in asap — we can make changes in the future, but it would be breaking changes that high seas would then need to adapt to.

app/controllers/api/v4/card_grants_controller.rb
24c02: is there a use case for pulling card grant transactions that _are_ receipted?

app/views/api/v4/card_grants/show.json.jbuilder
24c02: that sounds good, but the rest of v4 would probably have to be adapted to match though

app/controllers/api/v4/card_grants_controller.rb
24c02: could be, not sure when a card grant would ever have a pagination-worthy amount of transactions in its lifecycle
also, lol:
CleanShot 2024-11-27 at 16 28 45@2x

app/views/api/v4/card_grants/show.json.jbuilder
24c02: one issue i've seen with that in the past in larger enterprise stuff is ya gotta be careful what objects will accept which expansions in what context: you could thrash and effectively DoS the DB on a certain SIS by expanding fields that are fine when loading one object but really expensive when loading them in the index of all objects – not sure how to reason about what makes sense where, i just want my darn card grants ;w;

garyhtou: Expanding `disbursements` should be possible in production now due to https://github.com/hackclub/hcb/pull/8825. We no longer need to add this line!

To use, add `?expand=disbursements` to your URL.

app/controllers/api/v4/card_grants_controller.rb
garyhtou: yeah, let's scratch pagination for now

garyhtou (commented):
I realized that our Stripe Card controller [allows for listing transactions](https://github.com/hackclub/bank/blob/3776e2486e066a777af42a10a476aa2f2567e1fc/app/controllers/api/v4/stripe_cards_controller.rb#L20-L28). So, I'd suggest using that existing endpoint. What we have here does work, however, it creates a new endpoint that "proxies" the same information.

Since at the moment, there's no way to:
1. Get a Card Grant's Stripe Card ID
2. Filter Card's transactions for missing receipts

I've created https://github.com/hackclub/hcb/pull/8826 to fill in that remaining functionality. I believe with https://github.com/hackclub/hcb/pull/8826, you should hopefully have everything you need!

1. GET `/card_grants/:id`
2. using the response's `card_id`, GET `/cards/:id/transactions?missing_receipts=true`

Let me know what you think!

24c02:
ty gary, i think that should tick most of my boxes in a smarter way! only thing i'm missing after that is [getting the balance](https://github.com/hackclub/hcb/pull/8717/commits/1f83d7833355376562399f72dec5a07b5a787f94), but that's not super important for HS ^_^

garyhtou:
I totally forgot about that — i'll get that added!

garyhtou:
Here we go! https://github.com/hackclub/hcb/pull/8827

24c02:
awesome, tysm!

PR #8705, opened 2024-11-26
Fixed GIF won't load in Gmail #8682

MntRushmore:
## Summary of the problem
This hopefully will make the GIF show up in the email..

## Describe your changes
Updated GIF to Base64 format. (Didn't Work)
Uploaded to CDN (Didn't work)
Trying with Giphy now

PR #8718, opened 2024-11-27
Fixed gif rendering issue

46-HA:
## Summary of the problem

The issue was probably caused because the gif wouldn't load so I created a new link using #cdn and a fallback just in case

## Describe your changes

I added a new link that would work better


garyhtou:
Thanks for this, however, the GIF as-is started to work all of a sudden.

#8682

garyhtou:
It looks like https://github.com/hackclub/hcb/pull/8705 solved it. Thanks!

PR #8729, opened 2024-11-29
Fix receipt rendering in popover

sampoder:
Fixes https://github.com/hackclub/hcb/issues/8687, this is related to when we switched to turbo frames


app/assets/stylesheets/components/_receipts.scss
```diff
@@ -136,7 +136,8 @@
   display: none;
 }
 
-.modal--popover--receipt-expanded
+.modal--popover--receipt-expanded
```
sampoder:
```suggestion
.modal--popover--receipt-expanded
```

PR #8731, opened 2024-11-29
Don't attempt to render the balance graph when rendering mock total

sampoder:
Essentially this disables the balance graph which I think is causing this


sampoder:
actually doesn't fix anything but removes a JS error

PR #8732, opened 2024-11-29
Make `amount` on a mock transaction a `Money` object

sampoder:
`render_money` assumes any number inputted as cents. so when we had a $10.37 mock transaction it was being rendered as 10 cents.

closes https://github.com/hackclub/hcb/issues/8722


sampoder:
(needs to be tested in staging)

PR #8733, opened 2024-11-29
Create a `CardGrantSetting` on loading the new card grant page

sampoder:
Closes https://github.com/hackclub/hcb/issues/8680


sampoder:
It would be nice to be able to do this in model callbacks, but we don't have that with Flipper.

PR #8727, opened 2024-11-29
Order users in filter menu alphabetically

sampoder:
Closes https://github.com/hackclub/hcb/issues/8689

PR #8738, opened 2024-11-29
Options ordered alphabetically in filter menu [transactions view]

Aryankpoor:
[#8727](https://github.com/hackclub/hcb/pull/8727) might not have fully resolved #8689.
Also noticed some glitches after the push

image

## Here the order is not correct
image

This PR should be able to order user options alphabetically.


Aryankpoor:
Note: This PR only fixes the sorting not the glitches seen in the attached image

Aryankpoor:
Seems to be working fine on deployment
image

sampoder (changes requested):
Thanks! For the organisers one, I'd prefer for us to do it in SQL. Here's what I got to work:

```sql
@event.organizer_positions.joins(:user).order(Arel.sql("CONCAT(preferred_name, full_name) ASC"))
```

We should do the same for tags, something like `@event.tags.order(label: :asc)`.

Aryankpoor:
Done! Let me know if everything looks good. 
Also wanted to mention that this PR also sorts ```Type``` alphabetically

## Original View
image

## Updated view
image

app/views/events/_filter_menu.html.erb
```diff
@@ -3,7 +3,7 @@
     type="button"
     aria-label="Add filter..."
     class="tooltipped tooltipped--s pop menu__toggle menu__toggle--arrowless overflow-visible z-10"
-    data-menu-target="toggle" data-action="menu#toggle click@document->menu#close keydown@document->menu#keydown">
+    data-menu-target="toggle" data-action="menu#toggle click@document->menu#close keydown@document->keydown">
```
sampoder: Any reason for this change?

app/views/events/_filter_menu.html.erb
```diff
@@ -37,10 +36,9 @@
         </div>
       <% end %>
       <div data-tabs-target="tab" id="user">
-        <% @organizers.each do |position| %>
+        <% @event.organizer_positions.joins(:user).order(Arel.sql("CONCAT(users.preferred_name, users.full_name) ASC")).each do |position| %>
```
sampoder: Can this be moved into `EventsController`? There should already be an `@organizers =` line.

sampoder (commented):
Two quick comments; thanks!

app/views/events/_filter_menu.html.erb
Aryankpoor: Aah sorry, that should not have been removed. That was just me testing something out as an attempt to resolve the glitches I mentioned earlier. 
I have added that back in

app/views/events/_filter_menu.html.erb
Aryankpoor: Done :+1:

Aryankpoor: I have kept it in ```filter_menu``` for now as moving it to EventsController seems to be returning an error in deployment

Aryankpoor: I tested out the deployment after making this change and the transaction page broke. I can't figure out why though

Aryankpoor: I was trying out something like ```@organizers = @event.organizer_positions.joins(:user).order(Arel.sql("CONCAT(users.preferred_name, users.full_name) ASC"))```

sampoder: [`b5c4da0` (#8738)](https://github.com/hackclub/hcb/pull/8738/commits/b5c4da02731a8dee2795d4229dbc79016a726dd2) seems to be working for me locally. I think it's good to merge, might be staging timing out.

sampoder (approved):
LGTM!

PR #8742, opened 2024-11-30
Allow for burning inactive cards

Luke-Oldenburg:
## Summary of the problem
I accidentally ordered a card to the wrong address so I ordered a new one to my house. I was contacted about it but at that time I did not need the card anymore. I tried to burn it to remove it from my list of cards as it would never be used but I could not.

## Describe your changes
I moved the cancel stripe card action outside of the check to see if it has been initially activated.


sampoder (approved):
LGTM! Thank you for catching this.

PR #8749, opened 2024-11-30
Create `ApplicationHelper#error_boundary`

sampoder:
Closes https://github.com/hackclub/hcb/issues/8700

View PR for example of using `fallback_text`, using `fallback` would look something like:

```erb
<% fallback = capture do &>
  <p>No recent activities available.</p>
<% end %>
<%= error_boundary fallback: do %>
  render partial here
<% end %>
```


garyhtou:
I'm glad my suggestion worked!

PR #8756, opened 2024-11-30
Rename account transfer filter to HCB transfer

Luke-Oldenburg:
## Summary of the problem
The filter was called `Account transfer` despite being referred to as a `HCB transfer` on the Transfers page.

## Describe your changes
I renamed the filter from `Account transfer` to `HCB transfer`.


sampoder (changes requested):
Thanks for the PR! This should be changed on Line 184 of event's controller, that feels cleaner.

Luke-Oldenburg:
I think I had looked at that and wasn't sure how many other places referenced it as "Account" behind the scenes. I didn't want to go down a rabbit hole of having to change things but if that works I'll correct it there.

sampoder (approved):
Made a couple of edits, this is good to merge!

Luke-Oldenburg:
Ah yep, those were the things I was worried about missing if I renamed it from the root.

PR #8760, opened 2024-12-01
Fix canceled unactivated Stripe Card status

sampoder:
Closes https://github.com/hackclub/hcb/issues/8753, thanks reporting Luke!


Luke-Oldenburg:
Do we want to leave the "Card for X is on the way!" message even after it has been cancelled?

Luke-Oldenburg:
re: my last comment, if HQ has shipping info then maybe there could be a "Card for X has arrived!" message instead.

sampoder:
Re both comments: if the card is cancelled or activated, we should probably hide that message

sampoder:
We don't have tracking

Luke-Oldenburg:
Will do :+1:

PR #8788, opened 2024-12-03
`validates_length_of :payment_for` on `Wire`

sampoder:


app/models/wire.rb
```diff
@@ -75,6 +75,8 @@ class Wire < ApplicationRecord
   end
 
   validates_length_of :remittance_info, maximum: 140
+
```
```suggestion

```

PR #8798, opened 2024-12-03
Handle failed wires

sampoder:
Still got a bit of work to do imo (need to better document that some money will be lost due to exchange fees), but this is a good to go into production.


app/views/wire_mailer/notify_failed.html.erb
```diff
@@ -0,0 +1,5 @@
+We're writing to let you know your <%= link_to "recent wire to #{@wire.recipient_name}", hcb_code_url(@wire.local_hcb_code) %> was rejected and failed to send. Here's the explanation given by the receiving bank / country:
+
+<pre><%= @reason %></pre>
+
+The funds will be returned to your account shortly (due to exchange rate fluctuations, some money may have been lost). After verifying that the account details for the recipient are correct, you can re-attempt this transfer <%= link_to "here", new_event_wire_url(@wire.event) %>.
```
garyhtou: Should the exchange rate portion only be included if the currency is not USD?

PR #8808, opened 2024-12-04
Install `rack-timeout`

sampoder:
And configures it to a 30 second timeout.

Closes https://github.com/hackclub/hcb/issues/8793


Gemfile
```diff
@@ -222,3 +222,5 @@ gem "email_reply_parser"
 gem "eu_central_bank"
 
 gem "whitesimilarity"
+
+gem "rack-timeout", require: "rack/timeout/base"
```
sampoder:
```suggestion
gem "rack-timeout", require: "rack/timeout/base"

```

PR #8825, opened 2024-12-04
[API v4] Add expandable keys via params

garyhtou:
## Summary of the problem

The v4 API did not support expanding keys via params. They were all hard coded within the view when rendering a partial.

## Describe your changes

This PR allow for expanding specific keys using an `expand` param. For example: `?expand=organization,user`. It also provides two handy methods:
- `expand(*keys, &block)`
   Adds keys to expand for object rendered inside the block.

   ```ruby
   # previously
   json.partial! @card_grant, expand: [:organization] # it didn't consider query params

   # now
   expand :organization do
     json.partial! @card_grant # if param `expand=user`, it'd also expand user in addition to organization
   end
   ```
   Manually expanded keys (using the `expand` method) are always only scoped within their block. Keys expanded via query parameters will always be expanded — this latter behavior could change in the future.

- `expand?(key)`
   checks whether a key should be expanded.

   ```ruby
   # previously
   if local_assigns[:expand].include?(:user)
     json.user card_grant.user, partial: "api/v4/users/user", as: :user
   end

   # now
   if expand?(:user)
     json.user card_grant.user, partial: "api/v4/users/user", as: :user
   end
   ```


garyhtou:
This will allow us to expand card grant disbursements in https://github.com/hackclub/hcb/pull/8717

24c02:
sweet!

PR #8831, opened 2024-12-05
Add multiple filters in transactions page

### LINKED ISSUE #8587: [Filter] Being able to add multiple filters of the same type
With the awesome new filter UI and functionality, users can now filter by different types. I think a great addition would be to allow users to sort by multiple subtypes. For example, filter by multiple user. My intuition for this is it would have or logic. 

Example:
I want to sort by all transactions that are Dylan's OR Gary's. I goto the filter UI, click Gary and then it adds it as a filter, including both user's transactions. 
Image

I think this behavior would be true for Tag, User, and Type. I think we can keep the existing functionality for Date, Amount, and Receipts.

Aryankpoor:
Resolves #8587 (hopefully)

Tried something out with checkboxes although that's not the aim. If it works then will be trying out a different approach than checkboxes


Aryankpoor:
Something is breaking in deployment.

garyhtou:
Superseded by https://github.com/hackclub/hcb/pull/9862 due to git rewrite

PR #8839, opened 2024-12-05
Re-add HCB Wrapped 2024 under Flipper flag

YodaLightsabr:
## Summary of the problem
HCB Wrapped no longer exists on HCB

## Describe your changes
Re-adds the following:
- HCB Wrapped banner (under Flipper flag)
- HCB Wrapped React component & page (under Flipper flag)
- HCB Wrapped data download page


app/policies/user_policy.rb
```diff
@@ -81,4 +81,8 @@ def disable_feature?
     user.admin? || record == user
   end
 
+  def wrapped?
+    user.admin? || record == user
+  end
+
```
garyhtou: is this used anywhere?

app/controllers/users/wrapped_controller.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+module Users
+  class WrappedController < ApplicationController
+    skip_after_action :verify_authorized
+    before_action :set_data
+
+    def index
+      unless Flipper.enabled?(:bank_wrapped, current_user)
```
garyhtou: Before merging, double check that the flipper flag is disabled

app/models/metric/hcb/new_events.rb
```diff
@@ -26,7 +26,7 @@ def calculate
                .not_hidden
                .not_demo_mode
                .approved
-               .where("EXTRACT(YEAR FROM created_at) = ?", 2023)
+               .where("EXTRACT(YEAR FROM events.created_at) = ?", 2023)
```
garyhtou: should be switched to 2024 at some point lol

garyhtou (commented):
🪄

garyhtou (commented):
🪄

app/controllers/users/wrapped_controller.rb
YodaLightsabr: Yep

image

app/policies/user_policy.rb
YodaLightsabr: Since we switched to the wrapped_controller I don't think so

YodaLightsabr:
```suggestion
```

PR #8850, opened 2024-12-06
Implement keyword locks on card grants

sampoder:
Closes https://github.com/hackclub/hcb/issues/7242


app/controllers/card_grants_controller.rb
```diff
@@ -24,14 +24,15 @@ def new
       @card_grant.amount_cents = last_card_grant.amount_cents
       @card_grant.merchant_lock = last_card_grant.merchant_lock
       @card_grant.category_lock = last_card_grant.category_lock
+      @card_grant.keyword_lock = last_card_grant.keyword_lock
```
```suggestion
```

app/models/card_grant.rb
```diff
@@ -187,6 +188,10 @@ def allowed_category_names
     allowed_categories.map { |category| YellowPages::Category.lookup(key: category).name || "#{category}*" }.uniq
   end
 
+  def keyword_lock
+    super || setting&.keyword_lock
+  end
```
sampoder:
```suggestion
  end
```

PR #8855, opened 2024-12-06
[v4 API] Fix PublicActivity attribution

24c02:
before, v4 api controllers didn't tell public_activity about themselves, resulting in `owner` being nil and thus any action taken via the api showing as performed by 'a user'.

CleanShot 2024-12-06 at 10 18 19@2x

this PR fixes that by including the StoreController concern and exposing current_user :-)

CleanShot 2024-12-06 at 10 22 32@2x


sampoder (approved):
LGTM - thank you!

garyhtou:
amazing! great work hunting this down @24c02 !

PR #8861, opened 2024-12-07
Hide shipping info on canceled cards

Luke-Oldenburg:
## Summary of the problem
Canceled cards had shipping info showing.

## Describe your changes
I disabled shipping info on canceled cards.

Note: didn't test because my codespace is still messed up


app/views/stripe_cards/shipping.html.erb
```diff
@@ -2,7 +2,9 @@
   <% if @stripe_cards.any? %>
     <div class="grid grid--split mb3">
       <% @stripe_cards.each do |card| %>
-        <%= render partial: "stripe_cards/shipping", locals: { stripe_card: card } %>
+        <% if card.status_text != "canceled" %>
```
garyhtou: Instead of using an `if` to skip canceled cards inside the `for` loop, we can use SQL to filter them out.

try this: We're going to update how `@stripe_cards` is defined.
https://github.com/hackclub/bank/blob/95cb186ca5ddb1e0fcb021c8b95c5e2964bb725d/app/controllers/stripe_cards_controller.rb#L12-L21
^ In there, we'll add `where.not(stripe_status: "canceled")` after `physical_shipping`.

app/controllers/stripe_cards_controller.rb
```diff
@@ -12,6 +12,7 @@ def index
   def shipping
     # Only show shipping for phyiscal cards if the eta is in the future (or 1 week after)
     @stripe_cards = current_user.stripe_cards.physical_shipping.reject do |sc|
+      where.not(stripe_status: "canceled")
```
sampoder:
```suggestion
    @stripe_cards = current_user.stripe_cards.physical_shipping.where.not(stripe_status: "canceled").reject do |sc|
```

Luke-Oldenburg:
Whoops lol. Should be good now I think.

app/controllers/stripe_cards_controller.rb
```diff
@@ -11,7 +11,7 @@ def index
 
   def shipping
     # Only show shipping for phyiscal cards if the eta is in the future (or 1 week after)
-    @stripe_cards = current_user.stripe_cards.physical_shipping.reject do |sc|
+    @stripe_cards = current_user.stripe_cards.physical_shipping.where.not(stripe_status: "canceled").reject do |sc|
```
sampoder:
```suggestion
    @stripe_cards = current_user.stripe_cards.where.not(stripe_status: "canceled").physical_shipping.reject do |sc|
```

sorry for some reason the `physical_shipping` scope returns an array? that's mb

sampoder (approved):
Tested; looks good to me!

PR #8877, opened 2024-12-10
Fix unattached wire receipts / HCB code memoisation pre-save

sampoder:
lol. umm this was tricky to figure out.


app/models/wire.rb
```diff
@@ -329,7 +329,7 @@ def local_hcb_code
   end
 
   def usd_amount_cents
-    return -1 * local_hcb_code.amount_cents unless local_hcb_code.no_transactions?
+    return -1 * local_hcb_code.amount_cents unless id.nil? || local_hcb_code.no_transactions? # don't access local_hcb_code before saving.
```
garyhtou: How about we make `local_hcb_code` return `nil` when `!persisted?`

garyhtou:
Context: https://hackclub.slack.com/archives/C068U0JMV19/p1733863063616519?thread_ts=1733862750.942139&cid=C068U0JMV19

image

PR #8884, opened 2024-12-11
Hide snow toggle on small computers

sampoder:
Things are just too cluttered right now. Closes https://github.com/hackclub/hcb/issues/8867


app/assets/stylesheets/application.scss
```diff
@@ -501,3 +501,9 @@ summary {
     bottom: 64px !important;
   }
 }
+
+@media (min-width: 56em) and (max-width: 84rem) {
```
sampoder:
```suggestion
@media (min-width: 56em) and (max-width: 88rem) {
```

PR #8894, opened 2024-12-11
[v4 API] [Card Grants] allow changing merchant/category/keyword locks!

24c02:
a common High Seas problem has been "this merchant uses a different MID in my country" 

easy to fix for grants going forward, but requires manually canceling and regranting cards :-(

this PR fixes that plus a bunch of other random stuff i ran into along the way:
- adds a v4 update route for card grants & an accompanying policy 
  - (just merchant locks, etc.)
- restricts creating/editing card grants to users with the manager role
- adds status to card grant v4 partial
- raises error on trying to cancel a canceled/expired grant
- adds a v4 cancel route


app/policies/card_grant_policy.rb
garyhtou, line 1: Great! I confirmed in prod that this behavior change doesn't interrupt any YSWS team's work. Only managers and admins have been working with card grants — no members.
```sql
SELECT * FROM "card_grants" cd
join "users" u on cd.sent_by_id = u.id
join "organizer_positions" op on u.id = op.user_id and cd.event_id = op.event_id

where op.role != 100 and u.access_level = 0 # not manager and is normal user (non-admin)
-- it showed no results
```

app/models/card_grant.rb
```diff
@@ -135,6 +135,8 @@ def expire!
   end
 
   def cancel!(canceled_by = User.find_by!(email: "bank@hackclub.com"), expired: false)
+    raise ArgumentError, "Grant is already #{status}" unless status == :active
```
garyhtou: good catch!

app/controllers/api/v4/card_grants_controller.rb
```diff
@@ -40,6 +40,34 @@ def topup
...
+      def update
+        @card_grant = CardGrant.find_by_public_id!(params[:id])
+
+        authorize @card_grant
+
+        begin
+          @card_grant.update!(params.permit(:merchant_lock, :category_lock, :keyword_lock))
+        rescue ArgumentError => e
+          return render json: { error: "invalid_operation", messages: [e.message] }, status: :bad_request
+        end
```
```suggestion
        @card_grant.update!(params.permit(:merchant_lock, :category_lock, :keyword_lock))
```

`update!` would likely raise `ActiveRecord::RecordInvalid` error which V4's [ApplicationController](https://github.com/hackclub/bank/blob/29241509505f4dbd4af3b678813717afee6d3418/app/controllers/api/v4/application_controller.rb#L25-L27) can handle. So, no need for the rescue

garyhtou (commented):
LGTM

garyhtou:
If [`e438d23` (#8894)](https://github.com/hackclub/hcb/pull/8894/commits/e438d2312583c7217c530799ac499deb48b6a62d) ends up causing a different behavior than you expect, feel free to open a PR to revert it. I didn't test it, but I'm 99% sure it works.

I'm going to merge this since I imagine you'd prefer this in prod sooner rather than later 😊

garyhtou (approved):

24c02:
tysm!! no real preference either way on the rescue, if it's error it's error :-P

PR #8899, opened 2024-12-12
Hide perks and documents in transparency mode

manuthecoder:
This pull request includes changes to the `app/views/events/_nav.html.erb` file to enhance security by ensuring that only signed-in organizers can access certain event features. See #8897 

Improvements:

* Modified the condition for the `promotions` dock item to check if the organizer is signed in.
* Modified the condition for the `documentation` dock item to check if the organizer is signed in.
* Updated the "Promotions & perks" link to ensure it is only visible to signed-in organizers.
* Updated the "Documents" link to ensure it is only visible to signed-in organizers.


sampoder (changes requested):
Can we do this by updating the Pundit policy instead?

manu-gurudath:
Done! On desktop, I kept the condition so the policy button still renders and shows the tooltip when the user is signed in, if promotions are disabled for the organization.

sampoder (approved):
LGTM!

PR #8902, opened 2024-12-12
Validate the format of HCB codes

sampoder:
Closes https://github.com/hackclub/hcb/issues/8879

There's some work to be done on migrating the broken production data, see the thread at https://hackclub.slack.com/archives/C047Y01MHJQ/p1733975901963519 for more context


app/models/hcb_code.rb
```diff
@@ -61,6 +61,12 @@ class HcbCode < ApplicationRecord
     receipts "Has receipt?" do |receipts| receipts.exists? end
   end
 
+  validate on: :create do
+    unless hcb_code.match?(/\AHCB-\d{3}-\S+\z/)
+      errors.add(:hcb_code, "does not meet the required format.")
+    end
+  end
```
garyhtou: i'd suggest

```ruby
validates :hcb_code, format: { with: /regex/ }
```

and doing it on all contexts (not just create)

PR #8904, opened 2024-12-12
Change text selection color

manuthecoder:
This small tweak changes the default text selection color to reflect HCB's brand colors

| Before | After |
|--------|--------|
|  image| image|


garyhtou:
I think the red background here will be too bright/contrasty. During our call, @manuthecoder suggested inverting the contrast and doing dark text on a light background. We'll try a shade of the default browser select color first. If that doesn't work we'll try one of the hack club theme colors

sampoder:
#338eda might be nice?

sampoder:
Screenshot 2025-01-03 at 12 23 57 AM

I don't really think this fits - to be honest, I prefer the default my computer provides as it's more subtle. Maybe we do what GitHub does and use our own shade of blue? It's subtle but it works.

Screenshot 2025-01-03 at 12 26 41 AM

manu-gurudath:
I'll close this for now since it's been a while and I think everyone prefers the default color

PR #8926, opened 2024-12-14
Add `teenager` boolean to `User`

sampoder:


northeastprince:
Shouldn't teenager just be a method if we store birthdays?

Luke-Oldenburg:
> Shouldn't teenager just be a method if we store birthdays?

I also had that thought looking at this.

sampoder:
The column is encrypted so you can’t do SQL filtering based on value :/

PR #8936, opened 2024-12-15
Separate to and for in transfers

Luke-Oldenburg:
## Summary of the problem
The to column had both the recipient and the memo for certain transaction types.

## Describe your changes
I separated these into multiple columns.


garyhtou (commented):
It looks like the goal of this PR is to add another table column named `For`. Glancing at the code, it looks like only Disbursements were given a new cell (`<td>`). There's a chance that other transfer types (e.g. Check, Ach Transfer, etc.) would have some layout shift since it's missing a cell.

Would you mind sharing a screenshot of the new look?

Luke-Oldenburg:
Screenshot From 2025-01-12 15-58-50

garyhtou:
Superseded by https://github.com/hackclub/hcb/pull/9859 due to git rewrite.

PR #8937, opened 2024-12-15
Remove snow toggle

Luke-Oldenburg:
## Summary of the problem
https://github.com/hackclub/hcb/issues/8930#issuecomment-2543371399

## Describe your changes
Removed the snow toggle


garyhtou:
Now that it's January, the winter theme has been disabled across all of HCB. However, I'm still in favor of removing the toggle at this time. Our UI is about to have a large change which means that I'll give us a chance to rework this toggle for next year.

PR #8938, opened 2024-12-15
Hide expired sessions on the security page

sampoder:
Closes https://github.com/hackclub/hcb/issues/8934


Luke-Oldenburg:
@sampoder are you sure that the issue was that it was showing expired sessions? There was another section for those but I had to press the button to expire them manually in order for them to switch before.

PR #8941, opened 2024-12-16
Bump Ruby version to 3.3.6

emergenitro:
## Summary of the problem
Update to Ruby version 3.3.6. This is a routine update that includes minor bug fixes. It also stops warning missing default gem dependencies that will be bundled gems in Ruby 3.5. For more details, please refer to [the release notes on GitHub](https://github.com/ruby/ruby/releases/tag/v3_3_6).

Just updated .ruby-version and Dockerfile.


Gemfile.lock
```diff
@@ -961,7 +961,7 @@ DEPENDENCIES
   yellow_pages!
 
 RUBY VERSION
-   ruby 3.3.5p100
+   ruby "3.3.6"
 
 BUNDLED WITH
    2.5.17
```
garyhtou:
```suggestion
   ruby 3.3.6p108

BUNDLED WITH
   2.5.22
```

Gemfile.lock
```diff
@@ -961,7 +961,7 @@ DEPENDENCIES
   yellow_pages!
 
 RUBY VERSION
-   ruby 3.3.5p100
+   ruby 3.3.6p108
```
garyhtou: Running `bundle update --ruby` will usually update this on your behalf!

emergenitro: Yep thank you so much! That's my bad

PR #8949, opened 2024-12-16
[User Session] Clean up expiration logic

garyhtou:
It retains the same behavior but cleans up the code a bit using scopes.


sampoder (approved):
LGTM!

PR #8953, opened 2024-12-17
[Yellow Pages] Add service for listing merchants missing YP

garyhtou:
## Summary of the problem

while working on https://github.com/hackclub/hcb/pull/8952, @YodaLightsabr and I noticed that some top merchants don't have yellow page entries.

## Describe your changes

so, this allows us to easily list top merchants (by amount spent) that are missing yellow page entries.


garyhtou:
the sql query was copied from metric's SpendingByMerchant, but it needs a bit more work

PR #8957, opened 2024-12-17
[Admin] Clean up Check Deposit pages

garyhtou:
1. Prevent accessing check deposit "manual processing" page when the check deposit _**does not**_ require manual processing
2. Hide "process" button if check deposit does not require manual processing. Instead, show "view" button.


sampoder (approved):

sampoder:
LGTM - much needed!

PR #8968, opened 2024-12-17
Update snow z-index

emergenitro:
## Summary of the problem
Just updated the snow to have a higher z-index as it is currently going behind the search bar in orgs. Here's an image before:
image

and after:
image

Thanks and let me know if there are any issues :)


garyhtou (commented):
I appreciate the PR! I do agree it's a bit odd how we had the snow under the cards/inputs. However, I think there's a flip side to it. Having the snow under cards/inputs makes the page a bit more useable when the snow is on.

Do you have any thoughts on this?

emergenitro:
Oh I see what you mean but honestly as a user, the snow is bugging me more than the usability when it's above. It doesn't really interfere with how usable the inputs are as far as I've tested it.

Nevertheless, I do understand and I'm not totally sure how I'd go about it. Do you have any suggestions? Otherwise, we can leave it as-is

Luke-Oldenburg:
For me, the snow is too distracting in general and I have it off. But if you have it on, I don't see it going over text inputs that big of a deal.

garyhtou:
Superseded by https://github.com/hackclub/hcb/pull/9865 due to git rewrite.

PR #8979, opened 2024-12-17
Add `bestFriend` to Wrapped's data

sampoder:


app/controllers/users/wrapped_controller.rb
```diff
@@ -34,6 +35,7 @@ def set_data
           name: @user.full_name,
           firstName: @user.first_name,
           id: @user.id,
+          profilePicture: profile_picture_for(@user, 256),
```
garyhtou: Should we also update https://github.com/hackclub/bank/blob/d45782786bce34570480009cd898627cc4bbf79d/app/models/metric/user/most_interacted_with.rb#L61 to be size 256?

PR #8985, opened 2024-12-18
Deprecate `HcbCode#url` & switch to `popover_path`

sampoder:
Progress towards https://github.com/hackclub/hcb/issues/8942


app/models/hcb_code.rb
```diff
@@ -57,17 +57,18 @@ class HcbCode < ApplicationRecord
     hcb_code "HCB Code"
     created_at "Created at"
     date "Transaction date"
-    url "URL" do |url| "https://hcb.hackclub.com#{url}" end
+    hashid "URL" do |hashid| "https://hcb.hackclub.com/hcb/#{hashid}" end
```
```suggestion
    hashid "URL" do |hcb_coide| Rails.application.routes.url_helpers.hcb_code_url(hashid) end
```

give this a test

PR #9002, opened 2024-12-18
Make the "recurring" column either "ACTIVE", "INACTIVE" or "FALSE"

sampoder:
Struggling to squeeze this into the already complex query, thought this would be cleaner.

Closes https://github.com/hackclub/hcb/issues/8944


garyhtou:
Thank you!

PR #9013, opened 2024-12-19
Enable timestamp honeypots on reimbursement reports

sampoder:


garyhtou (commented):
Just as an FYI, by default, the `timestamp_threshold` is 4 seconds. Do you think this might cause some false positives if the user only needs to type the report name (assuming email is autocomplete by the browser).

sampoder:
Let's start with 2 seconds?

app/controllers/reimbursement/reports_controller.rb
```diff
@@ -8,7 +8,7 @@ class ReportsController < ApplicationController
     skip_before_action :signed_in_user, only: [:show, :start, :create, :finished]
     skip_after_action :verify_authorized, only: [:start, :finished]
 
-    invisible_captcha only: [:create], on_spam: :on_spam
+    invisible_captcha only: [:create], on_spam: :on_spam, timestamp_enabled: true
```
sampoder:
```suggestion
    invisible_captcha only: [:create], on_spam: :on_spam, timestamp_enabled: true, timestamp_threshold: 2
```

PR #9016, opened 2024-12-19
Don't double render activities using a `@before` parameter

sampoder:
Essentially, we will paginate through all the reports before the page was originally loaded:

Screenshot 2024-12-18 at 5 24 05 PM

That way we're always paginating through the same set of activities.

Closes https://github.com/hackclub/hcb/issues/9008


app/controllers/events_controller.rb
```diff
@@ -69,7 +69,9 @@ def show
     @money_in = all_transactions.reject { |t| t.amount_cents <= 0 }.first(3)
     @money_out = all_transactions.reject { |t| t.amount_cents >= 0 }.first(3)
 
-    @activities = PublicActivity::Activity.for_event(@event).order(created_at: :desc).first(5)
+    @before = params[:before] || Time.now
```
garyhtou: Should we give this a more specific name? This action is used for more than just activities

sampoder: Yeah, that's going to get confusing. I'm going to get the user stuff in and then tackle the event stuff

app/views/my/_activities.html.erb
```diff
@@ -6,7 +6,7 @@
       <%= turbo_frame_tag(
             "activities-#{@activities.next_page}",
             loading: :lazy,
-            src: "/my/activities?page=#{@activities.next_page}",
+            src: "/my/activities?page=#{@activities.next_page}&before=#{@before}",
```
sampoder:
```suggestion
            src: "/my/activities?page=#{@activities.next_page}&before=#{@before || Time.now}",
```

PR #9018, opened 2024-12-19
Limits on issuing cards

emergenitro:
Reverting back to Issue #8267, there is excessive card issuing which has to be prevented. In order to prevent that, as discussed, there's been a limit on user's issuing of virtual cards to 10 and physical cards to 4. This was done by validating it when the card is being issued.

Below is how it looks if the max is reached:
image

To achieve this, we added validation methods in the `StripeCard` model: `user_physical_card_limit_not_exceeded` and `user_virtual_card_limit_not_exceeded`. These methods check the number of cards a user has issued within the organization (event) and prevent issuing new cards if the user has reached the maximum allowed.

I added a `:non_granted` scope to ensure these aren't grant cards, for which, there will be no limit. When staging, please check it out and ensure that everything is a-okay. Thanks!


db/migrate/20241219000855_add_card_grant_id_to_stripe_cards.rb
sampoder, line 1: We shouldn't need to add `card_grant_id` to `StripeCard`, you can already do `card.subledger.card_grant` to get the `CardGrant` and you can use the `on_main_ledger` scope to filter these out.

```
bank(prod)> Event.friendly.find("onboard").stripe_cards.count
=> 666
bank(prod)> Event.friendly.find("onboard").stripe_cards.on_main_ledger.count
=> 12
```

app/models/stripe_card.rb
```diff
@@ -61,17 +62,19 @@ class StripeCard < ApplicationRecord
 
   after_create_commit :notify_user, unless: :skip_notify_user
 
-  attr_accessor :skip_notify_user
+  attr_accessor :skip_notify_user, :enforce_card_limit
```
sampoder: Hmm, what's happening here?

app/models/stripe_card.rb
```diff
@@ -115,13 +118,35 @@ class StripeCard < ApplicationRecord
...
   end
 
+  def enforce_card_limit?
+    enforce_card_limit != false
+  end
+
+  def user_physical_card_limit_not_exceeded
+    return unless physical?
+
+    if user.stripe_cards.physical.non_granted.where(event:).count >= 4
```
sampoder:
```suggestion
    if user.stripe_cards.physical.on_main_ledger.where(event:).count >= 4
```

app/models/user.rb
```diff
@@ -116,6 +116,7 @@ class User < ApplicationRecord
   has_many :approved_expenses, class_name: "Reimbursement::Expense", inverse_of: :approved_by
 
   has_many :card_grants
+  has_many :cards
```
```suggestion
```

db/migrate/20241219000855_add_card_grant_id_to_stripe_cards.rb
emergenitro: Oh alrighty, perfect, I'll change that out then. Thanks!

app/models/stripe_card.rb
emergenitro: It's needed to be accessed right? I tried it without earlier and got this:
image

sampoder: Where's the value for `enforce_card_limit` coming from?

garyhtou:
Superseded by https://github.com/hackclub/hcb/pull/9863 due to git rewrite.

PR #9020, opened 2024-12-19
Fix of the number of cards

emergenitro:
Just a minor fix which was slightly bugging me. The card number on top of the org cards and personal cards page included the frozen and burned cards which felt wrong. I think active cards should only be considered in the number there.

Before: 
image

After:
image

i.e. in the above case I have 5 frozen and/or burnt cards.

I totally understand if it was intentioned to include those as well. However, I just thought it makes more sense to not include them.


garyhtou:
Superseded by https://github.com/hackclub/hcb/pull/9864 due to git rewrite.

PR #9024, opened 2024-12-19
Disable verifying phone number on behalf of another user

manuthecoder:
This pull request includes a minor but important change to the `app/views/users/edit.html.erb` file. The change ensures that the phone number verification warning is only shown to the user who owns the account.

* [`app/views/users/edit.html.erb`](diffhunk://#diff-7a664d6e0b1d39425223431cffbf717843cbac2e33ac4fe6a58ecb959e38f6abL54-R54): Added a condition to check if the current user is the owner of the account before displaying the phone number verification warning.

See #8907


app/views/users/edit.html.erb
```diff
@@ -51,7 +51,7 @@
               <%= form.label :phone_number, "Mobile phone number" %>
               <%= form.hidden_field :phone_number, required: true, placeholder: "555-555-5555", id: "phone_number" %>
               <input type="tel" id="phone_raw" style="max-width:100%;" value="<%= @user.phone_number || nil %>" required="required">
-              <% if @user.phone_number_in_database.present? && !@user.phone_number_verified? %>
+              <% if @user.phone_number_in_database.present? && !@user.phone_number_verified? && current_user == @user %>
```
garyhtou: > The change ensures that the phone number verification warning is only shown to the user who owns the account.

As an admin, it would still be nice to know if a user's phone number is not verified.

app/views/users/edit.html.erb
```diff
@@ -51,7 +51,7 @@
...
                 <p class="h5 mt1 m0 warning flex items-center">
                   <%= inline_icon "important-fill", class: "warning mr1" %>
                   <span>Your phone number is unverified. <%= link_to "Click here", "#", class: "info", data: { behavior: "modal_trigger", modal: "verify_phone" } %>
```
sampoder:
```suggestion
                  <span>Your phone number is unverified. <%= link_to "Click here", "#", class: "info", data: { behavior: "modal_trigger", modal: "verify_phone" } if current_user == @user %>
```

test this works, but I think it should do the trick

sampoder (approved):

PR #9036, opened 2024-12-19
show Merchant ID on non-card-grant Stripe charges

24c02:
this adds the MID to the merchant detail expand section & modal as well as making category slugs copiable ^_^


sampoder (approved):
LGTM!

PR #9059, opened 2024-12-23
Create `CardGrant::ZeroJob`

sampoder:
Closes https://github.com/hackclub/hcb/issues/9053. ZERO!


app/models/card_grant.rb
```diff
@@ -134,25 +134,27 @@ def expire!
...
-      ).run
+    disbursement = DisbursementService::Create.new(
+      source_event_id: event_id,
+      destination_event_id: event_id,
+      name: custom_memo,
+      amount: balance.amount,
+      source_subledger_id: subledger_id,
+      requested_by_id: canceled_by.id,
+    ).run
+
```
sampoder:
```suggestion

```

Issue #6880, opened 2024-06-24
User Tasks


garyhtou:
# Goal
The goal is to consolidate all action items that a user has. This may include uploading receipts or reviewing reimbursement reports.

# Design

https://www.figma.com/design/R9F4Qc0EUovbWJfDsDnV8z/Tasks?node-id=147-154&t=IUVsCGnPWdfpQ3hK-0

See the annotations section on the Figma for more details

malted:
What are the action items we're consolidating? To me, there is only one action item that operations cares about - receipts. The only other action item I can think of is reimbursement review, which is not an operations bottleneck and is solely user-to-user dependent. So it's not really an action items page, it's a receipt and reimbursement page right?

garyhtou:
We're adding more items soon! Ideas include:
- Sign your contract
- Approve a member's ACH transfer

garyhtou:
![image](https://github.com/hackclub/hcb/assets/20099646/d96d74ab-b3c2-401f-aa90-631ecf3ea493)
https://hackclub.slack.com/archives/C047Y01MHJQ/p1719304671539609?thread_ts=1719260469.748239&cid=C047Y01MHJQ

garyhtou:
@sampoder will take on building the model and basic page. Afterwards, we'll handle building the popover separately

YodaLightsabr:
Would a page like this still serve any meaningful purpose? It seems like it just duplicates tasks from the receipts page and the reimbursements page, but it doesn't actually eliminate the need for those pages.

sampoder:
We still don't have a concrete use of tasks - it's only receipts at the moment, so yeah, at the moment we don't need this page until we add more tasks (or abandon tasks)

PR #9060, opened 2024-12-24
Add app notification badging for admin tasks

manuthecoder:
### See #8563 for more information

This pull request introduces a new feature to update the Progressive Web App (PWA) badge based on the status of tasks in the admin interface. The most important changes include the addition of a new Stimulus controller and the integration of this controller into the admin tasks view.

New feature implementation:

* [`app/javascript/controllers/admin_tasks_pwa_badge_controller.js`](diffhunk://#diff-b2005edb91badf6f85f61d454cdd28d9b737a64586ef1e4fb3266db27153358fR1-R25): Added a new Stimulus controller to observe changes in the task grid and update the PWA badge accordingly. The controller uses a `MutationObserver` to monitor the task grid and sets the app badge based on the sum of the badge values once all tasks are updated.

Integration with admin tasks view:

* [`app/views/static_pages/_admin.html.erb`](diffhunk://#diff-8d09560bad3191cd2ddb9a7c9496f1c78fccfaa4821ef746a33dcdeabed79762L29-R31): Integrated the new `admin-tasks-pwa-badge` controller into the admin tasks section by adding the `data-controller` attribute to the relevant HTML element.

#### Preview
image


sampoder (approved):
This looks great! However, I think the intention was for this to be done using the `Task` model, see https://github.com/hackclub/hcb/issues/6880 and https://github.com/hackclub/hcb/pull/7513. You can do something like `current_user.tasks.incomplete.count` to find the number.

sampoder (changes requested):
See above - sorry!

app/views/static_pages/_admin.html.erb
```diff
@@ -26,7 +26,8 @@
     isPinned(id) {
       return this.pinned_cards.find(card => card.id === id)
     }
-    }">
+    }"
+    data-controller="admin-tasks-pwa-badge">
```
sampoder:
```suggestion
    }">
```

app/views/my/tasks.html.erb
sampoder, line 1: This should be on the homepage, this page is a hidden page and isn't really ever loaded

app/controllers/my_controller.rb
```diff
@@ -29,6 +29,10 @@ def tasks
     @tasks = current_user.tasks
   end
 
+  def tasks_count
+    render json: { count: current_user.tasks.count }
+  end
+
```
sampoder: We could probably simplify this into one tasks route:

```ruby
def tasks
	@tasks = current_user.tasks
	respond_to do |format|
	  format.html
	  format.json { render json: { count: @tasks.count } }
	end
end
```

And then:

```javascript
if (navigator.setAppBadge) {
  window.addEventListener("load", async () => {
    const { count } = await fetch("/my/tasks.json").then(res => res.json())
    navigator.setAppBadge(count)
  })
}
```

Just sanity check that this works - looking forward to getting this in!

https://stackoverflow.com/questions/20188047/rails-respond-to-json-and-html <- here's what I based this on

manu-gurudath:
ty!

sampoder (approved):
LGTM - let's get this in!

PR #9076, opened 2024-12-27
[Policy] Let grant cardholders see all card charge info

Luke-Oldenburg:
## Summary of the problem
Grant cardholders were unable to see info on charges they made. This was discovered while trying to view receipts on HCB Mobile. The only allowed for admins and organizers but grant card users are just a cardholder and not a member of the organization.

## Describe your changes
I added a check to see if the user from the transaction record was the user being checked for policy.


sampoder (changes requested):
Screenshot 2024-12-27 at 5 06 32 PM

FYI

Luke-Oldenburg (commented):
Removed the change to transaction policy and fixed the variables being references in the others.

Luke-Oldenburg:
Whoops I was trying to see if I could mark the request as resolved but I don't think I can because you marked it as requested changes and not a regular comment.

sampoder (changes requested):
The authorisation is done on a `HcbCode` not a CPT or CT:

```ruby
@hcb_code = authorize HcbCode.find_by_public_id!(params[:id])
```

Additionally, these policies have to not error for non-Stripe transactions

Luke-Oldenburg:
@sampoder so I need to be editing `HcbCodePolicy` instead? And as for not erroring, I just need to check that `stripe_cardholder` exists before checking the `user_id` equality right?

24c02:
safe navigation operator is your friend!
`stripe_cardholder&.user_id`

garyhtou:
Superseded by https://github.com/hackclub/hcb/pull/9861 due to git rewrite.

PR #9078, opened 2024-12-27
[V4 API] Add more info to checks

Luke-Oldenburg:
Adds more info to checks in the API. Doesn't support legacy checks but the file didn't already going off of the comments.


app/views/api/v4/transactions/_check.json.jbuilder
```diff
@@ -1,6 +1,13 @@
...
+json.address_line2 check.address_line2
+json.address_state check.address_state
+json.address_zip check.address_zip
 json.recipient_name check.recipient_name
-json.memo check.local_hcb_code.memo
+json.recipient_email check.recipient_email
+json.memo check.memo
 json.payment_for check.payment_for
+json.check_number check.check_number
 json.status check.is_a?(IncreaseCheck) ? check.state_text.parameterize(separator: "_") : nil # TODO: handle statuses for old Lob checks
```
sampoder: For things that'd break for old checks, let's handle it like this

sampoder (approved):

PR #9081, opened 2024-12-28
[V4 API] Add more info to ACHs

Luke-Oldenburg:
- Rename description to payment_for to match other endpoints
- Add full account number and routing number


sampoder (changes requested):
We're working towards hiding these numbers / making them less visible with the eventual goal of never showing an account number after it's been inputted. We could add routing numbers (though visually I think a bank name is more helpful) but we shouldn't expose the entire account number.

Luke-Oldenburg:
Sounds good. I was trying to mirror the data visible on the app as closely as possible to site.

sampoder (approved):
LGTM! Just if you get a chance in the app, update where we reference `.description` if we do

PR #9084, opened 2024-12-30
Make frozen cards icy

24c02:
here's how it looks is on virtual, black, and white cards:
CleanShot 2024-12-29 at 20 12 05@2x
forgive me if this isn't a design direction y'all want to take but i think it'd be cute!


sampoder (approved):
LGTM - thank you!

leowilkin:
that looks so cute

24c02:
CleanShot 2024-12-30 at 18 44 14@2x
think it's cooler not fully grayscaled

PR #9096, opened 2025-01-01
Add `signed_out_at` to `UserSession`

sampoder:
Progress towards https://github.com/hackclub/hcb/issues/8950. I'll work on a migration script so we can drop `deleted_at` 

@garyhtou do we also want to drop `peacefully_expired`?

Here's the migration script:

```ruby
UserSession.with_deleted.where("deleted_at < expiration_at").update_all("expiration_at = deleted_at, signed_out_at = deleted_at")
```


garyhtou:
Test: ensure previously logged out sessions remain logged out

PR #9109, opened 2025-01-01
Remove arcade card from admin dash

leowilkin:
## Summary of the problem
Arcade has ended, the card is no longer required.

## Describe your changes
Closes #9104


sampoder (approved):
LGTM - thank you!

PR #9119, opened 2025-01-02
Add "Upload another receipt" copy to inline receipt upload buttons

sampoder:
Closes https://github.com/hackclub/hcb/issues/8903


sampoder:
need to add to CPT + test if this slows things down

PR #9130, opened 2025-01-03
Create `OrganizerPosition::Contract#creator`

sampoder:


app/models/organizer_position/contract.rb
```diff
@@ -178,6 +178,13 @@ def one_non_void_contract
       end
     end
 
+    def creator
+      user_id = versions.first&.whodunnit
+      return nil unless user_id
+
```
sampoder:
```suggestion

```

PR #9150, opened 2025-01-06
Create the `Export` model

sampoder:


app/models/export.rb
```diff
@@ -0,0 +1,54 @@
...
+#
+#  fk_rails_...  (requested_by_id => users.id)
+#
+class Export < ApplicationRecord
+  belongs_to :requested_by, class_name: "User", optional: true
+  validates_presence_of :requested_by, if: -> { async? }
+
+  after_initialize do
+    raise "Cannot directly instantiate an Export" if self.instance_of? Export
+  end
```
garyhtou, line 28: `self.abstract_class = true`

app/models/export.rb
```diff
@@ -0,0 +1,54 @@
...
+#
+#  index_exports_on_requested_by_id  (requested_by_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (requested_by_id => users.id)
+#
+class Export < ApplicationRecord
+  belongs_to :requested_by, class_name: "User", optional: true
+  validates_presence_of :requested_by, if: -> { async? }
```
garyhtou, line 24: after_save :upload_file
has_one_attached :file

then, ExportMailer would send a link to the s3 file.

app/controllers/exports_controller.rb
```diff
@@ -12,18 +12,55 @@ def transactions
...
+                      public_only: !organizer_signed_in?
+                    )
+                  end
+
+        if @export.async?
+          @export.requested_by = User.find_or_create_by(email: params[:email]) if params[:email]
+
+          if @export.requested_by.nil?
+            redirect_to collect_email_exports_path(file_extension:, event_slug: params[:event]) and return
+          end
```
garyhtou, line 49: Let's always set requested_by. It would be `params[:email] || current_user`

app/models/export.rb
sampoder:
```suggestion
  self.abstract_class = true
```

app/controllers/exports_controller.rb
sampoder: We should already do this in lines above

app/models/export.rb
sampoder: I'm going to get this in because I need it to solve another ticket, I'll add in S3 as soon as I can (most likely when files get that large) but I think this PR is still a net positive to get in in its current state

app/models/export.rb
sampoder: FYI: Screenshot 2025-03-27 at 11 38 10 PM

PR #9156, opened 2025-01-07
Give [Card Grants] some Purpose

24c02:
if you had multiple card grants from the same event w/ different restrictions (currently i think only High Seas is doing this, but that'll probably change!) it wasn't always easy to tell which one you could spend where (users have to guess based off MIDs, not all of which are yellow_page'd)
no more!
this PR adds a purpose column to card grants!
CleanShot 2025-01-07 at 13 01 48@2x
CleanShot 2025-01-07 at 12 53 06@2x
CleanShot 2025-01-07 at 12 53 14@2x
CleanShot 2025-01-07 at 12 56 30@2x
CleanShot 2025-01-07 at 12 56 43@2x


app/controllers/card_grants_controller.rb
```diff
@@ -46,6 +46,18 @@ def create
     redirect_to event_transfers_path(@event)
   end
 
+  def update
+    authorize @card_grant
+
+    if @card_grant.update(params.require(:card_grant).permit(:purpose))
+      flash[:success] = "Grant's purpose has been successfully updated!"
+    else
+      flash[:error] = @card_grant.errors.full_messages.to_sentence || "Grant's purpose could not be updated"
```
garyhtou: The left side will always be a truthy value, meaning `"Grant's purpose could not be updated"` will never be shown

```suggestion
      flash[:error] = @card_grant.errors.full_messages.to_sentence
```

app/views/card_grants/_details.html.erb
```diff
@@ -76,6 +76,12 @@
         <span>All categories allowed</span>
       <% end %>
     </p>
+    <% if @card_grant.purpose.presence %>
```
garyhtou: What you have technically works, but this is more clear
```suggestion
    <% if @card_grant.purpose.present? %>
```

app/views/card_grants/_edit_form.html.erb
```diff
@@ -0,0 +1,26 @@
...
+    </div>
+  </div>
+
+  <div class="actions">
+    <%= form.submit "Save" %>
+  </div>
+<% end %>
+
+<script>
+  window.clearName = () => {
```
garyhtou: Should we rename this to `clearPurpose`?

app/views/card_grants/actions/_edit.erb
```diff
@@ -0,0 +1,17 @@
+<% if card_grant.status == "active" %>
```
garyhtou:
```suggestion
<% if card_grant.active? %>
```

app/views/card_grants/_actions.html.erb
```diff
@@ -8,5 +8,8 @@
 
     <%= render "card_grants/actions/topup", card_grant: %>
     <%= render "card_grants/actions/cancel", card_grant: %>
+    <% if policy(card_grant).update? %>
+      <%= render "card_grants/actions/edit", card_grant: %>
+    <% end %>
```
garyhtou: I'd recommend moving the policy check inside the partial. So, the first line of the partial will be:

```erb
<% if card_grant.active? && policy(card_grant).update? %>
```

app/views/card_grants/actions/_edit.erb
```diff
@@ -0,0 +1,17 @@
...
+  <div class="tooltipped tooltipped--n" aria-label="Set this grant's purpose">
+    <%= link_to edit_card_grant_path(card_grant),
+                class: "btn bg-success",
+                data: { behavior: "modal_trigger", modal: "name_card" } do %>
+      <%= inline_icon "edit" %>
+      Set purpose
+    <% end %>
+  </div>
+
+  <% if policy(card_grant).update? %>
```
garyhtou: Since you're moving the policy check from outside of the partial to the first line of this partial.

This line 11 should no longer be needed.

app/views/card_grants/show.html.erb
```diff
@@ -14,6 +14,9 @@
   <% elsif @card_grant.user == current_user %>
     <div>
       <%= render partial: "canceled_warning", locals: { card_grant: @card_grant } %>
+      <% if @card_grant.purpose.presence %>
```
garyhtou:
```suggestion
      <% if @card_grant.purpose.present? %>
```

app/views/stripe_cards/_stripe_card.html.erb
```diff
@@ -44,8 +44,11 @@
     </p>
     <div class="stripe-card__overlay"></div>
   </div>
-  <div class="mt2 muted center"><%= stripe_card.name %></div>
-
+  <% if defined? card_label %>
+    <div class="mt2 muted center"><%= card_label %></div>
+  <% else %>
+      <div class="mt2 muted center"><%= stripe_card.name %></div>
```
garyhtou:
```suggestion
    <div class="mt2 muted center"><%= stripe_card.name %></div>
```

app/views/card_grants/show.html.erb
```diff
@@ -14,6 +14,9 @@
...
+        <h2><%= @card_grant.purpose %></h2>
+      <% end %>
       <%= render @card_grant.stripe_card, headless: true, show_card_number: true %>
```
garyhtou: Instead of adding a H2 here, should we pass `card_label` in line 20?

app/models/card_grant.rb
```diff
@@ -10,6 +10,7 @@
 #  email           :string           not null
 #  keyword_lock    :string
 #  merchant_lock   :string
+#  purpose         :string
```
garyhtou: Can we validate a max length? Maybe 30 or 40 chars?

just so that we don't have a lot of overflow on pages

app/views/card_grants/show.html.erb
24c02: i kinda visually prefer the heading here but i'm not married to it

app/models/card_grant.rb
24c02: sure!

24c02:
@garyhtou rspec is failing bc migrations are pending and i'm not sure how to solve that, do you have any advice?

garyhtou:
try running `bin/rails db:migrate`. it should update schema.rb, although be wary of unrelated schema changes. for this migration, i'd expected two lines changed.
1. the timestamp at the top
2. adding the purpose column to card grants

24c02:
oh i forgot to commit schema.rb lmao

24c02:
there are a bunch of other changes from past migrations though, is it cool to just not add them?

app/views/card_grants/_edit_form.html.erb
```diff
@@ -0,0 +1,26 @@
+<%= form_with(model: @card_grant, id: "card_grant_edit_form") do |form| %>
+  <div class="field">
+    <%= form.label :purpose %>
+
+    <div class="flex gap-3 items-stretch">
+      <%= form.text_field :purpose, placeholder: "OpenAI Credits" %>
+      <% if @card_grant.purpose.present? %>
+        <button class="btn bg-error px-3 tooltipped tooltipped--s overflow-visible" aria-label="Clear purpose" id="clear_purpose_btn" onclick="clearPurpose();" type="button">
+          <%= inline_icon "delete", class: "!m-0" %>
+        </button>
```
sampoder: Imo, this should be done as a post request to a route like `CardGrant#clear_purpose`

```ruby
def clear_purpose
  authorize @card_grant, :update?
  @card_grant.update(purpose: nil)
  redirect_back fallback_location: card_grant_url(@card_grant)
end
```

To me, that feels cleaner than doing this client side. Thoughts?

app/views/card_grants/actions/_edit.erb
sampoder, line 1: This should probably be a HTML file

app/views/card_grants/new.html.erb
```diff
@@ -22,6 +22,11 @@
       </div>
     </div>
 
+    <div class="field">
+      <%= form.label :purpose, "Purpose (optional, shown to recipient)" %>
+      <%= form.text_field :purpose, placeholder: "OpenAI Credits" %>
```
sampoder:
```suggestion
      <%= form.text_field :purpose, placeholder: "Pizza for a club meeting" %>
```
Feels more neutral to me

app/views/card_grants/_edit_form.html.erb
24c02: i'd be cool with that, i was following the way stripe card naming works

app/views/card_grants/actions/_edit.erb
24c02: how does that work with the form getting rendered in there?

sampoder: Something like `_edit.html.erb` should work, not sure why the form wouldn't work

app/views/card_grants/_edit_form.html.erb
sampoder: Ah... we should change that at some point imo

24c02:
CleanShot 2025-01-23 at 14 03 04@2x
snazzy new clear button:

app/views/static_pages/index.html.erb
```diff
@@ -34,7 +34,7 @@
 
   <div class="grid grid--wide mb-12">
     <% current_user.card_grants.includes(:stripe_card).activated.each do |grant| %>
-      <%= render grant.stripe_card, headless: true, href: card_grant_path(grant) %>
+      <%= render grant.stripe_card, headless: true, href: card_grant_path(grant), card_label: grant.purpose %>
```
garyhtou: Should we do the same for card grants listed on `/my/cards`?

24c02: oh yeah absolutely! might add that tomorrow :-)

app/controllers/card_grants_controller.rb
```diff
@@ -46,6 +46,26 @@ def create
...
+    end
+
+    redirect_to card_grant_url(@card_grant)
+  end
+
+  def clear_purpose
+    authorize @card_grant, :update?
+    @card_grant.update(purpose: nil)
+    # Dark Souls-style because who's ever gonna see this
+    flash[:success] = "PURPOSE CLEARED..."
```
```suggestion
```
lol I'm going to remove this so it doesn't confuse people in the future and they expect a flash

sampoder (approved):
LGTM - let's do it!

PR #9157, opened 2025-01-08
Create `Column::SweepJob`

sampoder:
Goal is to resolve https://github.com/hackclub/hcb/issues/9155


app/jobs/column/sweep_job.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      ColumnService.post("/transfers/ach", {
+        idempotency_key:,
+        amount: difference.abs,
+        currency_code: "USD",
+        type:,
+        entry_class_code: "CCD",
+        counterparty: {
+          account_number: Rails.application.credentials.svb[:account_number],
+          routing_number: Rails.application.credentials.svb[:routing_number],
```
garyhtou: I just sent u a message via Slack— we should be wary of overdraft here.

garyhtou (commented):
Once we're ready, this job should be scheduled.

PR #9173, opened 2025-01-09
Fix Balance Graph

rluodev:
## Summary of the problem

The balance graph didn't work

## Describe your changes

the balance graph now works

took me like an hour to find root cause 💀


app/javascript/controllers/balance_graph_controller.js
```diff
@@ -19,7 +19,7 @@ export default class extends Controller {
     )
   }
   connect() {
-    fetch(`/${this.slugValue}/balance_by_date`)
+    fetch(`${window.location.pathname.trimStart("/")}/balance_by_date`)
```
sampoder: This breaks the balance graph on `/epoch/transactions`. We need to be setting the value through the Stimulus controller. Probably worth checking the data attributes

Thanks for finding this as the cause!

sampoder (changes requested):

sampoder:
Chatted with Ruien, fyi: https://github.com/hackclub/hcb/pull/9175

PR #9174, opened 2025-01-09
Fix associated report links

sampoder:
Interesting reads linked on https://stackoverflow.com/questions/75738570/getting-a-turbo-frame-error-of-content-missing


app/views/hcb_codes/reimbursement/_expense_payout.html.erb
```diff
@@ -42,7 +42,7 @@
       <% end %>
       <p>
         <strong>Associated Report</strong>
-        <%= link_to expense.report.name, expense.report %>
+        <%= link_to expense.report.name, expense.report, data: { turbo_frame:"_top" } %>
```
sampoder:
```suggestion
        <%= link_to expense.report.name, expense.report, data: { turbo_frame: "_top" } %>
```

PR #9179, opened 2025-01-09
Paginate donation page

rluodev:
## Summary of the problem

donation page was long

## Describe your changes

now is short and has pages if it was long before


sampoder (approved):
LGTM - nice work here!

PR #9192, opened 2025-01-12
Remove post mortem

Luke-Oldenburg:
This PR removes the post mortem from the dev-docs. I don't think it is relevant to new contributors as I haven't seen anything done like this recently and it is from 2 years ago. If there are other post mortems I'm unaware of maybe it is best to keep them all in one place and link to them as examples for those wishing to document their changes more thoroughly.


sampoder:
We should keep this for historical purposes

PR #9198, opened 2025-01-14
Fix cards showing from organizations you are no longer part of

Ved235:
closes #9193


garyhtou (dismissed):
Good catch! I'd suggest updating the `@stripe_cards` variable defined here: https://github.com/hackclub/bank/blob/002aa011daf74a80ba8991198541026901ce3c09/app/controllers/my_controller.rb#L21

Ved235:
> Good catch! I'd suggest updating the `@stripe_cards` variable defined here: https://github.com/hackclub/bank/blob/002aa011daf74a80ba8991198541026901ce3c09/app/controllers/my_controller.rb#L21

I have created a new commit implementing this

app/controllers/my_controller.rb
```diff
@@ -21,7 +21,7 @@ def cards
     @stripe_cards = current_user.stripe_cards.includes(:event).order(
       Arel.sql("stripe_status = 'active' DESC"),
       Arel.sql("stripe_status = 'inactive' DESC")
-    )
+    ).select { |card| current_user.events.include?(card.event) }
```
sampoder:
```suggestion
    ).where(event: current_user.events)
```

sampoder (changes requested):
We need to think a bit about this, there are a couple of cases where we want to show these cards, admins might still want to see these and will still have access

Ved235:
> We need to think a bit about this, there are a couple of cases where we want to show these cards, admins might still want to see these and will still have access

Should we implement this for non admins then?

app/controllers/my_controller.rb
```diff
@@ -21,7 +21,7 @@ def cards
     @stripe_cards = current_user.stripe_cards.includes(:event).order(
       Arel.sql("stripe_status = 'active' DESC"),
       Arel.sql("stripe_status = 'inactive' DESC")
-    )
+    ).where(event: current_user.events)
```
sampoder:
```suggestion
    ).where(event: current_user.admin? ? Event.all : current_user.events)
```

sampoder (approved):

sampoder:
LGTM - let's go!

app/controllers/my_controller.rb
```diff
@@ -21,8 +21,7 @@ def cards
     @stripe_cards = current_user.stripe_cards.includes(:event).order(
       Arel.sql("stripe_status = 'active' DESC"),
       Arel.sql("stripe_status = 'inactive' DESC")
-    )
-
+    ).where(event: current_user.admin? ? Event.all : current_user.events)
```
garyhtou: I believe this may hide card grants; if they were shown before.

sampoder: That is correct, let me get that fixed rq

PR #9201, opened 2025-01-14
Add `Metric::Hcb::Stats` to calculate stats every hour

YodaLightsabr:
## Summary of the problem

Our `/stats` endpoint is really slow and can take down the entire app when spammed.

## Describe your changes

Moves the `/stats` endpoint into a job that runs every hour


app/controllers/stats_controller.rb
```diff
@@ -28,60 +28,8 @@ def stats_custom_duration
   def stats
     return render plain: "too many requests", status: :too_many_requests unless Flipper.enabled?(:stats_endpoint)
 
-    json = Rails.cache.fetch("stats", expires_in: 1.hour, race_condition_ttl: 60 * 5) do
-      now = params[:date].present? ? Date.parse(params[:date]) : DateTime.current
-      year_ago = now - 1.year
-      qtr_ago = now - 3.months
-      month_ago = now - 1.month
-      week_ago = now - 1.week
+    json = Metric::Hcb::Stats.last
```
garyhtou:
```suggestion
    json = Metric::Hcb::Stats.metric
```

app/jobs/metric_jobs/calculate_single.rb
```diff
@@ -5,6 +5,9 @@ class CalculateSingle < ApplicationJob
     queue_as :wrapped
 
     def perform(metric_class, record = nil)
+      if metric_class.is_a?(String) && metric_class.starts_with?("Metric::")
+        metric_class = metric_class.safe_constantize
+      end
```
garyhtou: Rather than updating this method, i'd suggest creating a new job (called by cron) that does:

```ruby
MetricJob::CalculateSingle.perform_now(Metric::Hcb::Stats)
```

garyhtou:
it's almost here!

garyhtou (approved):

PR #9206, opened 2025-01-14
[Card Grants] When merchant is disallowed, redirect users for help

YodaLightsabr:
## Summary of the problem
People always ask in #hcb when their card grant doesn't accept DigiKey purchases (among others).

## Describe your changes
The new declined mailer instructs users to reach out to the team behind their card grant.


garyhtou (approved):
good call

PR #9211, opened 2025-01-15
[My Cards] Hide cancelled cards inside an accordion 

manuthecoder:
This PR hides canceled cards in a `<details>` tag, as requested by #9200

image


app/views/my/cards.html.erb
```diff
@@ -22,19 +22,30 @@
...
+    </article>
+  </details>
 <% end %>
 
 <%= turbo_frame_tag :my_missing_receipts_list, src: my_missing_receipts_list_path, target: "_top" %>
 
 <% if @emburse_cards.any? %>
   <details class="mt3">
-    <summary><h2 class="inline-block mt0">Emburse cards</h2></summary>
+  <summary><h2 class="inline-block mt0">Emburse cards</h2></summary>
```
sampoder:
```suggestion
    <summary><h2 class="inline-block mt0">Emburse cards</h2></summary>
```

app/controllers/my_controller.rb
```diff
@@ -22,7 +22,11 @@ def cards
       Arel.sql("stripe_status = 'active' DESC"),
       Arel.sql("stripe_status = 'inactive' DESC")
     )
+
     @emburse_cards = current_user.emburse_cards.includes(:event)
+
+    @active_stripe_cards = @stripe_cards.where(stripe_status: "active")
+    @canceled_stripe_cards = @stripe_cards.where(stripe_status: "canceled")
```
sampoder: Currently frozen cards and cards not initially activated will be hidden entirely. We should probably show the ones pending initial activation outside of the accordion and the frozen ones inside the accordion. We'd have to change the langauge to accommodate that change.

Luke-Oldenburg: Imo frozen cards should still be grouped with the rest just moved to the bottom. I typically will freeze cards I'm not currently using but I will still access them again at some point unlike cancelled cards.

garyhtou: @Luke-Oldenburg  +1

sampoder: Cool with that too!

app/controllers/my_controller.rb
```diff
@@ -22,7 +22,11 @@ def cards
       Arel.sql("stripe_status = 'active' DESC"),
       Arel.sql("stripe_status = 'inactive' DESC")
     )
+
     @emburse_cards = current_user.emburse_cards.includes(:event)
+
+    @active_stripe_cards = @stripe_cards.where(stripe_status: "active")
```
garyhtou:
```suggestion
    @active_stripe_cards = @stripe_cards.where.not(stripe_status: "canceled")
```

garyhtou (approved):
amazing!

PR #9212, opened 2025-01-15
Update donation settling logic

sampoder:
Because we're no longer doing fee reimbursements, we're charging ourselves the fee


garyhtou:
thank you for this

PR #9228, opened 2025-01-15
[Rails] Finish upgrade to 7.1

KiKoS0:
Finishing enabling Rails 7.1 defaults, just after this PR and following https://github.com/discourse/discourse/commit/84823550d48a730a8ee1922e5f2e7565435a422e I'll create a follow up one that:

1. Bumps `config.load_defaults` to 7.1
2. Copies any config value that isn't the default Rails 7.1
3. Set the `cache_format_version` to 7.1
	https://github.com/hackclub/hcb/blob/b8c1174edd26c15144e050c54778651e1a947161/config/initializers/new_framework_defaults_7_1.rb#L240-L244
4. Deletes the migrating file


config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -17,8 +17,9 @@
 #
 # This will reduce the size of the load path, making `require` faster if you don't use bootsnap, or reduce the size
 # of the bootsnap cache if you use it.
-#++
-# Rails.application.config.add_autoload_paths_to_load_path = false
+#
+# To set this configuration, add the following line to `config/application.rb` (NOT this file):
+#   config.add_autoload_paths_to_load_path = false
```
KiKoS0: Seems like the `new_framework_defaults_7_1.rb' migration file was
generated before this commit that fixes it was made (before it the option wasn't being applied):

https://github.com/rails/rails/commit/07715ba3ce29cb4c5b9df4d0acb3ac80b544165e#diff-b80cd24acfb76b1dea49328100ec246cd42aa6abce8b824c1f10cc6de2508f99R20

But in general I don't think there are any occurrences of requiring
autoloaded files unless I missed it so let me know if I did.

I verified that the `$LOAD_PATH` is changed after the config change, this is the diff:
image

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -152,24 +153,24 @@
...
 #
 # `config.load_defaults 7.1` does not set this value for environments other than
 # development and test.
 #++
-# if Rails.env.local?
-#   Rails.application.config.log_file_size = 100 * 1024 * 1024
-# end
+if Rails.env.local?
+  Rails.application.config.log_file_size = 100 * 1024 * 1024
+end
```
KiKoS0: Should be safe since it's only applied for dev and test environments.

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -208,7 +209,7 @@
 # leave this optimization off on the first deploy, then enable it on a
 # subsequent deploy.
 #++
-# Rails.application.config.active_record.marshalling_format_version = 7.1
+Rails.application.config.active_record.marshalling_format_version = 7.1
```
KiKoS0: I'm not exactly sure where marshalling happens in the app (possibly
before caching?) but it seems that it's a backwards compatible change;
Meaning that payloads dumped with the older algorithm can still be
recreated into records by the new one.
https://github.com/rails/rails/pull/47747/files#diff-210d0f245e618b7f8c6fcb53f518bd7fd343ce7e5d8c609ecfdb149418beca19R1138-R1141

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -152,24 +153,24 @@
...
+  Rails.application.config.log_file_size = 100 * 1024 * 1024
+end
 
 ###
 # Enable raising on assignment to attr_readonly attributes. The previous
 # behavior would allow assignment but silently not persist changes to the
 # database.
 #++
-# Rails.application.config.active_record.raise_on_assign_to_attr_readonly = true
+Rails.application.config.active_record.raise_on_assign_to_attr_readonly = true
```
KiKoS0: There is no usage of `attr_readonly` in the codebase.

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -152,24 +153,24 @@
 # leave this optimization off on the first deploy, then enable it on a
 # subsequent deploy.
 #++
-# Rails.application.config.active_support.use_message_serializer_for_metadata = true
+Rails.application.config.active_support.use_message_serializer_for_metadata = true
```
KiKoS0: This one is backwards compatible as well and should work fine since we
are already running a 7.1+ version.

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -139,7 +140,7 @@
 # servers, first deploy without changing the serializer, then set the serializer
 # in a subsequent deploy.
 #++
-# Rails.application.config.active_support.message_serializer = :json_allow_marshal
+Rails.application.config.active_support.message_serializer = :json_allow_marshal
```
KiKoS0: Backwards compatible as well since `json_allow_marshal` option will
fallback to unmarshalling from older nodes. The option is still present
in Rails 7.2 so it should be fine.

Since we have `msgpack` as an indirect dependency we can switch to
it later after switching to Rails 7.1 defaults. It's more compact and
should perform better.

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -82,7 +83,7 @@
 ###
 # Disable deprecated singular associations names.
 #++
-# Rails.application.config.active_record.allow_deprecated_singular_associations_name = false
+Rails.application.config.active_record.allow_deprecated_singular_associations_name = false
```
KiKoS0: If I understand correctly, this does absolutely nothing given that we
are running the 7.2 version in which this option has been deprecated:
https://github.com/rails/rails/blob/8c7e39497f069e354d67ed14c63fa31383871e5d/guides/source/7_2_release_notes.md#deprecations-5

garyhtou:
Thank you for these really thorough comments!

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -152,24 +153,24 @@
...
 # Set the maximum size for Rails log files.
 #
 # `config.load_defaults 7.1` does not set this value for environments other than
 # development and test.
 #++
-# if Rails.env.local?
-#   Rails.application.config.log_file_size = 100 * 1024 * 1024
-# end
+if Rails.env.local?
+  Rails.application.config.log_file_size = 100 * 1024 * 1024
```
northeastprince:
```suggestion
  Rails.application.config.log_file_size = 100.megabytes
```

config/routes.rb
```diff
@@ -2,7 +2,6 @@
 
 require "sidekiq/web"
 require "sidekiq/cron/web"
-require "admin_constraint"
```
albertchae: I see from https://github.com/hackclub/hcb/pull/9228/commits/f01fc801154774117e9ba1357eaddef18d0bab90 it is removed because it's autoloaded

albertchae (approved):

garyhtou:
Great work here @KiKoS0!

PR #9235, opened 2025-01-16
[Wires] Store and display failure reason to user and ops

rluodev:
## Summary of the problem

people got confused why wire failed

## Describe your changes

now we show why the wire failed

image
image
image
image


app/models/wire.rb
```diff
@@ -294,6 +295,7 @@ class Wire < ApplicationRecord
       after do |reason: nil|
         WireMailer.with(wire: self, reason:).notify_failed.deliver_later
         create_activity(key: "wire.failed", owner: nil)
+        failure_reason = reason
```
```suggestion
        self.failure_reason = reason
```
And sorry I know this is pedantic but it should be `return_reason` imo

sampoder (changes requested):
Largely looks good - thank you!

app/models/wire.rb
rluodev: @sampoder rationale for it being failure reason is because the state we have is "failed"

sampoder (approved):

PR #9236, opened 2025-01-16
Remove `preferred_first_name`

rluodev:
## Summary of the problem

we don't need this method

## Describe your changes

method no longer exists


garyhtou (approved):
amazingg

PR #9237, opened 2025-01-16
Remove non-prod master.key before precompiling assets

KiKoS0:
@garyhtou I think the issue is that when initializing the app during precompilation it tries to decrypt the `credentials.yml.enc` file with the non-prod master key. So I believe we should delete the file before running it and make the calls to `Rails.application.credentials` lazy during init.

Let me know if this works or not, i also encountered this but I'm not yet sure if it's only me:
```
Step 25/29 : RUN groupadd --system --gid 1000 rails &&     useradd rails --uid 1000 --gid 1000 --create-home --shell /bin/bash &&     chown -R rails:rails log tmp storage db
 ---> Running in 086420a5e197
chown: cannot access 'storage': No such file or directory
```


Dockerfile.production
```diff
@@ -62,6 +62,9 @@ RUN yarn install --frozen-lockfile
 # Copy application code
 COPY . .
 
+# Remove non-prod master key
+RUN rm -f config/master.key
```
KiKoS0: I tried using a custom `<dockerfile-name>.dockerignore` but it only works with BuildKit: https://stackoverflow.com/a/57774684/9437824

config/initializers/preload_stis.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#option-3-preload-a-regular-directory
-unless Rails.application.config.eager_load
+if ENV["SECRET_KEY_BASE_DUMMY"].nil? && !Rails.application.config.eager_load
```
KiKoS0: I'm open for a better way to do this

garyhtou:
Thank you so much for this @KiKoS0. It looks like I'm also running into the same issue:

```
 => ERROR [stage-2 3/3] RUN groupadd --system --gid 1000 rails &&     use  7.5s
------
 > [stage-2 3/3] RUN groupadd --system --gid 1000 rails &&     useradd rails --uid 1000 --gid 1000 --create-home --shell /bin/bash &&     chown -R rails:rails log tmp storage db:
6.638 chown: cannot access 'storage': No such file or directory
------
Dockerfile.production:81
--------------------
  80 |     # Run and own only the runtime files as a non-root user for security
  81 | >>> RUN groupadd --system --gid 1000 rails && \
  82 | >>>     useradd rails --uid 1000 --gid 1000 --create-home --shell /bin/bash && \
  83 | >>>     chown -R rails:rails log tmp storage db
  84 |     USER 1000:1000
--------------------
ERROR: failed to solve: process "/bin/sh -c groupadd --system --gid 1000 rails &&     useradd rails --uid 1000 --gid 1000 --create-home --shell /bin/bash &&     chown -R rails:rails log tmp storage db" did not complete successfully: exit code: 1
```

garyhtou:
I believe `storage` is a folder that's created only if Active Storage is configured to use `Disk` service. We only have this [configured for development](https://github.com/hackclub/bank/blob/b822950bd51cabc5abc064d893f2b9b82bf7b8ef/config/storage.yml#L5-L7).

So, it is likely that the `storage` folder will not exist when we're building for production (especially on a CI server).

garyhtou:
I see a few solutions:
1. Remove `storage` from the command
2. Add a `.keep` to `storage`
3. Update the command to not error if the folder does not exist

garyhtou:
For now, I've created a `.keep` file for the `storage` directory to unblock us. Looking at a newly generated Rails app (Rails 8), it comes with a `storage/.keep`

config/initializers/preload_stis.rb
garyhtou: I'm assuming this change is because eager loading the app will run code that accesses `Rails.application.credentials`?

> make the calls to Rails.application.credentials lazy during init

Dockerfile.production
garyhtou: Clever attempt!

garyhtou:
I'm going to merge this into https://github.com/hackclub/hcb/tree/rluodev/3609-productionize-dockerfile to unblock us.

PR #9241, opened 2025-01-17
Add a little bit of margin to canceled cards section

sampoder:
Testing it in production and it felt like it needed this


garyhtou:
+1, thank you for this

PR #9245, opened 2025-01-17
Make Rails asset host configurable

garyhtou:
## Summary of the problem

When deploying to a non-`hcb.hackclub.com` site with Rail's asset host set to our usual production AWS CloudFront host, the site will attempt to fetch assets from the `hcb.hackclub.com` site. I believe this fails because the digest keys are different.

## Describe your changes

By omitting `ASSET_HOST` from env var, Rails will bypass the CloudFront CDN; allowing the deployment to serve images directly. This is useful for testing an alternative production deployment, but not ideal for a _real_ production deployment (CDN's still better, ofc).


garyhtou:
I've already set the `ASSET_HOST` env var in real production (hcb.hackclub.com). So this PR should deploy seamlessly since it's merged.

PR #9249, opened 2025-01-17
Ensure users are members of event they are pinning to + fix pinning disbursements

sampoder:
Closes https://github.com/hackclub/hcb/issues/9218


app/controllers/hcb_codes_controller.rb
```diff
@@ -87,9 +87,10 @@ def edit
 
   def pin
     @hcb_code = HcbCode.find(params[:id])
-    @event = @hcb_code.event
+    @event = @hcb_code.events.select { |e| e.id == Event.friendly.find_by_friendly_id(params[:event]).id }.first || @hcb_code.event
```
```suggestion
    param_event = Event.friendly.find_by_friendly_id(params[:event])
    @event = param_event ? @hcb_code.events.find_by(id: param_event.id) : @hcb_code.event
```

PR #9251, opened 2025-01-17
Remove email & sms methods from `Receipt::SYNCHRONOUS_SUGGESTION_UPLOAD_METHODS`

sampoder:
Because these methods didn't use direct uploads, HCB needs a little more time before it can extract the receipt text. Because uploads are guaranteed to be done after a commit. 

https://github.com/hackclub/hcb/issues/4067#issuecomment-2591701863


sampoder:
By putting them into a job, we'll retry extracting which should help

PR #9266, opened 2025-01-18
Revert "Revert "Ignore columns on PaymentRecipient post-migration""

sampoder:
Reverts hackclub/hcb#9265


sampoder:
Migrated that old data - turned out they didn't have emails, hence the issues but this is all fixed

PR #9269, opened 2025-01-19
Bump Rails config defaults to 7.1

KiKoS0:
## Summary of the problem
Bumping Rails config defaults to 7.1 following this finished migration https://github.com/hackclub/hcb/pull/9228.


config/application.rb
```diff
@@ -45,7 +45,7 @@ class Application < Rails::Application
 
     config.react.camelize_props = true
 
-    config.add_autoload_paths_to_load_path = false
```
albertchae: cool looks like `load_defaults 7.1` handles the autoload config

https://github.com/rails/rails/blob/14c115b120ed089331ff3dc13f36bd9129ced33d/railties/lib/rails/application/configuration.rb#L272

It looks like it may handle `cache_format_version` as well but I agree it's safer to set this explicitly since the codebase is already on 7.2

albertchae (approved):

garyhtou (approved):
Great, thank you!

PR #9278, opened 2025-01-20
[Receipts] Fix early return bug in pairing model

YodaLightsabr:
## Summary of the problem
We had an early `return` inside of a block in the `::ReceiptService::Suggest` service.

## Describe your changes
Refactors the code to use a nested if statement instead of an early return


sampoder (approved):
LGTM!

PR #9279, opened 2025-01-20
Fix transaction table width on mid-size laptop screens

YodaLightsabr:
## Summary of the problem
The transaction table doesn't fully expand to the width of the screen on certain mid-size laptop screens

## Describe your changes
Comparison images can be found here: [https://hackclub.slack.com/archives/C06JEH8HJSK/p1737358235237309](https://hackclub.slack.com/archives/C06JEH8HJSK/p1737358235237309) (before: left, after: right)


sampoder:
FYI on mobile, we probably don't want to do this because it the width of the transaction table becomes more than 100%:

Screenshot 2025-01-20 at 10 49 32 AM

PR #9284, opened 2025-01-20
[Rack::Attack] Safelist Stripe webhook IPs

rluodev:
## Summary of the problem

stripe could end up ratelimited

## Describe your changes

no longer can stripe be ratelimited... unlimited powah!!!


sampoder (approved):
LGTM!

PR #9296, opened 2025-01-21
Is it a bird? Is it a plane? No It's 1Password!

devramsean0:
## Summary of the problem

Reactivates the 1Password perk with updated copy, and a new form.

## Describe your changes

N/A

> [!NOTE]
> The disabled button is just an artifact of an old screenshot. Assume it is identical to the active one

image
Screenshot 2025-01-21 210141


devramsean0:
Ready to merge

app/views/events/promotions.html.erb
```diff
@@ -73,6 +73,41 @@
...
+    <%= link_to "Apply",
+      @perks_available ? fillout_form("7jzG7ADmuLus", { "Project_Name" => @event.name, "Bank_URL" => event_url(@event), "Your_Name" => current_user.name, "Email" => current_user.email }, prefix: "prefill_") : root_path,
+      class: "btn mr1",
+      target: :_blank %>
+  </div>
+
+  <% if @event.hackathon? %>
+  <p class="mb0">
+    <%= link_to "1Password", "https://1password.com" %> is offering one year’s worth of 1Password
+    for families (which is just a normal 1Password plan, for up to 5 people)
```
garyhtou:
```suggestion
    for families (for up to 5 people)
```

app/views/events/promotions.html.erb
```diff
@@ -73,6 +73,41 @@
...
+    to all Hackathons running on HCB for you to hand out as a prize to your participants.
```
garyhtou:
```suggestion
    to all hackathons running on HCB for you to hand out as a prize to your participants.
```

garyhtou: This feels like a run on sentence.

garyhtou: Question: What happens if someone uses this for their org instead of giving it out as a prize?

app/views/events/promotions.html.erb
```diff
@@ -73,6 +73,41 @@
...
+    This can only be claimed once per Hackathon. 
+  </p>
+    <% else %>
+      <h3 class="mb0 mt0">Do you make Open Source Software?</h3>
+      <p class="mb0">
+      <%= link_to "1Password", "https://1password.com" %> is offering a Lifetime 1password Teams
+      license (Up to 10 free seats) with access to their world class development tools (That’s up
+      to 10 team members!) to any Open Source and non-commercial project running on HCB!
```
garyhtou: All projects on HCB should be non-commerical. Unless projects like ArchiveBox are disallowed, I think we might be able to just say Open Source project running on HCB. ArchiveBox as a nonprofit and company side.

app/views/events/promotions.html.erb
```diff
@@ -73,6 +73,41 @@
...
+      status as part of the Hack Foundation). This can only be claimed once per organization
+      </p>
+      <h3 class="mb0 mt2">Are you neither a hackathon, or a OSS project?</h3>
+      <p class="mb0">
+      <%= link_to "1Password", "https://1password.com" %> is offering an indefinite 25% off of
+      a business plan to nonprofits running through HCB! While we can’t apply for you, we can
+      help you through the process (and provide you proof of your status as part of the Hack Foundation).
+      This can only be claimed once per organization.
+    </p>
+  <% end%>
```
garyhtou: Can you please fix the indent here?

app/views/events/promotions.html.erb
devramsean0: The verbosity was due to daisy's concerns/confusion surrounding the plan name.
We can replace it, but I'd like something that gets the same message across

app/views/events/promotions.html.erb
devramsean0: Nothing (on our side anyway)!
1Password have made it clear to me though, that it's designed to be offered as a prize, although there is nothing technically requiring it.

app/views/events/promotions.html.erb
devramsean0: This is a weird one, again it comes down to a specific program that 1Password run [(1Password for Open Source)](https://github.com/1Password/for-open-source). which is specifically for organisations that make Open Source software.

Unless, we are talking about the stuff on LN 97, and in that case I agree

app/views/events/promotions.html.erb
devramsean0: Done!

PR #9297, opened 2025-01-21
Waive fees on returned wires

sampoder:


app/services/fee_engine/create.rb
```diff
@@ -39,6 +39,7 @@ def determine_reason
       reason = :revenue_waived if canonical_transaction.likely_check_clearing_dda? # this typically has a negative balancing transaction with it
       reason = :revenue_waived if canonical_transaction.likely_card_transaction_refund? # sometimes a user is issued a refund on a transaction
       reason = :revenue_waived if canonical_transaction.local_hcb_code.ach_transfer? # outgoing ACH transfers are sometimes returned to the account upon failure
+      reason = :revenue_waived if canonical_transaction.local_hcb_code.wire? # outgoing wires are sometimes returned to the account upon failure
```
garyhtou: I think we should give it a different reason; similar to `donation_refunded`

garyhtou (dismissed):

app/services/fee_engine/create.rb
sampoder: Maybe we create a new reason `transfer_returned`?

garyhtou: yuy, that's good

PR #9304, opened 2025-01-22
Add 1 day / 7 days reminders to reimbursement reports

sampoder:
Closes https://github.com/hackclub/hcb/issues/9292. I did it on creating the report as to not send more than two reminders. I'm using the reports `updated_at` because we touch the report through expense so that variable can be used (also makes the email copy most accurate).

Flipper flagging because i want to test this in production on my user before enabling for everyone


app/jobs/reimbursement_job/one_day_reminder.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module ReimbursementJob
+  class OneDayReminder < ApplicationJob
+    queue_as :low
+    def perform(report)
+      return unless report.draft?
+
+      if report.updated_at < 24.hours.ago
```
garyhtou:
```suggestion
      if report.updated_at < 23.hours.ago
```

If we did 24 hours, then in practice, people would only receive an email after 2 days. Here's why:
- Jan 1st, 1:00 am: create report
- Jan 1st, 1:05 am: add expense
- Jan 2nd, 1:00 am: `OneDayReminder` runs, but requeues because `Jan 1, 1:05am < Jan 1, 1:00 am` is `false`.
- Jan 3rd, 1:00 am: `OneDayReminder` runs and actually sends email (`Jan 1, 1:05am < Jan 2, 1:00 am` is `true`.

So, I suggest we make the check here `23` hours to make the "1 day" more realistic. This assumes that when someone creates a report, they generally finish their edits within an hour.

app/jobs/reimbursement_job/seven_days_reminder.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module ReimbursementJob
+  class SevenDaysReminder < ApplicationJob
+    queue_as :low
+    def perform(report)
+      return unless report.draft?
+
+      if report.updated_at < 7.days.ago
```
garyhtou:
```suggestion
      if report.updated_at < 6.days.ago
```

For the same reason as above, maybe we do 6 days here? Or:

```suggestion
      if report.updated_at < (6.days + 23.hours).ago
```

sampoder: 6 days should be safe to do!

app/jobs/reimbursement_job/one_day_reminder.rb
sampoder: Good catch, thanks! Agreed.

PR #9322, opened 2025-01-22
Don't omit stats on the CardOnly plan

sampoder:
Closes https://github.com/hackclub/hcb/issues/9314


garyhtou:
We should probably update this to remove it.
https://github.com/hackclub/bank/blob/739454b4018c8b9ad891109abfa8b1f382e82bd2/app/models/metric/hcb/total_raised.rb#L36

Even better is to use `Event::Plan.that(:omit_stats)` there

garyhtou:
We might also have some stuff in Metabase that needs updating

sampoder:
Everything really should be using `Event::Plan.that(:omit_stats)`

garyhtou:
> Everything really should be using `Event::Plan.that(:omit_stats)`

Yeah +1. I don't think there's a way to make that happen in metabase tho

PR #9349, opened 2025-01-24
[Donation] Store referrer and UTM parameters

YodaLightsabr:
## Summary of the problem

Fixes #9348 

Right now, it's hard to find out where donors are coming from

## Describe your changes

- Stores the initial request referer as well as the `utm_source`, `utm_medium`, and `utm_campaign` parameters
- If present, renders them on the HCB code show page

image


app/models/donation.rb
```diff
@@ -317,6 +317,14 @@ def email
     recurring_donation&.email || super
   end
 
+  def referrer_domain
+    Addressable::URI.parse(referrer.presence)&.host
+  end
+
+  def referrer_favicon_url
+    referrer_domain && "https://t0.gstatic.com/faviconV2?client=SOCIAL&type=FAVICON&fallback_opts=TYPE,SIZE,URL&url=https://#{URI::Parser.new.escape(referrer_domain)}&size=256"
+  end
```
garyhtou: Very nice touch!

app/models/donation.rb
```diff
@@ -317,6 +317,14 @@ def email
     recurring_donation&.email || super
   end
 
+  def referrer_domain
+    Addressable::URI.parse(referrer.presence)&.host
+  end
+
+  def referrer_favicon_url
+    referrer_domain && "https://t0.gstatic.com/faviconV2?client=SOCIAL&type=FAVICON&fallback_opts=TYPE,SIZE,URL&url=https://#{URI::Parser.new.escape(referrer_domain)}&size=256"
```
garyhtou: If `referrer_domain` is `nil`, I believe this method would return false. How about we make it return `nil` too?

```suggestion
    return unless referrer_domain

    "https://t0.gstatic.com/faviconV2?client=SOCIAL&type=FAVICON&fallback_opts=TYPE,SIZE,URL&url=https://#{URI::Parser.new.escape(referrer_domain)}&size=256"
```

garyhtou: I'm recommend trying `URI::HTTP.build`

https://stackoverflow.com/questions/6152231/is-there-a-ruby-library-gem-that-will-generate-a-url-based-on-a-set-of-parameter

app/models/donation.rb
garyhtou, line 1: I think you're missing model annotations. Try running `bin/rails annotate_models`

app/views/hcb_codes/_donation.html.erb
```diff
@@ -196,3 +213,42 @@
...
+      <%= @donation.utm_source || unknown %>
+    </p>
+    <p>
+      <strong>UTM medium</strong>
+      <%= @donation.utm_medium || unknown %>
+    </p>
+    <p>
+      <strong>UTM campaign</strong>
+      <%= @donation.utm_campaign || unknown %>
+    </p>
```
garyhtou: It looks like there are technically 5 UTM params. Do we want all 5? Or are the other two almost never used?

app/views/hcb_codes/_donation.html.erb
```diff
@@ -196,3 +213,42 @@
     </article>
   </section>
 <% end %>
+
+<section class="modal modal--scroll bg-snow" data-behavior="modal" role="dialog" id="referrer_details_<%= instance %>">
+  <% header = capture do %>
+    <% if @donation.referrer_favicon_url.present? %>
+      <img src="<%= @donation.referrer_favicon_url %>" height="36" width="36" class="mr-2 rounded-lg align-bottom" alt="Icon indicating referrer">
+    <% end %>
+    <%= @donation.referrer_domain || @donation.referrer || "Unknown" %>
```
garyhtou:
```suggestion
    <%= @donation.referrer_domain || @donation.referrer || "Referrer and UTM parameters" %>
```

i think this might look better? although I don't know for sure since I'm just reading the code and don't have the page rendered

app/models/donation.rb
garyhtou, line 1: Add validation to `strip` all fields and silently truncate them to some number of characters

garyhtou (commented):
this is amazing! just a few small things

app/views/hcb_codes/_donation.html.erb
```diff
@@ -196,3 +213,42 @@
...
+      <%= @donation.utm_medium || unknown %>
+    </p>
+    <p>
+      <strong>UTM campaign</strong>
+      <%= @donation.utm_campaign || unknown %>
+    </p>
+  </div>
+  <article class="card__banner secondary border-top" style="margin: -2em; margin-top: 0;">
+    <p class="my0 muted">
+      Referrer and UTM information are used to track the source of donations. This data comes from the donor's browser and may not always be accurate.
```
garyhtou:
```suggestion
      Referrer and UTM information are used to understand the source of donations. This data is provided by the donor's browser and may not always be accurate.
```

app/models/donation.rb
YodaLightsabr: I don't think it would return false if `referrer_domain` is `nil`, but I might be misunderstanding what you're saying

image

app/views/hcb_codes/_donation.html.erb
YodaLightsabr: We could include those as well

app/views/hcb_codes/_donation.html.erb
YodaLightsabr: Good catch, maybe something along the lines of "Referrer and UTM parameters"

app/views/hcb_codes/_donation.html.erb
YodaLightsabr: Good way of putting it

app/models/donation.rb
YodaLightsabr: I tried using this, but it messes with the encoding in a way we can't control. Google's server is picky about it for some reason.

app/views/donations/_donation_form.html.erb
```diff
@@ -106,6 +106,11 @@
       <%= form.hidden_field :background, value: @background, name: :background %>
     <% end %>
 
+    <%= form.hidden_field :utm_source %>
+    <%= form.hidden_field :utm_medium %>
+    <%= form.hidden_field :utm_campaign %>
```
garyhtou: Missing UTM term and content

app/views/hcb_codes/_donation.html.erb
```diff
@@ -133,6 +134,22 @@
         </span>
       </span>
     </p>
+    <% if [@donation.referrer, @donation.utm_source, @donation.utm_medium, @donation.utm_campaign].any?(&:present?) %>
```
garyhtou: Missing UTM term and content

garyhtou (commented):
🪄 largely good to go. just a few things to clean up:

- missing UTM term and content in a few places
- figure out why rspec is failing

app/models/donation.rb
garyhtou: Oh, interesting! you are right. I had no idea. TIL

app/views/hcb_codes/_donation.html.erb
```diff
@@ -196,3 +213,50 @@
...
+      <%= @donation.utm_term || unknown %>
+    </p>
+    <p>
+      <strong>UTM content</strong>
+      <%= @donation.utm_content || unknown %>
+    </p>
+  </div>
+  <article class="card__banner secondary border-top" style="margin: -2em; margin-top: 0;">
+    <p class="my0 muted">
+      Referrer and UTM information are used to understand the source of donations. This data is provided by the donor's browser and may not always be accurate.
```
sampoder:
```suggestion
      Referrer and UTM information are used to understand the source of donations. This data is provided by the donor's browser and may not be accurate.
```
I think if we drop the "always" this won't overflow to a third line - sorry pet peeve

sampoder (approved):
Looks good to me!

app/views/hcb_codes/_donation.html.erb
YodaLightsabr: That's fair; I hadn't tested how it looked after the copy change

garyhtou:
Good to merge! You've got the 🪄!

PR #9362, opened 2025-01-26
Report Puma stats to AppSignal

sampoder:
See https://github.com/hackclub/hcb/issues/9332


garyhtou:
image

this is why i love appsignal

PR #9382, opened 2025-01-27
Create Rails 7.2 new framework defaults

KiKoS0:
## Summary of the problem
Creates new framework default config for Rails 7.2

- https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-7-1-to-rails-7-2

The discourse example migrated to the 7.2 config without going through this process, the only thing that they seem to have disabled is yjit: https://github.com/discourse/discourse/pull/28420


config/initializers/new_framework_defaults_7_2.rb
```diff
@@ -0,0 +1,72 @@
...
+# topic in the database.
+#
+# If `enqueue_after_transaction_commit` is set to `:default`, the queue adapter
+# will define the behaviour.
+#
+# Note: Active Job backends can disable this feature. This is generally done by
+# backends that use the same database as Active Record as a queue, hence they
+# don't need this feature.
+#++
+Rails.application.config.active_job.enqueue_after_transaction_commit = :default
```
KiKoS0: As far as i understand this keeps the same behavior since sidekiq's default's to true which means it only enqueues after commiting a transaction: https://github.com/sidekiq/sidekiq/blob/66e44316385d7a76e5e49dc27772da164ff70615/lib/active_job/queue_adapters/sidekiq_adapter.rb#L30-L32

Reference: https://github.com/rails/rails/pull/53375#issuecomment-2550640069

config/initializers/new_framework_defaults_7_2.rb
```diff
@@ -0,0 +1,72 @@
...
+
+###
+# Adds image/webp to the list of content types Active Storage considers as an image
+# Prevents automatic conversion to a fallback PNG, and assumes clients support WebP, as they support gif, jpeg, and png.
+# This is possible due to broad browser support for WebP, but older browsers and email clients may still not support
+# WebP. Requires imagemagick/libvips built with WebP support.
+#++
+Rails.application.config.active_storage.web_image_content_types = %w[image/png image/jpeg image/gif image/webp]
```
KiKoS0: Do we still need this if it's now an accepted type?
https://github.com/hackclub/hcb/blob/bea8f3b7579d11a3b3ed8fa18cad75acb160706d/config/initializers/active_storage.rb#L7

config/initializers/new_framework_defaults_7_2.rb
```diff
@@ -0,0 +1,72 @@
...
+
+###
+# Controls whether the PostgresqlAdapter should decode dates automatically with manual queries.
+#
+# Example:
+#   ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.select_value("select '2024-01-01'::date") #=> Date
+#
+# This query used to return a `String`.
+#++
+Rails.application.config.active_record.postgresql_adapter_decode_dates = true
```
KiKoS0: This one should be fine as we never call the adapter directly?

config/initializers/new_framework_defaults_7_2.rb
albertchae: oh cool, reduces the need for https://github.com/palkan/isolator

config/initializers/new_framework_defaults_7_2.rb
albertchae: interesting, maybe not but we can resolve that as a follow up

config/initializers/new_framework_defaults_7_2.rb
albertchae: yep seems fine and even if there was code that got a String I'm sure it called some Date new or parse on it immediately, which would be a pretty obvious and fixable error

config/initializers/new_framework_defaults_7_2.rb
```diff
@@ -0,0 +1,72 @@
...
+
+###
+# Enables YJIT as of Ruby 3.3, to bring sizeable performance improvements. If you are
+# deploying to a memory constrained environment you may want to set this to `false`.
+#++
+# Rails.application.config.yjit = true
```
albertchae: looks like hcb already has yjit enabled

https://github.com/hackclub/hcb/blob/ac967bb90c81157d8a73e078fa6564d62ae6dd54/config/initializers/yjit.rb

we can probably turn this setting on and delete the yjit initializer in this PR

albertchae (commented):
looks good, reviewed while comparing this to https://github.com/rails/rails/blob/v7.2.2.1/railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_7_2.rb.tt

albertchae (approved):

PR #9383, opened 2025-01-27
[CI] Suggest Rubocop changes

YodaLightsabr:
## Summary of the problem
Frequently, Rubocop will get upset because someone puts one too many spaces. It's good that Rubocop is looking out for us, but it takes a while to open the IDE back up and address the changes. It would be nice if Rubocop did it for us.

## Describe your changes
Rubocop will now comment suggested changes in GitHub using the [Suggested Changes Action](https://github.com/marketplace/actions/suggest-changes-action).


sampoder (approved):
LGTM!

app/controllers/admin/ledger_audits/tasks_controller.rb
```diff
@@ -8,8 +8,8 @@ def index
       end
 
       def show
-        @task = Admin::LedgerAudit::Task.find(params[:id])
-        @ledger_audit = @task.admin_ledger_audit
+         @task = Admin::LedgerAudit::Task.find(params[:id])
+         @ledger_audit = @task.admin_ledger_audit
```
YodaLightsabr:
```suggestion
        @task = Admin::LedgerAudit::Task.find(params[:id])
        @ledger_audit = @task.admin_ledger_audit
```

YodaLightsabr:
I'm merging this for now to try it out. It shouldn't affect the outcomes of the Rubocop check, it's just providing additional information.

If we find this to be annoying after a while, we can get rid of it.

garyhtou:
oh i'm going to love this sooo much

PR #9384, opened 2025-01-28
Provide breakdown of fees for PayPal transfers

YodaLightsabr:
## Summary of the problem
I was honestly a little confused when I saw the PayPal fee info on the new PayPal transfer page.
image

## Describe your changes
image

I think this should be a slightly better explanation? @sampoder correct me if any if the information is wrong.


sampoder:
Thanks for doing this! Yes, that's so much cleaner.

sampoder (approved):
Thank you!

PR #9402, opened 2025-01-29
Auto-accept receipts with a low distance on their `SuggestedPairing`

sampoder:


YodaLightsabr:
I think `< 20` might be a little bit too high. What do you think? I think it might also be better to just improve the pairing suggestion UI to show multiple at once (& use Turbo Streams, etc.) if that's becoming a bottleneck.

sampoder:
Sorry this PR isn't super well documented yet

Screenshot 2025-02-01 at 10 58 24 AM

I'd based the 20 number on this - though we could certainly drop it

> I think it might also be better to just improve the pairing suggestion UI to show multiple at once (& use Turbo Streams, etc.) if that's becoming a bottleneck.

The suggestion here was just for receipts that get emailed in, and I guess in the future we could also do SMS receipts as well.

Screenshot 2025-02-01 at 11 01 56 AM

@zachlatta can probably expand on the original thinking behind the suggestion but my understanding was it was when you're in a rush so you're emailing in receipts.

I do agree we could allow for viewing multiple suggestions at a time

garyhtou:
LGTM

PR #9432, opened 2025-01-31
[User] Require 2FA for admins

sampoder:
Closes https://github.com/hackclub/hcb/issues/9423


app/models/user.rb
```diff
@@ -160,6 +160,12 @@ class User < ApplicationRecord
 
   validate :profile_picture_format
 
+  validate on: :create do
```
```suggestion
  validate on: :update do
```

app/models/user.rb
```diff
@@ -160,6 +160,12 @@ class User < ApplicationRecord
 
   validate :profile_picture_format
 
+  validate on: :update do
+    if admin_override_pretend? && !use_two_factor_authentication?
+      errors.add(:access_level, "two factor authentication is required for this access level")
```
garyhtou: This error message would read funny: "Access level two factor authentication is required for this access level".

I suggest either adding the error on the `:base` or `:use_two_factor_authentication`.
- `errors.add(:base, "two factor authentication is required for this access level")`
   "two factor authentication is required for this access level"
- `errors.add(:use_two_factor_authentication, "is required for this access level")`
  "Use two factor authentication is required for this access level"

The first one (`:base`) reads better!

PR #9436, opened 2025-02-01
[Receipts] Compare missing receipt count to string (because it's a string)

YodaLightsabr:
## Summary of the problem
image

https://hackclub.slack.com/archives/C068U0JMV19/p1738407930525849


app/views/my/missing_receipts_icon.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= turbo_frame_tag "/my/missing_receipts_icon" do %>
-  <% if @missing_receipt_count != 0 %>
+  <% if @missing_receipt_count != "0" %>
```
garyhtou: We should figure out why this is a string. An int might make more sense.

YodaLightsabr: This was because of #9280. I've added a comment in #9448 about what we could do going forward.

PR #9437, opened 2025-02-02
Protect from double-refunding

sampoder:
Should be a little cleaner


garyhtou (commented):
In this job, we should return without erroring. If we error, the job will just keep being retried.

I suggest:
1. Inside the job, return without erroring
2. Before queueing, check that the donation has not yet been refunded

PR #9441, opened 2025-02-03
Add safe navigators to ct on unknown HCB codes

sampoder:


garyhtou:
Hm, i'm curious how this is possible. An unknown HCB code means that there's no linked object and no pending transactions. That means in order for an unknown HCB code to be created, there should be at least one CT.

However, maybe we're getting errors on this due to old HCB Codes that exist but no longer hold transactions?

garyhtou:
The issue is that HackPad accidentally issued two grants (old system). They were never and will never be processed. So I deleted them and their pending transactions. I've asked Alex Ren to re-issue them as actual card grants.

PR #9447, opened 2025-02-04
Clarify that we need an IBAN number

sampoder:


app/views/wires/new.html.erb
```diff
@@ -120,7 +120,7 @@
   </div>
 
   <div class="field">
-    <%= form.label :account_number %>
+    <%= form.label :account_number, "IBAN account number" %>
```
garyhtou: How about "International account number (IBAN)" to prevent saying "international bank account number account number"?

sampoder: Yeah - I like that
```suggestion
    <%= form.label :account_number, "International account number (IBAN)" %>
```

PR #9454, opened 2025-02-04
fix dependabot?

garyhtou:


.github/dependabot.yml
```diff
@@ -14,6 +14,7 @@ updates:
       timezone: "America/New_York"
   # Update Yellow Pages daily
   - package-ecosystem: "bundler"
+    directory: "/"
     schedule:
       interval: "daily"
```
garyhtou:
```suggestion
      interval: "daily"
      time: "10:00"
      timezone: "America/New_York"
```

garyhtou:
>Update configs must have a unique combination of 'package-ecosystem', 'directory', and 'target-branch'. Ecosystem 'bundler' has overlapping directories.

i guess it's not possible to have `yellow_pages` update more often than other bundler gems

PR #9472, opened 2025-02-04
Remove 2FA requirement for `db:seed` in development

manuthecoder:
## Summary of the problem
Running `bin/rails db:seed` fails in development because the seed file does not generate two-factor authentication codes. This PR removes the requirement for 2FA in development.


garyhtou:
good catch!

PR #9475, opened 2025-02-04
Fix the spelling of `minimum_wire_amount_cents`

sampoder:
Very embarrassing


sampoder:
thanks @YodaLightsabr for finding this...

PR #9478, opened 2025-02-05
Make `InvoicePayout#create_stripe_payout` public

sampoder:


garyhtou:
Was this temporary to manually retry the payouts? Is so, in the future, i'd recommend
```
invoice_payout.send(:create_stripe_payout)
```

The `send` method allows you to call private methods

PR #9495, opened 2025-02-06
Report to rails error reporter instead of airbrake

rluodev:


app/services/partners/plaid/transactions/get.rb
```diff
@@ -55,7 +55,7 @@ def fetch_transactions(offset: 0)
 
             results
           rescue ::Plaid::ApiError => error
```
garyhtou:
```suggestion
          rescue ::Plaid::ApiError => e
```

config/initializers/airbrake_subscriber.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AirbrakeSubscriber
+  def report(error, handled:, severity:, context:, source: nil)
+    Airbrake.notify(error, { context: context, handled: handled, level: severity })
```
garyhtou:
```suggestion
    Airbrake.notify(error, { context:, handled:, severity:, source: })
```

garyhtou (approved):

PR #9497, opened 2025-02-06
Track state changes in Check Deposits via webhooks

sampoder:
Closes https://github.com/hackclub/hcb/issues/9493, and also refactors the old system that was built in the transaction engine?


garyhtou (dismissed):
This should be done via a webhook from Column

PR #9498, opened 2025-02-06
Add commas to money input

manuthecoder:
This PR enhances the user experience by automatically adding commas to large numbers while typing in the transaction filter fields. This improves readability and usability when dealing with large amounts.

See #9442


app/assets/javascripts/ui.js
```diff
@@ -208,27 +208,28 @@ $(document).on('turbo:load', function () {
...
+      }
+  
+      $(this).val(value);
+    });
+  }
+    
+  attachMoneyInputListener();
+  
+  const observer = new MutationObserver(() => attachMoneyInputListener());
+  observer.observe(document.body, { childList: true, subtree: true });
```
sampoder: What's the intention behind using a `MutationObserver` here vs. how it was done before? Is it because of Turbo changing the document?

manu-gurudath: Rendering a menu re-renders its items and event listeners fail

sampoder:
```suggestion
  function attachMoneyInputListener() {
    $('input[data-behavior~="money"]').off('input').on('input', function () {
      let value = $(this)
        .val()
        .replace(/,/g, '') // remove all commas
        .replace(/[^0-9.]+/g, '') // remove non-numeric/non-dot characters
        .replace(/\B(?=(\d{3})+(?!\d))/g, ','); // add commas for thousands
  
      if (value.includes('.')) {
        let parts = value.split('.');
        value = parts[0] + '.' + (parts[1] ? parts[1].substring(0, 2) : '');
      }
  
      $(this).val(value);
    });
  }

  attachMoneyInputListener();

  // Used to attach the money input listener to inputs inside of menus / popups.

  const observer = new MutationObserver(() => attachMoneyInputListener());
  observer.observe(document.body, { childList: true, subtree: true });
```

sampoder (approved):
LGTM!

PR #9499, opened 2025-02-06
[1/4] Migrate from CountryEnumable to ISO3166::Country Alpha2

### LINKED ISSUE #8386: [`CountryEnumable`] Migrate to storing alpha2 and drop `country_enum_list`
At the moment, our `CountryEnumable` uses both the `countries` gem (`ISO3166::Country`) in addition to our own custom-defined hash located in [`CountryEnumable::country_enum_list`](https://github.com/hackclub/bank/blob/cd3adb8039c58e491c356bdef03f2b3dfaa7cef9/app/models/concerns/country_enumable.rb#L22)

The goal is to drop our custom-defined `CountryEnumable::country_enum_list` and solely depend on the `countries` gem for our list of countries. To make this happen, we need to migrate from storing a country as an integer to storing it as alpha2.

manuthecoder:
### 🧠 Context
We're deprecating our custom `CountryEnumable::country_enum_list` (which maps countries to integers) in favor of `ISO3166::Country`, which uses standard alpha2 codes (`"US"`, `"IN"`, etc.). This shift will simplify country handling across the codebase and reduce redundancy.

This is **Step 1** of a 4 part migration process

---

### ✅ Changes in this PR

- Adds new `country_alpha2` (string) columns via migration:
  - `events.country_alpha2`
  - `wires.country_alpha2`
- Adds `before_save` hooks to auto-populate `country_alpha2` based on existing:
  - `event.country`
  - `wire.recipient_country`
- Removes:
  - `include CountryEnumable`
  - `has_country_enum`
  from the `Event` model to begin decoupling from the legacy enum system.

---

### Overview

1. Add `country_alpha2` columns + save hooks (this PR ✅
2. Backfill alpha2 values from existing integer data
3. Update app logic to use `country_alpha2` (UI, filters, validations)
4. Drop old enum fields + delete `CountryEnumable`

---

Existing behavior using `event.country` or `wire.recipient_country` should remain intact for now


app/models/concerns/country_enumable.rb
```diff
@@ -16,10 +16,9 @@ def self.countries_for_select
     def has_country_enum(field: :country)
       enum field, self.country_enum_list, prefix: field
     end
+  end
 
-    private
-
-    def country_enum_list
+    def self.country_enum_list
```
garyhtou: fix indent

db/migrate/20241212164625_add_country2_alpha_to_wires.rb
```diff
@@ -0,0 +1,5 @@
+class AddCountry2AlphaToWires < ActiveRecord::Migration[7.2]
+  def change
+    add_column :wires, :country_alpha2, :string
```
garyhtou:
```suggestion
    add_column :wires, :address_country_alpha2, :string
```

garyhtou (commented):
We should also update [`User::PayoutMethod::Wire`](https://github.com/hackclub/hcb/blob/29da00915de572899f114bf5bbb0f6da3b2f3ebe/app/models/user/payout_method/wire.rb#L17)

PR #9501, opened 2025-02-06
[Google Workspace] Fix a bug that hid account listing for `g_suite`s created before papertrail was added

rluodev:
## Summary of the problem

function that relied on papertrail wasn't returning true if org didn't have papertrail before

## Describe your changes

now we add a bypass


app/views/events/g_suite_overview.html.erb
```diff
@@ -274,7 +274,7 @@
           </ol>
         </article>
       </details>
-      <% if !@g_suite.verified? && @g_suite.previously_verified? %>
+      <% if (!@g_suite.verified? && @g_suite.previously_verified?) || @g_suite.accounts.present? %>
```
```suggestion
      <% if (!@g_suite.verified? && @g_suite.previously_verified?) || @g_suite.accounts.any? %>
```

PR #9503, opened 2025-02-07
Create AirbrakeSubscriber.rb for initial rails error reporter testing

rluodev:
we create the subscriber and subscribe it to hear errors and also create a new route guaranteed to error so we can test how these errors show up on Airbrake and Appsignal


app/controllers/error_controller.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class ErrorController < ApplicationController
```
garyhtou: Can you make this controller something else? We have a similar named `ErrorsController` and I don't wanna them to mess with each other — although I doubt it will.

garyhtou:
Based on reading the docs, i think we need to manually pass request and user stuff to the `context` param. We'll see if this is actually the case.

PR #9508, opened 2025-02-07
[Stripe Card] Improve shipping ETA

garyhtou:
## Summary of the problem

Based on user feedback, the Stripe-provided shipping ETA is fairly inaccurate.

## Describe your changes

I'm adding an extra 2 days to the Stripe ETA to hopefully provide our users with more realistic expectations.

This PR also adds `RecrusiveOpenStruct` to better stub/mock Stripe's API objects in case they cannot be retrieved (e.g., in an outage or in development without proper API keys). Previously, it'd just error the whole page.

No UI changes.


app/views/stripe_card_mailer/physical_card_ordered.html.erb
```diff
@@ -1,4 +1,9 @@
 <p>Hey! Your new card from HCB is on the way 💳💸.</p>
-<p><strong>It will probably arrive around <%= Time.at(@eta).strftime("%A") %> (<%= Time.at(@eta).strftime("%d").to_i.ordinalize %>).</strong></p>
+<% if @eta %>
+  <p><strong>
+    It will probably arrive around <%= @eta.strftime("%A") %>
+    (<%= @eta.strftime("%d").to_i.ordinalize %>).
+  </strong></p>
+<% end %>
```
garyhtou: According to Stripe's docs, it's technically possible to have a null ETA, but we never handled that case in this mailer.

PR #9510, opened 2025-02-07
Route based on `.to` and not using the standard "routing" regex

sampoder:
Closes https://github.com/hackclub/hcb/issues/9509, this feels natural? At least to me.


garyhtou (commented):
Yeah, this LGTM.

Just as a FYI, this like means that these emails will go to the fallback mailbox and end up forwarded to hcb@

That's good since it'll get some human review/filtering. However, we don't want too many to go down that route — given that this is the first time we've seen this type of error. I'd think it's rare. So, we're good here!

sampoder:
Sounds good - I'll cancel the retries as well

PR #9513, opened 2025-02-08
Add card grant user callout & add UTM params to outbound links

YodaLightsabr:
## Summary of the problem
We have a ton of teens using HCB, but many of them only use card grants! It would be great to get card grant users inspired to run nonprofits and hackathons using HCB.

## Describe your changes
This PR adds UTM tags to outbound https://hackclub.com links to help us track how people are finding the HCB landing page and what strategies increase conversion. It also adds a navbar callout for card grant users:

image


app/models/user.rb
```diff
@@ -351,6 +351,10 @@ def sync_with_loops
     UserService::SyncWithLoops.new(user_id: id, new_user:).run
   end
 
+  def only_card_grant_user?
+    creation_method == "card_grant" && card_grants.count == 1
```
garyhtou: i'm curious: why exactly one card grant?

app/views/organizer_position/contracts_mailer/notify.html.erb
garyhtou, line 1: Can you do some research on how UTM links may affect email deliverability?

I wouldn't want these emails to end up on the Promotions page of Gmail because we added UTM query params.

app/helpers/static_pages_helper.rb
```diff
@@ -148,8 +148,8 @@ def airtable_info
     }
   end
 
-  def apply_form_url(user = current_user)
-    "https://hackclub.com/fiscal-sponsorship/apply/?#{URI.encode_www_form({ userEmail: user.email, firstName: user.first_name, lastName: user.last_name, userPhone: user.phone_number, userBirthday: user.birthday&.year }.compact)}"
+  def apply_form_url(user = current_user, campaign: nil, content: nil)
+    "https://hackclub.com/fiscal-sponsorship/apply/?#{URI.encode_www_form({ userEmail: user.email, firstName: user.first_name, lastName: user.last_name, userPhone: user.phone_number, userBirthday: user.birthday&.year, utm_source: "hcb", utm_medium: "web", utm_campaign: campaign, utm_content: content }.compact)}"
   end
```
garyhtou: Would it be better to allow any query param to be passed in?

So, something like:

```suggestion
  def apply_form_url(user = current_user, **query_params)
    query_params = { userEmail: user.email, firstName: user.first_name, lastName: user.last_name, userPhone: user.phone_number, userBirthday: user.birthday&.year, utm_source: "hcb", utm_medium: "web" }.merge(query_params) # allow method arguments to override default.
    "https://hackclub.com/fiscal-sponsorship/apply/?#{URI.encode_www_form(query_params.compact)}"
  end
```

Usage would then change to:

```ruby
apply_form_url(utm_campaign: "hcb-new-event", content: "first-time")
```

This more "natural" since it behaves the same as built-in Rails route helpers. For example,
```ruby
event_url(Event.find(1), hello: "world")
# =>  "http://localhost:3000/example?hello=world"
```

Essentially, `apply_form_url` would act like any other normal URL helper. Except that it provides some default query params.

app/views/users/_nav.html.erb
```diff
@@ -78,6 +78,23 @@
     </script>
   <% end %>
 
+  <% if (current_user.only_card_grant_user? && current_user.teenager?) %>
```
garyhtou: Good for checking that they're a teenager.

garyhtou: Instead of checking for only card grant. It's check for no events.

garyhtou: Please move into partial!

app/views/organizer_position/contracts_mailer/notify.html.erb
YodaLightsabr: I don't think they impact it that much, and we have a pretty strong reputation score on SendGrid. We can forgo the params for emails though, if you want.

app/helpers/static_pages_helper.rb
YodaLightsabr: That's reasonable, we can switch it up

app/models/user.rb
```diff
@@ -351,6 +351,10 @@ def sync_with_loops
     UserService::SyncWithLoops.new(user_id: id, new_user:).run
   end
 
+  def only_card_grant_user?
+    creation_method == "card_grant" && card_grants.count >= 1 && events.count == 0
```
garyhtou: It makes sense why we're checking `creation_method`. However, since it was added recently, the old card grant users wouldn't be filtered out by this check.
```suggestion
    card_grants.size >= 1 && events.size == 0
```

We also want to use size instead of count: https://medium.com/unagi/ruby-on-rails-length-vs-size-vs-count-with-examples-a32e17274c79

app/views/users/_card_grant_conversion_callout.html.erb
garyhtou, line 1: fix indent in file

app/views/users/_nav.html.erb
```diff
@@ -78,6 +78,10 @@
...
```
garyhtou: `teenager?` is less resource intensive than `only_card_grant_user?`, so let's run it first.
```suggestion
  <% if (current_user.teenager? && current_user.only_card_grant_user?) %>
```

sampoder (approved):
LGTM!

PR #9516, opened 2025-02-08
[Airbrake] Remove Ruien's error testing controller

garyhtou:
After https://github.com/hackclub/hcb/pull/9514 deploys and we verify it errors are reporting to both AppSignal and Airbrake, we can merge this PR.


rluodev:
@garyhtou I confirmed that errors go to both airbrake and appsignal, this looks good to merge!

PR #9518, opened 2025-02-08
Fix authentication redirection cycle for card grants

manuthecoder:
This pull request fixes a redirection loop when a user's HCB account does not match with their card grant account.

includes a small change to the `show` method in the `GrantsController`. The change adds a `method: :delete` parameter to the `redirect_to` call when handling a `Pundit::NotAuthorizedError`.


garyhtou (dismissed):
Two things here:
1. I think we want to update the CardGrant controller. It's kinda confusing, but the GrantController is apart of an unused grant feature. We want the CardGrant controller instead.
2. Automatically signing out a user because they visited a card grant that they don't have access to feels a bit heavy. Maybe we should show a dedicated page to handle this?

I'm envisioning a page that says:

```
[User badge]

This account does not have access to this grant. Please sign in with the same email you received the invitation at.

[Go Home link] [Sign out button]
```

The layout would be the one without any navs.

app/views/users/logout.html.erb
```diff
@@ -10,7 +11,12 @@
   <% end %>
   <%= render "logins/badge", user: current_user %>
 
-  <h3 class="medium max-width-2 mt-5">You’re currently signed into HCB, would you like to <%= params[:return_to].present? ? "continue" : "head to your dashboard" %>?</h3>
+  <h3 class="medium max-width-2 mt-5">
+    <% if params[:return_to]&.match?(%r{^/card_grants/[^/]+$}) %>
+      This account does not have access to this grant. Please sign in with the same email you received the invitation at.
+    <% else %>
+      You’re currently signed into HCB, would you like to <%= params[:return_to].present? ? "continue" : "head to your dashboard" %>?</h3>
+    <% end %>
```
sampoder: Could we generalise it so that a URL parameter or something can set this copy? Maybe like `?error=unauthorised_card_grant` could render this or something.

app/views/users/logout.html.erb
```diff
@@ -10,7 +10,12 @@
   <% end %>
   <%= render "logins/badge", user: current_user %>
 
-  <h3 class="medium max-width-2 mt-5">You’re currently signed into HCB, would you like to <%= params[:return_to].present? ? "continue" : "head to your dashboard" %>?</h3>
+  <h3 class="medium max-width-2 mt-5">
+    <% if params[:unauthorised_card_grant] %>
```
sampoder: Wouldn't the parameter be called `error`?

manu-gurudath: how did i not catch that 💀

app/controllers/card_grants_controller.rb
```diff
@@ -83,7 +83,7 @@ def show
     render :show, layout: !@frame
 
   rescue Pundit::NotAuthorizedError
-    redirect_to auth_users_path(return_to: card_grant_path(@card_grant)), flash: { info: "Please sign in with the same email you received the invitation at." }
+    redirect_to auth_users_path(email: @card_grant.user.email, return_to: card_grant_path(@card_grant), error: "unauthorised_card_grant")
```
```suggestion
    redirect_to auth_users_path(return_to: card_grant_path(@card_grant), error: "unauthorised_card_grant")
```

sampoder (approved):

PR #9520, opened 2025-02-08
[`production.Dockerfile`] `yarn cache clean`

garyhtou:
```
❯ docker image ls | grep yarn-cache
hcb-no-yarn-cache                   latest                                     779fbe9e6093   10 seconds ago   3.11GB
hcb-yarn-cache                      latest                                     5673b088b69d   6 minutes ago    3.11GB
```

I'm actually not sure if I see a difference. Maybe something about our Dockerfile already has it clear its cache by default?


garyhtou:
@albertchae, i'd love your eyes where if you find some time.

YodaLightsabr:
Is there any downside to including this in the Dockerfile?

garyhtou:
Probably not, i think we can merge

PR #9525, opened 2025-02-10
[Card Grants] Remove email from URL when redirecting to auth page

YodaLightsabr:
## Summary of the problem
https://hackclub.slack.com/archives/CN523HLKW/p1739154902294569

## Describe your changes
Remove `email:` from the `redirect_to auth_users_path` in the `GrantsController`

There may be other implications, and this PR may break the card grant flow. I'll continue to investigate and make sure the flows still work, but security is the #1 priority.


garyhtou (commented):
This controller is for the unused Grants features. Please update the CardGrantsController too!

And organizer position invites

image

PR #9532, opened 2025-02-10
[Org Card] Status filter option to be called "Frozen" + Add "Cancelled"

manuthecoder:
### See #9238

This pull request includes updates to the `card_overview` functionality in the `events_controller` and corresponding view updates to support the new "canceled" card status. The changes ensure that the system can handle and display cards with a "canceled" status, alongside the existing "active" and "inactive" statuses.

### Controller Updates:

* [`app/controllers/events_controller.rb`](diffhunk://#diff-e5ba780d6ceaadee8d430a3b77cd3d401a7f0aeacec1ee78d7aa073f027af8c8L423-R423): Modified the `card_overview` method to include "canceled" as a valid status and updated the status filtering logic to handle "canceled" cards. [[1]](diffhunk://#diff-e5ba780d6ceaadee8d430a3b77cd3d401a7f0aeacec1ee78d7aa073f027af8c8L423-R423) [[2]](diffhunk://#diff-e5ba780d6ceaadee8d430a3b77cd3d401a7f0aeacec1ee78d7aa073f027af8c8R444-R445)

### View Updates:

* [`app/views/events/_filter_cards_menu.html.erb`](diffhunk://#diff-257539c2ec10f46349199b836415696acfe82733b1cf40e5ea9b4bbef8310ae8L27-R28): Updated the card status filter menu to include "canceled" status and changed the display name of "inactive" to "frozen".

### Preview

Screenshot2025021012018PM-3955169613


app/controllers/events_controller.rb
```diff
@@ -416,7 +416,7 @@ def emburse_card_overview
   end
 
   def card_overview
-    @status = %w[active inactive].include?(params[:status]) ? params[:status] : nil
+    @status = %w[active inactive canceled].include?(params[:status]) ? params[:status] : nil
```
garyhtou: If we can change this to frozen to eliminate the need to gsub inactive to frozen elsewhere

```suggestion
    @status = %w[active frozen canceled].include?(params[:status]) ? params[:status] : nil
```

app/controllers/events_controller.rb
```diff
@@ -437,6 +437,8 @@ def card_overview
                          all_stripe_cards.active
                        when "inactive"
```
garyhtou: If we can rename it @status array to frozen, then this should say frozen here.

sampoder (approved):

sampoder:
LGTM!

PR #9539, opened 2025-02-12
[Dev Docs] Add research on Coolify

garyhtou:
wip

todo:

- [ ] explain _**why**_ this setup.


northeastprince:
I wonder if we could setup some sort of config management instead like Terraform/Chef so it's self-documenting instead

northeastprince:
And is there a specific reason in particular we're using Docker Swarm as opposed to just Coolify's multi-server support?

garyhtou:
> And is there a specific reason in particular we're using Docker Swarm as opposed to just Coolify's multi-server support?

Docker Swarm is more reliable and flexible, and I'm familiar with it. Unfortunately, Coolify's support for Docker Swarm is **very** experimental. I encountered a lot of bugs and ultimately resorted to Coolify's multi-server with a Hetzner LB.

> I wonder if we could setup some sort of config management instead like Terraform/Chef so it's self-documenting instead

To be explored in the future. TBH, I'm considering something like Rancher/Portainer with Ansible/cloud-init for the initial sever configuration

northeastprince:
If (we're considering other options than Coolify and) we're going with more [declarative deployment tooling](https://dev.37signals.com/leaning-imperative) instead of Kamal, I'd highly recommend taking a look at HashiCorp Nomad - it's basically 95% of Kubernetes with 5% of the complexity.

PR #9552, opened 2025-02-12
Allow for custom fees via the Fee model

sampoder:
WIP.


app/models/event.rb
```diff
@@ -283,6 +283,7 @@ class Event < ApplicationRecord
   scope :dormant, -> { where.not(id: Event.engaged) }
 
   has_many :fees, through: :canonical_event_mappings
+  # must transition to has_many :fees
```
garyhtou: good catch

app/models/fee.rb
```diff
@@ -7,14 +7,17 @@
 #  id                         :bigint           not null, primary key
 #  amount_cents_as_decimal    :decimal(, )
 #  event_sponsorship_fee      :decimal(, )
+#  memo                       :string
```
garyhtou: Do we also want to add a `payment_for`? I think it's not really necessary, but maybe we should stay consistent with other models

app/models/event.rb
sampoder: this should be done after I backfill the column

app/models/fee.rb
```diff
@@ -23,13 +26,15 @@
...
 
-  belongs_to :canonical_event_mapping
+  belongs_to :canonical_event_mapping, optional: true
   has_one :canonical_transaction, through: :canonical_event_mapping
 
   validates :reason, presence: true
   validates :amount_cents_as_decimal, numericality: { greater_than_or_equal_to: 0 }
   validates :event_sponsorship_fee, numericality: { greater_than_or_equal_to: 0 }
 
+  validates :memo, presence: true, unless: -> { canonical_event_mapping.present? }
```
garyhtou: validate nil if cem.present?

OR, we can choose to fill in `memo` for all Fees. May be a good idea tbh — it'll make rendering memos easier

app/models/fee.rb
```diff
@@ -23,13 +26,15 @@
 class Fee < ApplicationRecord
   has_paper_trail
 
-  belongs_to :canonical_event_mapping
+  belongs_to :canonical_event_mapping, optional: true
```
garyhtou: after backfilling, add `belongs_to :event` which will also validate that it must be present.

app/models/fee.rb
```diff
@@ -58,4 +63,16 @@ def waived?
     reason != "revenue"
   end
 
+  validate do
+    if canonical_event_mapping.present? && (canonical_event_mapping.event_id != event_id)
+      errors.add(:event_id, "must match the CanonicalTransaction's event")
+    end
+  end
+
+  before_validate do
```
garyhtou:
```suggestion
  before_validation do
```

app/models/fee.rb
```diff
@@ -7,14 +7,17 @@
 #  id                         :bigint           not null, primary key
 #  amount_cents_as_decimal    :decimal(, )
 #  event_sponsorship_fee      :decimal(, )
+#  memo                       :string
 #  reason                     :text
 #  created_at                 :datetime         not null
 #  updated_at                 :datetime         not null
-#  canonical_event_mapping_id :bigint           not null
+#  canonical_event_mapping_id :bigint
+#  event_id                   :bigint
```
garyhtou: after backfill, not null it

db/migrate/20250217050700_drop_fee_cemfk.rb
garyhtou, line 1: Can you rename this class and file to `DropFeeCemNotNull`. Dropping FK implies that we're getting rid of the whole column.

garyhtou (commented):
lgtm

db/migrate/20250217050700_drop_fee_cemfk.rb
sampoder: Yes! Sorry this was when this migration did something very different

PR #9554, opened 2025-02-13
Add temp endpoint for getting client IP

garyhtou:
Testing whether IPs are passed correctly after LB proxy on Hetzner, and how it compares to Heroku


garyhtou:
We're using `request.ip` in some places and `request.remote_ip` in others. I think we may need to switch to only use `remote_ip`. However, this test will verify that.

garyhtou:
localhost
```json
{
  "remote_ip": "127.0.0.1",
  "ip": "127.0.0.1",
  "forwarded_for": null,
  "forwarded": null,
  "forwarded_proto": null,
  "forwarded_port": null,
  "forwarded_host": null,
  "forwarded_server": null
}
```

coolify
```json
{
  "remote_ip": "<REDACTED, my external ip>",
  "ip": "<REDACTED, my external ip>",
  "forwarded_for": "<REDACTED, my external ip>",
  "forwarded": null,
  "forwarded_proto": "https",
  "forwarded_port": "443",
  "forwarded_host": null,
  "forwarded_server": null
}
```

heroku
```json
{
  "remote_ip": "<REDACTED, my external ip>",
  "ip": "<REDACTED, my external ip>",
  "forwarded_for": "<REDACTED, my external ip>",
  "forwarded": null,
  "forwarded_proto": "https",
  "forwarded_port": "443",
  "forwarded_host": null,
  "forwarded_server": null
}
```

so it looks like it all functions as-is without configuration. likely because the Heztner LB is hitting the servers via the local private network (10.0.0.0/8) so it's already trusted by Rails. Although, I did expect to see the LB's ip in `forwared_for` but it's not there. Maybe Rails already took it out?

PR #9569, opened 2025-02-13
[Sidekiq] Set concurrency WRT RAILS_MAX_THREADS

garyhtou:
easier to scale


config/initializers/sidekiq.rb
```diff
@@ -5,9 +5,41 @@
...
+  # once they finish can it pick up the 5 `critical` jobs. After after
+  # finishing the `critical` jobs, it will then pick up the remaining 5 `low`
+  # jobs. This means `critical` jobs can be blocked by long-running `low` jobs.
+  #
+  # Giving these long-running jobs it's own capsule guarantees the other queues
+  # (like `critical`) are not blocked by long-running jobs.
   config.capsule("throttled") do |capsule|
-    capsule.queues = %w[wrapped]
-    capsule.concurrency = 3
+    capsule.queues = %w[metrics]
```
garyhtou: It's fine to do the rename like this without keeping/draining the old queue name (`wrapped`) since there's nothing queue to `wrapped` and we don't have anything that would queue to `wrapped`.

PR #9571, opened 2025-02-14
Remove index on CEM from `Fee`

sampoder:
Blocking https://github.com/hackclub/hcb/pull/9552


garyhtou:
Are you getting an error with this index in place?

sampoder:
Yes - not null failure

sampoder:
(Rspec has an example of this)

garyhtou:
Should we remove the `NOT NULL` requirement on `canonical_event_mapping_id` rather than removing the index?

PR #7071, opened 2024-07-09
Implement 1099 Payroll

sampoder:
1099s! They are my favourite number things.

PR #9572, opened 2025-02-14
Revert "Implement 1099 Payroll"

garyhtou:
Reverts hackclub/hcb#7071

Migration error causing downtime on Coolify. Deploys still go through on Coolify even if entrypoint errors out


garyhtou:
Dispute Heroku having ran this migration (the shared DB is migrated), the Coolify rails server is erroring

garyhtou:
Found the issue. Coolify's DB user doesn't have access to change the schema

garyhtou:
reverting until we get coolify under control

PR #9577, opened 2025-02-15
[Org Cards] Fix mock data in table view

garyhtou:
## Summary of the problem

We attempted to render a popover for mock data stripe cards. and it doesn't go well (500). Bug existed only for Org card's table view.

## Describe your changes

Don't render popovers for demo mode orgs (on the cards page).

and refactor for readability


sampoder (approved):
LGTM!

PR #9581, opened 2025-02-16
Disable domains for Scrapyard orgs

sampoder:
Closes https://github.com/hackclub/hcb/issues/9575 - should be reverted after Scrapyard


app/views/events/promotions.html.erb
```diff
@@ -105,7 +105,7 @@
   <div
     class="card__banner card__banner--top flex justify-between items-center promo-domains">
     <h3 class="h1 p2 my0 color-black">Domains</h3>
-      <% if !@event.g_suites.none? || !@perks_available || !(@event.organized_by_teenagers?) %>
+      <% if (!@event.g_suites.none? || !@perks_available || !(@event.organized_by_teenagers?)) && event_tags.where(name: "Scrapyard").none? %>
```
```suggestion
      <% if (!@event.g_suites.none? || !@perks_available || !(@event.organized_by_teenagers?)) && @event.event_tags.where(name: "Scrapyard").none? %>
```

PR #9583, opened 2025-02-16
Create `StripeWebhookJob`

sampoder:
Closes https://github.com/hackclub/hcb/issues/9570


config/schedule.yml
```diff
@@ -175,3 +175,7 @@ user_seen_at_history_snapshot_job:
 column_sweep_job:
   cron: "0 0 * * MON" # run first day of each week
   class: "Column::SweepJob"
+
+stripe_webhook_job:
+  cron: "0,5 * * * *" # run every 5 minutes
```
garyhtou: I think this should be `"*/5 * * * *"`

PR #9584, opened 2025-02-16
Upgrade `actions/cache@v4`

sampoder:
Closes https://github.com/hackclub/hcb/issues/9549 - hopefully it's this simple.

PR #9594, opened 2025-02-18
[User] Allow locking admin User accounts

garyhtou:
## Summary of the problem

Couldn't lock out an admin account without using the console.

## Describe your changes

Allow locking out admins, but not super admins via the dashboard.


garyhtou:
CI is failing because of https://github.com/hackclub/hcb/pull/9584

PR #9595, opened 2025-02-18
[Lint] Prettier

garyhtou:
Running `bin/lint` gave these results. Changes likely due to https://github.com/hackclub/hcb/pull/9545


garyhtou:
hm prettier failed

sampoder:
Closing because `main` appears to be inline with our current Prettier version: https://github.com/hackclub/hcb/pull/9608

PR #9619, opened 2025-02-20
Fix login codes

manuthecoder:
This pull request fixes the login code bug and TOTP input process by removing the custom JavaScript controller and updating the corresponding HTML views to use a single input field instead of multiple number fields. This is intended to address #9616 

Changes to simplify the login code input:

* [`app/javascript/controllers/login_code_controller.js`](diffhunk://#diff-464dde7cbdd98507805c47e73c4fe18dee8739896b4984738d59b3616904e677L1-L94): Removed the entire custom JavaScript controller that handled the login code input logic.
* [`app/views/logins/login_code.html.erb`](diffhunk://#diff-e728e1322c534c6f9af44af548f2287d033297fce0080bc43c74075dc191f71fL21-R21): Replaced the multiple number fields for login code input with a single text field.

Changes to simplify the TOTP input:

* [`app/views/logins/totp.html.erb`](diffhunk://#diff-e1644be2bc2c0fed3d0d53f7363f242bf2bc0835e42eef1dbd306c3326d100fcL18-R19): Replaced the multiple number fields for TOTP input with a single text field.


garyhtou (approved):
ty

PR #9622, opened 2025-02-20
Force profile pictures to be variable

sampoder:
Fixes the cause of this issue: https://github.com/hackclub/hcb/issues/9600 

Prod data needs to be fixed


app/models/user.rb
```diff
@@ -388,6 +388,7 @@ def slug_candidates
   def profile_picture_format
     return unless profile_picture.attached?
     return if profile_picture.blob.content_type.start_with? "image/"
+    return if profile_picture.blob.variable?
```
```suggestion
    return if profile_picture.blob.content_type.start_with?("image/") && profile_picture.blob.variable?
```

PR #9642, opened 2025-02-22
Hide sign in button on donation pages

sampoder:
Feels unnatural to me - plus it's causing a UI frenzy. I believe this closes https://github.com/hackclub/hcb/issues/8912.


garyhtou:
Good solution

PR #9643, opened 2025-02-22
Add preview for docx files

Ved235:
Before to close #7733 I was using an approach based on Libre office, but it didn't work with Heroku. Instead, now using Google Drive Viewer a preview can be generated for docx files.  Sample images below:
image
image

_Note: I couldn't actually try this right now because the files have to be hosted on a public URL and in Codespace environment localhost is used_


garyhtou:
Superseded by https://github.com/hackclub/hcb/pull/9856 due to git rewrite.

PR #9700, opened 2025-02-26
[Event] Create onboarding mailer

garyhtou:
## Summary of the problem

closes https://github.com/hackclub/hcb/issues/9683

@leowilkin said he wouldn't be able to get around to this due to exams and asked that I wrap this up for him.

## Describe your changes

wip. untested and only implement the first email.

- [ ] CDN gifs — cdn's broken rn
- [ ] test scheduling/jobs
- [ ] double check 1 hour is reasonable based on ops workflow


app/jobs/event/onboarding/getting_started_email_job.rb
```diff
@@ -0,0 +1,22 @@
...
+
+class Event
+  module Onboarding
+    class GettingStartedEmailJob < ApplicationJob
+      queue_as :default
+
+      def perform(event)
+        return if event.demo_mode?
+
+        managers = event.organizer_positions.where(role: :manager).includes(:user).map(&:user)
```
garyhtou, line 11: Also include organizer positions where role is manager, so that if a user hasn't accepted their initial invite, they still get the email.

garyhtou:
FYI, we're considering using Loops for these types of welcome/marketing-ish emails

leowilkin:
I thought engr had a good few reasons why we weren't going to use loops, but that's ok! As long as it actually gets created I'm happy :)
________________________________

polypixeldev:
Now exists in the new application flow!

PR #9704, opened 2025-02-27
Create `Event#pending_reimbursements_balance` and use it in `balance_available_v2_cents`

sampoder:
Screenshot 2025-02-26 at 7 04 59 PM

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -505,13 +505,17 @@ class Event < ApplicationRecord
     cpt.sum(:amount_cents)
   end
 
+  def pending_reimbursements_balance
+    Reimbursement::Expense.where(report: reimbursement_reports.reimbursement_requested).approved.sum(:amount_cents)
+  end
+
   def balance_available_v2_cents
     @balance_available_v2_cents ||= begin
       fee_balance = can_front_balance? ? fronted_fee_balance_v2_cents : fee_balance_v2_cents
       if fee_balance.positive?
-        balance_v2_cents - fee_balance
+        balance_v2_cents - fee_balance - pending_reimbursements_balance
       else # `fee_balance` is negative, indicating a fee credit
-        balance_v2_cents
+        balance_v2_cents - pending_reimbursements_balance
       end
     end
   end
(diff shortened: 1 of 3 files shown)


garyhtou (commented):
This solution likely would work, however, I'm afraid we're straying further into an unmaintainable transaction engine. Please know that this comment isn't directed towards this code or your work.

We have pending transactions intended to handle use cases like these—temporarily modifying the balance until a settled transaction arrives. However, from previous attempts, we learned that managing pending transactions for Reimbursements is cumbersome. Because the `CanonicalPendingTransaction`s are associated to `Reimbursement::ExpensePayout`s, it requires us to create immutable `Reimbursement::ExpensePayout`s BEFORE the `Reimbursement::Report`'s amount is locked in.

One possible alternative solution I have not explored is refactoring `CanonicalPendingTransaction` to be associated with `Reimbursement::Expense` rather than `Reimbursement::ExpensePayout`. This would allow us to easily update the amount of `CanonicalPendingTransaction`s when the expense/report amount changes. It would also allow us to keep the creation of `Reimbursement::ExpensePayout` as-is (created AFTER the `Reimbursement::Report` has been fully approved and we know the final expense amounts).

Can you look further into this alternative solution? Or let me know if you have other ideas!

sampoder:
I do agree that I'm not super comfortable straying away from CPTs, my logic here was that this is what we've done before when an amount varies significantly before settling (fees) and that this PR would be fairly easy to revert / rip out. 

> One possible alternative solution I have not explored is refactoring CanonicalPendingTransaction to be associated with Reimbursement::Expense rather than Reimbursement::ExpensePayout. This would allow us to easily update the amount of CanonicalPendingTransactions when the expense/report amount changes. It would also allow us to keep the creation of Reimbursement::ExpensePayout as-is (created AFTER the Reimbursement::Report has been fully approved and we know the final expense amounts).

This is do-able but it could be weird. For example we could find ourselves declining and then undeclining CPTs. It would also create a funny situation with HCB codes because the current system's HCB codes contain an ExpensePayout's ID. We could create a HCB code purely used for pending reimbursements, but that again feels messy?

PR #9716, opened 2025-02-27
Bump Rails config defaults to 7.2

KiKoS0:
Closes: https://github.com/hackclub/hcb/issues/8471

## Summary of the problem
Bumping Rails config defaults to 7.2 following this finished migration https://github.com/hackclub/hcb/pull/9382.


garyhtou (approved):
ty!

PR #9738, opened 2025-03-04
Remove increase

manuthecoder:
This PR removes Increase functionality across the platform. Instead of outright deleting all references to "Increase" in the codebase, this update specifically removes the ability to create, modify, or delete related data. While frontend access to view this data has also been removed, the underlying records remain stored in case they are needed in the future.

cc @sampoder

sorry for getting to this so late I was a little caught up last week


app/controllers/increase_controller.rb
sampoder, line 1: We can probably remove this file?

app/jobs/increase_check_job/remind_undeposited_recipient.rb
sampoder, line 1: We still use this, checks sent using Column are still called `IncreaseCheck`. Yes, I know.

app/mailers/increase_check_mailer.rb
sampoder, line 1: We still use this, checks sent using Column are still called `IncreaseCheck`.

app/models/increase_check.rb
sampoder, line 1: We still use this, checks sent using Column are still called `IncreaseCheck`.

app/policies/increase_check_policy.rb
sampoder, line 1: We still use this, checks sent using Column are still called `IncreaseCheck`.

app/controllers/increase_checks_controller.rb
sampoder, line 1: We still use this, checks sent using Column are still called `IncreaseCheck`.

app/views/admin/increase_checks.html.erb
sampoder, line 1: We still use this, checks sent using Column are still called `IncreaseCheck`.

app/views/admin/increase_check_process.html.erb
sampoder, line 1: We still use this, checks sent using Column are still called `IncreaseCheck`.

sampoder (changes requested):
Transaction engine stuff looks good - though there may be more to do. We need to keep all of the `IncreaseCheck` stuff

sampoder:
Other things I think we can drop:

https://github.com/hackclub/hcb/blob/c1886a4cf73d8ad4a31e65688082f1e56bf8c532/config/initializers/increase.rb#L4
https://github.com/hackclub/hcb/blob/c1886a4cf73d8ad4a31e65688082f1e56bf8c532/Gemfile.lock#L409
https://github.com/hackclub/hcb/blob/c1886a4cf73d8ad4a31e65688082f1e56bf8c532/app/services/increase_service.rb#L4

manu-gurudath:
Done! Let me know if any further changes are needed. (Some of the files might touch parts of the platform which I might not be aware of so heads up)

config/routes.rb
```diff
@@ -367,13 +367,6 @@
...
   resources :checks, only: [:show]
 
-  resources :increase_checks, only: [] do
-    member do
-      post "approve"
-      post "reject"
-    end
-  end
-
   resources :paypal_transfers, only: [] do
```
sampoder: These need to be kept

config/routes.rb
```diff
@@ -714,7 +706,7 @@
     get "activation_flow", to: "events#activation_flow", as: :activation_flow
     post "activate", to: "events#activate", as: :activate
     resources :disbursements, only: [:new, :create]
-    resources :increase_checks, only: [:new, :create], path: "checks"
```
sampoder: These also need to stay

app/services/increase_service.rb
sampoder, line 1: Let's drop this entire file

sampoder (approved):
LGTM - thank you Manu!

PR #9739, opened 2025-03-05
Allow organizations to create sub-organizations

sampoder:
Allows pre-approved events to create sub-organizations. This version of sub-organizations only supports creating new FSOs by adding a manager / contract signee. The plan can be configured by admins in event settings.

image

_Sub-organizations page_

image

_Modal to create a new sub-organization_


sampoder:
Contracts? Do you inherit from parent? 

Delete create project button

Some sort of linking?

Dropdown for parent on new organisation page / prefill maybe

app/controllers/events_controller.rb
```diff
@@ -286,6 +286,25 @@ def team
...
+      end
+
+      ancestor_roles = {}
+      ancestor_ops.each do |op|
+        number = OrganizerPosition.roles[op.role]
+        ancestor_roles[op.user] = number if (ancestor_roles[op.user] || 0) < number && (direct_roles[op.user] || 0) < number
+      end
+
+      @indirect_access = ancestor_roles.sort_by { |_, role| role }.reverse!.to_h.transform_values { |number| OrganizerPosition.roles.key(number) }
+    end
```
```suggestion
    if @event.parent
      ancestor_ops = @event.ancestor_organizer_positions.includes(:user)
      direct_ops = @event.organizer_positions.includes(:user)

      # this hash contains every organizer's role as a number
      # higher numbers = higher access level
      direct_roles = {}
      direct_ops.each do |op|
        number = OrganizerPosition.roles[op.role]
        direct_roles[op.user] = number if (direct_roles[op.user] || 0) < number
      end

      # if you have a higher level of access in an ancestor event
      # you will inherit that level of access in this event
      ancestor_roles = {}
      ancestor_ops.each do |op|
        number = OrganizerPosition.roles[op.role]
        ancestor_roles[op.user] = number if (ancestor_roles[op.user] || 0) < number && (direct_roles[op.user] || 0) < number
      end

      @indirect_access = ancestor_roles.sort_by { |_, role| role }.reverse!.to_h.transform_values { |number| OrganizerPosition.roles.key(number) }
    end
```

feedback: this code is really difficult to read, I've tried adding comments but I can't understand the intention behind the last line.

are we making the keys of the hash the user object?

why not create a method on `User` which is `access_level_for(event)` that contains this logic?

app/models/event.rb
```diff
@@ -234,6 +279,15 @@ class Event < ApplicationRecord
 
   has_many :organizer_position_invites, dependent: :destroy
   has_many :organizer_positions, dependent: :destroy
+
+  def ancestor_organizer_positions
+    OrganizerPosition.where(event_id: ancestor_ids)
+  end
+
+  def access_organizer_positions
```
```suggestion
  def organizer_positions_with_access
```
imo this is more like English

app/models/concerns/organizer_position/has_role.rb
```diff
@@ -6,7 +6,18 @@ module HasRole
...
-      enum :role, { reader: 5, member: 25, manager: 100 }
+
+      enum :role, {
+        reader: 5,
+        member: 25,
+        manager: 100
+      }
+
+      roles.each do |role_name, role_value|
+        scope role_name, -> { where(role: role_name) }
```
sampoder: Screenshot 2025-07-23 at 12 19 32 AM

I believe we get this for free with `enum`?

app/models/event.rb
```diff
@@ -118,6 +121,48 @@ class Event < ApplicationRecord
       .where("flipper_gates.feature_key = ? AND flipper_gates.key = ?", flag, "actors")
   }
 
+  def ancestor_ids
+    [id] + Event.connection.execute(<<-SQL).map { |row| row["id"] }
```
sampoder, line 124: is an event an ancestor of itself?

app/models/event.rb
```diff
@@ -118,6 +121,48 @@ class Event < ApplicationRecord
...
+        FROM events e
+        INNER JOIN child_events ce ON e.parent_id = ce.id
+      )
+      SELECT id FROM child_events;
+    SQL
+  end
+
+  def ancestors
+    Event.where(id: ancestor_ids)
+  end
```
sampoder, line 155: if there's any way we could make this a scope that would be much appreciated: https://x.com/nateberkopec/status/1907240790725890074.

app/views/events/settings/_admin.html.erb
```diff
@@ -109,6 +109,38 @@
...
+          </li>
+        <% end %>
+      </ul>
+
+      <div class="field">
+        <%= form.label :subevent_plan, "Sub-organization default plan" %>
+        <%= form.select :subevent_plan, { selected: @event.subevent_plan }, { disabled: } do %>
+          <optgroup label="🚫 Disabled">
+            <option value="">🚫 Turn off sub-organizations</option>
+          </optgroup>
```
sampoder: It's not the end of the world but I don't feel great about this versus adding a boolean to `Event::Configuration`. also both of these fields should exist in `Event::Configuration`.

app/models/event.rb
```diff
@@ -772,6 +836,10 @@ def active_teenagers
     organizer_positions.joins(:user).count { |op| op.user.teenager? && op.user.active? }
   end
 
+  def subevents_enabled?
+    subevent_plan.present? && !subevent_plan.empty?
+  end
+
```
sampoder: see: https://github.com/hackclub/hcb/pull/9739/files#r2224312897

app/models/event.rb
YodaLightsabr: This is how Ruby handles ancestors:

image

app/controllers/events_controller.rb
```diff
@@ -286,6 +286,20 @@ def team
...
+    if @event.parent
+      ops = @event.ancestor_organizer_positions.includes(:user)
+      users = ops.map(&:user).uniq
+
+      access_levels = users.map do |user|
+        access_level = user.access_level_for(@event, ops)
+        next nil if access_level[:access_level] == :direct
+
+        [user, access_level[:role]]
+      end.compact!.sort_by { |_, role| role }.to_h
```
garyhtou: Try `filter_map`

app/controllers/events_controller.rb
```diff
@@ -766,6 +780,42 @@ def employees
     @employees = @employees.search(params[:q]) if params[:q].present?
   end
 
+  def projects
```
garyhtou: Rename projects to sub-organizations

app/models/concerns/organizer_position/has_role.rb
garyhtou: yup!

app/models/event.rb
```diff
@@ -118,6 +120,48 @@ class Event < ApplicationRecord
...
+    Event.where(id: ancestor_ids)
+  end
+
+  def descendant_balance
+    descendants.to_a.sum(&:balance_available_v2_cents)
+  end
+
+  belongs_to :parent, class_name: "Event", optional: true
+  has_many :subevents, class_name: "Event", foreign_key: "parent_id"
+  alias_method :descendants, :subevents
```
garyhtou: Let's try not to alias associations— there are some weird edge cases. Let's use the actual association name or rename the association

app/models/event.rb
```diff
@@ -369,7 +418,7 @@ class Event < ApplicationRecord
   end
 
   before_validation do
-    build_plan(type: Event::Plan::Standard) if plan.nil?
+    build_plan(type: parent&.plan&.class || Event::Plan::Standard) if plan.nil?
```
garyhtou: Shouldn't we also check parent's `subevent_plan`?

app/models/event.rb
```diff
@@ -735,6 +784,16 @@ def omit_stats?
...
 
+  validate do
+    if id && id == parent_id
+      errors.add(:parent, "can't be self-referential.")
+    end
+  end
+
+  def plan
+    parent&.plan || super
+  end
```
garyhtou: All events should have it's own plan record, I don't think we want to subsitute in the parent's here

app/models/event.rb
```diff
@@ -772,6 +831,10 @@ def active_teenagers
     organizer_positions.joins(:user).count { |op| op.user.teenager? && op.user.active? }
   end
 
+  def subevents_enabled?
+    config.subevent_plan.present? && !config.subevent_plan.empty?
+  end
```
garyhtou: I think the `present?` should also handle `!empty?`, although double check

app/models/organizer_position.rb
```diff
@@ -55,11 +55,20 @@ def tourable_options
...
-    return true if user&.admin?
+    if role.to_s == "reader"
+      return event.ancestor_organizer_positions.reader_access.where(user:).exists?
+    end
 
-    current = find_by(user:, event:)&.role
-    current && roles[current] >= roles[role]
+    if role.to_s == "member"
+      # Only check direct organizer positions, unless the user is a manager of an ancestor
+      return event.organizer_positions.member_access.where(user:).exists? || event.ancestor_organizer_positions.member_access.where(user:).exists?
```
garyhtou: please double check if it's a typo

```suggestion
      return event.organizer_positions.member_access.where(user:).exists? || event.ancestor_organizer_positions.reader_access.where(user:).exists?
```

app/models/user.rb
```diff
@@ -436,6 +436,20 @@ def disable_backup_codes!
...
+    access_level = nil
+
+    organizer_positions.select { |op| op.user == self }.each do |op|
+      if role.nil? || OrganizerPosition.roles[op.role] > OrganizerPosition.roles[role]
+        role = op.role
+        access_level = op.event == event ? :direct : :indirect
+      end
+    end
+
+    { role:, access_level: }
```
garyhtou, line 452: Maybe this method should just return `nil` if no OPs are found for the user

app/views/events/_nav.html.erb
```diff
@@ -122,6 +122,11 @@
       tooltip: "View legal documents and financial statements",
       icon: "docs",
       selected: local_assigns[:selected] == :documentation if policy(@event).documentation? %>
+    <%= dock_item "Projects",
```
garyhtou: rename to suborganizations

app/views/events/_title.html.erb
```diff
@@ -1,13 +1,19 @@
-<div data-controller="menu" class="pt2">
-  <%= link_to @event, class: "#{"pointer-events-none" unless signed_in?} text-decoration-none gap-2 text-inherit leading-9 event-title -m-2 my-2 px-2 rounded-lg transition-colors flex items-center tracking-tight hover:text-primary", style: "font-size:2em; font-weight: 600; word-break: break-word;", data: ({ action: "click->menu#toggle click@document->menu#close keydown@document->menu#keydown", menu_target: "toggle" } if show_org_switcher?) do %>
+<div class="muted">
+  <% @event.ancestors[1..].each do |ancestor| %>
```
garyhtou, line 2: should we have a `parents` (plural) method?

app/models/event.rb
```diff
@@ -118,6 +120,48 @@ class Event < ApplicationRecord
...
+      )
+      SELECT id FROM child_events;
+    SQL
+  end
+
+  def ancestors
+    Event.where(id: ancestor_ids)
+  end
+
+  def descendant_balance
```
```suggestion
  def descendant_total_balance_cents
```

app/views/events/settings/_admin.html.erb
```diff
@@ -109,6 +109,40 @@
     <% end %>
   <% end %>
 
+  <% admin_tool do %>
+    <%= form_with(model: event, local: true, html: { "x-data" => "{ship_type: '#{event.stripe_card_shipping_type}'}" }) do |form| %>
```
garyhtou: lol, i don't think we're shipping here

app/views/events/settings/_admin.html.erb
```diff
@@ -109,6 +109,40 @@
...
+        <div class="field">
+          <%= config.label :subevent_plan, "Sub-organization default plan" %>
+          <%= config.select :subevent_plan, { selected: @event.subevent_plan }, { disabled: } do %>
+            <optgroup label="🚫 Disabled">
+              <option value="">🚫 Turn off sub-organizations</option>
+            </optgroup>
+            <optgroup label="✅ Enabled">
+              <%= options_for_select(Event::Plan.available_plans.map { |p| [p.new.label, p.name] }, @event.config.subevent_plan) %>
+            </optgroup>
+          <% end %>
```
garyhtou: this is nice!

app/views/events/settings/_admin.html.erb
```diff
@@ -109,6 +109,40 @@
...
+        </div>
```
garyhtou, line 136: Turning off subevents will NOT delete existing subevents

app/views/events/team.html.erb
```diff
@@ -32,6 +32,24 @@
   <% end %>
 </h1>
 
+<% if @event.parent.present? %>
+  <%= render "callout", title: "The team behind #{@event.parent.name} also has access to #{@event.name}", type: "info", icon: "leader" do %>
+    <p>Because <%= @event.name %> is a subsidiary of <%= @event.parent.name %>, all team members of <%= @event.parent.name %> can access and view this organization.</p>
```
garyhtou:
```suggestion
    <p>Because <%= @event.name %> is a sub-organization of <%= @event.parent.name %>, all team members of <%= @event.parent.name %> can access and view this organization.</p>
```

db/migrate/20250304082509_add_parent_event_id_to_event.rb
```diff
@@ -0,0 +1,7 @@
+class AddParentEventIdToEvent < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :events, :parent, index: {algorithm: :concurrently}
+  end
+end
```
garyhtou:
```suggestion
end

```

db/migrate/20250721195537_add_subevent_plan_to_event_configuration.rb
```diff
@@ -0,0 +1,5 @@
+class AddSubeventPlanToEventConfiguration < ActiveRecord::Migration[7.2]
+  def change
+    add_column :event_configurations, :subevent_plan, :string
```
garyhtou, line 3: good!

db/schema.rb
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
```
garyhtou: 🥶🧵💬

app/policies/event_policy.rb
```diff
@@ -152,6 +152,14 @@ def employees?
     auditor_or_reader?
   end
 
+  def projects?
+    admin_or_reader? && (record.subevents_enabled? || record.descendants.any?)
```
garyhtou: good for checking `record.descendants.any?`

app/models/organizer_position.rb
YodaLightsabr: This was actually a typo the other way around—we only want parent _manager_ access to trickle down. Good catch though, thanks!

app/models/event.rb
sampoder: Interesting TIL - if we can, I'd vote to avoid replicating this because I think it's confusing or doing what Gary suggested: https://github.com/hackclub/hcb/pull/9739/files#r2226423504

app/controllers/events_controller.rb
```diff
@@ -766,6 +780,42 @@ def employees
...
+
+    if params[:email].blank?
+      flash[:error] = "Organizer email is required"
+      return redirect_back fallback_location: event_sub_organizations_path(@event)
+    end
+
+    subevent = ::EventService::Create.new(
+      name: params[:name],
+      emails: [params[:email]],
+      is_signee: true,
```
sampoder, line 814: are we going to require contracts for these? how will we collect parent's emails?

app/models/event.rb
sampoder:
```ruby
scope :ancestors_of, -> (event){ 
      find_by_sql(
        ["WITH RECURSIVE parent_events AS (
          SELECT id, parent_id
          FROM events
          WHERE id = ?
          UNION ALL
          SELECT e.id, e.parent_id
          FROM events e
          INNER JOIN parent_events pe ON e.id = pe.parent_id
        )
        SELECT DISTINCT events.* from events WHERE id IN (SELECT id FROM parent_events)",
        event.id
      ]
    )
  }
```

app/views/events/_title.html.erb
YodaLightsabr: We only use it here, I don't think we need to

app/views/events/settings/_admin.html.erb
YodaLightsabr: oopsies

app/controllers/events_controller.rb
YodaLightsabr: could you remind me how the new contract flow works? do we collect parent emails at event creation time?

app/models/event.rb
YodaLightsabr: I originally excluded the current event from `event.ancestors` but added it back because of Gary's suggestion. I guess adding `parents` could be a good middle-ground, but it might be confusing having 2 very similar methods that include/exclude the current event.

app/models/event.rb
YodaLightsabr: If we make this a scope, would it force us to do `Event.ancestors_of(event)` every time? Or would we be able to alias it to `event.ancestors` without issue now.

app/models/event.rb
```diff
@@ -369,7 +417,7 @@ class Event < ApplicationRecord
   end
 
   before_validation do
-    build_plan(type: Event::Plan::Standard) if plan.nil?
+    build_plan(type: parent&.subevent_plan&.class || parent&.plan&.class || Event::Plan::Standard) if plan.nil?
```
garyhtou, line 420: Brakeman may complain about `constantize`. maybe we don't need `constantize`.
```suggestion
    build_plan(type: parent&.config&.subevent_plan&.constantize || parent&.plan&.class || Event::Plan::Standard) if plan.nil?
```

garyhtou (approved):
There are things to clean up, but i'm going to merge this

PR #9759, opened 2025-03-08
Update and rename infrafi-sweeps.md to intrafi-sweeps.md

sampoder:
Also updates the name in the document


garyhtou (commented):
The number of times i've called it InfraFi is crazy!

PR #9765, opened 2025-03-10
Attempt to speed up reimbursements

sampoder:
Because it's timing out. And I love time outs.


app/views/reimbursement/expenses/_receipts.html.erb
```diff
@@ -25,7 +25,7 @@
 
       <% if expense.receipts.any? %>
         <div class="flex flex-col p1" id="link_receipt__form">
-          <% column_amount = [((expense.receipts.count - 1) / 2.to_f).ceil, expense.receipts.count == 2 ? 2 : 1].max %>
+          <% column_amount = [((expense.receipts.length - 1) / 2.to_f).ceil, expense.receipts.length == 2 ? 2 : 1].max %>
```
garyhtou: Use `size` instead of `length`. `size` will choose between length and count depending on if the association is in memory.

PR #9766, opened 2025-03-10
[Org Home] Fix running balance graph

YodaLightsabr:
## Summary of the problem
Right now, the X-axis scale on the running balance graph is "whenever we have another data point", which makes it pretty difficult to understand your financial situation over time.

## Describe your changes
Now it's fixed!

image

Sparkline weirdly has some issues when you try and provide repeat values with more than a few hundred data points, so I've also updated it to cut it off at 365 days (1 year).


app/javascript/controllers/balance_graph_controller.js
```diff
@@ -43,6 +51,21 @@ export default class extends Controller {
               ? this.availableValue
               : parseFloat(balanceByDate[date])
 
+          const mostRecentDate = balances[balances.length - 1]?.date
+          const range = getDates(new Date(date), new Date(mostRecentDate));
+          console.log({ mostRecentDate: new Date(mostRecentDate), date: new Date(date) })
+          console.log({ range })
```
sampoder:
```suggestion
```

sampoder (approved):
Thank you!

app/javascript/controllers/balance_graph_controller.js
YodaLightsabr: Oh oops

PR #9774, opened 2025-03-11
Revert "Load Doppler tokens using the Credentials module"

sampoder:
Reverts hackclub/hcb#9771


sampoder:
This worked in staging but crashed in production - frustrating. I think I'm going to hold off on doing this transition until we're off Heroku.

PR #9780, opened 2025-03-12
Transitioning `OFFICE_IP`, `SLACK_NOTIFICATIONS_EMAIL`, and `LIVE_URL_HOST` to Doppler

sampoder:


sampoder:
Long term `LIVE_URL_HOST` should be rewritten to move out of creds, but I want to decouple that from the open sourcing project

app/mailers/admin_mailer.rb
```diff
@@ -2,7 +2,7 @@
 
 class AdminMailer < ApplicationMailer
   include Rails.application.routes.url_helpers
-  default to: -> { Rails.application.credentials.admin_email[:slack] }
+  default to: -> { Credentials.fetch(:SLACK_NOTIFICATIONS_EMAIL) }
```
garyhtou: When we had built out `Credentials`, I remember you wanted to preserve the [nesting format](https://github.com/hackclub/bank/blob/22d134ea20da90702ca334023c09f19db0297770/app/lib/credentials.rb#L9-L9). Is that something we still plan to use?

sampoder: I still plan on at least trying this for Stripe / Column etc! But in these cases there's just one `admin_email` in use so I wanted to cut down on the complexity.

sampoder: https://github.com/hackclub/hcb/pull/9785/files

PR #9814, opened 2025-03-13
Remove usage of Production encrypted credentials

garyhtou:
## Summary of the problem

Now that our production encrypt credentials file is empty, we can remove it and any remaining references.

This also removes `USE_PROD_CREDENTIALS` env var since that only was only needed to choose between loading production vs development encrypted credentials.

## Describe your changes

Delete `config/credentials/production.yml.enc` and usages of `USE_PROD_CREDENTIALS`.


app/services/twilio_verification_service.rb
```diff
@@ -9,7 +9,7 @@ class TwilioVerificationService
   )
 
   # This isn't private/sensitive so it's okay to keep here
-  VERIFY_SERVICE_ID = Rails.env.production? || ENV["USE_PROD_CREDENTIALS"]&.downcase == "true" ? "VAa06a66dad4c1ca3c199a46334ff11945" : "VAe30d49e92f634419aacdc8648948dc75"
+  VERIFY_SERVICE_ID = Credentials.fetch(:TWILIO, :SMS_VERIFY, :SERVICE_ID, fallback: "VAe30d49e92f634419aacdc8648948dc75")
```
garyhtou: For simplicity, I moved the production value to Doppler.

PR #9815, opened 2025-03-13
Create `HasBalanceMonitoring` concern

sampoder:
Don't merge for now - wait until production is back up

image


garyhtou:
bonus points if this was written on bart! 🚇

sampoder:
I wish! (unless you consider my couch to be BART)

PR #9852, opened 2025-03-16
Model out a system for "financial freezes"

### LINKED ISSUE #9817: [Admin][Org] Financial freeze
For each organization, we want to be able to financially freeze all new transactions. We want to prevent new transactions, but if any slip through the cracks, we must **fully process** them. In other words, we must block them at the source.

This is an admin-only feature and should be configured under the admin tab for each org.

- Stripe cards: decline all authorizations
- Prevent the creation of disbursements, achs, wires, donations etc.
- disallow final approval of reimbursement reports
- disallow final approval of payroll payment
- Disable donation page
- auto-return column bank account transactions

---

This column should be located on the `Event` model (not `Event::Configuration`). It's a column that will be accessed frequently, so I'd like to avoid having to load another model (`Event::Configuration`) to get that info.

Also, anyone got a cooler name for this feature?

sampoder:
WIP - and yes it should be a database column before this hits production!


app/models/event.rb
```diff
@@ -646,6 +646,10 @@ def dormant?
     !engaged?
   end
 
+  def frozen?
```
garyhtou, line 652:
```suggestion
  def finanically_frozen?
```

garyhtou (changes requested):
I gave this some more thought. This frozen check should be located inside [`ample_balance?`](https://github.com/hackclub/hcb/blob/main/app/services/shared/ample_balance.rb#L5-L5), and we should ensure that all balance checks are done with `ample_balance?`.

There are a few exceptions to this (e.g. Check Deposit where we don't need to check for ample_balannce).

sampoder:
I'd prefer to not - I think that `ample_balance?` should stick to what it's name is and not do unexpected things. For example, we use it to render different error messages in reimbursements. I don't think we should error with you don't have enough money if it's because you're frozen. 

Separately - we should redo ample balance because it's a nightmare to use. having something like `Event.ample_balance_for(amount)` would be nice

app/models/event.rb
```diff
@@ -646,6 +646,10 @@ def dormant?
...
+    Flipper.enabled?(:frozen, self)
```
garyhtou, line 653: Let's make this a db column on Event when you get a chance

sampoder: Oh yes lol

PR #9062, opened 2024-12-25
[V4 API] Create cards

Luke-Oldenburg:
Adds an endpoint to create HCB cards

POST `/api/v4/cards`

PR #9858, opened 2025-03-17
@Luke-Oldenburg [V4 API] Create cards

garyhtou:
Supersedes https://github.com/hackclub/hcb/pull/9062 due to git rewrite. Original PR by @Luke-Oldenburg

---

Adds an endpoint to create HCB cards

POST `/api/v4/cards`

### DIFF
--- a/app/controllers/api/v4/stripe_cards_controller.rb
+++ b/app/controllers/api/v4/stripe_cards_controller.rb
@@ -75,6 +75,7 @@ module Api
         
         rescue => e
           notify_airbrake(e)
+          render json: { error: "internal_server_error" }, status: :internal_server_error
       end
 
       def update
--- a/app/controllers/api/v4/stripe_cards_controller.rb
+++ b/app/controllers/api/v4/stripe_cards_controller.rb
@@ -31,7 +31,22 @@ module Api
         event = authorize(Event.find(params[:stripe_card][:event_id]))
         authorize event, :create_stripe_card?, policy_class: EventPolicy
 
-        sc = stripe_card_params
+        params.require(:stripe_card).permit(
+          :event_id,
+          :card_type,
+          :stripe_cardholder_id,
+          :stripe_shipping_name,
+          :stripe_shipping_address_city,
+          :stripe_shipping_address_line1,
+          :stripe_shipping_address_postal_code,
+          :stripe_shipping_address_line2,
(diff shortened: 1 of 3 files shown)


app/controllers/api/v4/stripe_cards_controller.rb
```diff
@@ -27,6 +27,57 @@ def transactions
         @has_more = false # TODO: implement pagination
       end
 
+      def create
+        event = authorize(Event.find(params[:stripe_card][:event_id]))
+        authorize event, :create_stripe_card?, policy_class: EventPolicy
+
+        sc = params.require(:stripe_card).permit(
+          :event_id,
```
garyhtou, line 35: Let's call this `organization_id`

app/controllers/api/v4/stripe_cards_controller.rb
```diff
@@ -27,6 +27,57 @@ def transactions
...
+          :card_type,
+          :stripe_shipping_name,
+          :stripe_shipping_address_city,
+          :stripe_shipping_address_line1,
+          :stripe_shipping_address_postal_code,
+          :stripe_shipping_address_line2,
+          :stripe_shipping_address_state,
+          :stripe_shipping_address_country,
```
garyhtou, line 43: Let's drop `stripe` from the name.

app/controllers/api/v4/stripe_cards_controller.rb
```diff
@@ -27,6 +27,57 @@ def transactions
...
+          :stripe_card_personalization_design_id,
+          :birthday
+        )
+
+        if current_user.birthday.nil?
+          user_params = sc.slice("birthday(1i)", "birthday(2i)", "birthday(3i)")
+          current_user.update(user_params)
+        end
```
garyhtou, line 51: For the API, it may be better to error if birthday is unset. They should first set the birthday, then recall this endpoint to issue the card.

app/controllers/api/v4/stripe_cards_controller.rb
```diff
@@ -27,6 +27,57 @@ def transactions
...
+
+        return render json: { error: "internal_server_error" }, status: :internal_server_error if current_user.birthday.nil?
```
garyhtou, line 53: There may be a better error code this. Maybe 400?

PR #9859, opened 2025-03-17
@Luke-Oldenburg Separate to and for in transfers

garyhtou:
Supersedes https://github.com/hackclub/hcb/pull/8936 due to git rewrite. Original PR owned by @Luke-Oldenburg

---

## Summary of the problem
The to column had both the recipient and the memo for certain transaction types.

## Describe your changes
I separated these into multiple columns.


app/views/events/transfers.html.erb
```diff
@@ -129,6 +128,15 @@
               <%= d.name %>
             <% end %>
           </td>
+          <td>
+            <% if d.is_a?(Disbursement) %>
+              <%= d.name %>
+            <% elsif d.is_a?(CardGrant) %>
+              Card Grant
```
garyhtou, line 135: we now have `purpose`

app/views/events/transfers.html.erb
```diff
@@ -129,6 +128,15 @@
...
+            <% else %>
+              <%= d.payment_for %>
```
garyhtou, line 137:
```suggestion
              <%= d.try(:payment_for) %>
```

sampoder:
Seems like we're using https://github.com/hackclub/hcb/pull/10133 now

PR #9007, opened 2024-12-19
Standardize transaction title bar

Luke-Oldenburg:
## Summary of the problem
Not all transaction types had an edit button and memo in the title bar.

## Describe your changes
I updated the title bar for all other title bars to have an edit button and transaction type. This PR is moreso just a demo of what it looks like. I'm honestly not sure that I like the changes after looking at it. Another idea would be to move the rename button for all charges into the meatball menu and make card charges look like the rest (e.g. "USER made a card charge at MERCHANT for $X.XX" instead of being special and having the edit button and memo as the title unlike the others.

PR #9860, opened 2025-03-17
@Luke-Oldenburg Standardize transaction title bar

garyhtou:
Supersedes https://github.com/hackclub/hcb/pull/9007 due to git rewrite. Original PR owned by @Luke-Oldenburg 

---

## Summary of the problem
Not all transaction types had an edit button and memo in the title bar.

## Describe your changes
I updated the title bar for all other title bars to have an edit button and transaction type. This PR is moreso just a demo of what it looks like. I'm honestly not sure that I like the changes after looking at it. Another idea would be to move the rename button for all charges into the meatball menu and make card charges look like the rest (e.g. "USER made a card charge at MERCHANT for $X.XX" instead of being special and having the edit button and memo as the title unlike the others.


garyhtou:
image

PR #9861, opened 2025-03-17
@Luke-Oldenburg [Policy] Let grant cardholders see all card charge info

garyhtou:
Supersedes https://github.com/hackclub/hcb/pull/9076 due to git rewrite. Original PR by @Luke-Oldenburg

---

## Summary of the problem
Grant cardholders were unable to see info on charges they made. This was discovered while trying to view receipts on HCB Mobile. The only allowed for admins and organizers but grant card users are just a cardholder and not a member of the organization.

## Describe your changes
I added a check to see if the user from the transaction record was the user being checked for policy.


Luke-Oldenburg:
This should be closed.

PR #9876, opened 2025-03-18
Fix `.ruby-version` for GitHub Codespaces

manuthecoder:
This pull request includes a small change to the `.ruby-version` file. The change updates the Ruby version from `3.3.6` to `3.3.7` to make GitHub Codespaces work


sampoder (changes requested):
Production is currently running on 3.3.6: https://github.com/hackclub/hcb/blob/main/production.Dockerfile.

Could we instead downgrade Codespaces' version of Ruby?

garyhtou:
Would it help if we upgraded the ruby version to latest? (3.4.2)

https://github.com/hackclub/hcb/issues/9880

garyhtou:
> Could we instead downgrade Codespaces' version of Ruby?

Microsoft only publishes minor versions Ruby in the Ruby devcontainer image.

https://mcr.microsoft.com/v2/devcontainers/ruby/tags/list

It'll default to the latest patch. So, I've upgraded us to 3.3.7 (latest patch). https://github.com/hackclub/hcb/pull/9922

PR #9877, opened 2025-03-18
Create a `auditor` admin role

sampoder:
Works towards https://github.com/hackclub/hcb/issues/5525. Thoughts?

PR #9878, opened 2025-03-18
[Organizer Position] Add Reader role

### LINKED ISSUE #9828: [User Roles] Add a Read only role
We already have a `manager` and `member` role. Let's add a `read only` role.

The goal is that adding this goal is increase the number of teenagers on HCB. At the moment, we have a large number of FIRST robotics teams. But those organizations usually only have the coach (no teenagers). By adding this role, we can encourage coaches to add their team.

@manuthecoder, would you be interested in working on this?

manuthecoder:
The goal of this pull request is to introduce the ability for users to invite members with read-only permissions.

Work in progress, resolves #9828


app/models/concerns/organizer_position/has_role.rb
```diff
@@ -6,7 +6,7 @@ module HasRole
     included do
       # The enum values will allow us to have a hierarchy of roles in the future.
       # For example, managers have access to everything below them.
-      enum :role, { member: 25, manager: 100 }
+      enum :role, { read_only: 5, member: 25, manager: 100 }
```
sampoder: lol suddenly I really want to call these people `readers` - see https://github.com/hackclub/hcb/pull/9877. thoughts?

i feel like it'd more inline with the member / manager terms where it's an "er"

garyhtou: The reader makes sense. However, I request that we give the admin read-only role a different name to prevent confusion.

Maybe we can call the admin roller "auditor"?

sampoder: Yes let's do it!

app/policies/check_deposit_policy.rb
```diff
@@ -31,14 +31,18 @@ def check_deposits_enabled?
     record.event.plan.check_deposits_enabled?
   end
 
-  def admin_or_user?
-    admin? || user?
+  def admin_or_standard_user?
```
garyhtou:
```suggestion
  def admin_or_member_user?
```

I suggest that in policies, when we say `member?`, it implies that their a member or above. We could also make that more clear by defining a method called `member_access?`

app/policies/comment_policy.rb
```diff
@@ -13,31 +13,31 @@ def resolve
   end
 
   def new?
-    user.admin? || users.include?(user)
+    !reader? && (user.admin? || users.include?(user))
```
garyhtou: Instead of `!reader?`, our logic should be "is the user's role greater than or equal to member?"

app/models/concerns/organizer_position/has_role.rb
```diff
@@ -6,7 +6,7 @@ module HasRole
     included do
       # The enum values will allow us to have a hierarchy of roles in the future.
       # For example, managers have access to everything below them.
-      enum :role, { member: 25, manager: 100 }
+      enum :role, { reader: 5, member: 25, manager: 100 }
```
garyhtou, line 9: Good number to choose! Good job here

garyhtou (commented):
This is on the right track!

app/policies/comment_policy.rb
garyhtou: Also, i personally think we should allow people to see the forms/`new` pages, but disallow them to submit them. We largely already have that logic in a lot of places since members can see the event settings page, but can't submit the forms.

any thoughts on this? I could also see the argument that the UI looks very cluttered with a lot of forms that you can't submit. At the same time, some information is only accessible via the form (e.g. event settings page).

app/models/organizer_position.rb
```diff
@@ -54,6 +54,13 @@ def tourable_options
     }
   end
 
+  def self.role_at_least?(user, role)
+    return true if user&.admin?
+
+    current = find_by(user: user)&.role
+    current && roles[current] >= roles[role]
+  end
+
```
garyhtou: I believe we need to further filter down this search for `user` with `role` in `event`. Otherwise, reader role in one organization may accidentally grant permissions in another organization

app/controllers/events_controller.rb
```diff
@@ -270,7 +272,7 @@ def team
 
     @all_positions = @event.organizer_positions
                            .joins(:user)
-                           .where(role: @filter || %w[member manager])
+                           .where(role: @filter || %w[member manager reader])
                            .where(organizer_signed_in? ? "users.full_name ILIKE :query OR users.email ILIKE :query" : "users.full_name ILIKE :query", query: "%#{User.sanitize_sql_like(@q)}%")
                            .order(created_at: :desc)
```
garyhtou: I think this should have the same behavior without us needing to maintain a list of roles here

```suggestion
@all_positions = @all_positions.where(role: @filter) if @filter
@all_positions = @all_positions.where(organizer_signed_in? ? "users.full_name ILIKE :query OR users.email ILIKE :query" : "users.full_name ILIKE :query", query: "%#{User.sanitize_sql_like(@q)}%")
                               .order(created_at: :desc)
```

app/policies/comment_policy.rb
```diff
@@ -13,31 +13,31 @@ def resolve
   end
 
   def new?
-    user.auditor? || users.include?(user)
+    user.auditor? || OrganizerPosition.role_at_least?(user, event, :member)
```
garyhtou:
```suggestion
    user.auditor? || OrganizerPosition.role_at_least?(user, event, :reader)
```

app/policies/comment_policy.rb
```diff
@@ -13,31 +13,31 @@ def resolve
...
+    OrganizerPosition.role_at_least?(user, event, :member)
   end
 
   def react?
     show?
   end
 
   def show?
-    user&.auditor? || (users.include?(user) && !record.admin_only)
+    user&.auditor? || OrganizerPosition.role_at_least?(user, event, :member)
```
garyhtou:
```suggestion
    user&.auditor? || OrganizerPosition.role_at_least?(user, event, :reader)
```

app/policies/event_policy.rb
```diff
@@ -7,7 +7,7 @@ def index?
 
   # Event homepage
   def show?
-    is_public || auditor_or_user?
+    is_public || allowed_user?
```
garyhtou: Because of the way allow users is written, readers can't see this.

I'd would also recommend renaming `allow_user?` to be more specific.

app/policies/hcb_code_policy.rb
```diff
@@ -61,4 +61,8 @@ def present_in_events?
     record.events.select { |e| e.try(:users).try(:include?, user) }.present?
   end
 
+  def event
+    record.events.find { |e| e.try(:users).try(:include?, user) }
```
garyhtou: There may be a bug where it picks an event with lower permissions. I'd suggest using the `has_role?` approach from Comment Policy

app/policies/invoice_policy.rb
```diff
@@ -14,35 +14,35 @@ def index?
   end
 
   def new?
-    !unapproved? && (is_public || admin_or_user)
+    !unapproved? && (is_public || OrganizerPosition.role_at_least?(user, record&.sponsor&.event, :member))
```
garyhtou: I think we remove the safe nav here.

app/policies/organizer_position_deletion_request_policy.rb
```diff
@@ -10,7 +10,7 @@ def new?
   end
 
   def create?
-    return false unless user.present?
+    return false unless OrganizerPosition.role_at_least?(user, record.event, :reader)
```
garyhtou:
```suggestion
    return false unless OrganizerPosition.role_at_least?(user, record.event, :member)
```

garyhtou:
Todo from meeting:
1. Apply fixes from above
2. Double check all your policies
3. Record a video of you going through HCB as a reader

app/controllers/invoices_controller.rb
```diff
@@ -133,6 +133,8 @@ def create
     end
 
     redirect_to @invoice
+  rescue Pundit::NotAuthorizedError
+    redirect_to event_invoices_path(@event), flash: { error: "You are not authorized to perform this action." }
```
garyhtou:
```suggestion
    raise
```

garyhtou (approved):
Nice work here!

PR #9896, opened 2025-03-18
Render "USD" on all non-US locales

sampoder:
Closes https://github.com/hackclub/hcb/issues/9853 - didn't use IP because that would be potentially expensive.

### DIFF
--- a/app/controllers/donations_controller.rb
+++ b/app/controllers/donations_controller.rb
@@ -15,7 +15,12 @@ class DonationsController < ApplicationController
   before_action :hide_seasonal_decorations
   skip_before_action :redirect_to_onboarding
 
-  before_action { @force_fullstory = true }
+  before_action do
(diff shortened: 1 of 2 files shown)


app/controllers/donations_controller.rb
```diff
@@ -15,7 +15,12 @@ class DonationsController < ApplicationController
   before_action :hide_seasonal_decorations
   skip_before_action :redirect_to_onboarding
 
-  before_action { @force_fullstory = true }
+  before_action do
+    @force_fullstory = true
+    if (request&.env&.[]("HTTP_ACCEPT_LANGUAGE") && !request&.env&.[]("HTTP_ACCEPT_LANGUAGE")&.include?("-US")) || !@event.country_US?
```
garyhtou, line 20: Add a safe navigator on `@event` since `set_event` doesn't run on every action

PR #9900, opened 2025-03-18
Begin using `auditor_signed_in?`

sampoder:
Needs testing + I need to figure out `organizer_signed_in?`


sampoder:
Tested and LGTM to me - will merge tonight

PR #9908, opened 2025-03-19
Document getting keys

sampoder:


dev-docs/development.md
```diff
@@ -163,6 +163,10 @@ brew install imagemagick
 
 We used [Doppler](https://www.doppler.com/) to manage our credentials; if you have access to Doppler, you can set a `DOPPLER_TOKEN` in your `.env` file. Otherwise, you can provide credentials via a `.env.development` file [(view example)](.env.development.example).
 
+HCB relies on two services for the majority of it's financial features: Stripe and Column. We recommend creating a Stripe account in "test mode". Read more here: [docs.stripe.com/test-mode](https://docs.stripe.com/test-mode#test-mode). You can register for a Column account [here](https://dashboard.column.com/register); after their onboarding questions, select "Skip to Sandbox".
+
+We also include OpenAI and Twilio keys in our `.env.development` file. Information about obtaining these keys is available in these articles on [help.openai.com](https://help.openai.com/en/articles/4936850-where-do-i-find-my-openai-api-key) and [twilio.com](https://www.twilio.com/docs/iam/api-keys/keys-in-console).
```
garyhtou, line 168: I'd suggest adding: If you're a hack clubber, you can get a free, limited-use OpenAI key from the slack channel.

However, this would require us to make the OpenAI client's base URL configurable, if not already. (since the free OpenAI service proxies requests to OpenAI and is hosted on a different domain).

PR #9920, opened 2025-03-19
[Column][Returned Wired] Use `return_reason`

sampoder:


app/controllers/column/webhooks_controller.rb
```diff
@@ -53,7 +53,7 @@ def handle_ach_outgoing_transfer_returned
     end
 
     def handle_swift_outgoing_transfer_returned
-      Wire.find_by(column_id: @object[:id])&.mark_failed!(@object[:return_details].pick(:description)&.gsub(/\(trace #: \d+\)\Z/, "")&.strip)
+      Wire.find_by(column_id: @object[:id])&.mark_failed!(@object[:return_reason]&.gsub(/\(trace #: \d+\)\Z/, "")&.strip)
```
garyhtou, line 56: Looks like you beat me to this!  Do we want to keep the existing `return_details` logic? I'm assuming it was used before, but I'm not certain.

```ruby
      reason = @object[:return_details]&.pick(:description)&.gsub(/\(trace #: \d+\)\Z/, "")&.strip ||
        @object[:return_reason]

      Wire.find_by(column_id: @object[:id])&.mark_failed!(reason)
```

sampoder: `return_details` is the format Column uses for ACHs - for Wires they only use `return_reason`

Screenshot 2025-03-19 at 11 56 50 AM

garyhtou: Okay, perfect

PR #9930, opened 2025-03-20
Add Help Scout Beacon integration and update support button

leowilkin:
## Describe your changes

Added the HelpScout Beacon to the support button, and passes parameters like name & email to Helpscout which is pretty sick ngl

### What did I change?

I added:
- the embed to the main application (will only run once logged in)
- changed the 'get help' button to be the trigger for the widget
- added the relevant env vars to the `.env.development.example`, and `.env.development`

my code might break? i'm not an engineer... however this is really cool, so yeah.

cc. @garyhtou

### DIFF
--- a/dev-docs/production/migrations/sidekiq-redis.md
+++ b/dev-docs/production/migrations/sidekiq-redis.md
@@ -1,0 +1,18 @@
+# Migrating Sidekiq's Redis
+
+We store our queue of jobs in a Redis instance. When / if we change hosting providors, we can't lose that data.
+
+Here are the steps to do the migration:
+
+1) Ensure that AOF persistence is enabled in the new server's Redis configuration, [view Redis' docs](https://redis.io/docs/latest/operate/oss_and_stack/management/persistence/) for more context. This should be as simple as changing `appendonly no` to `appendonly yes` in the `redis.conf` (`/etc/redis/redis.conf` when using Hatchbox).
+2) Install [RIOT](https://github.com/redis/riot) locally (or any where you like to work!).
+3) Turn on maintenance mode and disable Sidekiq - this is to prevent new jobs being added to the queue.
+4) Run the following command to migrate the keys from your old Redis instance to your new Redis instance:
+  
+    ```
+    riot replicate [old_connection_string] [new_connection_string]
+    ```
+    
+    If migrating from Heroku, you'll need to add `--source-insecure --no-ttl` [(context)](https://stackoverflow.com/questions/65042551/ssl-certification-verify-failed-on-heroku-redis/). Read more on [redis.github.io](https://redis.github.io/riot/#_replication).
+5) Change `REDIS_URL` in your enviroment variables.
+6) Turn off maintenance mode, enable Sidekiq and verify the queue has migrated!


app/views/application/_support_button.html.erb
```diff
@@ -36,5 +33,8 @@
     <a class="support-menu__item" href="<%= docs_api_index_path %>" target="_blank">
       <%= inline_icon "code", size: 24 %> API docs
     </a>
+    <a class="support-menu__item" href="javascript:void(0);" onclick="window.Beacon && (window.Beacon('identify', { name: '<%= current_user&.name %>', email: '<%= current_user&.email %>' }), window.Beacon('session-data', { url: '<%= url_params[:url] %>', location: '<%= url_params[:location] %>', user_id: '<%= url_params[:user_id] %>' }), window.Beacon('open'))">
+      <%= inline_icon "email", size: 24 %> Get Help
```
sampoder:
```suggestion
      <%= inline_icon "email", size: 24 %> Get help
```
Tiny thing but @dwahbe will appreciate the consistency

app/views/application/_support_button.html.erb
```diff
@@ -14,12 +11,12 @@
...
       <% url_params = {
-           name: current_user&.name,
-           email: current_user&.email,
            url: "#{request.base_url}#{request.original_fullpath}",
            location: "#{routing[:controller]}##{routing[:action]} #{routing[:id]}".strip,
+           user_id: current_user&.id
          } %>
-
+    <% end %>
+    <% if current_user&.events&.not_demo_mode&.any? %>
```
sampoder: Any reason for these two lines? They seem redundant since

```
<% if current_user&.events&.not_demo_mode&.any? %>
```

is also on Line 6

leowilkin: whoopsie - didn't see that there. all good to remove.

app/views/layouts/application.html.erb
```diff
@@ -33,6 +33,27 @@
...
+          }
+          if(e.Beacon=n=function(t,n,a){
+            e.Beacon.readyQueue.push({method:t,options:n,data:a})
+          },n.readyQueue=[],"complete"===t.readyState)return a();
+          e.attachEvent?e.attachEvent("onload",a):e.addEventListener("load",a,!1)
+        }(window,document,window.Beacon||function(){});
+      </script>
+
+      <script type="text/javascript">
+        window.Beacon('init', Credentials.fetch(:HELPSCOUT,:BEACON_ID))
```
sampoder: We can honestly simplify this by just putting the key inline, it's a public key so it's not secret

leowilkin: https://hackclub.slack.com/archives/C026RKHLPNJ/p1742503794546119?thread_ts=1742493033.584299&cid=C026RKHLPNJ ?

sampoder: Putting it in this file will make it public - this Javascript runs client side

leowilkin: I think it would be better not to hardcode this into HCB because then it would be functional on development, which may confuse users & annoy ops with `foobar` support requests

app/views/application/_support_button.html.erb
sampoder:
```suggestion
```

app/views/layouts/application.html.erb
sampoder: Hmm - that's fair - let's keep it for now, can you send me the Beacon ID and I'll add it to Doppler?

leowilkin: It's on the thread on Slack! Thanks :)

sampoder (approved):
LGTM - nice work!

leowilkin:
Not sure what happened with the erb thingymagik, so I'll step off this and let you two figure it out..!

garyhtou:
oops

garyhtou:
one sec— looks like I messed something up

PR #9948, opened 2025-03-24
[RUBY] Bump to 3.4.2

KiKoS0:
## Summary of the problem
Bumps ruby version to [`3.4.2`](https://www.ruby-lang.org/en/news/2025/02/14/ruby-3-4-2-released/), both [`3.4.1`](https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-1-released/) and [`3.4.2`](https://www.ruby-lang.org/en/news/2025/02/14/ruby-3-4-2-released/) are minor bug fixes / metadata update. The main one with different/deprecated behavior is the [`3.4.0`](https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-0-released/) version.

I looked through the extended list of changes in https://github.com/ruby/ruby/blob/v3_4_2/NEWS.md, along with an [OSS example of bumping to a 3.4+ version](https://github.com/opf/openproject/pull/17651/files) and it doesn't seem that we have anything to change as far as I know.

The `Kernel#Float()` now accepting a decimal string with decimal part omitted like `1.` will now make this match in `numeric?` but i'm not sure if that's unwanted, possibly not:
https://github.com/hackclub/hcb/blob/068bb991c2d1e550a44256601d3729bfa43af0b3/app/models/metric/event/words.rb#L43-L72


garyhtou:
Very thorough work! Nice job. I'll let @albertchae handle the review here.

garyhtou:
> The Kernel#Float() now accepting a decimal string with decimal part omitted like 1. will now make this match in numeric? but i'm not sure if that's unwanted, possibly not:
> https://github.com/hackclub/hcb/blob/068bb991c2d1e550a44256601d3729bfa43af0b3/app/models/metric/event/words.rb#L43-L72


cc @YodaLightsabr

albertchae (approved):
LGTM

We have `RUBYOPT=-W:deprecated` so we should see deprecation warnings.

And the main thing from https://github.com/ruby/ruby/blob/v3_4_2/NEWS.md#compatibility-issues does seem to be the Float parsing for trailing `.`. We should also maybe look at uses of `to_f`

PR #9950, opened 2025-03-24
Admin > Auditor

sampoder:
Display the one that makes more sense


garyhtou (dismissed):
I would advocate for "X is an admin" even for auditors. I think saying "X is an auditor" can become confusing for our users. The details of auditor vs admin vs superadmin are not important for normal users.

sampoder:
Going to merge this in because it fixes a bug - I've kept them separate because being an admin suggests that they can take action, while auditors can't

PR #9956, opened 2025-03-24
Adds Network IDs to the Card Charge Transaction v4 API

Mohamad-Mortada:
Adds Network IDs to the Card Charge Transaction v4 API

## Summary of the problem
I can't access network IDs from card charges via the v4 API

## Describe your changes
Added another field under merchants called "network_id" to return this


sampoder (approved):
LGTM!

PR #9958, opened 2025-03-25
Un-hardcode some SMTP/Email stuff

cyteon:
This PR unhardcodes the selection between letter opener or SMTP on dev, by making some env vars.
- If SMTP__ENABLED is false or not set, it will use letter opener

It also un-harcodes the emails in app/mailers/application_mailer.rb!
- if SMTP__DOMAIN is not set, it falls back to hcb.hackclub.com for prod, and staging.hcb.hackclub.com for dev
- if SMTP__USERNAME is not set, it falls back to "hcb"

### DIFF
--- a/config/credentials/development.yml.enc
+++ b/config/credentials/development.yml.enc
@@ -1,1 +1,0 @@
-GxZcLC0/przYcpYa9wTLGIOSSmmrHczJ4jRWZhF0e/Gm6gUpTt8YrCrrtM7A8qPI2Z1WLDxsqXcE+U2Ekw==--2j640khvlDEpHtcy--+s4x3Zm9OTDPkD/38ccf6A==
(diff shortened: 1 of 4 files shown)


config/credentials/development.yml.enc
```diff
@@ -0,0 +1 @@
+GxZcLC0/przYcpYa9wTLGIOSSmmrHczJ4jRWZhF0e/Gm6gUpTt8YrCrrtM7A8qPI2Z1WLDxsqXcE+U2Ekw==--2j640khvlDEpHtcy--+s4x3Zm9OTDPkD/38ccf6A==
```
NeonGamerBot-QK: ummm u prob shouldnt upload ur enc creds to the repo

NeonGamerBot-QK (changes requested):

config/credentials/development.yml.enc
cyteon: Its encrypted and there is nothing in it, it was just generated, i checked it only includes comments :P

cyteon: imma remove it

garyhtou:
Can you help me understand the benefits of using a real STMP server in development? Letter opener has been working very well for us so far. I would prefer to keep development as letter opener only unless we have a strong reason to make it configurable.

cyteon:
Well the primary reason is making the dev instance closer to production, and would provide more realistic email testing. 
Is it absolutely necessary? No. But is it nice to have? Yes, in my opinion.

cyteon:
Github broke my fork, see #10009

PR #9960, opened 2025-03-26
Remove erroring line

sampoder:


app/controllers/events_controller.rb
```diff
@@ -359,7 +359,7 @@ def card_overview
     @view = cookies[:card_overview_view] || "grid"
 
     @user_id = params[:user].presence
-    @user = User.find(params[:user]) if params[:user]
+    @user = User.find_by_id(params[:user]) if params[:user]
```
garyhtou: I'm not sure if `@user` is used anywhere. maybe we can drop this line entirely

PR #9964, opened 2025-03-26
Fix codespace

cyteon:
Cause the ruby version was changed to 3.4.2, the codespace no longer works.
Fixed this by changing variant to `3.4-bookworm` in Dockerfile and devcontainer.json

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-ARG VARIANT="3.3-bookworm"
+ARG VARIANT="3.4-bookworm"
 FROM mcr.microsoft.com/devcontainers/ruby:${VARIANT}
 # Install Rails
 RUN su vscode -c "gem install rails webdrivers"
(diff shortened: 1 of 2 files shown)


ploszukiwacz:
very nerdy

AbanobGRH (approved):

garyhtou:
Thanks for putting this together! I think this might have been resolved by https://github.com/hackclub/hcb/pull/9979.

Let me know if that works for you!

cyteon:
May i ask, why was a duplicate of my PR merged instead of just merging mine?

garyhtou:
It looks like https://github.com/hackclub/hcb/pull/9979 was created before this one. We aim to merge the first PR that solves the issue. In this case, both PRs solved the issue, but https://github.com/hackclub/hcb/pull/9979 was created first. Sorry about that. We're totally happy to accept PRs from you!

cyteon:
I think mine was created first, merged was 9979, mine is 9964. 9964 is smaller than 9979, which means it was made first :/
Also, 9979:
image

Mine, aka 9964:
image

:p

garyhtou:
Oop, my bad! You're totally right. I was looking at the date when it would have been easier and more accurate to look at the PR number. I apologize.

cyteon:
Dont worry, its fine :p

PR #9977, opened 2025-03-26
Allow creation of orgs from Airtable ID

rluodev:
## Summary of the problem

need an easy way to create orgs from airtable id

## Describe your changes

now there's a page to do this


db/migrate/20250321200657_add_event_assoc_airtable_record_id_col.rb
```diff
@@ -0,0 +1,8 @@
+class AddEventAssocAirtableRecordIdCol < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def change
+    add_column :events, :application_airtable_record_id, :string
+    add_index :events, :application_airtable_record_id, unique: true, algorithm: :concurrently
```
sampoder: IMO instead of setting it in a column on the `Event` model, we should set the HCB ID on Airtable from HCB

app/controllers/admin_controller.rb
```diff
@@ -93,6 +96,23 @@ def event_create
...
+      name: application["Event Name"],
+      country: country&.alpha2,
+      point_of_contact_id: current_user.id,
+      approved: true,
+      organized_by_teenagers: application["TEEN"] == "Teen",
+      demo_mode: true
+    ).run
+
+    redirect_to events_admin_index_path, flash: { success: "Successfully created #{params[:name]}" }
+  rescue => e
```
sampoder: So in here we would update the application record's HCB ID

db/migrate/20250321200657_add_event_assoc_airtable_record_id_col.rb
sampoder: I guess the think here is that the Airtable record belongs to the Event the Event doesn't belong to the airtable record

garyhtou: I have asked Ruien to set it on both ends
- `application_airtable_record_id` on the events model
- and an HCB URL on the airtable

sampoder: I think we should have one source of truth and honestly I'd prefer not to tie our codebase back into Airtable

garyhtou: That's reasonable. Since we want to move applications into HCB in the future, I'm working with Ruien to set an `HCB ID` on Airtable so that we have a reliable foreign key to the Event on HCB. We'll drop the new column (`application_airtable_record_id`).

app/controllers/admin_controller.rb
```diff
@@ -93,6 +96,30 @@ def event_create
...
+    country = ISO3166::Country.find_country_by_any_name(application["Event Location"])
+    event = ::EventService::Create.new(
+      name: application["Event Name"],
+      country: country&.alpha2,
+      point_of_contact_id: current_user.id,
+      approved: true,
+      organized_by_teenagers: application["TEEN"] == "Teen",
+      demo_mode: true,
+      application_airtable_record_id: params[:airtable_record_id]
+    ).run
```
```suggestion
    event = ::EventService::Create.new(
      name: application["Event Name"],
      country: country&.alpha2,
      point_of_contact_id: current_user.id,
      approved: true,
      organized_by_teenagers: application["TEEN"] == "Teen",
      demo_mode: true
    ).run
```

app/controllers/admin_controller.rb
```diff
@@ -93,6 +96,29 @@ def event_create
...
+      organized_by_teenagers: application["TEEN"] == "Teen",
+      demo_mode: true
+    ).run
+
+    record["HCB account URL"] = "https://hcb.hackclub.com/#{event.slug}"
+    record["HCB ID"] = event.id
+
+    record.save
+
+    redirect_to event_path(event), flash: { success: "Successfully created #{params[:name]}" }
```
```suggestion
    redirect_to event_path(event), flash: { success: "Successfully created #{event.name}" }
```

PR #9984, opened 2025-03-26
Added personal grant card locks to creation UI

Mohamad-Mortada:
## Summary of the problem
The API allows for personal card locks but the UI doesn't yet. This changes things :)

## Describe your changes
I added new fields in a collapsable part of the card grant creation UI

image

### DIFF
--- a/app/controllers/card_grants_controller.rb
+++ b/app/controllers/card_grants_controller.rb
@@ -20,11 +20,6 @@ class CardGrantsController < ApplicationController
 
     last_card_grant = @event.card_grants.order(created_at: :desc).first
 
-    if last_card_grant.present?
-      @card_grant.amount_cents = last_card_grant.amount_cents
-      @card_grant.merchant_lock = last_card_grant.merchant_lock
-      @card_grant.category_lock = last_card_grant.category_lock
-    end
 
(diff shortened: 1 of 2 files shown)


app/controllers/card_grants_controller.rb
```diff
@@ -20,11 +20,6 @@ def new
 
     last_card_grant = @event.card_grants.order(created_at: :desc).first
 
-    if last_card_grant.present?
-      @card_grant.amount_cents = last_card_grant.amount_cents
-      @card_grant.merchant_lock = last_card_grant.merchant_lock
-      @card_grant.category_lock = last_card_grant.category_lock
-    end
```
sampoder, line 27: Maybe we add a checkbox that allows for inheriting the previous grant's policies? I don't want to push a breaking change

app/views/card_grants/new.html.erb
```diff
@@ -27,6 +27,24 @@
       <%= form.text_field :purpose, placeholder: "Pizza for a club meeting" %>
     </div>
 
+    <details>
+      <summary class="mb-5">Personal Card Locks (optional)</summary>
```
sampoder, line 31:
```suggestion
      <summary class="mb-5">Card-specific locks (optional)</summary>
```

PR #9985, opened 2025-03-27
Added Merchant Info to V3 Card Charge endpoint

Mohamad-Mortada:
## Summary of the problem

I wasn't able to fetch the info of card charges via the v3 api.

## Describe your changes

I added merchant info: name, smart name, country, and network id and made sure the docs reflected it elegantly

### DIFF
--- a/app/api/api/entities/card_charge.rb
+++ b/app/api/api/entities/card_charge.rb
@@ -11,6 +11,23 @@ module Api
           expose :date
         end
 
+        expose :merchant, using: Merchant do |hcb_code|
+          stripe_transaction = hcb_code.ct&.raw_stripe_transaction&.stripe_transaction
+          stripe_authorization = hcb_code.pt&.raw_pending_stripe_transaction&.stripe_transaction
+          merchant_data = (stripe_transaction || stripe_authorization)&.dig("merchant_data")
+
+          {
+            name: merchant_data&.dig("name"),
+            smart_name: begin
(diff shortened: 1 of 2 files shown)


garyhtou:
Is this related to a YSWS/card grant integration?

Mohamad-Mortada:
This is more related to a yellow page integration as it'll allow me to sift through most organizations' card charges and add network IDs that are missing.

YodaLightsabr (commented):
Currently, only the merchant logo and name are publicly available in Transparency Mode. I'm comfortable adding those to the API. However, exposing the network ID can share a lot of information with the public. In some cases, it could even expose the exact physical store something was purchased at, which could give away the geographical location of our users.

sampoder:
I think I'm comfortable with network IDs being public - for a lot of card transactions you can figure out the geographic location of the merchant from their name.

Just looking back at a couple on my account, you can work out locations based on these:

Screenshot 2025-04-22 at 1 37 21 AM

garyhtou:
For context here, I had chatted with @thedev132 about the goals of this PR. The end goal there is to push data from HCB into Yellow Pages. We decided that I'll provide @thedev132 with a dump of data for updating yellow pages and he'll make a PR for it. Then, afterwards, to identify which merchants need to be pushed from HCB to yellow pages, we'll build a page on HCB for that. This solution is more clean/easier than using the v3 API to pull all that data.

PR #9996, opened 2025-03-29
Improve analytics

YodaLightsabr:
## Summary of the problem
We're open source now! And we're probably going to make some noise & generate some publicity about HCB.

## Describe your changes
Adds Plausible analytics to track basic pageviews for our homepage & transparent organizations (doesn't track activity like FullStory would)


garyhtou (approved):

garyhtou:
Great! We're almost there!

PR #10007, opened 2025-03-30
Move hack on HCB to GitHub

sampoder:
I guess this form is no longer needed

### DIFF
--- a/config/constants.yml
+++ b/config/constants.yml
@@ -6,6 +6,6 @@
 # Learn more: https://guides.rubyonrails.org/configuring.html#custom-configuration
 
 shared:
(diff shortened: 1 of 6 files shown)


config/constants.yml
```diff
@@ -6,6 +6,6 @@
 # Learn more: https://guides.rubyonrails.org/configuring.html#custom-configuration
 
 shared:
-  hack_on_hcb_form_url: https://forms.hackclub.com/t/mWH2ycXiDdus
+  github_url: https://github.com/hackclub/hcb
   security_reporting_email: hcb-security@hackclub.com
   github_repo: https://github.com/hackclub/hcb
```
garyhtou, line 11: Looks like we have duplicate values here

PR #10021, opened 2025-04-01
change yt video link and change copy (#10017)

sampoder:


skyfallwastaken:
fyi “Before signing please watching” should be “Before signing please watch”

PR #10028, opened 2025-04-01
Fix `return_to` leading to non-HCB URLs

sampoder:


garyhtou (commented):
Please do this for all other usages of the `return_to` params. Either make that a `before_action`. Or define a method like:

```
def return_to_url
  url_from(params[:return_to])
end
```
that's used in the views

PR #10031, opened 2025-04-02
 [Loops] Sync hcbHasActiveOrg

garyhtou:
I'm adding `hcbHasActiveOrg`for better Loops.so audience filtering. It will be used to send the teenager referral program email tomorrow.

### DIFF
--- a/app/services/user_service/sync_with_loops.rb
+++ b/app/services/user_service/sync_with_loops.rb
@@ -3,7 +3,7 @@
 module UserService
   class SyncWithLoops
     def initialize(user_id:, queue: Limiter::RateQueue.new(2, interval: 1), new_user: false)
-      @user = User.find(user_id)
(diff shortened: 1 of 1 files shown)


app/services/user_service/sync_with_loops.rb
```diff
@@ -3,7 +3,7 @@
 module UserService
   class SyncWithLoops
     def initialize(user_id:, queue: Limiter::RateQueue.new(2, interval: 1), new_user: false)
-      @user = User.find(user_id)
+      @user = User.find(user_id).includes(:events)
```
sampoder:
```suggestion
      @user = User.includes(:events).find(user_id)
```

garyhtou: Whoops, thanks. We can drop the includes entirely here. Since it's just one record, there's no advantage eager loading it

PR #10055, opened 2025-04-03
Add message and UTM params to donation api

karmanyaahm:
## Summary of the problem
My organization wants to check whether members have paid dues automatically through the API. To do so, they will put their ID number in the HCB message. However, there's no way to access the message through the API.

## Describe your changes
Expose the parameters on the API. In addition, expose utm params as they might be useful for analyzing donations

### DIFF
--- a/app/api/api/entities/donation.rb
+++ b/app/api/api/entities/donation.rb
@@ -22,7 +22,16 @@ module Api
           ]
         }
         expose :recurring?, as: :recurring, documentation: { type: "boolean" }
-
+        expose :message
+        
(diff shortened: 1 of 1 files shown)


sampoder:
Unfortunately for many organisations this data would be private - I'm not sure we can put it in the public API

karmanyaahm:
Makes sense, I'll just figure out how to use the v4 API then

sampoder:
DM me on Slack - I can help

karmanyaahm:
Will do, thanks!

PR #10074, opened 2025-04-04
Create a custom timeout-related view

### LINKED ISSUE #10068: [Error] Give timeout errors a unique screen
This will make it easier for us to tell if it's a timeout or a 500

sampoder:
Closes https://github.com/hackclub/hcb/issues/10068

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -52,6 +52,10 @@ class ApplicationController < ActionController::Base
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -52,6 +52,10 @@ class ApplicationController < ActionController::Base
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
+  rescue_from Rack::Timeout::RequestTimeoutException do
+    redirect_to timeout_path
```
garyhtou, line 56: Instead of triggering another request, could we render the timeout template here?

Issue #10075, opened 2025-04-04
Allow HCB physical cards to be activated via the API


scooterthedev:
### Describe your goals

Looking around the HCB Mobile app, playing around, and ended up activating a physical card without the last 4 digits.

### Describe the bug

**This bug was most likely caused by the HCB API not validating if the last 4 digits were inputted**. 

The bug happened through hcb mobile. I went to the cards section, and went to a card I was going to activate shortly, and clicked activate thinking it will ask me for the last 4 digits like the web hcb does, but it didn't and just activated it. Then once I checked hcb web, I saw that it was actually activated and not just a glitch.

scooterthedev:
Temp fix via commit #10076

PR #10085, opened 2025-04-06
[V4 API] Added last4 for v4 api physcial HCB card activations.

scooterthedev:
## Summary of the problem

This method under the v4 api allowed anyone to activate a physical HCB card without the last 4 digits being submitted. #10075

## Describe your changes

- Adds a check to mandate and verify the last 4 digits of the physical HCB card before activation. This can be called via a `PATCH` request to `/api/v4/cards/:card_id` with the Authorization token being your Bearer and the URL parameters being `last4: last 4 digits` and `status: active`. `Status: frozen` does not require last4.

- Also removed the ability to freeze cancelled cards because the card is frozen FOREVER already ;)

This will affect the HCB mobile app as it will now require the last4 before it can authorize the activation CC: @thedev132


app/controllers/api/v4/stripe_cards_controller.rb
```diff
@@ -30,11 +30,47 @@ def transactions
...
 
-        if params[:status] == "frozen" && !@stripe_card.frozen?
+        if params[:status] == "frozen"
+          if @stripe_card.canceled?
+            return render json: { error: "Card has been cancelled, it can't be frozen" }, status: :unprocessable_entity
+          end
+
           @stripe_card.freeze!
+        elsif params[:status] == "active"
+          if params[:last4].blank?
```
sampoder, line 94: If the card was initially activated (`@card.initially_activated?`) we should allow it to be be activated again without the last four

scooterthedev: Isn't that handled on line 56?

scooterthedev: I _think_ I fixed it. Not the most efficient way, but it works.

scooterthedev:
@sampoder quick reminder ;)! I think I fixed your review, but if you could double check, that would also be awesome.

app/controllers/api/v4/stripe_cards_controller.rb
```diff
@@ -75,11 +75,52 @@ def create
...
+        if params[:status] == "frozen"
+          if @stripe_card.canceled?
+            return render json: { error: "Card has been cancelled, it can't be frozen" }, status: :unprocessable_entity
+          end
+
           @stripe_card.freeze!
+        elsif params[:status] == "active"
+          if @stripe_card.initially_activated?
+            @stripe_card.defrost! unless @stripe_card.stripe_status == "active"
+            return render json: { success: "Card activated!" }
```
sampoder, line 91:
```suggestion
            return render json: { success: "Card unfrozen!" }
```

app/controllers/api/v4/stripe_cards_controller.rb
```diff
@@ -75,11 +75,52 @@ def create
...
+          end
+
+          if @stripe_card.initially_activated?
+            if @stripe_card.created_at < Date.new(2024, 2, 22)
+              @stripe_card.defrost! unless @stripe_card.stripe_status == "active"
+              return render json: { success: "Card activated!" }
+            else
+              return render json: { error: "Card already activated" }, status: :unprocessable_entity
+            end
+          end
```
sampoder: This is probably redundant

sampoder (approved):
LGTM - thanks!

PR #10129, opened 2025-04-09
[Changelog] Add notification badge

YodaLightsabr:
Adds a notification badge to indicate unread posts

image


YodaLightsabr:
Merging to test fetch logic. Shouldn't be an issue since it's frontend-only and gated behind Flipper

PR #10130, opened 2025-04-10
Remove "event" from disbursements#show

zachlatta:
Before:

Screenshot 2025-04-10 at 10 23 46 AM

After:

Screenshot 2025-04-10 at 10 24 03 AM

Make it a little cleaner, and we don't use "event" elsewhere in the HCB interface.


sampoder (approved):
LGTM! FWIW this is an admin-only page

PR #10137, opened 2025-04-10
Let grant cardholders see all card charge info

Luke-Oldenburg:
Supersedes #9861


app/policies/hcb_code_policy.rb
```diff
@@ -2,7 +2,7 @@
 
 class HcbCodePolicy < ApplicationPolicy
   def show?
-    user&.admin? || present_in_events?
+    user&.admin? || present_in_events? || record.stripe_cardholder&.user_id == user&.id
```
sampoder:
```suggestion
    user&.admin? || present_in_events? || (record.stripe_cardholder.present? && record.stripe_cardholder.user == user)
```
Screenshot 2025-04-14 at 5 48 21 PM

If `user` and `record.stripe_cardholder` were nil this would return true

sampoder (approved):
LGTM!

PR #10138, opened 2025-04-10
Remove comments from grant card invite page

Luke-Oldenburg:
## Summary of the problem
The comments section would be shown underneath the invite for a grant card. It does not make sense to give a user access to comments if they have not yet accepted the invite.

## Describe your changes
I moved the comments to be inside of the grantee view and the admin view. Before it was just at the bottom of the file outside of any logic checking the grant status so it would always be showing.

### DIFF
--- a/app/views/card_grants/show.html.erb
+++ b/app/views/card_grants/show.html.erb
@@ -24,12 +24,6 @@
       </div>
     </div>
     <%= render partial: "spending_instructions", locals: { card_grant: @card_grant } %>
-    <% if policy(@card_grant.comments.build).create? %>
-      <h2>Comments</h2>
-      <%= render "comments/list", comments: @card_grant.comments %>
-      <%= render "comments/form", commentable: @card_grant %>
-    <% end %>
-
   <%# Organizers and admins' view %>
   <% else %>
     <% if @card_grant.stripe_card %>
@@ -53,11 +47,6 @@
       <% end %>
       <%= render partial: "canceled_warning", locals: { card_grant: @card_grant } unless @card_grant.stripe_card %>
     </div>
-    <% if policy(@card_grant.comments.build).create? %>
-      <h2>Comments</h2>
-      <%= render "comments/list", comments: @card_grant.comments %>
-      <%= render "comments/form", commentable: @card_grant %>
-    <% end %>
   <% end %>
 </div>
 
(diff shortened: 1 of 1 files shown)


app/views/card_grants/show.html.erb
```diff
@@ -57,9 +66,3 @@
 <% unless @card_grant.user == current_user && @card_grant.pending_invite? %>
   <%= render partial: "transactions", locals: { hcb_codes: @hcb_codes } %>
 <% end %>
-
-<h2>Comments</h2>
-<% if policy(@card_grant.comments.build).create? %>
-  <%= render "comments/list", comments: @card_grant.comments %>
-  <%= render "comments/form", commentable: @card_grant %>
-<% end %>
```
sampoder: I think the fix is actually to move the heading into the if block. Because the policy should make it organisers or admin-only already.

Luke-Oldenburg: I agree that may also be part of the issue but I believe it was showing the form as well on the invite page not just the heading.

sampoder: Hmm - can you send me a screen recording?

Luke-Oldenburg: I need to figure out how to issue card grants on dev mode unless you could send me another one on DevHacks that I'll just leave pending for my future testing.

sampoder: Head to `https://localhost:3000/flipper` and enable `card_grants_2023_05_25` for your event

Luke-Oldenburg:
@sampoder this is not fixed. Maybe the policy is wrong?
Screenshot From 2025-04-12 13-32-39

sampoder:
Are you an organiser on the event?

Luke-Oldenburg:
It happens for both members and managers.

sampoder:
That's expected behaviour

Luke-Oldenburg:
I was on a call with Gary and we decided it was strange to have that show if they have not yet accepted the invite.

sampoder:
There are a couple of times when this is needed, for example, providing
context about a card grant and why it was sent immediately after someone
sends one. It’s awkward here because you’re sending a grant to a user
already on the organisation - grantees don’t have access to comments by
default

Luke-Oldenburg:
To address the first part, isn't that what purpose is for? Not sure that it's visible on the invite though. That makes sense as to why the comments are showing though. I don't think I ever tested with sending a grant to somebody not in the organization. For those people the comments would be hidden correct?

sampoder:
That example would be if you didn’t want to grantee to see why - grantees
that aren’t organisers can’t see comments.

PR #10140, opened 2025-04-10
[Transfers] Slight tweaks to text on buttons in transfer money modal

Luke-Oldenburg:
All transfer modal buttons are now sentence case. Quiz item names make more sense now.


app/views/events/_transfer_form.html.erb
sampoder, line 1: Sorry to be pedantic but it should be the other way around and we should be using sentence case

Luke-Oldenburg: Oh whoops. Is Wire always capitalized or is that a mistake?

sampoder: it should be `International wire`

Luke-Oldenburg:
@sampoder should the capitalization also be changed here? https://github.com/hackclub/hcb/blob/main/app/views/events/transfers.html.erb

sampoder:
If it isn't sentence case yes

app/views/events/transfers.html.erb
sampoder, line 1: LGTM

app/views/events/_transfer_form.html.erb
```diff
@@ -20,19 +20,18 @@
...
             <strong>HCB transfer</strong>
-            <small>(between HCB accounts)</small>
+            <small>(between HCB organizations)</small>
           <% end %>
           <% if Flipper.enabled?(:card_grants_2023_05_25, @event) %>
             <%= link_to new_event_card_grant_path(@event), class: "text-decoration-none btn--form-option", data: { turbo: true } do %>
               <%= inline_icon "card", size: 28 %>
-              <strong>Grant</strong>
-              <small>A grant sent through an HCB card.</small>
+              <strong>Card grant</strong>
```
sampoder, line 34: Can we keep the additional context?

app/javascript/controllers/transfer_form_controller.js
sampoder, line 1: @manuthecoder can you help review this?

app/views/events/_transfer_form.html.erb
Luke-Oldenburg: Sure!

sampoder (approved):
LGTM - seems to be working

PR #10142, opened 2025-04-10
Add codespaces setup script

Luke-Oldenburg:
This PR adds a bash script that performs the final setup to prepare HCB for development on Codespaces. There was already a script made for Docker to make it a one liner to finish setup so I figured there should be an equivalent for Codespaces.

### DIFF
--- a/.devcontainer/boot.sh
+++ b/.devcontainer/boot.sh
@@ -1,4 +1,5 @@
 bundle install
+bin/rails db:migrate
 yarn install
 yarn build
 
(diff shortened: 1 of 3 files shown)


sampoder (commented):
Any reason we can't put this in https://github.com/hackclub/hcb/blob/main/.devcontainer/devcontainer.json?

Luke-Oldenburg:
I know things have broken before when being added to devcontainer but it's worth a shot.

sampoder (approved):
Got this working in codespaces - LGTM!

PR #10146, opened 2025-04-11
Remove Map to a PayPal transfer on Ledger page

leowilkin:
## Summary of the problem
We don't use PayPal anymore, so would be nice to de-bloat the ledger mapping page a tad more, and might even make it load _fractionally_ faster!

## Describe your changes
I just nuked the Map to a PayPal transfer bit - hopefully I haven't broken anything :)

### DIFF
--- a/app/views/admin/transaction.html.erb
+++ b/app/views/admin/transaction.html.erb
@@ -35,7 +35,6 @@
   <thead>
     <tr>
       <th>Map to an Event</th>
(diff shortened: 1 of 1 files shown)


leowilkin:
the failed check doesn't seem to be my fault

sampoder (approved):
LGTM - rest in peace, PayPal

PR #10148, opened 2025-04-12
Add monthly donation summary

### LINKED ISSUE #8323: [Recurring Donation] Disable `DonationMailer#donation_notification`. Replace with monthly summary email
The `DonationMailer#donation_notification` gets quite repetitive for monthly donations.

Let's disable them. However, we should still send them for the 1st round of a recurring donation.

Then, at the end of each month, we'll send a monthly summary email of all donations (including recurring donations)

polypixeldev:
## Summary of the problem

Previously, recurring donations sent an email to all organizers every month - this was a lot of emails and was disabled in #8407. However, organizers now do not get any notification of recurring donations after the initial donation.

Closes #8323 

## Describe your changes

I created a new EventMailer (as the monthly summaries don't really fit into the singular DonationMailer) with a monthly_donation_summary email. A new MonthlyDonationSummaryJob that is scheduled on the 1st of each month sends the summary email to the organizers of each event with at least 1 donation in the past month.

image

### DIFF
--- a/app/mailers/event_mailer.rb
+++ b/app/mailers/event_mailer.rb
@@ -8,6 +8,9 @@ class EventMailer < ApplicationMailer
     @event = params[:event]
 
     @donations = @event.donations.where(created_at: Time.now.last_month.beginning_of_month..).order(:created_at)
+
+    return if @donations.none?
+
     @total = @donations.sum(:amount)
 
     mail to: @emails, subject: "Monthly donation summary for #{@event.name}"
--- a/app/mailers/event_mailer.rb
+++ b/app/mailers/event_mailer.rb
@@ -13,7 +13,7 @@ class EventMailer < ApplicationMailer
 
     @total = @donations.sum(:amount)
 
-    mail to: @emails, subject: "Monthly donation summary for #{@event.name}"
+    mail to: @emails, subject: "#{@event.name} received #{@donations.length} donations this past month"
   end
 
   private
--- a/app/mailers/event_mailer.rb
+++ b/app/mailers/event_mailer.rb
@@ -7,8 +7,8 @@ class EventMailer < ApplicationMailer
   def monthly_donation_summary
     @event = params[:event]
 
-    @donations = @event.donations.select { |donation| donation.created_at >= 1.month.ago }.sort_by(&:created_at)
-    @total = @donations.reduce(0) { |sum, donation| sum + donation.amount }
+    @donations = @event.donations.where(created_at: Time.now.last_month.beginning_of_month..).order(:created_at)
+    @total = @donations.sum(:amount)
 
     mail to: @emails, subject: "Monthly donation summary for #{@event.name}"
   end
--- a/app/mailers/event_mailer.rb
+++ b/app/mailers/event_mailer.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+class EventMailer < ApplicationMailer
+  before_action { @event = params[:event] }
+  before_action :set_emails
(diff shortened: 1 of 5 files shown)


spec/mailers/previews/event_mailer_preview.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class EventMailerPreview < ActionMailer::Preview
+  def monthly_donation_summary
+    EventMailer.with(event: Event.not_demo_mode.first).monthly_donation_summary
```
sampoder:
```suggestion
    EventMailer.with(event: Donation.last.event).monthly_donation_summary
```
So there's a donation basically always?

app/views/event_mailer/monthly_donation_summary.html.erb
```diff
@@ -0,0 +1,24 @@
+<p><%= @emails.length > 1 ? "Hi all," : "Hi there," %></p>
+
+<p>
+  Here's your monthly donation summary for <%= @event.name %>:
+</p>
+
+<ul>
+  <% @donations.each do |donation| %>
+    <li>
+      <%= donation.name(show_anonymous: true) %> <%= donation.anonymous ? "anonymously" : "" %> donated <%= render_money donation.amount %>
```
sampoder:
```suggestion
      <%= donation.name(show_anonymous: true) %> <%= "anonymously" if donation.anonymous %> donated <%= render_money donation.amount %>
```

app/mailers/event_mailer.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class EventMailer < ApplicationMailer
+  before_action { @event = params[:event] }
+  before_action :set_emails
+
+  def monthly_donation_summary
+    @event = params[:event]
+
+    @donations = @event.donations.select { |donation| donation.created_at >= 1.month.ago }.sort_by(&:created_at)
```
sampoder:
```suggestion
    @donations = @event.donations.select { |donation| donation.created_at >= 1.month.ago }.sort_by(&:created_at)
    
    return if @donations.none?
    
```

sampoder:
```suggestion
    @donations = @event.donations.where(created_at: Time.now.last_month.beginning_of_month..).order(:created_at)
```
I rewrote this in SQL so we can sum for L11

app/mailers/event_mailer.rb
```diff
@@ -0,0 +1,23 @@
...
+    @total = @donations.reduce(0) { |sum, donation| sum + donation.amount }
```
sampoder:
```suggestion
    @total = @donations.sum(:amount)
```

app/mailers/event_mailer.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    mail to: @emails, subject: "Monthly donation summary for #{@event.name}"
```
sampoder: Maybe we could make it say "DevHacks received 2 donations this past month"

app/mailers/event_mailer.rb
sampoder, line 1: I feel like this belongs in `DonationMailer` but I'm not too fussed either way

polypixeldev: Yeah, the reason why I didn't put it in `DonationMailer` was because it seemed to be focused on sending emails about a single donation rather than the event as a whole

sampoder (approved):
LGTM! Thanks for building this

PR #10152, opened 2025-04-12
Fix inconsistent transfer page behavior for members

### LINKED ISSUE #10139: Inconsistent transfer behavior for members
### Describe your goals

I was visiting various pages to create transfers as a member. You should not be able to create a transfer as a member but the behavior across these pages was inconsistent and strange. Arguably, the "Send" button could be hidden altogether if you are not a manager.

### Describe the bug

ACH transfer- Returns to home page with the warning of "You are not authorized to perform this action."
Image
Mailed check- Page opens but all fields are disabled.
Image
International wire- Page opens but only certain fields are disabled.
Image
HCB transfer- Page opens but all fields are disabled.
Image
Card grant- A popover opens with the text "Content missing"
Image

### LINKED ISSUE #10149: [Transfers] No back button on card grant transfer modal
### Describe your goals

I was trying to send a transfer and clicked on card grant.

### Describe the bug

After closing the modal to select a separate transfer type, I noticed subsequent presses of the "Send" button would only display the card grant modal rather than the one to select transfer type. This behavior disappears after refreshing the page.
https://github.com/user-attachments/assets/0c1c6c20-f48f-4a49-b315-e418b1cbae7a

Luke-Oldenburg:
This PR makes it so all types of transfers have the same behavior when a member is viewing the UI to send them (fields and buttons are disabled). Before, there was an inconsistent mix of approaches to preventing members from sending transfers in the UI.

Closes #10139 and #10149


app/views/wires/new.html.erb
sampoder, line 1: Good catch - thank you!

app/views/events/_transfer_form.html.erb
```diff
@@ -29,7 +29,7 @@
             <small>(between HCB accounts)</small>
           <% end %>
           <% if Flipper.enabled?(:card_grants_2023_05_25, @event) %>
-            <%= link_to new_event_card_grant_path(@event), class: "text-decoration-none btn--form-option", data: { turbo: true } do %>
```
sampoder: I'd prefer to keep the in-page form if possible, it's nice how quick card grants can be made and we can add a tooltip for members

Luke-Oldenburg: Any idea on how to handle the content missing error and #10149 then? Making it a full page was just the easiest solution for me and I figured it would be okay given that there are other short full page forms as well such as disbursements.

sampoder: I think the changes to the policy should fix the content missing error. I think the fix is to look at how we're using JQuery modal and instead of having it just hide the modal, we have destroy the modal and then reopening it will recreate the initial state

Luke-Oldenburg: I reverted it back to a modal and the content missing error is gone. I'm not sure where the modal source is located to resolve the closing issue.

sampoder:
I added the unauthorised callback back to the ACH form because the other forms seems to have it. Can we add the disabled variable to the ACH transfer form?

Luke-Oldenburg:
@sampoder there was some weird behavior where it acted differently on dev mode vs production as to rendering the unauthorized callback which is why I had removed it. Might have been related to being signed in as an admin? I will test in a bit.

Luke-Oldenburg:
ACH seems to be working so I'll say it's good. I guess I gotta test it on production after it's merged to see how it behaves. There is a disabled on ACH btw (it's on the form). With your patch to the card grant modal it sometimes can take a bit to open the next time or miss a press of the Send button.

sampoder:
> ACH seems to be working so I'll say it's good. I guess I gotta test it on production after it's merged to see how it behaves. There is a disabled on ACH btw (it's on the form). With your patch to the card grant modal it sometimes can take a bit to open the next time or miss a press of the Send button.

Sounds good re: ACH - sorry I missed that!

I'm going to merge this in if the card grant stuff is bad in production, I can play around with it a bit more. A screen recording would be helpful if you have one.

sampoder (approved):

PR #10155, opened 2025-04-13
Give Stagey McStageface a phone number

polypixeldev:
## Summary of the problem

Currently, the default user created during seeding does not have a phone number - this causes issues when trying to create a cardholder with this user.

## Describe your changes

Added a default phone number (the Hack Club number) to the default user.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -6,7 +6,7 @@ email = Rails.env.staging? ? "staging@bank.engineering" : "admin@bank.engineerin
 
 if user.nil?
   puts "Woah there, there aren't any users! Creating an user (#{email})."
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -6,7 +6,7 @@
 
 if user.nil?
   puts "Woah there, there aren't any users! Creating an user (#{email})."
-  user = User.create!(email:, full_name: "Stagey McStageface")
+  user = User.create!(email:, full_name: "Stagey McStageface", phone_number: "+18556254225")
```
sampoder:
```suggestion
  user = User.create!(email:, full_name: "Stagey McStageface", phone_number: "+19064225632")
```
Just so we have a way of accessing messages if needed

sampoder (approved):
LGTM!

PR #10156, opened 2025-04-13
Add one time job to cancel the stripe cards for old returned and expired grants

### LINKED ISSUE #9077: [v4 API][Card Grants] Update pundit policy to disallow freezing/unfreezing a card when card grant is cancelled
### Describe your goals

@thedev132 was viewing a cancelled grant card on HCB mobile. He noticed that there was buttons to freeze and unfreeze it because he was a manager of the grant.

### Describe the bug

Despite the buttons not being available on the website, after pressing, they reactivated the grant card after it had been cancelled. It still displayed that the grant had been cancelled but the stripe card was active again.

Luke-Oldenburg:
When card grants were returned or expired in the past, their associated stripe cards would be frozen rather than canceled. This has since been changed but those cards still remain frozen. This job will go through and cancel all cards associated with returned or expired grants that have not already been canceled.

Closes #9077

### DIFF
--- a/app/jobs/one_time_jobs/cancel_legacy_returned_grants.rb
+++ b/app/jobs/one_time_jobs/cancel_legacy_returned_grants.rb
@@ -5,7 +5,7 @@ module OneTimeJobs
     def self.perform()
       CardGrant.where(status: :canceled).or(CardGrant.where(status: :expired)).find_each do |card_grant|
         @card = StripeCard.find(card_grant.stripe_card_id)
-        if !@card.canceled?
+        unless @card.canceled?
           @card.cancel!
         end
       end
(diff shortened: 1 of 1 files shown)


app/jobs/one_time_jobs/cancel_legacy_returned_grants.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module OneTimeJobs
+  class CancelLegacyReturnedGrants
+    def self.perform()
+      CardGrant.where(status: :canceled).or(CardGrant.where(status: :expired)).find_each do |card_grant|
+        @card = StripeCard.find(card_grant.stripe_card_id)
+        if !@card.canceled?
```
sampoder:
```suggestion
        unless @card.canceled?
```
handy thing

sampoder (approved):
LGTM - will run in production

PR #10157, opened 2025-04-13
[Transactions] Improve consistency in views

Luke-Oldenburg:
Before, the headings for transactions did not follow a consistent structure and also had differences in their text formatting. This PR aims to address those issues and make the headings more consistent and also set a standard for text style (names get none, background text is "regular muted", and money is "regular").


YodaLightsabr:
Thanks for working on this! It'll be nice to have some consistency going forward. I have a few suggestions though:
- For card charges, we're now listing the spender twice. I think this is a bit redundant, and it takes a little more effort to see where the card was spent now
image
- I noticed we're also removing the rename button; what if we added it to every type to keep consistency?

Luke-Oldenburg:
I agree about card charges. I'll go ahead and remove the name from the title. As for the edit button, I tested adding that to all of them and found it to look a bit ugly. It's also redundant as you can already edit it by double clicking the memo in the body(for transactions that have the memo listed as a field in the body) or shift-clicking from the transaction list(for all transactions). See #9007 and #9860 for more info.

sampoder:
I think I'd prefer having an edit icon on each of them? Rendering things like "Transferring Max Kern $5,000,000" if it hasn't been edited but then the actual memo if it has

Luke-Oldenburg:
I see what you mean although personally I feel like the prettified name is more fitting in the heading there as the transaction memo is used to identify it in other places.

YodaLightsabr:
The edit button on the transaction page is especially important to keep around because it's the only way to edit the transaction that is clearly defined in the UI. Shift clicking on the transaction name is faster, but it's not intuitively accessible by a button.

app/views/card_grants/_details.html.erb
```diff
@@ -1,76 +1,93 @@
-<%# locals: (card_grant:) %>
```
sampoder, line 1: Generally it's best to use locals in partials and avoid instance variables: https://andycroll.com/ruby/dont-use-instance-variables-in-partials/. Yes I know our codebase commits this sin many times but I don't want to make it worse.

app/views/card_grants/show.html.erb
```diff
@@ -38,7 +38,7 @@
     <% end %>
 
     <div class="flex flex-col g2">
-      <%= render partial: "details", locals: { card_grant: @card_grant } %>
```
sampoder: See comment re instance variables in partials

app/views/hcb_codes/_wire.html.erb
```diff
@@ -95,5 +106,3 @@
     </div>
   <% end %>
 </article>
-
-<% title "#{Money.from_cents(@hcb_code.wire.amount_cents, @hcb_code.wire.currency).format} wire to #{@hcb_code.wire.recipient_name}" %>
```
sampoder, line 99: Where / what are these titles being replaced with?

sampoder: I see it - thanks

app/views/card_grants/_details.html.erb
```diff
@@ -1,76 +1,93 @@
...
-    <%= pop_icon_to "edit",
-        [:edit, card_grant.disbursement.local_hcb_code],
-        class: "mr2 align-middle", "aria-label": "Rename transaction" %>
-    Grant to &nbsp; <%= user_mention card_grant.user %>
+  <h2 class="h2 mt0 mx0 border-none flex items-center justify-between" style="gap: 8px">
+    <%= turbo_frame_tag @hcb_code, class: "flex items-center" do %>
+      <%= pop_icon_to "edit",
+          edit_hcb_code_path(@hcb_code),
+          class: "mr2 align-middle tooltipped tooltipped--e", "aria-label": "Rename transaction",
+          data: { turbo: true } %>
```
sampoder: We're using this code a lot - can we put it in a partial?

app/views/hcb_codes/_stripe_card.html.erb
```diff
@@ -62,12 +42,10 @@
             <%= inline_icon :info, size: 20 %>
           </span>
         <% elsif @hcb_code.pt.declined? %>
-          <action data-behavior="modal_trigger" data-modal="declined_information_<%= instance %>" class="pointer" tabindex="0">
-            <span class="badge h4 md-right bg-accent">
```
sampoder: How come the modal is being removed?

Luke-Oldenburg: Honestly not sure. What is it for?

app/models/hcb_code.rb
```diff
@@ -133,6 +133,11 @@ def amount_cents
     end
   end
 
+  def display_amount
+    return disbursement.amount.abs if disbursement?
+    return amount.abs
+  end
```
sampoder: I believe we should be using `amount_cents_by_event` instead of creating a new method

app/views/hcb_codes/edit.html.erb
```diff
@@ -25,7 +25,7 @@
                             style: "border-color: #338eda; min-height: 0; font-size: 1.5rem; padding: 0.125rem 0 0.125rem 4px; letter-spacing: -0.02em" %>
 
         <span class="regular muted ml1 mr1"> <%= @hcb_code.stripe_refund? ? "refunded" : "for" %> </span>
-        <span class="mr3"><%= number_to_currency(@hcb_code.amount.abs) %></span>
+        <span class="mr3 regular"><%= number_to_currency(@hcb_code.display_amount / 100) %></span>
```
sampoder: What are you trying to achieve with this code and why does it need to be `/ 100`? There's probably a different combinations of helpers we can use to avoid dividing by 100

Luke-Oldenburg: It fixes the disbursement amount appearing as 0. There might be a way I can swap it to amount cents somewhere but Ive ran into scenarios in the past where I needed /100.

app/views/hcb_codes/_stripe_card.html.erb
sampoder: It shows why a card transaction was declined

app/views/hcb_codes/edit.html.erb
sampoder: https://github.com/hackclub/hcb/pull/10157#discussion_r2082249515

app/views/hcb_codes/_stripe_card.html.erb
Luke-Oldenburg: Think I accidentally deleted it while refactoring.

app/views/hcb_codes/edit.html.erb
Luke-Oldenburg: I tried what you wrote here and it worked for regular disbursements but card grants are appearing as zero in both their regular title and edit partial title.

app/views/card_grants/_details.html.erb
Luke-Oldenburg: Great idea!

app/views/hcb_codes/edit.html.erb
Luke-Oldenburg: I just noticed a bug where the amount gets divided by 100 in the edit partial. I'm going to keep playing around with stuff to try and fix it.

Luke-Oldenburg: Well that's gone randomly now so I'm back to where I started with the display_amount function. It works but I'm not sure how to clean it up like you described.

sampoder: I think you can use `amount_cents_by_event` but edit it slightly so it has `subledger_id: nil` in the SQL queries

app/views/hcb_codes/_ach_transfer.html.erb
```diff
@@ -5,29 +5,19 @@
...
-        <%= inline_icon @hcb_code.ach_transfer.state_icon, size: 20 if @hcb_code.ach_transfer.state_icon %>
-      </span>
-
-      <%= render "hcb_codes/memo_stream", hcb_code: @hcb_code %>
-      <span style="flex-grow: 1;"></span>
-      <%= render partial: "hcb_codes/ai_memo", locals: { hcb_code: @hcb_code } %>
-    <% end %>
-    <%= render partial: "hcb_codes/meatballs", locals: { hcb_code: @hcb_code } %>
-  </h2>
+  <% content_for :heading_content do %>
```
sampoder: Smart! This can just be `content_for :heading` I think

Luke-Oldenburg: It gave me a syntax error when I didn't have `do`.

Luke-Oldenburg: oh wait you meant rename that. will do.

sampoder: Oh yeah I meant to rename `:heading_content` to `:heading`

app/views/hcb_codes/edit.html.erb
Luke-Oldenburg: I'm still lost. I don't see subledger_id on any of those models. How do I do that?

Luke-Oldenburg: It was on the event mapping :smile:. Thanks for the advice! It worked!

Luke-Oldenburg:
Getting `'nil' is not an ActiveModel-compatible object. It must implement #to_partial_path.` on all transaction popovers.

Luke-Oldenburg:
> Getting `'nil' is not an ActiveModel-compatible object. It must implement #to_partial_path.` on all transaction popovers.

This was caused by explicity rendering a partial as `partial:` for some reason.

app/views/card_grants/_details.html.erb
```diff
@@ -1,11 +1,19 @@
 <%# locals: (card_grant:) %>
+<% hcb_code ||= card_grant.disbursement.local_hcb_code %>
```
davidcornu: **suggestion**: Do we ever have another `hcb_code`?

```suggestion
<% hcb_code = card_grant.disbursement.local_hcb_code %>
```

davidcornu (approved):

PR #10158, opened 2025-04-13
Fix order of permissions checking on card grants

Luke-Oldenburg:
## Summary of the problem
When organizers visited card grants they had received from the same organization they were an organizer for, they would receive the lower permission view instead of the higher permission view.

## Describe your changes
I modified the code to check for organizer or admin status before grantee status.


sampoder (approved):
Sure - let's run with this

PR #10159, opened 2025-04-13
Don't include OS version in fingerprint if it doesn't exist

polypixeldev:
## Summary of the problem

On Linux, there isn't an OS version included in the user agent string. This caused logins from Linux machines to show as coming from "Linux undefined".

## Describe your changes

`fingerprint.js` now only includes the OS version in the fingerprint if it is defined.


sampoder (approved):
LGTM!

PR #10160, opened 2025-04-13
Only topup stripe if it will be >=5k

### LINKED ISSUE #10147: [Stripe Issuing Balance Topups] Only do a topup if `topup_amount` >= 5k
Stripe issuing balance topups are creating a lot of noise in our bank statements. Since they already aren't one-to-one with transactions, let's only do a topup when the `topup_amount` is greater than $5k

polypixeldev:
## Summary of the problem

There's a lot of noise in bank statements from a lot of small Stripe topups.

Closes #10147 

## Describe your changes

Modifies the topup job to return if the amount is less than 5k.


sampoder (approved):
LGTM - thanks!

PR #10161, opened 2025-04-13
[DX] Add flippers to dev-docs

Luke-Oldenburg:
Added a full list of flippers to dev-docs alongside instructions on how to use Flipper.


dev-docs/development.md
```diff
@@ -193,3 +193,6 @@ We've transitioned to using development keys and seed data in development, but h
 - Override the `LOCKBOX`, `ACTIVE_RECORD__ENCRYPTION__DETERMINISTIC_KEY`, `ACTIVE_RECORD__ENCRYPTION__KEY_DERIVATION_SALT`, and `ACTIVE_RECORD__ENCRYPTION__PRIMARY_KEY` secrets by defining them in `.env.development`. Use the values from the [`production` enviroment in Doppler](https://dashboard.doppler.com/workplace/2818669764d639172564/projects/hcb/configs/production).
 
 - Run the [docker_setup.sh](https://github.com/hackclub/hcb/docker_setup.sh) script to set up a local environment with Docker. The script will use a dump of our production database from Heroku.
+
+## Flipper
+Flipper is used to toggle certain features on HCB. Flipper can be accessed at [http://127.0.0.1:3000/flipper/features](http://127.0.0.1:3000/flipper/features). To enable a feature, press "Add Feature", paste in the name of a feature from [this list](flippers.txt), and then press "Fully Enable".
```
sampoder:
```suggestion
[Flipper](https://github.com/flippercloud/flipper) is used to toggle certain features on HCB. Flipper can be accessed at [http://localhost:3000/flipper/features](http://localhost:3000/flipper/features). To enable a feature, press "Add Feature", paste in the name of a feature from [this list](flippers.txt), and then press "Fully Enable".
```

dev-docs/flippers.txt
sampoder, line 1: Hmm. I'm worried that we won't keep this up to date, we have https://github.com/hackclub/hcb/blob/d7b0af1504c5b7bdc0b267f1547542da1d97f54a/app/controllers/features_controller.rb#L4 which is the public facing list of Flipper flags but it doesn't include card grants and a couple others.

Luke-Oldenburg: I wasn't aware that existed! Would it be possible to add the missing flags and link to that file instead?

sampoder: Not necessarily because there are some Flippers that shouldn't be whitelisted for everyone to enable

Luke-Oldenburg: A little jank but what about an unused array of the others next to it.

sampoder: We can point people to hcb.hackclub.com/api/flags https://github.com/hackclub/hcb/pull/10176

sampoder (approved):
LGTM!

PR #10166, opened 2025-04-14
[Memos] Standardize sentence case

Luke-Oldenburg:
All memos now use sentence case.


sampoder:
There's been a bit of back and forward about this in the past but I think the consensus we landed on was that they should be sentence case - which ones are still uppercase?

Luke-Oldenburg:
I know ACH and wires are as well as I think a few others. I'll make those all sentence case then.

sampoder:
Sounds good!

sampoder (approved):
LGTM!

PR #10167, opened 2025-04-14
[Card Grants] Check if card grant is already canceled before allowing it to be returned

Luke-Oldenburg:
## Summary of the problem
If a card grant was canceled by multiple people at the same time, it would return the money to the organization multiple times resulting in money being created for the organization out of thin air.

## Describe your changes
I made the card grant policy check if it was already canceled before allowing it to be returned.

### DIFF
--- a/app/policies/card_grant_policy.rb
+++ b/app/policies/card_grant_policy.rb
@@ -22,7 +22,7 @@ class CardGrantPolicy < ApplicationPolicy
   end
 
   def cancel?
-    admin_or_user || record.user == user
+    (admin_or_user || record.user == user) && !record.canceled?
   end
 
   def topup?


sampoder:
We should already be doing this in the `cancel!` method on the model - can you link me the card where this happened? Want to track down the order of events

Luke-Oldenburg:
I believe it is https://hcb.hackclub.com/grants/yWIdl3. Mohamad and Gary were discussing this a few days ago.

sampoder:
Wow - these were at essentially the same time. I don't think this PR would have fixed this because the two requests would have already been authorised?

Screenshot 2025-04-14 at 5 16 16 PM


Screenshot 2025-04-14 at 5 16 12 PM

sampoder:
I appreciate you making this - but honestly I'm struggling to see a good fix. Maybe we can do the zero-ing outside of the return so there's a delay before we zero?

Luke-Oldenburg:
Where should that go instead?

sampoder:
We have the ZeroJob that does this rn (but it happens a week later): https://github.com/hackclub/hcb/blob/25a8002d493063c4b9b0cb5e45c82bfb629d04b1/app/jobs/card_grant/zero_job.rb#L4

Luke-Oldenburg:
Is it possible to run that sooner?

sampoder:
It was buggy so we added the delay (we thought unsettled transactions were causing the bug?)

Luke-Oldenburg:
What if we made zero job nightly and then removed zeroing from the return? Only other idea that I have that is kind of jank is waiting for a random time offset.

PR #10175, opened 2025-04-15
Truncate long transfer text w/ CSS

sampoder:
FYI @Luke-Oldenburg


Luke-Oldenburg:
What was the @ for? Am I missing something lol

Luke-Oldenburg:
Ohhh I see now

PR #10177, opened 2025-04-15
Properly namespace jobs with aliases

### LINKED ISSUE #8317: Fix incorrect job namespacing

polypixeldev:
## Summary of the problem

Jobs are currently incorrectly namespaced!

Closes #8317

## Describe your changes

Went through the incorrect jobs and renamed them to fit the rules. Also added aliases to the old names in case they are currently in Redis - these should be able to be removed in a month when the old names are gone, as per the [Sidekiq FAQ](https://github.com/sidekiq/sidekiq/wiki/FAQ#how-do-i-safely-rename-a-worker)


sampoder (commented):
We probably have to replace the references to each of these jobs as well - thank you for getting this going!

sampoder (approved):
LGTM - thank you!

PR #10179, opened 2025-04-15
Remove expired promos from command bar and the airtable checky thing

leowilkin:
## Summary of the problem
We don't do the wallet thing anymore, so why have it in the admin command bar?

see also: stickermule, github grants, replit, etc.

There was also a dead reference to /admin/grants/ that I nuked as well :)

## Describe your changes
I found the file that controls the command bar stuff, and i deleted the offending files.

RIP wallets


sampoder (approved):
LGTM!

PR #8341, opened 2024-10-28
Create `exempt_from_wire_minimum?` on `Event::Plan`

sampoder:
WIP

PR #10180, opened 2025-04-15
[Disbursements] Send anywhere by slug for HackClubAffiliates

Luke-Oldenburg:
I was discussing this with @YodaLightsabr and he advised it would be smarter to use a system based on slugs than the organization picker he is building (#8591). There probably needs to be some type of tooltip added explaining how to get an organization's slug but I'm not sure the best way to implement it design wise.
Screenshot From 2025-04-15 15-36-00
Screenshot From 2025-04-15 15-33-21


app/policies/disbursement_policy.rb
```diff
@@ -7,16 +7,13 @@ def show?
 
   def new?
     user&.admin? || (
-      (record.destination_event.nil? || record.destination_event.users.include?(user)) &&
+      record.destination_event.nil? &&
```
sampoder:
```suggestion
      record.destination_event.nil? &&
```
I don't think you want to be only approving disbursements with `nil` `destination_event`s

sampoder:
Is the intention of this PR to allow people to make disbursements to any event - not just the ones they're a part of?

Luke-Oldenburg:
Yes

sampoder:
I think we should direct people to use donations - especially if the organisation isn't transparent. That way people get a receipt, the organisation gets notified etc.

sampoder:
I guess the one exception is coming from HQ organisations because they often have to go to clubs etc. that people running YSWSs aren't a part of

sampoder:
One thing I had thought about was enabling HQ organisations to send disbursements anywhere based on a flag in `Event::Plan`

Luke-Oldenburg:
> One thing I had thought about was enabling HQ organisations to send disbursements anywhere based on a flag in `Event::Plan`

So only use the new UI I made for if it's an HQ organization?

Luke-Oldenburg:
> I think we should direct people to use donations - especially if the organisation isn't transparent. That way people get a receipt, the organisation gets notified etc.

Thought of that but it wastes money on credit card processing fees for no reason.

Luke-Oldenburg:
> I guess the one exception is coming from HQ organisations because they often have to go to clubs etc. that people running YSWSs aren't a part of

In that case wouldn't it typically be an admin sending money and they already had the ability to send anywhere (or at least I thought)?

app/policies/disbursement_policy.rb
Luke-Oldenburg: I was very confused by the nil checks but it worked when I tested it so I left it.

sampoder: Were you signed in as an admin?

Luke-Oldenburg: I might've been.

sampoder:
> So only use the new UI I made for if it's an HQ organization?

That's what I was thinking

> Thought of that but it wastes money on credit card processing fees for no reason.

Yeah - personally I'm comfortable with that though

> In that case wouldn't it typically be an admin sending money and they already had the ability to send anywhere (or at least I thought)?

Normally that's what people do but it's a bit of a pain to have people continuously asking for transfers (not their fault tbf). It'd be nice to smoothen that process.

Luke-Oldenburg:
I removed the from field to make the disbursement page more similar to the other transfer pages in the sense of it being event-specific.

app/controllers/disbursements_controller.rb
```diff
@@ -50,18 +50,18 @@ def new
                              else
                                current_user.events.not_hidden.filter_demo_mode(false)
                              end
-    @allowed_destination_events = if current_user.admin?
-                                    Event.all.reorder(Event::CUSTOM_SORT).includes(:plan)
-                                  else
-                                    current_user.events.not_hidden.without(@source_event).filter_demo_mode(false)
-                                  end
+    @allowed_destination_events = if @source_event.plan.id == 4
```
sampoder: Each `Event::Plan` should have a method like `unrestricted_disbursements_allowed?`. Standard can have false and then the HQ plan can override that

https://github.com/hackclub/hcb/pull/8341 is an example of this

app/policies/disbursement_policy.rb
sampoder: This policy still needs to be updated so that it's based on the source event's plan

app/views/disbursements/_form.html.erb
```diff
@@ -3,20 +3,12 @@
...
-    <%= form.label :source_event_id, "From" %>
-    <%= form.collection_select(:source_event_id,
-                                @allowed_source_events,
-                                :id,
-                                current_user.admin? ? :admin_dropdown_description : :disbursement_dropdown_description,
-                                { prompt: "Select one…", disabled: ->(e) { e != @source_event && !policy(e).create_transfer? || (!current_user.admin? && e.balance_available <= 0) } },
-                                required: true,
-                                disabled: ) %>
-  </div>
-
```
sampoder: Can we keep this? I'm not sure why it's being removed - admins also often go to https://hcb.hackclub.com/disbursements/new

sampoder (changes requested):

app/views/disbursements/_form.html.erb
```diff
@@ -3,20 +3,12 @@
...
   <div class="field event-select-target">
     <%= form.label :event_id, "To" %>
-    <%= form.select(:event_id,
+    <% if @source_event.plan.id == 4 %>
+      <%= form.text_field :event_id, placeholder: "hack_the_seas", required: true, disabled: %>
+    <% else %>
+      <%= form.select(:event_id,
```
sampoder: Imo we should keep the drop down and then have an "Other" option that then shows the text field

app/views/disbursements/_form.html.erb
```diff
@@ -3,20 +3,12 @@
...
-                                current_user.admin? ? :admin_dropdown_description : :disbursement_dropdown_description,
-                                { prompt: "Select one…", disabled: ->(e) { e != @source_event && !policy(e).create_transfer? || (!current_user.admin? && e.balance_available <= 0) } },
-                                required: true,
-                                disabled: ) %>
-  </div>
-
   <div class="field event-select-target">
     <%= form.label :event_id, "To" %>
-    <%= form.select(:event_id,
+    <% if @source_event.plan.id == 4 %>
```
sampoder: FYI every event has it's own `Event::Plan` so the ids will differ

app/views/disbursements/_form.html.erb
Luke-Oldenburg: Ah yea that was me being dumb

app/controllers/disbursements_controller.rb
Luke-Oldenburg: Ah I didn't understand how plans worked at first but it makes sense now.

app/views/disbursements/_form.html.erb
Luke-Oldenburg: Ah I thought id was the index of it in some list storing all of them.

app/models/event/plan/argosy_2024.rb
```diff
@@ -31,6 +31,10 @@ def features
         super - %w[promotions google_workspace]
       end
 
+      def unrestricted_disbursements_allowed?
+        false
+      end
+
```
sampoder: You can inherit from standard - we don't need to explicitly define it on every plan

Luke-Oldenburg: I didn't realize those inherited at first. Will fix.

app/policies/disbursement_policy.rb
Luke-Oldenburg: I think I just fixed what you said but not sure. I don't fully understand what the nil checks are for still.

app/policies/disbursement_policy.rb
```diff
@@ -8,7 +8,7 @@ def show?
   def new?
     user&.admin? || (
       record.destination_event.nil? &&
```
sampoder: We shouldn't only be approving disbursements with `nil` destination events

app/controllers/disbursements_controller.rb
```diff
@@ -50,18 +50,18 @@ def new
                              else
                                current_user.events.not_hidden.filter_demo_mode(false)
                              end
-    @allowed_destination_events = if current_user.admin?
-                                    Event.all.reorder(Event::CUSTOM_SORT).includes(:plan)
-                                  else
-                                    current_user.events.not_hidden.without(@source_event).filter_demo_mode(false)
-                                  end
+    @allowed_destination_events = if @source_event.plan.unrestricted_disbursements_allowed?
```
sampoder:
```suggestion
    @allowed_destination_events = if @source_event.plan.unrestricted_disbursements_allowed? || current_user.admin?
```

Luke-Oldenburg:
lgtm!

app/policies/disbursement_policy.rb
```diff
@@ -5,18 +5,29 @@ def show?
     user.auditor?
   end
 
+  def can_send?
+    return true if user&.admin?
+    return true if record.source_event.nil?
+    return true if record.source_event.users.include?(user)
```
garyhtou: Does this allow readers (in source event) to create HCB transfers?

Luke-Oldenburg: On another note, should new and create both have the same policy?

Luke-Oldenburg:
tested :+1:

sampoder (approved):
LGTM!

PR #10181, opened 2025-04-16
Convert options positional arguments to keyword arguments

### LINKED ISSUE #8292: [Refactor] Convert `options` positional arguments to keyword arguments
In a lot of our helpers, we have an `options` positional argument. Here's a (fake) example of what that means:

```ruby
# This method has two positional arguments.
# 1. The first argument (data) is required; it has no default value.
# 2. The second argument has a default value. The value is a hash.
def render_chart(data, options = {color: "blue", height: 100, width: 200})
  # do something (like render the chart... lol)
  puts options # print options so we can see what they are
end
```

So, what's the problem?

```ruby
# This way to defining options work well if you're just using the default options
render_chart([1, 2, 3, 4])
#=> {color: "blue", height: 100, width: 200}
# ^ yay, those are the default options defined in the method's parameters

# However, it breaks unintuitively when you try to pass in options. Say we want the usual default options, but want to override the color to be red.
render_chart([1, 2, 3, 4], {color: "red"})
#=> {color: "red"}
# ^ wait... what happened to the rest of the default options?! Since the second argument is just seen as a single hash in Ruby, passing in a second argument (our `{color: "red"}`) completely overrides the argument's default value.
```

**So, how do we fix this?**

Use [keyword arguments](https://noelrappin.com/blog/2024/02/better-know-keyword-arguments/).

```ruby
def render_chart(data, color: "blue", height: 100, width: 200, **options)
  # do something (like render the chart... lol)
  puts options.merge(color:, height:, width:) # print options so we can see what they are
end

# Calling the method without passing any options remains the same
render_color([1, 2, 3, 4])
#=> {color: "blue", height: 100, width: 200}
# ^ those are the default options

# Color the method with options will still keep other defaults; unlike the positional argument `options` above.
render_color([1, 2, 3, 4], color: "red')
#=> {color: "red", height: 100, width: 200}
# ^ we get color red and all the usual defaults.
```

Problem solved!

### Migration

Notice how the syntax for calling the two types of definitions is different? When `options` is a positional argument, we must explicitly pass in a hash. Keyword arguments doesn't need the curly braces (`{ }`). Also, some calls with `options` may duplicate the default values. For example:

```ruby
def render_chart(data, options = {color: "blue", height: 100, width: 200})
  # do something
end

render_chart([1, 2, 3, 4], {color: "red", height: 100, width: 200})
# we changed the color to red, but still need to pass in the default height & width
```

So, keep these in mind when migrating method from using positional options to keyword optoins.

image

polypixeldev:
## Summary of the problem

Using `options` positional arguments causes issues when used with default values as one provided value overrides the rest. Keyword arguments don't have this issue and don't require curly braces.

Closes #8292

## Describe your changes

Went through all the helpers with `options` positional arguments, changed them to use keyword arguments, and updated where they are used. One exception was in commit `c599344` - see commit body for details.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -58,23 +58,23 @@ module ApplicationHelper
     content_tag(:span, content.join.html_safe)
   end
 
-  def blankslate(text, options = {})
+  def blankslate(text, **options)
     other_options = options.except(:class)
     content_tag(:p, text, class: "center mt0 mb0 pt4 pb4 slate bold h3 mx-auto rounded-lg border #{options[:class]}", **other_options)
   end
 
-  def list_badge_for(count, item, glyph, options = { optional: false, required: false })
-    return nil if options[:optional] && count == 0
+  def list_badge_for(count, item, glyph, optional: false, required: false, **options)
+    return nil if optional && count == 0
 
     icon = inline_icon(glyph, size: 20, 'aria-hidden': true)
 
     content_tag(:span,
                 icon + count.to_s,
                 'aria-label': pluralize(count, item),
-                class: "list-badge tooltipped tooltipped--w #{options[:required] && count == 0 ? 'b--warning warning' : ''} #{options[:class]}")
+                class: "list-badge tooltipped tooltipped--w #{required && count == 0 ? 'b--warning warning' : ''} #{options[:class]}")
   end
 
-  def badge_for(value, options = {})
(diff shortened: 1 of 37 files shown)


sampoder (approved):
LGTM!

garyhtou:
Oh this is amazing @polypixeldev 🙏. I've been wanting this for a very long time

PR #10184, opened 2025-04-16
Fix `Statsd is not a module (TypeError)`

garyhtou:
## Summary of the problem

```
Statsd is not a module (TypeError)
```

### DIFF
--- a/app/jobs/user/statsd_gauge_online_user_count_job.rb
+++ b/app/jobs/user/statsd_gauge_online_user_count_job.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-class Statsd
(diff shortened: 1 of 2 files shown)


app/jobs/user/statsd_gauge_online_user_count_job.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-module Statsd
-  class GaugeOnlineUserCountJob < ApplicationJob
+class User
+  class StatsdGaugeOnlineUserCountJob < ApplicationJob
```
garyhtou, line 4: not a great name, but i'm getting this in for the sake of time since sidekiq jobs aren't running due to this error

PR #10206, opened 2025-04-18
Fix `company_name` to "HCB (Hack Club)"

sampoder:


app/models/ach_transfer.rb
```diff
@@ -173,7 +173,7 @@ def payment_recipient_attributes
   before_validation { self.recipient_name = recipient_name.presence&.strip }
 
   before_validation do
-    self.company_name = event.short_name if company_name.blank?
+    self.company_name = "HCB (Hack Club)" # https://hackclub.slack.com/archives/C065957QPC7/p1745005224201699
```
garyhtou, line 176: Please add a comment instead of linking to Slack. Column's slack messages have a data retention policy, so this message be unviewable soon.

PR #10208, opened 2025-04-20
Add setting for monthly donation summaries

polypixeldev:
## Summary of the problem

Following #10148, all users are sent an email each month for each organization they are a member of that has received a donation, without any option to opt-out of these emails.

## Describe your changes

Added a setting to the notification settings that allows users to enable or disable being sent these emails.
image

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -357,6 +357,7 @@ class UsersController < ApplicationController
       :sessions_reported,
       :session_duration_seconds,
       :receipt_report_option,
+      :donation_summary_option,
       :birthday,
       :seasonal_themes_enabled,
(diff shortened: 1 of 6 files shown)


sampoder:
Hmm - long term we probably want to build this into https://github.com/hackclub/hcb/issues/5367 - are people asking for this? We can merge it in short-term if people are but otherwise I'm chill with us sending it to everyone.

polypixeldev:
Hasn't been asked for (no donation summaries have been sent out yet though), just figured it would be good to have it. Building it into #5367 would be good though

sampoder:
We've pushed these emails out a couple of times now and people seem positive about them. I think it's alright if we don't make them configurable until the big notifications rehaul

PR #10212, opened 2025-04-20
[V4 API] Invoices & Sponsors

Luke-Oldenburg:
Adds the ability to create and view invoices and sponsors.


app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+module Api
+  module V4
+    class InvoicesController < ApplicationController
```
YodaLightsabr, line 5: Since we're adding the ability to create and load single invoices, we should also add the ability to list invoices. If we implement it similarly to Stripe cards, we can have a route for indexing by organization as well as by user:
https://github.com/hackclub/hcb/blob/70685d707103fe83fde631f9b5fc3a1803a2b271/app/controllers/api/v4/stripe_cards_controller.rb#L6-L15

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      def index
+        if params[:event_id].present?
+          @event = authorize(Event.find_by_public_id(params[:organization_id]) || Event.friendly.find(params[:organization_id]), :is_public)
```
garyhtou: Create a new policy for index

app/policies/api/invoice_policy.rb
```diff
@@ -6,6 +6,14 @@ def show?
       record.event.is_public?
```
garyhtou: Check to see if the user is at least a reader

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+          :item_amount,
+          :sponsor_id,
+          sponsor_attributes: policy(Sponsor).permitted_attributes
+        )
+    
+        sponsor_attrs = filtered_params[:sponsor_attributes]
+    
+        due_date = Date.civil(filtered_params["due_date(1i)"].to_i,
+                              filtered_params["due_date(2i)"].to_i,
+                              filtered_params["due_date(3i)"].to_i)
```
garyhtou: Let's take in a timestamp instead of multiple params.


```ruby
params["due_date"].to_datetime
```

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+          sponsor_address_line2: sponsor_attrs[:address_line2],
+          sponsor_address_city: sponsor_attrs[:address_city],
+          sponsor_address_state: sponsor_attrs[:address_state],
+          sponsor_address_postal_code: sponsor_attrs[:address_postal_code],
+          sponsor_address_country: sponsor_attrs[:address_country]
+        ).run
+    
+        unless OrganizerPosition.find_by(user: @invoice.creator, event: @event)&.manager?
+          InvoiceMailer.with(invoice: @invoice).notify_organizers_sent.deliver_later
+        end
```
garyhtou: Let's move this logic into the model as a `after_create_commit` callback.

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    
+          sponsor_id: sponsor_attrs[:id],
+          sponsor_name: sponsor_attrs[:name],
+          sponsor_email: sponsor_attrs[:contact_email],
+          sponsor_address_line1: sponsor_attrs[:address_line1],
+          sponsor_address_line2: sponsor_attrs[:address_line2],
+          sponsor_address_city: sponsor_attrs[:address_city],
+          sponsor_address_state: sponsor_attrs[:address_state],
+          sponsor_address_postal_code: sponsor_attrs[:address_postal_code],
+          sponsor_address_country: sponsor_attrs[:address_country]
```
garyhtou: Let's create a Sponsor controller for creating/showing/indexing sponsors.

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+          InvoiceMailer.with(invoice: @invoice).notify_organizers_sent.deliver_later
+        end
+    
+        render :show
+      rescue Pundit::NotAuthorizedError
+        raise
+      rescue => e
+        Rails.error.report(e)
+    
+        @sponsor = Sponsor.new(event: @event)
```
garyhtou: These two lines are probably not necessary.

app/views/api/v4/invoices/_invoice.json.jbuilder
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+json.created_at invoice.created_at
+json.id invoice.id
```
garyhtou:
```suggestion
json.id invoice.public_id
```

app/controllers/api/v4/sponsors_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+          contact_email: sponsor[:contact_email],
+          name: sponsor[:name],
+          event_id: sponsor[:organization_id]
+        )
+        authorize @sponsor
+      
+        if @sponsor.save
+          render :show
+        else
+          return render json: { error: e.message }, status: :internal_server_error
```
garyhtou: let's return a 400 (bad request) instead of 500

app/views/api/v4/sponsors/_sponsor.json.jbuilder
```diff
@@ -0,0 +1,14 @@
...
+json.address_city sponsor.address_city
+json.address_country sponsor.address_country
+json.address_line1 sponsor.address_line1
+json.address_line2 sponsor.address_line2
+json.address_postal_code sponsor.address_postal_code
+json.address_state sponsor.address_state
+json.contact_email sponsor.contact_email
+json.name sponsor.name
+json.slug sponsor.slug
+json.event_id sponsor.event_id
```
garyhtou:
```suggestion
json.event_id sponsor.event_id

```

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        filtered_params = params.require(:invoice).permit(
+          :due_date,
+          :item_description,
+          :item_amount
+        )
+    
+        due_date = filtered_params["due_date"].to_datetime
+
+        sponsor = authorize Sponsor.find_by_public_id(params[:sponsor_id]) || Sponsor.friendly.find(params[:sponsor_id])
```
garyhtou: Double check to see which policy this is using. Should it have a v4 API policy or is it okay to use the existing policy for the UI.

Luke-Oldenburg: I believe it's https://github.com/hackclub/hcb/blob/main/app/policies/sponsor_policy.rb. Let's just switch everything over to the UI policies when I cleanup the API. I'll also make a create_invoice check in event policy so we're not calling invoice policy with an event.

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+module Api
+  module V4
+    class InvoicesController < ApplicationController
+      def index
+        @event = authorize(Event.find_by_public_id(params[:event_id]) || Event.friendly.find(params[:event_id]), :index?)
+        @invoices = @event.invoices.order(created_at: :desc)
+      end
+
+      def show
+        @invoice = authorize Invoice.find_by_public_id(params[:id]) || Invoice.friendly.find(params[:id])
```
sampoder: Any reason for the friendly find? I think it should be safe to have

app/policies/event_policy.rb
```diff
@@ -27,6 +23,10 @@ def toggle_hidden?
     user&.admin?
   end
 
+  def index?
+    is_public || admin_or_user?
+  end
+
```
sampoder: The index of events should be public to everyone as it's user-specific

sampoder: It's the route used to generate your list of events in the command bar

app/controllers/api/v4/invoices_controller.rb
Luke-Oldenburg: I wanted to allow for both public and private IDs

Luke-Oldenburg: I'm planning on creating a helper function that does `find_by_public_id || friendly.find` and using that across the API in a subsequent PR. Discussed that with Gary.

app/policies/event_policy.rb
Luke-Oldenburg: I think I was using this policy to determine if members of the event should be indexable. Ex. invoices. What I did might be wrong though.

Luke-Oldenburg: > The index of events should be public to everyone as it's user-specific

Wait did you mean to say shouldn't? I'm still confused looking back at this.

sampoder: This is to determine whether you can access an index of events - it should be accessible to everyone. `InvoicePolicy` should define the index of invoices.

Luke-Oldenburg: This might be a dumb question but what is an index of events?

Luke-Oldenburg: I think I chose a bad name for the policy but what it was doing was correct I believe. I was using it to authorize an event not an invoice. Invoices should be able to be indexed on an event as long as you can authorize the event.

app/controllers/api/v4/invoices_controller.rb
Luke-Oldenburg: My implementation of invoice indexing is very similar to stripe card indexing in the API.

app/policies/event_policy.rb
sampoder: An index of events is a list of events you have access to. Head to https://hcb.hackclub.com/events.json to see yours. This method should not be changed, deleted or used for any other purposes.

> I think I chose a bad name for the policy but what it was doing was correct I believe. I was using it to authorize an event not an invoice. Invoices should be able to be indexed on an event as long as you can authorize the event.

This should be done in the `InvoicePolicy` as is currently being done in `InvoicesController`

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module Api
+  module V4
+    class InvoicesController < ApplicationController
+      def index
+        @event = authorize(Event.find_by_public_id(params[:event_id]) || Event.friendly.find(params[:event_id]), :invoice_list?)
+        @invoices = @event.invoices.order(created_at: :desc)
```
garyhtou:
```suggestion
        @event = Event.find_by_public_id(params[:event_id])
        @invoices = authorize(@event.invoices.order(created_at: :desc))
```

https://github.com/hackclub/hcb/blob/367cf2d85d213e21295c8aade7407a1d8027f13a/app/controllers/invoices_controller.rb#L10-L11
https://github.com/hackclub/hcb/blob/edb0fe4e9e7aaeee2c4a33fee2e366d21e109806/app/policies/invoice_policy.rb#L4-L14

app/controllers/api/v4/invoices_controller.rb
Luke-Oldenburg: Actually we're going to remove the friendly find and only use that on events.

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  module V4
+    class InvoicesController < ApplicationController
+      def index
+        @event = Event.find_by_public_id(params[:event_id]) || Event.friendly.find(params[:event_id])
+        @invoices = authorize(@event.invoices.order(created_at: :desc))
+      end
+
+      def show
+        # why does this cause "unable to find policy `NilClassPolicy` for `nil`" with a bad id?
+        @invoice = authorize Invoice.find_by_public_id(params[:id])
```
davidcornu: This is because `Invoice.find_by_public_id` returns `nil` when nothing is found

```
[4] pry(main)> Invoice.find_by_public_id("asdadsd")
=> nil
```

and `Pundit` uses the object you pass to `authorize` to determine which policy class to use (e.g. `Invoice` -> `InvoicePolicy`).

I would split this up into two steps


```suggestion
        @invoice = Invoice.find_by_public_id!(params[:id])
        authorize @invoice
```

- We first use `find_by_public_id!` which will raise `ActiveRecord::RecordNotFound` if nothing is found (and Rails' handling will render this as 404)
- Once we know we have a valid object we then pass it to `authorize`

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+      rescue => e
+        return render json: { error: e.message }, status: :bad_request
```
davidcornu: **concern**: What exceptions are you expecting to catch here?

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+      end
+
+      def create
+        event = authorize Event.find_by_public_id(params[:organization_id]) || Event.friendly.find(params[:organization_id])
+        authorize event, :create?, policy_class: InvoicePolicy
```
davidcornu: **concern**: Why do we need authorization logic on both the event and the invoice?

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        filtered_params = params.require(:invoice).permit(
+          :due_date,
+          :item_description,
+          :item_amount
+        )
+
+        due_date = filtered_params["due_date"].to_datetime
+
+        sponsor = authorize Sponsor.find_by_public_id(params[:sponsor_id])
```
davidcornu: - **concern**: I find it a bit odd to have the params to create the invoice under `params` but have `sponsor_id` hang off the side.
- **concern**: `Sponsor.find_by_public_id` can return `nil` (which will also make `authorize` fail). Either
    - `:sponsor_id` is optional in which case we should raise a 404 if it was provided but we can't find a corresponding record
    - `:sponsor_id` is required in which case we should always raise if we can't find a record. This seems to be the case as lot of the subsequent code assumes `sponsor` isn't `nil`.
-  **suggestion**:  Can we split the fetching and the authorization into separate lines?

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+        ).run
+
+        render :show
+      rescue Pundit::NotAuthorizedError
+        raise
+      rescue => e
+        Rails.error.report(e)
+
+        return render json: { error: e.message }, status: :internal_server_error
+      end
```
davidcornu: **concern**: I'm not exactly sure why this is necessary
1. We should have a default API response for `Pundit::NotAuthorizedError`. If we don't we should add one as relying on every endpoint to handle this is going to get messy
2. I'm not sure we want to blindly be sending exception messages to API clients. Again, we should have a default API responder for unexpected errors rather than rely on every endpoint doing this.

app/controllers/api/v4/sponsors_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+        @event = Event.find_by_public_id(params[:event_id]) || Event.friendly.find(params[:event_id])
+        @sponsors = authorize(@event.sponsors.order(created_at: :desc))
+      end
+
+      def show
+        # why does this cause "unable to find policy `NilClassPolicy` for `nil`" with a bad id?
+        @sponsor = authorize Sponsor.find_by_public_id(params[:id])
+      rescue => e
+        return render json: { error: e.message }, status: :bad_request
+      end
```
davidcornu, line 17: See https://github.com/hackclub/hcb/pull/10212/files#r2271048068

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+      def show
+        # why does this cause "unable to find policy `NilClassPolicy` for `nil`" with a bad id?
+        @invoice = authorize Invoice.find_by_public_id(params[:id])
+      rescue => e
+        return render json: { error: e.message }, status: :bad_request
+      end
+
+      def create
+        event = authorize Event.find_by_public_id(params[:organization_id]) || Event.friendly.find(params[:organization_id])
```
davidcornu: **suggestion**: I see this repeated a lot in this PR. Do we not have an API equivalent of https://github.com/hackclub/hcb/blob/990d33adb497e6a7ea1e323365a52fd83ffa5c26/app/controllers/concerns/set_event.rb? We could at least put this a `before_action`.

app/controllers/api/v4/sponsors_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+          contact_email: sponsor[:contact_email],
+          name: sponsor[:name],
+          event_id: event.id
+        )
+        authorize @sponsor
+
+        if @sponsor.save
+          render :show
+        else
+          return render json: { error: e.message }, status: :bad_request
```
davidcornu: **concern**: What is `e` in this context?

app/controllers/api/v4/sponsors_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+        # why does this cause "unable to find policy `NilClassPolicy` for `nil`" with a bad id?
+        @sponsor = authorize Sponsor.find_by_public_id(params[:id])
+      rescue => e
+        return render json: { error: e.message }, status: :bad_request
+      end
+
+      def create
+        event = authorize Event.find_by_public_id(params[:organization_id]) || Event.friendly.find(params[:organization_id])
+
+        sponsor = params.require(:sponsor).permit(
```
davidcornu: - **suggestion**: It's a pretty common pattern to put this logic in a private `create_params` method or something like that
- **suggestion**: I'd consider naming this `sponsor_params` at least as having both `@sponsor` and `sponsor` is confusing.

app/controllers/api/v4/sponsors_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+          address_city: sponsor[:address_city],
+          address_country: sponsor[:address_country],
+          address_line1: sponsor[:address_line1],
+          address_line2: sponsor[:address_line2],
+          address_postal_code: sponsor[:address_postal_code],
+          address_state: sponsor[:address_state],
+          contact_email: sponsor[:contact_email],
+          name: sponsor[:name],
+          event_id: event.id
+        )
```
davidcornu: **suggestion**: We can simplify this quite a bit


```suggestion
        @sponsor = event.sponsors.new(sponsor)
```

app/controllers/api/v4/sponsors_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+        authorize @sponsor
+
+        if @sponsor.save
+          render :show
```
davidcornu: **concern**: I'm not sure what our API conventions are but I'm used to seeing creation endpoints respond with a [201 Created](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/201) and setting the `Location` header to the resource's show URL.

app/models/invoice.rb
```diff
@@ -186,6 +186,12 @@ class NoAssociatedStripeCharge < StandardError; end
 
   before_create :set_defaults
 
+  after_create_commit -> {
+    unless OrganizerPosition.find_by(user: self.creator, event: event)&.manager?
```
davidcornu: **suggestion**: You shouldn't need `self` here

```suggestion
    unless OrganizerPosition.find_by(user: creator, event: event)&.manager?
```

app/policies/api/invoice_policy.rb
```diff
@@ -2,8 +2,20 @@
 
 module Api
   class InvoicePolicy < ApplicationPolicy
+    def index?
+      OrganizerPosition.role_at_least?(user, record, :reader)
```
davidcornu, line 6: **concern**: Using `record` here is a bit confusing as this is `InvoicePolicy` but we're dealing with a `Event`. Not sure if there's a better pattern though.

app/policies/api/invoice_policy.rb
```diff
@@ -2,8 +2,20 @@
...
+    end
+
     def show?
-      record.event.is_public?
+      OrganizerPosition.role_at_least?(user, record, :reader)
```
davidcornu: **concern**: Shouldn't `record` be an `Invoice` in this context?

app/policies/event_policy.rb
```diff
@@ -210,6 +210,10 @@ def activate?
 
   private
 
+  def admin_or_user?
+    admin? || record.users.include?(user)
```
davidcornu: **question**: Should this be `admin? || reader?`?

app/policies/api/invoice_policy.rb
Luke-Oldenburg: I don't really want to end up with this mess here as well.
https://github.com/luke-oldenburg/hcb/blob/lro-v4-api-invoice-sending/app/policies/invoice_policy.rb#L4-L13

app/controllers/api/v4/invoices_controller.rb
Luke-Oldenburg: Before, it was the NilClassPolicy error. Now there might not be anything. I like doing it so the user can always expect JSON back instead of HTML.

app/controllers/api/v4/invoices_controller.rb
Luke-Oldenburg: I think `sponsor_id` could go either way. `event_id` also isn't inside of the `filtered_params`.

app/controllers/api/v4/invoices_controller.rb
Luke-Oldenburg: Just tested and you can get an `ActiveRecord::RecordNotFound`

app/controllers/api/v4/application_controller.rb
```diff
@@ -15,6 +15,12 @@ class ApplicationController < ActionController::API
       before_action :set_expand
       before_action :set_paper_trail_whodunnit
 
+      if Rails.env.production?
+        rescue_from StandardError do |e|
+          render json: { error: "internal_server_error" }, status: :internal_server_error
+        end
+      end
```
davidcornu: **concern**: This feels very broad to me. 
- `StandardError` is the superclass for a lot of the exceptions we have below - will this override those?
	>  The handler of the first class for which exception.is_a?(klass) holds true is the one invoked, if any.
    [Source](https://api.rubyonrails.org/classes/ActiveSupport/Rescuable/ClassMethods.html#method-i-rescue_from)
- This doesn't tell future readers why this `rescue_from` was necessary
- This also might mean we aren't reporting errors that we should be aware of

app/controllers/concerns/set_event.rb
```diff
@@ -40,6 +40,14 @@ def set_event
       redirect_to params.to_unsafe_h
     end
 
+    def set_api_event
+      if params[:event_id]
+        @event = Event.find_by_public_id(params[:event_id]) || Event.friendly.find(params[:event_id])
+      else
+        @event = Event.find_by_public_id(params[:organization_id]) || Event.friendly.find(params[:organization_id])
```
davidcornu, line 47: **concern**: This feels to me like we should update the relevant routes to be consistent rather than add fallback cases here.

app/controllers/api/v4/events_controller.rb
```diff
@@ -40,10 +42,6 @@ def followers
 
       private
 
-      def set_event
-        @event = Event.find_by_public_id(params[:id]) || Event.find_by!(slug: params[:id])
-      end
```
davidcornu, line 47: **concern**: This is checking `params[:id]` whereas `set_api_event` checks `params[:event_id]`. Does it still work?

app/controllers/api/v4/invoices_controller.rb
davidcornu: > Just tested and you can get an `ActiveRecord::RecordNotFound`

That's fine though as there's already a handler for this in `Api::V4::ApplicationController`

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+      def show
+        skip_authorization
+        @invoice = Invoice.find_by_public_id!(params[:id])
+        authorize @invoice
+      rescue => e
+        return render json: { error: e.message }, status: :bad_request
+      end
+
+      def create
+        authorize @event, :create?
```
davidcornu: **question**: Does this end up using `EventPolicy#create?`? If so, that strikes me as a bit odd.

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        due_date = invoice_params["due_date"].to_datetime
+
+        sponsor = Sponsor.find_by_public_id(params[:sponsor_id])
```
davidcornu: **concern**: Can this be `nil`? What happens then?

app/controllers/api/v4/invoices_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+          current_user:,
+          sponsor_id: sponsor.id,
+          sponsor_name: sponsor.name,
+          sponsor_email: sponsor.contact_email,
+          sponsor_address_line1: sponsor.address_line1,
+          sponsor_address_line2: sponsor.address_line2,
+          sponsor_address_city: sponsor.address_city,
+          sponsor_address_state: sponsor.address_state,
+          sponsor_address_postal_code: sponsor.address_postal_code,
+          sponsor_address_country: sponsor.address_country
```
davidcornu, line 41: **suggestion**: This has me wondering whether `InvoiceService::Create` should just take a `sponsor` argument.

app/controllers/api/v4/sponsors_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+      include SetEvent
+
+      before_action :set_api_event, only: [:index, :create]
+
+      def index
+        @sponsors = authorize(@event.sponsors.order(created_at: :desc))
+      end
+
+      def show
+        skip_authorization
```
davidcornu: **question**: Why do we need to `skip_authorization` if we're subsequently calling `authorize`?

app/controllers/api/v4/sponsors_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+        @sponsor = Sponsor.find_by_public_id!(params[:id])
+        authorize @sponsor
+      rescue => e
+        return render json: { error: e.message }, status: :bad_request
```
davidcornu: **concern**: See previous comments about generic error handling. We should be able to rely on `Api::V4::ApplicationController` for this.

app/controllers/api/v4/sponsors_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      def create
+        authorize @event
+
+        @sponsor = @event.sponsors.new(sponsor_params)
+        authorize @sponsor
+
+        if @sponsor.save
+          render :show, status: :created, location: api_v4_sponsor_path(@sponsor)
+        else
```
davidcornu, line 27: **suggestion**: Do we want to pass along `@sponsor.errors.full_messages.to_sentence` here?

app/controllers/concerns/set_event.rb
Luke-Oldenburg: This is for backwards comparability.

app/controllers/api/v4/invoices_controller.rb
Luke-Oldenburg: It would require a mild refactor of the frontend but I would be on board with that. Updating the sponsor from the invoice feels wrong to me.

app/controllers/api/v4/sponsors_controller.rb
Luke-Oldenburg: I was getting authorization not performed errors.

app/controllers/api/v4/application_controller.rb
Luke-Oldenburg: What should we do then? When I was doing individual rescues, you had pointed me here.

app/controllers/api/v4/invoices_controller.rb
Luke-Oldenburg: I think I accidentally dropped the policy class from here.

app/controllers/api/v4/invoices_controller.rb
Luke-Oldenburg: No. I'll make it raise an error.

app/controllers/api/v4/events_controller.rb
Luke-Oldenburg: I'm not sure. I'm going to revert my changes to this file to be safe.

app/controllers/api/v4/invoices_controller.rb
davidcornu: Ah I handn't realized that the other use of `InvoiceService::Create` relies on passing all of these values from params https://github.com/hackclub/hcb/blob/141416a533aabdc625c217802c6eb836d6fb8548/app/controllers/invoices_controller.rb#L124-L140

app/controllers/api/v4/application_controller.rb
davidcornu: My recommendation would be to 
1. Remove this `rescue_from` as `StandardError` is _very_ broad. If we do get errors we don't know about then we'll get a 503 response and a AppSignal alert which is what we want (i.e. this would be an exceptional circumstance which needs addressing)
2. If there are _specific_ exceptions from our application code for which the default handling (either via the `rescue_from` causes in `Api::V4::ApplicationController` or the default handlers), either
    1. Add a new `rescue_from` to handle them if the response is generic to multiple endpoints
    2. Add `rescue`s for those specific exceptions where specific handling is needed

In general, I try to avoid `rescue StandardError` as much as possible as it can lead to errors not being reported and makes refactoring harder as future readers have no idea what exact exceptions are being caught here.

app/controllers/api/v4/invoices_controller.rb
Luke-Oldenburg: This might be good for a future PR.

app/controllers/api/v4/sponsors_controller.rb
davidcornu: This strikes me as extremely odd (e.g. we need it here but not on `index`). Have you had a chance to dig into those errors?

davidcornu:
This looks good to me. I have two remaining concerns but after that we should be good to go

- https://github.com/hackclub/hcb/pull/10212#discussion_r2353508640
- https://github.com/hackclub/hcb/pull/10212#discussion_r2353414042

Luke-Oldenburg:
> This looks good to me. I have two remaining concerns but after that we should be good to go
> 
>     * [[V4 API] Invoices & Sponsors #10212 (comment)](https://github.com/hackclub/hcb/pull/10212#discussion_r2353508640)
> 
>     * [[V4 API] Invoices & Sponsors #10212 (comment)](https://github.com/hackclub/hcb/pull/10212#discussion_r2353414042)

Both of those issues appear to have sorted themselves out over the course of working on this PR.

davidcornu (approved):

davidcornu:
I'm not getting any failures locally that I don't get against `main` so I think we're good to merge without the CI check.

PR #10213, opened 2025-04-21
[DX] Add instructions on getting an OAuth access token to dev-docs

Luke-Oldenburg:

### DIFF
--- a/dev-docs/development.md
+++ b/dev-docs/development.md
@@ -199,7 +199,13 @@ We've transitioned to using development keys and seed data in development, but h
 
 ## Getting an OAuth token
 
-To begin, open the rails console by running `bin/rails c`. Then, run `app = Doorkeeper::Application.create(name: "tester", redirect_uri: "http://localhost:3000/", scopes: ["read", "write"], confidential: false)` inside the console and save the output (you will need this later). After this, open `http://localhost:3000/api/v4/oauth/authorize?client_id=<UID>&redirect_uri=http://localhost:3000/&response_type=code&scope=read write`. Press the button to approve access on the page, then copy the code that appears in the address bar after being redirected. Now, send a POST request to `http://localhost:3000/api/v4/oauth/token` with a content type of `application/x-www-form-urlencoded` (you can use a tool like Postman to do this!). The request body fields are as follows:
+There is two different ways you can accomplish the first step of getting an OAuth token, either using a webpage, or the terminal depending on your preference.
+
+1. Go to [http://127.0.0.1:3000/api/v4/oauth/applications](http://127.0.0.1:3000/api/v4/oauth/applications). Press "New Application" and then set the name to anything of your choosing and the redirect URI to `http://localhost:3000/`. Press "Submit" and then save the info on the new page that appears and press "Authorize."
+
+2. Open the rails console by running `bin/rails c`. Then, run `app = Doorkeeper::Application.create(name: "tester", redirect_uri: "http://localhost:3000/", scopes: ["read", "write"], confidential: false)` inside the console and save the output (you will need this later). After this, open `http://localhost:3000/api/v4/oauth/authorize?client_id=<UID>&redirect_uri=http://localhost:3000/&response_type=code&scope=read write`.
+
+Press the button to approve access on the page, then copy the code that appears in the address bar after being redirected. Now, send a POST request to `http://localhost:3000/api/v4/oauth/token` with a content type of `application/x-www-form-urlencoded` (you can use a tool like Postman to do this!). The request body fields are as follows:
 ```
 grant_type=authorization_code
 code=<CODE>
--- a/dev-docs/development.md
+++ b/dev-docs/development.md
@@ -199,12 +199,12 @@ We've transitioned to using development keys and seed data in development, but h
 
 ## Getting an OAuth token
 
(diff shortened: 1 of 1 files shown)


dev-docs/development.md
```diff
@@ -195,3 +195,16 @@ We've transitioned to using development keys and seed data in development, but h
...
+
+
+## Getting an OAuth token
+
+To begin, open the rails console by running `bin/rails c`. Then, run `app = Doorkeeper::Application.create(name: "tester", redirect_uri: "https://hackclub.com/", scopes: ["read", "write"], confidential: false)` inside the console and save the output (you will need this later). After this, open `http://127.0.0.1/api/v4/oauth/authorize?client_id=<UID>&redirect_uri=https://hackclub.com/&response_type=code&scope=read write`. Press the button to approve access on the page, then copy the code that appears in the address bar after being redirected. Now, send a POST request to `http://127.0.0.1/api/v4/oauth/token` with a content type of `application/x-www-form-urlencoded` (you can use a tool like Postman to do this!). The request body fields are as follows:
+```
+grant_type=authorization_code
+code=<CODE>
+client_id=<UID>
+client_secret=<SECRET>
```
polypixeldev, line 212: afaik `confidential: false` removes the need for passing a client secret?

dev-docs/development.md
```diff
@@ -195,3 +195,16 @@ We've transitioned to using development keys and seed data in development, but h
 ## Flipper
 
 [Flipper](https://github.com/flippercloud/flipper) is used to toggle feature flags on HCB. Flipper can be accessed at [localhost:3000/flipper/features](http://localhost:3000/flipper/features). To enable a flag, press "Add Feature", paste in the name of a feature from [this list](https://hcb.hackclub.com/api/flags), and then press "Fully Enable".
+
+
+## Getting an OAuth token
+
+To begin, open the rails console by running `bin/rails c`. Then, run `app = Doorkeeper::Application.create(name: "tester", redirect_uri: "https://hackclub.com/", scopes: ["read", "write"], confidential: false)` inside the console and save the output (you will need this later). After this, open `http://127.0.0.1/api/v4/oauth/authorize?client_id=<UID>&redirect_uri=https://hackclub.com/&response_type=code&scope=read write`. Press the button to approve access on the page, then copy the code that appears in the address bar after being redirected. Now, send a POST request to `http://127.0.0.1/api/v4/oauth/token` with a content type of `application/x-www-form-urlencoded` (you can use a tool like Postman to do this!). The request body fields are as follows:
```
polypixeldev: small thing but the redirect_uri could probably just be a localhost URL since it doesn't actually need to exist and to not send requests with codes to other servers

polypixeldev: assumes your instance of HCB is running on port 80 (`http://127.0.0.1`)

dev-docs/development.md
Luke-Oldenburg: Not sure but this also works. If you confirm then I will switch it.

dev-docs/development.md
Luke-Oldenburg: Thanks for catching that!

Luke-Oldenburg: Good idea

dev-docs/development.md
polypixeldev: Yeah it might work but not necessary, I haven't been passing a client secret on my apps. HCB mobile also doesn't (as it'd have to be included with the app code and therefore useless)

dev-docs/development.md
```diff
@@ -195,3 +195,16 @@ We've transitioned to using development keys and seed data in development, but h
 ## Flipper
 
 [Flipper](https://github.com/flippercloud/flipper) is used to toggle feature flags on HCB. Flipper can be accessed at [localhost:3000/flipper/features](http://localhost:3000/flipper/features). To enable a flag, press "Add Feature", paste in the name of a feature from [this list](https://hcb.hackclub.com/api/flags), and then press "Fully Enable".
+
+
+## Getting an OAuth token
+
+To begin, open the rails console by running `bin/rails c`. Then, run `app = Doorkeeper::Application.create(name: "tester", redirect_uri: "http://localhost:3000/", scopes: ["read", "write"], confidential: false)` inside the console and save the output (you will need this later). After this, open `http://localhost:3000/api/v4/oauth/authorize?client_id=<UID>&redirect_uri=http://localhost:3000/&response_type=code&scope=read write`. Press the button to approve access on the page, then copy the code that appears in the address bar after being redirected. Now, send a POST request to `http://localhost:3000/api/v4/oauth/token` with a content type of `application/x-www-form-urlencoded` (you can use a tool like Postman to do this!). The request body fields are as follows:
```
scooterthedev: Everything up to sending the `POST` request for the Bearer token can be done via `/api/v4/oauth/applications` which is a UI application. This method is a lot more straight-forward and simpler.

Luke-Oldenburg: @thedev132 ...

Mohamad-Mortada: Ah I never knew that existed lol. It'd be good to include both as using the API is good for oAuth apps

scooterthedev: True, but if were gonna add both, it will need to .md formatting ;)

Mohamad-Mortada: Yea for sure :) Both are very handy

scooterthedev (approved):
LGTM!

sampoder (approved):
LGTM - thank you for creating this!

Luke-Oldenburg:
That was 127.0.0.1 because localhost broke it.

Luke-Oldenburg:
Idk why but with localhost it would just redirect to root

sampoder:
Hmm - that happens when you aren't signed in as an admin, it seems to be working for me

Luke-Oldenburg:
Just it not working on localhost or in general? I didn't change which user I was signed in to while I was testing URLs.

Luke-Oldenburg:
Hm yea it's working now. Not sure what was up before.

PR #10230, opened 2025-04-23
[UI] Make the "Jump to" keybinds change for Windows

scooterthedev:
## Summary of the problem

It used to show ⌘  K in the `Jump To...` menu even while on Windows

## Describe your changes

It now shows Ctrl K instead of ⌘ K while on Windows. On Mac, it stays as ⌘ K

Before on Windows
image 

After on Windows
image

### DIFF
--- a/app/views/users/_nav.html.erb
+++ b/app/views/users/_nav.html.erb
@@ -10,8 +10,8 @@
       style="border-radius: 8px"
       data-behavior="command_bar_trigger">
       <span class="flex-grow text-left font-medium">Jump to...</span>
-      <span class="bg-gray-300 dark:bg-gray-700 opacity-75 <%= request.user_agent&.to_s&.upcase&.include?('MAC') ? 'w-6' : 'w-9' %> rounded-md mr-2 text-sm">
-        <%= request.user_agent&.to_s&.upcase&.include?('MAC') ? '⌘'  : 'Ctrl' %>
+      <span class="bg-gray-300 dark:bg-gray-700 opacity-75 <%= request.user_agent.to_s.upcase.include?('MAC') ? 'w-6' : 'w-9' %> rounded-md mr-2 text-sm">
+        <%= request.user_agent.to_s.upcase.include?('MAC') ? '⌘'  : 'Ctrl' %>
       </span>
       <span class="bg-gray-300 dark:bg-gray-700 opacity-75 w-6 rounded-md text-sm">
         K
--- a/app/views/users/_nav.html.erb
+++ b/app/views/users/_nav.html.erb
@@ -10,8 +10,8 @@
       style="border-radius: 8px"
       data-behavior="command_bar_trigger">
       <span class="flex-grow text-left font-medium">Jump to...</span>
-      <span class="bg-gray-300 dark:bg-gray-700 opacity-75 w-6 rounded-md mr-2 text-sm">
-        ⌘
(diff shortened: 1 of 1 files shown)


app/views/users/_nav.html.erb
```diff
@@ -10,8 +10,8 @@
       style="border-radius: 8px"
       data-behavior="command_bar_trigger">
       <span class="flex-grow text-left font-medium">Jump to...</span>
-      <span class="bg-gray-300 dark:bg-gray-700 opacity-75 w-6 rounded-md mr-2 text-sm">
-        ⌘
+      <span class="bg-gray-300 dark:bg-gray-700 opacity-75 w-9 rounded-md mr-2 text-sm">
```
YodaLightsabr: Using `w-9` for both MacOS and Windows results in a really wide ⌘:

image

Could you also check the user agent here and use `w-6` for Mac and `w-9` for Windows?

YodaLightsabr (commented):
Thanks for working on this!

app/views/users/_nav.html.erb
scooterthedev: I think w-9 for windows looks fine. I also just added the user agent check here. If it's mac = w-6, and the rest is w-9. Also added linux to the Ctrl group while I was at it ;).

app/views/users/_nav.html.erb
```diff
@@ -10,8 +10,8 @@
       style="border-radius: 8px"
       data-behavior="command_bar_trigger">
       <span class="flex-grow text-left font-medium">Jump to...</span>
-      <span class="bg-gray-300 dark:bg-gray-700 opacity-75 w-6 rounded-md mr-2 text-sm">
-        ⌘
+      <span class="bg-gray-300 dark:bg-gray-700 opacity-75 <%= request.user_agent.to_s.upcase.include?('MAC') ? 'w-6' : 'w-9' %> rounded-md mr-2 text-sm">
+        <%= request.user_agent.to_s.upcase.include?('WIN') || request.user_agent.to_s.upcase.include?('LINUX') ? 'Ctrl' : '⌘' %>
```
sampoder:
```suggestion
      <span class="bg-gray-300 dark:bg-gray-700 opacity-75 <%= request.user_agent&.to_s&.upcase&.include?('MAC') ? 'w-6' : 'w-9' %> rounded-md mr-2 text-sm">
        <%= request.user_agent&.to_s&.upcase&.include?('MAC') ? '⌘'  : 'Ctrl' %>
```

sampoder: Tiny thing just incase the user agent is nil - thank you so much for building this!

scooterthedev: Thanks! Why does it need to be in-cased in nil- just for the future?

YodaLightsabr: `nil.to_s` = `""`

YodaLightsabr: We should be good to leave out the safe navigation

sampoder: Oh ok - that's chill then

scooterthedev: So you don't want the `&` safe nav, or is it fine as is?

YodaLightsabr: We can leave it out, it doesn't make a difference

scooterthedev: Okay, I switched it back. @sampoder does it look fine now?

sampoder (approved):
LGTM - thank you for putting this together!

PR #10231, opened 2025-04-23
Improve reader UI

manuthecoder:
This pull request introduces a consistent check for user roles by updating calls to the `organizer_signed_in?` method across multiple views. The method now explicitly checks if the user is signed in as a member by passing the `as: :member` argument. This change ensures that only users with the appropriate role can access or interact with certain UI elements.

### Updates to role-based access control:

* Updated `organizer_signed_in?` calls in various views to include the `as: :member` argument, ensuring role-based restrictions are applied consistently:
  - `app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb`
  - `app/views/canonical_transactions/_canonical_transaction.html.erb`
  - `app/views/events/_nav.html.erb` [[1]](diffhunk://#diff-1cbff28c76eb6b28dd25ab919a0bcc05df5cd5d8c67abad63346478beb6ff918L60-R60) [[2]](diffhunk://#diff-1cbff28c76eb6b28dd25ab919a0bcc05df5cd5d8c67abad63346478beb6ff918L118-R118) [[3]](diffhunk://#diff-1cbff28c76eb6b28dd25ab919a0bcc05df5cd5d8c67abad63346478beb6ff918L143-R143) [[4]](diffhunk://#diff-1cbff28c76eb6b28dd25ab919a0bcc05df5cd5d8c67abad63346478beb6ff918L172-R172)
  - `app/views/events/card_overview.html.erb`
  - `app/views/events/reimbursements.html.erb`
  - `app/views/events/transfers.html.erb`
  - `app/views/hcb_codes/_meatballs.html.erb`
  - `app/views/hcb_codes/_transaction_history.html.erb`
  - `app/views/hcb_codes/_unknown.html.erb`
  - `app/views/hcb_codes/reimbursement/_payout_holding.html.erb`
  - `app/views/invoices/index.html.erb`
  - `app/views/transactions/show.html.erb`


app/views/card_grants/_details.html.erb
```diff
@@ -4,7 +4,7 @@
   <h2 class="border-none mx0 mt0 inline-flex items-center">
     <%= pop_icon_to "edit",
         [:edit, card_grant.disbursement.local_hcb_code],
-        class: "mr2 align-middle", "aria-label": "Rename transaction" %>
+        class: "mr2 align-middle", "aria-label": "Rename transaction" (as: :member)%>
```
garyhtou: is this a typo?

app/views/events/_nav.html.erb
```diff
@@ -57,7 +57,7 @@
       event_check_deposits_path(@event),
       tooltip: "Deposit checks",
       icon: "cheque",
-      selected: local_assigns[:selected] == :deposit_check if policy(@event.check_deposits.build).index? %>
+      selected: local_assigns[:selected] == :deposit_check if policy(@event.check_deposits.build).index? && organizer_signed_in?(as: :member) %>
```
garyhtou: Since we're already calling the policy here, we might want to just include the member check inside the policy for `index?`

app/views/events/_nav.html.erb
```diff
@@ -140,7 +140,7 @@
           <%= link_to "Donations", event_donation_overview_path(@event) if policy(@event).donation_overview? %>
           <%= link_to "Invoices", event_invoices_path(@event) if @event.approved? && @event.plan.invoices_enabled? %>
           <%= link_to "Account numbers", account_number_event_path(@event) if policy(@event).account_number? && !Flipper.enabled?(:event_home_page_redesign_2024_09_21, @event) %>
-          <%= link_to "Check deposits", event_check_deposits_path(@event) if policy(@event.check_deposits.build).index? %>
+          <%= link_to "Check deposits", event_check_deposits_path(@event) if policy(@event.check_deposits.build).index? && organizer_signed_in?(as: :member) %>
```
garyhtou: Move check into policy

app/views/events/reimbursements.html.erb
```diff
@@ -24,7 +24,7 @@
         data: { behavior: "modal_trigger", modal: "create_reimbursement_report" } %>
   <%= link_to "#", class: "btn bg-success xs-hide", data: { behavior: "modal_trigger", modal: "create_reimbursement_report" } do %>
       <%= inline_icon "plus" %> New
-  <% end %>
+  <% end if organizer_signed_in?(as: :member) %>
```
garyhtou: Let's wrap line 20 through 27 with the `if`.

app/views/reimbursement/reports/show.html.erb
```diff
@@ -35,7 +35,7 @@
                       icon_size: 26,
                       class: "tooltipped tooltipped--w",
                       style: "height: 32px; width: 32px; background: #ec375020; color: #ec3750",
-                      'aria-label': "Delete this report." %>
+                      'aria-label': "Delete this report." if organizer_signed_in?(as: :member) %>
```
garyhtou: I'd suggest combining this with the `if` above.

app/views/reimbursement/reports/show.html.erb
```diff
@@ -44,7 +44,7 @@
                       icon_size: 28,
                       class: "tooltipped tooltipped--w",
                       style: "height: 32px; width: 32px;",
-                      'aria-label': "Edit this report's properties." %>
+                      'aria-label': "Edit this report's properties." if organizer_signed_in?(as: :member) %>
```
garyhtou: Combine with `unless` above.

app/views/events/_nav.html.erb
```diff
@@ -169,7 +169,7 @@
         <%= link_to "Promotions & perks", event_promotions_path(event_id: @event.slug) if policy(@event).promotions? %>
         <%= link_to "Google Workspace", event_g_suite_overview_path(event_id: @event.slug) if policy(@event).g_suite_overview? %>
         <%= link_to "Documents", event_documents_path(event_id: @event.slug) if policy(@event).documentation? %>
-        <%= link_to "Settings", edit_event_path(@event) %>
+        <%= link_to "Settings", edit_event_path(@event) if organizer_signed_in?(as: :member) %>
```
Luke-Oldenburg, line 172: Readers should be able to view settings.

app/views/events/card_overview.html.erb
```diff
@@ -22,7 +22,7 @@
     <%= render "filter_cards_menu" unless @has_filter %>
     <%= pop_icon_to @view == "list" ? "grid" : "list",
                     "?view=#{@view == 'list' ? 'grid' : 'list'}#{"&q=#{@q}" if @q}#{"&status=#{@status}" if @status}" %>
-    <% if organizer_signed_in? %>
+    <% if organizer_signed_in?(as: :member) %>
```
Luke-Oldenburg, line 25: I think it's okay to show this page even though it will unauthorized when they press order because they could also do it from their personal cards page and change the event selected.

app/views/events/reimbursements.html.erb
```diff
@@ -17,13 +17,15 @@
...
         class: "tooltipped tooltipped--s", "aria-label": "Public form" %>
   <% end %>
-  <%= pop_icon_to "plus",
-        "#",
-        class: "tooltipped tooltipped--s success sm-hide md-hide lg-hide",
-        "aria-label": "New report",
-        data: { behavior: "modal_trigger", modal: "create_reimbursement_report" } %>
-  <%= link_to "#", class: "btn bg-success xs-hide", data: { behavior: "modal_trigger", modal: "create_reimbursement_report" } do %>
-      <%= inline_icon "plus" %> New
+  <% if organizer_signed_in?(as: :member) %>
```
Luke-Oldenburg, line 20: I think it's better to have the button disabled instead of hiding it.

app/views/events/transfers.html.erb
```diff
@@ -8,7 +8,7 @@
     <%= badge_for @transfers.total_count, class: "bg-muted" %>
   </span>
 
-  <% if organizer_signed_in? %>
+  <% if organizer_signed_in?(as: :member) %>
```
Luke-Oldenburg, line 12: Does this hide the send button? That should be visible because all of the individual transfer pages have their fields disabled unless you are a manager.

Luke-Oldenburg (commented):
I've talked about this with @sampoder some in the past and have made PRs working towards making the UI more consistent in terms of behavior. I had thought the approach we were going towards was making as much non sensitive info as possible visible but just disable any action buttons / fields rather than hide them entirely.

garyhtou:
I'd like for Reader to behave more like transparency mode, where users can't read settings and some buttons for creating/modifying are hidden. I think for an entirely read-only role, disabling buttons instead of hiding them makes a lot of clutter.

I see the value in having a consistent UI, but I'm not sure that outweighs the value of having a cleaner UI.

app/views/events/_nav.html.erb
```diff
@@ -140,7 +140,7 @@
           <%= link_to "Donations", event_donation_overview_path(@event) if policy(@event).donation_overview? %>
           <%= link_to "Invoices", event_invoices_path(@event) if @event.approved? && @event.plan.invoices_enabled? %>
           <%= link_to "Account numbers", account_number_event_path(@event) if policy(@event).account_number? && !Flipper.enabled?(:event_home_page_redesign_2024_09_21, @event) %>
-          <%= link_to "Check deposits", event_check_deposits_path(@event) if policy(@event.check_deposits.build).index? %>
+          <%= link_to "Check deposits", event_check_deposits_path(@event) if policy(@event.check_deposits.build).create? %>
```
garyhtou: Readers should be able to see the index page

app/views/events/_nav.html.erb
garyhtou: Im cool with readers not being able to view settings. Do you have strong feeling otherwise?

Luke-Oldenburg: I thought about it more and I think this approach could make sense:
Readers- pages are hidden
Members- pages are disabled
Managers- pages work

I don't think I ever fully discussed how readers should work with Sam just members and I recall him saying for members we should show the pages disabled.

I'm not particular either way as long as we make sure to update the roles table.

sampoder:
I personally lean towards disabling items because it gives people a better sense of how HCB works but I'm also down to reuse the stripped down UI we built for transparency mode.

app/views/events/_nav.html.erb
manu-gurudath: >  Readers- pages are hidden
> Members- pages are disabled
> Managers- pages work

I agree, this approach feels intuitive

manu-gurudath:
just fixed check deposit settings! let me know if any further changes are needed

garyhtou (approved):

PR #10243, opened 2025-04-27
Enable Sidekiq in development

polypixeldev:
## Summary of the problem

HCB in development currently uses the default async backend for Active Job, which lacks Sidekiq's features like the dashboard and cron jobs.

## Describe your changes

I enabled using Sidekiq in the development environment file. While this makes it so that new jobs use Sidekiq, I couldn't figure out why it didn't automatically load the jobs from `schedule.yml` - which it should be doing automatically according to `sidekiq-cron`'s [README](https://github.com/sidekiq-cron/sidekiq-cron?tab=readme-ov-file#loading-jobs-from-schedule-file) (and if not, in the sidekiq initializer). For now, I've added dev docs on how to load the schedule manually.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -101,6 +101,8 @@ Rails.application.configure do
   # SMTP config
   config.action_mailer.delivery_method = :letter_opener_web
 
+  config.active_job.queue_adapter = :sidekiq
+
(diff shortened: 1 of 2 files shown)


sampoder (approved):
LGTM!

PR #10252, opened 2025-04-28
Fix new policy for ACHs and checks

Luke-Oldenburg:
Fixes unauthorized redirect bug related to #10152

### DIFF
--- a/app/policies/ach_transfer_policy.rb
+++ b/app/policies/ach_transfer_policy.rb
@@ -53,7 +53,7 @@ class AchTransferPolicy < ApplicationPolicy
   end
 
   def admin_or_user?
-    OrganizerPosition.role_at_least?(user, record.event, :member)
+    user&.admin? || record.event.users.include?(user)
   end
 
   def admin_or_manager?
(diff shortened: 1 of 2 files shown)


app/policies/ach_transfer_policy.rb
```diff
@@ -52,6 +52,10 @@ def user_who_can_transfer?
     EventPolicy.new(user, record.event).create_transfer?
   end
 
+  def admin_or_user?
+    OrganizerPosition.role_at_least?(user, record.event, :member)
```
sampoder: Should we be letting readers see this? Not sure what the other pages do

Luke-Oldenburg: I copied that policy from another one but arguably they all could allow for that. Should I change it?

sampoder: Sure and probs on the other ones as well

sampoder: Sure and probs on the other ones as well

Luke-Oldenburg: Turns out it was just checks too

sampoder (approved):
lgtm!

PR #10253, opened 2025-04-28
Improve card grant modal behavior

Luke-Oldenburg:
Switched from jquery refresh after doing turbo to using js to hide and reveal it with a back button. Works much smoother and faster. Only issue is the back button looks a little out of place.


sampoder:
Screenshot 2025-05-08 at 4 10 40 PM

I would prefer if we went with this setup - any reason we can't?

Luke-Oldenburg:
I couldn't get the modal to close when opening the next one. closeExisiting just hides the entire div.

sampoder:
Can you send me the code you had? I can give it a shot

Luke-Oldenburg:
https://github.com/hackclub/hcb/tree/1bf6952679938c45a4553f00f3d40bb3ebee2f49 this is after the "Small tweaks" commit. It got it to a separate modal and form but they were stacking on top of each other.

sampoder:
I figured it out! It's a bug with `jquery-modal`. 

The problem is we having fading setup - so when you call close(), it doesn't immediately close it but queues the close() function to run in 10 seconds, but the close function closes all modals. So once the new one is opened, they both get closed.

Here's how I fixed it: https://github.com/hackclub/hcb/tree/fix-jquery-modal

Luke-Oldenburg:
> I figured it out! It's a bug with `jquery-modal`.
> 
> The problem is we having fading setup - so when you call close(), it doesn't immediately close it but queues the close() function to run in 10 seconds, but the close function closes all modals. So once the new one is opened, they both get closed.
> 
> Here's how I fixed it: https://github.com/hackclub/hcb/tree/fix-jquery-modal

It worked! Thank you!

garyhtou (approved):
great work here @Luke-Oldenburg and @sampoder !

garyhtou:
It should auto merge after you resolve merge conflicts

PR #10255, opened 2025-04-29
Implement dropdown buttons

polypixeldev:
## Summary of the problem

Better UI! Similar actions should be able to be grouped in a single dropdown button, like how GitHub does it.

Closes #8219 

## Describe your changes

Adds a helper and corresponding Stimulus controller and CSS for dropdown buttons, which are implemented in the reimbursement page for standard/mileage expenses.

https://github.com/user-attachments/assets/f54a9524-cd3f-47e4-a394-ed59e8f44673

### DIFF
--- a/app/models/reimbursement/expense/mileage.rb
+++ b/app/models/reimbursement/expense/mileage.rb
@@ -34,7 +34,7 @@ module Reimbursement
   class Expense
     class Mileage < ::Reimbursement::Expense
       def rate
-        event.plan.mileage_rate(created_at || Date.today)
+        event.plan.mileage_rate(created_at || Date.today) # rate may be used on creation, however, created_at isn't set at that point.
       end
 
       def value_label
--- a/app/models/reimbursement/expense/mileage.rb
+++ b/app/models/reimbursement/expense/mileage.rb
@@ -34,7 +34,7 @@ module Reimbursement
   class Expense
     class Mileage < ::Reimbursement::Expense
       def rate
-        event.plan.mileage_rate(created_at)
+        event.plan.mileage_rate(created_at || Date.today)
       end
 
       def value_label
--- a/app/views/reimbursement/reports/_actions.html.erb
+++ b/app/views/reimbursement/reports/_actions.html.erb
@@ -14,12 +14,12 @@
       <div class="flex justify-start">
         <% if !report.locked? %>
           <%= form_with(method: :post, url: reimbursement_expenses_path(report_id: report.id), data: { controller: "form file-drop", turbo: "true" }, class: "tooltipped tooltipped--n", html: { "aria-label": "Drag a file into this button to create a new expense with a receipt attached." }) do |form| %>
-            <div class="relative">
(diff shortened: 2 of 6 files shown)


sampoder:
Wow - this looks like it was really nicely done!

app/views/reimbursement/reports/_actions.html.erb
```diff
@@ -14,17 +14,24 @@
...
-                data: { "file-drop-target" => "fileInput" } %>
+            <div class="add-expense-dropzone">
+              <%= dropdown_button form:,
+                options: [
+                  ["Standard expense", "standard", "Generic expense with a specified amount."],
+                  ["Mileage expense", "mileage", "Expense for miles traveled based on the standard mileage rate."]],
+                button_icon: "plus",
+                name: "type",
+                button_container_options: { data: { "file-drop-target": "dropzone", action: "dragover->file-drop#dragover drop->file-drop#drop dragenter->file-drop#dragenter dragleave->file-drop#dragleave" }},
+                template: "Add #{report.expenses.count > 0 ? "another" : "an"} [VALUE] expense"
```
sampoder: Can we define these as lambdas instead of string templates?

app/models/reimbursement/expense/mileage.rb
```diff
@@ -34,7 +34,7 @@ module Reimbursement
   class Expense
     class Mileage < ::Reimbursement::Expense
       def rate
-        event.plan.mileage_rate(created_at)
+        event.plan.mileage_rate(created_at || Date.today)
```
sampoder: Hmm - what's the thought behind this?

sampoder:
Screenshot 2025-04-29 at 10 47 16 PM

Thoughts on making these as rounded as the button?

app/models/reimbursement/expense/mileage.rb
polypixeldev: When creating a new mileage reimbursement using the dropdown button, the reimbursement record was not created yet, and `created_at` would be nil, causing an error in `mileage_rate`.

app/views/reimbursement/reports/_actions.html.erb
polypixeldev: I don't think we can since the dropdown button stimulus controller has to be able to update the text, not just the ruby code. It'd have to be encoded in HTML and executable by JS somehow

app/models/reimbursement/expense/mileage.rb
sampoder:
```suggestion
        event.plan.mileage_rate(created_at || Date.today) # rate may be used on creation, however, created_at isn't set at that point.
```

sampoder: Sounds good - I added a comment to explain this

app/views/reimbursement/reports/_actions.html.erb
sampoder: Just pushed some code to do this - I'm putting the label into each piece of HTML, thoughts? After you take a look at that I think this is good to merge

polypixeldev: Ahh I see, looks good to me!

PR #10257, opened 2025-04-29
Reduce negative margin on home page merchants graph

polypixeldev:
## Summary of the problem

Negative margin was added to the home page merchants graph to avoid a lot of white space on the left, added by the graph library - but this cut off some of the labels when dealing with larger amounts:
image

## Describe your changes

Reduces the negative margin to avoid cutting off larger values on the merchants graph


sampoder (approved):
Thanks!

PR #10258, opened 2025-04-29
Increase text contrast on warning buttons

polypixeldev:
## Summary of the problem

White text on the warning button background is hard to read:
image

## Describe your changes

Used the same darkened filter as other button backgrounds on the warning background:

image


sampoder:
Thoughts on just doing this in dark mode? I feel comfortable with the contrast in light mode

sampoder (approved):
Looks fire!

PR #10153, opened 2025-04-13
Add docs for using Stripe in test mode

polypixeldev:
## Summary of the problem

I recently discovered Stripe's awesome test mode, but had to spend some time figuring out how it worked and setting it up for HCB.

## Describe your changes

Added dev docs for using Stripe in test mode, specifically for donations and simulating card charges.

PR #10275, opened 2025-04-29
Document capturing a card authorization in development

polypixeldev:
## Summary of the problem

#10153 left out the process of capturing an authorization.

## Describe your changes

Documented the process of capturing and the job you have to run for HCB to mark the transaction as settled.


sampoder (approved):
LGTM!

PR #10280, opened 2025-04-30
[Login] Encrypt Browser Token (1/3)

manuthecoder:
This PR starts the migration to encrypt `logins.browser_token` using Lockbox.

#### Changes:
- Adds `browser_token_ciphertext` column
- Enables `has_encrypted :browser_token, migrating: true` in the model

This keeps current behavior unchanged while allowing encrypted values to be written in parallel. Safe to deploy with zero downtime.

---

### ⏭️ Next Step (after deploy):
Someone with production console access should run:

```ruby
Lockbox.migrate(Login)
```

---

### Upcoming PR:
A follow-up PR will:
- Finalize Lockbox usage
- Remove `has_secure_token`
- Drop the old column

A follow-up PR will be created once this is merged and run

(Let me know if this isn’t the right approach — happy to adjust)

### DIFF
--- a/app/models/login.rb
+++ b/app/models/login.rb
@@ -25,7 +25,7 @@ class Login < ApplicationRecord
   belongs_to :user
   belongs_to :user_session, optional: true
 
-  has_secure_token :browser_token
+  has_encrypted :browser_token, migrating: true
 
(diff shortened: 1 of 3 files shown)


sampoder (approved):
LGTM!

PR #10281, opened 2025-04-30
 [Login] Encrypt Browser Token (2/3)

manuthecoder:
### 🔐 Finalize & clean up `browser_token` encryption with Lockbox

This PR completes the Lockbox migration for `logins.browser_token` and removes the legacy column. *Do NOT merge this until all steps in #10280 are complete.*

#### Changes:

* Removes `has_secure_token :browser_token`
* Finalizes `has_encrypted :browser_token` by removing `migrating: true`
* Adds `before_validation :ensure_browser_token` fallback logic
* Adds `self.ignored_columns += ["browser_token"]` to safely ignore the legacy column
* Adds `ensure_browser_token` method to generate secure token if not already set

> [!NOTE] 
> Assumes `Lockbox.migrate(Login)` has already been run in production.

_A follow-up PR will be opened to safely remove the browser_token column from the database once this is deployed._


sampoder:
FYI, you need to ignore the column in your first PR and then drop it in a second one

Screenshot 2025-04-30 at 11 48 02 AM

garyhtou:
Good work here!

garyhtou (approved):

PR #10292, opened 2025-05-01
Send donation summary correction emails

polypixeldev:
## Summary of the problem

Monthly donation summaries that just got sent out included all donations, including not processed and refunded donations. Fixed for future summaries in #10289

## Describe your changes

Added a one time job that will send a correction email to all events that were sent an incorrect summary. Also fixes the subject to correctly pluralize "donation".
This job will need to be run on production when merged (might want to wait until #10291 is merged as well)

### DIFF
--- a/app/mailers/event_mailer.rb
+++ b/app/mailers/event_mailer.rb
@@ -13,7 +13,7 @@ class EventMailer < ApplicationMailer
 
     @total = @donations.sum(:amount)
 
-    mail to: @emails, subject: "#{"[CORRECTION] " if params[:correction]} #{@event.name} received #{@donations.length} donations this past month"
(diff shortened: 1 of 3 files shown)


app/mailers/event_mailer.rb
```diff
@@ -5,15 +5,15 @@ class EventMailer < ApplicationMailer
   before_action :set_emails
 
   def monthly_donation_summary
-    @event = params[:event]
+    @correction = params[:correction]
```
garyhtou, line 8: I think we need to still set `@event`

polypixeldev: I noticed that it was already being set with the `before_action` on line 4 - it worked when I tested in development

garyhtou: good catch! you're all good here

PR #10293, opened 2025-05-01
Remove donation summary correction code

polypixeldev:
## Summary of the problem

Job ran, no longer need code from #10292 

## Describe your changes

Removed correction related code


sampoder (approved):
thanks!

PR #10302, opened 2025-05-02
Add tags tab to organization settings

polypixeldev:
## Summary of the problem

Organization members can't currently modify existing transaction tags or delete them without going to add one first.

Closes #9488 

## Describe your changes

Adds a tab to organization settings if transaction tags are enabled that shows all tags that event has and allows members to create, modify, and delete them.

Screenshot from 2025-05-01 22-54-11
Screenshot from 2025-05-01 22-54-24

### DIFF
--- a/app/views/events/settings/_tags.html.erb
+++ b/app/views/events/settings/_tags.html.erb
@@ -6,13 +6,32 @@
           <div class="tag__preview tag-<%= tag.color || "muted" %>"></div>
           <p class="bold"><%= tag.label %></p>
         </div>
-        <div>
-          <%= pop_icon_to "edit",
-            nil,
-            class: "tooltipped tooltipped--w mr2 h-8 w-8", "aria-label": "Edit this tag",
-            data: { turbo: true } %>
+        <div class="flex">
+          <div>
+            <%= pop_icon_to "edit",
+              "#",
+              class: "tooltipped tooltipped--w mr2 h-8 w-8", "aria-label": "Edit this tag",
+              data: { behavior: "modal_trigger", modal: "edit_tag_#{tag.id}" } %>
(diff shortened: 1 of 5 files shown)


Luke-Oldenburg:
this is cool :eyes:

sampoder:
Thanks for making this! I think we should split the card into one card for pre-existing tags and another card for the form to create a new tag. I also we should vertically the stack the form fields, just so it's in the same format as other tag-related forms we have.

app/views/events/settings/_tags.html.erb
```diff
@@ -0,0 +1,67 @@
...
+            'aria-label': "Delete this tag" %>
+        </div>
+      </div>
+    <% end %>
+    <h3 class="mb1">Create a new tag</h3>
+    <%= form_with(url: event_tags_path(@event)) do |form| %>
+      <div class="flex gap-5 justify-between items-end">
+        <div>
+          <%= form.label :label, "Tag name" %>
+          <%= form.text_field :label %>
```
sampoder: Let's add a placeholder

app/views/events/settings/_tags.html.erb
```diff
@@ -0,0 +1,67 @@
...
+          <% Tag::COLORS.each do |color| %>
+            <label class="tags__radio w-8 mt-[2px] mx-1">
+              <%# Default color: Muted %>
+              <%= form.radio_button :color, color, checked: color == "muted" %>
+              <div class="radio__control tag-darker tag-<%= color %>"></div>
+            </label>
+          <% end %>
+        </div>
+
+        <%= form.submit "Create tag", class: "h-10 m-1" %>
```
sampoder: The centring on this button feels a little off

sampoder (approved):
LGTM! I made a couple of small edits to the styling

Screenshot 2025-05-05 at 11 24 00 PM

PR #10306, opened 2025-05-02
Add a hide links option to the `user_mention` helper

sampoder:
such a dumb bug - ugh.


sampoder:
hmm I'm going to fix this another way

PR #10309, opened 2025-05-02
Fix organizer position policy

Luke-Oldenburg:
## Summary of the problem
Managers could demote and remove other managers. Same with contract signees and other contract signees.

## Describe your changes
It checks if you are currently a manager and if the user you are attempting to modify is not a manager.
Same process for contract signee check.
change_position_role? and can_request_removal? have been updated.

### DIFF
--- a/app/policies/organizer_position_policy.rb
+++ b/app/policies/organizer_position_policy.rb
@@ -27,10 +27,7 @@ class OrganizerPositionPolicy < ApplicationPolicy
   end
 
   def can_request_removal?
-    return true if record.user == user
-    return true if user&.admin?
-    return true if OrganizerPosition.find_by(user:, event: record.event)&.manager? && !OrganizerPosition.find_by(user: record.user, event: record.event)&.manager?
-    return true if OrganizerPosition.find_by(user:, event: record.event)&.is_signee && !OrganizerPosition.find_by(user: record.user, event: record.event)&.is_signee
+    admin_or_manager? || record.user == user
   end
 
   def view_allowances?
--- a/app/policies/organizer_position_policy.rb
+++ b/app/policies/organizer_position_policy.rb
@@ -21,11 +21,16 @@ class OrganizerPositionPolicy < ApplicationPolicy
     return false unless user
     return false if record.user == user
 
-    admin_or_manager?
+    return true if user&.admin?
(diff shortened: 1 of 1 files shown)


app/policies/organizer_position_policy.rb
```diff
@@ -21,11 +21,16 @@ def change_position_role?
...
+    return true if OrganizerPosition.find_by(user:, event: record.event)&.is_signee && !OrganizerPosition.find_by(user: record.user, event: record.event)&.is_signee
   end
 
   def can_request_removal?
-    admin_or_manager? || record.user == user
+    return true if record.user == user
+    return true if user&.admin?
+    return true if OrganizerPosition.find_by(user:, event: record.event)&.manager? && !OrganizerPosition.find_by(user: record.user, event: record.event)&.manager?
+    return true if OrganizerPosition.find_by(user:, event: record.event)&.is_signee && !OrganizerPosition.find_by(user: record.user, event: record.event)&.is_signee
   end
```
sampoder: I think we should keep the existing setup - these are reviewed by a human so it's alright if we're loose on who can request removal

Luke-Oldenburg: Good point

app/policies/organizer_position_policy.rb
```diff
@@ -21,7 +21,9 @@ def change_position_role?
     return false unless user
     return false if record.user == user
 
-    admin_or_manager?
+    return true if user&.admin?
+    return true if OrganizerPosition.find_by(user:, event: record.event)&.manager? && !OrganizerPosition.find_by(user: record.user, event: record.event)&.manager?
+    return true if OrganizerPosition.find_by(user:, event: record.event)&.is_signee && !OrganizerPosition.find_by(user: record.user, event: record.event)&.is_signee
```
sampoder, line 26: I think I'm comfortable with managers demoting managers - a validation already prevents signees from dropping from manager

Luke-Oldenburg: It seems strange imo. This was to prevent signee from dropping other signee too. Not manager dropping signee.

sampoder: It should be impossible for signees to not be managers: https://github.com/hackclub/hcb/blob/3c81e4e91cc5626895620785539e11b71005d6d1/app/models/concerns/organizer_position/has_role.rb#L21

Luke-Oldenburg: Ah I understand now

PR #10310, opened 2025-05-02
Grammar and info fixes on roles page

Luke-Oldenburg:
https://hcb.hackclub.com/roles
It was missing readers and also said "an HCB" instead of "a HCB"

### DIFF
--- a/app/views/static_pages/roles.html.erb
+++ b/app/views/static_pages/roles.html.erb
@@ -22,7 +22,7 @@
       <strong>Members</strong> can <em>spend</em>, but not <em>send</em> money,
       and can't edit the organization's settings.
     </li>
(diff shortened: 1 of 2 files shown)


app/views/static_pages/roles.html.erb
```diff
@@ -22,6 +22,8 @@
       <strong>Members</strong> can <em>spend</em>, but not <em>send</em> money,
       and can't edit the organization's settings.
     </li>
+    <li><strong>Readers</strong> can only view basic information on an organization.
```
sampoder:
```suggestion
    <li><strong>Readers</strong> have read-only access to an organization.
```

sampoder (approved):
LGTM

PR #10314, opened 2025-05-03
show roles on team tabs for eveyone

anscg:
## Summary of the problem

https://hackclub.slack.com/archives/C068U0JMV19/p1746288968205959
user roles are hidden on organization cards if you're not a member, even though you can still filter by roles

## Describe your changes

displays user roles on organization for everyone


sampoder (approved):
lgtm! thank you

PR #10311, opened 2025-05-03
[Policy] Allow grant card holders to access stripe card info 

Mohamad-Mortada:
## Summary of the problem
For the v4 API you can't use v4/cards/crd_xxxxxxx for grant cards that you own. This policy fixes it by also allowing the card holder to have access.

PR #10315, opened 2025-05-03
Use cardholder? instead of record.stripe_cardholder value

Mohamad-Mortada:


Luke-Oldenburg (commented):
Might be missing something but I don't see a cardholder? function?

Mohamad-Mortada:
@sampoder requested this change for this policy so I'm assuming it should apply here? I may be wrong tho, just doing this for sake of consistency :) https://github.com/hackclub/hcb/pull/10311

app/policies/hcb_code_policy.rb
```diff
@@ -2,7 +2,7 @@
 
 class HcbCodePolicy < ApplicationPolicy
   def show?
-    user&.auditor? || present_in_events? || (record.stripe_cardholder.present? && record.stripe_cardholder.user == user)
+    user&.auditor? || present_in_events? || cardholder?
```
sampoder, line 5: Cardholder doesn't seem to be defined in `HcbCodePolicy`, it's defined here: https://github.com/thedev132/hcb/blob/04aaf858e636c316bc2755be639d7579672486f2/app/policies/stripe_card_policy.rb#L62 in `StripeCardPolicy`

Mohamad-Mortada: ohhhhh I see.

PR #10320, opened 2025-05-04
Fix stripe cards count on event card

Luke-Oldenburg:
Removes card grants from count to match the cards page on the event


sampoder (approved):
LGTM!

PR #10325, opened 2025-05-05
Update reader roles

Luke-Oldenburg:


Luke-Oldenburg:
Found some buggy behavior in general with readers while testing for this. Might go ahead and fix that in this PR so I can fully confirm the table is correct.
- [x] check deposits are buggy (not showing up in admin page when submitted by manager; unauthorizes for readers instead of disabled)
- [x] account number content missings for readers instead of unauthorized
- [x] order card unauthorizes for readers instead of disabled
- [x] google workspace makes html error for readers

sampoder:
What does this mean:

> not showing up in admin page when submitted by manager

That seems weird

Luke-Oldenburg:
@sampoder thinking dev mode was being weird for me but I couldn't thoroughly test reader permissions without that working.

Luke-Oldenburg:
Added an error message to check deposits failing and restricted the file types. No clue why they're still failing even with the correct file types though. Also made the form disabled for readers.

Luke-Oldenburg:
There's no good way to disable the cards form because of the way you can select events from your personal cards page so I'm just going to check that off and skip it.

Luke-Oldenburg:
Google Workspace error was unrelated to readers. Just a dev mode thing I think.

app/policies/event_policy.rb
```diff
@@ -117,6 +117,10 @@ def new_transfer?
     admin_or_manager? && !record.demo_mode?
   end
 
+  def create_check_deposit?
```
sampoder: We should be using `CheckDepositPolicy#create?` - and doing something like `policy(@event.check_deposits.build).create?`

app/policies/event_policy.rb
```diff
@@ -145,6 +149,10 @@ def reimbursements?
     auditor_or_reader? && record.plan.reimbursements_enabled?
   end
 
+  def create_reimbursement?
```
sampoder: Same comment as above

app/views/check_deposits/index.html.erb
```diff
@@ -34,7 +34,7 @@
   <div class="table-container">
     <table>
       <thead>
-        <% if @check_deposits.any? { |cd| policy(cd).view_image? } %>
```
sampoder: What's the context behind this change?

app/controllers/check_deposits_controller.rb
```diff
@@ -21,6 +21,11 @@ def create
     check_deposit.save!
 
     redirect_to url_for(check_deposit.local_hcb_code), flash: { success: "Your check deposit is on the way!" }
+
+  rescue => e
+    Rails.error.report(e)
+
+    redirect_to event_check_deposits_path(@event), flash: { error: e.message }
```
sampoder, line 28: (also responding to your DM here) Maybe we can change from using `processable` to using `content_type: [:png, :jpeg]` like we do on `Event`:

```ruby
validates :stripe_card_logo, content_type: [:png, :jpeg]
```

I think that would achieve the same limits we need for images to work with Column

Luke-Oldenburg: That would work. Do we actually need to be checking processable?

app/views/check_deposits/index.html.erb
Luke-Oldenburg: Readers should be able to view the check deposit but not the check images.

app/policies/event_policy.rb
Luke-Oldenburg: How do I handle this with reimbursements since they don't have their own policy? Reports and expenses are broken apart.

app/views/check_deposits/index.html.erb
sampoder: Yes - but this code is hiding / showing the column with the check images

app/policies/event_policy.rb
sampoder: You can use the report policy

Luke-Oldenburg: Would it be `policy(@event.reimbursements.build).create?` still or am I missing something to get the report itself?

app/controllers/check_deposits_controller.rb
sampoder: PNGs and JPEGs should be guaranteed to processable

app/views/check_deposits/index.html.erb
Luke-Oldenburg: Oh I thought it was the entire check deposit view there. I had put a show image around the image itself. I'll swap that back.

app/policies/event_policy.rb
sampoder: `policy(@event.reimbursement_reports.build).create?`

app/views/reimbursement/reports/_create_form.html.erb
```diff
@@ -83,7 +83,7 @@
       <% end %>
     <% end %>
     <div class="tooltipped tooltipped--e inline-block" aria-label="<%= @event&.demo_mode? ? "You don't have access to this feature at the moment." : "This will start your reimbursement report." %>">
-      <%= form.submit "Start report", class: "mt2", disabled: @event && (!organizer_signed_in? || @event&.demo_mode?) %>
+      <%= form.submit "Start report", class: "mt2", disabled: !policy(@event.reimbursement_reports.build).create? %>
```
sampoder:
```suggestion
      <%= form.submit "Start report", class: "mt2", disabled: @event && !policy(@event.reimbursement_reports.build).create? %>
```

sampoder (approved):
LGTM!

PR #10009, opened 2025-03-30
 Un-hardcode some SMTP/Email settings

cyteon:
Recreation of #9958 cause github disassociated my fork from this repo :(

This PR unhardcodes the selection between letter opener or SMTP on dev, by making some env vars.
- If SMTP__ENABLED is false or not set, it will use letter opener

It also un-harcodes the emails in app/mailers/application_mailer.rb!
- If SMTP__DOMAIN is not set, it falls back to hcb.hackclub.com for prod, and staging.hcb.hackclub.com for dev
- If SMTP__USERNAME is not set, it falls back to "hcb"

PR #10330, opened 2025-05-06
Revert " Un-hardcode some SMTP/Email settings"

sampoder:
Reverts hackclub/hcb#10009


sampoder:
This broke our production setup - at this time, I'm just going to keep it the same.

PR #10331, opened 2025-05-06
Disable/hide the help button and command menu for logged out users

anscg:
## Summary of the problem

The help button doesn't work for logged out users (which i assume it's intended) but they're shown while visiting a transparent org.

The command menu doesn't make much sense to a logged out user since most of the option requires them to be logged in, also taking over the cmd+k keybind on a simple static page visit can be an unexpected and potentially disruptive user experience esp while it's a default chrome shortcut for searching.


sampoder (approved):
LGTM!

PR #10335, opened 2025-05-07
[dev-docs::OAuth] Add instructions for scopes and confidential checkbox.

Luke-Oldenburg:


dev-docs/development.md
```diff
@@ -200,7 +200,7 @@ We've transitioned to using development keys and seed data in development, but h
 
 There is two different ways you can accomplish the first step of getting an OAuth token, either using a webpage, or the terminal depending on your preference.
 
-1. Go to [http://localhost:3000/api/v4/oauth/applications](http://localhost:3000/api/v4/oauth/applications). Press "New Application" and then set the name to anything of your choosing and the redirect URI to `http://localhost:3000/`. Press "Submit" and then save the info on the new page that appears and press "Authorize."
+1. Go to [http://localhost:3000/api/v4/oauth/applications](http://localhost:3000/api/v4/oauth/applications). Press "New Application" and then set the name to anything of your choosing, the redirect URI to `http://localhost:3000/`, and scopes to `read write`. For the purposes of this guide, you should leave confidential checked. Press "Submit" and then save the info on the new page that appears and press "Authorize."
```
sampoder: Hmm - I feel like we should be offering context behind the field: https://oauth.net/2/client-types/. Things like the HCB mobile app and the Raycast extension have `confidential: false`

Luke-Oldenburg: It can be false it's just that I instructed them to include the secrets later in the guide. I haven't tested it without confidential so I wasn't sure which fields you could safely ignore when generating the token.

sampoder (approved):
LGTM!

PR #10346, opened 2025-05-07
Check if Grant card flipper is enabled in v4 org API

Mohamad-Mortada:
I want to be able to check if grant cards are supported in a specific organization before allowing organizers to create grant cards in hcb mobile

### DIFF
--- a/app/views/api/v4/events/_event.json.jbuilder
+++ b/app/views/api/v4/events/_event.json.jbuilder
@@ -11,6 +11,7 @@ json.playground_mode_meeting_requested event.demo_mode_request_meeting_at.presen
 json.transparent event.is_public?
 json.fee_percentage event.revenue_fee.to_f
 json.background_image event.background_image.attached? ? Rails.application.routes.url_helpers.url_for(event.background_image) : nil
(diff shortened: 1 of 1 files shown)


Luke-Oldenburg (approved):

sampoder (commented):
I would like to think about how we want to structure the API more here - I don't want this endpoint to grow into a never ending list of fields. It might make sense to have a FlipperController in the API.

Luke-Oldenburg:
How exactly should the flipper controller work? I could implement something more complicated like that. I feel like for fields such as this though listing it on the event is fine for now.

Luke-Oldenburg:
If we separate it to another endpoint, that's an additional request being made here to determine the status of card grants. Unless we have a way for that new flipper controller to return all the flippers for a given event and then this event view will just pull from that.

sampoder:
We could do something like `events/features` route to access enabled Flippers. Or we could do something like `/features/:flag_id/:entity_id` that returns a true / false.

Luke-Oldenburg:
What about `/events/:event_id/features` since it's serving as a type of index function?

sampoder:
Oh yes good point - that's what I meant

PR #10356, opened 2025-05-08
[Filters] Remove delete and add check to tags

### LINKED ISSUE #10344: Tag filters don't show a checkmark when applied unlike other filters
### Describe your goals

Apply a tag filter.

### Describe the bug

It did not have a check mark in the list to indicate which filter was selected.

Image
Image

Luke-Oldenburg:
Delete button is no longer necessary with new tags page. There was also no check to indicate which tag was selected to filter by.

Closes #10344


sampoder (approved):
LGTM!

sampoder:
We should probably get the word out about the new tags page, @polypixeldev would you like to make a changelog post?

polypixeldev:
sure! are we going to be taking tags out of beta as well?

PR #10358, opened 2025-05-08
[Cards] Reset page number to 1 after applying filter

### LINKED ISSUE #10341: Adding filter on cards page doesn't reset page number
### Describe your goals

Add a filter while on the last page of HQ cards.

### Describe the bug

The page number did not reset so it displayed the error message "You don't have any frozen cards yet." It should return you back to page 1. 

[Link](https://hcb.hackclub.com/hq/cards?page=11&status=frozen&view=list) that is created after applying a filter from the final page.

[Link](https://hcb.hackclub.com/hq/cards?status=frozen&view=list) that should be created after applying a filter from a page other than page 1.

Luke-Oldenburg:
Closes #10341

P.S. I updated the tooltip text to make it more consistent with the other filter tooltips in the UI.

### DIFF
--- a/app/views/events/_filter_cards_menu.html.erb
+++ b/app/views/events/_filter_cards_menu.html.erb
@@ -2,7 +2,7 @@
   <button
     type="button"
     class="pop my-2 cursor-pointer tooltipped tooltipped--s"
-    aria-label="Add filters..."
+    aria-label="Add filter..."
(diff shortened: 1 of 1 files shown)


Luke-Oldenburg:
Now that I think about it they really had all ought to say "Add filters..." instead of "Add filter..."

sampoder (approved):
LGTM! Thanks for the copy edit as well

PR #10366, opened 2025-05-08
Specificy shift clicking on rename transaction preview text

Luke-Oldenburg:
before:
Screenshot From 2025-05-08 15-01-28


sampoder (approved):
LGTM!

PR #10367, opened 2025-05-08
[Card Grants] Properly handle overwithdrawal error

### LINKED ISSUE #10363: Overwithdrawaling from card grant displays error page instead of banner
### Describe your goals

Withdrawal more money than was on a card grant

### Describe the bug

It redirected to a generic error page instead of showing a warning banner

Image

Luke-Oldenburg:
Closes #10363
Screenshot From 2025-05-08 15-09-58

### DIFF
--- a/app/controllers/card_grants_controller.rb
+++ b/app/controllers/card_grants_controller.rb
@@ -136,6 +136,11 @@ class CardGrantsController < ApplicationController
     @card_grant.withdraw!(amount_cents: Monetize.parse(params[:amount]).cents, withdrawn_by: current_user)
 
     redirect_to @card_grant, flash: { success: "Successfully withdrew from grant." }
+
+  rescue => e
+    Rails.error.report(e)
+
+    redirect_to @card_grant, flash: { error: e.message }
   end
 
   def edit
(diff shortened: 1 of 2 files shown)


app/controllers/card_grants_controller.rb
```diff
@@ -136,6 +136,11 @@ def withdraw
     @card_grant.withdraw!(amount_cents: Monetize.parse(params[:amount]).cents, withdrawn_by: current_user)
 
     redirect_to @card_grant, flash: { success: "Successfully withdrew from grant." }
+
+  rescue => e
+    Rails.error.report(e)
```
sampoder: I don't think we should be reporting `ArgumentError`s that we're raising ourselves

sampoder: We should be reporting other errors though - sorry should have been clearer there

Luke-Oldenburg: Oh lol. How can I detect if it is an error we are raising ourselves?

sampoder: I think you can do something like `unless e.is_a?(ArgumentError)` (I think - test this)

Luke-Oldenburg: Tested it. I didn't really notice it logging anything different either way. I tried searching in the terminal for the string that is flashed in the UI.

Luke-Oldenburg: It didn't break anything though.

sampoder:
Error reporting happens in production

PR #10370, opened 2025-05-08
Make contract signature requirement configurable on `Event::Plan` + create `HackClubHQ` plan

### LINKED ISSUE #10368: Can't edit HQ orgs without contract signees
### Describe your goals

I was trying to change the name of a HQ YSWS org. As it's a HQ org, they don't have contract signees.

### Describe the bug

Image

I can't change it's name because it doesn't have any contract signees, and I can't auto-override this like we used to be able to.

Luke-Oldenburg:
Closes #10368


app/models/event.rb
```diff
@@ -764,7 +764,7 @@ def demo_mode_limit
   end
 
   def contract_signed
-    return if organizer_position_contracts.signed.any? || organizer_position_contracts.none?
+    return if organizer_position_contracts.signed.any? || organizer_position_contracts.none? || plan.is_a?(Event::Plan::HackClubAffiliate)
```
sampoder: This should be a method on plan like `requires_contract_signature?` or something

Luke-Oldenburg: We are going with this though? Of not requiring HQ to have contract?

sampoder: Oh I'm not sure - just if we do - I'd suggest that change

PR #10380, opened 2025-05-09
Show add tag button on grant page transactions

polypixeldev:
## Summary of the problem

It's currently a little difficult to add tags on a grant card transaction as an organizer - you have to go to the grant page and then go to the hcb code page (not popovers) to add a tag.

## Describe your changes

This makes it a little easier by adding the add tag button to the grant page.


sampoder:
Could you attach some screenshots to this PR? Thanks! I'm a little confused by the context - are these organisers tagging transactions that happened on a card grant card?

polypixeldev:
Yeah, organizers tagging transactions on a card grant - here's a screenshot
image

sampoder:
Is this something that has been commonly requested? I guess I'm struggling to see how these be used

Luke-Oldenburg:
I think it's good to have for the sake of consistency. Ive been using tags to mark whether I thought a transaction was approved spending or not for apex grants. It's annoying to have to open the individual hcb code pages from the subledger rather than just add it from the grant page.

sampoder:
Oh interesting - I didn't know this was a use case of tags - good to know

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -76,7 +76,7 @@
         <% end %>
       </div>
 
-      <% if !defined?(receipt_upload_button) && !defined?(hide_tags) && organizer_signed_in? && !pt.instance_of?(OpenStruct) && Flipper.enabled?(:transaction_tags_2022_07_29, @event || pt.local_hcb_code.event) %>
+      <% if !@frame && !defined?(hide_tags) && organizer_signed_in? && !pt.instance_of?(OpenStruct) && Flipper.enabled?(:transaction_tags_2022_07_29, @event || pt.local_hcb_code.event) %>
```
sampoder: What's the reason for adding `@frame`?

polypixeldev: The reason for adding `@frame` was because the add tag button was already hidden on the HcbCode popover, since the menu does not display on top of the modal. If I increase the z-index of `.menu__content` to > 200 though, it seems like it works. Should I do that in this or a separate PR?

sampoder: Cool - could we make it opt-in instead of using `hide_tags`?

sampoder: I'm worried we'll miss a case where we don't want to show tags

polypixeldev: Why would we not want to show tags somewhere else? I think it makes sense to show tags by default - from what I can tell, `hide_tags` is only used in a couple places where `@event` isn't defined (which breaks the create tag modal), and that should be fixed soon when it's changed to be a local instead of an instance variable

sampoder: I'm not sure off the top of my head but that's why I'm worried. Certainly shouldn't be on the receipt bin page.

sampoder (approved):
thank you!

PR #10381, opened 2025-05-09
Allow readers to view settings

Luke-Oldenburg:


app/views/events/settings/_card_grants.html.erb
Luke-Oldenburg, line 1: I couldn't get expires after to be disabled after applying it so I left it alone.

sampoder: I think my latest commit should have done this

sampoder: See https://stackoverflow.com/questions/61642122/trying-to-disable-the-whole-f-select-in-rails

sampoder (approved):
LGTM!

PR #10383, opened 2025-05-09
Fix `GSuite::Revocation` issues

rluodev:
Fixed mailers and `GSuiteAccount` Policy to account for the fact that we're passing in a GSuite to GSuiteAccountPolicy sometimes.


app/policies/g_suite_account_policy.rb
```diff
@@ -40,7 +40,11 @@ def toggle_suspension?
   private
 
   def admin_or_manager?
-    user&.admin? || (OrganizerPosition.find_by(user:, event: record.event)&.manager? && !record.g_suite.revocation.present?)
+    user&.admin? || (
+      OrganizerPosition.find_by(user:, event: record.event)&.manager? && !(
+        record.is_a?(GSuite) ? (record&.revocation.present? && record&.revocation.revoked?) : (record&.g_suite&.revocation.present? && record&.g_suite&.revocation.revoked?)
+      )
+    )
```
garyhtou:
```suggestion
    return true if user&.admin?

    revocation = record.is_a?(GSuite) ? record.revocation : record&.g_suite&.revocation
    OrganizerPosition.find_by(user:, event: record.event)&.manager? && !revocation&.revoked?
```

garyhtou (approved):
tested by Ruien

PR #10396, opened 2025-05-11
Remove instance variables from partials (Part 1)

Luke-Oldenburg:


YodaLightsabr:
:pray: Thank you for this! It seems like there are a few errors and a conflict with the disbursements form, but ping me once you're ready for a review.

PR #10399, opened 2025-05-12
Add link to event in receipt bin list

polypixeldev:
## Summary of the problem

I was in the receipt bin and thought it would be useful to link to the event there!

## Describe your changes

Changed the event name text to be a link (without an underline, matching the stripe card link next to it)

image
image


sampoder (approved):
Sounds good to me - thanks!

PR #10404, opened 2025-05-13
Make donation Gravatars more random

skyfallwastaken:
## Summary of the problem

email.to_i looks like it takes the leading numbers in the string, im guessing none of them start with numbers and so it's always zero.


sampoder:
Hmm - I think try making a new branch this is showing zero LOC changed

skyfallwastaken:
I love git 😔 
Anyway, new PR: https://github.com/hackclub/hcb/pull/10405

PR #10414, opened 2025-05-13
Add inactive filter for cards

Luke-Oldenburg:
You can now filter by inactive cards separately from frozen cards!
Screenshot From 2025-05-13 15-40-29

### DIFF
--- a/app/models/stripe_card.rb
+++ b/app/models/stripe_card.rb
@@ -68,8 +68,9 @@ class StripeCard < ApplicationRecord
 
   scope :deactivated, -> { where.not(stripe_status: "active") }
   scope :canceled, -> { where(stripe_status: "canceled") }
-  scope :frozen, -> { where(stripe_status: "inactive") }
+  scope :frozen, -> { where(stripe_status: "inactive", initially_activated: true) }
   scope :active, -> { where(stripe_status: "active") }
+  scope :inactive, -> { where(stripe_status: "inactive", initially_activated: false) }
   scope :physical_shipping, -> { physical.includes(:user, :event).reject { |c| c.stripe_obj[:shipping][:status] == "delivered" } }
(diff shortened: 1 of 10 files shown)


app/models/stripe_card.rb
```diff
@@ -68,8 +68,9 @@ class StripeCard < ApplicationRecord
 
   scope :deactivated, -> { where.not(stripe_status: "active") }
   scope :canceled, -> { where(stripe_status: "canceled") }
-  scope :frozen, -> { where(stripe_status: "inactive") }
+  scope :frozen, -> { where(stripe_status: "inactive", initially_activated: true) }
   scope :active, -> { where(stripe_status: "active") }
+  scope :inactive, -> { where(stripe_status: "inactive", initially_activated: false) }
```
sampoder, line 73: We use `.frozen` and `.frozen?` in a couple of places - can you update them to match this change? Eg. this chunk of code from the API needs it's documentation updated:

```ruby
expose :status, documentation: {
    type: "string", values: %w[active frozen canceled]
  } do |stripe_card, options|
  next "frozen" if stripe_card.frozen? # rename 'inactive' to 'frozen'

  stripe_card.stripe_status
end
```

Luke-Oldenburg: Correct me if I'm wrong but I think the API is the only place I missed with updating the values array.

sampoder: I think we'd also want to update `cpt&.stripe_card&.frozen?` to be `cpt&.stripe_card&.frozen? || cpt&.stripe_card&.inactive?`. That was the sort of thing I was referring to.

Luke-Oldenburg: Ah I get it now

sampoder (approved):
LGTM - thank you!

Luke-Oldenburg:
Card grants would never be inactive because they're virtual cards btw

PR #10416, opened 2025-05-13
 Add a "share donor graph" button on transparent orgs 

skyfallwastaken:
## Summary of the problem
https://graph.hcb.hackclub.com is a thing now! However, it's not really advertised anywhere except in the #hcb Slack thread, so users won't even know it exists :P

## Describe your changes

Adds a modal to copy a donor graph URL. The button only gets shown to orgs with transparency mode turned on.

image 
image


sampoder (approved):
LGTM!

PR #10419, opened 2025-05-14
Allow for members to view disbursement form

Luke-Oldenburg:
It redirects as unauthorized for members currently but the page should load with fields disabled.

### DIFF
--- a/app/policies/disbursement_policy.rb
+++ b/app/policies/disbursement_policy.rb
@@ -5,7 +5,7 @@ class DisbursementPolicy < ApplicationPolicy
     user.auditor?
   end
 
-  def can_send?(role: :manager)
+  def can_send?
     return true if user&.admin?
     return true if record.source_event.nil?
     return true if OrganizerPosition.role_at_least?(user, record.source_event, :manager)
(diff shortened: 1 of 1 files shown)


app/policies/disbursement_policy.rb
```diff
@@ -23,7 +23,7 @@ def can_receive?(role: :manager)
   end
 
   def new?
-    can_send?(role: :reader) && can_receive?(role: :reader)
```
sampoder, line 26: I think the fix here is to pass role into:

```
OrganizerPosition.role_at_least?(user, record.destination_event, :manager)
```

so:

```
OrganizerPosition.role_at_least?(user, record.destination_event, role)
```

Luke-Oldenburg: `can_send?` and `can_receive?` are used for create though. I don't think policies for create and new should be intertwined unless there's something I'm missing.

sampoder: Maybe you're right :/ I was trying to be fancy here and get them to line up as closely as possible. Feel free to go with your original fix but please then rip out the role param

sampoder (approved):
LGTM!

PR #10429, opened 2025-05-14
[Transfers] Default to transaction memo if purpose is blank or empty

Luke-Oldenburg:
Screenshot From 2025-05-14 13-14-14

### DIFF
--- a/app/views/events/transfers.html.erb
+++ b/app/views/events/transfers.html.erb
@@ -132,11 +132,7 @@
             <% if d.is_a?(Disbursement) %>
               <%= d.name %>
             <% elsif d.is_a?(CardGrant) %>
-              <% if !d.purpose || d.purpose == "" %>
-                <%= d.disbursement.local_hcb_code.memo %>
-              <% else %>
-                <%= d.purpose %>
-              <% end %>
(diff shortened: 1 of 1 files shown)


app/views/events/transfers.html.erb
```diff
@@ -132,7 +132,11 @@
             <% if d.is_a?(Disbursement) %>
               <%= d.name %>
             <% elsif d.is_a?(CardGrant) %>
-              <%= d.purpose %>
+              <% if !d.purpose || d.purpose == "" %>
+                <%= d.disbursement.local_hcb_code.memo %>
+              <% else %>
+                <%= d.purpose %>
+              <% end %>
```
sampoder:
```suggestion
              <%= d.presence || d.disbursement.local_hcb_code.memo %>
```
Check out `Object#presence`: https://stackoverflow.com/questions/19637499/what-is-the-point-of-objectpresence-in-rails

Luke-Oldenburg: Cool! I didn't know about that. d.purpose.presence because d is a card grant.

sampoder: Ah - yes good catch

PR #10434, opened 2025-05-14
[Login] Backup Codes

### LINKED ISSUE #6707: [Login] Create back up codes

rluodev:
## Summary of the problem

People might get locked out of their accounts with no way to get back in, which is bad.

## Describe your changes

Added backup codes, securely generated, SHA-512 BLT seasoned with fresh S&P.

Screenshot 2025-05-24 at 11 11 25 PM

Screenshot 2025-05-24 at 11 11 48 PM

Screenshot 2025-05-24 at 11 07 44 PM

Screenshot 2025-05-24 at 11 12 30 PM

Screenshot 2025-05-24 at 11 12 55 PM

Screenshot 2025-05-24 at 11 10 55 PM

Screenshot 2025-05-24 at 11 24 17 PM

Screenshot 2025-05-24 at 11 27 51 PM

Screenshot 2025-05-24 at 11 29 08 PM

Screenshot 2025-05-24 at 11 23 22 PM

Screenshot 2025-05-24 at 11 23 48 PM

Screenshot 2025-05-24 at 11 25 59 PM


app/mailers/user/backup_code_mailer.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+class User
+  class BackupCodeMailer < ApplicationMailer
+    before_action :set_user
+
+    default to: -> {
+      emails = @user.email_address_with_name
+    }
```
garyhtou:
```suggestion
    default to: -> { @user.email_address_with_name }
```

app/models/login.rb
```diff
@@ -59,7 +59,7 @@ class Login < ApplicationRecord
     event :mark_complete do
       transitions from: :incomplete, to: :complete do
         guard do
-          authentication_factors_count == (user.use_two_factor_authentication? ? 2 : 1)
+          authentication_factors_count >= (user.use_two_factor_authentication? && !authenticated_with_backup_code ? 2 : 1)
```
garyhtou:
```suggestion
          authentication_factors_count == (user.use_two_factor_authentication? ? 2 : 1) || authenticated_with_backup_code
```

app/models/user.rb
```diff
@@ -370,6 +372,61 @@ def only_card_grant_user?
...
+
+    codes
+  end
+
+  def redeem_backup_code(code)
+    found = nil
+    pepper = Credentials.fetch(:BACKUP_CODE_PEPPER)
+    # make sure we do not short circuit
+    unused_backup_codes.each do |backup_code|
+      hash = OpenSSL::KDF.pbkdf2_hmac(code + pepper, hash: "sha512", salt: Base64.decode64(backup_code.salt), iterations: 20_000, length: 64).unpack1("H*")
```
garyhtou: Create a class method on BackupCode for calculating a hash.

app/models/user.rb
```diff
@@ -370,6 +372,61 @@ def only_card_grant_user?
...
+      if ActiveSupport::SecurityUtils.secure_compare(hash, backup_code.hash)
+        found = backup_code
+      end
+    end
```
garyhtou: If someone is brute forcing, the loop will already run its entirety. And is someone puts in a valid backup code, that means they've already gained access and we don't need to keep running the loop to prevent leaking timing information.

Let's instantly return when found.

app/models/user.rb
```diff
@@ -370,6 +372,61 @@ def only_card_grant_user?
...
+
+    if found.present?
+      found.transaction do
+        found.mark_used!
+      end
```
garyhtou: Since we're only doing one write here, we don't need a transaction
```suggestion
      found.mark_used!
```

app/models/user.rb
```diff
@@ -370,6 +372,61 @@ def only_card_grant_user?
...
+      BackupCodeMailer.with(user_id: id).code_used.deliver_now
```
garyhtou: I would recommend making this a callback on mark_used

app/models/user.rb
```diff
@@ -370,6 +372,61 @@ def only_card_grant_user?
...
+
+      return true
+    end
+    false
+  end
+
+  def disable_backup_codes
+    backup_codes.where.not(aasm_state: [:used, :invalidated]).find_each &:mark_invalidated!
```
garyhtou: I'd suggest

```ruby
def invalidate_backup_codes
  backup_codes.where(aasm_state: :unsaved).destroy_all
  backup_codes.where(aasm_state: :unused).find_each &:mark_invalidated!
end
```

db/migrate/20250513051747_create_user_backup_codes.rb
```diff
@@ -0,0 +1,13 @@
+class CreateUserBackupCodes < ActiveRecord::Migration[7.2]
+  def change
+    create_table :user_backup_codes do |t|
+      t.references :user, null: false, foreign_key: true
+      t.string :aasm_state
+      t.text :hash, null: false
```
garyhtou: Add a unique index

db/migrate/20250513051747_create_user_backup_codes.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.text :salt, null: false
+      t.datetime :deleted_at
```
garyhtou: Drop acts as parnoid and this column

app/models/user/backup_code.rb
```diff
@@ -0,0 +1,64 @@
...
+      state :invalidated
+
+      event :mark_unused do
+        transitions from: :unsaved, to: :unused
+      end
+      event :mark_used do
+        transitions from: :unused, to: :used
+
+        after do
+          case user.backup_codes.where(aasm_state: :unused).count
```
garyhtou:
```suggestion
          case user.backup_codes.unused.size
```

app/models/user/backup_code.rb
```diff
@@ -0,0 +1,64 @@
...
+          when 0
+            User::BackupCodeMailer.with(user_id: user.id).no_codes_remaining.deliver_now
+          when 1..2
```
garyhtou:
```suggestion
          when 1..3
```

app/controllers/logins_controller.rb
```diff
@@ -204,6 +210,7 @@ def set_available_methods
     @sms_available = @user&.phone_number_verified && !@login.authenticated_with_sms
     @webauthn_available = @user&.webauthn_credentials&.any? && !@login.authenticated_with_webauthn
     @totp_available = @user&.totp.present? && !@login.authenticated_with_totp
+    @backup_code_available = @user&.unused_backup_codes&.any? && !@login.authenticated_with_backup_code
```
sampoder: In theory - `@login.authenticated_with_backup_code` should always be false, right? As if `authenticated_with_backup_code` is true, the transition to `mark_complete` must have happened.

app/models/user/backup_code.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    belongs_to :user
+
+    validates :code_hash, presence: true, uniqueness: true
+
+    aasm do
+      state :unsaved, initial: true
+      state :unused
+      state :used
+      state :invalidated
```
sampoder: `unsaved` feels confusing to me, because it is saved in the DB. Thoughts on using `previewed`? Which we use in `MailboxAddress`.

Similarly, I'd advocate for `active` instead of `unused`. I was a bit confused because I originally read `unused` as meaning that the user had invalidated the code without using it.

I also appreciate the `discarded` state on `MailboxAddress` (equivalent to `invalidated` here) but I feel less strongly about this one.

app/controllers/users_controller.rb
```diff
@@ -189,6 +189,30 @@ def disable_totp
...
+    @unsaved_backup_codes = @user.generate_backup_codes!
+  end
+
+  def activate_backup_codes
+    @user = params[:id] ? User.friendly.find(params[:id]) : current_user
+    authorize @user
+    previously_enabled = @user.unused_backup_codes&.any?
+    @user.backup_codes&.unused&.map(&:mark_invalidated!) if previously_enabled
+    @user.backup_codes&.unsaved&.map(&:mark_unused!)
+    User::BackupCodeMailer.with(user_id: @user.id).new_codes_generated.deliver_now
```
sampoder: Is this mailer for when backup codes are generated or when they're activated?

app/controllers/users_controller.rb
```diff
@@ -189,6 +189,30 @@ def disable_totp
...
+    authorize @user
+    @unsaved_backup_codes = @user.generate_backup_codes!
+  end
+
+  def activate_backup_codes
+    @user = params[:id] ? User.friendly.find(params[:id]) : current_user
+    authorize @user
+    previously_enabled = @user.unused_backup_codes&.any?
+    @user.backup_codes&.unused&.map(&:mark_invalidated!) if previously_enabled
+    @user.backup_codes&.unsaved&.map(&:mark_unused!)
```
sampoder: See comments about state names - I think that'd make this code more readable.

app/models/user.rb
```diff
@@ -370,6 +372,53 @@ def only_card_grant_user?
...
+      salt = Base64.strict_encode64(SecureRandom.random_bytes(64))
+      begin
+        # backup code pepper must be at least 32 bytes
+        backup_codes.create!(code_hash: User::BackupCode.gen_hash(code:, salt:, pepper:), salt: salt)
+      rescue ActiveRecord::RecordInvalid
+        # if the code is already in use, skip it
+        next
+      end
+      codes << code
+    end
```
sampoder: Is there a reason this isn't included in `before_validate` callbacks on `User::BackupCode` or a similar callback?

app/views/user/backup_code_mailer/backup_codes_disabled.html.erb
```diff
@@ -0,0 +1,16 @@
+<p>
+  Hey <%= @user.first_name %>,
+</p>
+
+<p>
+  You've just disabled backup codes for HCB, and your unused backup codes, if any, were invalidated.
+</p>
+
+<p>
+  If this was you, great! If you didn't do this, please <%= link_to "contact us", "mailto:hcb@hackclub.com?subject=My%20Backup%20Codes%20were%20disabled%20and%20it%20was%20not%20me" %> immediately.
```
sampoder: Recommend using the `mail_to` helper.

app/models/user/backup_code.rb
garyhtou: +1 on using `previewed` and `active`. i don't have a strong feeling for `discarded` vs `invalidated`, but maybe we should do `discarded` for consistency.

app/models/user.rb
rluodev: We need to get the list of unhashed codes to return to show, before_validate means we need to include a new column on BackupCode which is set and then unset after we retrieve the codes for showing them, and we need to omit the column from papertrail if we were to include it in before_validate.

app/controllers/users_controller.rb
rluodev: When we generate codes but the user doesn't save them, we treat it as if it never happened, which means that to the user, the codes are only treated as generated if they confirm they save them, so this is intended behavior.

app/views/users/generate_backup_codes.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= turbo_frame_tag :backup_codes_card do %>
+  <div class="card border b--info mt1 mb3 flex flex-col items-center justify-center">
+    <p>
+      Save the following codes in a safe place. You can use these codes to log in if you lose access to your other 2FA methods. <span class="font-bold">You will not be able to see these codes again.</span>
+    </p>
+    <%= copy_to_clipboard(@unsaved_backup_codes.join("\n")) do %>
+      <pre class="text-lg m-1 p-4 text-center bg-gray-100 rounded mx-auto"><% left_column, right_column = @unsaved_backup_codes.each_slice((@unsaved_backup_codes.size / 2.0).ceil).to_a %><%= left_column.zip(right_column).map { |left, right| "#{left.to_s.ljust(15)}#{right}" }.join("\n").rstrip %></pre>
+    <% end %>
+    <p class="font-bold">These codes have not been activated yet. You must click the below button to confirm you have saved your codes and to activate them.</p>
```
garyhtou:
```suggestion
    <p class="font-bold">These codes have not been activated yet. You must click the button below to confirm you have saved your codes and to activate them.</p>
```

app/views/users/generate_backup_codes.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= turbo_frame_tag :backup_codes_card do %>
+  <div class="card border b--info mt1 mb3 flex flex-col items-center justify-center">
+    <p>
+      Save the following codes in a safe place. You can use these codes to log in if you lose access to your other 2FA methods. <span class="font-bold">You will not be able to see these codes again.</span>
+    </p>
+    <%= copy_to_clipboard(@unsaved_backup_codes.join("\n")) do %>
+      <pre class="text-lg m-1 p-4 text-center bg-gray-100 rounded mx-auto"><% left_column, right_column = @unsaved_backup_codes.each_slice((@unsaved_backup_codes.size / 2.0).ceil).to_a %><%= left_column.zip(right_column).map { |left, right| "#{left.to_s.ljust(15)}#{right}" }.join("\n").rstrip %></pre>
```
garyhtou:
```suggestion
      <pre class="text-lg m-1 p-4 text-center bg-gray-100 rounded mx-auto fs-mask"><% left_column, right_column = @unsaved_backup_codes.each_slice((@unsaved_backup_codes.size / 2.0).ceil).to_a %><%= left_column.zip(right_column).map { |left, right| "#{left.to_s.ljust(15)}#{right}" }.join("\n").rstrip %></pre>
```

garyhtou: This is fine for now, but at some point, it would be nice to switch it to a CSS grid

app/controllers/users_controller.rb
sampoder: This should be more clear semantically - I think that using a consistent set of words (see https://github.com/hackclub/hcb/pull/10434#discussion_r2097161262) would help with this

app/models/user.rb
sampoder: I think you can use https://www.rubyguides.com/2018/11/attr_accessor/ to make it accessible on the object when initially created but never again after that. I think that would be cleaner? It feels weird for this logic to be in a controller to me - @garyhtou thoughts?

app/views/users/generate_backup_codes.html.erb
sampoder: Here's a version of this code using a CSS grid:

```erb
<%= copy_to_clipboard(@unsaved_backup_codes.join("\n")) do %>
    <pre class="text-lg m-1 p-4 text-center bg-gray-100 rounded mx-auto grid grid-cols-2">
    <% @unsaved_backup_codes.each do |code|%>
      <div class="px-6"><%= code %></div>
    <% end %>
  </pre>
<% end %>
```

sampoder: I would also recommend some copy edits:

```erb
<%= turbo_frame_tag :backup_codes_card do %>
  <div class="card border b--info mt1 mb3 flex flex-col items-center justify-center text-center">
    <p>
      Save the following codes in a safe place. <span class="font-bold">You will not be able to see these codes again.</span>
    </p>
    <%= copy_to_clipboard(@unsaved_backup_codes.join("\n")) do %>
        <pre class="text-lg m-1 p-4 text-center bg-gray-100 rounded mx-auto grid grid-cols-2">
        <% @unsaved_backup_codes.each do |code|%>
          <div class="px-6"><%= code %></div>
        <% end %>
      </pre>
    <% end %>
    <p class="font-bold">After saving your backup codes, click the button below to activate them.</p>
    <%= button_to "I have saved my codes", user_activate_backup_codes_path(@user), method: :post, class: "btn px2 h4", data: { turbo: false } %>
  </div>
<% end %>

```

rluodev: thanks, applied!

rluodev: applied

app/views/user/backup_code_mailer/backup_codes_disabled.html.erb
rluodev: done!

app/models/user/backup_code.rb
rluodev: done!

app/controllers/logins_controller.rb
rluodev: Yep, but imo this is good to have for consistency and it doesn't hurt to have it

app/controllers/users_controller.rb
sampoder: `new_codes_generated` should be `new_codes_activated` imo

app/controllers/users_controller.rb
```diff
@@ -189,6 +189,30 @@ def disable_totp
...
+
+  def activate_backup_codes
+    @user = params[:id] ? User.friendly.find(params[:id]) : current_user
+    authorize @user
+    previously_enabled = @user.active_backup_codes&.any?
+    @user.backup_codes&.active&.map(&:mark_discarded!)
+    @user.backup_codes&.previewed&.map(&:mark_active!)
+    User::BackupCodeMailer.with(user_id: @user.id).new_codes_generated.deliver_now
+    User::BackupCodeMailer.with(user_id: @user.id).backup_codes_enabled.deliver_now unless previously_enabled
+    redirect_back_or_to security_user_path(@user)
```
sampoder: imo these two mailers can be combined into one

app/views/user/backup_code_mailer/backup_codes_disabled.html.erb
```diff
@@ -0,0 +1,16 @@
+<p>
+  Hey <%= @user.first_name %>,
+</p>
+
+<p>
+  You've just disabled backup codes for HCB, and your unused backup codes, if any, were invalidated.
+</p>
+
+<p>
+  If this was you, great! If you didn't do this, please <%= mail_to "hcb@hackclub.com", "contact us", subject: "My Backup Codes were disabled and it was not me" %> immediately.
```
sampoder: We should be sentence casing

app/views/users/edit_security.html.erb
sampoder, line 1: Please check with sentence casing

spec/mailers/previews/user/backup_code_mailer_preview.rb
sampoder, line 1: Thank you!

app/controllers/users_controller.rb
rluodev: Should we just go to only having "new codes generated"?

app/views/user/backup_code_mailer/backup_codes_disabled.html.erb
rluodev: done!

app/views/users/edit_security.html.erb
rluodev: done!

app/models/user.rb
garyhtou: Since this is located in the model, I'm cool with how it is. It would be slightly more organized to use attr_accessor, but we'd still need to keep very similar code in the model for generating these codes

app/controllers/users_controller.rb
garyhtou: Yes!

app/views/user/backup_code_mailer/new_codes_generated.html.erb
```diff
@@ -0,0 +1,16 @@
+<p>
+  Hey <%= @user.first_name %>,
+</p>
+
+<p>
+  You've just generated new backup codes for HCB, and have 10 codes available to use if you lose access to other means of accessing your account. Please make sure to save the codes in a safe place. HCB staff will never ask for your backup codes.
```
garyhtou:
```suggestion
  You've just generated new backup codes for HCB, and have 10 codes available to use in case you're unable to authenticate using other means. Please make sure to save the codes in a safe place. HCB staff will never ask for your backup codes.
```

app/controllers/users_controller.rb
garyhtou: +1 on that

app/controllers/users_controller.rb
rluodev: done!

app/controllers/users_controller.rb
rluodev: done!

app/policies/user_policy.rb
```diff
@@ -25,6 +25,18 @@ def disable_totp?
     user.admin? || record == user
   end
 
+  def generate_backup_codes?
+    user.admin? || record == user
+  end
+
+  def activate_backup_codes?
+    user.admin? || record == user
+  end
```
garyhtou:
```suggestion
  def generate_backup_codes?
    record == user
  end

  def activate_backup_codes?
    record == user
  end
```

app/controllers/users_controller.rb
```diff
@@ -189,6 +189,29 @@ def disable_totp
...
+  def generate_backup_codes
+    @user = params[:id] ? User.friendly.find(params[:id]) : current_user
+    authorize @user
+    @previewed_backup_codes = @user.generate_backup_codes!
+  end
+
+  def activate_backup_codes
+    @user = params[:id] ? User.friendly.find(params[:id]) : current_user
+    authorize @user
+    previously_enabled = @user.active_backup_codes&.any?
```
garyhtou: I don't think this is being used anywhere

```suggestion
```

app/controllers/users_controller.rb
```diff
@@ -189,6 +189,29 @@ def disable_totp
...
+    @user.backup_codes&.active&.map(&:mark_discarded!)
+    @user.backup_codes&.previewed&.map(&:mark_active!)
+    User::BackupCodeMailer.with(user_id: @user.id).new_codes_generated.deliver_now
```
```suggestion
    User::BackupCodeMailer.with(user_id: @user.id).new_codes_activated.deliver_now
```

app/mailers/user/backup_code_mailer.rb
```diff
@@ -0,0 +1,37 @@
...
+    def new_codes_activated
+      mail subject: "You've generated new backup codes for HCB"
+    end
+
+    def code_used
+      mail subject: "You've used a backup code to login to HCB"
+    end
+
+    def backup_codes_disabled
+      mail subject: "HCB backup codes are disabled"
```
garyhtou:
```suggestion
      mail subject: "You've disabled your HCB backup codes"
```

app/controllers/logins_controller.rb
```diff
@@ -207,6 +213,7 @@ def set_available_methods
     @sms_available = @user&.phone_number_verified && !@login.authenticated_with_sms
     @webauthn_available = @user&.webauthn_credentials&.any? && !@login.authenticated_with_webauthn
     @totp_available = @user&.totp.present? && !@login.authenticated_with_totp
+    @backup_code_available = @user&.active_backup_codes&.any? && !@login.authenticated_with_backup_code
```
garyhtou:
```suggestion
    @backup_code_available = @user&.backup_codes_enabled? && !@login.authenticated_with_backup_code
```

app/models/user/backup_code.rb
```diff
@@ -0,0 +1,67 @@
...
+          when 0
+            User::BackupCodeMailer.with(user_id: user.id).no_codes_remaining.deliver_now
+          when 1..3
+            User::BackupCodeMailer.with(user_id: user.id).three_or_fewer_codes_remaining.deliver_now
+          end
+          User::BackupCodeMailer.with(user_id: user.id).code_used.deliver_now
+        end
+      end
+      event :mark_discarded do
+        transitions from: [:active, :previewed], to: :discarded
```
garyhtou: we never transition from previewed to discarded
```suggestion
        transitions from: :active, to: :discarded
```

app/views/user/backup_code_mailer/backup_codes_disabled.html.erb
```diff
@@ -0,0 +1,16 @@
+<p>
+  Hey <%= @user.first_name %>,
+</p>
+
+<p>
+  You've just disabled backup codes for HCB, and your unused backup codes, if any, were invalidated.
```
garyhtou:
```suggestion
  You've just disabled backup codes for HCB. If you had any unused backup codes, they were invalidated.
```

app/views/user/backup_code_mailer/backup_codes_enabled.html.erb
```diff
@@ -0,0 +1,16 @@
+<p>
+  Hey <%= @user.first_name %>,
+</p>
+
+<p>
+  You've just enabled backup codes for HCB, and have 10 codes available to use if you lose access to other means of accessing your account. Please make sure to save the codes in a safe place. HCB staff will never ask for your backup codes.
```
garyhtou:
```suggestion
  You've just enabled backup codes for HCB and have 10 codes available. Please make sure to save the codes in a safe place in case you lose access to your account. HCB staff will never ask for your backup codes.
```

app/views/user/backup_code_mailer/new_codes_activated.html.erb
```diff
@@ -0,0 +1,16 @@
+<p>
+  Hey <%= @user.first_name %>,
+</p>
+
+<p>
+  You've just generated new backup codes for HCB, and have 10 codes available to use in case you're unable to authenticate using other means. Please make sure to save the codes in a safe place. HCB staff will never ask for your backup codes.
```
garyhtou:
```suggestion
  You've just generated backup codes for HCB and have 10 codes available. Please make sure to save the codes in a safe place in case you lose access to your account. HCB staff will never ask for your backup codes.
```

app/views/user/backup_code_mailer/no_codes_remaining.html.erb
```diff
@@ -0,0 +1,16 @@
+<p>
+  Hey <%= @user.first_name %>,
+</p>
+
+<p>
+  You've just used the last backup code remaining on your account. We strongly urge you to regenerate more codes at your earliest convenience in your user settings to ensure you can always access your account.
```
garyhtou:
```suggestion
  You've just used the last backup code remaining on your account. We strongly urge you to regenerate more codes in your <%= link_to "user settings", security_user_url(@user) %> to ensure you can always access your account.
```

app/views/users/edit_security.html.erb
```diff
@@ -92,6 +92,36 @@
       Sign in to HCB using your fingerprint or a hardware security key.
     </div>
   <% end %>
+  <div class="flex items-center justify-between flex-wrap gap-x-5 gap-y-2 mt3 pb2" id="backup-codes">
+    <legend class="flex flex-row justify-start items-center">
+      <span class="h2 bold">Backup codes</span>
+    </legend>
+    <div class="flex flex-row gap-x-2">
+      <% if @user.backup_codes_enabled? %>
+        <%= link_to user_disable_backup_codes_path(@user), class: "btn px2 h4 bg-error", method: :post, data: { turbo: false, confirm: "Are you sure? This will invalidate your currently active codes immediately." } do %>
```
garyhtou:
```suggestion
        <%= link_to user_disable_backup_codes_path(@user), class: "btn px2 h4 bg-error", method: :post, data: { turbo: false, confirm: "Are you sure? This will invalidate your current codes immediately." } do %>
```

app/controllers/users_controller.rb
rluodev:
```suggestion
    authorize @user
```

garyhtou:
When they're previewing back up codes and attempt to close the page, please add one of those "Are you sure you want to leave the page" warnings. I think you can use `onbeforeunload`

app/models/user.rb
```diff
@@ -79,6 +79,8 @@ class User < ApplicationRecord
 
   has_many :logins
   has_many :login_codes
+  has_many :backup_codes, class_name: "User::BackupCode", inverse_of: :user, dependent: :destroy
+  has_many :active_backup_codes, -> { where(aasm_state: :active) }, class_name: "User::BackupCode", inverse_of: :user
```
davidcornu: **suggestion**: Can we use a scope on `User::BackupCode` instead of a separate relation? `aasm` gives us these for free (https://github.com/aasm/aasm?tab=readme-ov-file#automatic-scopes) so we can just do `user.backup_codes.active`.

app/controllers/users_controller.rb
```diff
@@ -189,6 +189,28 @@ def disable_totp
...
+    @user = params[:id] ? User.friendly.find(params[:id]) : current_user
+    authorize @user
+    @previewed_backup_codes = @user.generate_backup_codes!
+  end
+
+  def activate_backup_codes
+    @user = params[:id] ? User.friendly.find(params[:id]) : current_user
+    authorize @user
+    @user.active_backup_codes&.map(&:mark_discarded!)
+    @user.backup_codes&.previewed&.map(&:mark_active!)
```
davidcornu: 1. **concern**: We don't need as many `&` here. These methods return instances of `ActiveRecord::Relation` which always behaves like an array (meaning `.map` is a no-op)

    ```suggestion
        @user.active_backup_codes.map(&:mark_discarded!)
        @user.backup_codes.previewed.map(&:mark_active!)
    ```
2. **concern**: This feels like the kind of logic that
    1. Should be wrapped in a transaction so the entire operation either commits or fails
    2. Does not belong in the controller (maybe `@user.activate_new_backup_codes!` or something to that effect)

app/controllers/users_controller.rb
```diff
@@ -189,6 +189,28 @@ def disable_totp
...
+    User::BackupCodeMailer.with(user_id: @user.id).new_codes_activated.deliver_now
```
davidcornu: **concern**: Are we sure we want `deliver_now` vs. putting in in the queue with `deliver_later`?

app/models/login.rb
```diff
@@ -58,7 +58,7 @@ class Login < ApplicationRecord
     event :mark_complete do
       transitions from: :incomplete, to: :complete do
         guard do
-          authentication_factors_count == (user.use_two_factor_authentication? ? 2 : 1)
+          authentication_factors_count == (user.use_two_factor_authentication? ? 2 : 1) || authenticated_with_backup_code
```
davidcornu: **concern**: My understanding of TOTP backup codes is they act as a replacement for the TOTP factor and not _all_ factors. For example, GitHub requires you to authenticate with your email and password _before_ you get the option to use a recovery code: https://docs.github.com/en/authentication/securing-your-account-with-two-factor-authentication-2fa/recovering-your-account-if-you-lose-your-2fa-credentials#using-a-two-factor-authentication-recovery-code.

If I understand the above logic we're effectively allowing the user to bypass their 2FA setting in the case where they have a recovery code, which I don't think we should be doing.

app/models/user.rb
```diff
@@ -372,6 +374,53 @@ def only_card_grant_user?
...
+        # backup code pepper must be at least 32 bytes
+        backup_codes.create!(code_hash: User::BackupCode.gen_hash(code:, salt:, pepper:), salt: salt)
+      rescue ActiveRecord::RecordInvalid
+        # if the code is already in use, skip it
+        next
+      end
+      codes << code
+    end
+
+    codes
```
davidcornu: 1. **concern**: I find it surprising that this logic lives in the `User` model. Might I suggest we add a class method on `User::BackupCode` that generates the code, saves the record, and returns both the code and the record?
2. **concern**: What's our reasoning here for rolling our own crypto? Can we make use of `has_secure_password` (https://api.rubyonrails.org/classes/ActiveModel/SecurePassword/ClassMethods.html#method-i-has_secure_password) instead?
3. **concern**: This is tricky logic we should definitely have tests for. Happy to help if need be.

app/models/user.rb
```diff
@@ -372,6 +374,53 @@ def only_card_grant_user?
...
+  end
+
+  def redeem_backup_code!(code)
+    found = nil
+    pepper = Credentials.fetch(:BACKUP_CODE_PEPPER)
+    active_backup_codes.each do |backup_code|
+      hash = User::BackupCode.gen_hash(code:, salt: backup_code.salt, pepper:)
+      if ActiveSupport::SecurityUtils.secure_compare(hash, backup_code.code_hash)
+        backup_code.mark_used!
```
davidcornu, line 423: **concern**: We might have a race condition here as we're first reading the active backup codes and later marking one as used (assuming it hasn't already been used in the meantime). 

I'd recommend we re-fetch the backup code, apply pessimistic locking (https://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html), and only then mark it as used.

```suggestion
        ActiveRecord::Base.transaction do
            backup_code = User::BackupCode
              .lock # performs a SELECT ... FOR UPDATE https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE
              .active # makes sure that it hasn't already been used
              .find(backup_code.id) # will raise `ActiveRecord::NotFound` and abort the transaction

            backup_code.mark_used!
            return true
          end
```

app/models/user.rb
```diff
@@ -372,6 +374,53 @@ def only_card_grant_user?
...
+        return true
+      end
+    end
+
+    false
+  end
+
+  def disable_backup_codes!
+    backup_codes.previewed.destroy_all
+    active_backup_codes.map(&:mark_discarded!)
```
davidcornu: **concern**: This should probably be wrapped in a transaction

app/models/user.rb
```diff
@@ -372,6 +374,53 @@ def only_card_grant_user?
...
+    BackupCodeMailer.with(user_id: id).backup_codes_disabled.deliver_now
```
davidcornu, line 436: **concern**: Same question here about whether we should be sending these emails inline vs. queued

app/models/user/backup_code.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: user_backup_codes
+#
+#  id         :bigint           not null, primary key
+#  aasm_state :string
```
davidcornu: **concern**: Should we have a `NOT NULL` with a default here?

app/controllers/users_controller.rb
garyhtou: This is a good topic to raise. Historically, we've found `deliver_now` to be more reliable compared to `deliver_later`. For example, queues may be backed up, or Sidekiq/Redis downtime can cause deliveries to be skipped.

I think we've generally been using `deliver_now` for critical mailers, but I'm happy to explore a more proper solution

app/models/login.rb
garyhtou: Good catch! This reasoning makes sense to me. Any thoughts @rluodev?

rluodev: I think since all our methods of auth are effectively 2FA that the user could lose access to, the backup code makes sense for replacing all the methods

rluodev: Update: discussed w/ Gary in call and we settled on still requiring 2 methods of auth for security.

app/models/user.rb
garyhtou: @davidcornu would you mind setting up a pairing session with @rluodev on these three points?

app/controllers/users_controller.rb
```diff
@@ -189,6 +189,28 @@ def disable_totp
...
+    @user = params[:id] ? User.friendly.find(params[:id]) : current_user
+    authorize @user
+    @previewed_backup_codes = @user.generate_backup_codes!
+  end
+
+  def activate_backup_codes
+    @user = params[:id] ? User.friendly.find(params[:id]) : current_user
+    authorize @user
+    @user.active_backup_codes.map(&:mark_discarded!)
+    @user.backup_codes.previewed.map(&:mark_active!)
```
davidcornu: **concern**: see #2 in https://github.com/hackclub/hcb/pull/10434#discussion_r2169222154 regarding wrapping this in a transaction.

app/models/user.rb
```diff
@@ -79,6 +79,8 @@ class User < ApplicationRecord
 
   has_many :logins
   has_many :login_codes
+  has_many :backup_codes, class_name: "User::BackupCode", inverse_of: :user, dependent: :destroy
+  has_many :active_backup_codes, -> { active }, class_name: "User::BackupCode", inverse_of: :user
```
davidcornu: **concern**: See https://github.com/hackclub/hcb/pull/10434#discussion_r2169215463 - can use use a scope on `User::BackupCode` instead of a separate relation?

app/models/user.rb
```diff
@@ -381,6 +383,56 @@ def only_card_grant_user?
...
+        backup_codes.create!(code: code)
+      rescue ActiveRecord::RecordInvalid
+        # if the code is already in use, skip it
+        next
+      end
+      codes << code
+    end
+
+    codes
+  end
```
davidcornu: **concern**: Should we wrap this in a transaction? Otherwise you could end up doing half the work in the case of a failure.

app/models/user.rb
```diff
@@ -381,6 +383,56 @@ def only_card_grant_user?
...
+        return true
+      end
+      return true
+    end
+
+    false
+  end
+
+  def disable_backup_codes!
+    Base.transaction do
```
davidcornu: **concern**: What is `Base` here? 

```suggestion
    ActiveRecord::Base.transaction do
```

app/models/user/backup_code.rb
```diff
@@ -0,0 +1,64 @@
...
+#  id          :bigint           not null, primary key
+#  aasm_state  :string           default("previewed"), not null
+#  code_digest :text             not null
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#  user_id     :bigint           not null
+#
+# Indexes
+#
+#  index_user_backup_codes_on_code_digest  (code_digest) UNIQUE
```
davidcornu: **concern**: What is this index for? The bcrypt hash is salted so even if the code is identical we'll still end up with different values here.

app/models/user/backup_code.rb
```diff
@@ -0,0 +1,64 @@
...
+  class BackupCode < ApplicationRecord
+    has_paper_trail
+
+    has_secure_password :code
+
+    include AASM
+
+    belongs_to :user
+
+    validates :code_digest, presence: true, uniqueness: true
```
davidcornu: **concern**: (see previous comment)

```suggestion
    validates :code_digest, presence: true
```

app/models/user/backup_code.rb
```diff
@@ -0,0 +1,64 @@
...
+      event :mark_used do
+        transitions from: :active, to: :used
+
+        after do
+          User::BackupCodeMailer.with(user_id: user.id).code_used.deliver_now
+          case user.active_backup_codes.size
+          when 0
+            User::BackupCodeMailer.with(user_id: user.id).no_codes_remaining.deliver_now
+          when 1..3
+            User::BackupCodeMailer.with(user_id: user.id).three_or_fewer_codes_remaining.deliver_now
```
davidcornu: **question**: Do we need separate mailers for this? I'd consider including the number of remaining codes in the "code used" mailer

spec/controllers/logins_controller_spec.rb
davidcornu, line 1: **praise**: Thanks for adding this

app/models/user.rb
```diff
@@ -381,6 +383,56 @@ def only_card_grant_user?
...
+    while codes.size < 10
+      code = SecureRandom.alphanumeric(10)
+      next if codes.include?(code)
+
+      begin
+        backup_codes.create!(code: code)
+      rescue ActiveRecord::RecordInvalid
+        # if the code is already in use, skip it
+        next
+      end
```
davidcornu: **concern**: Given the salt + hash are we likely to ever hit the `rescue` branch?

app/models/user.rb
davidcornu: image

davidcornu:
UI Feedback

Screen | Comment
---- | ----
image | Should backup codes be available as an option regardless of whether the user has configured WebAuthn/TOTP credentials and turn on 2FA?
CleanShot 2025-07-10 at 13 18 10 | I feel like this should read "You last generated a set of 10 backup codes on July 10, 2025, of which 9 are still valid."

app/models/user/backup_code.rb
davidcornu: [Slack discussion](https://hackclub.slack.com/archives/C047Y01MHJQ/p1752166951048399)

app/models/user/backup_code.rb
rluodev: I think we leave it in just to make extra sure that a code can't be the same?

rluodev: Happy to remove it though

spec/controllers/logins_controller_spec.rb
rluodev: Np!

app/models/user/backup_code.rb
rluodev: We do include the number of codes remaining! The 0 codes left email is specifically because we add more urgency in the copy iirc since they used all their codes.

app/models/user.rb
rluodev: Happy to do this, cc @garyhtou since I don't remember the context on why we did this

app/models/user/backup_code.rb
davidcornu: > We do include the number of codes remaining!

I meant in the `code_used` mailer instead of a separate one

Right now we have

Mailer | Sent when | Copy
---- | ---- | ----
`code_used` | Always | "You've just used one of your backup codes to login to HCB..."
`no_codes_remaining` | When the last code is used | "You've just used the last backup code remaining on your account..."
`three_or_fewer_codes_remaining` | When there are between 1 and 3 codes left | "You've just used a backup code to login, and only have N left..."

So as soon as you have fewer than 3 you get `code_used` _and_ `three_or_fewer_codes_remaining`/`no_codes_remaining`.

What I'm suggesting is we combine these three into one and adjust the subject line and copy accordingly. We might also consider redirecting the user to their settings page with a `flash` if they logged in using a backup code and they have none left so they are encouraged to configure new ones immediately.

app/models/user/backup_code.rb
davidcornu: > I think we leave it in just to make extra sure that a code can't be the same?

The only case where that could happen is if we duplicated the value of another row's `code_digest`.  Because this is a `SecureRandom.alphanumeric(10)` combined with a random 16 byte salt, the odds of a collision are very small and IMHO not worth guarding against.

rluodev: yep! i committed your suggestion below and am going to db migrate to reannotate the file

app/models/user.rb
rluodev: If something does go wrong we probably don't want to tank the whole generation process, so it might still be good to have that rescue branch in there?

Not very sure about this tbh

app/models/user.rb
rluodev: sounds good to me

app/models/user.rb
davidcornu: > If something does go wrong we probably don't want to tank the whole generation process

If something does happen it would be highly unexpected and TBH I'd rather fail loudly than try to recover.

rluodev:
> UI Feedback
> 
> Screen	Comment
> image	Should backup codes be available as an option regardless of whether the user has configured WebAuthn/TOTP credentials and turn on 2FA?

I think I discussed with Gary and we thought it should be fine for backup codes to be a replacement authentication method for 1fa situations

> CleanShot 2025-07-10 at 13 18 10	I feel like this should read "You last generated a set of 10 backup codes on July 10, 2025, of which 9 are still valid."

This sounds good to me!

app/models/user/backup_code.rb
garyhtou: I see the benefit of combining them into one email; as well as

>We might also consider redirecting the user to their settings page with a flash if they logged in using a backup code and they have none left so they are encouraged to configure new ones immediately.

app/models/user.rb
rluodev: true, I'll remove the rescue statement

app/models/user/backup_code.rb
rluodev: yep, I merged your suggestion below

app/models/user/backup_code.rb
rluodev: done!

app/models/user.rb
rluodev: I did this, not sure if we should wrap more or less of the code in a transaction if you want to take a look!

app/models/user.rb
rluodev: done!

app/controllers/users_controller.rb
rluodev: done!

app/controllers/users_controller.rb
rluodev: done!

app/controllers/users_controller.rb
rluodev: re: this - David and I discussed during our pair session that it's fine to keep it as deliver_now for now, but he eventually wants to figure out the root cause of why our queues can get so long so we no longer have to do the deliver_now workaround just to avoid a possibly stuck queue.

rluodev:
@davidcornu I think I've addressed all the feedback you left; feel free to do a final review!

app/models/user.rb
```diff
@@ -381,6 +382,61 @@ def only_card_grant_user?
...
+
+      ActiveRecord::Base.transaction do
+        backup_code = User::BackupCode
+                      .lock # performs a SELECT ... FOR UPDATE https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE
+                      .active # makes sure that it hasn't already been used
+                      .find(backup_code.id) # will raise `ActiveRecord::NotFound` and abort the transaction
+        backup_code.mark_used!
+        return true
+      end
+      return true
```
garyhtou: Do we need both of these `return true`?

rluodev:
```suggestion
      end
```

davidcornu (approved):
Looks good to me! Thanks for all the follow-ups ❤️

PR #10435, opened 2025-05-14
Fix event tag creation

polypixeldev:
## Summary of the problem

Event tag creation seemed to be a little broken - when you created one from the admin tab of an event's settings, it would create the tag but wouldn't link it due to the slug that rails sets as `params[:event_id]` being used as a numeric record ID.

## Describe your changes

Removed the redundant hidden field from the new event tag form and changed the controller to interpret the ID as the event's slug.

### DIFF
--- a/app/controllers/event_tags_controller.rb
+++ b/app/controllers/event_tags_controller.rb
@@ -10,8 +10,7 @@ class EventTagsController < ApplicationController
 
     @event_tag.save!
 
-    if params[:event_id]
-      @event = Event.find_by(slug: params[:event_id])
+    if @event.present?
       authorize @event, :toggle_event_tag?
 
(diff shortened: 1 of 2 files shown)


app/controllers/event_tags_controller.rb
```diff
@@ -11,7 +11,7 @@ def create
     @event_tag.save!
 
     if params[:event_id]
-      @event = Event.find(params[:event_id])
+      @event = Event.find_by(slug: params[:event_id])
```
sampoder: Can we use `@event` coming from `SetEvent`?

polypixeldev: good idea, thanks!

sampoder (approved):
LGTM!

PR #10436, opened 2025-05-14
Fix user lock update detection logic

polypixeldev:
## Summary of the problem

`UsersController` was incorrectly detecting if an update had been requested to a user's locked status by checking the presence of the form value, but that value was always present in the form. This prevented other user updates from being made by an admin.

## Describe your changes

It now just checks if the form value for locked is different from the current locked status of the user.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -220,8 +220,8 @@ class UsersController < ApplicationController
         end
       end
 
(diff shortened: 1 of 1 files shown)


sampoder (approved):
LGTM!

sampoder (approved):
LGTM!

PR #10437, opened 2025-05-15
Fix development packages

polypixeldev:
## Summary of the problem

Got an error in development when `convert` wasn't found - `imagemagick` was not included in the Dockerfile for the devcontainer. Also, I noticed `wkhtmltopdf` was included in the packages you need to install if setting up HCB locally, but it seems that binary is provided by the `wkhtmltopdf-binary` in the `Gemfile`.

## Describe your changes

Added `imagemagick` to the Dockerfile and removed the `wkhtmltopdf` package from `development.md`


sampoder (approved):
LGTM

PR #10438, opened 2025-05-15
Link to hcb code in public activity

Luke-Oldenburg:
The card grant purchases were cluttering and I don't think were particularly relevant as they weren't by members of the organization. I think there should be a page somewhere for organizers that shows all transactions that have a subledger id set (ex. show all card grant transactions) to better stay on top of them for fraud.

Screenshot From 2025-05-14 22-46-31

### DIFF
--- a/app/views/public_activity/raw_pending_stripe_transaction/_create.html.erb
+++ b/app/views/public_activity/raw_pending_stripe_transaction/_create.html.erb
@@ -15,10 +15,8 @@
       withdrew <%= render_money hcb_code.stripe_atm_fee ? hcb_code.amount_cents.abs - hcb_code.stripe_atm_fee : hcb_code.amount_cents.abs %> from <i><%= humanized_merchant_name(hcb_code.stripe_merchant) %></i> for <%= link_to hcb_code.event&.name, hcb_code.event %>
     <% end %>
   <% else %>
-    <% if !hcb_code.stripe_card&.card_grant %>
-      <%= render layout: "/public_activity/common", locals: { activity:, current_user:, url: hcb_code_path(hcb_code) } do %>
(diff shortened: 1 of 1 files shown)


sampoder (changes requested):
3 things:

1. I'd recommend using `unless` instead of `if !`
2. This would hide the activity everywhere, including the admin view which is used for auditing / casual auditing, instead we should be editing the `for_event` scope: https://github.com/hackclub/hcb/blob/main/config/initializers/public_activity.rb
3. Lastly, I don't think we should make this change. I appreciate that people are constantly seeing what has happened most recently and if something weird happens, they're more likely to catch it.

+1 on the idea of an audit log for organisers. not sure where that would belong though.

Luke-Oldenburg:
> 3 things:
> 
>     1. I'd recommend using `unless` instead of `if !`
> 
>     2. This would hide the activity everywhere, including the admin view which is used for auditing / casual auditing, instead we should be editing the `for_event` scope: https://github.com/hackclub/hcb/blob/main/config/initializers/public_activity.rb
> 
>     3. Lastly, I don't think we should make this change. I appreciate that people are constantly seeing what has happened most recently and if something weird happens, they're more likely to catch it.
> 
> 
> +1 on the idea of an audit log for organisers. not sure where that would belong though.

Oh yea those are good points. I found them to be kind of useless without a link to the code but then I also removed them because I thought they were unhelpful at the same time. I think they actually would serve a purpose for casual auditing now with the code easier to find. Before you would have to go digging in transfers to find the user's grant card.

I'll make a discussion to track the organizer audit log.

Luke-Oldenburg:
https://github.com/hackclub/hcb/discussions/10443

sampoder (approved):
Works for me

PR #10444, opened 2025-05-16
Create the `supervisor` role

sampoder:


YodaLightsabr:
Are we planning on moving forward with this now that subevents exist

garyhtou:
We are not moving forward with this. However, unrelated, @polypixeldev will be adding an "Owner" role to replace the is_signee boolean.

garyhtou:
Thanks for checking @YodaLightsabr!

PR #10445, opened 2025-05-16
Allow members to request disbursements

YodaLightsabr:
## Summary of the problem
Members can't create disbursements. This PR lets members create disbursements (with manager approval).


app/models/disbursement.rb
```diff
@@ -7,6 +7,7 @@
 #  id                       :bigint           not null, primary key
 #  aasm_state               :string
 #  amount                   :integer
+#  approval_methods         :jsonb
```
sampoder: Imo this should be `approved_by` sorry

app/models/disbursement.rb
```diff
@@ -138,11 +145,14 @@ class Disbursement < ApplicationRecord
...
     event :mark_approved do
-      after do |fulfilled_by|
-        update(fulfilled_by:)
+      after do
         canonical_pending_transactions.update_all(fronted: true)
       end
-      transitions from: [:reviewing, :scheduled], to: :pending
+      transitions from: [:reviewing, :scheduled], to: :pending do
+        guard do
+          can_approve?
```
sampoder: We can probably use the `may_` methods https://github.com/aasm/aasm#guards

app/models/disbursement.rb
YodaLightsabr: Yeah that sounds better

app/models/disbursement.rb
YodaLightsabr: TIL—I'll swap it out

app/policies/event_policy.rb
```diff
@@ -110,11 +110,11 @@ def async_balance?
...
 
   def create_transfer?
-    admin_or_manager? && !record.demo_mode?
+    !record.demo_mode?
   end
 
   def new_transfer?
-    admin_or_manager? && !record.demo_mode?
+    !record.demo_mode?
   end
```
sampoder: I believe these policies are used for all transfers not just disbursements

app/policies/disbursement_policy.rb
```diff
@@ -22,6 +22,10 @@ def can_receive?(role: :manager)
     false
   end
 
+  def manager_approve?
+    OrganizerPosition.role_at_least?(user, record.source_event, :manager)
+  end
+
```
sampoder: Where are we using this method?

YodaLightsabr: Did you mean to resolve this?

YodaLightsabr:
- [x] ~~TODO: Test that there are no changes to card grant issuing (members should still not be able to issue these)~~

app/models/disbursement.rb
```diff
@@ -338,6 +361,11 @@ def fee_waived?
     !should_charge_fee?
   end
 
+  def needs_manager_review?
```
garyhtou: Can this be stored on the disbursement at the time of creation?

Having it stored will give us more flexibility to later customize which transactions require approval

app/views/admin/disbursement_process.html.erb
```diff
@@ -35,6 +35,16 @@
...
+    <% if @disbursement.needs_manager_review? %>
+      <tr>
+        <td style="text-align: right;">Authorized by:</td>
+        <% if @disbursement.authorized_by.present? %>
+          <td><%= user_mention @disbursement.authorized_by %></td>
+        <% else %>
+          <td>Awaiting manager review</td>
+        <% end %>
+      </tr>
+    <% end %>
```
garyhtou: Nice!

app/models/disbursement.rb
garyhtou: In the future, we'll also store _**why**_ a manger review is required:
- **member:** You're a member without spending permissions. You may request transfers, but they require review by a manager.
- **member with insufficient funds:** You're a member with spending controls configured, however, your spending control balance ($100) is insufficient for covering this requested transfer. As a result, it requires manager approval.

config/routes.rb
```diff
@@ -408,6 +408,7 @@
     post "mark_fulfilled"
     post "reject"
     post "cancel"
+    post "manager_approve"
```
garyhtou: Let's keep the naming consistent. Can we rename all "approve"/"approval" for this feature to "authorize"/"authorization"?

db/schema.rb
```diff
@@ -600,8 +600,11 @@
     t.datetime "deposited_at", precision: nil
     t.bigint "destination_subledger_id"
     t.bigint "source_subledger_id"
-    t.date "scheduled_on"
     t.boolean "should_charge_fee", default: false
+    t.date "scheduled_on"
```
garyhtou, line 606: Let's revert the change to `scheduled_on`

app/models/disbursement.rb
```diff
@@ -162,27 +172,40 @@ class Disbursement < ApplicationRecord
...
+
+    if scheduled_on.present?
+      mark_scheduled!
+    else
+      mark_approved!
+    end
+
+  end
+
+  def approve_by_manager(user)
```
garyhtou:
```suggestion
  def authorize_by_manager(user)
```

app/models/disbursement.rb
YodaLightsabr: Got it. Yeah, this will be necessary as we expand the system out further. Would we ever want to skip the manager review step, if they, say, get promoted to a manager while it's awaiting manager authorization?

config/routes.rb
YodaLightsabr: Yep, I've been trying to keep things consistent, I guess I missed this one. Good catch!

Luke-Oldenburg:
Closing as this is outdated. We should standardize a way to request transfers across all transfer types in the new payments system.

PR #10449, opened 2025-05-18
Stop tracking encrypted browser_token in PaperTrail

manuthecoder:
Just some additional cleanup from #10440. Now that browser_token is encrypted with Lockbox, we no longer want to track changes to it in version history.


sampoder (approved):
LGTM! Thank you. If you get a chance, doing a general audit of these would be appreciated.

PR #10455, opened 2025-05-19
Shorten grant return memo

Luke-Oldenburg:
This feels way too long

### DIFF
--- a/app/models/card_grant.rb
+++ b/app/models/card_grant.rb
@@ -199,7 +199,7 @@ class CardGrant < ApplicationRecord
   def cancel!(canceled_by = User.find_by!(email: "bank@hackclub.com"), expired: false)
     raise ArgumentError, "Grant is already #{status}" unless active?
 
-    zero!(custom_memo: "Return of funds from grant to #{user.name}", requested_by: canceled_by) if balance > 0
+    zero!(custom_memo: "Returning #{expired ? "expired" : "canceled"} grant to #{user.name}", requested_by: canceled_by) if balance > 0
 
     update!(status: :canceled) unless expired
     update!(status: :expired) if expired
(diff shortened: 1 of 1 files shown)


sampoder (changes requested):
The extra detail is good context as we don't have another way to track whether it's cancellation or expiration

Luke-Oldenburg:
Could we add a field to the grant of cancelation reason with values of :canceled, :returned, :expired?

Luke-Oldenburg:
Tracking important data like that in the memo feels wrong.

sampoder:
That seems like a good idea to me. Though I would also add that having in the memo gives an organisers immediate context

Luke-Oldenburg:
Could we remove the return of funds part then or rework it somehow? Trying to make it a bit shorter because I had one that was so long by default it got the ellipses truncation when rendering and that seemed excessive to me.

app/models/card_grant.rb
```diff
@@ -199,7 +199,7 @@ def zero!(custom_memo: "Return of funds from grant to #{user.name}", requested_b
   def cancel!(canceled_by = User.find_by!(email: "bank@hackclub.com"), expired: false)
     raise ArgumentError, "Grant is already #{status}" unless active?
 
-    zero!(custom_memo: "Return of funds from #{expired ? "expiration" : "cancellation"} of grant to #{user.name}", requested_by: canceled_by) if balance > 0
+    zero!(custom_memo: "Return of funds from grant to #{user.name}", requested_by: canceled_by) if balance > 0
```
sampoder: It can just be Returning expired grant to X or Returning cancelled grant to Y

Luke-Oldenburg:
> That seems like a good idea to me. Though I would also add that having in the memo gives an organisers immediate context

It already exists as the status enum. I missed that. Might want to add a value to separate returns from cancellations though.

sampoder (approved):
LGTM!

sampoder:
I think what we have in the enum right now is good.

PR #10457, opened 2025-05-19
Remove aliases from job renames

polypixeldev:
## Summary of the problem

Each job renamed in #10177 had an alias added to ensure that jobs already loaded in Sidekiq would not break. They should no longer be needed - **someone with admin on HCB should still check https://hcb.hackclub.com/sidekiq to make sure the old names are no longer there before merging this PR.**

## Describe your changes

Removed the remaining aliases from renamed jobs.


sampoder (approved):
LGTM! Thank you.

PR #10466, opened 2025-05-20
Remove PayPal payout section

Luke-Oldenburg:


YodaLightsabr:
There are currently still users that have their payout information set to PayPal. It might be better to add a notice that PayPal is going away and advise them to switch to another method before we remove the form fields from the page.

Luke-Oldenburg:
The notice was there for a while with the button being disabled and since has been removed so I figured it was time to get rid of the inaccessible form.

PR #10471, opened 2025-05-20
Add job to destroy stale playground events

### LINKED ISSUE #5332: Organization: Delete old playground orgs

polypixeldev:
## Summary of the problem

There's a ton of stale/unused playground events that could be (soft-)deleted!

Supersedes #5995 and closes #5332 

## Describe your changes

Adds a job to destroy these events and related records, to be executed when we switch away from playground events.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -230,7 +230,6 @@ class Event < ApplicationRecord
 
   has_many :organizer_position_invites, dependent: :destroy
   has_many :organizer_positions, dependent: :destroy
-  has_many :all_organizer_positions, -> { with_deleted }, class_name: "OrganizerPosition"
   has_many :organizer_position_contracts, through: :organizer_position_invites, class_name: "OrganizerPosition::Contract"
   has_many :organizer_position_deletion_requests, through: :organizer_positions, dependent: :destroy
   has_many :users, through: :organizer_positions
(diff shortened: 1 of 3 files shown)


app/models/event.rb
```diff
@@ -366,6 +368,14 @@ class Event < ApplicationRecord
 
   after_update :generate_stripe_card_designs, if: -> { attachment_changes["stripe_card_logo"].present? && stripe_card_logo.attached? && !Rails.env.test? }
 
+  # We can't do this through a normal dependent: :destroy since ActiveRecord does not support deleting records through indirect has_many associations
+  # https://github.com/rails/rails/commit/05bcb8cecc8573f28ad080839233b4bb9ace07be
+  after_destroy_commit do
+    all_organizer_positions.each do |position|
```
sampoder: Any reason we couldn't do:
```suggestion
    organizer_positions.with_deleted.each do |position|
```

polypixeldev: that works!

sampoder (approved):
This LGTM! @garyhtou can you take a look?

sampoder (approved):

PR #10472, opened 2025-05-20
Use wkhtmltopdf-binary to ensure consistent version

polypixeldev:
## Summary of the problem

Currently, `wkhtmltopdf` is being installed as an apt package on production but using the `wkhtmltopdf-binary` gem on development. There could be a version mismatch between production and development, which would make a bug related to this package confusing to solve.

## Describe your changes

Removes installing `wkhtmltopdf` on production and always use the `wkhtmltopdf-binary` gem instead.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -171,6 +171,8 @@ gem "query_count"
 gem "rack-mini-profiler", "~> 3.3"
 gem "stackprof" # used by `rack-mini-profiler` to provide flamegraphs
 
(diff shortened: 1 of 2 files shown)


sampoder (approved):
LGTM - will test in production after merging

sampoder:
PDFs appear to be working in production but @polypixeldev I'd appreciate if you could do some extra sanity checking (I just did a quick skim)

polypixeldev:
Thanks, PDFs seem to be working in production for me as well!

PR #10473, opened 2025-05-20
[Card Grants] Let grantees freeze their grant

Luke-Oldenburg:
Only lets them unfreeze if they were the one that froze it.
Screenshot From 2025-05-20 15-25-00

### DIFF
--- a/app/policies/stripe_card_policy.rb
+++ b/app/policies/stripe_card_policy.rb
@@ -10,11 +10,11 @@ class StripeCardPolicy < ApplicationPolicy
   end
 
   def freeze?
-    admin_or_manager? || organizer_and_cardholder? || grantee?
+    admin_or_manager? || member_and_cardholder? || grantee?
   end
 
   def defrost?
-    if grantee? && !organizer? && (record.last_frozen_by != current_user)
+    if grantee? && !member_and_cardholder? && (record.last_frozen_by != current_user)
       return false
     end
 
@@ -22,11 +22,11 @@ class StripeCardPolicy < ApplicationPolicy
   end
 
   def cancel?
-    admin_or_manager? || organizer_and_cardholder?
+    admin_or_manager? || member_and_cardholder?
   end
 
   def activate?
-    (user&.admin? || organizer_and_cardholder?) && !record&.canceled? && !record.event.financially_frozen?
+    (user&.admin? || member_and_cardholder?) && !record&.canceled? && !record.event.financially_frozen?
   end
 
   def show?
@@ -34,11 +34,11 @@ class StripeCardPolicy < ApplicationPolicy
   end
 
   def edit?
-    admin_or_manager? || organizer_and_cardholder?
+    admin_or_manager? || member_and_cardholder?
   end
 
   def update?
-    admin_or_manager? || organizer_and_cardholder?
+    admin_or_manager? || member_and_cardholder?
   end
 
   def transactions?
@@ -55,12 +55,12 @@ class StripeCardPolicy < ApplicationPolicy
 
   private
 
-  def organizer_and_cardholder?
-    organizer? && cardholder?
+  def member_and_cardholder?
+    member? && cardholder?
   end
 
-  def organizer?
-    OrganizerPosition.role_at_least?(user, record&.event, :reader)
+  def member?
+    OrganizerPosition.role_at_least?(user, record&.event, :member)
   end
 
   def cardholder?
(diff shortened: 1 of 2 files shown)


app/policies/stripe_card_policy.rb
```diff
@@ -10,11 +10,11 @@ def shipping?
   end
 
   def freeze?
-    admin_or_manager? || organizer_and_cardholder?
+    admin_or_manager? || organizer_and_cardholder? || (cardholder? && record.card_grant)
```
sampoder: Just to keep these readable can we make `(cardholder? && record.card_grant && !organizer?)` a method called `grantee?`

app/policies/stripe_card_policy.rb
```diff
@@ -10,11 +10,11 @@ def shipping?
...
   end
 
   def defrost?
-    freeze? && !record.event.financially_frozen?
+    !(record.card_grant && record.last_frozen_by != record.user && !organizer?) && freeze? && !record.event.financially_frozen?
```
sampoder:
```suggestion
    if grantee?
    	return false if record.last_frozen_by != current_user
    end
    
    freeze? && !record.event.financially_frozen?
```

app/policies/stripe_card_policy.rb
```diff
@@ -56,7 +60,7 @@ def organizer_and_cardholder?
   end
 
   def organizer?
-    OrganizerPosition.role_at_least?(user, record&.event, :member)
+    OrganizerPosition.role_at_least?(user, record&.event, :reader)
```
sampoder: What's the reasoning for this idea?

Luke-Oldenburg: I believe that line was written before readers existed. I think there's a more proper way to check if someone's an organizer that's used in some of the other policy classes.

sampoder: This would let readers freeze cards though?

Luke-Oldenburg: If they're the cardholder, yea. Id argue this organizer policy is doing what it should be and it's the other that should be changed.

app/policies/stripe_card_policy.rb
```diff
@@ -10,10 +10,14 @@ def shipping?
   end
 
   def freeze?
-    admin_or_manager? || organizer_and_cardholder?
+    admin_or_manager? || organizer_and_cardholder? || grantee?
   end
 
   def defrost?
+    if grantee? && !organizer? && (record.last_frozen_by != current_user)
```
sampoder:
```suggestion
    if grantee? && !(admin_or_manager? || organizer_and_cardholder?) && (record.last_frozen_by != current_user)
```
I think otherwise readers could defrost cards they were the grantee for if an admin froze it. Yes, that's a weird edge case.

sampoder: No - this doesn't fix that bug. I think we should keep organizer at member level (we can rename it to member_and_cardholder?) because readers normally shouldn't be cardholders.

Luke-Oldenburg: Yea that works. Should we account for a niche case though of a reader being demoted? Should it be as if they got removed and cancel their cards?

sampoder: Good question - so many edge cases. Do we currently cancel their cards when they're demoted?

app/policies/stripe_card_policy.rb
sampoder, line 1: thank you!

app/policies/stripe_card_policy.rb
Luke-Oldenburg: We don't. We really should imo.

app/policies/stripe_card_policy.rb
```diff
@@ -10,31 +10,35 @@ def shipping?
   end
 
   def freeze?
-    admin_or_manager? || organizer_and_cardholder?
+    admin_or_manager? || member_and_cardholder? || grantee?
   end
 
   def defrost?
+    if grantee? && !member_and_cardholder? && (record.last_frozen_by != current_user)
```
garyhtou:
```suggestion
    if grantee? && !member? && (record.last_frozen_by != current_user)
```

garyhtou (approved):

app/policies/stripe_card_policy.rb
Luke-Oldenburg: https://github.com/hackclub/hcb/issues/10512

PR #9602, opened 2025-02-19
[Google Workspace] Create the revocation feature

rluodev:
## Summary of the problem

Closes https://github.com/hackclub/hcb/issues/9500

## Describe your changes

Screenshot 2025-05-08 at 5 14 47 PM

Screenshot 2025-05-08 at 5 15 07 PM

Screenshot 2025-05-08 at 5 15 46 PM

Screenshot 2025-05-08 at 5 16 45 PM

PR #10474, opened 2025-05-20
Fix namespacing of new GSuiteJobs

polypixeldev:
## Summary of the problem

Looks like some new `GSuiteJob`s got added with improper namespacing in #9602.

## Describe your changes

Fixed the namespacing for these jobs, added aliases, and updated references.


sampoder (approved):
Thanks for keeping your eye out for these!

PR #10475, opened 2025-05-20
Fallback to hcb_code's event for flipper in transaction partials

polypixeldev:
## Summary of the problem

Small issue where tags won't show up if `@event` isn't defined (although the add tag button already did since the fallback was done correctly there)

## Describe your changes

Added the HCB code event fallback to the flipper check for the tags render


polypixeldev:
Closing as it makes more sense to include in #10380

PR #10476, opened 2025-05-20
Merge receipt bin pair email with upload success email

polypixeldev:
## Summary of the problem

Recently I uploaded a receipt using the receipt bin email on HCB, but was confused when I read the upload success email saying it would be in my receipt bin and it wasn't. Turns out another email has been sent after saying that the receipt had been automatically matched, but it would've been nice for those emails to be merged into one.

## Describe your changes

Merged the receipt bin pair email with the upload success email by making it a synchronous suggestion upload method. The `@reverse` action was also removed as it was already broken and would be difficult to implement if multiple receipts were uploaded in the same email.


sampoder:
> The @reverse action was also removed as it was already broken and would be difficult to implement if multiple receipts were uploaded in the same email.

Unfortunately, this is kind of a key feature that I believe we need to continue supporting. Our AI suggestions aren't perfect and I think it would frustrate users if they have to reupload receipts into their receipt bin.

app/mailboxes/receipt_bin_mailbox.rb
```diff
@@ -25,10 +25,15 @@ def process
 
     return bounce_error if result.empty?
 
+    pairs = result.map do |receipt|
+      receipt.reload.receiptable
+    end.select { |receiptable| receiptable.present? }
```
sampoder: Is the pairing job ran asynchronously?

polypixeldev: No, the pairing job is ran when the receipt is created, since the email receipt bin upload method has been added to `SYNCHRONOUS_SUGGESTION_UPLOAD_METHODS`: https://github.com/hackclub/hcb/blob/9b09c9e60ad4678e50f3a2aa73bdbeee9bb10d2c/app/services/receipt_service/create.rb#L19

polypixeldev:
> > The @reverse action was also removed as it was already broken and would be difficult to implement if multiple receipts were uploaded in the same email.
> 
> Unfortunately, this is kind of a key feature that I believe we need to continue supporting. Our AI suggestions aren't perfect and I think it would frustrate users if they have to reupload receipts into their receipt bin.

I talked with Gary about this and he said that we could remove the `@reverse` action as it already wasn't working and it would take some explaining in the email if you uploaded multiple receipts at once but only wanted to reverse some of them of how to format the command. Another solution to still maintain the reverse pairing functionality would be to include either a link to each pairing in the email that reverses the pairing, or maybe another section in the receipt bin that displays recent pairings and a button to reverse them. What do you think would be best?

polypixeldev:
New solution after talking with Gary is a button on AI-matched receipts to send it back to the receipt bin (which uses the same reversal method that `@reverse` did):
image

sampoder:
I would strongly advocate for the reversal being down via email - this feature is only used by a small subset of users who are trying to avoid the web UI for managing their receipts. I think the link approach seems like the best idea - otherwise it's a lot of clicking through to find that button (though I appreciate that you made it!)

YodaLightsabr:
https://github.com/hackclub/hcb/pull/10476#issuecomment-2904573853 Is this only available on receipts added with a suggested pairing? The actual reversal doesn't seem like it involves AI, so I would advocate for enabling it everywhere and using a color that isn't purple.

garyhtou:
> [#10476 (comment)](https://github.com/hackclub/hcb/pull/10476#issuecomment-2904573853) Is this only available on receipts added with a suggested pairing? The actual reversal doesn't seem like it involves AI, so I would advocate for enabling it everywhere and using a color that isn't purple.

+1 that sounds good. I'd also recommend calling it "Move to receipt bin" to make it clearer that the receipt is being removed from this transaction.

app/controllers/receipts_controller.rb
```diff
@@ -209,6 +209,34 @@ def create
...
+
+    respond_to do |format|
+      format.turbo_stream { render turbo_stream: generate_streams }
+      format.html         {
+        if params[:popover]&.starts_with?("HcbCode:")
+          flash[:popover] = params[:popover].gsub("HcbCode:", "")
+        end
+
+        if success
+          flash[:success] = "Sent receipt to receipt bin"
```
garyhtou: Let's add a link to the receipt bin.

See this for an example: https://github.com/hackclub/hcb/blob/94b7663ecb25c0a5df3933df37f76784b2ff1e42/app/controllers/suggested_pairings_controller.rb#L37-L38

app/controllers/receipts_controller.rb
```diff
@@ -209,6 +209,34 @@ def create
     end
   end
 
+  def reverse
+    authorize @receipt
```
garyhtou, line 213:
```suggestion
    authorize @receipt
    @receiptable = @receipt.receiptable
```

app/mailboxes/receipt_bin_mailbox.rb
garyhtou:
```suggestion
    pairs = result.map { |r| r.reload.receiptable }.select(&:present?)
```

config/routes.rb
```diff
@@ -54,6 +54,10 @@
       post "link"
       get "link_modal"
     end
+
+    member do
+      get "reverse"
```
garyhtou: We shouldn't make updates to the database on a GET request. Let's change this to a POST.
https://softwareengineering.stackexchange.com/questions/188860/why-shouldnt-a-get-request-change-data-on-the-server

An example of why POST is preferred over GET: HCB will pre-fetch links. If someone hovers over a reverse link, but doesn't click it. There's a chance it may actually be reversed.


1. Change this to POST
2. If you're coming from email, on page load, open a modal that says "Are you sure you want to reverse this pairing?" (a Yes/confirmation button will then send that POST request).

app/assets/javascripts/jquery-modal.js
sampoder, line 1: oh Turbo - curious why this change?

app/views/hcb_codes/show.html.erb
```diff
@@ -21,7 +21,7 @@
       ☎️ moshi moshi?
       <%= link_to "Click to get an SMS (to #{current_user.phone_number}) of the notification", send_sms_receipt_hcb_code_path(@hcb_code), method: :post %>
     <% end %>
-    <%= render "receipts/list_v2" %>
+    <%= render "receipts/list_v2", show_bin_button: true %>
```
sampoder: nit: can we call this `show_receipt_bin_button`

app/views/hcb_codes/show.html.erb
```diff
@@ -53,3 +53,13 @@
...
+
+<% if @reverse_receipt_id.present? %>
+  <section id="reverse_receipt_modal" class="modal modal--scroll max-w-md bg-snow" data-behavior="modal" role="dialog" data-modal-auto-open="true" data-modal="reverse_receipt_modal">
+    <%= modal_header("Reverse receipt pairing") %>
+    <p>Are you sure you would like to reverse the receipt pairing for this transaction?</p>
+
+    <%= button_to "Yes", reverse_receipt_path(@reverse_receipt_id, format: :html), class: "left bg-success btn", method: :post %>
+    <a href="#" class="right bg-error btn" rel="modal:close">No</a>
+  </section>
+<% end %>
```
sampoder, line 65: Would recommend using Turbo for this sort of behaviour just for consistency https://arc.net/l/quote/upidddly

app/views/receipts/_receipt.html.erb
```diff
@@ -92,6 +92,17 @@
...
+      <% if defined?(show_bin_button) %>
+        <%= pop_icon_to "payment-docs",
+            reverse_receipt_path(receipt, { popover: local_assigns[:popover], format: (:html unless local_assigns[:turbo_for_deletion]) }.compact),
+            data: {
+              behavior: "modal_ignore",
+              turbo_method: :post,
+            },
+            size: 24,
+            class: "warning tooltipped tooltipped--w",
+            'aria-label': "Move to receipt bin" %>
```
sampoder:
```suggestion
            'aria-label': "Move to Receipt Bin" %>
```

app/mailboxes/receipt_bin_mailbox.rb
```diff
@@ -25,10 +25,13 @@ def process
...
     return bounce_error if result.empty?
 
+    pairs = result.map { |r| r.reload.receiptable }.select(&:present?)
+
     ReceiptBinMailer.with(
       mail: inbound_email,
       reply_to: mail.to.first,
-      receipts_count: result.size
+      receipts_count: result.size,
+      pairs:
```
sampoder: to make this more readable, I would recommend `paired_hcb_codes` / `paired_with` or a more succinct version of that because I was expected these to be `SuggestedPairing`s

app/views/receipt_bin_mailer/bounce_success.html.erb
```diff
@@ -1,7 +1,20 @@
 <p>
-We got your <%= "receipt".pluralize(@receipts_count) %>, thanks! You can find <%= @receipts_count == 1 ? "it" : "them" %> in your <%= link_to "Receipt Bin", my_inbox_url %>.
+We got your <%= "receipt".pluralize(@receipts_count) %>, thanks!
+<% unless @receipts_count == @pairs.size %>
+You can find <%= @receipts_count == 1 ? "it" : "them" %> in your <%= link_to("Receipt Bin", my_inbox_url) %>.
+<% end %>
 </p>
 
+<% if @pairs.size > 0 %>
```
sampoder:
```suggestion
<% if @pairs.any? %>
```

app/assets/javascripts/jquery-modal.js
polypixeldev: I added this because the code in ui.js to automatically open the confirmation modal was running before Turbo had loaded

app/views/hcb_codes/show.html.erb
polypixeldev: I'm not sure if `data-turbo-confirm` is applicable here - this modal is only opened when you click on the link to reverse from an email, since there needs to be a confirmation modal that actually sends the POST request to reverse the receipt pairing. But from what I understand, `data-turbo-confirm` only opens an additional dialog once the user clicks a link or submits a form with that attribute.

app/assets/javascripts/jquery-modal.js
```diff
@@ -153,7 +153,9 @@
     },
 
     show: function () {
-      Turbo.cache.clear()
+      if (typeof Turbo != "undefined") {
+        Turbo.cache.clear()
+      }
```
sampoder, line 158: This feels odd - any idea what was causing this?

app/services/receipt_service/suggest.rb
```diff
@@ -35,13 +35,14 @@ def run!(include_details: false)
 
           if @receipt.email_receipt_bin? && pair = @receipt.suggested_pairings
                                                            .unreviewed
-                                                           .where("distance <= ?", 20)
+                                                           .where("distance <= ?", 200)
```
sampoder: hmm - any reason behind this? (cc: @YodaLightsabr)

sampoder (approved):
thank you! sorry for all the delay with reviewing this just a couple of changes we might want to not make but then this seems good to go

app/services/receipt_service/suggest.rb
polypixeldev: Oops, this is just a debugging change - I had changed the distance threshold to 200 so that I could get it to match my receipts while testing, will remove

app/assets/javascripts/jquery-modal.js
polypixeldev: This seemed to be an issue with just Turbo taking a second to load, which became an issue here since jquery-modal tries to open the confirm reverse modal right when the page loads.

sampoder (approved):
thank you!

sampoder:
Could you spend some time confirming this works as expected when deployed?

PR #10477, opened 2025-05-20
[DX] Add more linking to code in dev-docs

Luke-Oldenburg:
Goal is to make it easier for somebody looking at the docs to find the file they're looking for especially if they are new to Ruby.


dev-docs/pull_requests.md
```diff
@@ -15,4 +15,4 @@
 - [ ] Tested by reviewer before merging <!-- leave this unchecked until right before merging -->
 - [ ] After merging: Close related issues
 - [ ] After merging: Notify the requestor of the issue. Post in `#hcb-team` if the feature impacts HCB's operations team
```
Luke-Oldenburg: I assume it's probably best not to link to a private channel to avoid confusion. Should this line even be here at all?

dev-docs/guides/transaction_imports.md
```diff
@@ -34,19 +34,19 @@ Each of these book transfers has a memo in this format: `HCB DISBURSE [ID]`.
 
 Based on the ID, `EventMappingEngine::Map::Disbursements` will map the transactions to their associated event and subledger. The logic for determining whether it should be mapped to the receiving event or the source event is based on `amount_cents` and is located in `EventMappingEngine::GuessEventId::Disbursement`. If it’s a positive transaction, we map it to the receiving event. If it’s a negative transaction, we map it to the source event.
```
Luke-Oldenburg, line 39: I don't think these files still exist.

sampoder: Yes - this code was rewritten.

Luke-Oldenburg: Should I just rip this out or could you rewrite this paragraph please? I'm unfamiliar with how it works currently.

dev-docs/pull_requests.md
```diff
@@ -14,5 +14,5 @@
 - [ ] Tested by submitter before requesting review _(Does the test plan pass in development or staging?)_
 - [ ] Tested by reviewer before merging <!-- leave this unchecked until right before merging -->
 - [ ] After merging: Close related issues
-- [ ] After merging: Notify the requestor of the issue. Post in `#hcb-team` if the feature impacts HCB's operations team
```
sampoder, line 17: Please keep this - the person merging the PR should do this

dev-docs/guides/transaction_imports.md
sampoder: We should rewrite it at some point - we now use HCB short code mapping for this

dev-docs/pull_requests.md
Luke-Oldenburg: Ah good point I didn't think of that.

YodaLightsabr (approved):

PR #10481, opened 2025-05-20
Fix memo underline after stream

polypixeldev:
## Summary of the problem

Underlines on linked memos were disappearing after a popover was opened. The `memo_stream` partial was rendering a turbo stream that replaced the memo with a memo without the underline.

## Describe your changes

Moved the underline style to the parent of the span that is replaced by `memo_stream`.

### DIFF
--- a/app/views/hcb_codes/_memo.html.erb
+++ b/app/views/hcb_codes/_memo.html.erb
@@ -37,7 +37,7 @@
               data-navigation-location-param="<%= edit_hcb_code_path(hcb_code, inline: true, prepended_to_memo:, location:, ledger_instance:) %>"
               data-navigation-frame-param="<%= "#{location}:#{ActionView::RecordIdentifier.dom_id(hcb_code)}:memo" %>">
             <span
-              class="mr2 tooltipped tooltipped--s memo-underline"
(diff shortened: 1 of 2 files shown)


app/views/hcb_codes/_memo.html.erb
```diff
@@ -37,11 +37,11 @@
               data-navigation-location-param="<%= edit_hcb_code_path(hcb_code, inline: true, prepended_to_memo:, location:, ledger_instance:) %>"
               data-navigation-frame-param="<%= "#{location}:#{ActionView::RecordIdentifier.dom_id(hcb_code)}:memo" %>">
             <span
-              class="mr2 tooltipped tooltipped--s"
+              class="mr2 tooltipped tooltipped--s memo-underline"
               style="color: inherit; text-decoration: none;"
               aria-label="Shift+click to rename this transaction">
               <%= prepended_to_memo %>
-              <span data-memo-for="<%= dom_id(hcb_code) %>" class="<%= "hover:underline" if ledger_instance %>">
```
sampoder, line 44: What's the reason for dropping the `if ledger_instance` check?

polypixeldev: Ah, I forgot to add it back on the parent div - added it back now

sampoder (approved):
LGTM!

PR #10485, opened 2025-05-21
Cut canceled cards in half!

Luke-Oldenburg:
Porting this from emburse cards to stripe ones too!
Screenshot From 2025-05-20 22-31-02


sampoder (dismissed):
I think we want DRY this up a little, it feels like a lot of code is being repeated here

app/views/stripe_cards/_stripe_card.html.erb
```diff
@@ -1,53 +1,16 @@
...
+  <% if stripe_card.canceled? %>
+    <div class="canceled-card-wrapper">
+      <%= render "stripe_cards/card", event: @event, stripe_card:, show_card_details: @show_card_details, side: "left" %>
+      <%= render "stripe_cards/card", event: @event, stripe_card:, show_card_details: @show_card_details, side: "right" %>
+    </div>
+
+  <% else %>
+    <%= render "stripe_cards/card", event: @event, stripe_card:, show_card_details: @show_card_details %>
+  <% end %>
+
```
sampoder, line 13: @YodaLightsabr what are your thoughts on how partials are being used here? I'm worried we're making this code more complicated than it needs to be. Maybe [`content_for` and `yield`](https://stackoverflow.com/questions/13150983/rails-what-is-the-difference-between-content-for-and-yield) would help or if there was an alternative way of doing this without duplicating the HTML?

YodaLightsabr: Hmm. I'm also worried that this isn't accessible. I haven't tried it locally but I'm worried that each card is two separate focus items and that screen readers will repeat cards if they're cancelled.

Luke-Oldenburg: @YodaLightsabr they are. Is that an issue?

sampoder: Yes - screen readers should ignore the second element https://stackoverflow.com/questions/672156/is-there-a-way-to-write-content-that-screen-readers-will-ignore

app/views/stripe_cards/_card.html.erb
```diff
@@ -0,0 +1,53 @@
+<%# locals: (event:, stripe_card:, show_card_details:, side: "left") %>
+
+<div
+  class="stripe-card
+        <%= stripe_card.status_text.downcase %>
+        <%= stripe_card.virtual? ? "virtual" : "" %>
```
davidcornu: **suggestion** (minor): Using the same pattern as the classes after might make this more readable

```suggestion
        <%= "virtual" if stripe_card.virtual? %>
```

app/views/stripe_cards/_card.html.erb
```diff
@@ -0,0 +1,53 @@
...
+        <%= "platinum" if stripe_card.platinum? %>
+        <%= "stripe-card--white" if stripe_card.personalization_design&.white? %>
+        <%= "stripe-card--black" if stripe_card.personalization_design&.black? %>
```
davidcornu: **suggestion**: We could make use of the `PersonalizationDesign#color` [method](https://github.com/hackclub/hcb/blob/ee5fc8af72dc219f1a7c64b77042f354e1ef3398/app/models/stripe_card/personalization_design.rb#L75-L77) to remove some duplication here

```suggestion
        <%= "stripe-card--#{stripe_card.personalization_design.color}" if stripe_card.personalization_design %>
```

app/views/stripe_cards/_card.html.erb
```diff
@@ -0,0 +1,53 @@
...
+        <%= "deactivated" if stripe_card.canceled? %>
+        <%= "canceled-left" if side == "left" && stripe_card.canceled? %>
+        <%= "canceled-right" if side == "right" && stripe_card.canceled? %>
```
davidcornu: **suggestion**: Another opportunity to reduce some duplication

```suggestion
        <%= "canceled-#{side}" if stripe_card.canceled? %>
```

Another option might be to 
1. add a more generic `stripe-card--#{side}` class 
2. add a `canceled` class when the card is cancelled
3. use the combination of (1) and (2) for styling

seen as the side isn't really specific to being cancelled.

app/views/stripe_cards/_card.html.erb
```diff
@@ -0,0 +1,53 @@
...
+        mt1
+  "
+  aria-hidden="<%= "true" if side == "right" && stripe_card.canceled? %>"
```
davidcornu: **question**: How does `aria-hidden=""` behave? Is that treated as `false`?

app/views/stripe_cards/_card.html.erb
```diff
@@ -0,0 +1,53 @@
...
+    <span class="bold caps h6 absolute flex items-center" style="opacity: 50%; right: 10px; top: 10px;">
+      <%= inline_icon "bank-account" %>
+      Platinum
+    </span>
+  <% end %>
+  <% if stripe_card.personalization_design.present? %>
+    <div class="stripe-card__personalization absolute" style="right: 16px; top: 16px;">
+      <div class="stripe-card__personalization_image_wrapper">
+        <div class="stripe-card__personalization_inner_image_wrapper">
+          <%= image_tag Rails.application.routes.url_helpers.rails_blob_url(stripe_card.personalization_design.logo), class: "stripe-card__personalization-image" %>
```
davidcornu: **question**: Do we need to fully quality the url helper?

```suggestion
          <%= image_tag rails_blob_url(stripe_card.personalization_design.logo), class: "stripe-card__personalization-image" %>
```

app/views/stripe_cards/_card.html.erb
```diff
@@ -0,0 +1,53 @@
...
+        </div>
+      </div>
+    </div>
+  <% end %>
+  <p class="stripe-card__number fs-mask">
+    <% if stripe_card.user != current_user %>
+      <%= stripe_card.hidden_card_number_with_last_four %>
+    <% elsif local_assigns[:show_card_number] %>
```
davidcornu: **question**: Where does this come from? Should there be a definition for it in the strict locals comment?

app/views/stripe_cards/_card.html.erb
```diff
@@ -0,0 +1,53 @@
...
+      <%= stripe_card.formatted_card_number %>
+    <% elsif @show_card_details %>
```
davidcornu: **concern**: We also seem to have a `show_card_details` local. Which one is correct?

app/views/stripe_cards/_card.html.erb
```diff
@@ -0,0 +1,53 @@
...
+        <%= stripe_card.formatted_card_number %>
+    <% else %>
+      <%= stripe_card.hidden_card_number_with_last_four %>
+    <% end %>
+  </p>
+  <p class="stripe-card__name flex justify-between items-center">
+    <span style="overflow: hidden; text-overflow: ellipsis; white-space: nowrap;">
+      <%= defined?(@event) ? stripe_card.user.name : stripe_card.event.name %>
```
davidcornu: **question**: Where does `@event` come from?

Luke-Oldenburg: Some magic in either the users or events controller. I believe this defined check is used for card grant rendering.

Luke-Oldenburg: Using instance variables in partials is bad but this isn't a new usage so it shouldn't break anything.

davidcornu: We also have a required `event` local though so it should always be there. Can we use that instead?

app/views/stripe_cards/_card.html.erb
Luke-Oldenburg: Good point I'll make sure it returns false.

app/views/stripe_cards/_card.html.erb
Luke-Oldenburg: These classes are specific to it being canceled though so I think it's fine. I went ahead and cleaned up the code a bit more.

app/views/stripe_cards/_card.html.erb
```diff
@@ -0,0 +1,53 @@
...
+    </div>
+  <% end %>
+  <p class="stripe-card__number fs-mask">
+    <% if stripe_card.user != current_user %>
+      <% if OrganizerPosition.role_at_least?(current_user, stripe_card.event, :reader) %>
+        <%= stripe_card.hidden_card_number_with_last_four %>
+      <% else %>
+        <%= stripe_card.hidden_card_number %>
+      <% end %>
+    <% elsif show_card_details || @show_card_details %>
```
davidcornu: **suggestion**: Given that this is a new partial with strict locals, we shouldn't need to rely on this instance variable


```suggestion
    <% elsif show_card_details %>
```

davidcornu (approved):

app/views/stripe_cards/_card.html.erb
Luke-Oldenburg: Should we drop it from the controller then too?

davidcornu: Aren't we using that to set it in the `render(partial:...` call?

Luke-Oldenburg: Ah good point.

davidcornu (approved):

davidcornu:
^ well that was overly enthusiastic of me 😅. forgot I'd already approved.

I'm logging off for the day so I can ship this in the morning (feel free to do so yourself before as well)

PR #10492, opened 2025-05-22
Fix external receipt uploading

polypixeldev:
## Summary of the problem

Receipt uploading on the attach receipt link by someone who isn't signed in as authorized to upload a receipt wasn't working.

## Describe your changes

Passed the secret from the upload form to the receipt controller.


sampoder (approved):
Wow - good catch here - thank you!

PR #10493, opened 2025-05-22
Add low spending balance warning email

polypixeldev:
## Summary of the problem

We need to make sure members with spending limits have their limits in mind, and spare them the frustration of having their card decline when they've run out of balance!

Closes #8397

## Describe your changes

Adds a warning email in a new `ControlsMailer` that is sent when a user's spending balance gets low (as described in the issue, except that the $25 threshold is only taken into account when crossed for the first time) after a card authorization or when one of their transactions settles at a different amount.

image


app/mailers/organizer_position/spending/controls_mailer.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class OrganizerPosition
+  module Spending
+    class ControlsMailer < ApplicationMailer
+      def warning
+        @control = params[:control]
+        @controls_path = Rails.application.routes.url_helpers.event_organizer_position_spending_controls_path(event_id: @control.organizer_position.event, organizer_position_id: @control.organizer_position)
```
sampoder: I recommend not storing paths in instance variables.

app/services/canonical_pending_transaction_service/settle.rb
```diff
@@ -25,6 +25,11 @@ def run!
           canonical_pending_transaction_id: @canonical_pending_transaction.id,
           canonical_transaction_id: @canonical_transaction.id,
         ).notify_settled.deliver_later
+
+        spending_control = @canonical_transaction.stripe_card.active_spending_control
+        if spending_control.present?
+          SpendingControlService.check_low_balance(spending_control, -@canonical_pending_transaction.amount_cents + @canonical_transaction.amount_cents)
+        end
```
sampoder: I would recommend making this as an `after_create_commit` on CPT as I think this service is ran after a pending transaction settles (so emails will be delayed by a couple of days)

app/services/spending_control_service.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module SpendingControlService
+  def self.check_low_balance(spending_control, amount_difference)
+    historic_max_balance_cents = spending_control.balance_cents - amount_difference
```
sampoder: Would recommend naming this variable `previous_balance`

app/services/spending_control_service.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module SpendingControlService
+  def self.check_low_balance(spending_control, amount_difference)
```
sampoder: I'd also recommend passing in a HcbCode and using it's amount instead of `amount_difference`

app/services/stripe_authorization_service/create_from_webhook.rb
```diff
@@ -43,6 +43,11 @@ def run
           if cpt.local_hcb_code&.stripe_cash_withdrawal?
             AdminMailer.with(hcb_code: cpt.local_hcb_code).cash_withdrawal_notification.deliver_later
           end
+
+          spending_control = cpt.stripe_card.active_spending_control
+          if spending_control.present?
+            SpendingControlService.check_low_balance(spending_control, cpt.amount_cents)
+          end
```
sampoder: Ah - I see this now. Good idea also putting it in the settle CT service.

app/services/canonical_pending_transaction_service/settle.rb
sampoder: I now see how this works - LGTM. I'd recommend passing in a HcbCode and working out the balance off of that.

app/mailers/organizer_position/spending/controls_mailer.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class OrganizerPosition
+  module Spending
+    class ControlsMailer < ApplicationMailer
+      def warning
```
garyhtou: rename to low balance warning

sampoder (approved):
code lgtm! will iterate in production.

PR #10494, opened 2025-05-22
Fix error in control policy

polypixeldev:
## Summary of the problem

There is currently an error raised when a user not in an event tries to access a spending control page, when it should just display the normal "You are not authorized to perform this action." toast.

## Describe your changes

Added a `&` to the policy method that finds the organizer position in case it does not exist.


sampoder (approved):
LGTM!

PR #10495, opened 2025-05-22
[Card Grants] Hide actions for members

### LINKED ISSUE #10469: [Popovers] Actions on card grants are only visible on invites
### Describe your goals

View a card grant in a popover.

### Describe the bug

I could only view the organizer actions on a pending invite for a card grant. After the invite was accepted, these actions were only visible on the full page and not in the popover.
https://hackclub.slack.com/archives/CN523HLKW/p1747523437660359

Luke-Oldenburg:
## Summary of the problem

Members don't have permissions to use the actions so they should see the regular view and not the organizer view.

manager/admin: 
image

member:
image

reader: 
image

## Describe your changes

Use card grant policy.

### DIFF
--- a/app/policies/card_grant_policy.rb
+++ b/app/policies/card_grant_policy.rb
@@ -31,7 +31,7 @@ class CardGrantPolicy < ApplicationPolicy
 
   def edit?
     admin_or_manager? && record.active?
-  end  
+  end
 
   def toggle_one_time_use?
     admin_or_manager? && record.active?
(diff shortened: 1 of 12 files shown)


sampoder:
What does a member see after this update if they aren't the grant holder?

Luke-Oldenburg:
Good catch! The page was empty.

garyhtou (commented):
1. Move all action buttons into `_action` partial
2. Add a if/else for return vs cancel
3. Add policy checks for all buttons inside `_action` partial
4. Update `app/views/card_grants/show.html.erb:50-53` to render `_action` partial

garyhtou (approved):
really nice work!

PR #10497, opened 2025-05-22
Add back IncreaseCheckJob alias

polypixeldev:
## Summary of the problem

This job still needs the alias since it gets scheduled a long time in advance.

## Describe your changes

Adds back the alias

Need to re-run the failed job in Sidekiq when merged


garyhtou (approved):
ty

PR #10499, opened 2025-05-23
Fix card grant expiration warnings

skyfallwastaken:
## Summary of the problem
The code for card grant expiration emails was already there, but they didn't work because the job was checking if `expires_at` was exactly 3 days/1 month before now.

## Describe your changes
Checks if it expires between now and the notification window (1 month/3 days/24 hours). If it does, sends out the expiration warning email and sets the `last_expiry_notification_sent_at` field so that emails aren't re-sent every day for the 1 month/3 days warnings.

### DIFF
--- a/app/jobs/card_grant/expiration_job.rb
+++ b/app/jobs/card_grant/expiration_job.rb
@@ -25,5 +25,7 @@ class CardGrant
         end
       end
     end
+
   end
+
 end
--- a/app/jobs/card_grant/expiration_job.rb
+++ b/app/jobs/card_grant/expiration_job.rb
@@ -3,22 +3,26 @@
 class CardGrant
   class ExpirationJob < ApplicationJob
     queue_as :low
+
+    NOTIFICATION_WINDOWS = {
+      "24 hours" => 24.hours,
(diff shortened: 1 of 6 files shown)


sampoder:
> The code for card grant expiration emails was already there, but they didn't work because the job was checking if expires_at was exactly 3 days/1 month before now.

This was intentional. If we check that it's exactly 3 days ago we can avoid an extra database column.

skyfallwastaken:
> This was intentional. If we check that it's exactly 3 days ago we can avoid an extra database column. 

Oh ic! Makes sense.

From my understanding though, `expires_on` isn’t taking account of the fact that the _time_ portion of Time.now isn’t the same as the expires_at, so the actual fix is probably to just fix that (but do correct me if I’m wrong! I’m still learning Rails)

sampoder:
> From my understanding though, expires_on isn’t taking account of the fact that the time portion of Time.now isn’t the same as the expires_at, so the actual fix is probably to just fix that (but do correct me if I’m wrong! I’m still learning Rails) 

That sounds right!

PR #10502, opened 2025-05-23
[Receipt Bin] Improve UI and be more strict with suggestions

YodaLightsabr:
## Summary of the problem
Receipt Bin frequently suggests receipts that are missing important information or don't match the transaction.

If HCB suggests a receipt, the user should feel confident that the suggestion is accurate.

## Describe your changes
This PR makes updates to the pairing algorithm by automatically disqualifying suggestions that are missing key pieces of information like date, amount, or card last 4. It provides more wiggle room on date matching (because merchants like Amazon sometimes charge the card at a later date) and requires the amount to be an exact match.


app/views/receipts/_receipt.html.erb
```diff
@@ -2,66 +2,60 @@
...
+        </span>
+      </span>
+    <% end %>
+  </div>
 
   <% if defined?(selectable) %>
     <span>
       <div class="receipt__icon"></div>
       <span style="display: inline-grid; border-radius: 0.5rem;">
+        <div class="receipt__icon--check">✓</div>
```
sampoder, line 26: Thank you!

app/services/receipt_service/suggest.rb
```diff
@@ -50,70 +50,34 @@ def run!(include_details: false)
...
+                      @extracted.extracted_total_amount_cents.nil? ||
+                      @extracted.extracted_date.nil? ||
+                      @extracted.extracted_card_last4.nil?
+
+      distance = 0
 
-      features = distances.values.reject { |data| data[:value].nil? }
+      unless (@extracted.extracted_total_amount_cents.abs - hcb_code.amount_cents.abs).abs == 0
+        distance += 200 # Automatically disqualify suggestion if not matching
+      end
```
sampoder, line 123: Should we provide 10% leeway? Feels a bit strict when totals on receipts aren't always the exact same especially because ChatGPT can be a bit off

app/services/receipt_service/suggest.rb
```diff
@@ -50,70 +50,34 @@ def run!(include_details: false)
...
+
+      unless @extracted.extracted_card_last4 == hcb_code.card&.last4
+        distance += 200 # Automatically disqualify suggestion if not matching
+      end
```
sampoder, line 127: What if `extracted_card_last4` is nil? Similar comments for below

app/services/receipt_service/suggest.rb
YodaLightsabr: We can play around with it, but I've found that the amounts ChatGPT pulls are usually either perfectly accurate or off by a lot. Even if small errors were more common, I'd rather have a 10% false negative rate than a 10% false positive rate. I'm open to other ideas though.

app/services/receipt_service/suggest.rb
YodaLightsabr: Line 54 should handle that case

app/services/receipt_service/suggest.rb
sampoder: Oh no - just like the number on the receipt doesn't match. For example Ubers or restaurants with tips later on or an unlisted CC charge (yes - I know you should upload a better receipt because the full one would have it but I don't think we should require perfect matches). But we can also put this in production and experiment, just my gut feels like there should be room for variability.

YodaLightsabr: That's true, but then it makes Receipt Bin less effective when you make multiple purchases from the same merchant on the same date. If the extra tip data just isn't in the original receipt, I don't think we should be trying to factor it in. Is there a way you could envision building this out that doesn't result in the bad suggestions we're currently getting?

sampoder: Talked it through with Ian in DMs 
Screenshot 2025-05-28 at 8 36 47 PM

app/services/receipt_service/suggest.rb
sampoder: I don't think that we should disqualify receipts that don't have a last four

YodaLightsabr: Are you suggesting that we should instead only disqualify when the data exists but differs from the transaction?

sampoder: Yes

PR #10503, opened 2025-05-24
add ephemeralKeyId to the ephemeral key cards v4 route

Mohamad-Mortada:
According to [this](https://docs.stripe.com/issuing/cards/digital-wallets?platform=react-native&lang=curl#push-provisioning), for HCB Mobile card wallet provisioning we need the ephemeral key id too as we have to pass in this entire object to the component, we have the secret and the stripe id but not the ephemeral id:

```
{
  "id": "ephkey_1G4V6eEEs6YsaMZ2P1diLWdj",
  "object": "ephemeral_key",
  "associated_objects": [
    {
      "id": "ic_1GWQp6EESaYspYZ9uSEZOcq9",
      "type": "issuing.card"
    }
  ],
  "created": 1586556828,
  "expires": 1586560428,
  "livemode": false,
  "secret": "ek_test_YWNjdF8xRmdlTjZFRHelWWxwWVo5LEtLWFk0amJ2N0JOa0htU1JzEZkd2RpYkpJdnM_00z2ftxCGG"
}
```

### DIFF
--- a/app/controllers/api/v4/stripe_cards_controller.rb
+++ b/app/controllers/api/v4/stripe_cards_controller.rb
@@ -129,7 +129,7 @@ module Api
 
         ahoy.track "Card details shown", stripe_card_id: @stripe_card.id, user_id: current_user.id, oauth_token_id: current_token.id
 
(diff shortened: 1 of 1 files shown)


sampoder (approved):
sg to me!

PR #10513, opened 2025-05-27
Stop members from unfreezing their card if it was frozen by a manager

Luke-Oldenburg:

### DIFF
--- a/app/policies/stripe_card_policy.rb
+++ b/app/policies/stripe_card_policy.rb
@@ -14,7 +14,7 @@ class StripeCardPolicy < ApplicationPolicy
   end
 
   def defrost?
-    if grantee? && !member? && (record.last_frozen_by != current_user)
+    if ((grantee? && !member?) || member_and_cardholder?) && (record.last_frozen_by != current_user)
       return false
     end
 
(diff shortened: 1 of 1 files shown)


app/policies/stripe_card_policy.rb
```diff
@@ -14,7 +14,7 @@ def freeze?
   end
 
   def defrost?
-    if grantee? && !member? && (record.last_frozen_by != user)
+    if ((grantee? && !admin_or_manager?) || member_and_cardholder?) && (record.last_frozen_by != user)
       return false
     end
```
sampoder:
```suggestion
    unless record.last_frozen_by == user || admin_or_manager?
        return false
    end
```
Thoughts on this code? I'm not sure if I'm missing an edge case but this feels a lot more readable.

sampoder (approved):

sampoder (approved):

app/policies/stripe_card_policy.rb
Luke-Oldenburg: ik this got merged already but it lgtm. was busy for a few days.

leowilkin:
@Luke-Oldenburg if i'm understanding this PR correctly, this allows members to unfreeze their own cards that managers froze? isn't this intentional behaviour?

Luke-Oldenburg:
@leowilkin it does the opposite of that.

leowilkin:
aha! good stuff

PR #10515, opened 2025-05-28
[StripeCardPolicy] Fix error while defrosting

garyhtou:
## Summary of the problem

image

fyi @Luke-Oldenburg


Luke-Oldenburg:
Oops😬

PR #10516, opened 2025-05-28
Add admin tool with user last active

### LINKED ISSUE #10470: [Admin] Add a thing to see who's been active, not just the number
### Describe the problem

ok - here's the pitch. on the team page as an admin, I can see how many people have been active in the last 30 days, but I can't tell who tf they are - esp when like 50% have been active, but the other half havent.

can you add something to the team page that you can toggle (?) to see who's been active, like a green pulse or something...

maybe when you click on the thing that says how many active teens there have been?

Image

_No response_

Luke-Oldenburg:
closes #10470 
Screenshot From 2025-05-28 09-41-02
Screenshot From 2025-05-28 09-40-54

### DIFF
--- a/app/views/organizer_positions/_organizer_position.html.erb
+++ b/app/views/organizer_positions/_organizer_position.html.erb
@@ -70,7 +70,12 @@
         <% end %>
         <% admin_tool("badge") do %>
           <div class="text-muted ml0">
-            last active: <%= format_date "last active: #{ organizer_position.user.last_seen_at}" %>
+            Last active:
+            <% if organizer_position.user.last_seen_at.present? %>
+              <%= format_date(organizer_position.user.last_seen_at) %>
+            <% else %>
+              unknown
+            <% end %>
           </div>
         <% end %>
       </div>
--- a/app/views/organizer_positions/_organizer_position.html.erb
+++ b/app/views/organizer_positions/_organizer_position.html.erb
@@ -90,6 +90,10 @@
       <% end %>
     </div>
   <% end %>
+
+  <%# This empty div pushes the content below to the bottom of the card. %>
+  <div class="mt-auto"></div>
+
   <% if auditor_signed_in? %>
     <div class="warning m0">
       Last active:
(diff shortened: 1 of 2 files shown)


garyhtou (commented):
This is a nice addition! I wonder if we could restructure this area— it's getting a bit cluttered for admins.

app/views/organizer_positions/_organizer_position.html.erb
```diff
@@ -68,6 +68,11 @@
                           disabled: !policy(organizer_position).can_request_removal? %>
           </div>
         <% end %>
+        <% admin_tool("badge") do %>
+          <div class="text-muted ml0">
+            last active: <%= format_date "last active: #{ organizer_position.user.last_seen_at}" %>
+          </div>
+        <% end %>
```
garyhtou: I believe `format_date` expects a date and not a string. And we should  probably handle the case where `last_seen_at` is null— which is possible for really old users who haven't signed in for a while.

```suggestion
        <% admin_tool("badge") do %>
          <div class="text-muted ml0">
            Last active:
            <% if organizer_position.user.last_seen_at.present? %>
              <%= format_date(organizer_position.user.last_seen_at) %>
            <% else %>
              unknown
            <% end %>
          </div>
        <% end %>
```

sampoder:
+1 on design. We could add a row below phone + email.

YodaLightsabr:
Another option worth considering is showing the user last active date in the tooltip when you hover over a user mention. It would make this information available in more places and not clutter up the existing team card design.

sampoder:
I'd be cool with that though idk how important the information is and want to avoid adding a bunch of extra SQL queries to more pages

app/views/organizer_positions/_organizer_position.html.erb
Luke-Oldenburg: Yea you're totally right. I'm kind of curious as to how it's working and not throwing an error with the typo I made.

Luke-Oldenburg:
I think we should keep it in the date invited column because it is also date info.
Screenshot From 2025-05-31 22-47-53

sampoder:
Yeah - that's chill, I think we're all referring to the card

Luke-Oldenburg:
Ah I see what you mean now. I'll move that.

Luke-Oldenburg:
Screenshot From 2025-06-01 17-05-27

garyhtou (approved):

leowilkin:
i like that - fab, thanks Luke & crew!

PR #10523, opened 2025-05-29
Skip RSpec PR run if the PR is from an external repo 

Mohamad-Mortada:
It is better to skip the PR run as it fails every time I make a PR as I make it on my own repo. Also a failed run emails me every time which is kinda annoying. This fixes that :)


sampoder:
Sorry - I won't be merging this. It's good to know that Rspec hasn't passed and it reminds the maintainer to run it locally if they believe their review needs that.

PR #10526, opened 2025-05-29
Remove PVSA from site

devramsean0:
## Summary of the problem

PVSA is dead (temporarily, but I don't have much hope for it's return any time soon :( )
[https://presidentialserviceawards.gov/](https://presidentialserviceawards.gov/)
## Describe your changes

Move the PVSA card down to the expired promotions section,
Also disabled all PVSA forms and the button

image


sampoder (approved):
LGTM! Rest in peace.

PR #10530, opened 2025-05-30
Add time frames to home page graphs

polypixeldev:
## Summary of the problem

The home page currently shows the text "ALL TIME" on the graphs, but there's no way to change the time frame of the data shown.

## Describe your changes

This PR makes the current "ALL TIME" text on each home page graph a menu that allows the user to select a valid time frame for the event. This is passed to the breakdown engine which uses the provided time frame to correctly calculate the sums.

The graphs, which were previously paired into turbo frames, are now separate turbo frames to allow independently updating the time frame for each graph.

image


app/models/event.rb
```diff
@@ -51,6 +51,11 @@
 #
 class Event < ApplicationRecord
   MIN_WAITING_TIME_BETWEEN_FEES = 5.days
+  BREAKDOWN_TIMEFRAMES = {
+    "Past year"  => 1.year,
+    "Past month" => 1.month,
+    "Past week"  => 1.week
+  }.freeze
```
garyhtou: Can we move this to `BreakdownEngine::TIMEFRAMES`
- Create a file called `app/services/breakdown_engine.rb`

app/services/breakdown_engine/categories.rb
```diff
@@ -2,9 +2,9 @@
 
 module BreakdownEngine
   class Categories
-    def initialize(event, past_month: false)
+    def initialize(event, **options)
       @event = event
-      @past_month = past_month
+      @timeframe = options[:timeframe]
```
garyhtou: Let's use keyword arguments:

```ruby
def initialize(event, timeframe: nil)
  @event = event
  @timeframe = timeframe
end
```

app/services/breakdown_engine/categories.rb
```diff
@@ -14,7 +14,12 @@ def run
       )
                                        .joins("LEFT JOIN canonical_transactions ct ON raw_stripe_transactions.id = ct.transaction_source_id AND ct.transaction_source_type = 'RawStripeTransaction'")
                                        .joins("LEFT JOIN canonical_event_mappings event_mapping ON ct.id = event_mapping.canonical_transaction_id")
-                                       .where("event_mapping.event_id = ? #{"AND raw_stripe_transactions.created_at > NOW() - INTERVAL '1 month'" if @past_month}", @event.id)
+                                       .where({
+                                         event_mapping: {
+                                           event_id: @event.id
+                                         },
+                                         raw_stripe_transactions: @timeframe.present? ? { created_at: (Time.now - @timeframe)..Time.now } : nil
```
garyhtou:
```suggestion
                                         raw_stripe_transactions: @timeframe.present? ? { created_at: (@timeframe.ago).. } : nil
```

app/services/breakdown_engine/tags.rb
```diff
@@ -2,13 +2,14 @@
...
-    def initialize(event)
+    def initialize(event, **options)
       @event = event
+      @timeframe = options[:timeframe]
     end
 
     def run
       tags = @event.tags.includes(hcb_codes: [:canonical_transactions, :canonical_pending_transactions]).each_with_object([]) do |tag, array|
-        amount_cents_sum = tag.hcb_codes.sum do |hcb_code|
+        amount_cents_sum = tag.hcb_codes.where(@timeframe.present? ? { created_at: (Time.now - @timeframe)..Time.now } : {}).sum do |hcb_code|
```
garyhtou:
```suggestion
        transactions = @timeframe.present? ? tag.hcb_codes.select { |hcb_code| hcb_code.date.after?(@timeframe.ago) } : tag.hcb_codes
        amount_cents_sum = transactions.sum do |hcb_code|
```

garyhtou (approved):

PR #10533, opened 2025-05-31
Ledger amount search

leowilkin:
## Summary of the problem

I'm building out my Wise extension, and I'd like to be able to click a button and search for an exact amount on the HCB ledger, unmapped, or mapped.

## Describe your changes

For this I'd need an amount box - so I added one to the admin controller, and the visual ledger html file.

image

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -308,6 +308,7 @@ class AdminController < ApplicationController
     @page = params[:page] || 1
     @per = params[:per] || 100
     @q = params[:q].present? ? params[:q] : nil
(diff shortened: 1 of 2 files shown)


leowilkin:
holdup am I super dumb and you can already search by amount using `@q`

sampoder (commented):
Code LGTM - maybe we remove amount sorting from `@q` and just use the dedicated field? This isn't too well documented anyways so I think a dedicated field would be better

leowilkin:
@sampoder yoink

sampoder (approved):
LGTM - thank you!

PR #10542, opened 2025-06-02
Add receipt deletion to v4 API

Mohamad-Mortada:
## Summary of the problem
In HCB Mobile you can add receipt but not delete it

## Describe your changes
This allows me to delete individual receipts.

### DIFF
--- a/app/controllers/api/v4/receipts_controller.rb
+++ b/app/controllers/api/v4/receipts_controller.rb
@@ -17,6 +17,17 @@ module Api
         render "show"
       end
 
+      def destroy
+        @receipt = Receipt.find(params[:id])
+        
+        if policy(@receipt).destroy?
+          @receipt.destroy!
+          render json: { message: "Receipt successfully deleted" }, status: :ok
+        else
+          render json: { error: "You are not authorized to delete this receipt" }, status: :forbidden
+        end
+      end
+      
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/api/v4/receipts_controller.rb
```diff
@@ -17,6 +17,18 @@ def create
...
+
+        if policy(@receipt).destroy?
+          @receipt.destroy!
+          render json: { message: "Receipt successfully deleted" }, status: :ok
+        else
+          render json: { error: "You are not authorized to delete this receipt" }, status: :forbidden
+        end
+      end
+
+
```
sampoder: Any reason we aren't directly doing `authorize @receipt`

Mohamad-Mortada: Nope, you're right it is more consistent to use authorize instead of using policy.

app/controllers/api/v4/receipts_controller.rb
```diff
@@ -17,6 +17,15 @@ def create
         render "show"
       end
 
+      def destroy
+        @receipt = Receipt.find(params[:id])
+        authorize @receipt
+
+        @receipt.destroy!
```
davidcornu, line 24: **note**: This had me wondering whether there is any additional logic in `ReceiptsController` but it's basically the same https://github.com/hackclub/hcb/blob/5d8171135465fe140559ae8ea9d972b855d1193d/app/controllers/receipts_controller.rb#L10-L33

davidcornu:
Ran the test suite locally

davidcornu (approved):

PR #10543, opened 2025-06-02
Added creating org invites to v4 api

Mohamad-Mortada:
## Summary of the problem
On HCB Mobile, I want to allow users to invite users to their org within the app. 

## Describe your changes
Added create to invitation route


app/controllers/api/v4/invitations_controller.rb
```diff
@@ -4,14 +4,44 @@ module Api
   module V4
     class InvitationsController < ApplicationController
       skip_after_action :verify_authorized, only: [:index]
-      before_action :set_invitation, except: [:index]
+      before_action :set_invitation, except: [:index, :create]
```
garyhtou:
```suggestion
      before_action :set_invitation, except: [:index]
```

app/controllers/api/v4/invitations_controller.rb
```diff
@@ -12,6 +12,36 @@ def index
 
       def show; end
 
+      def create
```
davidcornu, line 20: **concern**: We have a very similar non-API code path to create invitations https://github.com/hackclub/hcb/blob/78111dbd907ac185f7d52f2396d2804dd7d9fc63/app/controllers/organizer_position_invites_controller.rb#L20-L43 and it's hard to reason about the differences between the two. Can we centralize this logic in one service so it can be called from both places?

app/controllers/api/v4/invitations_controller.rb
```diff
@@ -28,17 +30,16 @@ def create
           return render json: { error: "User already has a pending invitation" }, status: :unprocessable_entity
         end
 
-        @invitation = OrganizerPositionInvite.new(
-          event: @event,
-          email: params[:email],
-          position_role: params[:role],
-          invited_by: current_user
-        )
+        service = OrganizerPositionInviteService::Create.new(event: @event, sender: current_user, user_email: params[:email], is_signee: false, role: params[:role], enable_spending_controls: params[:enable_spending_controls], initial_control_allowance_amount: params[:initial_control_allowance_amount])
```
Mohamad-Mortada, line 35: Ooooo I never knew we had a service for that

Luke-Oldenburg (approved):
lgtm!

sampoder (approved):
lgtm!

PR #10544, opened 2025-06-02
Improve transaction heatmap

polypixeldev:
## Summary of the problem

Currently, the transaction heatmap on the bottom of the home page never showed due to a bug in `BreakdownEngine::Heatmap`. It also did not include pending transactions and was showing the total income or spend on a certain day rather than the net difference, which was confusing.

## Describe your changes

Fixed the bug in `BreakdownEngine::Heatmap` (hash keys were strings, not dates), added pending transactions to the heatmap calculations, and changed the heatmap to show the net difference instead of the total income or spend.


garyhtou (approved):
Code looks good to me. @manuthecoder, are you cool with us changing the heat map to be calculated from the net transaction amount?

manu-gurudath:
yep, that sounds great

sampoder (approved):

PR #10545, opened 2025-06-02
Remove top_categories event route

polypixeldev:
## Summary of the problem

Noticed this route while working on another PR, but it doesn't have a controller method or view.

## Describe your changes

Removed the unused route and policy method.


sampoder (approved):
LGTM

PR #10546, opened 2025-06-02
Handle rate limiting in donation summary sending 

polypixeldev:
## Summary of the problem

> Net::SMTPServerBusy: 454 Throttling failure: Maximum sending rate exceeded.

This error was raised a couple days ago when the monthly donation summaries were sent since the job does not currently handle rate limits from SES.

## Describe your changes

Changed the monthly donation summary job to be a long-running job that will use exponential backoff when it is rate limited by SES.


app/jobs/monthly_donation_summary_job.rb
```diff
@@ -1,11 +1,22 @@
...
+      mail = nil
+      while mail.nil?
+        begin
+          mail = mailer.monthly_donation_summary.deliver_now
+        rescue Net::SMTPServerBusy
+          # Rate limited by SES
+          sleep [2**attempt, 32].min
+          attempt += 1
+        end
+      end
```
sampoder: Thoughts on using https://github.com/Shopify/limiter/blob/main/README.md#advanced-usage like we do for Loops?

garyhtou (approved):

PR #10553, opened 2025-06-03
[Flipper] Remove unused flags

Luke-Oldenburg:
- international_wires_2024_09_22
- user_permissions_2024_03_09
- receipt_bin_2023_04_07
- sms_receipt_notifications_2022_11_23
- email_updates_2024_05_24
- recently_on_hcb_2024_05_23
- card_logos_2024_08_27
- stripe_card_personalization_designs_2024_08_19

### DIFF
--- a/app/views/stripe_cards/_form.html.erb
+++ b/app/views/stripe_cards/_form.html.erb
@@ -79,28 +79,26 @@
     <% designs = [@event&.stripe_card_personalization_designs&.available, StripeCard::PersonalizationDesign.common.available].flatten.compact %>
     <% designs += StripeCard::PersonalizationDesign.unlisted.available if current_user.admin? %>
     <% if designs.any? %>
-      <% if Flipper.enabled?(:stripe_card_personalization_designs_2024_08_19, current_user) || (@event && Flipper.enabled?(:card_logos_2024_08_27, @event)) %>
-        <h2 class="grid__fill">Card design</h2>
-        <div class="field field--options" style="max-width: unset;">
-          <% designs.each do |design| %>
-            <%= form.radio_button :stripe_card_personalization_design_id, design.id %>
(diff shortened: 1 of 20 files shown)


app/views/stripe_cards/_form.html.erb
```diff
@@ -79,28 +79,26 @@
     <% designs = [@event&.stripe_card_personalization_designs&.available, StripeCard::PersonalizationDesign.common.available].flatten.compact %>
     <% designs += StripeCard::PersonalizationDesign.unlisted.available if current_user.admin? %>
     <% if designs.any? %>
-      <% if Flipper.enabled?(:stripe_card_personalization_designs_2024_08_19, current_user) || (@event && Flipper.enabled?(:card_logos_2024_08_27, @event)) %>
```
YodaLightsabr, line 82: This feature isn't fully enabled yet

YodaLightsabr (changes requested):
The feature `stripe_card_personalization_designs_2024_08_19` isn't fully enabled in production yet. Everything else looks good to go.

Luke-Oldenburg:
What is the difference between `stripe_card_personalization_designs_2024_08_19` and `card_logos_2024_08_27`?

YodaLightsabr:
I think `stripe_card_personalization_designs_2024_08_19` is the user permission. In this specific case, it should be fine though because the right half of the expression will always be true (`card_logos_2024_08_27` is fully enabled).

YodaLightsabr (approved):
I've just fully enabled `stripe_card_personalization_designs_2024_08_19`, so this is good to go!

Luke-Oldenburg:
I can't push to main.

garyhtou:
These have now been deleted from the Flipper dashboard in prod

PR #10567, opened 2025-06-04
Fix recent activity

Luke-Oldenburg:
Heading no longer shows if `@activities` is unset. Sets `@activities` from more actions so recent activity covers more of the my and users UI.


sampoder:
I think we should keep the current system - I'm just going to revert the changes made to this file in https://github.com/hackclub/hcb/commit/04c750a28265957c1e6846feb6d14787d9037c9d which caused the issue.

PR #10570, opened 2025-06-04
Add email to reimbursement reports

Luke-Oldenburg:
Before only admins could get a user's email from a reimbursement report.
Screenshot From 2025-06-04 15-19-37


YodaLightsabr:
This makes that header really tall. Has anyone requested this / found themselves needing this? This also isn't the first place something like this has come up; if we're looking to provide more information about mentioned users across the board, it might be beneficial to build out a custom tooltip.

sampoder:
Users should use the comment section to communicate with the person being reimbursed

sampoder:
Probably preemptive to close this I want to follow the discussion on Ian's questions but I don't think this should be in the header.

Luke-Oldenburg:
I think it'd be useful to have in a custom tooltip like it is already for admins. I wasn't sure that it would be okay to expose user emails everywhere though. I was running into some issues on Apex where I was trying to match a reimbursement to somebody in our DB and it is easier to search by email than name.

sampoder:
We shouldn't be doing that everywhere - can you add it to the settings panel as an uneditable field?

Luke-Oldenburg:
Like put the email in the reimbursement settings? That would work.

sampoder:
Yeah

app/views/reimbursement/reports/_edit_form.html.erb
```diff
@@ -1,6 +1,8 @@
 <%= form_with(model: @report, local: true) do |form| %>
   <%= form.label :report_name, "Report name" %>
   <%= form.text_field :report_name, placeholder: "Chicago Trip Expenses", required: true %>
+  <%= form.label :report_to, "Report to (this cannot be changed)", class: "mt2" %>
```
sampoder:
```suggestion
  <%= form.label :report_to, "Report to", class: "mt2" %>
```

app/views/reimbursement/reports/_edit_form.html.erb
```diff
@@ -1,6 +1,8 @@
 <%= form_with(model: @report, local: true) do |form| %>
   <%= form.label :report_name, "Report name" %>
   <%= form.text_field :report_name, placeholder: "Chicago Trip Expenses", required: true %>
+  <%= form.label :report_to, "Report to", class: "mt2" %>
+  <%= form.text_field :report_to, value: @user.email, readonly: true %>
```
sampoder:
```suggestion
  <%= form.text_field :report_to, value: @report.user.email, readonly: true %>
```

sampoder (approved):
LGTM1

PR #10573, opened 2025-06-04
Add tooltip to describe onetime grants

3kh0:
## Summary of the problem

How does someone figure out what a one-time-use grant is?

## Describe your changes

Sets the tooltip for the one time use button on grants to "Locks the card after a purchase" which helps people figure out what the heck it does


sampoder (changes requested):
Copy LGTM - can you add the tooltip directly on the button though?

Luke-Oldenburg:
@sampoder I don't think it's possible-- at least from my testing. Using a div is how we handled it on `stripe_cards/actions/freeze`. I superseded this in #10600 by the way because I had other changes to make to tooltips.

3kh0:
Superseded in #10600, thank you @Luke-Oldenburg!

PR #10577, opened 2025-06-05
Added create endpoint to V4 comments API

Mohamad-Mortada:
## Summary of the problem
I want to be able to add comments in HCB Mobile

## Describe your changes
Added a create method and a show which gets returned from create.


app/controllers/api/v4/comments_controller.rb
```diff
@@ -8,6 +8,28 @@ def index
         @comments = policy_scope(@hcb_code.comments).includes(:user).order(created_at: :asc)
       end
 
+      def create
+        @hcb_code = authorize HcbCode.find_by_public_id!(params[:transaction_id]), :show?
```
garyhtou: `authorize @comment` should be resposible for this

```suggestion
        @hcb_code = HcbCode.find_by_public_id!(params[:transaction_id])
```

garyhtou (approved):

PR #10583, opened 2025-06-05
WIP: Split disbursement into two HcbCodes

### LINKED ISSUE #10556: [HcbCode] Split Disbursements into two HcbCodes
Given this existing HcbCode: `HCB-500-15860` with `short_code` `NP4P3`, we'll split into two HcbCode:
- `HCB-501-15860`: Outgoing disbursement
  - Would contain one CanonicalPendingTransaction and one CanonicalTransaction. Both with negative amounts.
  - Short code: `NP4P3_legacy_outgoing`. We're choosing to include the existing HcbCode's short code in this new short code to make the paper trail more clear.
- `HCB-502-15860`: Incoming disbursement
  - Would contain one CanonicalPendingTransaction and one CanonicalTransaction. Both with positive amounts.
  - Short code: `NP4P3_legacy_incoming`
- The existing HcbCode record will remain and be unmodified
  - When visiting an existing HcbCode, it'll redirect.
    - Old urls include:
      - https://hcb.hackclub.com/hcb/HCB-500-15860 `<-- hcb_code string`
      - https://hcb.hackclub.com/hcb/PVHGR2x `<-- HcbCode.hashid`
    - Redirect to: (order of precedence)
      1. If you have access to outgoing, redirect to outgoing
      2. If you have access to incoming, redirect to incoming
      3. You don't have access to either, so 401

New Disbursements will have a randomly generated short code for incoming and separately for outgoing.

When visiting an outgoing/incoming disbursement, you should be able to click to the other as long as you have access to the other org.

We need to migrate associations that belong to HcbCode:
- `Receipt`
  - Copy to both new HcbCodes
  - Moving forward, new receipts and comments on an incoming/outgoing disbursement HcbCode will only be visible on that single HcbCode and won't be reflected in the other. This is how the code currently operates.
- `Comment`
  - Same as Receipt
- `HcbCode::Pin`
  - Only 3 for disbursements at the moment
- `Tag`
  - About 900 tag associations
- `SuggestedPairings`
  - Not as important to migrate
- Unrelated (no need to touch)
  - `HcbCode::PersonalTransaction`
  - At the time of writing (June 3rd), there are no personal transactions associated to any disbursements, which makes sense.
  - `reimbursement_expense_payout`
  - `reimbursement_payout_holding`

We'll need to update a handful of SQL queries that explicitly mention HcbCodes with `600`

## Process
1. Disbursement reading/mapping can use both formats
2. New disbursements create using new format
3. Migrate old disbursements
4. Remove old disbursement reading/mapping

garyhtou:
We were able to get a Two-HcbCode'ed Disbursement created using `DisbursementService::CreateV2`. It however errors when the async jobs are ran (largely due to error when rendering views for Turbostream broadcasts).

`Disbursement#hcb_code` and `Disbursement#local_hcb_code` may prove ot be a problem since it's likely called (duck-typing style) from the UI. Those two method currently don't have enough context as to which HcbCode to return (either outgoing/incoming) without adding a new parameter.

### DIFF
--- a/app/models/disbursement.rb
+++ b/app/models/disbursement.rb
@@ -10,6 +10,7 @@
 #  deposited_at             :datetime
 #  errored_at               :datetime
 #  in_transit_at            :datetime
+#  is_v2                    :boolean          default(FALSE), not null
 #  name                     :string
 #  pending_at               :datetime
 #  rejected_at              :datetime
@@ -191,7 +192,11 @@ class Disbursement < ApplicationRecord
   end
 
   # Eagerly create HcbCode object
-  after_create :local_hcb_code
+  after_create :create_local_hcb_code, if: -> { !is_v2 }
+
+  def create_local_hcb_code
+    HcbCode.create!(hcb_code: hcb_code)
+  end
 
   alias_attribute :approved_at, :pending_at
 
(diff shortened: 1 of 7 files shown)


app/models/disbursement.rb
```diff
@@ -202,10 +207,45 @@ def transferred_at
...
+    raise ArgumentError, "Disbursement is v2, use outgoing_local_hcb_code instead" unless is_v2
+
+    "HCB-#{TransactionGroupingEngine::Calculate::HcbCode::OUTGOING_DISBURSEMENT_CODE}-#{id}"
+  end
+
+  def outgoing_local_hcb_code
+    raise ArgumentError, "Disbursement is v2, use outgoing_local_hcb_code instead" unless is_v2
+
+    @local_hcb_code ||= HcbCode.find_or_create_by(hcb_code:)
+  end
```
sampoder, line 238: hmm - these error messages seem confusing?

garyhtou: yes, it is confusing. this code was just copied down while we were debugging. this PR is more in a proof of concept stage rather than actual code to be merged/reviewed.

garyhtou:
Handled by https://github.com/hackclub/hcb/pull/12753 and other PRs linked in https://github.com/hackclub/hcb/issues/12725

PR #10588, opened 2025-06-06
Only render freeze if stripe card is defined on card grant

sampoder:


Luke-Oldenburg:
It should be changed in the elsif too.

sampoder:
SGTM - thank you!

PR #10593, opened 2025-06-06
Fix add expense grammar

### LINKED ISSUE #10592: [Reimbursement] Fix grammar for new expense button on first expense
The new expense button currently says "Add an standard expense" which, to be grammatically correct, should be "Add a standard expense"

Image

Luke-Oldenburg:
Closes #10592


sampoder (approved):
lol thanks this is a carryover

PR #10600, opened 2025-06-08
[Card Grants] Stateful action tooltips

Luke-Oldenburg:
Tooltips on actions now update to show a warning if they are disabled or an informative message if enabled for all buttons.

credit to @3kh0 for original pr on updating one time use tooltip


app/views/card_grants/actions/_return.html.erb
```diff
@@ -1,23 +1,24 @@
 <%# locals: (card_grant:) %>
 
-<% if card_grant.active? %>
```
YodaLightsabr, line 3: Is there a reason we're removing this check?

app/views/card_grants/actions/_cancel.html.erb
```diff
@@ -1,13 +1,12 @@
 <%# locals: (card_grant:) %>
 
-<% if card_grant.active? %>
```
YodaLightsabr, line 3: Is there a reason we're removing this check?

app/views/card_grants/actions/_cancel.html.erb
```diff
@@ -1,13 +1,12 @@
...
-
+<% cancel_tooltip = policy(card_grant).cancel? ? "Cancel this grant and return funds to the organization" : "You don't have permission to perform this action" %>
```
YodaLightsabr, line 3: Since we're checking if a user has permissions before rendering the button, wouldn't it be better to hide it instead of showing a tooltip?

app/views/card_grants/actions/_return.html.erb
Luke-Oldenburg: It's checked inside of the actions partial already.

app/views/card_grants/actions/_cancel.html.erb
Luke-Oldenburg: It's checked inside of the actions partial already.

app/views/card_grants/actions/_cancel.html.erb
Luke-Oldenburg: Buttons are hidden for readers and grantees. Members get a disabled version that is consistent with the rest of the UI.

Luke-Oldenburg: See #10495

app/views/card_grants/actions/_toggle_one_time_use.html.erb
```diff
@@ -1,8 +1,11 @@
 <%# locals: (card_grant:) %>
 
-<%= link_to toggle_one_time_use_event_card_grant_path(id: card_grant.hashid, event_id: @event.slug),
-            method: :post,
-            class: "btn bg-slate",
-            disabled: !policy(card_grant).toggle_one_time_use? do %>
-  <%= inline_icon "private", size: 20 %> <%= card_grant.one_time_use ? "Disable" : "Enable" %> one time use
-<% end %>
+<% totu_tooltip = card_grant.one_time_use ? "Allow for unlimited authorizations on this card" : "Lock the card after one purchase" %>
+<div class="tooltipped tooltipped--n" aria-label="<%= policy(card_grant).toggle_one_time_use? ? totu_tooltip : "You don't have permission to perform this action" %>">
```
davidcornu: **suggestion**: I wonder if we'd be better off using full `if`/`else` expressions rather than having multiple ternaries (this applies to the other cases as well but this example is particularly illustrative)

```suggestion
<% 
  totu_tooltip = if policy(card_grant).toggle_one_time_use?
    if card_grant.one_time_use
      "Allow for unlimited authorizations on this card"
    else
      "Lock the card after one purchase"
    end 
  else
    "You don't have permission to perform this action"
  end
%>
<div class="tooltipped tooltipped--n" aria-label="<%= totu_tooltip %>">
```

app/views/card_grants/actions/_cancel.html.erb
davidcornu, line 3: **question**: There are a couple of cases where we no longer have this conditional. Is that intentional?

Luke-Oldenburg: Yes because it is already checked in the actions partial.

app/views/card_grants/actions/_toggle_one_time_use.html.erb
Luke-Oldenburg: I think that's the only case unless I'm missing something.

davidcornu (approved):

PR #10601, opened 2025-06-09
[Stripe Cards] Fix error on defrost

Luke-Oldenburg:
Cards were erroring when being defrosted despite the action going through because of the addition of one time use. This has been update to check for an actual error rather than the return value of the function.


YodaLightsabr (commented):
Is there an error you're encountering w/ the current logic? It seems like it would work just fine because we're running it at the end and checking if the grant is active first. If there is an error, it could probably be fixed by adding the same `if` statement to https://github.com/hackclub/hcb/blob/856a8d712eacad90e7e990be0fd67fd790c3a943/app/models/card_grant.rb#L207 as well.

Luke-Oldenburg:
Oh yea it should be fine then but I don't see any scenarios where that line is necessary.

sampoder:
Please keep that line - if we call the method `cancel!` method on a StripeCard it should also cancel the card grant and vice versa

sampoder (approved):

PR #10605, opened 2025-06-10
Use `X-Docuseal-Secret` header

sampoder:


YodaLightsabr (approved):

app/controllers/docuseal_controller.rb
```diff
@@ -8,6 +8,8 @@ def webhook
       contract = OrganizerPosition::Contract.find_by(external_id: params[:data][:submission_id])
       return render json: { success: true } unless contract # sometimes contracts are sent using Docuseal that aren't in HCB
 
+      return render json: { success: false } unless request.headers["X-Docuseal-Secret"] == Credentials.fetch(:DOCUSEAL, :WEBHOOK_SECRET)
```
garyhtou, line 11: Ideally, we use secure compare here.

PR #10609, opened 2025-06-10
[Receipts] Card locking

YodaLightsabr:
## Summary of the problem
Zach requested that we prevent users from spending their funds when they have 10 missing receipts

## Describe your changes
Adds the following, Flipper flagged for testing:
- Checks how many missing receipts a user has after every authorization
- If 10 or more, the user's cards are locked
- At certain intervals, we alert the user that they're nearing the 10 transaction limit

### DIFF
--- a/app/controllers/my_controller.rb
+++ b/app/controllers/my_controller.rb
@@ -88,7 +88,7 @@ class MyController < ApplicationController
 
   def inbox
     @count = current_user.transactions_missing_receipt.count
-    @locking_count = current_user.transactions_missing_receipt(since: Date.parse("2025-06-09")).count
+    @locking_count = current_user.transactions_missing_receipt(since: Receipt::CARD_LOCKING_START_DATE).count
 
     hcb_code_ids_missing_receipt = current_user.hcb_code_ids_missing_receipt
     @hcb_codes = Kaminari.paginate_array(HcbCode.where(id: hcb_code_ids_missing_receipt)
--- a/app/controllers/my_controller.rb
+++ b/app/controllers/my_controller.rb
@@ -88,6 +88,8 @@ class MyController < ApplicationController
 
   def inbox
     @count = current_user.transactions_missing_receipt.count
+    @locking_count = current_user.transactions_missing_receipt(since: Date.parse("2025-06-09")).count
+
     hcb_code_ids_missing_receipt = current_user.hcb_code_ids_missing_receipt
     @hcb_codes = Kaminari.paginate_array(HcbCode.where(id: hcb_code_ids_missing_receipt)
                  .includes(:canonical_transactions, canonical_pending_transactions: :raw_pending_stripe_transaction) # HcbCode#card uses CT and PT
--- a/app/controllers/stripe_controller.rb
+++ b/app/controllers/stripe_controller.rb
@@ -33,11 +33,13 @@ class StripeController < ActionController::Base
     # fire-and-forget update to grafana dashboard
     StatsD.increment("stripe_webhook_authorization", 1)
 
-    approved = ::StripeAuthorizationService::Webhook::HandleIssuingAuthorizationRequest.new(stripe_event: event).run
+    service = ::StripeAuthorizationService::Webhook::HandleIssuingAuthorizationRequest.new(stripe_event: event)
+    approved = service.run
 
     if approved
-      ::User::UpdateCardLockingJob.perform_later
-      ::User::SendCardLockingNotificationJob.perform_later
+      user = service.card.user
+      ::User::UpdateCardLockingJob.perform_later(user:)
+      ::User::SendCardLockingNotificationJob.perform_later(user:)
     end
 
     response.set_header "Stripe-Version", "2022-08-01"
--- a/app/controllers/stripe_controller.rb
+++ b/app/controllers/stripe_controller.rb
@@ -35,6 +35,11 @@ class StripeController < ActionController::Base
 
     approved = ::StripeAuthorizationService::Webhook::HandleIssuingAuthorizationRequest.new(stripe_event: event).run
 
+    if approved
+      ::User::UpdateCardLockingJob.perform_later
+      ::User::SendCardLockingNotificationJob.perform_later
+    end
+
     response.set_header "Stripe-Version", "2022-08-01"
 
     render json: {
--- a/app/models/receipt.rb
+++ b/app/models/receipt.rb
@@ -59,7 +59,7 @@ class Receipt < ApplicationRecord
   # - @sampoder
   PREPROCESSED_SIZES = ["1024x1024"].freeze
 
-  CARD_LOCKING_START_DATE = Date.new(2025, 6, 9)
+  CARD_LOCKING_START_DATE = Date.new(2025, 6, 13)
 
   has_one_attached :file do |attachable|
     PREPROCESSED_SIZES.each do |resize|
--- a/app/models/receipt.rb
+++ b/app/models/receipt.rb
@@ -87,7 +87,7 @@ class Receipt < ApplicationRecord
   end
 
   after_commit do
-    User::CardLockingJob.perform_later
+    User::UpdateCardLockingJob.perform_later
   end
 
   validate :has_owner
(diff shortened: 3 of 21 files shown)


app/models/user.rb
```diff
@@ -313,17 +314,28 @@ def hcb_code_ids_missing_receipt
     end
   end
 
-  def transactions_missing_receipt
+  def transactions_missing_receipt(since: nil)
```
sampoder: I don't think it's safe to do this, we're memoising this value but it can change based on the argument. 

I doubt this PR would directly introduce a bug but we should be avoiding this.

sampoder: `transactions_missing_receipt_count` looks good on this front

sampoder: Maybe we use https://github.com/matthewrudy/memoist - looks clean

app/services/user_service/send_card_locking_notification.rb
```diff
@@ -0,0 +1,27 @@
...
+module UserService
+  class SendCardLockingNotification
+    def initialize(user:)
+      @user = user
+    end
+
+    def run
+      return unless Flipper.enabled?(:card_locking_2025_06_09, @user)
+
+      count = @user.transactions_missing_receipt(since: Date.parse("2025-06-09")).count
```
sampoder:
```suggestion
      count = @user.transactions_missing_receipt(since: Date.new(2025, 6, 9)).count
```
instead of Date.parse

app/controllers/stripe_controller.rb
```diff
@@ -35,6 +35,11 @@ def handle_issuing_authorization_request(event)
 
     approved = ::StripeAuthorizationService::Webhook::HandleIssuingAuthorizationRequest.new(stripe_event: event).run
 
+    if approved
+      ::User::UpdateCardLockingJob.perform_later
+      ::User::SendCardLockingNotificationJob.perform_later
```
sampoder: Could we get a weird condition where the second job runs before the first? I feel like they should either be combined or `::User::UpdateCardLockingJob` should call `::User::SendCardLockingNotificationJob`

sampoder: Oh I see why you aren't doing that - to avoid sending these when someone uploads a receipt - hmm. Maybe `UpdateCardLockingJob` takes a parameter?

app/models/user.rb
YodaLightsabr: Oops, I meant to do it for both. Good catch. I might just ignore memorizing then, if the parameter exists.

app/services/user_service/send_card_locking_notification.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      if count.in?([5, 7, 9])
+        CardLockingMailer.warning(email: @user.email, missing_receipts: count).deliver_later
+
+        if @user.phone_number.present? && @user.phone_number_verified?
+          message = "You now have #{count}/10 transactions missing receipts. After 10, your cards will be locked. You can manage your receipts at #{Rails.application.routes.url_helpers.my_inbox_url}."
```
sampoder:
```suggestion
          message = "You now have #{count} transactions missing receipts. If you have ten or more missing receipts, your cards will be locked. You can manage your receipts at #{Rails.application.routes.url_helpers.my_inbox_url}."
```

app/controllers/stripe_controller.rb
YodaLightsabr: My concern with that is that then it gets confusing; it checks their mossing receipt count, potentially locks their card, and sends notifications. I would rather keep these separate just so that the naming is clear.

sampoder:
Can we make some improvements to the UI?

Screenshot 2025-06-09 at 11 56 18 PM

We should probably use the info callout for this that we use in other places. And if not, we should make sure that these are vertically aligned.

Screenshot 2025-06-09 at 11 58 50 PM

imo - we should drop this, it feels out of place on this page. also red != active.

Screenshot 2025-06-09 at 11 59 20 PM

imo the left badge should be info and the right badge should go from green (<= 3), orange (<=6) or red. And it should probably have a tooltip explaining it.

app/controllers/my_controller.rb
```diff
@@ -88,6 +88,8 @@ def missing_receipts_icon
 
   def inbox
     @count = current_user.transactions_missing_receipt.count
+    @locking_count = current_user.transactions_missing_receipt(since: Date.parse("2025-06-09")).count
```
garyhtou: Make the date a constant

app/services/user_service/update_card_locking.rb
```diff
@@ -0,0 +1,23 @@
...
+      return unless Flipper.enabled?(:card_locking_2025_06_09, @user)
+
+      count = @user.transactions_missing_receipt(since: Date.parse("2025-06-09")).count
+
+      if count >= 10 && !@user.cards_locked?
+        CardLockingMailer.cards_locked(email: @user.email, missing_receipts: count).deliver_later
+      end
+
+      cards_locked = count >= 10
+      @user.update!(cards_locked:)
```
garyhtou:
```suggestion
      cards_should_lock = count >= 10
      if cards_should_lock && !@user.cards_locked?
        CardLockingMailer.cards_locked(email: @user.email, missing_receipts: count).deliver_later
      end
      
      @user.update!(cards_locked: cards_should_lock)
```

app/views/card_locking_mailer/warning.html.erb
```diff
@@ -0,0 +1,3 @@
+Your HCB cards will risk getting locked if you don't upload your <%= @missing_receipts %> outstanding missing receipts. Once you hit 10 missing receipts, your cards will be locked until you upload them.
```
garyhtou:
```suggestion
Your HCB cards are at risk of getting locked if you don't upload your <%= @missing_receipts %> outstanding missing receipts. Once you hit 10 missing receipts, your cards will be locked until you upload them.
```

app/views/card_locking_mailer/cards_locked.html.erb
```diff
@@ -0,0 +1,3 @@
+Your HCB cards have been locked due to your <%= @missing_receipts %> outstanding missing receipts.
+<br><br>
+You can manage your receipts on HCB on the <%= link_to "receipts page", my_inbox_url %>.
```
garyhtou:
```suggestion
You can manage your receipts on HCB <%= link_to "receipts page", my_inbox_url %>.
```

app/views/card_locking_mailer/warning.html.erb
```diff
@@ -0,0 +1,3 @@
...
+<br><br>
+You can manage your receipts on HCB on the <%= link_to "receipts page", my_inbox_url %>.
```
garyhtou:
```suggestion
You can manage your receipts on HCB <%= link_to "receipts page", my_inbox_url %>.
```

app/views/hcb_codes/_decline_reason.html.erb
```diff
@@ -73,6 +75,8 @@
             If you believe that this merchant should be enabled, please reach out to the <%= hcb_code.event.name %> team.
         <% elsif hcb_decline_reason == :cash_withdrawals_not_allowed %>
             Cash withdrawals are disabled for most HCB cards. If you need to make a purchase with cash, try using personal funds and submitting a reimbursement report.
+        <% elsif hcb_decline_reason == :user_cards_locked %>
+            You can manage your receipts on HCB on the <%= link_to "receipts page", my_inbox_url %>. If you have more than 10 missing receipts, your cards will be locked until you submit them.
```
garyhtou:
```suggestion
            You can manage your receipts on HCB <%= link_to "receipts page", my_inbox_url %>. If you have 10 or more missing receipts, your cards will be locked until you submit them.
```

garyhtou (commented):
make threshold a constant

app/models/receipt.rb
```diff
@@ -59,6 +59,8 @@ class Receipt < ApplicationRecord
   # - @sampoder
   PREPROCESSED_SIZES = ["1024x1024"].freeze
 
+  CARD_LOCKING_START_DATE = Date.new(2025, 6, 9)
```
```suggestion
  CARD_LOCKING_START_DATE = Date.new(2025, 6, 13)
```

garyhtou: We'll make the announcement, and have it go live

garyhtou (approved):

garyhtou (approved):

PR #10611, opened 2025-06-10
[Card Grants] Make UI more intuitive for grant issuers and recipients

YodaLightsabr:
## Summary of the problem
Card grant users represent a significant chunk of our monthly active users. However, the experience is not great and feels like a very stripped-down UI. It doesn't integrate nicely with the rest of HCB, and it can feel quite bloated in certain areas that we haven't given much attention to.

## Describe your changes
This PR refactors many fields into a hero banner card to make certain elements more glanceable and reduce the monotony of the interface. It also abstracts organizer settings into a single settings view, accessible via a modal. Configuration options are split into tabs for organization.

image

image

image


app/controllers/card_grants_controller.rb
```diff
@@ -40,6 +40,62 @@ def create
...
+    @hcb_codes = @card_grant.visible_hcb_codes
+  end
+
+  def edit_withdraw
+    authorize @card_grant
+
+    @event = @card_grant.event
+    @card = @card_grant.stripe_card
+    @hcb_codes = @card_grant.visible_hcb_codes
+  end
```
davidcornu, line 97: **suggestion** (non-blocking): If we don't think these will need to diverge further we could probably delegate the instance variable setup to a method (e.g. `setup_edit_context`).

app/controllers/card_grants_controller.rb
```diff
@@ -67,7 +123,7 @@ def show
 
     authorize @card_grant
 
-    if @card_grant.pre_authorization_required? && @card_grant.pre_authorization&.draft? && !organizer_signed_in?
+    if @card_grant.pre_authorization_required? && (@card_grant.pre_authorization&.draft? || @card_grant.pre_authorization&.submitted? || @card_grant.pre_authorization&.rejected?) && !organizer_signed_in?
```
davidcornu, line 126: **suggestion** (non-blocking): I find it hard to reason about what this means from a business logic perspective. Might be helpful to name these concepts with local variables or methods.

app/models/card_grant.rb
```diff
@@ -120,6 +120,15 @@ def state_text
...
   end
 
+  def status_badge_type
+    s = state.to_sym
+    return :success if s == :success
+    return :error if s == :muted
+    return :warning if s == :info
+
+    :muted
+  end
```
davidcornu, line 130: **suggestion** (non-blocking): Seen as we're just returning `s` we could make this a bit more straightforward with a `case`.

```suggestion
  def status_badge_type
    s = state.to_sym
    case s
    when :success, :error, :warning
      s
    else
      :muted
    end
  end
```

app/policies/card_grant_policy.rb
```diff
@@ -17,6 +17,34 @@ def spending?
...
+    record.event.is_public? || user&.auditor? || user_in_event?
+  end
+
+  def edit_topup?
+    record.event.is_public? || user&.auditor? || user_in_event?
+  end
+
+  def edit_withdraw?
+    record.event.is_public? || user&.auditor? || user_in_event?
+  end
```
davidcornu, line 46: **suggestion** (non-blocking): Seen as we're repeating the same logic for all these actions we could either put it in a private method and call it from these methods or define one method and make the others aliases of it. Would make it clearer that they're all identical.

app/views/card_grant/pre_authorizations/show.html.erb
```diff
@@ -16,12 +16,18 @@
       <% end %>
     </h3>
   </div>
-  <div class="flex flex-row items-end justify-start gap-2 mb2">
-    <h3 class="m-0">Pre-authorization</h3>
-    <%= inline_icon "view-forward", class: "m-0 self-center", size: 20 %>
-    <p class="m-0">Before you can get spending, we need to know what you're planning to buy.</p>
+  <div class="flex flex-row items-center justify-start gap-2 mb2 lg:items-end">
+    <div class="flex items-start flex-col lg:flex-row lg:items-end justify-start gap-2">
+      <div class="flex flex-row items-end justify-start gap-2 w-min-content min-w-0 flex-shrink-0">
```
davidcornu, line 21: **concern** (non-blocking): This strikes me as _a lot_ of `flex`ing but without poking around in an inspector I can't really give you more useful feedback than that 😅.

app/views/card_grants/_edit_nav.html.erb
davidcornu, line 1: **concern**: This indentation is pretty funky in this file (e.g. lines 3, 30, 32)

app/views/card_grants/_header_banner.html.erb
```diff
@@ -0,0 +1,46 @@
...
+  <div class="flex flex-row justify-end items-start">
+    <% if organizer_signed_in?(as: :member) %>
+      <%= render "card_grants/actions/edit", card_grant: card_grant %>
+    <% end %>
+  </div>
+  <div class="flex-grow"></div>
+  <div class="flex flex-col gap-8 lg:flex-row justify-between items-start lg:items-center">
+    <div class="flex flex-col justify-center items-start mt-5">
+      <h2 class="border-none !mx-0 !px-0 !mt-0 inline-flex items-center">
+        Grant to &nbsp; <%= user_mention(card_grant.user, class: "mr-2") %> for <%= render_money card_grant.amount %>
```
davidcornu, line 13: **suggestion**: One space should be enough no?

```suggestion
        Grant to&nbsp;<%= user_mention(card_grant.user, class: "mr-2") %> for <%= render_money card_grant.amount %>
```

app/views/card_grants/_header_banner.html.erb
```diff
@@ -0,0 +1,46 @@
...
+      </h2>
+      <p class="my-0 flex items-baseline g-1">
+        <%= status_badge card_grant.status_badge_type %>
+        <%= card_grant.state_text %>
+        <% if card_grant.purpose.present? %>
+          • <%= card_grant.purpose %>
```
davidcornu, line 19: **nitpick**: Using a named escape code might make it clear this is intentional
```suggestion
          &middot; <%= card_grant.purpose %>
```

app/views/card_grants/_header_banner.html.erb
```diff
@@ -0,0 +1,46 @@
...
+    </p>
+    <p>
+      <strong>Sent</strong>
+      <span class="tooltipped" aria-label="<%= card_grant.created_at.to_formatted_s(:long) %>">
+        <%= time_ago_in_words card_grant.created_at %> ago
+      </span>
+    </p>
+  </section>
+  </div>
+  <% if defined?(footer) && footer.present? %>
```
davidcornu, line 41: **question**: I'm surprised we need `defined?` here as the `strict_locals` comment has a default value

```suggestion
  <% if footer.present? %>
```

Luke-Oldenburg (approved):

app/views/card_grants/edit_actions.html.erb
davidcornu, line 1: **concern**: More funky indentation (e.g. lines 4, 6)

app/views/card_grants/show.html.erb
```diff
@@ -1,12 +1,29 @@
...
+<% footer = nil %>
+<% if @card_grant.user == current_user && @card_grant.pending_invite? %>
+  <% footer = capture do %>
+    <div class="flex flex-row items-center justify-start gap-2 lg:items-end">
+      <div class="flex items-start flex-col lg:flex-row lg:items-end justify-start gap-2">
+        <div class="flex flex-row items-end justify-start gap-2 w-min-content min-w-0 flex-shrink-0">
+          <h3 class="m-0 p-0">Card activation</h3>
+          <%= inline_icon "view-forward", class: "m-0 lg:self-center", size: 20 %>
+        </div>
+        <p class="m-0">We just need one more thing before you can start spending&nbsp;your&nbsp;grant.</p>
```
davidcornu, line 13: **suggestion**: Instead of manually adding `&nbsp;` tags we could use a `<span>` around the text we don't want to wrap with a utility class.

davidcornu (approved):
No blockers but I'd appreciate a follow-up.

PR #10612, opened 2025-06-10
[Admin] Add phone number to users page

YodaLightsabr:
## Summary of the problem
#10610

## Describe your changes
Closes #10610

### DIFF
--- a/app/views/admin/users.html.erb
+++ b/app/views/admin/users.html.erb
@@ -30,6 +30,7 @@
       <th class="w-32">Created</th>
       <th>Name</th>
       <th>Email</th>
+      <th>Phone</th>
       <th>Org Positions</th>
       <th class="w-48">Card Grants</th>
       <th># Receipts Missing</th>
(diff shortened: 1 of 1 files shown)


app/views/admin/users.html.erb
```diff
@@ -52,6 +53,12 @@
             <br><small><i><%= email %></i></small>
           <% end %>
         </td>
+        <td>
+          <%= user.phone_number %>
+          <% unless user.phone_number_verified %>
+            <br><small><i>⚠️ Unverified</i></small>
+          <% end %>
```
garyhtou, line 60: nice!

garyhtou (commented):
Thanks for the quick turn around! Can we also make it searchable, if its not already

garyhtou:
Use case for searchability: @leowilkin wants to locate users with the same phone number (people who have multiple accounts).

YodaLightsabr:
I believe it is already! https://github.com/hackclub/hcb/blob/f7ccaa5bb9b185bf34af1824b607b9c00abc3bc2/app/models/user.rb#L57

garyhtou:
Perfect!

PR #10615, opened 2025-06-10
[Admin] Filter active teenagers within the last 30 days on the team page

scooterthedev:
## Summary of the problem

https://hackclub.slack.com/archives/C068U0JMV19/p1749575821982219 + @leowilkin ;)

## Describe your changes

Added a filter for admins only to see active teenagers for a specific org on the teams page.


app/controllers/events_controller.rb
```diff
@@ -274,7 +276,11 @@ def team
 
     @all_positions = @event.organizer_positions
                            .joins(:user)
-    @all_positions = @all_positions.where(role: @filter) if @filter
+    if @filter == "active_teens"
+      @all_positions = @all_positions.select { |op| op.user.teenager? && op.user.active? } # select if user is a teenager and active (stole from the other code ;))
```
sampoder, line 282: I think this will break as select forces the SQL query to run and then does the filtering in memory, on Line 284 though we seem to be adding more to the SQL query

scooterthedev: In my testing, nothing broke. but yah, that doesn't seem good. Any suggestions on how to fix it?

sampoder: I think move it to after L284

garyhtou: I'm not sure if the code's changed since this comment thread, but it seems to be fine now. L284 is inside an `elsif` so it won't run if L282 converts it into an array

garyhtou (approved):

PR #10616, opened 2025-06-11
[Google Workspace] Fix Revocation Timing Errors and Misc Errors

rluodev:
## Summary of the problem

Revocations failed because Google Workspace is slow to recognize that all accounts have been deleted after they were actually deleted. There were also a few other errors I took the time to fix.

## Describe your changes

Fixed errors by queueing jobs to run which re-queue up to two more times before erroring and requiring manual intervention to delete on Google Workspace end.


app/jobs/g_suite/delete_domain_users_job.rb
```diff
@@ -0,0 +1,30 @@
...
+      begin
+        Partners::Google::GSuite::DeleteUsersOnDomain.new(domain:).run
+        DeleteOrgUnitJob.set(wait: 1.minute).perform_later(domain:, remote_org_unit_path:)
+      rescue => e
+        if times_queued < 3
+          retry_job(wait: 2.minutes, domain:, remote_org_unit_path:, times_queued: times_queued + 1)
+          return
+        end
+        Rails.error.report("Failed to delete GSuite domain #{domain}'s users after 3 attempts (pls do manually): #{e.message}\nBacktrace: #{e.backtrace}")
+      end
```
garyhtou: Catch a more specific error, but also use https://api.rubyonrails.org/v8.0.2/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on

app/services/partners/google/g_suite/delete_domain.rb
```diff
@@ -12,13 +12,19 @@ def initialize(domain:)
...
             return
           end
           # groups and users must be deleted to be able to delete domain
           DeleteGroupsOnDomain.new(domain: @domain).run
-          DeleteUsersOnDomain.new(domain: @domain).run
-          directory_client.delete_domain(gsuite_customer_id, @domain)
+          gsuite = ::GSuite.find_by(domain: @domain)
+          gsuite&.accounts&.each do |account|
+            account.destroy
+          end
```
garyhtou:
```suggestion
          gsuite&.accounts&.destroy_all
```

app/services/partners/google/g_suite/delete_users_on_domain.rb
```diff
@@ -26,7 +26,7 @@ def run
             end
 
             Rails.error.report(e)
-            throw :abort
+            throw e
```
garyhtou:
```suggestion
            raise e
```

garyhtou (commented):
@davidcornu, could you please help review this PR? It may be a good idea to do a code walkthrough with @rluodev first since a bit of context is required for this section.

app/services/partners/google/g_suite/delete_domain.rb
```diff
@@ -12,13 +12,17 @@ def initialize(domain:)
...
           end
           # groups and users must be deleted to be able to delete domain
           DeleteGroupsOnDomain.new(domain: @domain).run
-          DeleteUsersOnDomain.new(domain: @domain).run
-          directory_client.delete_domain(gsuite_customer_id, @domain)
+          gsuite = ::GSuite.find_by(domain: @domain)
+          gsuite&.accounts&.destroy_all
+          # All jobs are queued one after another from within the delete domain users job
+          # to ensure that the domain is deleted only after all users and org unit are deleted
+          ::GSuite::DeleteDomainUsersJob.set(wait: 1.minute).perform_later(domain: @domain, remote_org_unit_path: gsuite.remote_org_unit_path)
```
davidcornu: **concern**: We're allowing for `gsuite` to be `nil` here by using `find_by` (vs. `find_by!`), which we're accounting for on line 22 (`gsuite&.accounts&.destroy_all`) but on line 25 (`gsuite.remote_org_unit_path`). 

1. Can it be `nil` at this point?
2. If (1), then let's update 25, if not we can use `find_by!` and remove the `&` on 22

app/services/partners/google/g_suite/delete_groups_on_domain.rb
```diff
@@ -17,7 +17,7 @@ def run
           end
           begin
             res = directory_client.list_groups(customer: gsuite_customer_id, domain: @domain)
-            res.groups.each do |group|
+            res.groups&.each do |group|
```
davidcornu, line 22: **suggestion**: Can we add a comment here explaining why the `&` is necessary. It isn't obvious that `.groups` can be `nil`.

app/services/partners/google/g_suite/delete_org_unit.rb
```diff
@@ -11,6 +11,8 @@ def initialize(org_unit_path:)
         end
 
         def run
+          return unless @org_unit_path.present?
+
```
davidcornu: **concern** (non-blocking): This reads as a bit odd to me. I'd lean towards pushing this check upstream and not calling `DeleteOrgUnit.new` with a blank `org_unit_path` in the first place, but I don't have context on those callsites.

app/services/partners/google/g_suite/delete_users_on_domain.rb
```diff
@@ -17,16 +17,14 @@ def run
           end
           begin
             res = directory_client.list_users(customer: gsuite_customer_id, domain: @domain, max_results: 500)
-            res.users.each do |user|
+            res.users&.each do |user|
```
davidcornu, line 22: Same comment as https://github.com/hackclub/hcb/pull/10616/files#r2183705778

app/services/partners/google/g_suite/delete_domain.rb
rluodev: done!

app/services/partners/google/g_suite/delete_org_unit.rb
rluodev: done!

davidcornu (approved):

PR #10630, opened 2025-06-12
Improve invoice amount UI

manuthecoder:
This PR improves the invoice creation experience by enhancing the aesthetic of the amount text

### Before
image

### After
image
image


sampoder:
Hmm - placing labels inside of fields doesn't feel like our design system. I don't think we should have the field itself expand to include errors.

PR #10642, opened 2025-06-16
Set paper trials in v4 api base controller

Mohamad-Mortada:
## Summary of the problem
Right now in the API I've noticed when I freeze a grantee's card it shows as "Last frozen by no user"and that's due to not setting paper trials 

## Describe your changes
Added paper trials to the application controller


sampoder (approved):
LGTM!

PR #10645, opened 2025-06-16
Add invoice filters

manuthecoder:
This pull request introduces enhancements to the filtering functionality for invoices, refactors the filtering logic for better flexibility, and updates the UI to dynamically render filter menus. 

### Backend Enhancements to Filtering Logic:

* **Refactored filter parameter handling in `def index`:** Changed the parameter from `filter` to `status` and added support for advanced filtering by amount (`amount_greater_than`, `amount_less_than`) and creation date (`created_after`, `created_before`). [[1]](diffhunk://#diff-2211725d26383c020697dc2008aa4044b9b56ebb010045878fff29caf36129f9L38-R38) [[2]](diffhunk://#diff-2211725d26383c020697dc2008aa4044b9b56ebb010045878fff29caf36129f9R51-R55)

* **Introduced `@filter_options` in `def index`:** Added a new instance variable defining available filters (`status`, `created_*`, `amount_*`) and their types, enabling dynamic filter configuration.

* **Updated `check_filters?` helper method:** Enhanced logic to support new filter types (`date_range`, `amount_range`) and ensure compatibility with the new filtering parameters.

### UI Updates for Dynamic Filters:

* **Replaced static filter links in `index.html.erb`:** Removed hardcoded filter links and replaced them with dynamic rendering of filter menus and filter bars based on `@filter_options` and `@has_filter`.

### Preview
image


app/controllers/invoices_controller.rb
```diff
@@ -90,6 +95,14 @@ def index
         @stats[:pending] += @invoices[i].item_amount
       end
     end
+
+    @filter_options = [
+      { key: "status", label: "Status", type: "select", options: %w[paid unpaid archived voided] },
+      { key: "created_*", label: "Date", type: "date_range" },
+      { key: "amount_*", label: "Amount", type: "amount_range" },
+    ]
```
davidcornu: **question**: Does this ever change? If not we could just store it in a constant

app/views/events/filters/_bar.html.erb
```diff
@@ -26,8 +26,8 @@
         <% param_key = "#{filter[:key].sub('_*', '')}_#{suffix}" %>
         <% if params[param_key].present? %>
           <div class="badge badge-large bg-muted">
-            <%= "#{filter[:label].gsub('Amount ', '').humanize} #{suffix.humanize}" %>
-            <%= Money.new(params[param_key].to_i, "USD").format %>
+            <%= suffix.humanize %>
+            <%= Money.new(params[param_key].to_i * 100, "USD").format %>
```
davidcornu, line 30: **question**: What exactly are we storing in `params[param_key]` and what has changed from before?

app/helpers/events_helper.rb
```diff
@@ -92,15 +92,63 @@ def show_org_switcher?
...
       end
     end
   end
 
+  def validate_filter_options(filter_options, params)
+    filter_options.each do |opt|
+      key = opt[:key].to_s
+      base = key.delete_suffix("_*")
+
+      case opt[:type]
```
davidcornu, line 111: **question**: Do we not want to validate `select`?

app/controllers/invoices_controller.rb
```diff
@@ -90,6 +95,14 @@ def index
         @stats[:pending] += @invoices[i].item_amount
       end
     end
+
+    @filter_options = [
+      { key: "status", label: "Status", type: "select", options: %w[paid unpaid archived voided] },
+      { key: "created_*", label: "Date", type: "date_range" },
+      { key: "amount_*", label: "Amount", type: "amount_range" },
```
davidcornu: **suggestion**: Is it important that `_*` be included in `key`? We seem to spend a lot of time removing it and we have logic that requires it to be there for the correct functioning of `date_range` and `amount_range` filters. I would either
- Split out `key` into `start_key` and `end_key` for range types
- Use `key` as the base and append the conventional suffixes

app/views/events/filters/_bar.html.erb
manu-gurudath: rewrote this logic so it would be a bit easier to read

app/helpers/events_helper.rb
manu-gurudath: `select` should generally be safe as this function is intended only for validating user input (such as date and amount ranges)

app/views/events/filters/_bar.html.erb
davidcornu: Right but if I'm reading this correctly we're now doing a `*100` where we weren't before

app/helpers/events_helper.rb
davidcornu: 🤔 my thinking was that we're specifying `options` for `select` ~and as far as I can tell we aren't actually checking that the param is one of those values~ but the only place we're validating that is in the `case` statement that applies the filters. That might be fine though.

davidcornu (approved):
Aside from the question about the new `* 100` (https://github.com/hackclub/hcb/pull/10645#discussion_r2212752850) this LGTM.

manu-gurudath:
> Aside from the question about the new `* 100` ([#10645 (comment)](https://github.com/hackclub/hcb/pull/10645#discussion_r2212752850)) this LGTM.

ah this was actually a bug with the old code which is now addressed in this pr

PR #10646, opened 2025-06-16
Enable `annotate_rendered_view_with_filenames` in development

davidcornu:
## Summary of the problem

Rails has a built-in mechanism to insert HTML comments around rendered template which makes it easier to figure out where parts of a page come from.

CleanShot 2025-06-16 at 16 47 43@2x

https://guides.rubyonrails.org/configuring.html#config-action-view-annotate-rendered-view-with-filenames

## Describe your changes

Flips the configuration setting to `true` in `config/environments/development.rb` (so it is only enabled in development mode).

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -80,7 +80,7 @@ Rails.application.configure do
   # config.i18n.raise_on_missing_translations = true
 
   # Annotate rendered view with file names.
-  # config.action_view.annotate_rendered_view_with_filenames = true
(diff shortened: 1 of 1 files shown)


sampoder (approved):
Thank you!

garyhtou (commented):
This is handy, thank you!

YodaLightsabr:
I didn't know this was a thing. Thanks for adding it!

PR #10647, opened 2025-06-17
[Org Home] Disable transactions if organizers aren't signed in

manuthecoder:
This pull request refines the logic for rendering transaction links in the `app/views/events/home` directory. The changes ensure that links are only clickable when the user is signed in as an organizer, improving the user experience and preventing unintended behavior for non-organizers.

### Updates to transaction link rendering:

* [`app/views/events/home/_money_movement.html.erb`](diffhunk://#diff-9d88ddf84e51d3d3103283019be589083a086d5f5a5d427078773bd62a62fbb4L13-R20): Updated the "money in" and "money out" transaction links to include a conditional check for `organizer_signed_in?`. Links are now only rendered with `href` and associated attributes if the user is an organizer. [[1]](diffhunk://#diff-9d88ddf84e51d3d3103283019be589083a086d5f5a5d427078773bd62a62fbb4L13-R20) [[2]](diffhunk://#diff-9d88ddf84e51d3d3103283019be589083a086d5f5a5d427078773bd62a62fbb4L51-R65)

* [`app/views/events/home/_transactions.html.erb`](diffhunk://#diff-f4b1ba8fc2c80c69b587c22ee12857e311dc00dcaee1208323ad47d27b8200f9L5-R13): Applied the same `organizer_signed_in?` conditional logic to the "recent transactions" links, ensuring consistent behavior across transaction displays.


app/views/events/home/_money_movement.html.erb
```diff
@@ -10,7 +10,13 @@
...
 
       <% @money_in.each do |transaction| %>
-        <a href="<%= url_for(transaction.local_hcb_code) %>" class="homepage-transaction <%= Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (@event || transaction.local_hcb_code.event).present? ? "cursor-pointer hover:bg-gray-50 active:bg-gray-100 transition-colors dark:hover:bg-neutral-700 dark:active:bg-neutral-600" : "" %>" data-behavior="modal_trigger" data-modal="money_movement_tx_details_<%= transaction.local_hcb_code.__id__ %>">
+        <a
+          class="homepage-transaction <%= organizer_signed_in? && Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (@event || transaction.local_hcb_code.event).present? ? "cursor-pointer hover:bg-gray-50 active:bg-gray-100 transition-colors dark:hover:bg-neutral-700 dark:active:bg-neutral-600" : "" %>"
+          <% if organizer_signed_in? %>
+          href="<%= url_for(transaction.local_hcb_code) %>"
+          data-behavior="modal_trigger"
+          data-modal="money_movement_tx_details_<%= transaction.local_hcb_code.__id__ %>"
+          <% end %>>
```
davidcornu: **suggestion**: If I'm reading this correctly, we're basically removing all useful classes and properties from this `<a>` tag when `organizer_signed_in?`, which leaves me with two questions:
1. Do we need this to be a `<a>` when there is no `href`?

    &rarr; The [`capture` helper](https://api.rubyonrails.org/classes/ActionView/Helpers/CaptureHelper.html#method-i-capture) might come in handy here so we can wrap the inner content with different outer elements.
3. Can we simplify the `class` attribute logic?

    &rarr; Would a local variable be more readable?

^ putting the above together we end up with something like

```erb
<% @money_in.each do |transaction| %>
  <% rendered_transaction = capture do %>
    <!-- What was previously in the `<a>` tag -->
  <% end %>  

  <% if organizer.signed_in? %>
    <!-- 👇 We should probably name this -->
    <% name_this_concept = Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (@event || transaction.local_hcb_code.event).present? %>
    <% css_classes = "cursor-pointer hover:bg-gray-50 active:bg-gray-100 transition-colors dark:hover:bg-neutral-700 dark:active:bg-neutral-600" if name_this_concept  %>
    <a
      class="homepage-transaction <%= css_classes %>" 
      href="<%= url_for(transaction.local_hcb_code) %>"
      data-behavior="modal_trigger"
      data-modal="money_movement_tx_details_<%= transaction.local_hcb_code.__id__ %>"
    >
      <%= rendered_transaction %>
    </a>
  <% else %>
    <div class="homepage-transaction">
      <%= rendered_transaction %>
    </div>
  <% end %>
<% end %>
```

manu-gurudath:
ty @davidcornu for the awesome detailed feedback! please let me know if anything else needs to be changed

garyhtou (approved):
Thank you! great review @davidcornu

PR #10648, opened 2025-06-17
Disable the HCB card logo field for members

### LINKED ISSUE #10464: Choose file isn't disabled for members despite having the property.
### Describe the bug

You are able to select a file to upload for the card logo as a reader even though it has the disabled property.

Image

https://github.com/hackclub/hcb/blob/9d308531227888e63cb5afe4ffbe405bc6b1b225/app/views/events/settings/_details.html.erb#L47

davidcornu:
## Summary of the problem

Fixes https://github.com/hackclub/hcb/issues/10464.

When a member views the event settings page, the HCB card logo field appears enabled despite the action being permitted. 

Note that they still wouldn't be able to submit the form as that action would be blocked by the pundit policy: https://github.com/hackclub/hcb/blob/7481706275b9d1492fa60965775438e49ece4ec7/app/policies/event_policy.rb#L57-L59

## Describe your changes

1. I've set the `disabled` property on `form.file_field` (mirroring other fields on the page https://github.com/hackclub/hcb/blob/f333eb2c034cc38cd43720c324fa5c0620988eaf/app/views/events/settings/_details.html.erb#L46-L47
2. I've added CSS to style the `<label>` and `<input>` contained in a `.field--fileupload` when it contains an element with the `disabled` property.
    CleanShot 2025-06-17 at 10 37 57@2x

### DIFF
--- a/app/assets/stylesheets/components/_forms.scss
+++ b/app/assets/stylesheets/components/_forms.scss
@@ -109,7 +109,8 @@ input[type='number'] {
   }
 
   &:has([disabled]) {
-    .field--fileupload__label, .field--fileupload__field {
+    .field--fileupload__label,
+    .field--fileupload__field {
       color: map-get($palette, slate);
       cursor: no-drop;
     }
--- a/app/assets/stylesheets/components/_forms.scss
+++ b/app/assets/stylesheets/components/_forms.scss
@@ -107,6 +107,23 @@ input[type='number'] {
 
     color: $muted;
   }
+
+  &:has([disabled]) {
+    .field--fileupload__label, .field--fileupload__field {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components/_forms.scss
```diff
@@ -107,6 +107,23 @@ input[type='number'] {
 
     color: $muted;
   }
+
+  &:has([disabled]) {
```
davidcornu, line 111: This should be safe to use: https://caniuse.com/css-has

sampoder (approved):
Awesome - code LGTM! Thank you for adding this!

PR #10649, opened 2025-06-17
Flipper for card grant popovers

manuthecoder:
This pull request introduces a feature flag to conditionally enable card popovers on the home page 

Feature flag implementation:

* [`app/views/static_pages/index.html.erb`](diffhunk://#diff-6fbbdc20d8dca32618e9da61910e0944c71a852662adaeb66108e0aad8940808L3-R3): Added a line to check if the `:hcb_code_popovers_2023_06_16` feature flag is enabled for the current user and set `@show_card_popovers` accordingly.

image


app/views/static_pages/index.html.erb
```diff
@@ -1,6 +1,6 @@
 <% title "Home" %>
 <% page_md %>
-
+<% @show_card_popovers = Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) %>
```
davidcornu, line 3: - **concern** (non-blocking): This threw me for a loop until I realized we reference this instance variable in a couple of partials and already have similar lines of code in other places. In general I don't expect templates to be setting instance variables as by convention that's usually done in the controller. There might be an opportunity to make this a bit less surprising by
  1. Moving the logic that sets `@show_card_popovers` into the appropriate controllers (or to a shared helper if it's the same everywhere)
  2. Using `locals` (https://guides.rubyonrails.org/layouts_and_rendering.html#passing-local-variables) to pass `@show_card_popovers` to partials rather than having partials reference it directly
- **question**: In [other cases](https://github.com/search?q=repo%3Ahackclub%2Fhcb%20%2F%40show_card_popovers%20%3D%2F&type=code) we've restricted this further by checking `organizer_signed_in?` as well (e.g. https://github.com/hackclub/hcb/commit/624af3f8bdfd998dba2e8b5bfa56fb25979c0295). Do we not need that here?

Luke-Oldenburg: @davidcornu long term goal is to replace instance variables in partials with locals. I've started work on this. I agree that it should be set in the controller.

manu-gurudath: @davidcornu that's a really good point, should we save it for another PR since `@show_card_popovers` touches multiple modals?

davidcornu: Yep we can do that

davidcornu (approved):

manu-gurudath:
interesting how the checks failed...

image

PR #10650, opened 2025-06-17
[Card Grants] Improve actions with meatball menu

manuthecoder:
The goal of this PR is to reduce the visual clutter for card grant settings:

### Before
image

### After
image

### DIFF
--- a/app/views/card_grants/_actions.html.erb
+++ b/app/views/card_grants/_actions.html.erb
@@ -3,7 +3,8 @@
 <% if card_grant.active? %>
   <div class="mt3 flex flex-wrap align-items-center justify-start g1 justify-center">
     <% if !organizer_signed_in?(as: :member) || card_grant.user == current_user %>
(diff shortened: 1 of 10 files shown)


Luke-Oldenburg:
I think freeze and purpose should still be under the card because on regular stripe cards the actions that are available are freeze, rename, and burn.

Luke-Oldenburg:
And even maybe get reimbursed. I like topup, withdraw, and one time use in the meatball though.

manu-gurudath:
I agree, keeping "Freeze" and "Purpose" does make it feel consistent with regular stripe cards. 

image

Maybe keeping "Get reimbursed" under the menu might be a good idea though, since it's not part of the regular cards UI, but I did move it to the top

Luke-Oldenburg:
Yea that seems good.

manu-gurudath:
See #10611

PR #10651, opened 2025-06-17
[Invoices] Add sorting ability

manuthecoder:
This PR introduces invoice sorting functionality for users, with plans to extend sorting to other parts of the platform soon.

image


app/views/events/_sort_table_header.html.erb
```diff
@@ -0,0 +1,17 @@
+<% @active = params[:sort] == sort || (params[:sort].nil? && defined?(default) && default) %>
```
garyhtou:
```suggestion
<% active = params[:sort] == sort || (params[:sort].nil? && defined?(default) && default) %>
```

app/controllers/invoices_controller.rb
```diff
@@ -50,11 +50,18 @@ def index
 
     relation = relation.search_description(params[:q]) if params[:q].present?
 
-    @invoices = relation.order(created_at: :desc)
+    allowed_sorts = %w[created_at status sponsors.name amount_due]
+    allowed_directions = %w[asc desc]
+
+    sort_column = params[:sort].in?(allowed_sorts) ? params[:sort] : "created_at"
+    sort_direction = params[:direction].in?(allowed_directions) ? params[:direction] : "desc"
+    relation = relation.order("#{sort_column} #{sort_direction}")
```
davidcornu: **concern**: We probably shouldn't be coupling sort parameters to the SQL needed to apply them

```suggestion
    allowed_directions = %w[asc desc]
    sort_direction = params[:direction].in?(allowed_directions) ? params[:direction] : "desc"
    
    sort_column = 
      case params[:sort]
      when "created_at", "status", "amount_due"
        params[:sort]
      when "sponsor_name"
        "sponsors.name"
      else
        "created_at"
      end

    relation = relation.order("#{sort_column} #{sort_direction}")
```

app/views/events/_sort_table_header.html.erb
```diff
@@ -0,0 +1,6 @@
+active = params[:sort] == sort || (params[:sort].nil? && defined?(default) && default)
+  direction = params[:sort] == sort ? (params[:direction] == "asc" ? "desc" : "asc") : (params[:direction] || "desc")
+  classes << "sort-button--active" if active
+  classes "sort-button--right" if defined
+<%= link_to upsert_query_params(sort:, direction:), class: classes.join(" ") do %>
+<% end %>
```
davidcornu: **concern**: This doesn't look right.
  - We have Ruby logic that doesn't appear to be wrapped in ERB tags
  - There's an empty block on the `link_to`

app/views/invoices/index.html.erb
```diff
@@ -59,10 +59,18 @@
...
+      </th>
+      <th>
+        <%= render "events/sort_table_header", sort: "date", default: true %>
+      </th>
+      <th>
+        <%= render "events/sort_table_header", sort: "sponsors.name", display: "To" %>
+      </th>
+      <th>
+        <%= render "events/sort_table_header", sort: "amount_due", right: true, display: "Amount" %>
+      </th>
```
davidcornu: - **suggestion**: Rather than deal with the indirection of a `render` here, might I suggest we write this as a loop over the sort parameters and their labels?
- **concern**: We have a dependency here between the `sort` values and the controller, which must me identical for this to work. It might be worth putting them in a constant so it's clear we're referring to the same thing.

app/views/events/_sort_table_header.html.erb
manu-gurudath: something weird happened here interesting

app/controllers/invoices_controller.rb
```diff
@@ -55,14 +62,34 @@ def index
...
+    sort_direction = params[:direction].in?(allowed_directions) ? params[:direction] : "desc"
+
+    sort_column =
+      case params[:sort]
+      when "created_at", "status", "amount_due"
+        params[:sort]
+      when "sponsor_name"
+        "sponsors.name"
+      else
+        "created_at"
```
garyhtou: This is duplicated logic since you defined `default: true` in `INVOICE_COLUMNS`

app/controllers/invoices_controller.rb
```diff
@@ -55,14 +62,34 @@ def index
...
-    @invoices = relation.order(created_at: :desc).includes(:sponsor).page(params[:page]).per(25)
+    allowed_directions = %w[asc desc]
+    sort_direction = params[:direction].in?(allowed_directions) ? params[:direction] : "desc"
+
+    sort_column =
+      case params[:sort]
+      when "created_at", "status", "amount_due"
+        params[:sort]
+      when "sponsor_name"
+        "sponsors.name"
```
garyhtou: It's okay if we include a `.` (dot) in the `key`

app/controllers/invoices_controller.rb
```diff
@@ -55,14 +62,34 @@ def index
...
+      else
+        "created_at"
+      end
+
+    if organizer_signed_in?
+      relation = relation.left_joins(:sponsor) if sort_column == "sponsors.name"
```
garyhtou: Maybe this should be configured in `INVOICE_COLUMNS` too

app/controllers/invoices_controller.rb
```diff
@@ -55,14 +62,34 @@ def index
...
+      @invoices = relation.order(sort_column => sort_direction).includes(:sponsor).page(params[:page]).per(25)
+    else
+      @invoices = relation.order(created_at: :desc).includes(:sponsor).page(params[:page]).per(25)
+    end
```
garyhtou: Instead of duplicating the `@invoices = ...` line, i would suggest setting `sort_column` `sort_direction` to `created_at` and `desc` when `!organizer_signed_in`

garyhtou (approved):

PR #10652, opened 2025-06-17
Mention required homebrew packages in setup docs

davidcornu:
## Summary of the problem

On a fresh install, `bundle install` will fail because of missing native dependencies.

## Describe your changes

Adds an extra line to the setup instructions to mention them.


sampoder (approved):
LGTM - thank you!

PR #10653, opened 2025-06-17
[WIP] Start drafting sudo mode RFC

davidcornu:
## Summary of the problem

#6346 

## Describe your changes

[Rendered](https://github.com/hackclub/hcb/blob/david/push-lywmwopxprpp/dev-docs/rfcs/001_sudo_mode/sudo_mode.md)


YodaLightsabr:
A couple of thoughts:
- I think keeping the current Login model (which you outlined as option 1) might be the best path forward. we would likely need to make some changes to make it support both initial logins and re-authentication steps, but it helps keep the controller/view logic in 1 place
- this idea of re-authentication could be extended beyond sudo mode. instead of logging people out completely, we could just require 1 factor to re-authenticate. Google does this a lot, and it helps keep your account secure even if sensitive actions aren’t being taken.
  - I think it might be beneficial to consider both time-based expiration and inactivity-based expiration. if we implement re-authentication, we could re-authenticate after inactivity but still force a full login after a set period of time (say, 30 days).
- for question #\2, I don’t think we should send re-authentication emails. that’s probably overkill.
- re question #\3, I’m totally onboard with ending impersonation sessions after 2 hours.
- for question #\4, I think it would be cool to include them inactivities.
- your approach to #\5 seems good

dev-docs/rfcs/001_sudo_mode/sudo_mode.md
```diff
@@ -0,0 +1,104 @@
...
+4. Should re-authentications appear in activities?
+5. How should this work within https://github.com/hackclub/hcb-mobile? Should we return a known error code so the mobile app can prompt the user to re-authenticate?
+
+### How do we model the re-authentication?
+
+**Option 1 - Treat it as a new `Login`**
+
+Given that we have an existing flow for authenticating users, we could treat re-authentication as a new login flow which, once complete, swaps in a new `UserSession`.
+
+We would ideally link the new `Login` to the previous one so we can distinguish it as a re-authentication, apply different rules (e.g. only requiring one factor), and maintain a chain of `Login` records (for auditing purposes).
```
sampoder, line 85: I think this a good idea, we could do something like `voided_by_id` to store the ID of the past `Login`

dev-docs/rfcs/001_sudo_mode/sudo_mode.md
```diff
@@ -0,0 +1,104 @@
...
+**Open questions**
+
+1. We currently show users a list of their sessions including how long ago they logged in. Should we also surface information about re-authentications?
+
+   <img src="sessions.png" width="200"/>
+2. We send users an email when there is a new login on their account. Would we want to do the same for re-authentications?
+
+   <img src="login_notification_email.png" width="200"/>
+3. How do we want to handle sudo actions for users that are being impersonated? Is re-authenticating the impersonator enough?
+    - _Opinion_: We should strongly consider making this question go away by capping impersonated sessions to 2 hours. I don't think there's a use case for having these persist for longer than that.
```
sampoder, line 75: Yes- this is a good idea

dev-docs/rfcs/001_sudo_mode/sudo_mode.md
```diff
@@ -0,0 +1,104 @@
...
+
+1. We currently show users a list of their sessions including how long ago they logged in. Should we also surface information about re-authentications?
+
+   <img src="sessions.png" width="200"/>
+2. We send users an email when there is a new login on their account. Would we want to do the same for re-authentications?
+
+   <img src="login_notification_email.png" width="200"/>
+3. How do we want to handle sudo actions for users that are being impersonated? Is re-authenticating the impersonator enough?
+    - _Opinion_: We should strongly consider making this question go away by capping impersonated sessions to 2 hours. I don't think there's a use case for having these persist for longer than that. 
+4. Should re-authentications appear in activities?
```
sampoder, line 76: Yes but with custom copy

dev-docs/rfcs/001_sudo_mode/sudo_mode.md
```diff
@@ -0,0 +1,104 @@
...
+1. Users who initiate a sudo action and subsequently abandon it (e.g. because they don't have access to their email or TOTP) should remain logged in.
+2. Re-authentication only requires a single factor of the user's choosing
+3. Once they are re-authenticated we won't require them to do so for another two hours
+
+**Open questions**
+
+1. We currently show users a list of their sessions including how long ago they logged in. Should we also surface information about re-authentications?
+
+   <img src="sessions.png" width="200"/>
+2. We send users an email when there is a new login on their account. Would we want to do the same for re-authentications?
```
sampoder, line 71: We normally only do this if the context is unrecognised, I think it's ok to not do this

dev-docs/rfcs/001_sudo_mode/sudo_mode.md
```diff
@@ -0,0 +1,104 @@
...
+
+**Assumptions**
+
+1. Users who initiate a sudo action and subsequently abandon it (e.g. because they don't have access to their email or TOTP) should remain logged in.
+2. Re-authentication only requires a single factor of the user's choosing
+3. Once they are re-authenticated we won't require them to do so for another two hours
+
+**Open questions**
+
+1. We currently show users a list of their sessions including how long ago they logged in. Should we also surface information about re-authentications?
```
sampoder, line 68: Yes but imo we should flatten them into one Login

PR #10660, opened 2025-06-18
Validate `User#session_duration_seconds`

davidcornu:
## Summary of the problem

We allow users to pick from a fixed list of session durations

Setting | Options
---- | ----
CleanShot 2025-06-18 at 10 30 28@2x | CleanShot 2025-06-18 at 10 30 36@2x

which come from https://github.com/hackclub/hcb/blob/8508ac95625ca08aef11b7919596ddb8f68c0665/app/helpers/sessions_helper.rb#L4-L11

However, we don't actually validate this setting which means you can technically set it to anything (something I have confirmed in production).

## Describe your changes

For the sake of documentation and consistency, I've added basic validations to the field to make sure it is always a value from the list.

I've also checked the production database with `select session_duration_seconds, count(*) from users group by session_duration_seconds` to make sure all the current values are valid.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -165,6 +165,8 @@ class User < ApplicationRecord
 
   validates :preferred_name, length: { maximum: 30 }
 
+  validates(:session_duration_seconds, presence: true, inclusion: { in: SessionsHelper::SESSION_DURATION_OPTIONS.values })
+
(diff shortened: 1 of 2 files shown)


sampoder (approved):
LGTM! Thank you!

PR #10663, opened 2025-06-19
Block readers from uploading + deleting receipts

sampoder:
Read-only role means read only receipts.

### DIFF
--- a/app/policies/receipt_policy.rb
+++ b/app/policies/receipt_policy.rb
@@ -4,7 +4,7 @@ class ReceiptPolicy < ApplicationPolicy
   def destroy?
     user&.admin? ||
       (record&.receiptable&.event &&
-        OrganizerPosition.role_at_least?(user, record&.receiptable&.event, :member) &&
+        OrganizerPosition.role_at_least?(user, record.receiptable.event, :member) &&
         unlocked?
       ) ||
       # Checking if receiptable is nil prevents unauthorized
--- a/app/policies/receipt_policy.rb
+++ b/app/policies/receipt_policy.rb
@@ -3,8 +3,8 @@
 class ReceiptPolicy < ApplicationPolicy
   def destroy?
     user&.admin? ||
-      (record&.receiptable&.event && 
-        OrganizerPosition.role_at_least?(user, record&.receiptable&.event, :member) && 
+      (record&.receiptable&.event &&
+        OrganizerPosition.role_at_least?(user, record&.receiptable&.event, :member) &&
         unlocked?
       ) ||
       # Checking if receiptable is nil prevents unauthorized
--- a/app/policies/receipt_policy.rb
+++ b/app/policies/receipt_policy.rb
@@ -3,7 +3,10 @@
 class ReceiptPolicy < ApplicationPolicy
   def destroy?
     user&.admin? ||
-      (record&.receiptable&.event&.users&.include?(user) && unlocked?) ||
+      (record&.receiptable&.event && 
+        OrganizerPosition.role_at_least?(user, record&.receiptable&.event, :member) && 
+        unlocked?
+      ) ||
       # Checking if receiptable is nil prevents unauthorized
       # deletion when user no longer has access to an org
       (record&.receiptable.nil? && record&.user == user) ||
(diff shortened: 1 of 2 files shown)


app/policies/receipt_policy.rb
```diff
@@ -3,7 +3,10 @@
 class ReceiptPolicy < ApplicationPolicy
   def destroy?
```
davidcornu, line 4: **concern**: This method is becoming very hard to reason about, particularly because it's all one big expression.

We could refactor it to something like this:

```ruby
def destroy?
  return false unless record && user

  return true if user.admin?

  event = record.receiptable&.event

  # In the case where there is an associated event, we rely on the user's role
  # within it to determine their access
  if event
    user_has_event_access = OrganizerPosition.role_at_least?(user, event, :member) 

    return user_has_event_access && unlocked?
  end

  # If there isn't an event, users are only allowed to destroy records they own
  return false unless record.user == user

  case record.receiptable
  # Checking if receiptable is nil prevents unauthorized deletion when user no
  # longer has access to an org
  when nil
    true
  when Reimbursement::Expense
    unlocked?
  else
    false
  end
end
```

app/policies/receiptable_policy.rb
```diff
@@ -18,7 +18,7 @@ def upload?
   def present_in_events?
     # Assumption: Receiptable has an association to Event
     events = record.try(:events) || [record.event]
-    events.select { |e| e.try(:users).try(:include?, user) }.present?
+    events.select { |e| e.try(:users).try(:include?, user) && OrganizerPosition.role_at_least?(user, e, :member) }.present?
```
davidcornu: **suggestion**: There are a few improvements we might be able to make here:

1. We can remove the need to treat `e` as potentially `nil` by doing `[record.event].compact` ([docs](https://docs.ruby-lang.org/en/3.4/Array.html#method-i-compact))
2. We can skip `e.try(:users).try(:include?, user)` altogether because `OrganizerPosition.role_at_least?` already checks that the user is part of the event: https://github.com/hackclub/hcb/blob/565f03d338fa6cb562fdb40028519a76ef8c0cf9/app/models/organizer_position.rb#L61 This is technically more permissive as it shortcuts the check for admins https://github.com/hackclub/hcb/blob/565f03d338fa6cb562fdb40028519a76ef8c0cf9/app/models/organizer_position.rb#L59 where `e.try(:users).try(:include?, user)` would have ensured the user was actually part of the event,  however we're allowing admins access anyway https://github.com/hackclub/hcb/blob/4d37755955b535389c637ab4739a221287a7aa10/app/policies/receiptable_policy.rb#L4-L14
3. Instead of `.select` + `.present?` we can can be a bit more efficient by using `.any?` ([docs](https://docs.ruby-lang.org/en/3.4/Array.html#method-i-any-3F)) which won't construct an intermediate array and halt as soon as we find a match.

Putting this all together we get

```ruby

  def present_in_events?
    # Assumption: Receiptable has an association to Event
    events = record.try(:events) || [record.event].compact
    events.any? { |event| OrganizerPosition.role_at_least?(user, event, :member) }
  end
```

app/policies/receipt_policy.rb
sampoder: Yes - this is reasonable, I'm going to merge this as it's fixing a bug but I've done this rewrite and I'm going to stick it in a PR:

```ruby
def destroy?  
    return true if user.admin?
    return false if record.nil?
    
    # any members of events should be able to modify receipts.
    if record.receiptable.event
      return true if OrganizerPosition.role_at_least?(user, record.receiptable.event, :member) && unlocked?
    end
    
    # the receipt is in receipt bin.
    if record.receiptable.nil?
      return record.user == user
    end
    
    # the receipt is on a reimbursement report. people making reports may not be in the organization.
    if record.receiptable.instance_of?(Reimbursement::Expense)
      return true if record.receiptable.report.user == user && unlocked?
    end
    
    return false
  end
```

app/policies/receiptable_policy.rb
sampoder: Thanks for investigating this - sgtm!

app/policies/receipt_policy.rb
sampoder: Spun into https://github.com/hackclub/hcb/pull/10665

PR #10664, opened 2025-06-19
Add an `invoiced_at` to `AchTransfer`

sampoder:
Mel was looking to keep track of these.

### DIFF
--- a/app/models/ach_transfer.rb
+++ b/app/models/ach_transfer.rb
@@ -14,6 +14,7 @@
 #  company_entry_description :string
 #  company_name              :string
 #  confirmation_number       :text
+#  invoiced_at               :date
 #  payment_for               :text
 #  recipient_email           :string
 #  recipient_name            :string
@@ -102,6 +103,8 @@ class AchTransfer < ApplicationRecord
   end
   validates :company_entry_description, length: { maximum: 10 }, allow_blank: true
   validates :company_name, length: { maximum: 16 }, allow_blank: true
+  # validates :invoiced_at, presence: true, on: :create
+  validate :invoiced_at_must_be_before_or_on_creation_date, on: :create
 
   has_one :t_transaction, class_name: "Transaction", inverse_of: :ach_transfer
   has_one :raw_pending_outgoing_ach_transaction, foreign_key: :ach_transaction_id
@@ -356,6 +359,12 @@ class AchTransfer < ApplicationRecord
   def scheduled_on_must_be_in_the_future
     if scheduled_on.present? && scheduled_on.before?(Date.today)
       errors.add(:scheduled_on, "must be in the future")
+    end
+  end
+
+  def invoiced_at_must_be_before_or_on_creation_date
+    if invoiced_at.present? && invoiced_at.after?(created_at&.to_date || Date.current)
+      errors.add(:invoiced_at, "cannot be after the transfer creation date")
     end
   end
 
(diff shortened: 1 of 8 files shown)


app/models/ach_transfer.rb
```diff
@@ -102,6 +103,8 @@ def payment_recipient_attributes
   end
   validates :company_entry_description, length: { maximum: 10 }, allow_blank: true
   validates :company_name, length: { maximum: 16 }, allow_blank: true
+  # validates :invoiced_at, presence: true, on: :create
+  validate :invoiced_at_must_be_before_or_on_creation_date, on: :create
```
davidcornu: 1. **suggestion**: I think you can get the same result from [the built-in validators](https://guides.rubyonrails.org/active_record_validations.html#comparison)

    ```suggestion
      validates(
        :invoiced_at, 
        comparison: { 
          less_than_or_equal_to: :created_at,
          message: "cannot be after the transfer creation date"
        },
        allow_nil: true,
        on: :create
      )
    ```
2. **suggestion**: Can we add a spec for this behaviour in `spec/models/ach_transfer_spec.rb`?

app/views/admin/ach_start_approval.html.erb
```diff
@@ -115,6 +115,12 @@
         <td style="text-align: right;">Payment for:</td>
         <td><%= @ach_transfer.payment_for %></td>
       </tr>
+      <% if @ach_transfer.invoiced_at.present? %>
+        <tr>
+          <td style="text-align: right;">Invoice date:</td>
+          <td><%= @ach_transfer.invoiced_at&.strftime("%B %-d, %Y") %></td>
```
davidcornu: **suggestion**: We've already checked `@ach_transfer.invoiced_at.present?` at this point so we don't need to use the safe navigation operator here.

```suggestion
          <td><%= @ach_transfer.invoiced_at.strftime("%B %-d, %Y") %></td>
```

db/schema.rb
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
```
davidcornu: **concern**: This is odd. Why was it removed?

db/schema.rb
```diff
@@ -12,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2025_06_10_000245) do
+ActiveRecord::Schema[7.2].define(version: 2025_06_19_045020) d
```
davidcornu: oops

```suggestion
ActiveRecord::Schema[7.2].define(version: 2025_06_19_045020) do
```

db/schema.rb
sampoder: This gets auto removed every time you run a migration, I was working on this on a plane, so I didn't have the internet to push my second commit

app/models/ach_transfer.rb
sampoder: Let's do it!

sampoder:
```suggestion
  validates(
    :invoiced_at, 
    comparison: { 
      less_than_or_equal_to: :created_at,
      message: "cannot be after the transfer creation date"
    },
    allow_nil: true,
    on: :create
  )
```

db/schema.rb
```diff
@@ -1,4 +1,4 @@
-# frozen_string_literal: true
+# frozen_string_literal: true
```
```suggestion
# frozen_string_literal: true
```

db/schema.rb
davidcornu: Probably not the end of the world TBH. If Rails is omitting it I don't think the upside of a teeny teeny bit less memory use is worthwhile.

PR #10665, opened 2025-06-19
Rewrite `ReceiptPolicy#destroy?`

sampoder:
More readable, spun out of https://github.com/hackclub/hcb/pull/10663


app/policies/receipt_policy.rb
```diff
@@ -3,13 +3,38 @@
...
       (record&.receiptable.instance_of?(Reimbursement::Expense) && record&.user == user && unlocked?)
   end
 
+  def destroy?  
+    return true if user.admin?
+    return false if record.nil?
+    
+    # any members of events should be able to modify receipts.
+    if record.receiptable.event
+      return true if OrganizerPosition.role_at_least?(user, record.receiptable.event, :member) && unlocked?
```
davidcornu: **question**: I don't believe the original logic falls through if the user doesn't have access for the event. Is the `if` here intentional?

```suggestion
      return OrganizerPosition.role_at_least?(user, record.receiptable.event, :member) && unlocked?
```

app/policies/receipt_policy.rb
```diff
@@ -3,13 +3,38 @@
...
+        OrganizerPosition.role_at_least?(user, record.receiptable.event, :member) &&
+        unlocked?
+      ) ||
       # Checking if receiptable is nil prevents unauthorized
       # deletion when user no longer has access to an org
       (record&.receiptable.nil? && record&.user == user) ||
       (record&.receiptable.instance_of?(Reimbursement::Expense) && record&.user == user && unlocked?)
   end
 
+  def destroy?
```
davidcornu: Looks like a merge gone wrong here (likely why CI is failing). We have two `destroy?` method definitions.

app/policies/receipt_policy.rb
```diff
@@ -3,13 +3,38 @@
...
       (record&.receiptable.nil? && record&.user == user) ||
       (record&.receiptable.instance_of?(Reimbursement::Expense) && record&.user == user && unlocked?)
   end
 
+  def destroy?  
+    return true if user.admin?
+    return false if record.nil?
+    
+    # any members of events should be able to modify receipts.
+    if record.receiptable.event
```
davidcornu: **concern**: If `record.receiptable` can be `nil`, then we can't just call `.event` here

```suggestion
    if record.receiptable&.event
```

app/policies/receipt_policy.rb
```diff
@@ -3,13 +3,38 @@
...
+      return true if OrganizerPosition.role_at_least?(user, record.receiptable.event, :member) && unlocked?
+    end
+    
+    # the receipt is in receipt bin.
+    if record.receiptable.nil?
+      return record.user == user
+    end
```
davidcornu, line 11: **suggestion**: We could push this condition further up as that would then let us assume `record.receiptable` exists in subsequent code (e.g. it could live on line 19).

app/policies/receipt_policy.rb
```diff
@@ -3,13 +3,38 @@
...
+    
+    # the receipt is on a reimbursement report. people making reports may not be in the organization.
+    if record.receiptable.instance_of?(Reimbursement::Expense)
+      return true if record.receiptable.report.user == user && unlocked?
```
davidcornu: **question**: We probably don't need this to fall through

```suggestion
      return record.receiptable.report.user == user && unlocked?
```

app/policies/receipt_policy.rb
```diff
@@ -2,16 +2,25 @@
...
-        OrganizerPosition.role_at_least?(user, record.receiptable.event, :member) &&
-        unlocked?
-      ) ||
-      # Checking if receiptable is nil prevents unauthorized
-      # deletion when user no longer has access to an org
-      (record&.receiptable.nil? && record&.user == user) ||
-      (record&.receiptable.instance_of?(Reimbursement::Expense) && record&.user == user && unlocked?) ||
-      (user && record&.receiptable.try(:card_grant)&.user == user && record.receiptable.card_grant.active?)
+    return true if user.admin?
+    return false if record.nil?
```
Luke-Oldenburg, line 5: Can the record ever be nil and if so should this trump the admin check?

app/policies/receipt_policy.rb
```diff
@@ -2,16 +2,25 @@
...
+
+    # the receipt is in receipt bin.
+    if record.receiptable.nil?
+      return record.user == user
+    end
+
+    # any members of events should be able to modify receipts.
+    if record.receiptable.event
+      return OrganizerPosition.role_at_least?(user, record.receiptable.event, :member) && unlocked?
```
Luke-Oldenburg, line 15: Do we want members to be able to modify receipts if it's for a reimbursement report that doesn't belong to them? We allow for it currently but I think it should be disabled.

app/policies/receipt_policy.rb
garyhtou: It probably should trump the admin check

Luke-Oldenburg:
```suggestion
    return false if record.nil?
    return true if user.admin?
```

app/policies/receipt_policy.rb
```diff
@@ -2,16 +2,25 @@
...
+    end
+
+    # the receipt is on a reimbursement report. people making reports may not be in the organization.
+    if record.receiptable.instance_of?(Reimbursement::Expense)
+      return record.receiptable.report.user == user && unlocked?
+    end
+
+    return false
```
garyhtou, line 23:
```suggestion
    false
```

garyhtou (approved):

PR #10666, opened 2025-06-19
Add `rails-pg-extras` so we can keep tabs on PostgreSQL

davidcornu:
## Summary of the problem

Given that AppSignal doesn't give us a tonne of visibility into the overall health of the database, I figured we could get a few more insights by leveraging https://github.com/pawurb/rails-pg-extras (e.g. missing or unused indexes).

CleanShot 2025-06-19 at 12 35 35@2x

## Describe your changes

- Adds `rails-pg-extras` to the `Gemfile`
- Adds a migration to enable to required extensions
- Adds an initializer to disable the built-in password authentication and limit available actions
- Mounts the web interface behind `AdminConstraint` at `/pg_extras`


db/migrate/20250619154630_enable_pg_extras_extensions.rb
```diff
@@ -0,0 +1,7 @@
+class EnablePgExtrasExtensions < ActiveRecord::Migration[7.2]
+  def change
+    enable_extension "pg_stat_statements"
```
davidcornu, line 3: For some reason this is present in the schema https://github.com/hackclub/hcb/blob/82d4a1b735250a89f184778c4f695c765c672d5d/db/schema.rb#L18 but isn't enable in production (checked this by running `select * from pg_extension` in production).

db/schema.rb
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
```
davidcornu, line 2: This gets removed by rails so may not be worth fighting

db/migrate/20250619154630_enable_pg_extras_extensions.rb
garyhtou: I wonder if it's an artifact of moving from Heroku

db/schema.rb
garyhtou: Yeah, the churn on this line is annoying. I think we've got a couple of options:
1. Have rubocop ignore the frozen string literal check for this file
2. Update the Rails generator template to include frozen string literal

Any thoughts on which is better?

garyhtou:
^ oops, that i think i did something wrong with that merge commit— i'll check that out in a sec

garyhtou:
properly merged now

garyhtou:
I ran into this error while testing locally
image


A quick search suggests configuring `shared_preload_libraries = 'pg_stat_statements'` inside postgresql.conf. @davidcornu do you happen to have more information on this, and what we'll need to do in production to get this deployed?

db/schema.rb
davidcornu: Given that this is auto-generated I'm inclined to go with (1) as it incurs less of a maintenance burden than (2).

davidcornu: Oh it's already ignored: https://github.com/hackclub/hcb/blob/8cf688f7c0b21fe030dbb153a374d2bc86b3c4cf/.rubocop.yml#L7

davidcornu:
> A quick search suggests configuring `shared_preload_libraries = 'pg_stat_statements'` inside postgresql.conf. 

That's what I had to do locally. Let me check production.

PR #10667, opened 2025-06-19
Expire impersonated sessions after 1 hour

davidcornu:
## Summary of the problem

Admin users have the ability to impersonate other users for debugging and support purposes. However while thinking through the implications of implementing a sudo mode (see https://github.com/hackclub/hcb/pull/10653#discussion_r2155544529) we realized these impersonated sessions live for as long as the impersonated user's `session_duration_seconds`, which could be much longer than necessary.

## Describe your changes

- `spec/rails_helper.rb` now eagerly-loads all files in `spec/support` so we can put shared code in there
- Added `SessionSupport` module to make writing controller tests easier
- Updated the `:user_session` factory to populate the `session_token` field
- Updated the logic in `SessionHelper#sign_in` to cap impersonated sessions to one hour
- Added a test for the above

### DIFF
--- a/app/helpers/sessions_helper.rb
+++ b/app/helpers/sessions_helper.rb
@@ -10,6 +10,10 @@ module SessionsHelper
     "30 days" => 30.days.to_i
   }.freeze
 
+  # For security reasons we severely restrict the duration of impersonated
+  # sessions
+  IMPERSONATED_SESSION_DURATION = SESSION_DURATION_OPTIONS.fetch("1 hour")
+
   def impersonate_user(user)
     sign_out
     sign_in(user:, impersonate: true)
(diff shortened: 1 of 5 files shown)


app/helpers/sessions_helper.rb
```diff
@@ -10,6 +10,10 @@ module SessionsHelper
     "30 days" => 30.days.to_i
   }.freeze
 
+  # For security reasons we severely restrict the duration of impersonated
+  # sessions
+  IMPERSONATED_SESSION_DURATION = SESSION_DURATION_OPTIONS.fetch("1 hour")
```
davidcornu, line 15: This is one hour instead of two (as we were discussing) because we still need it to be a valid value for `UserSession` (see #10660). If this is two short we can always add a two hour option to the model's validation.

sampoder (approved):
lgtm! thank you

app/helpers/sessions_helper.rb
garyhtou: 1 hour sounds good to me!

PR #10548, opened 2025-06-02
Add one time use card grants

polypixeldev:
## Summary of the problem

https://hackclub.slack.com/archives/CN523HLKW/p1746313970709819
Closes #10508 

## Describe your changes

Adds a one time use option to card grants that will automatically freeze its Stripe card when an authorization is approved. Managers can defrost, which will allow the grantee to make any amount of additional transactions on the card, but the option can be re-enabled with a new action on the card grant page.

image

PR #10668, opened 2025-06-19
Commit missing `one_time_use` annotation

davidcornu:
## Summary of the problem

https://github.com/hackclub/hcb/pull/10548 introduced a new `one_time_use` column on the `card_grants` table but the automatic annotation seems to be missing from the model file.

## Describe your changes

Commits the missing annotation


garyhtou (approved):
Thank you! Let's see if we can add a CI check for this

PR #10670, opened 2025-06-20
Fix Airbrake message in  `Login` validation

davidcornu:
## Summary of the problem

When we generate a message for Airbrake we reference `session` which does not exist.

## Describe your changes

Replaces `session` with `user_session`.


sampoder (approved):
Thank you!

PR #10671, opened 2025-06-20
Feature-flagged sudo mode for 2fa setting

davidcornu:
RFC: #10653

The following PRs were extracted from this one and shipped separately:
- https://github.com/hackclub/hcb/pull/10780
- https://github.com/hackclub/hcb/pull/10826
- https://github.com/hackclub/hcb/pull/10903

## Changes

- Adds the model relationships described in https://github.com/hackclub/hcb/pull/10903#issue-3220305247 
  - Adds `logins` and `initial_login` relations to `UserSession`
  - Adds `initial_login` relation to `Login`
- Updates the logic in `Login` to only require a single factor in the case of a reauthentication
- Adds `UserSession#sudo_mode?` to determine whether sudo mode is enabled based on the creation times of its associated logins
- Adds `SudoModeHandler` which handles intercepting requests and rendering the reauthentication page
- Requires sudo mode if the 2fa setting is changed and the user has the feature flag enabled
- Extensive testing

image


app/controllers/users_controller.rb
```diff
@@ -20,6 +20,7 @@ class UsersController < ApplicationController
                                                :complete_sms_auth_verification,
                                                :start_sms_auth_verification]
   before_action :set_shown_private_feature_previews, only: [:edit, :edit_featurepreviews, :edit_security, :edit_admin]
+  before_action :enforce_sudo_mode, only: [:update]
```
sampoder: This is super clean - thank you for implementing it this way

app/models/login.rb
sampoder, line 1: The changes to this model LGTM - only comment is I would have implemented `preceding_login` instead of inital

app/views/sudo_mode/reauthenticate.html.erb
```diff
@@ -0,0 +1,21 @@
+<h1>Confirm Access</h1>
+
+<p>Signed in as <%= current_user.email %></p>
+
+<%= form_tag(request.path, method: request.request_method_symbol) do %>
+  <%
+    Rack::Utils
+      .parse_query(request.request_parameters.to_query)
```
sampoder: I think we could make this a partial that we include in all the forms rendered from the `LoginsController` and then have some shared logic in `LoginsController#complete` to handle input from that.

sampoder (commented):
Model changes LGTM - as you mentioned in the video, I think the big thing will be reusing as much of the login controller code as possible. I'm happy to jump on a call to talk through that stuff at some point.

app/models/login.rb
davidcornu: That was my original plan but the approach I ended up landing on is to have everything point to the original login rather than have a chain in order to maintain the original login's lifetime (I felt it was simpler that way). Happy to switch it up if there's a compelling reason though.

davidcornu:
Rebased on top of https://github.com/hackclub/hcb/pull/10826 to shrink the diff

app/views/sudo_mode/reauthenticate.html.erb
```diff
@@ -0,0 +1,95 @@
...
+            <% when :totp %>
+              Use a one-time password
+            <% when :webauthn %>
+              Use your security key
+            <% end %>
+          <% end %>
+        </li>
+      <% end %>
+    </ul>
+  <% end %>
```
garyhtou, line 94: Oh my gosh, this rendering logic is sooo much cleaner than our current normal login UI

app/controllers/sudo_mode_handler.rb
```diff
@@ -0,0 +1,218 @@
+# frozen_string_literal: true
+
+class SudoModeHandler
+  # The default rank for authentication factors (lower is better)
+  # - 0 is deliberately left available so we can use it for the user's preference
+  # - `:backup_code` is explicitly omitted as we don't want to use that for
+  #   reauthentication
+  FACTOR_PREFERENCES = {
+    totp: 1,
+    webauthn: 2,
```
garyhtou: Should we switch the order of `webauthn` and `totp`? I wonder if WebAuthn is more secure espeically since it supports hardware security keys and biometrics.

app/helpers/sessions_helper.rb
```diff
@@ -158,4 +158,23 @@ def sign_out_of_all_sessions(user = current_user)
...
+  # It can either be used as a `before_action` callback or as part of an action
+  # implementation if you only want to require sudo mode in specific cases. In
+  # the latter scenario, you _MUST_ check the return value and only proceed if
+  # it is `true`.
+  #
+  # @return [Boolean] whether sudo mode was obtained and the controller action can proceed
+  def enforce_sudo_mode
+    return true if sudo_mode?
+
+    SudoModeHandler.new(controller_instance: self).call
```
garyhtou, line 178: clever!

app/models/user_session.rb
```diff
@@ -83,6 +90,20 @@ def expired?
...
   end
 
+  SUDO_MODE_TTL = 2.hours
+
+  # Determines whether the user can perform a sensitive action without
+  # reauthenticating.
+  #
+  # @return [Boolean]
+  def sudo_mode?
+    return true unless Flipper.enabled?(:sudo_mode_2015_07_21, user)
```
garyhtou, line 100: Such a small and clean way to feature flagging this!

garyhtou (commented):
So far so good! I left a few comments and will come back to finish my review

app/views/sudo_mode/reauthenticate.html.erb
```diff
@@ -0,0 +1,95 @@
...
+      For additional security, please enter the code from your authenticator app.
+    </p>
+
+    <%= code_field.call(placeholder: "Enter your one-time password") %>
+  <% when :webauthn %>
+    <p>
+      For additional security, please authenticate with your security key.
+    </p>
+
+    <%= hidden_field_tag("_sudo[webauthn_response]", nil) %>
```
davidcornu, line 57: **note**: This currently doesn't work as I haven't wired up the front-end logic. Will address that in a follow-up PR.

garyhtou (approved):
Very nice! I'll need to spend some more time understanding the parameters, but i'm happy with where we're at right now!

Please go ahead and merge this before the sync!

garyhtou:
Oh yes, and a screenshot or two in the PR description would be nice

app/controllers/sudo_mode_handler.rb
davidcornu: Can do!

PR #10677, opened 2025-06-23
Standardize canceled spelling

Luke-Oldenburg:
Only changed it in the UI.


davidcornu (approved):
Thanks @Luke-Oldenburg!

PR #10685, opened 2025-06-24
[Org Documents] Sort by `created_at: desc`

### LINKED ISSUE #10684: [Org Documents] Sort by created at desc
@LucyHackClub requested that we sort documents so that recently uploaded docs show up first.

At the moment, they are rendered with organization-specific docs first, then common docs afterwards. Within those "groups", they are unsorted.

I don't have a preference on whether we keep those groups or not, but I do think we should sort them by `created_at: :desc`.

manuthecoder:
Closes #10684

This PR orders org documents by their creation date, descending

### DIFF
--- a/app/views/documents/index.html.erb
+++ b/app/views/documents/index.html.erb
@@ -21,6 +21,7 @@
 
 <ul class="grid">
   <%= render partial: "document", collection: @active_documents %>
+  <%= render partial: "document", collection: @active_common_documents %>
 
   <% if @event.approved? && !@event.demo_mode? %>
     <%= link_to event_verification_letter_path(event_id: @event.slug, format: "pdf") do %>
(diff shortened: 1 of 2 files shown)


app/views/documents/index.html.erb
```diff
@@ -58,7 +57,6 @@
         <article>
           <ul class="grid">
             <%= render partial: "document", collection: @archived_documents %>
-            <%= render partial: "document", collection: @archived_common_documents %>
```
sampoder: How come this is being removed?

garyhtou: @manuthecoder, I think we want to keep the common documents (and archived docs too). The portion of this PR that adds `order(created_at: :desc)` looks good! Keep all the existing instance variables and apply the `order` to them. I believe the view can remain untouched.

manu-gurudath:
done!

sampoder (approved):
lgtm!

PR #10700, opened 2025-06-24
Small fixes for one time use grants

polypixeldev:
## Summary of the problem
One time use grants weren't setting a papertrail when frozen. The location in the details was also bad and the grantee had no indication of it being one time use.

## Describe your changes

Adds papertrail to be bank@hackclub.com when frozen, moves the detail row, and adds copy to show one time use to the grantee.


app/services/stripe_authorization_service/create_from_webhook.rb
```diff
@@ -45,7 +45,9 @@ def run
           end
 
           if cpt&.stripe_card&.card_grant&.one_time_use
-            cpt.stripe_card.freeze!
+            PaperTrail.request(whodunnit: User.find_by(email: "bank@hackclub.com").id) do
```
sampoder: Can you use the new system user created by @davidcornu?

davidcornu (approved):

PR #10701, opened 2025-06-24
Unpin IRB version

polypixeldev:
## Summary of the problem

The IRB version was pinned to 1.14.3 since console1984 did not work with the latest version. This has been fixed in the latest version of console1984.

## Describe your changes

Unpinned IRB in the Gemfile. Tested in development with console1984 enabled to make sure it works.


garyhtou (approved):
Thank you!

PR #10702, opened 2025-06-24
Fix the mega button

polypixeldev:
## Summary of the problem
The reimbursement add expense dropdown button was rendering as a "mega button" on mobile.

## Describe your changes

Enforced the width of the right button with an `!important`
image


garyhtou (approved):
great!

Issue #9659, opened 2025-02-25
Transition Column imports from reports to webhooks


sampoder:
For context, at the moment, our imports from Column work as described in here: https://github.com/hackclub/hcb/blob/66f2fbc4698ec9a6cfa3799495a0010ad2263920/dev-docs/guides/transaction_imports.md?plain=1#L3

Specifically they are idempotent using two factors: the report's ID and the index (row number) of the transaction in the report. 

This works, but it has caused us issues in the past, specifically when changing reporting timezone. It also means that we do this as a nightly job and money movement isn't instant.

It should also be possible to use a webhook to listen for certain event types, including but not limited to:

```
swift.outgoing_transfer.returned
ach.outgoing_transfer.initiated
wire.outgoing_transfer.initiated
swift.incoming_transfer.completed
swift.outgoing_transfer.initiated
interest.payout.completed
book.transfer.completed
check.outgoing_debit.returned
wire.incoming_transfer.completed
check.incoming_debit.settled
ach.outgoing_transfer.returned
ach.incoming_transfer.returned
ach.outgoing_transfer.canceled
ach.outgoing_transfer.settled
ach.incoming_transfer.settled
check.outgoing_debit.settled
```

We should determine the full list from: https://column.com/docs/workingwithapi/events-and-webhooks.

And, where appropriate, create a `RawColumnTransaction` using the webhook's data. The rest of the transaction will work as intended.

To tackle this, we'd need to:

1) Add `column_id` and `column_transaction_type` columns to `RawColumnTransaction`
2) Backfill this data into our existing `RawColumnTransaction`s and add code to the model to write to these columns.
    1) We probably also want to add a validation that this data is the same as the data stored in `column_transaction`
4) Add a not-null and unique constrain to these columns (it has to be a unique combination of the two).
5) Extend `Column::WebhooksController` to handle creating `RawColumnTransaction`s.
6) Write a job to import missed events using https://column.com/docs/api/#events/list-all-webhooks, just in case. This can replace `Column::StatementJob`

sampoder:
I did a bit more research into this, it's going to be slightly difficult to work out the amount in this system. Should be easy for initiations but we'll have to check that returns are as simple as `-amount`.

PR #10704, opened 2025-06-25
Begin creating `RawPendingColumnTransaction` and switch Column HCB codes to having the Column ID

sampoder:
Context behind this implementation is available at https://github.com/hackclub/hcb/issues/9659.


sampoder (commented):
TODO: filter out incoming money transactions without an event-specific account number

app/controllers/column/webhooks_controller.rb
```diff
@@ -11,6 +11,12 @@ def webhook
       type = params[:type]
       if type == "ach.incoming_transfer.scheduled"
         handle_ach_incoming_transfer_scheduled
+      elsif type == "ach.incoming_transfer.settled"
```
davidcornu, line 14: **nit**: A `case` would make this a bit easier to read

app/controllers/column/webhooks_controller.rb
```diff
@@ -50,6 +56,33 @@ def handle_ach_incoming_transfer_scheduled
       # at this point, the ACH is approved!
     end
 
+    def handle_ach_incoming_transfer_settled
```
davidcornu: **suggestion**: I'm not actually sure we're getting value out of the duplication between `handle_ach_incoming_transfer_settled`, `handle_wire_incoming_transfer_completed`, and `handle_swift_incoming_transfer_completed` here

1. It's not obvious what's actually changing vs. what _should_ be the same
2. We have to assume `column_event_type` is faithfully passed along

Might I suggest something along the lines of

```ruby
def handle_as_raw_pending_column_transaction(column_event_type:)
  RawPendingColumnTransaction.create!(
    # 💁‍♂️ note the use of `.fetch` here as we're assuming 
    # `@object` always has these properties
    amount_cents: @object.fetch("available_amount"), 
    date_posted: @object.fetch("effective_at_utc"),
    column_transaction: @object,
    column_event_type:,
  )
end
```

which would let the code in the `webhook` method do something like

```ruby
elsif type == "wire.incoming_transfer.completed"
        # there can be no doubt that this matches             👇
        handle_as_raw_pending_column_transaction(column_type: type)
```

or even better

```ruby
case type
# ...
when "ach.incoming_transfer.settled", "wire.incoming_transfer.completed", "swift.incoming_transfer.completed"
  handle_as_raw_pending_column_transaction(column_type: type)
```

app/models/raw_column_transaction.rb
```diff
@@ -20,11 +20,18 @@ class RawColumnTransaction < ApplicationRecord
...
 
   def canonize
-    create_canonical_transaction!(
+    ct = create_canonical_transaction!(
       amount_cents:,
       memo:,
       date: date_posted,
     )
+
+    raw_pending_column_transaction = RawPendingColumnTransaction.find_by(column_id: column_transaction["transaction_id"])
```
davidcornu, line 29: - **question**: can there ever be a `column_transaction` without `transaction_id`?
  ```suggestion
      raw_pending_column_transaction = RawPendingColumnTransaction.find_by(column_id: column_transaction.fetch("transaction_id"))
  ```
- **question**: can we ever fail to find a matching record here?
  ```suggestion
      raw_pending_column_transaction = RawPendingColumnTransaction.find_by!(column_id: column_transaction["transaction_id"])
  ```

db/migrate/20250625003228_create_raw_pending_column_transactions.rb
```diff
@@ -0,0 +1,13 @@
+class CreateRawPendingColumnTransactions < ActiveRecord::Migration[7.2]
+  def change
+    create_table :raw_pending_column_transactions do |t|
+      t.string :column_id
+      t.integer :column_event_type
+      t.jsonb :column_transaction
+      t.text :description
+      t.date :date_posted
+      t.integer :amount_cents
```
davidcornu: - **question**: should we expect all of these columns to potentially have `NULL` values?
- **question**: should `column_id` be unique?

db/migrate/20250625004725_add_raw_pending_column_transaction_id_to_canonical_pending_transaction.rb
```diff
@@ -0,0 +1,7 @@
+class AddRawPendingColumnTransactionIdToCanonicalPendingTransaction < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :canonical_pending_transactions, :raw_pending_column_transaction, index: { algorithm: :concurrently, name: "index_canonical_pending_txs_on_rpct_id" }
```
davidcornu, line 5: - **question**: Should we add a foreign key?
- **question**: Are we expecting this to be `NULL` sometimes?
- **question**: Should this be a unique index?

app/models/raw_pending_column_transaction.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  enum :column_event_type, {
+    "swift.incoming_transfer.completed": 0,
+    "wire.incoming_transfer.completed": 1,
+    "ach.incoming_transfer.settled": 2
+  }
+
+  def create_canonical_pending_transaction
+    column_account_number = Column::AccountNumber.find_by(column_id: column_transaction["account_number_id"])
+    return unless column_account_number
```
davidcornu, line 34: **question**: When might we run into this scenario? Should we fail louder?

app/models/raw_pending_column_transaction.rb
```diff
@@ -0,0 +1,49 @@
...
+    )
+  end
+
+  def memo
+    if column_id.start_with? "acht"
+      return "#{column_transaction["company_name"]} #{column_transaction["company_entry_description"]}"
+    elsif column_id.start_with?("wire") || column_id.start_with?("swft_")
+      return column_transaction["originator_name"]
+    end
+  end
```
davidcornu: **concern**: This is very mysterious.
1. Where do these rules come from? 
    1. Is there some documentation we can link to? 
    2. Do we have similar code elsewhere in the code base?
2. Can these fields not exist? - if not can we switch to `.fetch` so we can catch bad cases?

app/services/transaction_grouping_engine/calculate/hcb_code.rb
```diff
@@ -265,7 +265,7 @@ def unknown_hcb_code
         [
           HCB_CODE,
           UNKNOWN_CODE,
-          @ct_or_cp.id
+          @ct_or_cp.column_transaction_id || @ct_or_cp.id
```
davidcornu, line 268: **question**: What is the impact of this?

app/models/raw_pending_column_transaction.rb
sampoder: This is common and we shouldn't fail for this - this is any time a transaction hits our main account number and can't be matched to an event. Examples include internal book transfers and all Stripe transactions.

db/migrate/20250625004725_add_raw_pending_column_transaction_id_to_canonical_pending_transaction.rb
sampoder: This will often be `NULL`, in the overwhelming amount of cases this will be `null`. Can be made unique. No opinion on making it a foreign key.

db/migrate/20250625003228_create_raw_pending_column_transactions.rb
sampoder: None should be null, good idea re making `column_id` unique and I'll enforce no null on any of these

app/controllers/column/webhooks_controller.rb
sampoder: I agree but I think that's out of scope for this PR. There's differing opinions on case & switches on the team lol - I love them, others not so much.

app/models/raw_column_transaction.rb
sampoder: `transaction_id` should never be null. 

Yes we can fail to find a matching record - I'll add that handling

app/services/transaction_grouping_engine/calculate/hcb_code.rb
sampoder: To the best of my knowledge nothing - we seem to have no logic built using the `hcb_i2` field for `UNKNOWN_CODE`s.

app/controllers/column/webhooks_controller.rb
sampoder: This is clean - I like it! I'll switch to this

app/models/raw_pending_column_transaction.rb
sampoder: These come from `RawColumnTransaction` - these memos are cosmetic. Company name or originator name should always be set but `company_entry_description` is optional

PR #10705, opened 2025-06-25
[Receipt Bin] Fix `per` param

garyhtou:
## Summary of the problem

We use the `per` param to determine the layout of the page (whether transactions are rendered in groups, or a single list). Unfortunately, it 500s at the moment because we're comparing an integer to a string (`params[:per` is a string).

## Describe your changes

This fixes it by casting `per` to an int.


garyhtou:
cc @YodaLightsabr

PR #10710, opened 2025-06-25
Fix empty admin tool on receipt details

### LINKED ISSUE #10699: [Receipts] Empty admin tool on view receipt modal
Image

Luke-Oldenburg:
Closes #10699

### DIFF
--- a/app/views/receipts/_details.html.erb
+++ b/app/views/receipts/_details.html.erb
@@ -21,6 +21,6 @@
   </div>
 </section>
 
-<% admin_tool("mt3 p0") do %>
+<% admin_tool_if(receipt.textual_content.present?, "mt3 p0") do %>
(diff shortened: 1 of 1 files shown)


app/views/receipts/_details.html.erb
```diff
@@ -21,6 +21,6 @@
   </div>
 </section>
 
-<% admin_tool("mt3 p0") do %>
+<% admin_tool_if(receipt.textual_content.present?, "mt3 p0") do %>
   <pre class="m0 p1 bg-transparent"><code><%= receipt.textual_content %></code></pre>
 <% end %>
```
garyhtou: Although this gives us the results we want, I'd suggest wrapping the admin tool with a simple `if`. It's a bit easier to understand and doesn't run the risk of accidentally exposing admin data to users.

admin_tool_if will show its contents to everyone if the first argument is false.

PR #10713, opened 2025-06-25
Move raffle to modal

Luke-Oldenburg:


app/views/raffles/_raffle_form.html.erb
```diff
@@ -0,0 +1,14 @@
+<%# locals: (program:) %>
+
+<%= form_with(url: raffles_path, html: { onsubmit: "location.reload();", action: "" }) do |f| %>
```
davidcornu, line 3: **question**: The `onsubmit: "location.reload();"` is odd. Why do we need it?

app/views/raffles/new.html.erb
```diff
@@ -1,20 +1,10 @@
...
-        Program
-        <%= f.text_field :program %>
-      <% end %>
-    <% end %>
+<% title "Join the raffle" %>
+<% page_sm %>
 
-    <%= f.submit "Join", class: "btn mt-4" %>
-  <% end %>
+<div class="mx-auto max-w-[500px]">
```
davidcornu: **suggestion** (non-blocking): I generally recommend using relative units for things like this so that the width is proportional to the font size (which many users adjust for accessibility reasons). We have `max-w-xl` which seems close.

app/views/static_pages/index/_robotics_raffle.html.erb
```diff
@@ -23,7 +25,7 @@
           <% end %>
         </p>
         <div class="flex gap-2">
-          <img src="https://hc-cdn.hel1.your-objectstorage.com/s/v3/4f997b30150882281aecb29595abda0dddf40409_image.png" height="96" width="96" class="rounded align-bottom object-cover tooltipped tooltipped--w" alt="$100 Amazon Gift Card" aria-label="$100 Amazon Gift Card">
+          <img src="https://hc-cdn.hel1.your-objectstorage.com/s/v3/4f997b30150882281aecb29595abda0dddf40409_image.png" height="96" width="96" class="rounded align-bottom object-cover tooltipped tooltipped--w" alt="Flipper Zero" aria-label="Flipper Zero">
```
davidcornu: **suggestion**: This is a chonky image that we're only using 96x96 of. A crop to 192x192 (for 2x displays) + a run through ImageOptim cuts this down to 11kb (https://hc-cdn.hel1.your-objectstorage.com/s/v3/8f67aaf1a78939f2c3cd623bd33a3ecb1441e658_flipper_zero_banner.png)

app/views/static_pages/index/_robotics_raffle.html.erb
```diff
@@ -1,8 +1,10 @@
+<% raffle_program = "robotics-flipper-zero-raffle-june-2025" %>
+<%= render "raffles/raffle_modal", id: "join_robotics_raffle", program: raffle_program %>
+
 <% if current_user.teenager? && Referral::Program.find_by(id: 3)&.users&.where(id: current_user.id)&.any? && Date.current.month == 6 %>
-  <% raffle_program = "robotics-flipper-zero-raffle-june-2025" %>
   <% already_joined = Raffle.where(user: current_user, program: raffle_program).any? %>
 
-  <%= link_to already_joined ? "javascript:void(0);" : new_raffle_path(program: raffle_program), class: "w-full no-underline text-inherit #{"cursor-default" if already_joined}" do %>
+  <%= link_to already_joined ? "javascript:void(0);" : new_raffle_path(program: raffle_program), class: "w-full no-underline text-inherit #{"cursor-default" if already_joined}", data: { behavior: already_joined ? "" : "modal_trigger", modal: "join_robotics_raffle" } do %>
```
davidcornu: **suggestion**: Given that we're effectively making this `<a>` a non-`<a>` in certain scenarios I'd consider using `capture` ([docs](https://api.rubyonrails.org/classes/ActionView/Helpers/CaptureHelper.html#method-i-capture)) to assign the inner content to a variable and conditionally render the outer element as either an `<a>` or something else.

app/views/raffles/_raffle_form.html.erb
Luke-Oldenburg: Refresh the page after submitting the form so it closes the modal and update the raffle section.

app/views/raffles/new.html.erb
Luke-Oldenburg: I'm honestly not too sure on the styling because this code was patched together from a few locations.

app/views/raffles/_raffle_form.html.erb
davidcornu: Is this because of the `turbo-frame`?

Luke-Oldenburg: I'm not sure but this was the solution Gary and I landed on. I think it's okay to trigger a full page reload. Other forms do it but for some reason this one needed the additional `onsubmit`.

davidcornu (approved):

PR #10714, opened 2025-06-25
[Admin] Small UI tweaks

Luke-Oldenburg:
Fixes some capitalization and margin inconsistencies

### DIFF
--- a/app/views/comments/_form.html.erb
+++ b/app/views/comments/_form.html.erb
@@ -44,7 +44,7 @@
     <%= form.hidden_field :commentable_type, value: commentable&.class %>
     <%= form.hidden_field :commentable_id, value: commentable&.id %>
 
-    <% admin_tool("inline-flex", "span") do %>
+    <% admin_tool_if(!in_admin_tool, "inline-flex span") do %>
       <% if defined?(admin_only) %>
         <%= form.hidden_field :admin_only, value: true %>
         This comment will only be shown to admins.
--- a/app/views/comments/_form.html.erb
+++ b/app/views/comments/_form.html.erb
@@ -44,7 +44,7 @@
     <%= form.hidden_field :commentable_type, value: commentable&.class %>
     <%= form.hidden_field :commentable_id, value: commentable&.id %>
 
-    <% admin_tool_if(!in_admin_tool, "inline-flex span") do %>
+    <% admin_tool("inline-flex", "span") do %>
       <% if defined?(admin_only) %>
         <%= form.hidden_field :admin_only, value: true %>
         This comment will only be shown to admins.
--- a/app/views/events/settings/_admin.html.erb
+++ b/app/views/events/settings/_admin.html.erb
@@ -139,9 +139,11 @@
     <% end %>
   <% end %>
 
-  <% admin_tool if !Rails.env.development? && @event.airtable_record.present? do %>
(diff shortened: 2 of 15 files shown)


YodaLightsabr (approved):
Thank you! I don't even need to see the UI—I know exactly where all of these are and have just gotten used to the visual issues over the years. I'm not sure why Brakeman is failing though.

YodaLightsabr (dismissed):
Ah, you'll need to update [this file](https://github.com/hackclub/hcb/blob/main/config/brakeman.ignore#L118-L151) before merging.

https://github.com/hackclub/hcb/blob/main/config/brakeman.ignore#L118-L151

app/views/comments/_form.html.erb
```diff
@@ -44,7 +44,7 @@
     <%= form.hidden_field :commentable_type, value: commentable&.class %>
     <%= form.hidden_field :commentable_id, value: commentable&.id %>
 
-    <% admin_tool("inline-flex", "span") do %>
+    <% admin_tool_if(!in_admin_tool, "inline-flex span") do %>
```
sampoder: i'd advocate for keeping this as it was - this feels like unneeded complexity. We should also make sure this defaults to false.

app/views/stripe_cards/show.html.erb
```diff
@@ -3,8 +3,8 @@
 
 <%= turbo_frame_tag "stripe_card_#{@card.public_id}" do %>
   <% admin_tool("mt3") do %>
-    <%= link_to "View card on Stripe", @card.stripe_dashboard_url, class: "btn bg-accent" %>
-    <%= link_to "View cardholder on Stripe", @card.cardholder.stripe_dashboard_url, class: "btn bg-accent" %>
+    <%= link_to "View card on Stripe", @card.stripe_dashboard_url&.start_with?("https://dashboard.stripe.com"), class: "btn bg-accent m1" %>
+    <%= link_to "View cardholder on Stripe", @card.cardholder.stripe_dashboard_url&.start_with?("https://dashboard.stripe.com"), class: "btn bg-accent m1" %>
```
sampoder: Screenshot 2025-06-25 at 7 00 52 PM

is this intended?

Luke-Oldenburg: I may or may not have been vibecoding. I'll fix tomorrow.

app/views/comments/_form.html.erb
Luke-Oldenburg: That's fair I just thought the nested AT looked weird.

app/views/events/settings/_admin.html.erb
```diff
@@ -139,10 +139,10 @@
     <% end %>
   <% end %>
 
-  <% admin_tool do %>
+  <% admin_tool if !Rails.env.development? && @event.airtable_record.present? do %>
```
davidcornu: **suggestion**: This is a bit funky given the block afterwards (I wasn't even aware this was legal syntax 😅). Might I suggest something a bit more vanilla

```erb
<% if !Rails.env.development? && @event.airtable_record.present? %>
  <% admin_tool do %>
    <!-- stuff -->
  <% end %>
<% end %>
```

davidcornu:
Other than https://github.com/hackclub/hcb/pull/10714#discussion_r2169576830 this all looks good to me. Will approve & merge once addressed.

davidcornu (approved):

PR #10715, opened 2025-06-25
Don't cancel card grants when removing organizer

### LINKED ISSUE #10696: [Card Grants] Removing a user from an event cancels grants issued to them
This is probably because the stripe cards are getting canceled and then that calls card grant cancel. We should only be canceling stripe cards that don't have a card grant associated. Maybe we should consider freezing them though?

Luke-Oldenburg:
Closes #10696

### DIFF
--- a/app/controllers/organizer_positions_controller.rb
+++ b/app/controllers/organizer_positions_controller.rb
@@ -16,7 +16,7 @@ class OrganizerPositionsController < ApplicationController
     # also cancel all stripe cards from the organizer
     cards = @organizer_position.user.stripe_cards.where(event: @organizer_position.event)
     cards.each do |card|
-      card.cancel! unless card.stripe_status == "canceled"
+      card.cancel! unless card.stripe_status == "canceled" || card.card_grant.present?
     end
     # ...and auto-close all deletion requests
     @organizer_position.organizer_position_deletion_requests.under_review.each { |opdt| opdt.close(current_user) }


app/controllers/organizer_positions_controller.rb
```diff
@@ -16,7 +16,7 @@ def destroy
     # also cancel all stripe cards from the organizer
     cards = @organizer_position.user.stripe_cards.where(event: @organizer_position.event)
     cards.each do |card|
-      card.cancel! unless card.stripe_status == "canceled"
+      card.cancel! unless card.stripe_status == "canceled" || card.card_grant.present?
```
davidcornu, line 19: **question**: Is there a test for this endpoint? Might be useful to make sure this behaviour continues to work

sampoder (approved):
lgtm! though a note for the future: i feel like this should be in the model

Luke-Oldenburg:
Should it? We cancel cards from other places where this logic isn't applicable.

sampoder:
Sorry I meant that it should belong in the OrganizerPosition or OPDR model (all of this functionality)

Luke-Oldenburg:
Ah I can do that before merging.

sampoder:
Oh no need - just a note to self really

app/controllers/organizer_positions_controller.rb
garyhtou: There's likely not a test for this endpoint. It would be good to have, though.

PR #10720, opened 2025-06-25
Switch to `7-2-stable` branch of Rails

davidcornu:
## Summary of the problem

We are running into a Rails issue in production (https://github.com/rails/rails/issues/51780, https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/1136/samples/6596247683eb67648f30f807-897217264893116276917507606401) for which there are as-yet-unreleased mitigations in `7-2-stable` branch (https://github.com/rails/rails/commit/7fef4ae61ea80786b6f02f2f96b85d3eb72b510c).

We could wait for a `7.2.3` release but given that `7.2.2.1` dates back to December it could be awhile (and the 7.2.x series is nearing the end of its maintenance window: https://rubyonrails.org/maintenance).

## Describe your changes

Switches the `rails` dependency from Rubygems to Git on the `7-2-stable` branch (https://github.com/rails/rails/compare/v7.2.2.1...7-2-stable).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ ruby File.read(File.join(File.dirname(__FILE__), ".ruby-version")).strip
 
 gem "dotenv-rails", groups: [:development, :test]
 
-gem "rails", "~> 7.2"
+gem "rails", git: "https://github.com/rails/rails.git", branch: "7-2-stable"
 
 gem "puma", "~> 6.6" # app server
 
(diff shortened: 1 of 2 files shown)


sampoder (approved):
thank you!

garyhtou:
On that note, we should bring https://github.com/hackclub/hcb/issues/8470 to the front burners 🔥

PR #10721, opened 2025-06-25
Un-revert "Expire impersonated sessions after 1 hour"

davidcornu:
Reverts hackclub/hcb#10687 which reverted https://github.com/hackclub/hcb/pull/10667.

We hit several exceptions when the original PR was deployed, but upon further investigation I'm fairly certain we were running into a Rails issue. See https://github.com/hackclub/hcb/pull/10720 for details.


sampoder:
Let's do it!

garyhtou:
Amazing! I'd glad we found the root cause— and that it's unrelated haha

PR #10722, opened 2025-06-25
Introduce `User.system_user` for consistency

davidcornu:
## Summary of the problem

We use the `bank@hackclub.com` user in many places throughout the code base but have different ways of accessing the record, some of which would raise if it didn't exist, others which would return `nil`.

## Describe your changes

- Introduces `User.system_user` and `User::SYSTEM_USER_EMAIL`
- Replaces references to `bank@hackclub.com` with the above


sampoder (approved):
Thank you! Appreciate this.

garyhtou (commented):
Really nice!

PR #10726, opened 2025-06-25
Handle non-event mileage rates

### LINKED ISSUE #10679: [Reimbursement] Default mileage rate to standard rate when no Event is associated
## Problem

@zachlatta encountered an issue where his reimbursement report 500'ed because he had a mileage expense without the report being linked to an Event.

Without the Event association, the report can't determine the proper mileage rate to use (HQ has a lower mileage rate than fiscally sponsored orgs).

## Solution

To provide the smoothest experience for the majority of users, we'll default to the standard mileage rate when no Event is associated. Once an Event is associated, it'll use that Event's mileage reimbursement rate.

The only downside of this solution is that the mileage rate may change/decrease when it's later assigned to an HQ org. For example, Zach starts an Event-less reimbursement report. He adds a mileage expense of 500 miles at 70 cents (the standard rate). He later associates the report with HQ, which updates the mileage rate to 35 cents. This change could confuse users, but it outweighs the other solution of disallowing mileage reimbursements for Event-less reimbursements.

sampoder:
Closes https://github.com/hackclub/hcb/issues/10679


app/models/reimbursement/expense/mileage.rb
```diff
@@ -34,7 +34,7 @@ module Reimbursement
   class Expense
     class Mileage < ::Reimbursement::Expense
       def rate
-        event.plan.mileage_rate(created_at || Date.today) # rate may be used on creation, however, created_at isn't set at that point.
+        (event&.plan || Event::Plan::Standard).mileage_rate(created_at || Date.today) # rate may be used on creation, however, created_at isn't set at that point.
```
garyhtou: I think `mileage_rate` is defined on the instance, and not the class. Other than that, this looks good. Thanks!

```suggestion
        (event&.plan || Event::Plan::Standard.new).mileage_rate(created_at || Date.today) # rate may be used on creation, however, created_at isn't set at that point.
```

PR #10733, opened 2025-06-26
Remove headway script and change links

polypixeldev:
## Summary of the problem

First step of migrating from Headway to the new HCB Blog

## Describe your changes

Removes the Headway widget scripts and changes all links to headway posts to the post on blog.hcb.hackclub.com

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -53,7 +53,7 @@
     *@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@*
 
           Come hack on HCB with us!
-  <%= Rails.configuration.constants.github_url.to_json %>
+  <%= Rails.configuration.constants.github_url %> <%# erb_lint:disable ErbSafety %>
   `, 'color: #ec3750; font-weight: bold; font-size: 10px;')
       </script>
     <% end %>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -53,7 +53,7 @@
     *@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@*
 
           Come hack on HCB with us!
-  <%= Rails.configuration.constants.github_url %>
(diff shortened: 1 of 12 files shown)


app/views/layouts/application.html.erb
```diff
@@ -36,7 +36,6 @@
           } catch {}
         })
       </script>
-      <script async src="https://cdn.headwayapp.co/widget.js"></script>
```
davidcornu, line 39: Is this going to break this sparkles link?

CleanShot 2025-06-26 at 12 52 36@2x

davidcornu: We can probably drop this code as well https://github.com/hackclub/hcb/blob/d293e57763729f69a1612606e22095761650eccd/app/views/layouts/application.html.erb#L28-L38

(and the equivalent code on the docs layout)

polypixeldev: Good catch, thanks David! This is getting rid of the widget that currently opens if you click the sparkles button, but it will still link to blog.hcb.hackclub.com. I'm working on another PR that implements a similar widget in our own code that pulls from blog.hcb.hackclub.com

garyhtou (approved):

app/views/layouts/application.html.erb
```diff
@@ -68,7 +53,7 @@
     *@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@*
 
           Come hack on HCB with us!
-  <%= Rails.configuration.constants.github_url %>
+  <%= Rails.configuration.constants.github_url %> <%# erb_lint:disable ErbSafety %>
```
garyhtou, line 56: ERB lint complained of 
```
erb interpolation in javascript tag must call '(...).to_json' (not autocorrected)
```

@polypixeldev and i spent some time trying to find a solution. `to_json` wouldn't work for us since it warps the URL in quotes. we decided to disable the check, but if anyone has a better solution, definitely let us know.

PR #10747, opened 2025-06-26
[Admin] Fix event audit log order

Luke-Oldenburg:
Audit log was backwards before.

### DIFF
--- a/app/views/events/audit_log.html.erb
+++ b/app/views/events/audit_log.html.erb
@@ -1,6 +1,6 @@
 <%= turbo_frame_tag :audit_log do %>
   <ul class="list-reset">
-    <% @event.versions.select { |version| version.changeset.any? { |field, changes| changes.any?(&:present?) } }.map do |version| %>
+    <% @event.versions.reverse.select { |version| version.changeset.any? { |field, changes| changes.any?(&:present?) } }.map do |version| %>
       <li>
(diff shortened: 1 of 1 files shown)


app/views/events/audit_log.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= turbo_frame_tag :audit_log do %>
   <ul class="list-reset">
-    <% @event.versions.select { |version| version.changeset.any? { |field, changes| changes.any?(&:present?) } }.map do |version| %>
+    <% @event.versions.reverse.select { |version| version.changeset.any? { |field, changes| changes.any?(&:present?) } }.map do |version| %>
```
davidcornu, line 3: **note**: Double-checked to see what order was being applied by default

CleanShot 2025-06-26 at 17 07 44@2x

which is coming from 
- https://github.com/paper-trail-gem/paper_trail/blob/cae7f5d1a6730b04b3b213f9c1453fd1ad6119d2/lib/paper_trail/model_config.rb#L182-L185 
- https://github.com/paper-trail-gem/paper_trail/blob/cae7f5d1a6730b04b3b213f9c1453fd1ad6119d2/lib/paper_trail/model_config.rb#L218-L219 
- https://github.com/paper-trail-gem/paper_trail/blob/cae7f5d1a6730b04b3b213f9c1453fd1ad6119d2/lib/paper_trail/version_concern.rb#L61-L65

davidcornu: Given that we're loading all of `.versions` I think `@event.versions.reverse.select` is fine but we could push the reversal into the database with

```suggestion
    <% @event.versions.reorder(PaperTrail::Version.timestamp_sort_order("desc")).select { |version| version.changeset.any? { |field, changes| changes.any?(&:present?) } }.map do |version| %>
```

davidcornu (approved):

PR #10751, opened 2025-06-27
[User Session] Clear metadata on older sessions

### LINKED ISSUE #10744: [Login] Periodically clear out old `UserSession` entries
[Internal discussion](https://hackclub.slack.com/archives/C047Y01MHJQ/p1750259883680629)

For auditing purposes and in order to help users identify their active sessions, we keep quite a bit of metadata in the `user_sessions` table: https://github.com/hackclub/hcb/blob/320b029640bbac76f34cd72ac881a6d1eb803f98/db/schema.rb#L2132-L2138

At the moment there's no mechanism to clear this data out and its usefulness to us decreases over time. We should pick a threshold and automatically get rid of anything older than that.

manuthecoder:
### Closes #10744

This pull request introduces a new background job to clear outdated user session data and schedules it to run daily. The job ensures sensitive session information is removed for sessions older than one year. Below are the most important changes:

### Addition of a new job:

* [`app/jobs/user/clear_old_user_sessions_job.rb`](diffhunk://#diff-19bd9b2978ef88bf9ae1db00fbb346141d92e5d85c4cb816b8949d814f0c4792R1-R21): Added a new job `ClearOldUserSessionsJob` to the `User` namespace. This job iterates through user sessions older than one year and clears sensitive data such as `device_info`, `os_info`, `timezone`, `ip`, and location details (`latitude` and `longitude`).

### Scheduling the job:

* [`config/schedule.yml`](diffhunk://#diff-18c4a92c666266e8bead5b6ddb034501022877cdf773c0d337499ed0a7f6d341R175-R179): Scheduled the new `ClearOldUserSessionsJob` to run daily at 3:00 AM with a low priority queue.

[Internal discussion](https://hackclub.slack.com/archives/C047Y01MHJQ/p1750259883680629)


app/jobs/user/clear_old_user_sessions_job.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class User
```
davidcornu: **nitpick**: This could live under `UserSession` seen as that's the model it's operating on

app/jobs/user/clear_old_user_sessions_job.rb
```diff
@@ -0,0 +1,21 @@
...
+      UserSession.where("created_at < ?", 1.year.ago).find_each do |session|
+        session.update_columns(
+          device_info: nil,
+          os_info: nil,
+          timezone: nil,
+          ip: nil,
+          impersonated_by_id: nil,
+          latitude: nil,
+          longitude: nil
+        )
```
davidcornu: **suggestion**: This feels like the kind of logic that should live within the `UserSession` model (e.g. in a method like `clear_metadata!`) so that people looking at that code are aware that this is a possibility.

app/jobs/user/clear_old_user_sessions_job.rb
```diff
@@ -0,0 +1,21 @@
...
+    queue_as :low
+    
+    def perform
+      UserSession.where("created_at < ?", 1.year.ago).find_each do |session|
+        session.update_columns(
+          device_info: nil,
+          os_info: nil,
+          timezone: nil,
+          ip: nil,
+          impersonated_by_id: nil,
```
davidcornu: **suggestion**: We should definitely keep this if we are opting to clear out sensitive data instead of deleting the row as it changes the meaning of the `UserSession`.

manu-gurudath:
done! let me know if any further changes are needed, ty!

app/jobs/user_session/clear_old_user_sessions_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class UserSession
+  class ClearOldUserSessionsJob < ApplicationJob
+    queue_as :low
+
+    def perform
+      UserSession.where("created_at < ?", 1.year.ago).find_each(&:clear_metadata!)
```
garyhtou:
```suggestion
      UserSession.expired.where("created_at < ?", 1.year.ago).find_each(&:clear_metadata!)
```

Checking for `created_at < 1.year.ago` is safe, however, I think it should be safer to also check that it's expired.

app/models/user_session.rb
```diff
@@ -81,6 +81,14 @@ def expired?
     expiration_at <= Time.now
   end
 
+  def clear_metadata!
+    update_columns(
+      device_info: nil,
+      latitude: nil,
+      longitude: nil,
+    )
+  end
```
garyhtou: Do we want to use `update` to `update!` so that callbacks/validations run? Any thoughts @davidcornu?

davidcornu: > Do we want to use `update` to `update!` so that callbacks/validations run?

Inclined to say yes unless we have a good reason not to.

garyhtou: Yeah, i'm on the same page.

_typo in first message: to -> or*_

app/models/user_session.rb
```diff
@@ -81,6 +81,14 @@ def expired?
     expiration_at <= Time.now
   end
 
+  def clear_metadata!
+    update_columns(
```
garyhtou:
```suggestion
    update!(
```

PR #10761, opened 2025-06-29
Added v4 receipt bin

Mohamad-Mortada:
## Summary of the problem
I need to query all receipts where receiptable is nil for the user to add these to hcb mobile receipt bin. 

## Describe your changes
Added a receipt bin method to v4 receipt controller and a custom route.


garyhtou (approved):
I'm happy with where the route is located. As our UI changes, it will be difficult to match our API routes with the UI. I think we should start with the clean-ish slate for the API that's logical when it's standalone.

PR #10766, opened 2025-06-30
Update `poppler` and `dry-validation` to address warnings, add `pstore` to `Gemfile`

davidcornu:
## Summary of the problem

When running any command we get a few warnings which can addressed with some dependency housekeeping.

CleanShot 2025-06-30 at 10 25 10@2x

## Describe your changes

- Updates the `poppler` gem (along with its sub-dependencies) to address the `gobject-introspection` warning
- Updates the `dry-validation` gem to address the `ruby2_keywords` warning
- Adds `pstore` to the `Gemfile` (this used to be part of the standard library but [is being moved to a gem](https://bugs.ruby-lang.org/issues/20309))


garyhtou (approved):
Thank you so much for this!

davidcornu:
The `brakeman` step was failing as it was missing libraries. I'm surprised this worked before. My guess is it pulled the dependency from the bundler cache and never actually tried to load the library.

PR #10769, opened 2025-06-30
Remove grants flipper

Luke-Oldenburg:

### DIFF
--- a/app/views/events/_nav.html.erb
+++ b/app/views/events/_nav.html.erb
@@ -58,7 +58,7 @@
       tooltip: "Deposit checks",
       icon: "cheque",
       selected: local_assigns[:selected] == :deposit_check if policy(@event.check_deposits.build).index? %>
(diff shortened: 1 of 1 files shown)


app/views/events/_nav.html.erb
```diff
@@ -58,7 +58,7 @@
       tooltip: "Deposit checks",
       icon: "cheque",
       selected: local_assigns[:selected] == :deposit_check if policy(@event.check_deposits.build).index? %>
-    <% if policy(@event).transfers? || policy(@event).reimbursements? || policy(@event).card_overview? || Flipper.enabled?(:grants_2023_06_21, @event) %>
+    <% if policy(@event).transfers? || policy(@event).reimbursements? || policy(@event).card_overview? %>
```
davidcornu, line 61: **suggestion** (non-blocking): This is fine considering the scope of the PR but this reads to me like we're missing a policy method.

davidcornu (approved):

PR #10775, opened 2025-06-30
Add non-null constraint to tag emojis

polypixeldev:
## Summary of the problem

Emojis have been backfilled and are now required!

## Describe your changes

Ensures that emojis are in the database by adding a non-null constraint. Also updates the schema file to add the emoji field.


sampoder (approved):
lgtm!

PR #10776, opened 2025-06-30
Reader permission fixes

Luke-Oldenburg:


app/views/events/promotions.html.erb
```diff
@@ -18,7 +20,8 @@
       <%= link_to "Apply",
         @perks_available ? fillout_form("8NZy8a2ui9us") : root_path,
         class: "btn mr1",
-        target: :_blank %>
+        target: :_blank,
+        disabled: %>
```
davidcornu: **concern**: I don't believe `disabled` works on `<a>` tags (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/disabled#overview). We need to render an inert element here as a placeholder instead.

app/views/events/promotions.html.erb
```diff
@@ -32,12 +35,13 @@
     <%= image_tag "https://cloud-fz66hmje0-hack-club-bot.vercel.app/0theeventhelperbutbetter.png", width: 60 %>
     <%= link_to @perks_available ? (@event.country_US? ? "Purchase" : "Ineligible") : "Purchase",
       @perks_available ? fillout_form("hDJYyLd7qNus", { "Event Name" => @event.name, "Bank URL" => event_url(@event), "Submitter Name" => current_user.name, "Submitter Email" => current_user.email }, prefix: "prefill_") : root_path,
-      class: "btn mr1 #{'disabled' unless @perks_available && @event.country_US?}",
-      target: :_blank %>
+      class: "btn mr1",
+      target: :_blank,
+      disabled: !(@perks_available && @event.country_US? && OrganizerPosition.role_at_least?(current_user, @event, :manager)) %>
```
davidcornu: **concern**: This `link_to` is getting out of hand. Can we refactor this to make the conditionals easier to reason about?

app/views/receipts/_form_v3.html.erb
```diff
@@ -62,7 +62,7 @@
                                 "action"           => "change->form#submit"
                               } %>
           <span class="flex items-center justify-center flex-row flex-wrap" style="margin: -8px;">
-            <%= form.label :file, class: "btn m1 #{"bg-success" if local_assigns[:success]} #{"bg-error" if local_assigns[:error]}", id: "upload-receipt-button" do %>
+            <%= form.label :file, class: "btn m1 #{"bg-success" if local_assigns[:success]} #{"bg-error" if local_assigns[:error]} #{"disabled" unless ReceiptablePolicy.new(current_user, receiptable).upload?}", id: "upload-receipt-button" do %>
```
davidcornu: **suggestion**: I think we're hitting a breaking point here. 

```suggestion
            <% additional_classes = [].tap do |array|
                 array << "bg-success" if local_assigns[:success]
                 array << "bg-error" if local_assigns[:error]
                 array << "disabled" unless ReceiptablePolicy.new(current_user, receiptable).upload?
               end.join(", ") %>
            <%= form.label :file, class: "btn m1 #{additional_classes}", id: "upload-receipt-button" do %>
```

app/views/receipts/_form_v3.html.erb
```diff
@@ -80,7 +80,7 @@
               <% end %>
             <% end %>
             <% if defined?(enable_linking) %>
-              <%= link_to my_receipts_upload_path(current_user), class: "btn bg-primary m1", data: defined?(inline_linking) ? nil : { behavior: "modal_trigger", modal: "link_receipt_#{instance}" }, onclick: defined?(inline_linking) ? "event.preventDefault(); $('#link_receipt_#{instance}_select').show(); $('#link_receipt_#{instance}_form').hide();" : "" do %>
+              <%= link_to my_receipts_upload_path(current_user), class: "btn bg-primary m1 #{"disabled" unless ReceiptablePolicy.new(current_user, receiptable).link_modal?}", data: defined?(inline_linking) ? nil : { behavior: "modal_trigger", modal: "link_receipt_#{instance}" }, onclick: defined?(inline_linking) ? "event.preventDefault(); $('#link_receipt_#{instance}_select').show(); $('#link_receipt_#{instance}_form').hide();" : "" do %>
```
davidcornu, line 88: - **concern**: This is also getting hard to understand
- **question**: We seem to be using a `disabled` attribute in some places (which as mentioned previously isn't a valid attribute for `<a>` tags) and a `disabled` class in others. Do any of these have an effect?

app/views/events/promotions.html.erb
sampoder: I think this change is safe to leave out, the form is public

app/views/events/promotions.html.erb
sampoder, line 1: I feel like these are fine? You can modify `@perks_available` to require managers to be able to press the buttons

app/views/canonical_transactions/_canonical_transaction.html.erb
```diff
@@ -83,27 +83,29 @@
...
-                <%= button_to event_tag_path(@event || ct.local_hcb_code.event, tag), class: "menu__action", method: :delete, title: "Delete this tag", form: { "data-turbo" => "true", "data-turbo-confirm" => tag.removal_confirmation_message } do %>
-                  <%= inline_icon "delete", size: 18, style: "margin: 0" %>
-                <% end %>
-              </div>
-            <% end %>
-            <% if (@event || ct.local_hcb_code.event).tags.any? %>
-              <div class="menu__divider tags__divider"></div>
-            <% end %>
-            <%= render partial: "hcb_codes/create_tag", locals: { button: ct.local_hcb_code.hashid } %>
+        <% if OrganizerPosition.role_at_least?(current_user, @event, :member) %>
```
sampoder: Can we use Pundit for this?

app/policies/receiptable_policy.rb
sampoder, line 1: thank you - this is clean!

app/views/hcb_codes/_tags.html.erb
```diff
@@ -16,8 +18,10 @@
                 <%= tag.label %>
                 <%= "✓" if hcb_code.tags.include?(tag) %>
               <% end %>
-              <%= button_to event_tag_path(@event, tag), class: "menu__action", method: :delete, title: "Delete this tag", form: { "data-turbo" => "true", "data-turbo-confirm" => tag.removal_confirmation_message }  do %>
-                <%= inline_icon "delete", size: 16, style: "margin: 0" %>
+              <% if OrganizerPosition.role_at_least?(current_user, @event, :member) %>
```
sampoder: Can we use Pundit? Using Pundit makes fixes like this easier in the future

app/views/receipts/_form_v3.html.erb
sampoder: the `disabled` class greys out the button, makes it unclickable and should be used for links (it's a pure visual change though)

davidcornu (approved):
Save for the ERB issues this looks good to me 👍

PR #10780, opened 2025-06-30
Substantially increase test coverage of `LoginsController`

davidcornu:
## Summary of the problem

There's a lot of logic in `LoginsController` that is currently untested and before refactoring it as part of #10671 I'd like to have some guarantee that it continues to work.

## Describe your changes

Rewrite + expand the `LoginsController` spec (36% &rarr; 68% line coverage)


app/views/application/_logo.html.erb
```diff
@@ -5,12 +5,15 @@
       (user_birthday? ? "Happy birthday! #{['🎂', '🎈', '🎉', '🥳'].sample}" : "You’re in #{Rails.env} mode. #{development_mode_flavor}") :
       "",
     data: { "controller" => "analytics #{"ar" if current_page?(auth_users_path)}", "analytics-event-name-value" => "Icon right-clicked", "ar-usdz-href-value" => "/bank.usdz", "action" => "contextmenu->menu#toggle contextmenu->analytics#track click@document->menu#close keydown@document->menu#keydown", "menu-target" => "toggle" } do %>
-    <%= image_tag user_birthday? ? "logo-party-dark.gif" : by_season("logo-#{Rails.env}-dark.png", fall: "https://cloud-qmskqc293-hack-club-bot.vercel.app/0bank-pumpkin.png"),
+
+    <% logo_env = Rails.env.local? ? "development" : Rails.env %>
```
davidcornu, line 9: This is because we don't have `test` equivalents for these files

sampoder (approved):
wow - thank you for doing this!

PR #10787, opened 2025-07-01
[Event Settings] Fix card grant tab not showing

Luke-Oldenburg:

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -367,8 +367,6 @@ class Event < ApplicationRecord
 
   after_update :generate_stripe_card_designs, if: -> { attachment_changes["stripe_card_logo"].present? && stripe_card_logo.attached? && !Rails.env.test? }
 
-  after_update :create_card_grant_setting
-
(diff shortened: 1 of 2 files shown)


garyhtou (changes requested):
Instead of doing an `after_update` which creates a `CardGrant` for all organizations, let's always show the Card Grant settings tab if the event plan allows for it. On that tab, if the card grant setting doesn't exist, then show an "Enable card grants" button.

garyhtou:
The downside of `after_update` is that we'd have a bunch of unused card grant settings. When we need to change the default settings, we'd then need to determine which are actually used vs unused, communicate breaking changes, and roll out changes to these records.

app/models/event.rb
```diff
@@ -798,4 +800,8 @@ def enforce_transparency_eligibility
     end
   end
 
+  def create_card_grant_setting
```
sampoder, line 801: Should every org have a card grant setting?

PR #10789, opened 2025-07-01
Theme switching fixes

polypixeldev:
## Summary of the problem

https://github.com/hackclub/hcb/pull/10757 introduced a couple bugs with theme switching where the selected theme wouldn't get a checkmark on page load and users were unable to switch on an admin page.

## Describe your changes

Fixed the bugs by falling back to the system theme when there was no theme cookie and adding the full blog widget to the admin header


sampoder (approved):
lgtm!

PR #10790, opened 2025-07-01
Update raffles for July 2025

Luke-Oldenburg:
Screenshot From 2025-07-01 13-20-30


garyhtou (approved):

garyhtou:
We can come back later to remove the extra top margin in the image

Luke-Oldenburg:
I left it so the image would be square. I think it looks fine.

PR #10797, opened 2025-07-01
Allow blazer for auditors

Luke-Oldenburg:


garyhtou (approved):

garyhtou (commented):
It would have also been possible to update AdminConstraint to take in a parameter with the required role, but this implementation is fine!

PR #10801, opened 2025-07-02
Migrate all Airbrake usage to Appsignal

polypixeldev:
## Summary of the problem

There's only room for one error reporter in this codebase...
and Appsignal gives us better reports!

## Describe your changes

Changed all `Airbrake.notify` and `notify_airbrake` statements to use the Rails error reporter, which is handled by Appsignal. Also changed the frontend error reporting to use Appsignal, and updated docs.

Before deploying, make sure to set the `APPSIGNAL_API_KEY` environment variable to the appropriate key from Appsignal  - https://docs.appsignal.com/front-end/installation.html#creating-a-push-api-key


polypixeldev:
Need to figure out if Doppler credentials are available during webpack build

app/javascript/appsignal.js
```diff
@@ -0,0 +1,11 @@
...
+
+import Appsignal from '@appsignal/javascript'
+
+const shouldEnableAppsignal = !!APPSIGNAL_API_KEY
+
+export const appsignal = shouldEnableAppsignal
+  ? new Appsignal({
+      key: APPSIGNAL_API_KEY,
+    })
+  : undefined
```
davidcornu: **suggestion**: Rather than having `appsignal` potentially be `undefined`, I would consider initializing it without `key` so that it just runs in dev mode (https://docs.appsignal.com/front-end/configuration.html). This would save us a bunch of conditional logic downstream.

app/models/event.rb
```diff
@@ -659,7 +659,7 @@ def frozen?
   end
 
   def revenue_fee
-    plan&.revenue_fee || (Airbrake.notify("#{id} is missing a plan!") && 0.07)
+    plan&.revenue_fee || (Rails.error.unexpected("#{id} is missing a plan!") && 0.07)
```
davidcornu: **suggestion**: I know this isn't your code but it's a good opportunity to clean this up a little

```suggestion
    configured = plan&.revenue_fee
    return configured if configured.present?
    
    Rails.error.unexpected("#{id} is missing a plan!")
    
    0.07 # ideally we'd put this in a constant like `Plan::FALLBACK_REVENUE_FEE`
```

app/services/transaction_engine/nightly.rb
```diff
@@ -56,10 +56,9 @@ def import_raw_column_transactions!
       transactions_by_report.each do |report_id, transactions|
         transactions.each_with_index do |transaction, transaction_index|
           if transaction["effective_at"] == transaction["effective_at_utc"] && transaction["effective_at_utc"] < "2024-10-07T04:00:00Z"
-            notice = "Skipping the import of the following transaction in #{report_id}"
+            notice = "Skipping the import of the following transaction in #{report_id}: #{transaction}"
             puts notice
```
davidcornu: **suggestion**: Let's make sure this ends up in the logs
```suggestion
            Rails.logger.warn notice
```

app/javascript/controllers/index.js
```diff
@@ -3,14 +3,13 @@
...
 
 const application = Application.start()
 
-if (airbrake) {
-  application.handleError = error => {
-    airbrake.notify(error)
-  }
+if (appsignal) {
+  installErrorHandler(appsignal, application)
 }
```
davidcornu: **suggestion**: We don't need this `if` anymore now that we no longer conditionally set the export.

```suggestion
installErrorHandler(appsignal, application)
```

davidcornu (approved):
Minor comment - otherwise looks great!

webpack.config.js
```diff
@@ -41,8 +41,7 @@ module.exports = {
     }),
     new webpack.DefinePlugin({
       // prettier-ignore
-      AIRBRAKE_PROJECT_ID: JSON.stringify(process.env.AIRBRAKE_PROJECT_ID || null),
-      AIRBRAKE_API_KEY: JSON.stringify(process.env.AIRBRAKE_API_KEY || null),
+      APPSIGNAL_API_KEY: JSON.stringify(process.env.APPSIGNAL_API_KEY || null)
```
garyhtou: I've added it as `APPSIGNAL_FRONTEND`. Can you rename the code?

I gave it a different name to distinguish between the frontend vs backend push api key

PR #10811, opened 2025-07-02
[Org homepage] Refactor transactions

manuthecoder:
@davidcornu suggested the following change on #10647. This PR changes transactions as well to apply the feedback

> **suggestion**: If I'm reading this correctly, we're basically removing all useful classes and properties from this <a> tag when organizer_signed_in?, which leaves me with two questions:
> 1. Do we need this to be a <a> when there is no href?
> 
>     &rarr; The [capture helper](https://api.rubyonrails.org/classes/ActionView/Helpers/CaptureHelper.html#method-i-capture) might come in handy here so we can wrap the inner content with different outer elements.
> 3. Can we simplify the class attribute logic?
> 
>     &rarr; Would a local variable be more readable?
> 
> ^ putting the above together we end up with something like
> 
> ```erb
> <% @money_in.each do |transaction| %>
>   <% rendered_transaction = capture do %>
>     
>   <% end %>  
> 
>   <% if organizer.signed_in? %>
>     
>     <% name_this_concept = Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) && (@event || transaction.local_hcb_code.event).present? %>
>     <% css_classes = "cursor-pointer hover:bg-gray-50 active:bg-gray-100 transition-colors dark:hover:bg-neutral-700 dark:active:bg-neutral-600" if name_this_concept  %>
>     <a
>       class="homepage-transaction <%= css_classes %>" 
>       href="<%= url_for(transaction.local_hcb_code) %>"
>       data-behavior="modal_trigger"
>       data-modal="money_movement_tx_details_<%= transaction.local_hcb_code.__id__ %>"
>     >
>       <%= rendered_transaction %>
>     </a>
>   <% else %>
>     <div class="homepage-transaction">
>       <%= rendered_transaction %>
>     </div>
>   <% end %>
> <% end %>
> ```


davidcornu (approved):
Thanks for following up!

garyhtou:
nice work here!

PR #10815, opened 2025-07-02
Revert "Add missing raffle annotation and schema index (#10782)"

Luke-Oldenburg:
This reverts commit a9a891ed82b932c0937fd4a76e33da4af5b42623.

## Summary of the problem

This index was intentionally removed previously and shouldn't be added back.


db/schema.rb
```diff
@@ -1496,7 +1496,6 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.index ["program", "user_id"], name: "index_raffles_on_program_and_user_id", unique: true
-    t.index ["user_id"], name: "index_raffles_on_user_id"
```
davidcornu, line 1499: This is odd. I have this index locally and it's in production as well:

CleanShot 2025-07-02 at 15 47 15@2x

If we don't actually need it then we'll need a migration to drop it

PR #10818, opened 2025-07-02
Improve confirmation logic for removing donation goals

manuthecoder:
https://github.com/user-attachments/assets/00f5c7ca-bf27-47cc-a709-4e1fb44e9905

Currently, when a user tries to toggle off donation goals, a confirmation modal appears. However, if the user cancels the modal, the donation goals accordion still collapses, which can be confusing.

This PR fixes that by ensuring the accordion only collapses if the user confirms the action.

### DIFF
--- a/app/views/events/settings/_donations.html.erb
+++ b/app/views/events/settings/_donations.html.erb
@@ -66,7 +66,7 @@
               data: { action: "change->accordion#toggle", target: "accordion.checkbox" },
               disabled:,
               checked: @event.donation_goal.present?,
-              onchange: @event.donation_goal && "if (!this.checked && confirm('Remove donation goal?')) this.closest('form').requestSubmit();",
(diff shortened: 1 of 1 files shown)


app/views/events/settings/_donations.html.erb
```diff
@@ -66,7 +66,7 @@
               data: { action: "change->accordion#toggle", target: "accordion.checkbox" },
               disabled:,
               checked: @event.donation_goal.present?,
-              onchange: @event.donation_goal && "if (!this.checked && confirm('Remove donation goal?')) this.closest('form').requestSubmit();",
+              onchange: @event.donation_goal && "!this.checked && (!confirm('Remove donation goal?') ? (this.checked = true) : this.closest('form').requestSubmit())",
```
davidcornu, line 69: **suggestion** (non-blocking): Might be a nice improvement to move this into a stimulus controller that coordinates with the accordion controller via events (https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events) rather than having to reason about inline JS.

davidcornu (approved):

PR #10819, opened 2025-07-02
[Donation tiers] Improve functionality

manuthecoder:
## Donation Tiers Feature Improvements
This pull request makes donation tiers easier to manage and use. Here's what changed:

- Tier Management
   - **Publish/unpublish tiers** – Organizers can now hide tiers from the donation page without deleting them. All existing tiers are visible by default.
   - **10 tier limit** – Events can't have more than 10 donation tiers. The "Create tier" button gets disabled when you hit the limit.
   - Added a publish checkbox and cleaner controls for managing tiers.
- Code Organization
   - **Reusable setup code** – Created a shared `DonationPageSetup` module that both the donations and tiers controllers use, making the code easier to maintain.
   - **Better routing** – Tier-specific donation pages now use their own controller action (`donation/tiers#start`) instead of sharing logic with the main donations page.
- User Experience
   - **Improved donation page** – The tier selection interface is cleaner and handles custom amounts better.
   - **Loading feedback** – When editing tiers, loading feedback is shown so users know what's happening.
- Small Fixes
   - Removed duplicate flash message rendering from event headers

### Preview 
image


app/models/donation/tier.rb
garyhtou, line 1: You can use
```ruby
validates :donation_tiers, length: { maximum: 10 }
```

db/migrate/20250702221525_add_publish_toggle_to_donation_tier.rb
```diff
@@ -0,0 +1,5 @@
+class AddPublishToggleToDonationTier < ActiveRecord::Migration[7.2]
+  def change
+    add_column :donation_tiers, :published, :boolean, default: false, null: false
```
garyhtou: `Donation::Tier.update_all(published: true)`

app/controllers/donation/tiers_controller.rb
sampoder, line 1: Is there any way of not re-writing this code? We should be able to use the donations controller's start method with some modifications - no?

Luke-Oldenburg:
Could you include in the description what the fixes are please? It isn't initially clear what this PR is for.

app/controllers/donation/tiers_controller.rb
```diff
@@ -2,12 +2,31 @@
...
       @tiers = @event.donation_tiers
     end
 
+    def start
+      return not_found unless @event.donation_page_available?
+
+      authorize @event, :show?
+
+      @tier = @event.donation_tiers.find_by(id: params[:tier_id]) if params[:tier_id]
+      @monthly = true # Donation tiers are always monthly
```
garyhtou: This later gets overridden by `@monthly = params[:monthly].present?` in  `build_donation_page!`

app/controllers/donation/tiers_controller.rb
```diff
@@ -63,7 +82,8 @@ def update
         tier.update(
           name: tier_data[:name],
           description: tier_data[:description],
-          amount_cents: (tier_data[:amount_cents].to_f * 100).to_i
+          amount_cents: (tier_data[:amount_cents].to_f * 100).to_i,
+          published: tier_data[:published] == "1"
```
garyhtou:
```suggestion
          published:  ActiveRecord::Type::Boolean.new.cast(tier_data[:published])
```

app/controllers/donation/tiers_controller.rb
```diff
@@ -84,7 +104,7 @@ def destroy
     private
 
     def set_event
-      @event = Event.where(slug: params[:event_id]).first
+      @event = Event.where(slug: params[:event_name] || params[:event_id]).first
```
garyhtou: Can we check to see why we used our own `set_event` method instead of the one defined in the concern: 
https://github.com/hackclub/hcb/blob/166a8e163380fd3e1de82aebae5fba61300b34d8/app/controllers/concerns/set_event.rb#L5-L4

app/models/donation/tier.rb
```diff
@@ -28,6 +28,7 @@ class Tier < ApplicationRecord
 
     validates :name, :amount_cents, presence: true
     validates :amount_cents, numericality: { only_integer: true, greater_than: 0 }
+    validate { errors.add(:base, "Event can only have 10 donation tiers") if event&.donation_tiers&.size.to_i > 10 }
```
garyhtou:
```suggestion
    validate { errors.add(:base, "Organization can only have 10 donation tiers") if event.donation_tiers.size > 10 }
```

app/controllers/donation/tiers_controller.rb
```diff
@@ -2,12 +2,26 @@
...
     before_action :set_event, except: [:set_index]
 
     def index
       @tiers = @event.donation_tiers
     end
 
+    def start
+      return not_found unless @event.donation_page_available?
+
+      authorize @event, :show?
```
garyhtou: This really should be moved into an `EventPolicy#donation_page?`

app/controllers/donation/tiers_controller.rb
```diff
@@ -2,12 +2,26 @@
...
+
+      return unless build_donation_page!(event: @event, params:, request:)
+
+      @hide_flash = true
+      render "donations/start_donation"
+    end
+
     def set_index
```
garyhtou, line 23: The rest of these (`set_index`, `create`, `update`, and `destroy`) should be in an `Donation::TierPolicy`.

Maybe something like:
```ruby
class Donation
  class TierPolicy < ApplicationPOlicy

    alias :set_index? :can_update_event?
    alias :create? :can_update_event?
    alias :update? :can_update_event?
    alias :destroy? :can_update_event?

    private

    def can_update_event?
      EventPolicy.new(user, record.event).update?
    end
  end
end
```

app/views/events/landing/_header.html.erb
```diff
@@ -59,5 +59,3 @@
 <% end %>
 
 <%= render "application/fullstory" %>
-
-<%= render "application/flash" %>
```
garyhtou: With this change, let's make sure flash is still rendered exactly once per page!

app/models/donation/tier.rb
```diff
@@ -28,11 +28,20 @@ class Tier < ApplicationRecord
 
     validates :name, :amount_cents, presence: true
     validates :amount_cents, numericality: { only_integer: true, greater_than: 0 }
```
garyhtou, line 31: Validate `amount_cents % 100 == 0` since we don't want to allow cents

db/migrate/20250702221525_add_publish_toggle_to_donation_tier.rb
```diff
@@ -0,0 +1,17 @@
...
```
garyhtou:
```suggestion
    add_column :donation_tiers, :published, :boolean, default: true, null: false
    change_column_default :donation_tiers, :published, false
```

and then we can get rid of the reversilbe block

db/migrate/20250709020702_remove_index_event_follows_on_user_id.rb
```diff
@@ -1,5 +1,5 @@
 class RemoveIndexEventFollowsOnUserId < ActiveRecord::Migration[7.2]
   def change
-    remove_index :event_follows, name: "index_event_follows_on_user_id"
+    remove_index :event_follows, column: :user_id, name: "index_event_follows_on_user_id"
```
garyhtou: This seems unrelated. Let's drop it

app/views/wise_transfers/new.html.erb
```diff
@@ -35,6 +35,8 @@
 
   <%= form_errors(@wise_transfer, "wise transfer", "We couldn't send this") %>
 
+  <%= render "wise_transfers/delay_warning" %>
```
garyhtou: This is also unrelated

garyhtou (commented):
It appears there was a merge conflict that wasn't resolved properly

app/models/donation/tier.rb
```diff
@@ -28,11 +28,28 @@ class Tier < ApplicationRecord
...
+      return if event.donation_tiers.where.not(id: id).count < 10
+
+      errors.add(:base, "Organization can only have 10 donation tiers")
+    end
+
+    def amount_divisible_by_100
+      if amount_cents.present? && amount_cents % 100 != 0
+        errors.add(:amount_cents, "must be divisible by 100")
+      end
+    end
```
garyhtou:
```suggestion
    validate :amount_is_whole_dollar

default_scope { order(sort_index: :asc) }

acts_as_paranoid

    private

    def maximum_tiers_limit
      return if event.donation_tiers.where.not(id: id).count < 10

      errors.add(:base, "Organization can only have 10 donation tiers")
    end

    def amount_is_whole_dollar
      if amount_cents.present? && amount_cents % 100 != 0
        errors.add(:amount_cents, "must be whole dollars")
      end
    end
```

garyhtou (approved):

garyhtou:
force merging since reverisible migration check is failing due to this PR being older

Issue #10745, opened 2025-06-26
[StatsD] Disconnect reporting to telemetry.hackclub.com


garyhtou:
We have statsd reporting to telemetry.hackclub.com, but telemetry.hackclub.com got shut down. Let's stop it from uselessly reporting stats.

At some point, we'll want to report those same stats elsewhere. Either AppSignal or our own Grafana replacement.

davidcornu:
https://docs.appsignal.com/standalone-agent/statsd.html might be a good solution so we can keep using StatsD instead of rewriting everything using AppSignal's custom instrumentation (https://www.rubydoc.info/gems/appsignal#custom-instrumentation)

garyhtou:
We should also investigate https://docs.appsignal.com/api/public-endpoint/statsd.html

rluodev:
after discussing with Gary this is paused because telemetry.hackclub.com is coming back!

PR #10822, opened 2025-07-03
[DX] Perform fewer network requests on boot

davidcornu:
## Summary of the problem

When the HCB Rails app boots, we currently perform two requests which can substantially slow down day to day actions like running tests on slow networks (or prevent them altogether if there is no network).

1. We retrieve the list of IP addresses used by Stripe for webhooks
2. We sent StatsD events to `telemetry.hackclub.com`

Neither of these is necessary in development or CI.

## Describe your changes

1. Stops setting `STATSD_ENV` and `STATSD_ADDR` in test and development environments (this results in events being logged to stderr)
2. Moves the Stripe IP allow-list to a file
3. Sets up a github actions workflow to sync the allow-list every day and create a PR if it has changed (you can see a demo of this working here: https://github.com/davidcornu/test-repo/pull/1, https://github.com/davidcornu/test-repo/actions/runs/16052346301)


.github/workflows/stripe_ips.yml
garyhtou, line 1: This is clever!

garyhtou (approved):

config/initializers/statsd.rb
```diff
@@ -1,9 +1,12 @@
...
 Rails.application.configure do
-  # StatsD config here
-  ENV["STATSD_ENV"] = "production" # This won't send data unless set to production
-  ENV["STATSD_ADDR"] = "telemetry.hackclub.com:8125"
+  unless Rails.env.local?
+    # StatsD config here
+    ENV["STATSD_ENV"] = "production" # This won't send data unless set to production
+    ENV["STATSD_ADDR"] = "telemetry.hackclub.com:8125"
+  end
+
```
garyhtou, line 9: FYI, telementry.hackclub.com is no longer functional.

https://github.com/hackclub/hcb/issues/10745

davidcornu: Ah I was wondering where I could see the stats being collected

PR #10826, opened 2025-07-03
Extract core logic from `LoginsController` so it can be tested and reused

davidcornu:
Split out from #10671 

## Summary of the problem

Our login logic is very coupled to `LoginsController` making it harder to test and use elsewhere.

## Describe your changes

- Replaces a number of instance variables (set in `set_available_methods`) with methods on `Login`
- Extracts the bulk of logic in `#complete` into `ProcessLoginService`
- Moves and generalizes the WebAuthn test code in the `LoginsController` spec into `WebAuthnSupport` so it can be used in more places


app/controllers/logins_controller.rb
```diff
@@ -101,47 +100,43 @@ def totp
...
-        sms: params[:sms],
-      ).run
-
-      if params[:sms]
-        @login.update(authenticated_with_sms: true)
-      else
-        @login.update(authenticated_with_email: true)
+      ok = service.process_login_code(
+        code: params[:login_code],
+        sms: ActiveRecord::Type::Boolean.new.cast(params[:sms])
```
sampoder, line 126:
```suggestion
        sms: !!params[:sms]
```
This is mainly a personal preference

davidcornu: Those aren't strictly equivalent though (e.g. `""` is `true` if you use `!!` but `false` if you cast)

app/models/login.rb
sampoder, line 1: changes to this model look good and are cleaner - thank you!

app/views/logins/choose_login_preference.html.erb
sampoder, line 1: frontend changes feels good to me as well (using the `Login` model)

spec/controllers/logins_controller_spec.rb
sampoder, line 1: thank you for the tests!

app/services/process_login_service.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    unless verified
+      errors.add(:base, "Invalid one-time password")
+      return false
+    end
+
+    ActiveRecord::Base.transaction do
+      totp_credential.update!(last_used_at: DateTime.now)
+      login.update!(authenticated_with_totp: true)
+    end
```
sampoder: tiniest nit: but I would prefer using `user.totp` directly instead of `totp_credential` - thoughts?

sampoder (approved):
logic & abstraction looks good to me! some tiny comments

app/services/process_login_service.rb
davidcornu: I have no strong feelings either way so happy to switch

davidcornu:
Note - This is going to conflict with https://github.com/hackclub/hcb/pull/10434. Will see if we can get that merged first and I can update this PR accordingly.

Issue #10755, opened 2025-06-27
[DX] Stub out remote calls in the test suite


davidcornu:
There are several tests in our test suite making real API calls to Stripe and other services. We should consider stubbing these out with a gem like https://rubygems.org/gems/vcr or https://rubygems.org/gems/webmock so they're easier to run.

This might also let us reduce the number of credentials needed (or completely eliminate them) so anyone outside of HCB can run things.

garyhtou:
Yes please!!

PR #10829, opened 2025-07-03
Set up `WebMock` to catch network calls in tests

davidcornu:
See https://github.com/hackclub/hcb/issues/10755


brarnaudovski:
I'm happy to work on this. I see specs need adapting alongside `webmock` installation.

PR #10830, opened 2025-07-03
Organization Announcements (Tier 1)

Luke-Oldenburg:


app/views/announcements/_format_menu.html.erb
```diff
@@ -0,0 +1,58 @@
+<div class="flex items-center gap-2">
+  <div id="format-menus"></div>
+  <div data-controller="menu" data-menu-append-to-value="#format-menus">
+    <button
+      data-action="menu#toggle click@document->menu#close keydown@document->menu#keydown"
+      data-menu-target="toggle"
+      class="border-0 bg-transparent hover:!bg-dark px-2 py-1 rounded-md font-bold flex gap-1"
+      type="button">
+      H <%= inline_icon "down-caret", size: 18 %>
```
Luke-Oldenburg, line 9: What is the H for?

Gemfile
```diff
@@ -228,3 +228,5 @@ gem "rack-timeout", require: "rack/timeout/base"
 gem "irb"
 
 gem "pstore"
+
+gem "prosemirror_to_html"
```
davidcornu, line 232: **note**: This is https://github.com/etaminstudio/prosemirror_to_html. [Prosemirror](https://prosemirror.net/) stores its content as a JSON data structure rather than HTML, so rendering it requires a transformation step.

package.json
```diff
@@ -27,14 +27,22 @@
...
     "@shopify/draggable": "^1.1.4",
     "@stripe/stripe-js": "^7.4.0",
+    "@tiptap/core": "^2.24.0",
+    "@tiptap/extension-code-block-lowlight": "^2.25.0",
+    "@tiptap/extension-image": "^2.25.0",
+    "@tiptap/extension-link": "^2.25.0",
+    "@tiptap/extension-placeholder": "^2.24.0",
+    "@tiptap/extension-underline": "^2.24.0",
+    "@tiptap/pm": "^2.24.0",
+    "@tiptap/starter-kit": "^2.24.0",
```
davidcornu, line 37: **note**: This is https://tiptap.dev/ which builds on top of ProseMirror.

db/schema.rb
```diff
@@ -190,6 +190,19 @@
     t.index ["visit_token"], name: "index_ahoy_visits_on_visit_token", unique: true
   end
 
+  create_table "announcements", force: :cascade do |t|
+    t.string "title"
```
davidcornu: **concern**: Should this be `null: false`? Do we anticipate announcements without titles?

db/schema.rb
```diff
@@ -190,6 +190,19 @@
...
+    t.bigint "user_id"
```
davidcornu: 1. **suggestion**: What is this user? We might want to use a more descriptive column name like `author_id`
2. **concern**: Will all announcements have an associated user? If so we're missing a `null: false`
3. **concern**: I don't see a foreign key for this relation. I'd recommend we add one

db/schema.rb
```diff
@@ -190,6 +190,19 @@
...
+    t.boolean "draft"
```
davidcornu: 1. **concern**: If we're set storing this state as a separate column (see (2)) then I'd recommend we set a default value and `null: false`, otherwise we run the risk of ending up with `NULL` values in the DB which are open to interpretation.
2. **suggestion**: Might I suggest using `published_at is null` as the indication that this announcement is in a draft state?
    
    For example, in our current modelling we have
    published_at / draft | true | false
    ---- | ---- | ----
    present | ? | published
    null | draft | ?

    We get all the states we want out of just `published_at` and remove the invalid ones (which we'd have to enforce in application code).

db/schema.rb
```diff
@@ -190,6 +190,19 @@
...
+    t.datetime "deleted_at", precision: nil
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.text "content"
```
davidcornu: **suggestion**: If this is always going to be JSON we might as well store it as `JSONB` (https://guides.rubyonrails.org/active_record_postgresql.html#json-and-jsonb)

db/schema.rb
```diff
@@ -877,6 +890,16 @@
...
   end
 
+  create_table "event_follows", force: :cascade do |t|
+    t.bigint "user_id", null: false
+    t.bigint "event_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["event_id"], name: "index_event_follows_on_event_id"
+    t.index ["user_id", "event_id"], name: "index_event_follows_on_user_id_and_event_id", unique: true
+    t.index ["user_id"], name: "index_event_follows_on_user_id"
```
davidcornu: **suggestion**: We don't need the `["user_id"]` index because the `["user_id", "event_id"]` one can also be used to filter by `user_id`.

app/controllers/announcements_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def set_announcement
+    if params[:id].present?
+      @announcement = Announcement.find(params[:id])
+    else
+      @announcement = Announcement.new
+    end
+
+    if params[:event_id].present?
+      @event = Event.find_by(slug: params[:event_id])
```
davidcornu: **suggestion**: We can be a bit stricter here

```suggestion
      @event = Event.find_by!(slug: params[:event_id])
```

app/controllers/announcements_controller.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+class AnnouncementsController < ApplicationController
+  before_action :set_announcement
+
+  def index
+    authorize @announcement
+
+    @all_announcements = Announcement.where(event: @event).order(draft: :desc, published_at: :desc, created_at: :desc)
+    @announcements = Kaminari.paginate_array(@all_announcements).page(params[:page]).per(10)
```
davidcornu: **concern**: Any reason we're using `paginate_array` here? I believe that just loads all the records and paginates in-memory which might not be what we want.

app/controllers/announcements_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+    @announcement.publish
+
+    flash[:success] = "Published announcement"
+
+    redirect_to event_announcement_path(@event, @announcement)
+  end
+
+  private
+
+  def set_announcement
```
davidcornu, line 91: **concern**: There's a lot going on in this method and mixing loading from the DB and creating new records might lead to unexpected behaviour (e.g. `params` can also come from query parameters so one could do some shenanigans by passing those params to other endpoints).

I'd recommend 
- Restricting this method to just loading records from the DB and skipping in on actions that need to build a new `Announcement`
  - Those actions can just inline whatever logic they need
- I would consider splitting out a `set_event` method as well so it can be used for `create`
- I would try to scope access to announcements through `@event` by doing things like `@event.announcements.find` rather than `Announcement.find` as the former guarantees we're only ever dealing with announcements for a specific event.

app/controllers/announcements_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+class AnnouncementsController < ApplicationController
+  before_action :set_announcement
+
+  def index
+    authorize @announcement
+
+    @all_announcements = Announcement.where(event: @event).order(draft: :desc, published_at: :desc, created_at: :desc)
+    @announcements = Kaminari.paginate_array(@all_announcements).page(params[:page]).per(10)
+
+    raise ActionController::RoutingError.new("Not Found") if !@event.is_public && @all_announcements.empty? && !organizer_signed_in?
```
davidcornu, line 17: **suggestion**: I'm not sure this is the right place to be doing all of this:

1. `@event.is_public` we might be able to get rid of by only loading public events in `set_announcement` (e.g. `Event.transparent.find`)
2. `organizer_signed_in?` feels like it should be addressed in the pundit policy
3. `@all_announcements.empty?` is curious. Do we not want to render this page at when there are no results?

app/controllers/announcements_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+  def new
+    authorize @announcement
+  end
+
+  def create
+    @announcement = @event.announcements.build(params.require(:announcement).permit(:title, :content, :draft).merge(user: current_user))
+
+    authorize @announcement
+
+    @announcement.save!
```
davidcornu, line 32: **question**: Do we plan on performing any validation at this phase?

davidcornu (commented):
Not quite done but figured I'd publish my first batch of comments

app/controllers/announcements_controller.rb
Luke-Oldenburg:
```
  def card_grant_overview
    card_grants_page = (params[:card_grants_page] || 1).to_i
    card_grants_per_page = (params[:card_grants_per_page] || 20).to_i

    transactions_page = (params[:transactions_page] || 1).to_i
    transactions_per_page = (params[:transactions_per_page] || 20).to_i

    authorize @event

    @card_grants = @event.card_grants.order(created_at: :desc)
    @paginated_card_grants = @card_grants.page(card_grants_page).per(card_grants_per_page)

    @all_stripe_cards = @event.stripe_cards.where.associated(:card_grant)
    @hcb_codes = HcbCode.where(hcb_code: @all_stripe_cards.flat_map(&:all_hcb_codes)).order(created_at: :desc)
    @paginated_hcb_codes = @hcb_codes.page(transactions_page).per(transactions_per_page)
  end
  ```
  This is an example of better pagination I believe.

File: app/controllers/announcements_controller.rb (hunk 13 above)
Comment (Luke-Oldenburg) [reply app/controllers/announcements_controller.rb:32]: I don't think it's necessary in the current state but maybe in future tiers.

File: db/schema.rb (hunk 7 above)
Comment (Luke-Oldenburg) [reply db/schema.rb]: This is being switched to HTML in tier 2 so it is fine to leave for now.

**Comment by Luke-Oldenburg** [comment]
There should also be a follow button on the announcements page. Make sure the org being nontransparent doesn't prevent announcements from showing unless there are none.

File: app/controllers/announcements_controller.rb (hunk 10 above)
Comment (davidcornu) [reply app/controllers/announcements_controller.rb]: Yeah I believe `.page` and `.per` is the kaminari way (although you don't necessarily have to make `.per` user-configurable.

File: app/controllers/event/follows_controller.rb [hunk 14]
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+class Event
+  class FollowsController < ApplicationController
+    include SetEvent
+    skip_before_action :signed_in_user, only: :create
+    before_action :set_event, only: :create
+
+    def create
+      if !signed_in?
```
davidcornu: **nit**: `if !` is what `unless` is for 😬
```suggestion
      unless signed_in?
```

app/controllers/event/follows_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+        return redirect_to auth_users_path(return_to: event_url(@event), require_reload: true)
+      end
+
+      attrs = {
+        user_id: current_user.id,
+        event: @event
+      }
+      follow = Event::Follow.new(attrs)
+      if authorize follow
+        Event::Follow.create(attrs)
```
davidcornu: - **suggestion**: Why throw away a perfectly good `follow`?
- **suggestion**: If we're not checking the return value we should use a `!` method so we'll `raise` if something goes wrong.


```suggestion
        follow.save!
```

- **suggestion**: We might want to rescue `ActiveRecord::RecordNotUnique` in case the user is already following the same event and just continue

app/controllers/event/follows_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+        skip_authorization
+        return redirect_to auth_users_path(return_to: event_url(@event), require_reload: true)
+      end
+
+      attrs = {
+        user_id: current_user.id,
+        event: @event
+      }
+      follow = Event::Follow.new(attrs)
+      if authorize follow
```
davidcornu: **question**: Do we need to check the return value of `authorize`? I was under the impression that it raises an exception if the user doesn't have access.

app/controllers/event/follows_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+      else
+        flash[:error] = "You aren't allowed to follow this event."
+      end
+    end
+
+    def destroy
+      @event_follow = Event::Follow.find(params[:id])
+      if authorize @event_follow
+        slug = @event_follow.event.slug
+        @event_follow.destroy
```
davidcornu: **suggestion**: Ditto here re: avoiding silent failures


```suggestion
        @event_follow.destroy!
```

app/controllers/my_controller.rb
```diff
@@ -141,4 +141,10 @@ def payroll
     @payout_method = current_user.payout_method
   end
 
+  def feed
+    @event_follows = Event::Follow.where(user_id: current_user.id)
```
davidcornu: **suggestion** (minor): Do we have a relation for this? Might be more straightforward as `current_user.event_follows` or something to that effect.

app/controllers/my_controller.rb
```diff
@@ -141,4 +141,10 @@ def payroll
...
+    @all_announcements = Announcement.where(draft: false, event: @event_follows.map { |follow| follow.event }).order(draft: :desc, published_at: :desc, created_at: :desc)
```
davidcornu: 1. **suggestion**: Might be worth defining a `published `scope on `Announcement` so we can do `Announcement.published.where`
2. **nitpick**: `@event_follows.map { |follow| follow.event }` can be shortened to `@event_follows.map(&:event)`
3. **suggestion**: If we're scoping the results to `draft: false` then we don't need to include `draft` in the `.order`. We probably don't also need `created_at` in there (`id` might be an acceptable tie breaker if we're worried about that)

app/controllers/my_controller.rb
```diff
@@ -141,4 +141,10 @@ def payroll
...
+    @announcements = Kaminari.paginate_array(@all_announcements).page(params[:page]).per(10)
```
davidcornu: Same concern as https://github.com/hackclub/hcb/pull/10830/files#r2193154693.

app/helpers/announcements_helper.rb
davidcornu, line 1: **question**: Do we need this?

app/helpers/application_helper.rb
```diff
@@ -392,18 +392,18 @@ def ugc_link_to(name, string, *options)
   def dropdown_button(button_class: "bg-success", template: ->(value) { value }, **options)
```
davidcornu, line 392: **concern** (non-blocking): We might want to consider breaking this out into a template as `content_tag` and friends are not ideal for complex cases like this.

app/jobs/announcement_published_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AnnouncementPublishedJob < ApplicationJob
+  queue_as :default
+  def perform(announcement:)
+    emails = announcement.event.followers.map(&:email_address_with_name)
+
+    emails.each do |email|
+      AnnouncementMailer.with(announcement:, email:).announcement_published.deliver_later
+    end
```
davidcornu: **suggestion**: Out of an abundance of precaution we might want to consider batching this query


```suggestion
    announcement.event.followers.find_each do |follower|
      AnnouncementMailer.with(
        announcement:, 
        email: follower.email_address_with_name
      ).announcement_published.deliver_later
    end
```

so we're not loading all followers into memory.

app/models/announcement.rb
```diff
@@ -0,0 +1,52 @@
...
+  has_paper_trail
+  acts_as_paranoid
+
+  belongs_to :user
+  belongs_to :event
+
+  alias_method :author, :user
+
+  def publish
+    AnnouncementPublishedJob.new.perform(announcement: self)
```
davidcornu: **suggestion**: Do we want to enqueue this rather than perform it inline?

app/jobs/announcement_published_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AnnouncementPublishedJob < ApplicationJob
+  queue_as :default
+  def perform(announcement:)
```
davidcornu, line 5: **suggestion**: We might want to make sure the announcement is still published at this point just to account for cases where it might have been unpublished between the time this job is enqueued and when it runs.

app/models/announcement.rb
```diff
@@ -0,0 +1,52 @@
...
+class Announcement < ApplicationRecord
+  has_paper_trail
+  acts_as_paranoid
+
+  belongs_to :user
+  belongs_to :event
+
+  alias_method :author, :user
+
+  def publish
```
davidcornu: **nit**: This should probably be `publish!` if it's going to call `save!`

app/assets/stylesheets/components/_tiptap.scss
```diff
@@ -0,0 +1,7 @@
+.tiptap p.is-editor-empty:first-child::before {
+  color: #adb5bd;
+  content: attr(data-placeholder);
+  float: left;
+  height: 0;
+  pointer-events: none;
+}
```
davidcornu, line 7: **note**: This strikes me as very specific (the selector + color). Will need to run locally to provide better feedback.

app/policies/announcement_policy.rb
```diff
@@ -0,0 +1,62 @@
...
+  def create?
+    admin_or_manager? && !record.event.demo_mode?
+  end
+
+  def show?
+    Flipper.enabled?(:organization_announcements_tier_1_2025_07_07, record.event)
+  end
+
+  def edit?
+    (admin? || record.user == user) && record.draft
```
davidcornu: **concern**: Does `record.user == user` mean a user that is no longer part of the event still have access to edit announcements?

app/views/ach_transfers/new.html.erb
```diff
@@ -14,7 +14,7 @@
 <% end %>
 
  <% if @event.present? && !policy(@event).create_transfer? %>
-  <%= render "events/transfers/unauthorized_callout" %>
+  <%= render "events/unauthorized_callout" %>
```
davidcornu, line 17: **question**: Is this related to announcements?

app/views/announcement_mailer/announcement_published.html.erb
```diff
@@ -0,0 +1,4 @@
+<h1><%= @announcement.title %></h1>
+<%= @announcement.render_html %>
+<br>
+<i><%= link_to "This email was sent on behalf of #{@event.name}. If you would like to unsubscribe, please click here.", event_url(@event) %></i>
```
davidcornu: **suggestion**: Should we mention that we are sending this email because the user follows the event? We could also tell them when they started following the event based on `created_at` on the follow.

app/views/announcements/_announcement_card.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    <%= link_to announcement.event.name, announcement.event %>
+  <% end %>
+  <div class="flex items-center justify-between">
+    <p class="font-bold text-xl my-0 no-underline text-white">
+      <% if announcement.draft %>
+        <span class="text-orange border-orange rounded-full border-solid border-[1px] text-sm p-[6px] mr-1 no-underline">Draft</span>
+      <% end %>
+      <%= link_to announcement.title, event_announcement_path(announcement.event, announcement) %>
+    </p>
+    <% if announcement.user == current_user && !show_event %>
```
davidcornu: **question**: Can other managers/admins for the event perform these actions?

app/views/announcements/_announcement_form.html.erb
```diff
@@ -0,0 +1,47 @@
+<%# locals: (content: nil) %>
+
+<%= form_with model: @announcement, url: content.present? ? event_announcement_path(@event, @announcement) : event_announcements_path(@event), data: { controller: "tiptap", "tiptap-target": "form", "tiptap-content-value": content, "tiptap-editable-value": true, "turbo-frame": "_top" }.compact do |form| %>
```
davidcornu: **question**: `content.present?` strikes me as a bit odd as a way to differentiate whether this is a create or update. Might I suggest `@announcement.persisted?` (https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-previously_persisted-3F)?

davidcornu (commented):
I think I've gotten through everything save for the front-end which I'd rather review by running locally.

app/views/ach_transfers/new.html.erb
Luke-Oldenburg: It was reused in announcements so got moved out of transfers.

app/assets/stylesheets/components/_tiptap.scss
polypixeldev: This was from the Tiptap docs, as they do not include CSS
https://tiptap.dev/docs/editor/extensions/functionality/placeholder#install

app/controllers/announcements_controller.rb
YodaLightsabr: I think the way we're doing this is alright because of a few unique circumstances, but generally I would agree. We want to ensure that we're 404ing on organizations that aren't public (kind of like GitHub does for private repos). The only case where we want to show announcements for a non-transparent organization is if they've published some, or if the organizer is signed in.

app/helpers/application_helper.rb
YodaLightsabr: Agree. This would be a larger lift though, so we can make an issue for it after we merged.

app/models/announcement.rb
```diff
@@ -0,0 +1,62 @@
...
+  belongs_to :event
+
+  scope :published, -> { where.not(published_at: nil) }
+
+  def publish!
+    AnnouncementPublishedJob.perform_later(announcement: self)
+
+    self.published_at = Time.now
+
+    save!
```
YodaLightsabr:
```suggestion
    update!(published_at: Time.now)
```

YodaLightsabr (approved):
Looks good. A few small nitpicks and linting fixes and we should be good to go.

YodaLightsabr (approved):

PR #10832, opened 2025-07-04
Fix events query in `WeeklyYswsEventSummaryJob`

davidcornu:
## Summary of the problem

The `created_at` condition was ambiguous because `.ysws` adds two `JOIN` clauses.

https://hackclub.slack.com/archives/C04LDFJ5X7S/p1751640227940719

## Describe your changes

Makes the `.where` more specific.


Luke-Oldenburg (approved):
Thank you!

PR #10834, opened 2025-07-06
Five Dollar Friends

Sarvesh-Mk:
## Summary of the problem

For the Five Dollar Friends TUB, to increase publicity there is now a red banner on any organization that is teen led and has less than 3 teens created/activated after July 3rd 2025.

## Describe your changes

Added a card on every tab that showcases my tub and links it to a fillout form. 

image
The bolded text underlines when mouse hovers over it


Luke-Oldenburg:
Should we be hiding this after they have three? What if they get 3 but haven't applied yet?

Sarvesh-Mk:
> Should we be hiding this after they have three? What if they get 3 but haven't applied yet?
In this case, due to how vibrant the banner is I don't feel like its necessary and is pretty hard to miss.

sampoder (approved):
lgtm! thank you Sarvesh

PR #10835, opened 2025-07-07
Migrate dependencies for Rails 8

KiKoS0:
## Summary of the problem
Bumping Flipper to `1.3.5` and switching from `annotate` to `annotaterb` before bumping Rails to 8.

https://github.com/hackclub/hcb/issues/8470

### Annotate
- Switch to the yaml configuration that `annotaterb` expects, following their migration guide: https://github.com/drwl/annotaterb/blob/main/MIGRATION_GUIDE.md#automatic-annotations-after-running-database-migration-commands
### Flipper
- Made sure that we are not using any of the methods that are in the [`1.0.0` breaking changes](https://github.com/flippercloud/flipper/releases/tag/v1.0.0):
  - `Flipper.bool`
  - `Flipper.boolean`
  - `Flipper.actor`
  - `Flipper.percentage_of_actors`
  - `Flipper.actors`
  - `Flipper.percentage_of_time`
  - `Flipper.time`
- Generate the update migration for Flipper introduced in [`v1.1.2`](https://github.com/flippercloud/flipper/releases/tag/v1.1.2).


config/initializers/flipper.rb
```diff
@@ -5,4 +5,8 @@
   # aren't running many betas at the time of writing. If Flipper is used
   # extensively in the future, this can be changed
   config.flipper.preload = false
+
+  # Setting to `true` or `:raise` will raise error when a feature doesn't exist.
+  # Use `:warn` to log a warning instead.
+  config.flipper.strict = false
```
KiKoS0: The default is `!Rails.env.production?` but my understanding is that we are adding feature flags ad-hoc in development so we need this set to `false`. Let me know otherwise.

db/migrate/20250706230741_change_flipper_gates_value_to_text.rb
```diff
@@ -0,0 +1,20 @@
...
+    # Ensure this incremental update migration is idempotent
+    return unless connection.column_exists? :flipper_gates, :value, :string
+
+    if index_exists? :flipper_gates, [:feature_key, :key, :value]
+      remove_index :flipper_gates, [:feature_key, :key, :value]
+    end
+    change_column :flipper_gates, :value, :text
+    safety_assured do
+      add_index :flipper_gates, [:feature_key, :key, :value], unique: true, length: { value: 255 }
+    end
```
KiKoS0, line 14: I'm not sure if how big is `flipper_gates`, if it is I'll move it to new migration that creates the index concurrently.

.annotaterb.yml
davidcornu, line 1: **question**: Are these the default settings? Do we absolutely need them committed or do we get the same behaviour without `.annotaterb.yml`?

config/initializers/flipper.rb
davidcornu: **suggestion**: I think `:warn` might be a good compromise here. Ideally we'd want to catch cases where we're checking flags that don't exist but I don't believe we're set up to pre-populate the ones we need.

db/migrate/20250706230741_change_flipper_gates_value_to_text.rb
davidcornu: Ran `select count(*) from flipper_gates` in production and got 2,731. I think we're safe here.

db/migrate/20250706230741_change_flipper_gates_value_to_text.rb
```diff
@@ -0,0 +1,20 @@
...
+  def up
+    # Ensure this incremental update migration is idempotent
+    return unless connection.column_exists? :flipper_gates, :value, :string
+
+    if index_exists? :flipper_gates, [:feature_key, :key, :value]
+      remove_index :flipper_gates, [:feature_key, :key, :value]
+    end
+    change_column :flipper_gates, :value, :text
+    safety_assured do
+      add_index :flipper_gates, [:feature_key, :key, :value], unique: true, length: { value: 255 }
```
davidcornu, line 13: **note**: This struck me as a bit odd but it's what Flipper recommends: https://github.com/flippercloud/flipper/blob/87e86745cdcfd2a728c30009cdee88ee057bca82/lib/generators/flipper/templates/migration.erb#L15

davidcornu (commented):
Thanks for the PR!

.annotaterb.yml
KiKoS0: Yeah there are a few different defaults, e.g: annotating tests

PR #10836, opened 2025-07-07
Upgrade to Rails 8

KiKoS0:
## Summary of the problem
Upgrading to Rails 8 & update configuration to new defaults.

## Describe your changes

- Bump all rails deps versions.
- Generate new framework defaults.
- Remove hard deprecated usages of `model: nil` in `form_with` calls: https://github.com/rails/rails/commit/0c150bacaaca7772121300ca124e3ed5392f6ab9
- Generate Rails 8 defaults and merge them with our config.


config/initializers/new_framework_defaults_8_0.rb
```diff
@@ -0,0 +1,30 @@
...
+# Read the Guide for Upgrading Ruby on Rails for more info on each option.
+# https://guides.rubyonrails.org/upgrading_ruby_on_rails.html
+
+###
+# Specifies whether `to_time` methods preserve the UTC offset of their receivers or preserves the timezone.
+# If set to `:zone`, `to_time` methods will use the timezone of their receivers.
+# If set to `:offset`, `to_time` methods will use the UTC offset.
+# If `false`, `to_time` methods will convert to the local system UTC offset instead.
+#++
+Rails.application.config.active_support.to_time_preserves_timezone = :zone
```
KiKoS0, line 20: I think this is safe if we are not overriding TZs right? Either in config or with [`Time.use_zone`](https://apidock.com/rails/v6.0.0/Time/use_zone/class).

config/environments/production.rb
```diff
@@ -150,6 +175,19 @@
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 
+  # Only use :id for inspections in production.
+  config.active_record.attributes_for_inspect = [:id]
```
KiKoS0, line 179: This is obviously safer but let me know if we don't want to have it for ops.

config/environments/development.rb
```diff
@@ -67,6 +67,12 @@
   # Highlight code that triggered database queries in logs.
   config.active_record.verbose_query_logs = true
 
+  # Append comments with runtime information tags to SQL queries in logs.
+  config.active_record.query_log_tags_enabled = true
```
davidcornu, line 71: 🙌 Hopefully AppSignal picks these up.

config/environments/production.rb
```diff
@@ -58,6 +60,21 @@
...
+      if path.start_with?("/assets/")
+        # Files in /assets/ are expected to be fully immutable.
+        # If the content change the URL too.
+        "public, immutable, max-age=#{1.year.to_i}"
+      else
+        # For anything else we cache for 1 minute.
+        "public, max-age=#{1.minute.to_i}, stale-while-revalidate=#{5.minutes.to_i}"
+      end
+    end
+  }
```
davidcornu, line 76: **concern**: While this is a sensible default for Rails apps I'm not sure how HCB manages its assets (e.g. do they even go through Rails or are they served by another server?). @garyhtou thoughts?

config/environments/production.rb
davidcornu: **concern**: We might be relying on this for the admin-specific views such as https://github.com/hackclub/hcb/blob/71883c21ee466faad47df57980439d828f55820e/app/views/application/_admin_viewer.html.erb#L9 but as far as I can tell we don't make direct use of `inspect`.

config/environments/test.rb
```diff
@@ -31,7 +34,7 @@
   config.cache_store = :null_store
 
   # Raise exceptions instead of rendering exception templates.
-  config.action_dispatch.show_exceptions = false
+  config.action_dispatch.show_exceptions = :rescuable
```
davidcornu, line 37: Neat! https://guides.rubyonrails.org/configuring.html#config-action-dispatch-show-exceptions

config/initializers/new_framework_defaults_8_0.rb
```diff
@@ -0,0 +1,32 @@
...
+# When both `If-Modified-Since` and `If-None-Match` are provided by the client
+# only consider `If-None-Match` as specified by RFC 7232 Section 6.
+# If set to `false` both conditions need to be satisfied.
+#++
+Rails.application.config.action_dispatch.strict_freshness = true
+
+###
+# Set `Regexp.timeout` to `1`s by default to improve security over Regexp Denial-of-Service attacks.
+#++
+Regexp.timeout = 1
```
davidcornu, line 32: Nice. We use user-provided regular expressions for card grants https://github.com/hackclub/hcb/blob/d293e57763729f69a1612606e22095761650eccd/app/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb#L96 so this is a good fail-safe.

config/initializers/new_framework_defaults_8_0.rb
davidcornu: This seems to me like a safer default.

Before, if you did something like

```ruby
local = ActiveSupport::TimeZone['America/Montreal'].now
```

then `local.to_time.zone` would be `nil` and there would just be `utc_offset` to figure out how it relates to UTC.

With the new setting, `local.to_time.zone` returns the same `zone` object as `local`, which is better as the UTC offset of a time zone is not constant (rules can change over time + daylight savings).

config/environments/production.rb
24c02: `awesome_print` doesn't respect `attributes_for_inspect` (or `filter_parameters` or love or money or any higher power, for that matter) – i dream of the world where it does though

app/models/invoice.rb
```diff
@@ -268,7 +268,14 @@ def set_fields_from_stripe_invoice(inv = remote_invoice)
...
     self.due_date = Time.at(inv.due_date).to_datetime # convert from unixtime
     self.ending_balance = inv.ending_balance
-    self.finalized_at = inv.respond_to?(:status_transitions) ? inv.status_transitions.finalized_at : inv.try(:finalized_at)
+
+    finalized_value = if inv.respond_to?(:status_transitions)
+                        inv.status_transitions.finalized_at
+                      else
+                        inv.try(:finalized_at)
+                      end
+    self.finalized_at = finalized_value ? Time.at(finalized_value.to_i).to_datetime : nil
```
KiKoS0, line 289: According to these two issues below, the Rails 7 behavior was to silently set it to `nil` and it was never being assigned correctly. ActiveRecord 8 behavior is to raise a `PG::DatetimeFieldOverflow` error.

- https://github.com/rails/rails/issues/53679
- https://github.com/rails/rails/issues/54110

```ruby
ActiveRecord::StatementInvalid:
 PG::DatetimeFieldOverflow: ERROR: date/time field value out of range: "1752413473"
 HINT: Perhaps you need a different "datestyle" setting.
 CONTEXT: unnamed portal parameter $14 = '...'
```

I'm not sure if there are other occurrences of this that aren't covered by tests; If there are it might not be a bad idea to monkey patch it like in: https://github.com/rails/rails/issues/53679#issuecomment-2530827971

KiKoS0: not really sure why we need `respond_to?` when `status_transitions` isn't nullable: https://docs.stripe.com/api/invoices/object?api-version=2024-06-20#invoice_object-status_transitions
also doesn't look like the Invoice object can ever have a `finalized_at` field.

albertchae:
I propose we merge https://github.com/hackclub/hcb/pull/11289 before this and then slice out some of the deprecation fixes from the Rails 8 PR and merge those separately, what do you think @davidcornu @garyhtou ?

davidcornu:
> I propose we merge #11289 before this and then slice out some of the deprecation fixes from the Rails 8 PR and merge those separately

Makes sense to me 👍

config/environments/production.rb
garyhtou: IIRC, assets are served by the same Rails server. It does get cached by Cloudfront. If I'm understanding Rails assets correctly, the 1 year should be fine because if the asset changes, there will be a new fingerprint

app/models/invoice.rb
albertchae: extracted this fix to https://github.com/hackclub/hcb/pull/11872

manu-gurudath:
Is this PR still relevant?

PR #10838, opened 2025-07-07
Fix banner showing in playground mode - Teenage Referral TUB  

Sarvesh-Mk:
## Summary of the problem

Banner shows in playground mode on top of playground mode banner if the org has the teenager tag.

## Describe your changes

Added clause in if statement to check if organization is in demo mode and only displays the banner if it is not in demo mode.

before fix:
image
after fix:
image
^ Organization activated
image
^ During playground mode


sampoder (approved):
lgtm!

PR #10841, opened 2025-07-07
Create card grant settings when the plan is set to affiliate

Mohamad-Mortada:
## Summary of the problem
Currently, when you create an affiliated org you don't see the card grants page until you create a grant which builds the settings. 

## Describe your changes
I added a condition in the modal to build settings when that event's plan is set to affiliate.

### DIFF
--- a/app/models/event/plan.rb
+++ b/app/models/event/plan.rb
@@ -91,7 +91,7 @@ class Event
         errors.add(:type, "is invalid")
       end
     end
-    
+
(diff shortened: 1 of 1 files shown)


Luke-Oldenburg:
Please see https://github.com/hackclub/hcb/pull/10787 for details on how to fix this issue.

Mohamad-Mortada:
Thanks will implement :)

PR #10861, opened 2025-07-08
[Admin] Add risk level dropdown at org creation

Sarvesh-Mk:
## Summary of the problem

There is no easy way for operations to determine the risk level of an organization when creating it.

## Describe your changes

Created a dropdown at the bottom of admin org creation page that can be used to set the organization risk level

image

### DIFF
--- a/app/services/event_service/create.rb
+++ b/app/services/event_service/create.rb
@@ -34,9 +34,9 @@ module EventService
         @emails.each do |email|
           OrganizerPositionInviteService::Create.new(event:, sender: point_of_contact, user_email: email, is_signee: @is_signee).run!
         end
-        
-        
-        event.update!(risk_level: @risk_level) if !@risk_level.blank?  
+
+
+        event.update!(risk_level: @risk_level) if @risk_level.present?
         event
       end
     end
--- a/app/services/event_service/create.rb
+++ b/app/services/event_service/create.rb
@@ -35,8 +35,6 @@ module EventService
           OrganizerPositionInviteService::Create.new(event:, sender: point_of_contact, user_email: email, is_signee: @is_signee).run!
         end
 
-
-        event.update!(risk_level: @risk_level) if @risk_level.present?
         event
       end
     end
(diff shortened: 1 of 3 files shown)


app/services/event_service/create.rb
```diff
@@ -34,6 +35,8 @@ def run
           OrganizerPositionInviteService::Create.new(event:, sender: point_of_contact, user_email: email, is_signee: @is_signee).run!
         end
 
+
+        event.update!(risk_level: @risk_level) if @risk_level.present?
```
davidcornu: - **suggesiton**: Given that `risk_level` is an attribute on `Event` I would include it in the `attrs` method instead of performing a separate update here. Something like

    ```diff
    diff --git a/app/services/event_service/create.rb b/app/services/event_service/create.rb
    index aaa47eeb6d..ed26fb35f5 100644
    --- a/app/services/event_service/create.rb
    +++ b/app/services/event_service/create.rb
    @@ -35,8 +35,6 @@
               OrganizerPositionInviteService::Create.new(event:, sender: point_of_contact, user_email: email, is_signee: @is_signee).run!
             end
     
    -
    -        event.update!(risk_level: @risk_level) if @risk_level.present?
             event
           end
         end
    @@ -54,7 +52,9 @@
             point_of_contact_id: @point_of_contact_id,
             demo_mode: @demo_mode,
             plan: Event::Plan.new(type: @plan)
    -      }
    +      }.tap do |hash|
    +        hash[:risk_level] = @risk_level if @risk_level.present?
    +      end
         end
     
         def point_of_contact
    ```
- **note**: I confirmed that https://github.com/hackclub/hcb/blob/ba6d9bd4c8295f86be9fde08531253df38a6715f/app/models/event.rb#L398-L403 will prevent invalid `risk_level` values from being set so we probably don't need any additional validation here.

Sarvesh-Mk: I implemented this, I had no idea what .tap was but its really cool!

davidcornu (approved):

PR #10862, opened 2025-07-08
Log to STDOUT in development

davidcornu:
## Summary of the problem

- When running `bin/dev` we don't actually see Rails logs in the console as they to go `log/development.log` by default
- They also differ from production where we use `ActiveSupport::TaggedLogging` https://github.com/hackclub/hcb/blob/4ff2c08b46a88485d5efd67556e751d1e6358413/config/environments/production.rb#L133-L137

## Describe your changes

Mirror the production config in development and set `RAILS_LOG_TO_STDOUT` in `Procfile.dev`


sampoder (approved):
looks good to me!

PR #10864, opened 2025-07-08
Mark Donation Tiers feature as fully shipped

manuthecoder:
This PR moves card grants to the "Past feature previews" and marks it as fully shipped

image


garyhtou (approved):
Image
cool stuff!

PR #10874, opened 2025-07-09
Set `autocomplete` attribute on totp + login code fields

davidcornu:
## Summary of the problem

Browsers use the `autocomplete` attribute on input fields to suggest useful values. In the case of login codes, this can involve looking at the user's SMS messages or emails and suggesting codes that were sent to them (which is very handy).

We started doing this in https://github.com/hackclub/hcb/pull/9625 for a new version of these fields but that flag was never rolled out, so until we figure out what to do with that logic ([Slack discussion](https://hackclub.slack.com/archives/C047Y01MHJQ/p1752068620075929)) I figured this was a nice quality of life improvement.

## Describe your changes

- Cleans up template formatting
- Switches from `text_field` to `text_field_tag` as we aren't using `form_for`
- Adds `autocomplete="one-time-password"` to the appropriate fields

### DIFF
--- a/app/views/logins/login_code.html.erb
+++ b/app/views/logins/login_code.html.erb
@@ -34,6 +34,7 @@
             "",
             placeholder: "Enter your login code",
             class: "!max-w-full w-max",
+            autocomplete: "one-time-code",
             required: true,
             autofocus: true
           ) %>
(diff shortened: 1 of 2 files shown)


sampoder (approved):
lgtm! thank you

PR #10881, opened 2025-07-09
[Feature Preview] Make org announcement emojis separated

garyhtou:
## Summary of the problem

Before:
image

Emojis are stuck together in one string

## Describe your changes

After:
_use your imagination_

Emojis are confetti separately.

Since `%w[]` creates an array split by whitespace, you need whitespace between the emojis so that they're separate elements in the array


Luke-Oldenburg:
haha this was a fun bug

davidcornu (approved):

PR #10885, opened 2025-07-09
[Announcements] More routing and error fixes

Luke-Oldenburg:

### DIFF
--- a/app/controllers/event/follows_controller.rb
+++ b/app/controllers/event/follows_controller.rb
@@ -18,9 +18,13 @@ class Event
       }
       follow = Event::Follow.new(attrs)
       authorize follow
-      follow.save!
+      begin
+        follow.save!
+      rescue ActiveRecord::RecordNotUnique
+        # Do nothing. The user already follows this event.
+      end
+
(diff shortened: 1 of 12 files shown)


davidcornu (approved):

davidcornu:
Hold up - noticed something fishy

app/controllers/event/follows_controller.rb
```diff
@@ -19,7 +19,8 @@ def create
       follow = Event::Follow.new(attrs)
       authorize follow
       follow.save!
-      redirect_to event_announcements_path(@event)
+      redirect_to event_announcement_overview_path(@event)
+    rescue ActiveRecord::RecordNotUnique
```
davidcornu: **concern**: This will result in Rails attempting to render a `create` template which does not exist. We should instead do something like

```ruby
begin
  follow.save!
rescue ActiveRecord::RecordNotUnique
  # Do nothing. The user already follows this event.
end

redirect_to event_announcement_overview_path(@event)
```

davidcornu (approved):

PR #10887, opened 2025-07-09
[Announcements] Fix publish path and missing text

Luke-Oldenburg:
Publish path was causing a 500 when loading the overview page in certain ways.
The missing text was only targeted at organizers originally.


YodaLightsabr (approved):

YodaLightsabr:
LGTM

PR #10892, opened 2025-07-10
Referral links

YodaLightsabr:


Luke-Oldenburg:
Can we add another alias route for referral links? Some people get suspicious of the word even though it's just for internal stats tracking.

PR #10893, opened 2025-07-10
[Announcements] Custom elements

polypixeldev:
## Summary of the problem

Adding custom elements to announcements will allow users to add elements to announcements that include HCB data. Currently:
- Donation goal progress
- HCB codes
- Donation summaries

## Describe your changes

Custom elements are rendered with a placeholder in the frontend editor, and then with real data on the backend. To do this while freezing financial data, HTML must be stored in the content column of announcements instead of ProseMirror JSON. A one time job (`OneTimeJobs::MigrateAnnouncementContent`) has been added to migrate existing announcements to store the HTML instead of JSON - **this should be run immediately after this is deployed.**

This PR also adds confetti on publish :tada:

Screenshot From 2025-07-10 13-37-53
Screenshot From 2025-07-10 13-37-18
Screenshot From 2025-07-10 13-21-39
Screenshot From 2025-07-10 13-19-59


Luke-Oldenburg:
These look so good.

app/controllers/announcements_controller.rb
```diff
@@ -23,13 +26,13 @@ def create
...
     flash[:success] = "Announcement successfully #{params[:announcement][:draft] == "true" ? "drafted" : "published"}!"
+    confetti! if @announcement.published?
     redirect_to announcement_path(@announcement)
-
   rescue => e
-    puts e.message
     flash[:error] = "Something went wrong. #{e.message}"
     Rails.error.report(e)
-    redirect_to event_announcement_overview_path(@announcement.event)
+    authorize @event, :announcement_overview?, policy_class: EventPolicy
```
davidcornu: **question**: Do we not get `EventPolicy` by default here?

app/services/prosemirror_service/donation_goal_node.rb
```diff
@@ -0,0 +1,28 @@
...
+    def tag
+      [{ tag: self.class.tag_name, attrs: (@node.attrs.to_h || {}).merge({ class: "donationGoal relative card shadow-none border flex flex-col py-2 my-2" }) }]
+    end
+
+    def matching
+      @node.type == self.class.node_type
+    end
+
+    def text
+      event = Event.find(@node.attrs.event.to_i)
```
davidcornu: **concern**: Can we trust `@node.attrs.event.to_i`? If it's user data then theoretically it could contain anything and let you render any event's data. We might want to constraint it to the announcement's `event_id` (or skip it altogether)

app/services/prosemirror_service/donation_summary_node.rb
```diff
@@ -0,0 +1,28 @@
...
+    def tag
+      [{ tag: self.class.tag_name, attrs: (@node.attrs.to_h || {}).merge({ class: "donationSummary relative card shadow-none border flex flex-col py-2 my-2" }) }]
+    end
+
+    def matching
+      @node.type == self.class.node_type
+    end
+
+    def text
+      event = Event.find(@node.attrs.event.to_i)
```
davidcornu: **concern**: Same concern here about whether we can trust `@node.attrs.event.to_i`

app/services/prosemirror_service/hcb_code_node.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    def matching
+      @node.type == self.class.node_type
+    end
+
+    def text
+      event = Event.find(@node.attrs.event.to_i)
+
+      hcb_code = HcbCode.find_by_hashid(@node.attrs.code)
```
davidcornu: **concern**: Same concern here about whether we can trust `@node.attrs.event.to_i` and `@node.attrs.code`. We should also check that the the event has access to the `HcbCode` otherwise this could let you have access to any `HcbCode`.

app/services/prosemirror_service/donation_goal_node.rb
davidcornu: Looking at https://github.com/etaminstudio/prosemirror_to_html, there doesn't seem to be a way to pass context to the renderer so maybe we could do something like

```ruby
module ProsemirrorService
  class Renderer
    CONTEXT_KEY = :prosemirror_service_render_context

    class << self
      def with_context(new_context, &)
        old_context = context
        Fiber[CONTEXT_KEY] = new_context

        yield
      ensure
        Fiber[CONTEXT_KEY] = old_context
      end

      def context
        Fiber[CONTEXT_KEY]
      end

      def render
        with_context({ event: @event }) do
          # do actual rendering
          # This allows our custom nodes to retrieve the event with
          # ProsemirrorService::Render.context.fetch(:event)
        end
      end
    end
  end
end
```

davidcornu: I see you've implemented some checks in the views (https://github.com/hackclub/hcb/pull/10893/files#diff-bd7143156019bb1f2baa42cccd8460201be5812d6500226fd40e387e4f421e51R3-R5), which is good, but I'm reticent to rely on us remembering to perform those checks there rather than prevent invalid data from being loaded in the first place.

app/assets/stylesheets/mailer.scss
```diff
@@ -1,6 +1,10 @@
 @import 'basscss';
 @import 'space';
 
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
```
davidcornu, line 6: **concern**: It's been awhile since I've written HTML markup so this may no longer be the case, but do things like `flex` work consistently? Have we tested these emails in different clients (https://www.litmus.com/ is good for this)?

app/controllers/announcements_controller.rb
polypixeldev: Oh, yeah we do, thanks - my understansing of Pundit was that it determined the policy class from the controller and not the model

app/services/prosemirror_service/donation_goal_node.rb
polypixeldev: good catch! The context looks great, that's exactly what I need

app/assets/stylesheets/mailer.scss
polypixeldev: do you know how to get that configured in development? i'm trying to configure it to actually send the email with an smtp server but having trouble

polypixeldev: I've tried some different things, it displays perfect in Thunderbird and Proton Mail. Outlook seems to have a couple bugs and I can't get Gmail to style the email at all - I think it's just removing styling from the emails since it doesn't trust the email? This might just be better to see how it works in production

davidcornu: > do you know how to get that configured in development? i'm trying to configure it to actually send the email with an smtp server but having trouble

I think litmus just lets you paste in the markup which we can recover from `letter_opener` (or via the previews / rails console).

> I've tried some different things, it displays perfect in Thunderbird and Proton Mail. Outlook seems to have a couple bugs and I can't get Gmail to style the email at all - I think it's just removing styling from the emails since it doesn't trust the email? This might just be better to see how it works in production

For emails you basically have to use 90s era HTML and do progressive enhancement. The way I see it we either
1. Figure out a way to use the same markup for both the site and email (which means limiting ourselves to what's supported in emails)
2. Have an email-specific renderer for these nodes that's a lot simpler

My current feeling is (2) might be a lot more straightforward and easier to maintain.

app/services/prosemirror_service/hcb_code_node.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    def matching
+      @node.type == self.class.node_type
+    end
+
+    def text
+      event = ProsemirrorService::Renderer.context.fetch(:event)
+
+      hcb_code = HcbCode.find_by_hashid(@node.attrs.code)
```
davidcornu, line 22: **concern**: I would move the check you have in the view to here 


```ruby
hcb_code = HcbCode.find_by_hashid(@node.attrs.code)

unless hcb_code.event == event
  hcb_code = nil
end
```

and later

```erb
<% if hcb_code.nil? %>
 <p>This transaction cannot be displayed</p>
```

so whoever is reading this code can be 100% sure that anything that runs afterwards does not have access to `hcb_code` if it shouldn't.

app/views/announcements/_announcement_form.html.erb
```diff
@@ -1,6 +1,6 @@
 <%# locals: (content: nil) %>
 
-<%= form_with model: @announcement, url: @announcement.persisted? ? announcement_path(@announcement) : announcements_path, data: { controller: "tiptap", "tiptap-target": "form", "tiptap-content-value": content, "tiptap-editable-value": true, "turbo-frame": "_top" }.compact do |form| %>
+<%= form_with model: @announcement, url: @announcement.persisted? ? announcement_path(@announcement) : announcements_path, data: { controller: "tiptap", "tiptap-target": "form", "tiptap-content-value": content, "tiptap-event-value": @event.id, "turbo-frame": "_top" }.compact do |form| %>
```
davidcornu: **question**: Do we still need `tiptap-event-value` now that the nodes no longer store their own event id?

app/views/layouts/application.html.erb
davidcornu, line 1: **question**: I'm not sure I understand these changes. Why do we need to wait for `JSConfetti` to be defined?

app/assets/stylesheets/mailer.scss
davidcornu: Btw feel free to ship and iterate in another PR

polypixeldev: Yeah, I think I'll look into (2) and just have render a simple version of the element for emails. I tested it in litmus and the only other client it renders well in is Apple Mail.

app/views/layouts/application.html.erb
polypixeldev: The confetti is really weird, and I'm honestly not sure why `JSConfetti` is not always defined when the script first runs, but that was the behavior I observed while testing this.

Luke-Oldenburg: I think `JSConfetti` is only loaded when confetti is needed but I may be wrong.

db/migrate/20250714200934_change_announcement_content_to_jsonb.rb
```diff
@@ -0,0 +1,32 @@
...
+      change_column :announcements, :content, :jsonb, null: false, using: 'CAST(content AS JSON)'
+    end
+  end
+
+  def down
+    safety_assured do
+      change_column :announcements, :content, :text
+    end
+  end
+end
```
garyhtou:
```suggestion
end

```

garyhtou (approved):

garyhtou:
Samuel's fixing brakeman and then it should be good to go

PR #10894, opened 2025-07-10
block files larger than 25mb on comments

3kh0:
## Summary of the problem

comments on transactions allow you to upload almost any file, and any size, which is really bad for that aws bill

## Describe your changes

this adds a simple limit to comment creations, checking the file size to not be larger than 25mb (which should be very generous for most peeps)

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -10,7 +10,7 @@ class CommentsController < ApplicationController
   end
 
   def create
-    if params[:comment][:file].present? && params[:comment][:file].size > 25.megabytes
+    if comment_params[:file].present? && comment_params[:file].size > 25.megabytes
       flash[:error] = "File size must be less than 25MB"
       redirect_back(fallback_location: @commentable) and return
     end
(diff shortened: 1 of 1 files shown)


app/controllers/comments_controller.rb
```diff
@@ -10,6 +10,11 @@ def new
   end
 
   def create
+    if params[:comment][:file].present? && params[:comment][:file].size > 25.megabytes
```
Luke-Oldenburg:
```suggestion
    if comment_params[:file].present? && comment_params[:file].size > 25.megabytes
```

Luke-Oldenburg:
This is what I get after uploading too large of a file.
Screenshot From 2025-07-12 15-58-58

garyhtou (changes requested):
Please use `active_storage_validations` to validate this!

https://github.com/igorkasyanchuk/active_storage_validations

Also, 25mb feels a bit small — we should determine what the 95 percentile is in prod

davidcornu:
I wonder whether this is something we can push into the reverse-proxy (https://caddyserver.com/docs/caddyfile/directives/request_body, https://hatchbox.io/announcements/69) rather than pushing the entire request through to Rails before bailing.

Luke-Oldenburg:
This has been superseded by https://github.com/hackclub/hcb/pull/11809. Thank you for originally suggesting the change!

PR #10897, opened 2025-07-10
[Admin] Modify balances export query to include total raised

Sarvesh-Mk:
## Summary of the problem

When exporting to a CSV, the balances sheet does not include the amount raised.

## Describe your changes

Added line when exporting CSV so it includes the total amount raised in a column next to balances.

### DIFF
--- a/app/models/export/event/balances.rb
+++ b/app/models/export/event/balances.rb
@@ -58,7 +58,7 @@ class Export
       end
 
       def header
-        ::CSV::Row.new(headers, ["id", "name", "postal_code", "contact email", "organizers", "revenue fee", "url", "balance", "omitted?"], true)
+        ::CSV::Row.new(headers, ["id", "name", "postal_code", "contact email", "organizers", "revenue fee", "url", "balance", "total raised", "omitted?"], true)
       end
 
       def row(event)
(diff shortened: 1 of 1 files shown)


app/models/export/event/balances.rb
```diff
@@ -58,7 +58,7 @@ def events
       end
 
       def header
-        ::CSV::Row.new(headers, ["id", "name", "postal_code", "contact email", "organizers", "revenue fee", "url", "balance", "omitted?"], true)
+        ::CSV::Row.new(headers, ["id", "name", "postal_code", "contact email", "organizers", "revenue fee", "url", "balance", "total raised", "omitted?"], true)
```
davidcornu, line 61: **concern** Where do these exports end up being used? If non-admins have access to them this could be a breaking change as we're changing the column ordering (e.g. for excel things and such). If so, we should consider adding columns at the end.

Sarvesh-Mk: This is the export balances button in the organizations query that only **admins** (HCB Operations Members) can access. No one else will be able to access this information. This function is mostly used by operations to check statistics, I want to be able to see the total amount raised of organization for some of the TUBs im planning out.

PR #10901, opened 2025-07-10
Strip out `toggle_signee_status` and error if creating a contract fails

sampoder:


Luke-Oldenburg:
What is this for?

sampoder:
This was to attempt to prevent cases where people became signees without contracts

PR #10903, opened 2025-07-10
Add `initial_login_id` to the `logins` table

davidcornu:
Extracted from https://github.com/hackclub/hcb/pull/10671

## Summary of the problem

In order to re-authenticate the user when they perform a sensitive operation more than 2 hours after their last authentication, we plan on leveraging the existing `Login` model.

There's an outline of how logins currently work here: https://github.com/hackclub/hcb/blob/david/push-lywmwopxprpp/dev-docs/rfcs/001_sudo_mode/sudo_mode.md#how-does-authentication-work-today

When a user is logged the first time we have the following

image

For the first two hours after `login.created_at` the user is allowed to perform sensitive actions but afterwards we want to re-authenticate them. To do this, 

1. We create a new `Login` record whose `initial_login_id` points to the first `Login`
2. We authenticate the user
3. Upon completion, we mark the new `Login` as complete and point it at the original `UserSession`
4. The user now has another two hours in which to perform sensitive actions (determined by the max `created_at` of the logins associated with their current session)

image

If we need to re-authenticate the user again later on we end up with

image

**Questions we need to ask of this data**

Question | Answer
---- | ----
What was the initial login for this session? | `user_session.logins.where(initial_login_id: nil)`
When was the last time the user authenticated? | `logins.max_by(&:created_at)&.created_at`
Is a given login a re-authentication (e.g. so we can disable 2FA)? | `login.initial_login_id.present?`

## Describe your changes

Adds a nullable `initial_login_id` to the `logins` table

### DIFF
--- a/app/models/login.rb
+++ b/app/models/login.rb
@@ -10,6 +10,7 @@
 #  browser_token_ciphertext :text
 #  created_at               :datetime         not null
 #  updated_at               :datetime         not null
+#  initial_login_id         :bigint
 #  user_id                  :bigint           not null
 #  user_session_id          :bigint
 #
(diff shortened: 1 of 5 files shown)


sampoder (approved):
LGTM - thank you!

PR #10904, opened 2025-07-10
[Announcements] Follow linking, monthly follow summary email, HCB announcements on home

Luke-Oldenburg:
- Link to org in manage following modal
- Monthly follow summary email for organizers
- Temporarily add HCB announcements to home page while we determine a better solution

### DIFF
--- a/app/jobs/monthly_follower_summary_job.rb
+++ b/app/jobs/monthly_follower_summary_job.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class MonthlyFollowerSummaryJob < ApplicationJob
+  queue_as :default
+  def perform
+    Event.includes(:event_follows).where("event_follows.created_at > ?", 1.month.ago).references(:event_follows).find_each do |event|
+      mailer = EventMailer.with(event: event)
+      mailer.monthly_follower_summary.deliver_later
+    end
+  end
+
+end
--- a/app/jobs/monthly_follower_summary_job.rb
+++ b/app/jobs/monthly_follower_summary_job.rb
@@ -3,7 +3,7 @@
 class MonthlyFollowerSummaryJob < ApplicationJob
   queue_as :default
   def perform
-    Event.includes(:event_follows).where("event_follows.created_at > ?", Time.now.last_month.beginning_of_month..).references(:event_follows).find_each do |event|
+    Event.includes(:event_follows).where(event_follows: { created_at: Time.now.last_month.beginning_of_month.. }).find_each do |event|
       mailer = EventMailer.with(event: event)
       mailer.monthly_follower_summary.deliver_later
     end
--- a/app/jobs/monthly_follower_summary_job.rb
+++ b/app/jobs/monthly_follower_summary_job.rb
@@ -3,7 +3,7 @@
 class MonthlyFollowerSummaryJob < ApplicationJob
   queue_as :default
   def perform
-    Event.includes(:event_follows).where("event_follows.created_at > ?", 1.month.ago).references(:event_follows).find_each do |event|
+    Event.includes(:event_follows).where("event_follows.created_at > ?", Time.now.last_month.beginning_of_month..).references(:event_follows).find_each do |event|
       mailer = EventMailer.with(event: event)
       mailer.monthly_follower_summary.deliver_later
     end
--- a/app/mailers/event_mailer.rb
+++ b/app/mailers/event_mailer.rb
@@ -18,6 +18,17 @@ class EventMailer < ApplicationMailer
     mail to: @emails, subject: "#{@event.name} received #{@donations.length} #{"donation".pluralize(@donations.length)} this past month"
   end
 
+  def monthly_follower_summary
+    @follows = @event.event_follows.where(created_at: Time.now.last_month.beginning_of_month..).order(:created_at)
+
+    return if @follows.none?
+    return if @emails.none?
+
+    @total = @follows.length
+
+    mail to: @emails, subject: "#{@event.name} got #{@total} #{"follower".pluralize(@total)} this past month"
+  end
+
   def donation_goal_reached
     @goal = @event.donation_goal
     @donations = @event.donations.succeeded.where(created_at: @goal.tracking_since..)
--- a/app/views/announcements/_announcement_card.html.erb
+++ b/app/views/announcements/_announcement_card.html.erb
@@ -1,6 +1,6 @@
 <%# locals: (announcement: nil, show_event: false) %>
 
-<div class="card flex flex-col gap-2 <%= "border-warning border-solid" if announcement.draft? %>">
+<div class="card flex flex-col gap-2 <%= "border-warning border-solid" if announcement.draft? %> w-full">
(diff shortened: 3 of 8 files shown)


app/mailers/event_mailer.rb
```diff
@@ -18,6 +18,17 @@ def monthly_donation_summary
     mail to: @emails, subject: "#{@event.name} received #{@donations.length} #{"donation".pluralize(@donations.length)} this past month"
   end
 
+  def monthly_follower_summary
+    @follows = @event.event_follows.where(created_at: Time.now.last_month.beginning_of_month..).order(:created_at)
```
davidcornu, line 22: **concern**: This is different than the query in `MonthlyFollowerSummaryJob`. Is that intentional?

app/jobs/monthly_follower_summary_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class MonthlyFollowerSummaryJob < ApplicationJob
+  queue_as :default
+  def perform
+    Event.includes(:event_follows).where("event_follows.created_at > ?", 1.month.ago).references(:event_follows).find_each do |event|
```
davidcornu: **concern**: Can you double check that this doesn't return duplicate events if they had multiple follows created in the time span? 

I have a sneaking suspicion that given the SQL this generates

```ruby
Event.includes(:event_follows).where("event_follows.created_at > ?", 1.month.ago).references(:event_follows).to_sql
```

```sql
SELECT
    events.* -- modified for brevity
FROM
	"events"
	LEFT OUTER JOIN "event_follows" ON "event_follows"."event_id" = "events"."id"
WHERE
	"events"."deleted_at" IS NULL
	AND (event_follows.created_at > '2025-06-10 19:37:49.741117')
ORDER BY
	"events"."id" ASC
```

we'll end up with one row for every matching row in `event_follows` unless Rails magically adds a `distinct` somewhere.

app/mailers/event_mailer.rb
Luke-Oldenburg: The query in `MonthlyFollowerSummaryJob` is to retrieve the list of Events to email whereas this is to retrieve the list of follows on an event.

davidcornu: Right but this one is doing `Time.now.last_month.beginning_of_month` and the other is `1.month.ago` so we're potentially leaving out a slice at the beginning of the month.

app/jobs/monthly_follower_summary_job.rb
Luke-Oldenburg: I'm only getting one Event record returned when the Event has two Follows.

davidcornu: hmm I'll try to break it locally after the feedback session

app/jobs/monthly_follower_summary_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class MonthlyFollowerSummaryJob < ApplicationJob
+  queue_as :default
+  def perform
+    Event.includes(:event_follows).where("event_follows.created_at > ?", Time.now.last_month.beginning_of_month..).references(:event_follows).find_each do |event|
```
davidcornu: **suggestion**: This currently fails because you can't pass a range to `"event_follows.created_at > ?"`

```suggestion
    Event.includes(:event_follows).where(event_follows: { created_at: Time.now.last_month.beginning_of_month.. }).find_each do |event|
```

app/jobs/monthly_follower_summary_job.rb
davidcornu: Ok so ran the following test

```ruby
# frozen_string_literal: true

require "rails_helper"

RSpec.describe Event::Follow do
  it "returns one event" do
    users = create_list(:user, 3)

    event1 = create(:event)
    users.each do |user|
      Event::Follow.create!(event: event1, user:)
    end

    returned_events = []

    Event
      .includes(:event_follows)
      .where(event_follows: { created_at: Time.now.last_month.beginning_of_month.. })
      .find_each do |event|
        returned_events << event
      end

    expect(returned_events.size).to eq(1)
  end
end
```

and looks like Rails is indeed handling this by doing this in two queries, one to get the `DISTINCT "events"."id"` that match our constraints and a subsequent one to populate the `Event` objects

```
SQL (1.4ms)  SELECT DISTINCT "events"."id" AS alias_0, "events"."id" FROM "events" LEFT OUTER JOIN "event_follows" ON "event_follows"."event_id" = "events"."id" WHERE "events"."deleted_at" IS NULL AND "event_follows"."created_at" >= $1 ORDER BY "events"."id" ASC LIMIT $2  [["created_at", "2025-06-01 04:00:00"], ["LIMIT", 1000]]
SQL (0.8ms)  SELECT "events"."id" AS t0_r0, "events"."name" AS t0_r1, "events"."address" AS t0_r2, "events"."created_at" AS t0_r3, "events"."updated_at" AS t0_r4, "events"."emburse_department_id" AS t0_r5, "events"."slug" AS t0_r6, "events"."point_of_contact_id" AS t0_r7, "events"."hidden_at" AS t0_r8, "events"."donation_page_enabled" AS t0_r9, "events"."donation_page_message" AS t0_r10, "events"."is_public" AS t0_r11, "events"."public_message" AS t0_r12, "events"."last_fee_processed_at" AS t0_r13, "events"."aasm_state" AS t0_r14, "events"."country" AS t0_r15, "events"."holiday_features" AS t0_r16, "events"."can_front_balance" AS t0_r17, "events"."demo_mode" AS t0_r18, "events"."demo_mode_request_meeting_at" AS t0_r19, "events"."is_indexable" AS t0_r20, "events"."deleted_at" AS t0_r21, "events"."activated_at" AS t0_r22, "events"."increase_account_id" AS t0_r23, "events"."website" AS t0_r24, "events"."description" AS t0_r25, "events"."stripe_card_shipping_type" AS t0_r26, "events"."donation_thank_you_message" AS t0_r27, "events"."donation_reply_to_email" AS t0_r28, "events"."public_reimbursement_page_enabled" AS t0_r29, "events"."public_reimbursement_page_message" AS t0_r30, "events"."postal_code" AS t0_r31, "events"."reimbursements_require_organizer_peer_review" AS t0_r32, "events"."short_name" AS t0_r33, "events"."risk_level" AS t0_r34, "events"."financially_frozen" AS t0_r35, "events"."donation_tiers_enabled" AS t0_r36, "event_follows"."id" AS t1_r0, "event_follows"."user_id" AS t1_r1, "event_follows"."event_id" AS t1_r2, "event_follows"."created_at" AS t1_r3, "event_follows"."updated_at" AS t1_r4 FROM "events" LEFT OUTER JOIN "event_follows" ON "event_follows"."event_id" = "events"."id" WHERE "events"."deleted_at" IS NULL AND "event_follows"."created_at" >= $1 AND "events"."id" = $2 ORDER BY "events"."id" ASC  [["created_at", "2025-06-01 04:00:00"], ["id", 11]]
```

TIL

davidcornu (approved):

PR #10906, opened 2025-07-10
[Announcements] Fix show and add follower count

Luke-Oldenburg:
## Summary of the problem

Show didn't work for non-signed in users.

## Describe your changes

Fix show.

Add follower count to announcements page.

### DIFF
--- a/app/views/events/announcement_overview.html.erb
+++ b/app/views/events/announcement_overview.html.erb
@@ -5,7 +5,7 @@
 <h1 class="flex heading items-center">
   <span class="flex-grow flex items-center flex-1">
     Announcements
-    <%= badge_for "#{@event.followers.count} #{"follower".pluralize(@event.followers.count)}", class: "bg-muted" %>
+    <%= badge_for pluralize(@event.followers.count, "follower"), class: "bg-muted" %>
   </span>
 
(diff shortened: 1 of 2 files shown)


app/views/events/announcement_overview.html.erb
```diff
@@ -2,8 +2,11 @@
 <% page_md %>
 <%= render "events/nav", selected: :announcements %>
 
-<h1 class="heading">
-  <span class="flex flex-grow items-center">Announcements</span>
+<h1 class="flex heading items-center">
+  <span class="flex-grow flex items-center flex-1">
+    Announcements
+    <%= badge_for "#{@event.followers.count} #{"follower".pluralize(@event.followers.count)}", class: "bg-muted" %>
```
davidcornu: **suggestion** Does

```suggestion
    <%= badge_for pluralize(@event.followers.count, "follower"), class: "bg-muted" %>
```

not work? https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-pluralize

Luke-Oldenburg: Ah I didn't know that was a thing.

davidcornu (approved):

PR #10907, opened 2025-07-10
Hide tag actions for readers and auditors

polypixeldev:
## Summary of the problem

Tag suggestions and the remove tag button were showing for readers and auditors.

## Describe your changes

Changed the signed in checks to be `as: member`


Luke-Oldenburg (commented):
Could you also fix delete being accessible in https://github.com/hackclub/hcb/blob/main/app/views/hcb_codes/_tags.html.erb and https://github.com/hackclub/hcb/blob/main/app/views/events/settings/admin/_tags.html.erb please? Maybe we should consider modifying https://github.com/hackclub/hcb/blob/main/app/views/canonical_transactions/_tag.html.erb to be usable everywhere.

sampoder (approved):

PR #10910, opened 2025-07-11
[Reimbursements] Only allow managers to be assigned reviewer

Sarvesh-Mk:
## Summary of the problem

Members can be requested to approve reimbursement requests despite not actually being able to.

## Describe your changes

Event now has a managers property to fetch all the managers in the current organization. 

image
image
^ member is not shown as an option in dropdown
Fixes #10739

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -236,7 +236,7 @@ class Event < ApplicationRecord
   has_many :organizer_position_contracts, through: :organizer_position_invites, class_name: "OrganizerPosition::Contract"
   has_many :users, through: :organizer_positions
   has_many :signees, -> { where(organizer_positions: { is_signee: true }) }, through: :organizer_positions, source: :user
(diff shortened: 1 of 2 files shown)


app/models/event.rb
```diff
@@ -236,6 +236,7 @@ class Event < ApplicationRecord
   has_many :organizer_position_contracts, through: :organizer_position_invites, class_name: "OrganizerPosition::Contract"
   has_many :users, through: :organizer_positions
   has_many :signees, -> { where(organizer_positions: { is_signee: true }) }, through: :organizer_positions, source: :user
+  has_many :managers, -> { where(organizer_positions: { role: "manager" }) }, through: :organizer_positions, source: :user
```
```suggestion
  has_many :managers, -> { where(organizer_positions: { role: :manager }) }, through: :organizer_positions, source: :user
```

davidcornu (approved):

garyhtou:
Good find (and fixed)!

PR #10911, opened 2025-07-11
[DX] Swap local_assigns for strict locals in wires

Luke-Oldenburg:
## Summary of the problem

We were already using strict locals which would require a value even if we had set the default to nil. I don't see a reason to use local assigns which does a similar thing.

## Describe your changes

Swap `local_assigns` for strict locals. Also fixed usages of instance variables, poor formatting, and use auditor helper instead of checking `current_user`.


app/views/wires/_requirements.html.erb
```diff
@@ -1,18 +1,18 @@
...
-    <% callout_footer += " and events on plans with exemptions" if local_assigns[:event].plan.exempt_from_wire_minimum? %>
-    <% callout_footer += "— including #{local_assigns[:event].name}!" unless local_assigns[:event].minimum_wire_amount_cents > 100 %>
-    <% callout_footer += " and exempted events" if !@event.plan.exempt_from_wire_minimum? && Flipper.enabled?(:exempt_from_wire_minimum, @event) %>
+<% if event.present? %>
+  <% callout_footer += " and events on plans with exemptions" if event.plan.exempt_from_wire_minimum? %>
+  <% callout_footer += "— including #{event.name}!" unless event.minimum_wire_amount_cents > 100 %>
+  <% callout_footer += " and exempted events" if !event.plan.exempt_from_wire_minimum? && Flipper.enabled?(:exempt_from_wire_minimum, event) %>
 <% end %>
-<% callout_footer += " (admins can bypass this)" if current_user.admin? %>
+<% callout_footer += " (admins can bypass this)" if auditor_signed_in? %>
```
garyhtou:
```suggestion
<% callout_footer += " (admins can bypass this)" if admin_signed_in? %>
```

garyhtou (approved):

PR #10912, opened 2025-07-11
[DX] Remove instance variables from partial (disbursements)

Luke-Oldenburg:
## Summary of the problem

Code cleanliness

## Describe your changes

Pass required values for disbursements form as strict locals rather than using instance variables.

### DIFF
--- a/app/views/disbursements/_form.html.erb
+++ b/app/views/disbursements/_form.html.erb
@@ -1,10 +1,12 @@
-<% disabled = @source_event.present? && !policy(@source_event).create_transfer? %>
+<%# locals: (disbursement:, event: nil, source_event: nil, allowed_source_events:, allowed_destination_events:) %>
+
+<% disabled = source_event.present? && !policy(source_event).create_transfer? %>
 
 <%= form_with(model: disbursement, local: true, url: disbursements_path, html: { "x-data" => "{fee: null, event_id: null}" }) do |form| %>
   <%= form_errors(disbursement, "Disbursements") %>
 
-  <%= render partial: "select_organization", locals: { form: form, field_name: "source_event_id", disabled: disabled, events: @allowed_source_events, default_event: @source_event, sending: true } %>
-  <%= render partial: "select_organization", locals: { form: form, field_name: "event_id", disabled: disabled, events: @allowed_destination_events, default_event: nil, receiving: true, allow_custom_events: @source_event&.plan&.unrestricted_disbursements_enabled? } %>
+  <%= render partial: "select_organization", locals: { form: form, field_name: "source_event_id", disabled: disabled, events: allowed_source_events, default_event: source_event, sending: true } %>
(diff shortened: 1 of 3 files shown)


app/views/disbursements/_form.html.erb
```diff
@@ -19,7 +21,7 @@
   <div class="field">
     <div class="flex items-center">
       <span class="bold muted" style="width: 1rem;">$</span>
-      <%= form.number_field :amount, value: (@disbursement&.amount.nil? ? nil : (@disbursement.amount.to_f / 100)),
+      <%= form.number_field :amount, value: (disbursement&.amount.nil? ? nil : (disbursement&.amount.to_f / 100)),
```
davidcornu, line 24: - **suggestion**: `disbursement` is non-optional so we can afford to be a bit less cautious here

  ```suggestion
        <%= form.number_field :amount, value: (disbursement.amount.nil? ? nil : (disbursement.amount.to_f / 100)),
  ```
- **concern** (non-blocking): I assume `disbursement.amount` is USD cents? Should we be using a `Money` here to do the conversion instead of round-tripping through a float?

app/views/disbursements/_form.html.erb
```diff
@@ -69,7 +71,7 @@
 
   <div
       class="<%= "actions inline-block #{"tooltipped tooltipped--e" if disabled}" %>"
-      aria-label="<%= @event&.demo_mode? ? "Transfers are disabled in Playground mode" : disabled ? "Only managers can make transfers!" : nil %>">
+      aria-label="<%= event&.demo_mode? ? "Transfers are disabled in Playground mode" : disabled ? "Only managers can make transfers!" : nil %>">
```
davidcornu: **nit**: Nested ternaries are a bit mind-bending. 

```suggestion
      aria-label="<%= event&.demo_mode? ? "Transfers are disabled in Playground mode" : (disabled ? "Only managers can make transfers!" : nil) %>">
```

app/views/disbursements/_form.html.erb
Luke-Oldenburg: 1. If it's nil though shouldn't we still use the safety operator?
2. There's probably a cleaner way to do it but I tend to just leave it alone because it's so easy to break it on accident.

davidcornu: > If it's nil though shouldn't we still use the safety operator?

Oh didn't realize it could be `nil`. Let's leave as-is then.

> There's probably a cleaner way to do it but I tend to just leave it alone because it's so easy to break it on accident.

Yeah hence the _(non-blocking)_. This has been this way for awhile so I'm not expecting it to be changed here.

Luke-Oldenburg: Yea, definitely something I'd like to clean up in the future but not now.

davidcornu (approved):

app/views/disbursements/_form.html.erb
Luke-Oldenburg: It theoretically could be nil but I don't know the scenario where that would occur so I think it's best to leave. Strict locals means you could still pass it as nil just you have to explicitly state that.

PR #10914, opened 2025-07-11
[DX] Use admin_signed_in? helper and variants

Luke-Oldenburg:
## Summary of the problem

We were directly checking the users permission type.

## Describe your changes

We have helpers for this, lets use them!


davidcornu (approved):
Nice clean up

PR #10917, opened 2025-07-11
Release announcements

Luke-Oldenburg:
## Summary of the problem

we want it rolled out fully

## Describe your changes

- taking it off flipper woohoo!
- link to announcement overview instead of event in announcement card
- fix dev mode bug by adding bank to seed file
- fix policy issues
- convert follows to hashids

### DIFF
--- a/app/policies/event/follow_policy.rb
+++ b/app/policies/event/follow_policy.rb
@@ -3,7 +3,7 @@
 class Event
   class FollowPolicy < ApplicationPolicy
     def create?
-      user == record.user && Flipper.enabled?(:organization_announcements_tier_1_2025_07_07, record.event)
+      user == record.user
     end
 
     def destroy?
--- a/app/policies/event_policy.rb
+++ b/app/policies/event_policy.rb
@@ -81,7 +81,7 @@ class EventPolicy < ApplicationPolicy
   end
 
   def announcement_overview?
-    Flipper.enabled?(:organization_announcements_tier_1_2025_07_07, record)
+    true
   end
 
   def emburse_card_overview?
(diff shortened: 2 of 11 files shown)


app/policies/event/follow_policy.rb
```diff
@@ -3,7 +3,7 @@
 class Event
   class FollowPolicy < ApplicationPolicy
     def create?
-      user == record.user && Flipper.enabled?(:organization_announcements_tier_1_2025_07_07, record.event)
+      user == record.user
```
garyhtou, line 6: For a separate PR: I think we need stricter rules on who can follow. IIRC, it's only possible to allow organizations that are either:
- transparent
- or, have at least one post
(^ i say that because that's the only time the button would be visible for following)

I think that logic probably should be located here as well.

app/policies/event_policy.rb
```diff
@@ -81,7 +81,7 @@ def team?
   end
 
   def announcement_overview?
-    Flipper.enabled?(:organization_announcements_tier_1_2025_07_07, record)
+    true
```
garyhtou, line 84: What if the org is not transparent or does not have any posts?

app/views/events/settings/_features.html.erb
garyhtou, line 1: Wanna move this down to the "past feature review" section?

Luke-Oldenburg: I would but it was only up for a few days and never publicly announced as a feature preview.

app/policies/event_policy.rb
Luke-Oldenburg: That logic is handled inside of the controller so it generates a 404 instead of an unauthorized.

davidcornu (approved):

PR #10919, opened 2025-07-12
Convert IDs to slugs

Luke-Oldenburg:
## Summary of the problem

The development environment did not work properly because it was missing `Event`s with specific IDs such as the reimbursement clearinghouse.

## Describe your changes

I converted these IDs to slugs and added them to the DB seed.
> [!CAUTION]
> This PR should be reviewed thoroughly. It has the potential to severely break production if something is wrong.

> [!IMPORTANT]
> Internal organizations should not have their slugs changed anymore (not that we ever should have in the first place). Should we make a plan that disallows this?


sampoder:
We shouldn't do this - sorry. You can seed with these IDs, I believe? But slugs can change and this feels risky.

garyhtou:
@sampoder, if you'd like to join the discussion on this, please see https://hackclub.slack.com/archives/C047Y01MHJQ/p1752263629850819

PR #10920, opened 2025-07-12
[v4] Add missing withdraw method

Mohamad-Mortada:
## Summary of the problem
The routes indicate there is a v4 api for withdraw but the controller doesn't 

## Describe your changes
added the controller method


Luke-Oldenburg (approved):
LGTM! We probably want more specific error messages and input validation in the future.

sampoder (approved):
lets do it!

PR #10922, opened 2025-07-12
Prevent contract signees from removing themselves

polypixeldev:
## Summary of the problem

Contract signees should not be able to remove themselves from the org without an OPDR

## Describe your changes

Changed the policy to disable contract signees from removing themselves without an OPDR


Luke-Oldenburg:
Should we even be allowing contract signees to bypass OPDRs at all?

PR #10923, opened 2025-07-13
[Admin] Added text to describe stripe issuing tiers

### LINKED ISSUE #10814: [Admin] Add text describing Stripe Issuing card shipping tiers
### Describe the problem

https://docs.stripe.com/issuing/cards/physical/ship-cards

_No response_

Sarvesh-Mk:
## Summary of the problem

When admins change the card issuing type (Standard, Express or priority), there is no clear way to determine what it actually does

## Describe your changes

When you update the dropdown, It displays a p tag with info on the selected issuing type. Does not display anything for standard issuing type.

image
image
image
P.S. Found that alpine variables break when using turbo so I disabled it to force a page reload so that the alpine variables could be set correctly. If a reload didn't occur, upon updating both p tags for express and priority would show.

FIxes #10814

### DIFF
--- a/app/views/events/settings/_admin.html.erb
+++ b/app/views/events/settings/_admin.html.erb
@@ -27,7 +27,7 @@
     </details>
   <% end %>
 
-  <%= form_with(model: event, local: true) do |form| %>
+  <%= form_with(model: event, local: true, html: { "x-data" => "{ship_type: '#{event.stripe_card_shipping_type}'}" }) do |form| %>
     <%= form_errors event, "organization" %>
     <% admin_tool do %>
       <%= link_to (event.hidden? ? "Un-hide project" : "Hide project"),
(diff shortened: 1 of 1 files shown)


app/views/events/settings/_admin.html.erb
```diff
@@ -60,7 +60,9 @@
 
       <div class="field">
         <%= form.label :stripe_card_shipping_type, "Card shipping method" %>
-        <%= form.collection_select :stripe_card_shipping_type, Event.stripe_card_shipping_types, :first, :first, {}, { disabled: } %>
+        <%= form.collection_select :stripe_card_shipping_type, Event.stripe_card_shipping_types, :first, :first, {}, { disabled:, "x-on:input" => "ship_type = $event.target.value" } %>
+        <p class="italic mt1" x-show="ship_type == 'express'">Express shiping will cost $13 per card and arrive in 4-5 buisness days (normal shipping costs $0.63 per card)</p>
```
Luke-Oldenburg:
```suggestion
        <p class="italic mt1" x-show="ship_type == 'express'">Express shipping will cost $13 per card and arrive in 4-5 buisness days (normal shipping costs $0.63 per card)</p>
```

Luke-Oldenburg (commented):
Can you move the info on standard shipping from parentheses to a p tag for when standard is selected?

app/views/events/settings/_admin.html.erb
Sarvesh-Mk: Done

Sarvesh-Mk:
image
image
image

Luke-Oldenburg (approved):
Nice work!

PR #10928, opened 2025-07-14
Small tweaks to raffles

Luke-Oldenburg:
- lock date on teen raffle
- use dynamic dates on robotics raffle

### DIFF
--- a/app/views/static_pages/index/_robotics_raffle.html.erb
+++ b/app/views/static_pages/index/_robotics_raffle.html.erb
@@ -1,5 +1,5 @@
 <% if current_user.teenager? && Referral::Program.find_by(id: 3)&.users&.where(id: current_user.id)&.any? && Date.current.month == 7 %>
-  <% raffle_program = "robotics-flipper-zero-raffle-july-2025" %>
+  <% raffle_program = "robotics-flipper-zero-raffle-#{Date.current.strftime("%m-%Y")}" %>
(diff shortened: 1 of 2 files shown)


app/views/static_pages/index/_teen_raffle.html.erb
```diff
@@ -20,8 +20,7 @@
           on <%= Date.current.end_of_month.strftime("%B %e, %Y") %>.
           <% if already_joined %>
             <span class="italic block mt-2 text-sm">
-              You have already joined the raffle this month. Remember to check
-              back in <%= next_month %> for the next Raffle!
+              You have already joined the raffle this month.
```
garyhtou: Let's keep the existing copy so they know they can come back next month. I think we're basically guaranteed to always have monthly raffles at this point.

app/views/static_pages/index/_robotics_raffle.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if current_user.teenager? && Referral::Program.find_by(id: 3)&.users&.where(id: current_user.id)&.any? && Date.current.month == 7 %>
-  <% raffle_program = "robotics-flipper-zero-raffle-july-2025" %>
+  <% raffle_program = "robotics-flipper-zero-raffle-#{Date.current.strftime("%m-%Y")}" %>
```
garyhtou, line 2: Do we know if we'll have monthly flipper zero raffles for robotics?

YodaLightsabr (approved):

PR #10929, opened 2025-07-14
[Announcement] Policy Changes & Auto-Following

Luke-Oldenburg:
- hide invoices in nav
- remove unused policy in event
- change manual 404 on announcement overview to policy check
- hide follow button if event is private and no announcements
- automatically follow organizers to their event on first announcement


app/policies/event_policy.rb
```diff
@@ -81,6 +74,8 @@ def team?
   end
 
   def announcement_overview?
+    return false if !@event.is_public && @all_announcements.empty? && !organizer_signed_in?
```
garyhtou: I'm not sure if `organizer_signed_in?` is defined here

app/views/events/_follow_button.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if !@event_follow.present? || !signed_in? %>
+<% if policy(@event).announcement_overview? && (!@event_follow.present? || !signed_in?) %>
```
garyhtou: Let's had a nested if. The outside will determine whether or not buttons should show. The inside will then choose between follow/unfollow.

app/policies/event/follow_policy.rb
```diff
@@ -3,6 +3,8 @@
 class Event
   class FollowPolicy < ApplicationPolicy
     def create?
+      return false if !record.event.is_public && !OrganizerPosition.role_at_least?(user, record.event, :reader)
+
       user == record.user
```
YodaLightsabr:
```suggestion
      (record.event.is_public || OrganizerPosition.role_at_least?(user, record.event, :reader)) && user == record.user
```

app/views/events/_follow_button.html.erb
```diff
@@ -1,11 +1,15 @@
...
-    <%= inline_icon "event-check" %>
-    Follow
-  <% end %>
-<% else %>
-  <%= button_to follow_path(@event_follow), method: :delete, class: "btn bg-blue flex items-center" do %>
-    <%= inline_icon "reply" %>
-    Unfollow
+<%# locals: (event:, event_follow: nil) %>
+
+<% if !@event.is_public && @all_announcements.empty? %>
```
YodaLightsabr:
```suggestion
<% if @event.is_public || @all_announcements.any? %>
```

PR #10934, opened 2025-07-14
[Admin] See target's email when processing deletion request

Sarvesh-Mk:
## Summary of the problem

When processing a deletion request, the deletee's email is not shown and isn't accessible easily. However the submitter's email is shown so I want to match that for the deletee too.

## Describe your changes

Added a line to show the deletee's email.

image
recipient email is what I added


app/views/organizer_position_deletion_requests/show.html.erb
```diff
@@ -32,6 +32,13 @@
       <%= user_mention @opdr.organizer_position.user %>
     </p>
 
+    <p>
+      <strong>reciepeint email</strong>
```
leowilkin:
```suggestion
      <strong>Recipient email</strong>
```

~that is _not_ how you spell that~ fixed spelling

leowilkin (changes requested):
Fixed spelling

app/views/organizer_position_deletion_requests/show.html.erb
Luke-Oldenburg, line 1: What if instead of a `copy_to_clipboard` we use a `mail_to` and also update the submitter email?

app/views/organizer_position_deletion_requests/show.html.erb
```diff
@@ -32,6 +32,13 @@
       <%= user_mention @opdr.organizer_position.user %>
     </p>
 
+    <p>
+      <strong>Recipient email</strong>
```
YodaLightsabr: Let's change the copy here. Recipient doesn't quite sound right here.

YodaLightsabr (dismissed):

app/views/organizer_position_deletion_requests/show.html.erb
sampoder:
```suggestion
      <strong>User email</strong>
```

sampoder (approved):
lgtm!

PR #10935, opened 2025-07-14
[Announcements] Change `content` column from `text` to `jsonb`

garyhtou:
## Summary of the problem

It stores json, but it's in a text column

## Describe your changes

convert it from a text column to a jsonb column

### DIFF
--- a/app/models/announcement.rb
+++ b/app/models/announcement.rb
@@ -5,23 +5,22 @@
 # Table name: announcements
 #
 #  id           :bigint           not null, primary key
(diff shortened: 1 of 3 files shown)


Luke-Oldenburg:
Isn't this being converted to html with custom elements?

garyhtou:
Lol i know it's confusing— we decided we still needed to keep the JSON. So we're adding an HTML column and keep the existing JSON. Also, i'm closing the PR since the migration is now copied into Samuel's branch

PR #10937, opened 2025-07-14
Fix searching

manuthecoder:
image

cc @garyhtou


garyhtou (approved):
Great! This fixes the flow of:
1. Add `:q` search
2. Add type `card charge`
3. Hit `x` to remove `:q` search

The following flow will still need to be fixed:
1. Add type `card charge`
2. Add `:q` search
It unexpectedly also clears all other filters (such as the type `card charge`)

PR #10938, opened 2025-07-15
Show sender available balance to admins on disbursement new

polypixeldev:
## Summary of the problem

@leowilkin requested to be able to see sender event balance when created a disbursement, which is currently hidden when an admin is signed in.

## Describe your changes

Added the balance below the org select.

image

### DIFF
--- a/app/controllers/disbursements_controller.rb
+++ b/app/controllers/disbursements_controller.rb
@@ -48,7 +48,7 @@ class DisbursementsController < ApplicationController
     user_event_ids = current_user.organizer_positions.reorder(sort_index: :asc).pluck(:event_id)
 
     @allowed_source_events = if admin_signed_in?
-                               Event.select(:name, :id, :demo_mode, :slug).all.reorder(Event::CUSTOM_SORT).includes(:plan)
+                               Event.select(:name, :id, :demo_mode, :slug, :can_front_balance).all.reorder(Event::CUSTOM_SORT).includes(:plan)
                              else
                                current_user.events.not_hidden.filter_demo_mode(false)
                              end.to_enum.with_index.sort_by { |e, i| [user_event_ids.index(e.id) || Float::INFINITY, i] }.map(&:first)
--- a/app/javascript/controllers/organization_select_controller.js
+++ b/app/javascript/controllers/organization_select_controller.js
@@ -10,7 +10,7 @@ export default class extends Controller {
     'wrapper',
     'field',
     'other',
-    'balance'
+    'balance',
   ]
   static values = {
     state: Boolean,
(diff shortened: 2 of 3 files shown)


app/controllers/disbursements_controller.rb
```diff
@@ -48,7 +48,7 @@ def new
     user_event_ids = current_user.organizer_positions.reorder(sort_index: :asc).pluck(:event_id)
 
     @allowed_source_events = if admin_signed_in?
-                               Event.select(:name, :id, :demo_mode, :slug).all.reorder(Event::CUSTOM_SORT).includes(:plan)
+                               Event.select(:name, :id, :demo_mode, :slug, :can_front_balance).all.reorder(Event::CUSTOM_SORT).includes(:plan)
```
davidcornu, line 51: **concern** (non-blocking): I find it odd that we're using a constrained `select` here (as opposed to skipping it or using a `pluck`) as we run the risk of downstream logic being wrong on account of columns not being loaded.

app/views/disbursements/_select_organization.html.erb
```diff
@@ -17,7 +17,7 @@
                     <% disabled_message = nil %>
                     <% disabled_message = "Insufficient balance" if sending && !admin_signed_in? && event.balance_available <= 0 %>
                     <% disabled_message = "HCB transfers disabled" if sending && !policy(event).create_transfer? %>
-                    <div data-id="<%= event.public_id %>" data-slug="<%= event.slug %>" data-name="<%= event.name %>" data-organization-select-target="organization" style="<%= i > 50 ? "display: none" : "" %>" data-index="<%= i %>">
+                    <div data-id="<%= event.public_id %>" data-slug="<%= event.slug %>" data-name="<%= event.name %>" data-organization-select-target="organization" style="<%= i > 50 ? "display: none" : "" %>" data-index="<%= i %>" data-balance="<%= sending && admin_signed_in? ? event.available_balance : 0 %>">
```
davidcornu, line 20: **concern**: If my very limited understanding of `available_balance` is correct and given that we have a lot of `event` records, this will result in _a lot_ of database and memory usage which we might not want to do just to populate a dropdown. 

Might I suggest we instead make use of the `EventsController#async_balance` logic that @YodaLightsabr added to show the balance once an event has been selected?

https://github.com/hackclub/hcb/blob/6e1deab06a36ee580aedc5b5abd8b4ed4a64b7b7/app/controllers/events_controller.rb#L542-L546

https://github.com/hackclub/hcb/blob/80fb00e7a7e8ed7b6ce4f7b16ff213b57136143b/app/views/events/async_balance.html.erb#L1-L7

app/javascript/controllers/organization_select_controller.js
```diff
@@ -128,6 +129,19 @@ export default class extends Controller {
...
+          let usdFormatter = new Intl.NumberFormat('en-US', {
+            style: 'currency',
+            currency: 'USD',
+          })
+
+          let formattedAmount = usdFormatter.format(
+            Number(organization.dataset['balance']) / 100
+          )
+
+          this.balanceTarget.innerText = formattedAmount
```
davidcornu, line 142: **suggestion**: If we're not actually manipulating `organization.dataset['balance']` we might be able to just put the formatted string in `data-balance` and save ourselves this logic.

YodaLightsabr:
Another potential approach is replacing the ID with the async balance that we already show for non-admins. If admins still need to see the ID for an organization, we can move it into a tooltip on the amount field.

polypixeldev:
Good idea - showing the balance in the picker + ID tooltip is what seems more useful, Leo agrees.

polypixeldev:
Created a new PR for the new approach - #11219

PR #10944, opened 2025-07-15
[v4 API] add user lookup routes, restricted oauth2 tokens

24c02:
adds `/api/v4/users/<public_id>` and `/api/v4/users/by_email/<email>` for admins...
also, brings in a restricted scope for `ApiToken`s + `require_oauth2_scope` on v4 controllers:
apps can opt in by taking the "restricted" scope – then, they won't be able to access any actions that `require_oauth2_scope` a scope they don't have or haven't been scoped yet.

### DIFF
--- a/app/controllers/api/v4/users_controller.rb
+++ b/app/controllers/api/v4/users_controller.rb
@@ -23,6 +23,8 @@ module Api
         render :show
       end
 
+      require_oauth2_scope "user_lookup", :show, :by_email
+
       def available_icons
         icons = {
           frc: current_user.events.robotics_team.any?,
--- a/app/controllers/api/v4/users_controller.rb
+++ b/app/controllers/api/v4/users_controller.rb
@@ -17,7 +17,6 @@ module Api
       end
 
       def by_email
-        puts params[:email]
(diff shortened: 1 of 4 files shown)


app/controllers/api/v4/users_controller.rb
```diff
@@ -4,11 +4,27 @@ module Api
...
+      end
 
       def show
-        @user = authorize current_user
+        @user = User.find_by_public_id!(params[:id])
+        authorize @user
       end
 
+      def by_email
+        puts params[:email]
```
garyhtou:
```suggestion
```

app/controllers/api/v4/users_controller.rb
```diff
@@ -4,11 +4,27 @@ module Api
...
+
+      def me
+        @user = authorize current_user, :show?
+        render :show
+      end
 
       def show
-        @user = authorize current_user
+        @user = User.find_by_public_id!(params[:id])
+        authorize @user
```
garyhtou, line 16: Double check policy for this

garyhtou (approved):

PR #10946, opened 2025-07-15
Use AASM for announcement draft state

polypixeldev:
## Summary of the problem

Preparing for announcement prompting which will need the state to be AASM

## Describe your changes

Added AASM state for announcement draft state and used its helpers. Added script to backfill the column which should be run as soon as this is deployed.

### DIFF
--- a/app/models/announcement.rb
+++ b/app/models/announcement.rb
@@ -38,7 +38,7 @@ class Announcement < ApplicationRecord
     state :draft, initial: true
     state :published
 
-    event :publish do
+    event :mark_published do
       transitions from: :draft, to: :published
 
       after do
--- a/app/models/announcement.rb
+++ b/app/models/announcement.rb
@@ -54,7 +54,6 @@ class Announcement < ApplicationRecord
 
   scope :published, -> { where(aasm_state: :published) }
 
-  before_save do
     if content_changed?
       self.rendered_html = ProsemirrorService::Renderer.render_html(content, event)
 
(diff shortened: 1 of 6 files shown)


app/models/announcement.rb
```diff
@@ -28,15 +29,32 @@
...
+    state :published
+
+    event :mark_published do
+      transitions from: :draft, to: :published
+
+      after do
+        update!(published_at: Time.now)
+
+        AnnouncementPublishedJob.perform_later(announcement: self)
+      end
```
garyhtou:
```suggestion
  aasm timestamps: true do
    state :draft, initial: true
    state :published

    event :mark_published do
      transitions from: :draft, to: :published

      after do
        AnnouncementPublishedJob.perform_later(announcement: self)
      end
```

app/models/announcement.rb
```diff
@@ -28,15 +29,30 @@
...
+
   validates :content, presence: true
 
   belongs_to :author, class_name: "User"
   belongs_to :event
 
-  scope :published, -> { where.not(published_at: nil) }
+  scope :published, -> { where(aasm_state: :published) }
 
   before_save do
```
garyhtou, line 55: This is automatically defined by AASM
```suggestion
```

garyhtou (approved):
ty

garyhtou:
I ended up running this code to also backfill deleted records:

```ruby
      Announcement.only_deleted.find_each do |announcement|
        da = announcement.deleted_at
        ua = announcement.updated_at
        announcement.update(deleted_at: nil)

        if announcement.published_at.present?
          announcement.update_column(:aasm_state, :published)
        else
          announcement.update_column(:aasm_state, :draft)
        end

        announcement.update_columns(deleted_at: da, updated_at: ua)
      end

```

PR #10948, opened 2025-07-15
Some more admin fixes

Luke-Oldenburg:
The better version of https://github.com/hackclub/hcb/pull/10925

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -47,7 +47,7 @@ class ApplicationController < ActionController::Base
 
   # Enable Rack::MiniProfiler for admins
   before_action do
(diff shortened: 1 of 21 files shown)


app/controllers/application_controller.rb
```diff
@@ -47,7 +47,7 @@ class ApplicationController < ActionController::Base
 
   # Enable Rack::MiniProfiler for admins
   before_action do
-    if current_user&.admin?
+    if admin_signed_in?
```
Luke-Oldenburg:
```suggestion
    if current_user&.admin?
```

YodaLightsabr (approved):

PR #10950, opened 2025-07-15
Refactor AnnouncementPublishedJob

### LINKED ISSUE #10947: Rename `AnnouncementPublishedJob` to `Announcement::PublishedJob`
Rename `AnnouncementPublishedJob` to `Announcement::PublishedJob`

Luke-Oldenburg:
Closes #10947


garyhtou (approved):
Thank you!

There's a slight chance we might get an error about `AnnouncementPublishedJob` being an uninitialized constant, but that's fine— I'm willing to take the risk since these aren't scheduled and we can manually fix it if we do see an error

PR #10951, opened 2025-07-16
Suppress missing receipt emails on salary event transactions

malted:
When card transactions are created on salary orgs and the user has missing receipts on a non-salary org, a "you have (5|7|9) missing receipts" message is sent every single time.


app/controllers/stripe_controller.rb
```diff
@@ -36,7 +36,9 @@ def handle_issuing_authorization_request(event)
     service = ::StripeAuthorizationService::Webhook::HandleIssuingAuthorizationRequest.new(stripe_event: event)
     approved = service.run
 
-    if approved
+    try_lock_cards? = service.event.plan != Event::Plan::SalaryAccount.name
```
malted: Blind change, not sure if this should be `.type`!

leowilkin:
~you're gonna hate me but this is the most underrated feature on HCB~

garyhtou (dismissed):
Thanks for the PR! We should add a `receipts_required?` method on all the `Event::Plan`, then use it here

Luke-Oldenburg:
That should probably be in the features string rather than as a separate method.

malted:
Adding to the features array doesn't seem to make sense - it isn't a feature.

I've made a `receipts_required?` method!

app/models/event/plan.rb
```diff
@@ -53,6 +53,10 @@ def standard?
       type == Event::Plan::Standard.name
     end
 
+    def receipts_required?
+      type == Event::Plan::SalaryAccount.name
+    end
```
davidcornu: - **concern**: This is odd. I would have assumed `Event::Plan::SalaryAccount` did not require receipts on account of https://github.com/hackclub/hcb/blob/5a6dde582fe07e5122881a41026ac9e706b778d1/app/models/event/plan/salary_account.rb#L30-L32
- **suggestion**: Given that we have an inheritance hierarchy for `Plan`, I would leverage it by doing
    ```ruby
    class Event::Plan
      def receipts_required?
        true
      end
    end
    ```
    and overriding it for specific subclasses
    ```ruby
    class Event::Plan::SalaryAccount
      def receipts_required?
        false
      end
    end
    ```

app/models/hcb_code.rb
```diff
@@ -500,6 +500,8 @@ def ct2
   def receipt_required?
     return false if pt&.declined?
 
+    return false if !event.plan.receipts_required?
```
davidcornu: **concern**: We can simplify this a bit with `unless` and we need to handle cases where `event` is `nil`

```suggestion
    return false unless event && event.plan.receipts_required?
```

app/controllers/stripe_controller.rb
```diff
@@ -36,7 +36,7 @@ def handle_issuing_authorization_request(event)
     service = ::StripeAuthorizationService::Webhook::HandleIssuingAuthorizationRequest.new(stripe_event: event)
     approved = service.run
 
-    if approved
+    if approved && service.event.plan.receipts_required?
```
sampoder: this logic should be in the job - we want to be doing as little as possible in this response as possible

app/models/event/plan.rb
sampoder: This should not be on `Event::Plan`. It should be on `Event::Plan::Standard`.

davidcornu: > It should be on `Event::Plan::Standard`.

That's a bit surprising. `Event::Plan` is the top of the hierarchy

Luke-Oldenburg: I think all plans inherit from standard though.

app/controllers/stripe_controller.rb
Luke-Oldenburg: This logic calls the job though?

davidcornu: @Luke-Oldenburg I think what @sampoder is suggesting here is that we leave this part of the code as-is and keep enqueueing the job, but modify the job to check whether receipts are required and bail out early if not (which means we're performing a bit more work in the job but save ourselves the extra check in the webhooks endpoint as these responses need to happen quickly so as not to block cards.

app/models/event/plan.rb
garyhtou: In reality, it should be on `Event::Plan`. At the moment, the bulk of the default settings are currently in `Event::Plan::Standard`. This should be fixed at some point.

garyhtou (commented):
Good to merge after addressing https://github.com/hackclub/hcb/pull/10951/files#r2373168871

davidcornu (approved):

PR #10952, opened 2025-07-16
[Ledger] Can now filter incoming/outgoing transactions on ledger

### LINKED ISSUE #10695: [Ledger] Incoming / outgoing transaction filter

Sarvesh-Mk:
## Summary of the problem

There is no clear way to filter transactions via incoming and outgoing.
image

## Describe your changes

Created direction parameter to filter the transactions from incoming and outgoing.

image
image
image
Fixes #10695


app/views/events/_filter.html.erb
```diff
@@ -114,6 +114,14 @@
           <% end %>
         </div>
       <% end %>
+      <% if @direction %>
+        <div class="badge badge-large bg-muted">
+          <%= @direction.humanize.gsub("In", "Incoming transactions").gsub("Out", "Outgoing transactions") %>
```
Luke-Oldenburg:
```suggestion
          <%= @direction.humanize.gsub("In", "Incoming").gsub("Out", "Outgoing") %>
```

Luke-Oldenburg (commented):
Does the check mark render on the left or right? It feels like there is a lot of padding around the new area.

app/views/events/_filter.html.erb
Sarvesh-Mk: image

Luke-Oldenburg: Are you able to get rid of all the extra padding? It looks really strange compared to the other tabs.

Sarvesh-Mk: I think incoming and outgoing works fine actually.

app/controllers/events_controller.rb
```diff
@@ -1079,13 +1081,15 @@ def set_transaction_filters
...
     @missing_receipts = params[:missing_receipts].present?
+    @direction = params[:direction]
 
     # Also used in Transactions page UI (outside of Ledger)
     @organizers = @event.organizer_positions.joins(:user).includes(:user).order(Arel.sql("CONCAT(preferred_name, full_name) ASC"))
   end
 
   def _show_pending_transactions
     return [] if params[:page] && params[:page] != "1"
+    return [] if params[:direction]
```
polypixeldev, line 1092: This should probably pass on the filter to the `PendingTransactionEngine::PendingTransaction:All.new` call instead of just not showing any pending transactions - you can add the `@expenses` and `@revenue` filters with similar SQL as in `TransactionGroupingEngine::Transaction:All`, but in https://github.com/hackclub/hcb/blob/main/app/services/pending_transaction_engine/pending_transaction/all.rb

app/views/events/ledger.html.erb
```diff
@@ -18,7 +18,7 @@
           </tr>
           </thead>
           <tbody data-behavior="transactions">
-          <%= render "pending_fee_transaction" if @event.fronted_fee_balance_v2_cents > 0 %>
+          <%= render "pending_fee_transaction" if @event.fronted_fee_balance_v2_cents > 0 && !@direction %>
```
polypixeldev: This can be rendered if the user is filtering for outgoing transactions, so it'd be better to check for:
```suggestion
          <%= render "pending_fee_transaction" unless @event.fronted_fee_balance_v2_cents <= 0 || @direction == "in" %>
```

leowilkin (changes requested):
LGTM - this is a handy feature, thank you!

YodaLightsabr (approved):
Looks good. Get another review from @garyhtou or @davidcornu before merging, though.

app/views/events/ledger.html.erb
```diff
@@ -18,7 +18,7 @@
           </tr>
           </thead>
           <tbody data-behavior="transactions">
-          <%= render "pending_fee_transaction" if @event.fronted_fee_balance_v2_cents > 0 %>
+          <%= render "pending_fee_transaction" unless @event.fronted_fee_balance_v2_cents <= 0 || @direction == "in" %>
```
sampoder:
```suggestion
          <%= render "pending_fee_transaction" unless @event.fronted_fee_balance_v2_cents <= 0 || @direction == "revenue" %>
```

sampoder (approved):
thank you! just pushed a small redesign to make this income / expenses

PR #10953, opened 2025-07-16
[Referrals] Store referral program on login and permit custom styling and copy

YodaLightsabr:
## Summary of the problem
Right now, we don't store `referral_program` on `logins`, which means we need to store the referral as a `return_to`.

## Describe your changes
This refactors that code to be more in line with our convention of storing login-based information on `login`.

It also allows `referral_programs` to provide custom text and a background for the login page.

image


app/controllers/logins_controller.rb
```diff
@@ -21,12 +21,16 @@ def new
...
     @prefill_email = params[:email] if params[:email].present?
+    @referral_program = Referral::Program.find(params[:referral]) if params[:referral].present?
   end
 
   # when you submit your email
   def create
     user = User.create_with(creation_method: :login).find_or_create_by!(email: params[:email])
-    login = user.logins.create
+
+    referral_program = Referral::Program.find(params[:referral_program_id]) if params[:referral_program_id].present?
```
garyhtou: can we do find_by so that it doesn't error?

garyhtou (approved):

garyhtou:
The goal here is to allow custom styling for login screens when users come through certain referral programs. For example, if a user come from Summer of Making, we'll give them a Summer of Making branded login page to smooth the transition from the SoM site to HCB.

PR #10962, opened 2025-07-16
[DX] Show Flipper actor names

davidcornu:
## Summary of the problem

When using the Flipper UI it's quite annoying to cross-reference the actor IDs with the corresponding models.

## Describe your changes

Configures `actor_names_source` to surface useful information about users and events (https://blog.flippercloud.io/flipper-1-1-0/#actor-names-737)

CleanShot 2025-07-16 at 17 29 27@2x


config/initializers/flipper.rb
```diff
@@ -10,3 +10,40 @@
...
+
+    grouped.each do |model, mapping|
+      case model
+      when "User"
+        User.where(id: mapping.keys).pluck(:id, :email, :full_name).each do |(id, email, full_name)|
+          actor_names[mapping.fetch(id)] = ActionMailer::Base.email_address_with_name(email, full_name)
+        end
+      when "Event"
+        Event.where(id: mapping.keys).pluck(:id, :name).each do |(id, name)|
+          actor_names[mapping.fetch(id)] = name if name.present?
```
garyhtou: **non-blocking**: Would `name` ever be blank? I think we might be safe to drop that check especially since we've now added the NOT NULL to the Event name column.

spec/misc/flipper_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      deleted_event.destroy!
+
+      deleted_user = create(:user)
+      deleted_user_flipper_id = deleted_user.flipper_id
+      deleted_user.destroy!
+
+      actor_names = Flipper::UI.configuration.actor_names_source.call(
+        [
+          user.flipper_id,
+          "User;✌️",
```
garyhtou, line 46: haha

garyhtou (approved):
Nice!

config/initializers/flipper.rb
davidcornu: Good point! I added the `NOT NULL` because I came across it in this PR 😬

PR #10967, opened 2025-07-16
Switch `events.name` to `NOT NULL`

davidcornu:
## Summary of the problem

While implementing `Event`-related code I noticed that we don't have any presence validations on `name` and the column is nullable, however
- We don't have any `NULL` or blank values in production
    CleanShot 2025-07-16 at 17 47 59@2x
- We validate that `name` is present on creation https://github.com/hackclub/hcb/blob/f66d59951366bbb5fa3544c807695710800d483e/app/services/event_service/create.rb#L22

## Describe your changes

- Adds a presence validation on `name`
- As per https://github.com/ankane/strong_migrations?tab=readme-ov-file#setting-not-null-on-an-existing-column,
  1. Adds a migration to add a non-validated check constraint on the column
  2. Adds a subsequent migration to validate (i) and add a `NOT NULL`


garyhtou (approved):
good stuff!

PR #10968, opened 2025-07-16
[Card Grants] Allow blocking merchants and categories on card grants

YodaLightsabr:
## Summary of the problem
Right now, pre-authorizations are fully functional, but you can't issue a grant with specific merchants or categories disallowed.

## Describe your changes
This PR adds the functionality to configure disallowed merchants and categories on an event card grant setting. It will then block those merchants and categories at authorization time before checking the merchant data against the allowed merchant logic.

image


app/models/card_grant_setting.rb
```diff
@@ -26,8 +28,11 @@ class CardGrantSetting < ApplicationRecord
   belongs_to :event
   serialize :merchant_lock, coder: CommaSeparatedCoder # convert comma-separated merchant list to an array
   serialize :category_lock, coder: CommaSeparatedCoder
+  serialize :banned_merchants, coder: CommaSeparatedCoder
+  serialize :banned_categories, coder: CommaSeparatedCoder
   alias_attribute :allowed_merchants, :merchant_lock
   alias_attribute :allowed_categories, :category_lock
+  alias_attribute :disallowed_merchants, :banned_merchants
```
garyhtou, line 35:
```suggestion
```

garyhtou (approved):

PR #10974, opened 2025-07-16
[Announcement] Auto-follow organizers

Luke-Oldenburg:
Organizers now automatically follow their event after the first announcement.

or if they join the event, and the event already has posts

### DIFF
--- a/app/models/announcement.rb
+++ b/app/models/announcement.rb
@@ -62,12 +62,35 @@ class Announcement < ApplicationRecord
   belongs_to :author, class_name: "User"
   belongs_to :event
 
+  before_save :autofollow_organizers
   before_save do
     if content_changed?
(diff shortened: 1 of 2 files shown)


app/models/announcement.rb
```diff
@@ -72,4 +73,26 @@ class Announcement < ApplicationRecord
     end
   end
 
+  private
+
+  def autofollow_organizers
+    # is this the first announcement to be published?
+    if published? && event.announcements.published.none?
+      event.users.find_each do |user|
+        unless event.followers.where({ id: user.id }).any?
```
YodaLightsabr:
```
      event.users.where.not(id: event.followers.pluck(:id)).find_each do |user|
```

This might work a little bit better and avoids another database query on each user.

garyhtou (approved):

PR #10979, opened 2025-07-17
Prevent updating the author just from editing the announcement

### LINKED ISSUE #10978: [Announcement] Prevent updating the author just from editing the announcement
We're updating `author` to `current_user` here:

https://github.com/hackclub/hcb/blob/2bf618333f8a0ed3be2514930d26d91c41ef597a/app/controllers/announcements_controller.rb#L56

We did this so that template_draft announcements' authors get switched from HCB to whoever edited it.

Let's update this logic so that we only pass in the author if the current author is HCB.

polypixeldev:
## Summary of the problem

Authors are being updated on every announcement edit, but it should only update if the author was HCB (created with a template draft).

Closes #10978 

## Describe your changes

Added a check to make sure the author is not updated when it was not the system user


sampoder (approved):
lgtm! thank you

app/controllers/announcements_controller.rb
```diff
@@ -55,7 +55,7 @@ def update
 
     content_hash = JSON.parse(params[:announcement][:json_content])
     @announcement.transaction do
-      @announcement.update!(announcement_params.merge(content: ProsemirrorService::Renderer.set_html(content_hash), author: current_user))
+      @announcement.update!(announcement_params.merge({content: ProsemirrorService::Renderer.set_html(content_hash), author: @announcement.author == User.system_user ? current_user : nil }.compact))
```
sampoder:
```suggestion
      @announcement.update!(announcement_params.merge({ content: ProsemirrorService::Renderer.set_html(content_hash), author: @announcement.author == User.system_user ? current_user : nil }.compact))
```
rubocop 🤖

PR #10980, opened 2025-07-17
Revert "Flipper for card grant popovers"

Luke-Oldenburg:
Reverts hackclub/hcb#10649

Card grants don't work in popovers for regular users because they can't render the stripe card view. It appears as fine to admins though.


YodaLightsabr:
I also think that popovers aren't the right move for card grants; it forces us to render a smaller view, which we already struggle with on the card grant initial funding transaction popover. I'm open to other suggestions though.

YodaLightsabr (approved):

PR #10982, opened 2025-07-17
Fix admin create event flow to create contract for invitee

### LINKED ISSUE #10738: [Admin][OrganizerPosition] Fix `is_signee` field
https://github.com/hackclub/hcb/blob/963338c8d20bf8969d3fca26100c6883dad5a97a/app/views/admin/event_new.html.erb#L20-L23

Make this the same UI/fields as the normal user invite form. It basically should ask whether or not to send a contract to the first user.

albertchae:
## Summary of the problem

Fixes https://github.com/hackclub/hcb/issues/10738

When creating a new event in the admin flow, invited users would not have an associated contract created. This PR fixes that.

## Describe your changes

**Best reviewed by commits**

- extract a partial to ask for signee + cosigner + onboarding videos from regular invite flow and share it with admin flow
- currently this has is_signee default to true since that was the existing behavior for admins, we should revisit if that's what we want for regular invites too. It's slightly awkward to set the default due to the partial being shared now.

small QOL improvement to hide the cosigner field if the user is not going to be a signee, as it's not relevant. This will still get sent to the backend if a user filled it out but then flipped the `is_signee` flag off, but the backend correctly ignores it.

We might want to fold in contract creation into the OrganizerPositionInvite service somehow since they seem related.

### DIFF
--- a/app/javascript/controllers/form_hide_field_controller.js
+++ b/app/javascript/controllers/form_hide_field_controller.js
@@ -14,10 +14,6 @@ export default class extends Controller {
   toggle() {
     const selectedValue = this.selectTarget.value
 
-    if (selectedValue === this.conditionValue) {
-      this.fieldTarget.style.display = 'none'
-    } else {
-      this.fieldTarget.style.display = 'block'
-    }
+    this.fieldTarget.hidden = selectedValue === this.conditionValue
   }
 }
--- a/app/javascript/controllers/form_hide_field_controller.js
+++ b/app/javascript/controllers/form_hide_field_controller.js
@@ -1,0 +1,23 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
+  static values = {
+    condition: String,
+  }
+
+  static targets = ['field', 'select']
+
+  connect() {
+    this.toggle()
+  }
+
+  toggle() {
+    const selectedValue = this.selectTarget.value
+
+    if (selectedValue === this.conditionValue) {
+      this.fieldTarget.style.display = 'none'
+    } else {
+      this.fieldTarget.style.display = 'block'
+    }
+  }
+}
--- a/app/views/organizer_position_invites/_admin_signee_check.html.erb
+++ b/app/views/organizer_position_invites/_admin_signee_check.html.erb
@@ -1,3 +1,5 @@
+<%# locals: (form:, default_selected: "") %>
+
 <div class="field field--select w-100"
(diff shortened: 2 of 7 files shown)


app/javascript/controllers/form_hide_field_controller.js
```diff
@@ -0,0 +1,23 @@
...
+    this.toggle()
+  }
+
+  toggle() {
+    const selectedValue = this.selectTarget.value
+
+    if (selectedValue === this.conditionValue) {
+      this.fieldTarget.style.display = "none"
+    } else {
+      this.fieldTarget.style.display = "block"
```
garyhtou: Instead of setting it to block, let's store whatever the display was on controller connect, then use that value

app/views/organizer_position_invites/_admin_signee_check.html.erb
```diff
@@ -0,0 +1,27 @@
+<div class="field field--select w-100"
```
garyhtou, line 3: Let's add a strict locals comment

https://blog.appsignal.com/2024/09/11/ruby-on-rails-7-1-partial-strict-locals-and-their-gotchas.html

garyhtou:
On the admin new organization page, let's default to contract signee true. On the org team page, deafult to false.

app/views/organizer_position_invites/_admin_signee_check.html.erb
albertchae: cool feature, done

app/javascript/controllers/form_hide_field_controller.js
albertchae: I instead decided to toggle `hidden` instead which seemed simpler than storing and restoring the display

Luke-Oldenburg:
- reader / member / manager does not affect visibility of the contract tool
- select dropdown has 3 options, one of them being to select despite being a binary option.
  - the select option shows the email box and should probably be removed.
  - this same behavior occurs on the create event page

garyhtou (approved):
we'll come back to fix those

albertchae:
> * reader / member / manager does not affect visibility of the contract tool
> 
>     * select dropdown has 3 options, one of them being to select despite being a binary option.
>       
>       * the select option shows the email box and should probably be removed.
>       * this same behavior occurs on the create event page

Addressing these in https://github.com/hackclub/hcb/pull/11011

PR #10989, opened 2025-07-17
Minor fix to arg parsing in docker_start.sh script

atikinf:
Prevent --with-solargraph string replacement from leaving "". `docker compose run` interprets the quotes as a command - which then fails

## Summary of the problem
Currently, if running `./docker_start.sh --with-solargraph`, the string replacement will replace `--with-solargraph` w empty quotes:`docker compose run ... ''`. This causes the script to fail since `run` interprets the quotes as a command.

Might be machine specific, but it's a minor tweak. Changing the replacement command to replace it w nothing also doesn't work for some reason : P

## Describe your changes
Fix the string replacement by iterating over the args w an array instead of `{@/.../}`.


sampoder (approved):
thanks - appreciate the fix!

PR #10990, opened 2025-07-17
Fix search clearing other filters

### LINKED ISSUE #10963: [Event Filter] Search drops other filters
See https://github.com/hackclub/hcb/pull/10937#pullrequestreview-3025624079

manuthecoder:
### Closes https://github.com/hackclub/hcb/pull/10937#pullrequestreview-3025624079

This pull request updates the search form in the `app/views/events/_search.html.erb` file to preserve additional query parameters when submitting the form. 

Key change:

* Added a loop to include hidden fields for all parameters except `:q`, `:controller`, and `:action`, ensuring that these parameters are retained during form submissions.

### DIFF
--- a/app/views/events/_search.html.erb
+++ b/app/views/events/_search.html.erb
@@ -6,4 +6,7 @@
       <%= inline_icon "view-close", size: 24 %>
     <% end %>
   <% end %>
+  <% params.except(:q, :controller, :action).each do |key, value| %>
+    <%= hidden_field_tag key, value %>
+  <% end %>
 </div>


app/views/events/_search.html.erb
```diff
@@ -6,4 +6,7 @@
       <%= inline_icon "view-close", size: 24 %>
     <% end %>
   <% end %>
+  <% params.except(:q, :controller, :action).each do |key, value| %>
+    <%= hidden_field_tag key, value %>
+  <% end %>
```
davidcornu, line 11: **concern**: I'm quite surprised we need this. I would have expected the `<form>`'s `action` to already contain the existing params so all the fields do is override the new ones

manu-gurudath: yea it's a little weird but date and amount range filters have the same lines of code as well (since they use forms)

sampoder (approved):
trying looking for a cleaner way on the internet and tested a bit locally - I think this is the way to go about this? approving to fix the bug!

sampoder:
https://stackoverflow.com/questions/63620943/how-to-create-a-form-and-keep-the-existing-params

PR #10995, opened 2025-07-18
[Stripe Cards] swap expiration date and CVV

24c02:
most checkout forms ask for card number, then exp, then CVV – having to copypaste #​1 then #​3 then #​2 when using a virtual card brings me mental anguish...

=>


Luke-Oldenburg:
I didn't know this! Thank you! Could you replicate this to `card_grants/card_details` too? I'm going to combine the two into a partial sometime soon.

24c02:
good catch, done!

YodaLightsabr (approved):
Thank you for this!

PR #10997, opened 2025-07-18
Refactor mailer set_emails into an event method

polypixeldev:
## Summary of the problem

The code to get a list of emails of all users in an event is replicated across several mailers, and some don't take into account the team contact email if it is set.

## Describe your changes

Added a method on `Event` that gets a list of all event users along with the team contact email, and used it everywhere mailers need it.


davidcornu (approved):
Nice clean-up!

PR #11000, opened 2025-07-18
[Admin] No longer need contract signee when creating org from Airtable

Sarvesh-Mk:
## Summary of the problem

When creating an organization from Airtable, it says that :cosigner_email and :onboarding_videos are not passed in:
image

## Describe your changes

Added default values for cosigner_email and onboarding videos to nil and false.

I can't test this in dev mode.

### DIFF
--- a/app/services/event_service/create.rb
+++ b/app/services/event_service/create.rb
@@ -4,8 +4,8 @@ module EventService
   class Create
     def initialize(name:,
                    point_of_contact_id:,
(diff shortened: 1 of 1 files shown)


garyhtou (approved):

albertchae:
whoops missed this in https://github.com/hackclub/hcb/pull/10982, thanks for fixing @Sarvesh-Mk

PR #11002, opened 2025-07-18
[Admin] Separate admin user details into a turbo frame

polypixeldev:
## Summary of the problem

Some user admin pages on production just don't load because of all of the data that has to be fetched and rendered on that page, but only the first couple admin tools are usually used.

## Describe your changes

Separates the slow-loading data on the user admin page into a lazy-loaded turbo frame so that it doesn't hold back the other parts of the page.


app/controllers/users_controller.rb
```diff
@@ -211,9 +211,12 @@ def disable_backup_codes
 
   def edit_admin
     @user = params[:id] ? User.friendly.find(params[:id]) : current_user
-    set_onboarding
-    show_impersonated_sessions = auditor_signed_in? || current_session.impersonated?
-    @sessions = show_impersonated_sessions ? @user.user_sessions : @user.user_sessions.not_impersonated
```
davidcornu, line 216: **note**: did some quick checking to make sure we aren't using these. I think we're ok but hard to be 100% sure.

app/policies/user_policy.rb
```diff
@@ -61,6 +61,10 @@ def edit_admin?
     user.auditor? || (record == user && user.admin_override_pretend?)
   end
 
+  def admin_details?
+    user.auditor?
```
davidcornu, line 65: **concern**: Why is this different than `edit_admin?` (`user.auditor? || (record == user && user.admin_override_pretend?)`) ?

app/controllers/users_controller.rb
polypixeldev: I also didn't see any usages - for some reason onboarding and sessions are loaded on all the user settings methods, even though onboarding is only used in `edit` and sessions is only used in `edit_security`. I'll make a separate PR to double-check and remove the rest of the unnecessary variables

app/policies/user_policy.rb
polypixeldev: When an admin is pretending not to be an admin, the only part of the admin page they see is the checkbox to stop pretending to be an admin. Since they shouldn't see any of the user details, we don't need to check if they're pretending not to be an admin

davidcornu (approved):

PR #11003, opened 2025-07-18
Fix announcement card when date object is `nil`

manuthecoder:
image

This pull request includes a minor update to the `app/views/announcements/_announcement_card.html.erb` file. The change ensures that the `strftime` method is only called if the date object is not `nil`, improving code safety.

* [`app/views/announcements/_announcement_card.html.erb`](diffhunk://#diff-5b4ff73e8fc14c2fee0863612408c16cf494ff6e79fd3d91df03b82532ec925dL25-R25): Added safe navigation (`&.`) to the `strftime` method call to handle potential `nil` values for `created_at` or `published_at`.

### DIFF
--- a/app/views/announcements/_announcement_card.html.erb
+++ b/app/views/announcements/_announcement_card.html.erb
@@ -22,7 +22,7 @@
     <% end %>
   </div>
   <p class="italic my-0">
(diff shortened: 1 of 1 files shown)


manu-gurudath:
update: idk if this relevant anymore

brarnaudovski:
Unsure why this is closed, but I'm having the same issue.

Trying to start with a contribution to this project, and this is one of the issues after running the app for the first time.

Maybe it's not relevant when using production data, but it seems some seed data is missing, so this issue is mitigated.

PR #11011, opened 2025-07-20
Follow up to UX issues with contract signee invitation

albertchae:
## Summary of the problem

Follow up to UX issues noted in https://github.com/hackclub/hcb/pull/10982#issuecomment-3089548257

- reader / member / manager does not affect visibility of the contract tool
- select dropdown has 3 options, one of them being to select despite being a binary option.
    - the select option shows the email box and should probably be removed
    - this same behavior occurs on the create event page

## Describe your changes

> reader / member / manager does not affect visibility of the contract tool

Now we only show the contract sub form if the invited user is a manager. This raises a small issue where you could have Manager selected, then set them to be a contract signee, then change the role to Reader or Member and try to submit. This would do the right thing and [result in an error](https://github.com/hackclub/hcb/blob/9c9a8cbca4c8b5b512e998d07b1597030e8f9cd9/app/models/concerns/organizer_position/has_role.rb#L21-L26), but I think is confusing UX, so I changed the backend to only honor the `is_signee` field when Manager is the selected role.

Screenshot 2025-07-20 at 9 49 31 AM

An alternative would have been to clear the contract signee form in the frontend when switching roles, but I think sanitizing params in the backend was easier in this case.

I didn't realize we were using alpinejs for this form, we might want to consider removing the stimulus controller I added for hiding and using alpinejs for that too.

> the select option shows the email box and should probably be removed

We now only show the cosigner option when "Contract Signee" is selected. This was implemented by changing `form_hide_field_controller` to take the option we want to show the conditional field on, not the ones we want to hide on.

> select dropdown has 3 options, one of them being to select despite being a binary option.

This one I left as is for now as it might be intentional/desired behavior.
- On the regular organizer invite flow, we are not setting a default option which would prompt the user to pick one.
- In create event, we default to contract signee since the first member of an event is probably a signee. Arguably we could remove the prompt option here but that seems like unnecessary complexity.

Screenshot 2025-07-20 at 10 03 21 AM

@garyhtou can you confirm whether this is fine, or should we consider setting a default option for the regular invite flow? This would make it easier to remove the prompt option in both places we are using this


garyhtou:
That behavior is perfect! Thank you

Luke-Oldenburg (approved):
lgtm!

Luke-Oldenburg (changes requested):
I resolved the merge conflicts but could you please use alpine instead of stimulus?

sampoder:
Why would we not use Stimulus?

Luke-Oldenburg:
> Why would we not use Stimulus?

We were already using alpine elsewhere here.

PR #11012, opened 2025-07-20
Exempt salary accounts from card locking

sampoder:
Context:

Screenshot 2025-07-20 at 11 32 19 PM

This will make approvals for gap years with locked cards but using a salary account slower but it should be within the time we have + is better than declining.


garyhtou (approved):
LGTM! In the future, we'll want to move this logic into the Event Plan itself

PR #11014, opened 2025-07-21
[DX] Only run `suggest_lint_changes` for pull requests

davidcornu:
## Summary of the problem

- We are currently using `if: ${{ github.event.pull_request }}` to conditionally run `parkerbxyz/suggest-changes` but I can't find any evidence that `github.event.pull_request` is still supported
- Our `if` is on the very last step of the job so we still run all the previous steps before bailing

## Describe your changes

- Move the `if` to the job definition
- Use the documented `github.event_name == 'pull_request'` (https://docs.github.com/en/actions/reference/contexts-reference#example-usage-of-the-github-context)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,7 @@ jobs:
   suggest_lint_changes:
     name: 'Suggest lint changes'
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


garyhtou (approved):
nice!

PR #11028, opened 2025-07-21
Add sudo mode feature preview

davidcornu:
## Summary of the problem

We want to allow users to opt-in to testing sudo mode before we roll it out to everyone.

## Describe your changes

- Add the feature flag to `FeaturesController::FEATURES`
- Add a new feature preview to the list
    image
- Require sudo mode to disable sudo mode


app/controllers/features_controller.rb
```diff
@@ -47,6 +48,11 @@ def disable_feature
             end
     feature = params[:feature]
     authorize actor
+
+    if feature == "sudo_mode_2015_07_21"
+      return unless enforce_sudo_mode # rubocop:disable Style/SoleNestedConditional
+    end
```
sampoder, line 54: just to sanity check - this is requiring people to pass sudo mode checks to enable sudo mode?

garyhtou: If so, i'm not sure if this works as expected since `enforce_sudo_mode` and `sudo_mode?` will always return true because of:

```ruby
def sudo_mode?
    return true unless Flipper.enabled?(:sudo_mode_2015_07_21, user)
    # ...
```

garyhtou (approved):
Nice!

app/controllers/features_controller.rb
davidcornu: > just to sanity check - this is requiring people to pass sudo mode checks to enable sudo mode?

To disable. This is in the `disable_feature` action (see the accompanying test https://github.com/hackclub/hcb/pull/11028/files#diff-dd99dd0a50f80265ce784f671ad5efb0e6ad9da9ae448b4711f5e1f5e72f945dR10)

davidcornu:
Rebase incoming

PR #11031, opened 2025-07-21
fix accounts_inactive? to not fail on nil users count

rluodev:
## Summary of the problem

we previously failed if the users came back as nil

## Describe your changes

we now use safe navigation to avoid this issue.


app/models/g_suite.rb
```diff
@@ -141,9 +141,9 @@ def previously_verified?
   def accounts_inactive?
     begin
       res = Partners::Google::GSuite::Users.new(domain:).run
-      res_count = res.users.count
+      res_count = res.users ? res.users.count : 0
```
sampoder:
```suggestion
      res_count = res.users&.count || 0
```

app/models/g_suite.rb
```diff
@@ -141,9 +141,9 @@ def previously_verified?
   def accounts_inactive?
     begin
       res = Partners::Google::GSuite::Users.new(domain:).run
-      res_count = res.users.count
+      res_count = res.users&.count || 0
       inactive_accounts = []
-      res.users.each do |user|
+      res&.users.each do |user|
```
rluodev:
```suggestion
      res&.users&.each do |user|
```

YodaLightsabr (approved):

PR #11033, opened 2025-07-21
[Announcement] Fix prompting when there is no event mission

polypixeldev:
## Summary of the problem

Users were being prompted to make an announcement when updating their event for the first time since the description was being changed from `nil` to `""`

## Describe your changes

Added a check to make sure the mission is not empty when prompting.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -334,7 +334,7 @@ class EventsController < ApplicationController
         end
 
 
-        if @event.description_previously_changed? && !@event.description.empty?
+        if @event.description_previously_changed? && @event.description.present?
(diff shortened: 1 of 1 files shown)


app/controllers/events_controller.rb
```diff
@@ -334,7 +334,7 @@ def update
         end
 
 
-        if @event.description_previously_changed?
+        if @event.description_previously_changed? && !@event.description.empty?
```
```suggestion
        if @event.description_previously_changed? && @event.description.present?
```
Can you check if this works? I prefer avoiding negation

polypixeldev: that works, thanks!

davidcornu (approved):

PR #11034, opened 2025-07-21
Disable HQ perks

### LINKED ISSUE #11018: Disable Perks for all HQ Orgs

alexbluo:
Resolves #11018


sampoder (changes requested):
Can we do this at a policy level and just not show this page to anyone not eligble?

leowilkin:
> Can we do this at a policy level and just not show this page to anyone not eligble?

Zach wanted us to 'dogfood' the perks page - this should be _visible_ to HQ orgs, however buttons should be disabled.

alexbluo:
changing to allow daydream on premium - dont merge yet

sampoder (dismissed):
If we're disabling all perks then we can set `@perks_available` to false then - right?

alexbluo:
only want to disable for orgs on hq/hq affiliated plan.

daydream will allegedly be zero percent plans so no change needed—please merge

YodaLightsabr (changes requested):
See Sam's comment—thanks!

app/views/events/promotions.html.erb
```diff
@@ -51,7 +51,7 @@
   </div>
 <% end %>
 
-<% if @event.hackathon? || @event.plan.is_a?(Event::Plan::HighSchoolHackathon) %>
+<% if (@event.hackathon? || @event.plan.is_a?(Event::Plan::HighSchoolHackathon)) %>
```
YodaLightsabr:
```suggestion
<% if @event.hackathon? || @event.plan.is_a?(Event::Plan::HighSchoolHackathon) %>
```

YodaLightsabr (approved):

YodaLightsabr (approved):

PR #11035, opened 2025-07-21
[Announcement] Migrate block JSON types to use Ruby identifiers

polypixeldev:
## Summary of the problem

We shouldn't have two different types for each block - changing all block types to be the class identifier is clearer.

## Describe your changes

Changed the types used in the ProseMirror JSON to use Ruby class identifier strings, and added a job to migrate existing content over.

> [!WARNING]
> The job `OneTimeJobs::MigrateBlockTypes` should be run once this is deployed.


sampoder (approved):
lgtm! will run one time job when this deploys

PR #11036, opened 2025-07-21
[Announcement::Block::DonationGoal] Save event donation goal on block create

polypixeldev:
## Summary of the problem

The donation goal block for announcements was not saving the ID of the donation goal is was referencing, so if it was refreshed after a new donation goal had been created, it would update to a different goal.

## Describe your changes

Saves the ID of the current donation goal in the parameters on save, and references it if it exists.

### DIFF
--- a/app/models/announcement/block/donation_goal.rb
+++ b/app/models/announcement/block/donation_goal.rb
@@ -25,9 +25,7 @@ class Announcement
   class Block
     class DonationGoal < ::Announcement::Block
       before_create do
-        if self.parameters["goal"].nil?
-          self.parameters["goal"] = announcement.event.donation_goal.id
-        end
+        goal_param
 
         self
       end
@@ -35,7 +33,7 @@ class Announcement
       def render_html(is_email: false)
         goal = nil
         begin
-          goal = Donation::Goal.find(parameters["goal"])
+          goal = Donation::Goal.find(goal_param)
         rescue ActiveRecord::RecordNotFound
           goal = announcement.event.donation_goal
         end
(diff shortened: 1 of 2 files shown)


app/models/announcement/block/donation_goal.rb
```diff
@@ -24,13 +24,31 @@
 class Announcement
   class Block
     class DonationGoal < ::Announcement::Block
+      before_create do
+        goal_param
+
+        self
```
davidcornu: **concern**: Not entirely sure what's going on here

- `goal_param` doesn't seem to have any externally-observable impact
- why return `self`?


Also, `before_create` can just take the method name as a symbol

```ruby
before_create :goal_param
```

app/models/announcement/block/donation_goal.rb
```diff
@@ -24,13 +24,31 @@
...
+      end
+
       def render_html(is_email: false)
-        goal = announcement.event.donation_goal
+        goal = nil
+        begin
+          goal = Donation::Goal.find(goal_param)
+        rescue ActiveRecord::RecordNotFound
+          goal = announcement.event.donation_goal
+        end
```
davidcornu: **suggestion**: This can be simplified quite a bit

```suggestion
        goal = Donation::Goal.find_by(id: goal_param) 
        goal ||= announcement.event.donation_goal
```

app/models/announcement/block/donation_goal.rb
polypixeldev: thank you for bringing that up! I had made a typo in `goal_param` and had thought that `before_create` needed to return self for changes to be saved

app/models/announcement/block/donation_goal.rb
```diff
@@ -24,13 +24,23 @@
 class Announcement
   class Block
     class DonationGoal < ::Announcement::Block
+      before_create :goal_param
+
       def render_html(is_email: false)
-        goal = announcement.event.donation_goal
+        goal = Donation::Goal.find_by(id: goal_param)
```
davidcornu: **concern**: We should scope this to the event rather than allow any goal to be rendered

```suggestion
        goal = Donation::Goal.find_by(event: announcement.event, id: goal_param)
```

davidcornu (approved):

PR #11038, opened 2025-07-21
[Announcement] Add validation for unauthorized HCB code block

polypixeldev:
## Summary of the problem

We're not currently doing any validations on when blocks are saved, and leaving checking that any parameters are valid until we render the block. Now that blocks are in the database, we can easily use validations to not save any invalid data in the first place.

## Describe your changes

Added a validation to the HCB code block that makes sure that it is in the same event as the announcement the block is a part of, and alerts it to the user with a browser alert (this will be changed to something nicer in a future PR). The existing checks in `render` are not removed since invalid data could still be in the database or be added with a Rails method that does not check validations.

### DIFF
--- a/app/controllers/announcements/blocks_controller.rb
+++ b/app/controllers/announcements/blocks_controller.rb
@@ -9,10 +9,8 @@ module Announcements
 
       authorize block, policy_class: Announcement::BlockPolicy
 
-      begin
-        block.save!
-      rescue ActiveRecord::RecordInvalid
-        return render json: { errors: block.errors.map { |error| error.full_message } }, status: :bad_request
+      unless block.save
+        return render json: { errors: block.errors.map(&:full_message) }, status: :bad_request
       end
 
       render json: { id: block.id, html: block.rendered_html }
--- a/app/controllers/announcements/blocks_controller.rb
+++ b/app/controllers/announcements/blocks_controller.rb
@@ -12,7 +12,7 @@ module Announcements
       begin
         block.save!
       rescue ActiveRecord::RecordInvalid
-        return render json: { errors: block.errors.map { |error| error.full_message } }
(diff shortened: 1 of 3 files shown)


app/controllers/announcements/blocks_controller.rb
```diff
@@ -9,7 +9,11 @@ def create
 
       authorize block, policy_class: Announcement::BlockPolicy
 
-      block.save!
+      begin
+        block.save!
+      rescue ActiveRecord::RecordInvalid
+        return render json: { errors: block.errors.map { |error| error.full_message } }, status: :bad_request
+      end
```
garyhtou:
```suggestion
      unless block.save
        return render json: { errors: block.errors.map(&:full_message) }, status: :bad_request
      end
```

app/models/announcement/block/hcb_code.rb
```diff
@@ -34,6 +36,16 @@ def render_html(is_email: false)
         Announcements::BlocksController.renderer.render partial: "announcements/blocks/hcb_code", locals: { hcb_code:, event: announcement.event, is_email:, block: self }
       end
 
+      private
+
+      def hcb_code_in_event
+        hcb_code = ::HcbCode.find_by_hashid(parameters["hcb_code"])
+
+        unless hcb_code&.event == announcement.event
```
garyhtou:
```suggestion
        if hcb_code.nil? || hcb_code.event != announcement.event
```

app/models/announcement/block/hcb_code.rb
```diff
@@ -34,6 +36,16 @@ def render_html(is_email: false)
...
+          errors.add(:base, "You are not authorized to insert this transaction")
```
garyhtou:
```suggestion
          errors.add(:base, "Transaction can not be found.")
```

sampoder (approved):
tested locally - works! lgtm

PR #11040, opened 2025-07-21
[Announcement] Drop HTML columns from announcements

polypixeldev:
## Summary of the problem

These columns are not being used anymore, and can be safely dropped now that the migration to JSON and blocks is complete.

## Describe your changes

Drops the columns and adds the ignored columns fields as recommended by Strong Migrations


davidcornu (approved):

davidcornu:
@polypixeldev happy to ship this as soon as the merge conflict is resolved

PR #11044, opened 2025-07-22
Fix inconsitencies between `db/schema.rb` and production

davidcornu:
## Summary of the problem

On a freshly setup database, the schema doesn't match the model annotations. Upon further investigation it appears `db/schema.rb` was erroneously altered so as to no longer match the actual state of production.

## Describe your changes

- Remove default value on `referral_programs.login_text_color`
- Add missing unique index on `(raffles.program, raffles.user_id)`

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1562,6 +1562,7 @@ ActiveRecord::Schema[7.2].define(version: 2025_07_21_124328) do
     t.string "program", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.index ["program", "user_id"], name: "index_raffles_on_program_and_user_id", unique: true
   end
 
   create_table "raw_column_transactions", force: :cascade do |t|
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1790,7 +1790,7 @@ ActiveRecord::Schema[7.2].define(version: 2025_07_21_124328) do
     t.string "background_image_url"
     t.string "login_header_text"
     t.text "login_body_text"
-    t.string "login_text_color", default: "#ffffff"
+    t.string "login_text_color"
   end
 
   create_table "reimbursement_expense_payouts", force: :cascade do |t|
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -1790,7 +1790,7 @@
     t.string "background_image_url"
     t.string "login_header_text"
     t.text "login_body_text"
-    t.string "login_text_color", default: "#ffffff"
+    t.string "login_text_color"
```
davidcornu, line 1794: This was introduced in https://github.com/hackclub/hcb/pull/10953/files#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3. 

For some reason

- The migration does not contain a default: https://github.com/hackclub/hcb/blob/1d799497ece0e4adafd8fd6065ec47e44eb47e2c/db/migrate/20250715203909_add_background_image_url_and_login_text_to_referral_program.rb#L6
- The schema has a default: https://github.com/hackclub/hcb/blob/1d799497ece0e4adafd8fd6065ec47e44eb47e2c/db/schema.rb#L1793
- Production does not having a default: CleanShot 2025-07-22 at 10 56 28@2x

db/schema.rb
```diff
@@ -1562,6 +1562,7 @@
     t.string "program", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.index ["program", "user_id"], name: "index_raffles_on_program_and_user_id", unique: true
```
davidcornu, line 1565: This was removed in https://github.com/hackclub/hcb/pull/10830/files#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3L1520 but is present (and necessary) in production:

CleanShot 2025-07-22 at 11 11 02@2x

garyhtou (approved):
Thank you so much for this 🙏

db/schema.rb
YodaLightsabr: Apologies for this one! Thank you for making the fix.

davidcornu: No worries! Happens to the best of us

PR #11045, opened 2025-07-22
[DX] Add a `/schema` endpoint

davidcornu:
## Summary of the problem

Issues like https://github.com/hackclub/hcb/pull/11044 are annoying to diagnose as they require us to use Blazer to query the state of the production schema and compare it to the `db/schema.rb` format.

## Describe your changes

Adds a `/schema` endpoint which effectively does `rails db:schema:dump`, letting us copy and paste the production state and immediately see a local diff.


sampoder (approved):
very neat - nice!

PR #11048, opened 2025-07-22
[Sudo mode] Add client-side support for WebAuthn

davidcornu:
## Summary of the problem

In #10671 I added server-side support for WebAuthn but intentionally omitted the client-side logic so I could address it in a follow-up.

## Describe your changes

Introduces a new stimulus controller that is bound when using the WebAuthn reauthentication method


YodaLightsabr (approved):
Looks good!

PR #11049, opened 2025-07-22
[Announcement::Block::DonationSummary] Add end date and fallbacks to donation summary block

polypixeldev:
## Summary of the problem

End dates for donation summaries are needed to properly support monthly announcements and prevent it from updating on an accidental refresh.

## Describe your changes

Add the end date parameter to the donation summary block, and refactor a bit to be better about default values for block parameters.


sampoder (approved):
lgtm - thank you!

PR #11053, opened 2025-07-22
Set turbo frame target to top on HCB code popovers

polypixeldev:
## Summary of the problem

https://hackclub.slack.com/archives/C068U0JMV19/p1753204464687409

Clicking on links inside HCB code popovers would be captured by the turbo frame instead of navigating the page, resulting in either an infinite loading state or "Content missing"

## Describe your changes

Added `target="_top"` to the turbo frame to disable capturing links.


sampoder (approved):
lgtm!

PR #11054, opened 2025-07-22
[Announcement] Remove unnecessary JS node attributes

polypixeldev:
## Summary of the problem

Since all data is now stored in the `parameters` part of blocks, we no longer need the attributes in the JS editor.

## Describe your changes

Removed the unused attributes from the JS node configuration


sampoder (approved):
lgtm!

PR #10973, opened 2025-07-16
[Announcement] Policy Fixes

Luke-Oldenburg:
- manual 404 -> policy restriction (we were exposing slugs because this caused unexpected behavior)
- make follow policy stricter
- remove unused `shipping?` policy from event
- hide invoices tab in nav based on policy
- use strict locals in follow button
- hide follow button if no announcements on private event

PR #11056, opened 2025-07-22
[Announcement] Add Atom feeds!

24c02:
hokay so:
we got atom xml:

it validates:

if you follow/organize the event you get this little guy:

it renders beautifully in your feed reader of choice:

what more could you want?

### DIFF
--- a/app/policies/event_policy.rb
+++ b/app/policies/event_policy.rb
@@ -85,7 +85,7 @@ class EventPolicy < ApplicationPolicy
   end
 
   def feed?
-    is_public
+    announcement_overview?
   end
 
   def emburse_card_overview?
--- a/app/policies/event_policy.rb
+++ b/app/policies/event_policy.rb
@@ -84,6 +84,10 @@ class EventPolicy < ApplicationPolicy
     true
   end
 
+  def feed?
+    is_public
+  end
+
   def emburse_card_overview?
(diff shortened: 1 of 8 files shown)


app/policies/event_policy.rb
```diff
@@ -84,6 +84,10 @@ def announcement_overview?
     true
   end
 
+  def feed?
+    is_public
```
24c02: is this right? can you view announcements on non-transparent orgs? i'm dunno

polypixeldev: you can view announcements on non-transparent orgs - the way that it should work is that if it has announcements it should respond with the feed, but if it is non-transparent and there are no announcements then it should respond with 404 to not expose the slug

app/policies/event_policy.rb
```diff
@@ -84,6 +84,10 @@ def announcement_overview?
...
+  end
```
polypixeldev:
```suggestion
  def feed?
    announcement_overview?
  end
```
I think this is fine to just make it the same as `announcement_overview?`, also since there are changes to that policy in #10973

24c02: yep, with those changes i'm cool with that!

garyhtou (approved):

PR #11061, opened 2025-07-22
Integrate web mocking library

### LINKED ISSUE #10755: [DX] Stub out remote calls in the test suite

brarnaudovski:
# WIP

Fixes #10755 Closes #10829 

## Summary of the problem

Summary from the issue #10755

> There are several tests in our test suite making real API calls to Stripe and other services. We should consider stubbing these out with a gem like https://rubygems.org/gems/vcr or https://rubygems.org/gems/webmock so they're easier to run.
> 
> This might also let us reduce the number of credentials needed (or completely eliminate them) so anyone outside of HCB can run things.
> 

## Describe your changes
`webmock` integration as library used in test group that mocks web requests.

integration of the library involves fixing the existing web requests called in the specs. Otherwise, specs that make real web requests fail with `webmock` exception.
Following is all the spec files that need fixing:
- [x] spec/controllers/ach_transfers_controller_spec.rb
- [x] UserService::SyncWithLoops
- [ ] others tba

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -151,6 +151,7 @@ end
 group :test do
   gem "factory_bot_rails" # Test data
   gem "simplecov", require: false # Code coverage
+  gem "webmock", "~> 3.25"
 end
 
 group :development, :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -238,6 +238,9 @@ GEM
       unaccent (~> 0.3)
     country_select (8.0.2)
       countries (~> 5.0)
+    crack (1.0.0)
+      bigdecimal
+      rexml
     crass (1.0.6)
     cronex (0.15.0)
       tzinfo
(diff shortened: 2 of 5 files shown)


davidcornu:
@brarnaudovski this is great! You might not actually need the Loops stuff now that https://github.com/hackclub/hcb/pull/11140 has shipped.

davidcornu:
Closing this on account of https://github.com/hackclub/hcb/pull/11751.

Thanks for having taken a stab at this @brarnaudovski ❤️

PR #11063, opened 2025-07-23
Include local amount on HCB code partial

YodaLightsabr:
## Summary of the problem
Right now, it's hard to see how much a card transaction is for in the local currency. We show the merchant currency, but not the amount. This can make it hard to pair receipts or understand finances in a local context.

## Describe your changes
This PR adds a line to the Stripe card partial on HCB codes to include the local currency, if applicable. It references a new method on the Raw Stripe Transaction. This makes it easier for users to see how much a transaction was for in the original currency before it was converted to USD by Visa.

cc @leowilkin

### DIFF
--- a/app/models/raw_stripe_transaction.rb
+++ b/app/models/raw_stripe_transaction.rb
@@ -24,7 +24,7 @@ class RawStripeTransaction < ApplicationRecord
 
   include ActiveSupport::NumberHelper
 
-  ZERO_DECIMAL_CURRENCIES = %w[BIF CLP DJF GNF JPY KMF KRW MGA PYG RWF UGX VND VUV XAF XOF XPF]
+  ZERO_DECIMAL_CURRENCIES = %w[BIF CLP DJF GNF JPY KMF KRW MGA PYG RWF UGX VND VUV XAF XOF XPF].freeze
 
   def memo
     @memo ||= stripe_transaction.dig("merchant_data", "name")
--- a/app/models/raw_stripe_transaction.rb
+++ b/app/models/raw_stripe_transaction.rb
@@ -22,6 +22,10 @@ class RawStripeTransaction < ApplicationRecord
   has_many :hashed_transactions
   has_one :canonical_transaction, as: :transaction_source
 
+  include ActiveSupport::NumberHelper
+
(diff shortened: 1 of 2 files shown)


sampoder:
Ian and I chatted about this on Slack but this already exists:

Screenshot 2025-07-22 at 10 40 19 PM

app/models/raw_stripe_transaction.rb
```diff
@@ -34,6 +38,19 @@ def refund?
...
+    amount = stripe_transaction["merchant_amount"].abs
+
+    return nil if currency.nil? || currency == "USD"
+
+    if ZERO_DECIMAL_CURRENCIES.include?(currency)
+      "#{number_to_delimited(amount)} #{currency}"
+    else
+      "#{number_to_currency(BigDecimal((amount || 0).to_s) / 100, unit: "")} #{currency}"
+    end
+  end
```
davidcornu, line 52: **suggestion**: Any reason we can't leverage the `money` gem here? If has built-in support for subunit handling

```suggestion
  def local_amount
    currency = stripe_transaction["merchant_currency"].upcase
    amount = stripe_transaction["merchant_amount"].abs

    return nil if currency == "USD"

    Money.new(amount, currency).format
  end
```

app/models/raw_stripe_transaction.rb
```diff
@@ -34,6 +38,19 @@ def refund?
     stripe_transaction["type"] == "refund"
   end
 
+  def local_amount
+    currency = stripe_transaction["merchant_currency"]&.upcase
+    amount = stripe_transaction["merchant_amount"].abs
+
+    return nil if currency.nil? || currency == "USD"
```
davidcornu, line 45: **suggestion**: as far as I can tell, `merchant_currency` and `merchant_amount` are always present (both in the API and in our database)

CleanShot 2025-07-30 at 14 34 57@2x

https://docs.stripe.com/api/issuing/transactions/object?api-version=2024-12-18.acacia#issuing_transaction_object-merchant_amount

```suggestion
    currency = stripe_transaction.fetch("merchant_currency").upcase
    amount = stripe_transaction.fetch("merchant_amount").abs

    return nil if currency == "USD"
```

sampoder:
Ian and talked about this on Slack but also putting this here, we already
have this built out

davidcornu:
@sampoder ah ok so we can close this?

PR #11065, opened 2025-07-23
[Documents] Can now categorize documents when uploading

### LINKED ISSUE #10716: [Admin] SubDocument
### Describe the problem

I wish there was a subdocument for "State Tax Exemptions". We're starting to get emails coming in asking for tax exemptions (mostly Robotics teams)

Fiscal orgs can view the states we have tax exemption for and in the process of getting

Sarvesh-Mk:
## Summary of the problem

Currently, there is no way to categorize documents on the documents tab and it makes the documents tab look very messy. Orgs are finding it hard to determine what document is used for what so it might be useful to have categories for them.
image
ew its so disorganized ^ 🤮 
## Describe your changes

Added an enum to store what category a document is part of. Also created a DB migration file to create a column on the table for this. Also added scopes to filter out by category.

image

Wow look at these categories ^ 🤩 
image
^ 😮 new sick dropdown to select categories when uploading a document (also seen in the edit page)

Fixes #10716


Sarvesh-Mk:
I know you might be thinking "What about preexisting documents?". Well almost every document on the documents tab is a common document which can be modified in the common documents section. In terms of organization specific documents, the only one that comes to my mind is fiscal sponsorship documents which are defaulted to general and are supposed to be general according to the attached issue.

app/views/documents/index.html.erb
```diff
@@ -21,8 +21,36 @@
...
+          <strong class="h3 block mt1 line-height-2">Fiscal Sponsorship Confirmation</strong>
+          <%= user_mention @event.point_of_contact %>
+        </li>
+      <% end %>
+    </div>
+
+    <h3> Nonprofit documents </h3>
+    <div class="grid grid-cols-3 gap-4">
+      <%= render partial: "document", collection: @active_common_documents.nonprofit_status %>
+    </div>
```
sampoder: What differentiates a general document from a nonprofit document?

db/migrate/20250721015626_create_document_categories.rb
YodaLightsabr, line 1: Let's rename this to `AddCategoryToDocuments` (`add_category_to_documents`)

app/views/documents/index.html.erb
Sarvesh-Mk: Its described in #10716 but I would assume documents like the statement of lost receipt is not a directly related to a non-profit document?

app/models/document.rb
```diff
@@ -47,6 +48,17 @@ class Document < ApplicationRecord
   validate :ensure_file_attached
 
   scope :common, -> { where(event_id: nil) }
+  scope :general, -> { where(category: :general) }
+  scope :nonprofit_status, -> { where(category: :nonprofit_status) }
+  scope :tax_exception, -> { where(category: :tax_exception) }
+  scope :forms, -> { where(category: :forms) }
```
YodaLightsabr: Great use of scopes! Thankfully, Rails takes care of all of these for us: https://api.rubyonrails.org/v5.1.7/classes/ActiveRecord/Enum.html

YodaLightsabr (dismissed):

app/models/document.rb
Sarvesh-Mk: Woah did not know that, removing the unnecessary scopes then!

db/migrate/20250721015626_create_document_categories.rb
Sarvesh-Mk: Done!

app/views/documents/index.html.erb
```diff
@@ -21,8 +21,36 @@
...
+          <div class="overflow-hidden" style="max-height: 250px;">
+            <%= image_tag event_fiscal_sponsorship_letter_url(event_id: @event.slug, format: "png"), width: 330, style: "max-width: 100%;" %>
+          </div>
+          <strong class="h3 block mt1 line-height-2">Fiscal Sponsorship Confirmation</strong>
+          <%= user_mention @event.point_of_contact %>
+        </li>
+      <% end %>
+    </div>
+
+    <h3> Nonprofit documents </h3>
```
sampoder:
```suggestion
    <h3>Nonprofit status</h3>
```

app/models/document.rb
```diff
@@ -48,6 +49,13 @@ class Document < ApplicationRecord
 
   scope :common, -> { where(event_id: nil) }
 
+  enum :category, {
+    general: 0,
+    nonprofit_status: 1,
+    tax_exception: 2,
```
sampoder:
```suggestion
    tax_exemption: 2,
```

sampoder (approved):
code lgtm! thank you Sarvesh

PR #11071, opened 2025-07-23
[Perks] Add Sticker Ninja expired perks

alexbluo:
## Summary of the problem

When I removed Sticker Ninja from perks to make way for Jukebox, I forgot to add it to expired perks.

## Describe your changes

image


YodaLightsabr (changes requested):
The code looks good, but please add a PR description as well as a screenshot. Thanks! (Slack me once you need a re-review)

PR #11073, opened 2025-07-23
[Announcement] Add new UI for block adding and editing

polypixeldev:
## Summary of the problem

The built-in browser modal for submitting data about blocks isn't great, and currently you can't edit the parameters of a block without removing it and adding it again. :(

## Describe your changes

This PR adds custom modals to add/edit the data of a block, for now all except `DonationGoal`. To allow this, creating new announcements using a modal has been removed, as modals cannot be easily layered. Clicking on the create button will just direct to the `announcements/new` page.

Some refactoring has also been done to the block database structure and Stimulus controllers to make rendering them in different contexts (like a turbo stream) easier and with less code.

https://github.com/user-attachments/assets/79c90201-3790-4ed7-9d8c-6a4a70bd3f9e


sampoder (changes requested):
this is a nice improvement - thank you! one thing I noticed was that the validations code from your other PR (https://github.com/hackclub/hcb/pull/11038/) doesn't seem to have carried over:

Screenshot 2025-07-26 at 12 40 18 PM

polypixeldev:
> this is a nice improvement - thank you! one thing I noticed was that the validations code from your other PR (#11038) doesn't seem to have carried over:

just merged main into this PR and added the validation to display in the block modal - thanks for reminding me!

sampoder (changes requested):
We should use the standard menu styles here:

Screenshot 2025-08-27 at 3 55 54 PM

Also, please use sentence case. Otherwise, lgtm. I think in the future it'd be nice to have a transaction picker for the transaction link page.

sampoder (approved):
thank you!

PR #11076, opened 2025-07-24
Redesign premium promotions to be more inline with our design system

sampoder:
@alexbluo I'd love you review here, new perk looks great! just I'm a little worried that we're veering too far from our design style / system.

Screenshot 2025-07-23 at 8 00 22 PM

Screenshot 2025-07-23 at 8 00 17 PM

these are my attempts - let me know what you think!


YodaLightsabr (approved):
LGTM

alexbluo:
Yea def agree, looks way better. Thanks for making this!

PR #11078, opened 2025-07-24
[Announcement] Add event creation as updated default in atom feed

polypixeldev:
## Summary of the problem

feed.atom currently throws an error if there are no announcements!

## Describe your changes

Adds `event.created_at` as the default `updated` field in the announcements atom feed


sampoder (approved):
lgtm!

PR #11080, opened 2025-07-24
V3 API improvements

24c02:
changes the V3 transparency API to:
expose the creator of `AchTransfer`s as a `User`
expose the actual `owner` of PublicActivity activities rather than only trying `#user`
and.....drumroll please:
correctly expose the linked HCB code on RawPendingStripeTransaction.create activities
(previously they were unlinked because rpst doesn't directly have a `local_hcb_code`, it's on the associated cpt)


3kh0:
image

sampoder (approved):
lgtm - thank you!

PR #11086, opened 2025-07-24
[Announcement] Add one time notice mailer

polypixeldev:
## Summary of the problem

A notice should be sent out to managers of organizations that have had monthly announcements automatically enabled to let them know about the feature.

## Describe your changes

Adds the notice email to the announcement mailer, and a one time job to send out the email.

image


app/jobs/one_time_jobs/send_announcement_notice.rb
```diff
@@ -0,0 +1,21 @@
...
+module OneTimeJobs
+  class SendAnnouncementNotice < ApplicationJob
+    def self.perform
+      # Rate limit is 14/s, but putting 12 here to be safe and allow for other emails to be sent
+      queue = Limiter::RateQueue.new(12, interval: 1)
+
+      Event.includes(:config).where(config: { generate_monthly_announcement: true }).find_each do |event|
+        monthly_announcement = event.announcements.monthly.last
+
+        if monthly_announcement.present?
```
davidcornu, line 12: **concern**: does this mean we'll keep sending these until they post an announcement?

polypixeldev: No, this was just meant to be a one time email - the check for monthly announcement is present is in case an event enabled generating monthly announcements after this month's job ran

sampoder (approved):
lgtm! are we confident these won't double send?

polypixeldev:
Not 100% - it's unlikely that they will ratelimit since when we run this job it will be the only mass email send running - however I can't be 100% sure that something else won't error and cause the job to re-run.

I'd like to get this sent out soon but it's better to wait until we have better error handling and ratelimiting for emails before running this, I'll be working on that next with the goal of getting it done to send this email out early next week

PR #11089, opened 2025-07-24
Add note about announcements visibility in transparency mode setting

### LINKED ISSUE #11087: [Announcements] Make it clear who can see announcements and where your announcements page is public
We already have clear defined rules on who can see announcements and whether an org's announcement index page is public. We should communicate these rules/behaviors to organizers.

Espeically the facts that your announcements page is private/deactivated/disabled (since some orgs care about staying private)

polypixeldev:
## Summary of the problem

Organizers should know how the announcement page works for non-transparent and transparent organizations!

Closes #11087

## Describe your changes

Adds a bullet point to the list about transparency mode and a note in the announcement overview blankslate about visibility.

### DIFF
--- a/app/views/events/announcement_overview.html.erb
+++ b/app/views/events/announcement_overview.html.erb
@@ -40,6 +40,9 @@
         <p class="text-gray-500 my-1 max-w-sm">
           Create a new announcement to keep your followers up to date!
         </p>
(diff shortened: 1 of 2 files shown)


sampoder (approved):
lgtm! thank you

PR #11091, opened 2025-07-24
Always show tags section in hcb code show for members

### LINKED ISSUE #11090: Tags missing
### Describe your goals

I was trying to add a tag when clicking into the transaction

### Describe the bug

the Tag button has gone missing 

shows on the ledger
Image

missing inside the transaction details 

Image

polypixeldev:
## Summary of the problem

While it doesn't make sense for readers or transparent viewers to see the tags section if there's nothing there, anyone able to add a tag should be able to see the section if there are no tags already.

Fixes #11090

## Describe your changes

Adds a check for `organizer_signed_in?(as: :member)`

### DIFF
--- a/app/views/hcb_codes/_tags.html.erb
+++ b/app/views/hcb_codes/_tags.html.erb
@@ -1,4 +1,4 @@
-<% if hcb_code.tags.any? || organizer_signed_in?(as: :member) %>
+<% if hcb_code.tags.any? || TagPolicy.new(current_user, @event).create? %>
   <div>
     <strong>Tags</strong>
     <div class="flex items-center flex-wrap" style="gap: 0.25rem">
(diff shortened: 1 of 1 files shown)


app/views/hcb_codes/_tags.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if hcb_code.tags.any? %>
+<% if hcb_code.tags.any? || organizer_signed_in?(as: :member) %>
```
Luke-Oldenburg: This returns true for auditors. It may be better to use a policy here.

sampoder:
```suggestion
<% if hcb_code.tags.any? || TagPolicy.new(current_user, @event).create? %>
```

sampoder (approved):
nice idea! thank you

PR #11101, opened 2025-07-25
Collect co-signer email when creating subevents

YodaLightsabr:
## Summary of the problem
Contracts for minors won't work because we aren't passing a co-signer email into `EventService::Create`

## Describe your changes
Pass a co-signer email into `EventService::Create`


sampoder:
Thank you!

sampoder (approved):
lgtm! thank you

PR #11106, opened 2025-07-25
Limit `transactions` in `Spending::Control` to main ledger

sampoder:
Closes https://github.com/hackclub/hcb/issues/10465


sampoder:
Tested in the Rails console

PR #11115, opened 2025-07-26
Add Tax Deductible field to Donations v4

Mohamad-Mortada:
## Summary of the problem
Tap to Pay needs a feature to collect donations for goods and services

## Describe your changes
Added a param

### DIFF
--- a/app/controllers/api/v4/donations_controller.rb
+++ b/app/controllers/api/v4/donations_controller.rb
@@ -12,7 +12,8 @@ module Api
                                    collected_by_id: current_user.id,
                                    in_person: true,
                                    name: params[:name] || nil,
-                                   email: params[:email] || nil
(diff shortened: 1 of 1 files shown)


app/controllers/api/v4/donations_controller.rb
```diff
@@ -12,7 +12,8 @@ def create
                                    collected_by_id: current_user.id,
                                    in_person: true,
                                    name: params[:name] || nil,
-                                   email: params[:email] || nil
+                                   email: params[:email] || nil,
+                                   tax_deductible: params[:tax_deductible]
```
sampoder:
```suggestion
                                   tax_deductible: params[:tax_deductible] || true
```

sampoder (approved):
code lgtm!

PR #11116, opened 2025-07-26
Allow admins to bypass OrganizerPosition.role_at_least?

YodaLightsabr:
## Summary of the problem
Admins are currently failing `OrganizerPosition.role_at_least?` checks due to a recent change in #9739 

## Describe your changes
This PR adds back two lines that prevented admins from bypassing `OrganizerPosition.role_at_least?` checks.

### DIFF
--- a/app/models/organizer_position.rb
+++ b/app/models/organizer_position.rb
@@ -55,6 +55,9 @@ class OrganizerPosition < ApplicationRecord
   end
 
   def self.role_at_least?(user, event, role)
+    return false unless event.present? && role.present?
+    return true if user&.admin?
+
     if role.to_s == "reader"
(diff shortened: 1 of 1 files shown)


app/models/organizer_position.rb
```diff
@@ -55,6 +55,9 @@ def tourable_options
   end
 
   def self.role_at_least?(user, event, role)
+    return false unless event.present? && role.present?
+    return true if user&.admin?
```
sampoder, line 59: We should give auditors read permissions right?

sampoder: ie. returning true if the user is an auditor and role is reader

YodaLightsabr: Was this done in the previous version? I'm onboard with this but this usually happens in policy checks IIRC.

PR #11118, opened 2025-07-26
[v4] Add PREMIUM as an available icon

Mohamad-Mortada:
## Summary of the problem
Premium HCB Icon perchance? 👀

## Describe your changes
Added condition to check if user is in an org with 10 or more active teenagers :)


sampoder (approved):
lgtm!

PR #11123, opened 2025-07-26
[Bug] Render verification letter in correct documents category

Sarvesh-Mk:
## Summary of the problem

I forgot to move the code to render the verification letter in the general tab so it creates a huge column and messes up the ui
image

## Describe your changes

Rending the verification letter in the same place as the contract.


sampoder (approved):
lgtm!

PR #11130, opened 2025-07-28
Switch `users.email` to `NOT NULL`

davidcornu:
## Summary of the problem

- We have a presence validation on `User#email`: https://github.com/hackclub/hcb/blob/1d41ad3bdc0c994cb879baca93d61df785320899/app/models/user.rb#L165
- We don't have any `NULL` or blank values in the database
    CleanShot 2025-07-28 at 09 33 30
- Yet `users.email` is a nullable column: https://github.com/hackclub/hcb/blob/fefb74bc9a399bd5d9ea2129f703a76fa76b7a4f/db/schema.rb#L2245

## Describe your changes

As per the strong migrations guidelines for changing a column's nullability:
1. Adds a non-validated check constraint
2. Validates (1), switches the column to `NOT NULL`, drops (1)

See #11015 for a similar change.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,7 +11,7 @@
 #  charge_notifications          :integer          default("email_and_sms"), not null
 #  comment_notifications         :integer          default("all_threads"), not null
 #  creation_method               :integer
-#  email                         :text
+#  email                         :text             not null
 #  full_name                     :string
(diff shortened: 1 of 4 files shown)


sampoder (approved):
lgtm! thank you.

PR #11131, opened 2025-07-28
[DX] Fix email display in Flipper

davidcornu:
## Summary of the problem

- In #10962 I configured `actor_names_source` to show the name and email of flipper actors https://github.com/hackclub/hcb/blob/fefb74bc9a399bd5d9ea2129f703a76fa76b7a4f/spec/misc/flipper_spec.rb#L20
- When testing I noticed we weren't displaying the email address 
    image

After some quick digging I found out Flipper assumes the output of `actor_names_source` is HTML so it performs additional sanitization before rendering: https://github.com/flippercloud/flipper/blob/fcb554ca5c5c9e324f81d52b186c111313504b39/lib/flipper/ui/views/feature.erb#L77-L78. This meant the email part of the output which is wrapped in `<` and `>` gets stripped out.

## Describe your changes

- Runs `ActionMailer::Base.email_address_with_name` through `CGI.escape_html`
- Updates the spec

which fixes the issue:

CleanShot 2025-07-28 at 10 20 10

### DIFF
--- a/config/initializers/flipper.rb
+++ b/config/initializers/flipper.rb
@@ -38,7 +38,7 @@ Flipper::UI.configure do |config|
       case model
       when "User"
         User.where(id: mapping.keys).pluck(:id, :email, :full_name).each do |(id, email, full_name)|
-          actor_names[mapping.fetch(id)] = ActionMailer::Base.email_address_with_name(email, full_name)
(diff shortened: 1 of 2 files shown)


sampoder (approved):
awesome - thank you! we can also put it in brackets if we'd like.

davidcornu:
> we can also put it in brackets if we'd like.

figured it was more straightforward to use the standard email format but wasn't aware that it was being treated as HTML output

PR #11132, opened 2025-07-28
Only show the latest announcement from HCB if one exists

davidcornu:
## Summary of the problem

On a fresh database we run into an exception because the HCB org doesn't have any announcements.

image

## Describe your changes

- Moves the announcement selection logic to the controller
- Hides the entire section depending on whether an announcement exists


sampoder (approved):
code lgtm - thanks for catching this one!

PR #11133, opened 2025-07-28
Require sudo mode for ACH transfers over $500

davidcornu:
## Summary of the problem

See #6346 

## Describe your changes

- Check for sudo mode in `AchTransfersController#create` when the amount is greater than $500
- Adds a basic test for the green path and the sudo mode case
- (somewhat unrelated) Adds a `turbo-visit-control` meta tag to the reauthentication page so it breaks out of turbo frames

### DIFF
--- a/app/views/sudo_mode/reauthenticate.html.erb
+++ b/app/views/sudo_mode/reauthenticate.html.erb
@@ -1,5 +1,15 @@
 <% content_for(:container_class, "container--sm py-20") %>
 
+<% content_for(:head) do %>
+  <%#
+    If sudo mode is triggered from within a Turbo frame, the default behaviour
+    would be to render the response within that frame. This meta tag tells Turbo
+    to treat it as a full page navigation.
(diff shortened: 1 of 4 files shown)


app/views/sudo_mode/reauthenticate.html.erb
```diff
@@ -1,5 +1,15 @@
...
 
+<% content_for(:head) do %>
+  <%#
+    If sudo mode is triggered from within a Turbo frame, the default behaviour
+    would be to render the response within that frame. This meta tag tells Turbo
+    to treat it as a full page navigation.
+    https://turbo.hotwired.dev/handbook/frames#%E2%80%9Cbreaking-out%E2%80%9D-from-a-frame
+  %>
+  <meta name="turbo-visit-control" content="reload">
+<% end %>
```
sampoder, line 11: thank you for this!

sampoder (approved):

PR #11135, opened 2025-07-28
Require sudo mode for wires above $500

davidcornu:
## Summary of the problem

See https://github.com/hackclub/hcb/issues/6346

## Describe your changes

- Check for sudo mode in `WiresController#create` when the amount is greater than $500
- Adds a basic test for the green path and the sudo mode case

### DIFF
--- a/app/controllers/wires_controller.rb
+++ b/app/controllers/wires_controller.rb
@@ -5,7 +5,6 @@ class WiresController < ApplicationController
 
   before_action :set_event, only: %i[new create]
   before_action :set_wire, only: %i[approve reject send_wire edit update]
-  before_action :enforce_sudo_mode, only: %i[create]
 
   def new
     @wire = @event.wires.build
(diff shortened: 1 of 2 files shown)


app/controllers/wires_controller.rb
```diff
@@ -5,6 +5,7 @@ class WiresController < ApplicationController
 
   before_action :set_event, only: %i[new create]
   before_action :set_wire, only: %i[approve reject send_wire edit update]
+  before_action :enforce_sudo_mode, only: %i[create]
```
sampoder: technically some organizations are exempt from the $500 minimum so we should implement similar logic to https://github.com/hackclub/hcb/pull/11133 imo

davidcornu: good catch! had noticed this but forgot to push my commit 😅

davidcornu:
rebase incoming

PR #11139, opened 2025-07-29
Return true for admins on `organizer_signed_in?`

sampoder:
I believe this will fix Ruien's issues


sampoder:
FYI @Luke-Oldenburg https://github.com/hackclub/hcb/pull/10883

PR #11142, opened 2025-07-29
Require sudo mode for checks above $500

davidcornu:
## Summary of the problem

See https://github.com/hackclub/hcb/issues/6346

## Describe your changes

- Adds basic test coverage for check creation
- Requires sudo mode if the check amount is greater than $500

### DIFF
--- a/app/controllers/sudo_mode_handler.rb
+++ b/app/controllers/sudo_mode_handler.rb
@@ -12,6 +12,8 @@ class SudoModeHandler
     email: 4,
   }.freeze
 
+  CENTS_THRESHOLD = 500_00
+
   Params = Struct.new(
     # When the form is submitted, `submit_method` determines which
     # authentication method the user selected, which in turn determines whether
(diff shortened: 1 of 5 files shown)


app/controllers/sudo_mode_handler.rb
```diff
@@ -12,6 +12,8 @@ class SudoModeHandler
     email: 4,
   }.freeze
 
+  CENTS_THRESHOLD = 500_00
```
YodaLightsabr:
```suggestion
  THRESHOLD_CENTS = 500_00
```
nitpick: We tend to tack on `_cents` at the end of a variable name as a unit. Unless there's a reason not to, it might be better to use `THRESHOLD_CENTS`

YodaLightsabr (approved):
Looks good!

app/controllers/sudo_mode_handler.rb
davidcornu: makes sense

PR #11146, opened 2025-07-30
[Announcements] Add spending summary blocks and improve monthly template

polypixeldev:
## Summary of the problem

We need more blocks! These spending summary blocks will be useful for monthly announcements, which will be generated soon and need a better template.

## Describe your changes

Adding in the graph blocks that are currently rendered on the new home page as announcement blocks:
- Top merchants
- Top categories
- Top users
- Top tags

In emails, the breakdowns are rendered just as bullet points since getting the chart to render as an image on the server is going to take some more work and debugging. The breakdown engine has also been modified to allow for start and end dates instead of just a duration.

The monthly announcement template has been improved to include the donation goal and new top categories block. Finally, the `empty?` method has been added on all blocks to make sure the monthly template does not include data that does not exist - this also will be used in a future PR to prevent sending monthly announcements with no data.

image
image
image


app/javascript/controllers/tiptap/nodes/top_categories_node.js
```diff
@@ -0,0 +1,55 @@
...
+export const TopCategoriesNode = Node.create({
+  name: 'Announcement::Block::TopCategories',
+  atom: true,
+  group: 'block',
+  priority: 2000,
+  addAttributes() {
+    return {
+      start_date: {},
+      id: {},
+      html: {},
```
davidcornu: **concern**: Should we be using more meaningful values here or is `{}` what tiptap expects?

app/javascript/controllers/tiptap/nodes/top_categories_node.js
```diff
@@ -0,0 +1,55 @@
...
+            })
+          }
+        }
+      })
+
+      observer.observe(document, {
+        attributes: true,
+        childList: true,
+        subtree: true,
+      })
```
davidcornu: 1. **concern**: I'm surprised we need this much fanfare to get `ReactRailsUJS` to mount the relevant elements. Does either tiptap or react_ujs provide a better way to handle this kind of lifecycle?
2. **suggestion**: We have basically the exact same code for every node. Can we centralize this logic in one place assuming the answer to (1) is that there is no other way.

app/models/announcement/block/donation_goal.rb
```diff
@@ -27,18 +27,24 @@ class DonationGoal < ::Announcement::Block
...
 
+      def empty?
+        goal.nil?
+      end
+
       private
 
+      def goal
+        @goal ||= Donation::Goal.find_by(event: announcement.event, id: goal_param)
+        @goal ||= announcement.event.donation_goal
```
davidcornu: **suggestion** (minor): I find it a bit odd to do this in two steps

```suggestion
        @goal ||= (Donation::Goal.find_by(event: announcement.event, id: goal_param) || announcement.event.donation_goal)
```

app/models/announcement/block/donation_summary.rb
```diff
@@ -27,27 +27,35 @@ class DonationSummary < ::Announcement::Block
...
         Announcements::BlocksController.renderer.render partial: "announcements/blocks/donation_summary", locals: { donations:, total:, start_date:, end_date:, is_email:, block: self }
       end
 
       private
 
+      def donations
+        start_date = start_date_param
+        end_date = end_date_param
+
+        @donations ||= announcement.event.donations.where(aasm_state: [:in_transit, :deposited], created_at: start_date..end_date)
```
davidcornu: **concern**: `.where(aasm_state: [:in_transit, :deposited]` feels to me like it should be a scope on `Donation` that we can descriptively name (and be aware of if we ever modify `Donation`)

app/models/announcement/block/donation_summary.rb
```diff
@@ -27,27 +27,35 @@ class DonationSummary < ::Announcement::Block
...
+      end
+
       def start_date_param
         self.parameters["start_date"] ||= 1.month.ago.to_s
 
-        Date.parse(self.parameters["start_date"])
+        DateTime.parse(self.parameters["start_date"])
```
davidcornu: **suggestion** (minor): I find it a bit odd that we're stringifying the fallback timestamp before parsing it

```ruby
        if self.parameters["start_date"].present?
          DateTime.parse(self.parameters["start_date"])
        else
          1.month.ago.to_s
        end
```

app/models/announcement/block/top_categories.rb
```diff
@@ -0,0 +1,63 @@
...
+#  index_announcement_blocks_on_announcement_id  (announcement_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (announcement_id => announcements.id)
+#
+class Announcement
+  class Block
+    class TopCategories < ::Announcement::Block
+      before_create :start_date_param
```
davidcornu: **concern**: As far as I can tell,  the `start_date_param` method doesn't mutate anything. Is that intentional?

app/models/announcement/block/top_merchants.rb
```diff
@@ -0,0 +1,63 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_...  (announcement_id => announcements.id)
+#
+class Announcement
+  class Block
+    class TopMerchants < ::Announcement::Block
+      before_create :start_date_param
+      before_create :end_date_param
```
davidcornu: **suggestion** I see a lot of repetition to handle these params. Should we move some of this logic to a reusable module?

app/services/breakdown_engine/tags.rb
```diff
@@ -2,21 +2,22 @@
...
       @event = event
-      @timeframe = timeframe
+      @start_date = start_date
+      @end_date = end_date
     end
 
     def run
       tags = @event.tags.includes(hcb_codes: [:canonical_transactions, :canonical_pending_transactions]).each_with_object([]) do |tag, array|
-        transactions = @timeframe.present? ? tag.hcb_codes.select { |hcb_code| hcb_code.date&.after?(@timeframe.ago) } : tag.hcb_codes
+        transactions = tag.hcb_codes.select { |hcb_code| (@start_date.nil? || hcb_code.date&.after?(@start_date)) && (@end_date.nil? || hcb_code.date&.before?(@end_date)) }
```
davidcornu: **concern**: This is getting a bit too complex for single expressions

```suggestion
        transactions = tag.hcb_codes.select do |hcb_code| 
          next false unless @start_date.nil? || hcb_code.date&.after?(@start_date)
          next false unless @end_date.nil? || hcb_code.date&.before?(@end_date)
          true
        end
```

app/javascript/controllers/tiptap/nodes/top_categories_node.js
polypixeldev: The object is used to configure properties for the attribute like default value or custom rendering (https://v2.tiptap.dev/docs/editor/extensions/custom-extensions/extend-existing#attributes) - I can't find any documentation on if anything else would work, but an empty object is what made sense to me

app/javascript/controllers/tiptap/nodes/top_categories_node.js
polypixeldev: Since the HTML is dynamically fetched and inserted into the DOM after initial page load, `ReactRailsUJS` doesn't automatically mount it, and just calling `mountComponents` inside `addNodeView` won't work since the HTML is not actually inserted in the DOM until after that function finishes. The `MutationObserver` is there to only call `mountComponents` once the node is added - I'll factor it out

app/models/announcement/block/top_categories.rb
polypixeldev: Yep - I wanted to keep the start date on the graph blocks `nil`, but forgot to get rid of the callback

app/models/announcement/block/top_merchants.rb
polypixeldev: Good idea, separated them out into concerns for `HasEndDate` and `HasFlexibleStartDate` (for the blocks that don't have a default start date)

sampoder (approved):
Code and demo LGTM!

PR #9953, opened 2025-03-24
Add `docs` layout for Brand Guidelines & Security pages

YodaLightsabr:
## Summary of the problem
https://hcb.hackclub.com/branding and https://hcb.hackclub.com/security are content-heavy pages with little support from a dedicated layout. The security page specifically is pretty ugly because it's just a bunch of content in the center of a page.

## Describe your changes
Adds the `docs` layout to be used on text-heavy documentation pages:

image

PR #11157, opened 2025-07-30
Properly center "Welcome to" on small screens

davidcornu:
Before | After
---- | ----
CleanShot 2025-07-30 at 13 33 16@2x | CleanShot 2025-07-30 at 13 32 52@2x 

## Summary of the problem

https://hackclub.slack.com/archives/C068U0JMV19/p1753892044644149

We have CSS that gives headings a bit more space on smaller screens https://github.com/hackclub/hcb/blob/272c735e134d77810e5bde25c4de8a150e44f18a/app/assets/stylesheets/application.scss#L406-L419 by stretching them out horizontally by 1rem. However the `<h3>` element with "Welcome to" has a `m0` class on it which zeroes out all the margin https://github.com/hackclub/hcb/blob/300969e63f9c63176e4c68da02e7f4a923756223/app/assets/stylesheets/space.css#L1-L3 and has a higher precedence. This results in the `<h3>` being rendered off-center as it is larger than its container.

CleanShot 2025-07-30 at 13 41 28@2x

## Describe your changes

Instead of using `m0`, use Tailwind's `my-0` which only zeroes out `margin-top` and `margin-bottom`.

### DIFF
--- a/app/views/layouts/docs.html.erb
+++ b/app/views/layouts/docs.html.erb
@@ -129,7 +129,7 @@
 
             <div class="welcome__content">
               <div class="center">
-                <h3 class="h3 caps secondary block m0">Welcome to</h3>
+                <h3 class="h3 caps secondary block my-0">Welcome to</h3>
                 <h1 class="primary mt0 border-none"><%= @organizer_position&.initial? ? "HCB" : @event.name %></h1>
               </div>
 
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -180,7 +180,7 @@
 
             <div class="welcome__content">
               <div class="center">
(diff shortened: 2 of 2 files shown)


app/views/layouts/docs.html.erb
```diff
@@ -129,7 +129,7 @@
 
             <div class="welcome__content">
               <div class="center">
-                <h3 class="h3 caps secondary block m0">Welcome to</h3>
```
davidcornu, line 132: This layout was introduced in https://github.com/hackclub/hcb/pull/9953 but I'm not actually sure it makes use of this code. I chose to update it anyway so both remain consistent.

YodaLightsabr (approved):
Lgtm

PR #11160, opened 2025-07-31
Remove `ignored_columns` from `Announcement`

davidcornu:
## Summary of the problem

Follow-up to #11040 

We no longer need to ignore these columns as they are no longer part of the production schema.

## Describe your changes

Remove the line.


davidcornu:
fyi @polypixeldev

sampoder (approved):

PR #11161, opened 2025-07-31
Make sure `LoginsController` and `SudoModeHandler` handle the correct logins

davidcornu:
## Summary of the problem

We create two types of logins (initial logins, and reauthentications) which are handled in different ways. When loading them we don't currently make sure we are dealing with the type we expect.

## Describe your changes

- Make sure `LoginsController` only handles initial logins
- Make sure `SudoModeHandler` only handles reauthentications


sampoder (approved):
lgtm! thank you

davidcornu:
Needs updating to account for https://github.com/hackclub/hcb/pull/11192

PR #11162, opened 2025-07-31
Move from `logins.initial_login_id` to `logins.is_reauthentication`

davidcornu:
## Summary of the problem

One of the underlying assumptions behind the design of sudo mode is that `UserSession` would always have at least one `Login` (referred to as the initial login), which could be referenced by subsequent logins (referred to as reauthentications) in their `initial_login_id`.

This is outlined here: https://github.com/hackclub/hcb/pull/10903

However, I recently realized that this assumption does not hold as we have a mechanism for admins to impersonate and unimpersonate users which creates `UserSession` records without any associated `Login`s whatsoever. This means
- This assertion fails in production: https://github.com/hackclub/hcb/blob/c772fdf8a2456e76e79d4f42c6735fb4058c0489/app/controllers/sudo_mode_handler.rb#L186
- We don't have a way to model reauthentications in that context as we don't have an `initial_login_id` to reference

https://github.com/hackclub/hcb/blob/21ec8d158f5d4ba52f4e7d780124dd01a26e611b/app/helpers/sessions_helper.rb#L17-L26

Internal discussions:
- https://hackclub.slack.com/archives/C047Y01MHJQ/p1753895603368189
- https://hackclub.slack.com/archives/C096P73LEMT/p1753889138870219

## Describe your changes

> [!WARNING]
> The commits in this PR are not intended to be shipped all at once. I included the complete set of changes here to make it easier for reviewers to reason about the big picture, but once approved I will extract them into separate PRs (as indicated by the numbers in the commit messages).
>
> **As such, I _highly_ recommend reviewing this commit by commit as there is necessary code that is added and later removed that will not appear in the combined diff**

**Stage 1**: https://github.com/hackclub/hcb/pull/11190
- Add `is_reauthentication` column to the `logins` table

**Stage 2**: https://github.com/hackclub/hcb/pull/11191
- Automatically populate `is_reauthentication` when logins are created
- Backfill `is_reauthentication` for existing logins

**Stage 3**: https://github.com/hackclub/hcb/pull/11192
- Switch application logic to use `is_reauthentication` and ignore `initial_login_id`

**Stage 4**: https://github.com/hackclub/hcb/pull/11248
- Drop the `initial_login_id` column from the `logins` table

**Stage 5**: https://github.com/hackclub/hcb/pull/11249
- Remove `initial_login_id` from `ignored_columns`

### DIFF
--- a/app/models/login.rb
+++ b/app/models/login.rb
@@ -22,8 +22,6 @@
 #  index_logins_on_user_session_id      (user_session_id)
 #
 class Login < ApplicationRecord
-  self.ignored_columns = %w[initial_login_id]
-
   include AASM
   include Hashid::Rails
 
--- a/app/models/login.rb
+++ b/app/models/login.rb
@@ -21,10 +21,6 @@
 #  index_logins_on_user_id              (user_id)
 #  index_logins_on_user_session_id      (user_session_id)
 #
-# Foreign Keys
-#
-#  fk_rails_...  (initial_login_id => logins.id)
-#
 class Login < ApplicationRecord
   self.ignored_columns = %w[initial_login_id]
 
--- a/app/models/login.rb
+++ b/app/models/login.rb
@@ -8,6 +8,7 @@
 #  aasm_state               :string
 #  authentication_factors   :jsonb
 #  browser_token_ciphertext :text
+#  is_reauthentication      :boolean          default(FALSE), not null
 #  created_at               :datetime         not null
 #  updated_at               :datetime         not null
 #  initial_login_id         :bigint
--- a/app/models/login.rb
+++ b/app/models/login.rb
@@ -87,6 +87,8 @@ class Login < ApplicationRecord
     mark_complete! if may_mark_complete?
   end
 
+  before_create(:sync_is_reauthentication)
+
   def authentication_factors_count
     return 0 if authentication_factors.nil?
 
@@ -147,6 +149,10 @@ class Login < ApplicationRecord
     else
       1
     end
+  end
+
+  def sync_is_reauthentication
+    self.is_reauthentication = reauthentication?
   end
 
 end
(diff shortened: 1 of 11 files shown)


app/models/login.rb
```diff
@@ -130,7 +120,7 @@ def available_factors
   end
 
   def reauthentication?
-    initial_login_id.present?
+    is_reauthentication?
   end
```
sampoder, line 124: Can we alias this?

db/migrate/20250731150352_drop_initial_login_id_from_logins.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class DropInitialLoginIdFromLogins < ActiveRecord::Migration[7.2]
+  def change
+    # The column is ignored and all references have been removed
```
sampoder, line 5: Is this ignore already in production?

app/models/login.rb
davidcornu: That was my first thought but `is_reauthentication` is a `method_missing` thing that isn't defined until use 🙃

db/migrate/20250731150352_drop_initial_login_id_from_logins.rb
davidcornu: It would be (see the big warning in the description). This would ship as the 4th PR.

sampoder: Perfect - thank you

app/models/login.rb
sampoder: Wack :/ oh well

sampoder (approved):
code lgtm! is this all shipping as one PR?

davidcornu:
This PR has been split up into 5 different ones (listed in the description) and as such is no longer necessary.

PR #11163, opened 2025-07-31
Add sneeze flavor text

YodaLightsabr:
## Summary of the problem
Our flavor texts are missing the sneeze flavor text

## Describe your changes
Add the sneeze flavor text


sampoder (approved):
ACHoo!

PR #11166, opened 2025-07-31
Drop wise transfers table

YodaLightsabr:


davidcornu:
For context - there was an accidental merge to `main` after which we rolled back the deploy and reverted the changes, but the migration ran in production so this is the clean-up.

davidcornu (approved):

PR #11168, opened 2025-07-31
💸 Support international transfers via Wise

YodaLightsabr:
Sending money internationally is hard. Wires are expensive, and Wise transfers always required coordination off-platform and created a heavy operations workload. This PR builds Wise transfers directly into HCB.

image


app/models/canonical_pending_transaction.rb
sampoder, line 1: I think instead of doing `amount_pending` we should make an estimate of the amount like we do with wire transfers. We could either use the Wise API (https://docs.wise.com/api-docs/api-reference/quote) for this or use the currency conversion we already have in the codebase (what wires use). I don't think we should introduce the concept of pending amount CPTs.

app/models/concerns/has_wise_recipient.rb
```diff
@@ -0,0 +1,161 @@
...
+    #   unless bic_code.nil? || bic_code.match(/[A-Z]{4}([A-Z]{2})[A-Z0-9]{2}([A-Z0-9]{3})?$/) # https://www.johndcook.com/blog/2024/01/29/swift/
+    #     errors.add(:bic_code, "is not a valid SWIFT / BIC code")
+    #   end
+    # end
+
+    # validate do
+    #   if IBAN_FORMATS[bank_country.to_sym] && !account_number.match(IBAN_FORMATS[bank_country.to_sym])
+    #     errors.add(:account_number, "does not meet the required format for this country")
+    #   end
+    # end
```
sampoder: We should remove this code if we don't plan to use it.

app/models/concerns/has_wise_recipient.rb
```diff
@@ -0,0 +1,161 @@
...
+
+    validate do
+      if POSTAL_CODE_FORMATS[recipient_country.to_sym] && !address_postal_code.match(POSTAL_CODE_FORMATS[recipient_country.to_sym])
+        errors.add(:address_postal_code, "does not meet the required format for this country")
+      end
+    end
```
sampoder, line 14: This was designed specifically for Column's required formats - not sure if Wise has different / looser requirements

app/models/concerns/has_wise_recipient.rb
```diff
@@ -0,0 +1,161 @@
...
+    "US": /\A\d{5}(?:-\d{4})?\z/,
+    "CN": /\A\d{6}\z/,
+    "JP": /\A\d{3}-\d{4}\z/,
+    "FR": /\A\d{5}\z/,
+    "DE": /\A\d{5}\z/
+  }.freeze
+
+  ACCOUNT_NUMBER_FIELD = { type: :text_field, key: "account_number", placeholder: "123456789", label: "Account number" }.freeze
+
+  AVAILABLE_CURRENCIES = (::EuCentralBank::CURRENCIES + ["EUR"] - ["USD"]).freeze
```
sampoder: Where did we get this list from? This was what we used for wires only because this is what we could convert

app/models/concerns/has_wise_recipient.rb
```diff
@@ -0,0 +1,161 @@
...
+
+    # View https://github.com/hackclub/hcb/issues/9037 for context. Limited in India only, at the moment.
+
+    validate on: :create do
+      if recipient_information[:purpose_code].present? && RESTRICTED_PURPOSE_CODES[recipient_country.to_sym]&.include?(recipient_information[:purpose_code])
+        errors.add(:purpose_code, "can not be used on HCB, please use a more specific purpose code or contact us.")
+      end
+    end
+
+    def self.information_required_for(currency) # country can be null, in which case, only the general fields will be returned.
```
sampoder: Would be good to include a comment about where this information came from + how to extend this

app/models/wire.rb
```diff
@@ -43,6 +43,7 @@
 #  fk_rails_...  (user_id => users.id)
 #
 class Wire < ApplicationRecord
+  puts "Wire model loaded"
```
```suggestion
```

app/views/canonical_pending_transactions/_canonical_pending_transaction.html.erb
```diff
@@ -124,7 +124,11 @@
     </div>
   </td>
   <td class="nowrap text-right">
-    <%= render_transaction_amount pt.amount %>
+    <% if pt.amount_pending? %>
+      PENDING
+    <% else %>
+      <%= render_transaction_amount pt.amount %>
+    <% end %>
```
sampoder: See earlier comment about pending amounts - I really don't think we should do this.

app/views/hcb_codes/_wise_transfer.html.erb
```diff
@@ -0,0 +1,90 @@
...
+      <%= format_datetime @hcb_code.wise_transfer.created_at %>
+    </p>
+  </section>
+  <% if @hcb_code.pt.amount_pending? %>
+    <section class="card__banner relative border-bottom bg-[#9FE870] dark:text-[#9FE870] text-[#163300] dark:bg-[#163300] flex flex-row items-center justify-start gap-4 md:gap-2">
+      <%= inline_icon "wise", size: 16 %>
+      <div class="flex md:flex-row flex-col items-start justify-start gap-1 flex-grow">
+        Funds have not been withdrawn from your account yet.
+        <span class="md:flex-grow"></span>
+        <%= link_to "Learn more about Wise transfers on HCB →", "https://help.hcb.hackclub.com/article/65-wise", class: "dark:!text-[#9FE870] !text-[#163300] bold no-underline after:content-[''] after:top-0 after:left-0 after:bottom-0 after:right-0 after:z-index-1 after:absolute", target: "_blank" %>
```
sampoder: 404 fyi

app/views/wise_transfers/new.html.erb
```diff
@@ -0,0 +1,110 @@
...
+<h1>Send a Wise transfer</h1>
+
+<%= render "callout",
+    type: "info",
+    icon: "wise",
+    title: "Important information about Wise transfers",
+    class: "!border-[#9FE870] [&>div.flex]:text-[#163300] dark:[&>div.flex]:text-[#9FE870]",
+    footer: link_to("Learn more about Wise on HCB →", "https://help.hcb.hackclub.com/article/65-wise") do %>
+  HCB is partnering with Wise to offer international transfers to over 100 countries. HCB imposes no fees or minimums, but any fees charged by Wise will be passed on to your organization.
+<% end %>
```
sampoder: I think we should mention here that we support fee-free international transfers over $500

app/models/wise_transfer.rb
```diff
@@ -0,0 +1,176 @@
...
+#  address_postal_code           :string
+#  address_state                 :string
+#  amount_cents                  :integer          not null
+#  approved_at                   :datetime
+#  bank_name                     :string
+#  branch_number_bidx            :string
+#  branch_number_ciphertext      :string
+#  currency                      :string           not null
+#  institution_number_bidx       :string
+#  institution_number_ciphertext :string
```
sampoder: I assume these are being stored as separate columns to encrypt them. Can we try directly encrypting `recipient_information`? I'd like to avoid us having to add new columns as we add support for new countries. See the `PaymentRecipient` as an example of this.

app/models/wise_transfer.rb
```diff
@@ -0,0 +1,176 @@
...
+
+  include PublicActivity::Model
+  tracked owner: proc { |controller, record| controller&.current_user }, event_id: proc { |controller, record| record.event.id }, only: [:create]
+
+  after_create do
+    create_canonical_pending_transaction!(
+      event:,
+      amount_cents: 0,
+      amount_pending: true,
+      memo: "#{Money.from_cents(amount_cents, currency).format} #{currency} Wise to #{recipient_name}", # this really _really_ sucks. I'm open to ideas here.
```
sampoder:
```suggestion
      memo: "#Wise transfer to #{recipient_name} ({Money.from_cents(amount_cents, currency).format} #{currency} )", # this really _really_ sucks. I'm open to ideas here.
```

app/models/wise_transfer.rb
```diff
@@ -0,0 +1,176 @@
...
+  monetize :amount_cents, as: "amount", with_model_currency: :currency
+  monetize :usd_amount_cents, as: "usd_amount", allow_nil: true
+
+  include PublicActivity::Model
+  tracked owner: proc { |controller, record| controller&.current_user }, event_id: proc { |controller, record| record.event.id }, only: [:create]
+
+  after_create do
+    create_canonical_pending_transaction!(
+      event:,
+      amount_cents: 0,
```
sampoder: We probably should give people an estimate, even if it's a straight currency conversion without fees (that's what we used to do for PayPal, the amount without fees).

app/models/wise_transfer.rb
```diff
@@ -0,0 +1,176 @@
...
+
+  private
+
+  def extract_wise_id
+    url_prefix = "https://wise.com/transactions/activities/by-resource/TRANSFER/"
+
+    return unless wise_id.present? && wise_id.starts_with?(url_prefix)
+
+    self.wise_id = wise_id[url_prefix.length..]
+  end
```
sampoder: Should this be done on the frontend?

app/views/admin/wise_transfer_process.html.erb
```diff
@@ -0,0 +1,231 @@
...
+              <%= form.label :usd_amount, "Amount (USD)", class: "bold mb1" %> <br>
+              $ <%= form.number_field :usd_amount,
+                placeholder: "500.00",
+                step: 0.01,
+                required: true,
+                data: { controller: "truncate-decimal", action: "truncate-decimal#truncate blur->truncate-decimal#pad" } %>
+            </div>
+            <%= form.submit "🏦 Check balance" %>
+          <% end %>
+        <% end %>
```
sampoder: Is there a reason we aren't doing the balance check on the frontend using Alpine?

app/views/admin/wise_transfer_process.html.erb
```diff
@@ -0,0 +1,231 @@
...
+  <div class="flex flex-row items-start justify-start gap-4">
+    <span class="-m-[2px] w-7 h-7 flex-shrink-0 rounded-full flex items-center justify-center text-xs font-bold border-2 border-solid border-smoke">2</span>
+    <p class="m-0">
+      Input transfer details in Wise and update the USD amount to perform an ample balance check against the organization.
+    </p>
+  </div>
+  <div class="flex flex-row items-start justify-start gap-4">
+    <span class="-m-[2px] w-7 h-7 flex-shrink-0 rounded-full flex items-center justify-center text-xs font-bold border-2 border-solid border-smoke">3</span>
+    <p class="m-0">
+      Complete the transfer in Wise and paste in the Wise transfer ID on HCB.
```
sampoder, line 246: Can you include instructions about how to get the ID?

sampoder (changes requested):
Mostly looking good - biggest comment is about the concept of pending amounts which I think we should avoid. 

Question: how do these transfers get mapped? I think you need to add some code to the admin CT mapping page so they can map to a Wise transfer. Look at the code for wires to do this.

app/views/admin/wise_transfer_process.html.erb
leowilkin: The ID is within the URL - there's a box for the 'completed transaction' URL that Wise gives you.

app/models/wise_transfer.rb
leowilkin: @sampoder if you think it would be worth it, I can get an average fee that we've been charged since we started using Wise - and factor that into the pending txn?

app/models/canonical_pending_transaction.rb
leowilkin: This is more of a quick fix, rather than a long-term solution. I've got about circa. 200-300 transfers to send out next week and we wanted to skunk works this with HQ events where it doesn't really matter how much the amount is.

sampoder: We can do a hack for setting amount_cents (it doesn't have to have perfect accuracy, as you mentioned in another comment we can do conversion + average fee) but I don't want to push this change to CPT

app/views/admin/wise_transfer_process.html.erb
sampoder: Yes - please include those instructions

app/views/hcb_codes/_wise_transfer.html.erb
leowilkin: I know - this is being made public when we push it

app/models/wise_transfer.rb
leowilkin: The average fee that Wise has charged us is 3.55990228% of the original USD value.

sampoder: Let's add 4% to everything then

app/views/wise_transfers/new.html.erb
YodaLightsabr: For the time being these will only be for HQ and it's alright if they pay fees. We also haven't worked the logistics of doing this yet—having ops do disbursements every time could get messy.

sampoder: Sorry - to clarify, I think we should always charge on the fee for Wise transfers. My original comment would have been better phrased as: 

> I think we should mention here that we support fee-free wire transfers over $500

app/models/wise_transfer.rb
YodaLightsabr: I'm not too sure about doing this on the frontend, it might be a bit complicated. I'm open to doing it in the controller action though.

app/models/wise_transfer.rb
YodaLightsabr: We've discussed this on Slack and I think the plan is to move forward with using the Wise API to get a non-persistent quote and render that on the frontend. It takes the same amount of effort as just converting the currency, but we have the added benefit of knowing how much the fee will be and rendering that for the user.

app/models/canonical_pending_transaction.rb
YodaLightsabr: Because we're moving forward with more accurate amount estimates, we're scrapping the idea of pending amounts.

app/views/wise_transfers/new.html.erb
YodaLightsabr: Ah, I see. Yeah, that's fair. I'm not sure what the best way of doing this is, so I'm open to ideas. Our transfer options are a lot more complicated now than they used to be...

app/models/concerns/has_wise_recipient.rb
YodaLightsabr: I have to imagine the postal code formats would be good to keep around? I don't see a world where Column has different format validations than the actual postal codes

app/policies/wise_transfer_policy.rb
```diff
@@ -0,0 +1,42 @@
...
+  def update?
+    user&.admin?
+  end
+
+  def mark_sent?
+    user&.admin?
+  end
+
+  def generate_quote?
+    user&.admin? || user.events.any?
```
sampoder:
```suggestion
    user&.auditor? || user.events.any?
```

app/models/wise_transfer.rb
sampoder: +1 for moving it to the controller. I think you could also just run this code after someone clicks off the input:

onblur = (e) => {
       url_prefix = "https://wise.com/transactions/activities/by-resource/TRANSFER/"

       return unless e.target.value.starts_with?(url_prefix)

       e.target.value = e.target.value [url_prefix.length..]
}

but controller works as well. I just want consistent inputs into the model.

app/controllers/admin_controller.rb
```diff
@@ -40,6 +40,14 @@ def transaction
     @canonical_pending_transactions = CanonicalPendingTransaction.unmapped.where(amount_cents: @canonical_transaction.amount_cents)
     @ahoy_events = Ahoy::Event.where("name in (?) and (properties->'canonical_transaction'->>'id')::int = ?", [::SystemEventService::Write::SettledTransactionMapped::NAME, ::SystemEventService::Write::SettledTransactionCreated::NAME], @canonical_transaction.id).order("time desc")
 
+    if @canonical_transaction.memo.include?("WISE INC")
+      potential_wise_transfers = WiseTransfer.sent.where(amount_cents: @canonical_transaction.amount_cents)
```
sampoder: Is this `amount_cents` in USD?

app/models/concerns/has_wise_recipient.rb
sampoder: Sgtm - let's run with Column's for now

app/models/wise_transfer.rb
```diff
@@ -0,0 +1,208 @@
...
+  monetize :amount_cents, as: "amount", with_model_currency: :currency
+  monetize :usd_amount_cents, as: "usd_amount", allow_nil: true
+
+  include PublicActivity::Model
+  tracked owner: proc { |controller, record| controller&.current_user }, event_id: proc { |controller, record| record.event.id }, only: [:create]
+
+  after_create do
+    create_canonical_pending_transaction!(
+      event:,
+      amount_cents: estimate_usd_amount_cents,
```
sampoder:
```suggestion
      amount_cents: estimated_usd_amount_cents,
```

app/models/wise_transfer.rb
```diff
@@ -0,0 +1,208 @@
...
+                         })
+
+    payment_option = response.body["paymentOptions"].first
+    price_after_fees = Money.from_dollars(payment_option["sourceAmount"], "USD")
+    fees = payment_option["price"]["items"]
+    payin_fee = Money.from_dollars(fees.find { |fee| fee["type"] == "PAYIN" }["value"]["amount"], "USD")
+
+    price_before_payin_fee = price_after_fees - payin_fee
+
+    wise_ach_fee = 1.0017 # The Wise API doesn't show profile-specific payment methods like ACH, but the ACH fee is a standard 0.17% of the amount sent.
```
sampoder, line 197: thank you for the comment!

app/models/wise_transfer.rb
```diff
@@ -0,0 +1,208 @@
...
+                           targetCurrency: money.currency_as_string,
+                           targetAmount: money.dollars
+                         })
+
+    payment_option = response.body["paymentOptions"].first
+    price_after_fees = Money.from_dollars(payment_option["sourceAmount"], "USD")
+    fees = payment_option["price"]["items"]
+    payin_fee = Money.from_dollars(fees.find { |fee| fee["type"] == "PAYIN" }["value"]["amount"], "USD")
+
+    price_before_payin_fee = price_after_fees - payin_fee
```
sampoder:
```suggestion
    pay_in_fee = Money.from_dollars(fees.find { |fee| fee["type"] == "PAYIN" }["value"]["amount"], "USD")

    price_before_pay_in_fee = price_after_fees - pay_in_fee
```
nit because I read this as "paying" fee at first and Wise seems to treat it as two words: `option.payIn`

app/views/events/_transfer_form.html.erb
```diff
@@ -6,30 +6,41 @@
...
         <div class="field field--options fit">
           <%= link_to new_event_ach_transfer_path(@event), class: "text-decoration-none btn--form-option", data: { turbo_frame: "_top" } do %>
             <%= inline_icon "payment-transfer", size: 28 %>
             <strong>ACH transfer</strong>
+            <small>United States</small>
           <% end %>
           <%= link_to new_event_increase_check_path(@event), class: "text-decoration-none btn--form-option", data: { turbo_frame: "_top" } do %>
             <%= inline_icon "email", size: 28 %>
             <strong>Mailed check</strong>
+            <small>United States</small>
```
sampoder, line 19: nit: Domestic?

app/views/events/_transfer_form.html.erb
```diff
@@ -6,30 +6,41 @@
...
-            <strong>International wire</strong>
+            <div class="badge badge--corner bg-info m1 mt0" style="width: fit-content;">$500 minimum</div>
+            <strong>Wire transfer</strong>
+            <small>International</small>
           <% end %>
           <%= link_to new_disbursement_path(source_event_id: @event), class: "text-decoration-none btn--form-option", data: { turbo_frame: "_top" } do %>
             <%= inline_icon "bank-account", size: 28 %>
             <strong>HCB transfer</strong>
-            <small>(between HCB organizations)</small>
+            <small>Any HCB organization</small>
```
sampoder, line 37: I think the original label was clearer

app/views/wise_transfers/new.html.erb
```diff
@@ -0,0 +1,112 @@
...
+
+<h1>Send a Wise transfer</h1>
+
+<%= render "callout",
+    type: "info",
+    icon: "wise",
+    title: "Important information about Wise transfers",
+    class: "!border-[#9FE870] [&>div.flex]:text-[#163300] dark:[&>div.flex]:text-[#9FE870]",
+    footer: link_to("Learn more about Wise on HCB →", "https://help.hcb.hackclub.com/article/65-wise") do %>
+  HCB is partnering with Wise to offer international transfers to over 100 countries. HCB imposes no fees or minimums, but any fees charged by Wise will be passed on to your organization.
```
sampoder:
```suggestion
  We've integrated Wise into HCB to offer international transfers to over 100 countries. HCB never charges fees to send transfers, however Wise will charge processing fees to your organization.
  <br />
  Sending over $500? Avoid all fees by sending an international wire.
```

app/views/wise_transfers/new.html.erb
```diff
@@ -0,0 +1,112 @@
...
+      </div>
+    </div>
+    <% @wise_transfer.errors.messages_for(:amount).each do |message| %>
+      <div class="primary"><%= message %></div>
+    <% end %>
+    <% if @event.balance <= 0 %>
+      <span class="error">There are no funds to transfer.<br>
+    <% end %>
+
+    <p x-text="(usd_quote == null || currency == null || amount == null) ? 'Enter amount to get USD estimate' : Number.isNaN(usd_quote) ? 'Loading quote from Wise...' : `Roughly ${usd_quote} USD`" class="muted mt-0"></p>
```
sampoder:
```suggestion
    <p x-text="(usd_quote == null || currency == null || amount == null) ? 'Enter an amount & currency to get an estimate in USD' : Number.isNaN(usd_quote) ? 'Loading quote from Wise...' : `Approximately ${usd_quote} USD`" class="muted mt-0"></p>
```

sampoder (commented):
looking good - just a couple of comments!

app/controllers/admin_controller.rb
```diff
@@ -40,6 +40,14 @@ def transaction
...
+
+      if potential_wise_transfers.size == 1
```
northeastprince:
```suggestion
      if potential_wise_transfers.count.one?
```

app/controllers/admin_controller.rb
YodaLightsabr: Oops, I fixed this locally but forgot to push

app/models/wise_transfer.rb
YodaLightsabr: ahhh camel case in a Ruby app is so annoying. I think I skipped the underscore because they refer to it as `PAYIN` in all caps without an underscore, but I agree that it's much cleaner with one.

app/views/events/_transfer_form.html.erb
YodaLightsabr: I see where you're coming from. The intent was to clarify the available destinations for all of them; is there another way of saying that that fits in better with the other labels?

app/views/wise_transfers/new.html.erb
YodaLightsabr: Good way of putting this, thanks.

app/views/events/_transfer_form.html.erb
YodaLightsabr: That's what I was thinking originally, but it might be strange for organizations outside of the Untied States. Maybe we conditionally render it? Not sure what the best option is here.

app/views/events/_transfer_form.html.erb
Luke-Oldenburg: Not sure but I also prefer the original.

app/views/events/_transfer_form.html.erb
sampoder: I think we should design our copy for US-based organisations, all orgs are technically operating in the US. We could do a conditional.

app/views/events/_transfer_form.html.erb
sampoder: To another HCB organization - though I'm not sure if we need a consistent label

YodaLightsabr: `Another HCB organization` sounds like it would work well?

app/assets/images/icons/payment-transfer.svg
YodaLightsabr, line 1: This updates the icon to add `fill="currentColor"` so that non-black colors work

sampoder (approved):
Tested E2E - lgtm!

leowilkin:
best PR of the year ~ operations

PR #11169, opened 2025-07-31
[Announcement] Fix bugs in monthly announcement job

polypixeldev:
## Summary of the problem

`Announcement::MonthlyJob` had a couple bugs:
1. it was using `publish!` instead of `mark_published!`
2. If a monthly announcement was not edited, it would still be a template draft when the job ran, which was not a valid transition

## Describe your changes

1. Change the job to use `mark_published!`
2. Allow template drafts to be published in the AASM configuration for `Announcement`


sampoder (approved):
lgtm!

PR #11177, opened 2025-08-01
Allow non-PII public info in v4

Mohamad-Mortada:
**Please look through this carefully**

## Summary of the problem
If you fetch any public org you aren't in you get a 403. It should return a public version like v3 does.

## Describe your changes
I removed the show in v4 policy check in events controller and individually added policy checks for PII in the view.


app/views/api/v4/transactions/_card_charge.json.jbuilder
```diff
@@ -3,7 +3,7 @@
 stripe_transaction = hcb_code.ct&.raw_stripe_transaction&.stripe_transaction
 stripe_authorization = hcb_code.pt&.raw_pending_stripe_transaction&.stripe_transaction
 
-json.merchant do
+json.merchant if policy(hcb_code.local_hcb_code).show? do
```
Luke-Oldenburg, line 6: Could we move this if statement to the outside?

app/views/api/v4/transactions/_donation.json.jbuilder
```diff
@@ -3,7 +3,7 @@
 json.recurring donation.recurring?
 json.donor do
   json.name donation.name
-  json.email donation.email
+  json.email donation.email if policy(donation.local_hcb_code).show?
```
Luke-Oldenburg, line 6: Can we make a specific policy for this instead of using the HCB code?

app/views/api/v4/transactions/_invoice.json.jbuilder
```diff
@@ -9,5 +9,5 @@ json.due_date invoice.due_date.to_date
 json.sponsor do
   json.id invoice.sponsor.public_id
   json.name invoice.sponsor.name
-  json.email invoice.sponsor.contact_email
+  json.email invoice.sponsor.contact_email if policy(invoice.local_hcb_code).show?
```
Luke-Oldenburg, line 12: Can we make a specific policy for this instead of using the HCB code?

app/controllers/api/v4/events_controller.rb
Luke-Oldenburg, line 1: Should the `show_in_v4?` policy be changed instead of updating this to use `show?`?

Mohamad-Mortada:
#11889 Not planned for the time being :(

PR #11179, opened 2025-08-01
Use strings for `start_date` fallback

sampoder:
This maintains consistency with the non-nils

### DIFF
--- a/app/models/announcement/block/donation_summary.rb
+++ b/app/models/announcement/block/donation_summary.rb
@@ -52,6 +52,7 @@ class Announcement
           DateTime.parse(self.parameters["start_date"])
         else
           self.parameters["start_date"] ||= 1.month.ago.to_s
+          1.month.ago
         end
       end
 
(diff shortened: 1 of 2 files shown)


app/models/announcement/block/donation_summary.rb
```diff
@@ -51,7 +51,7 @@ def start_date_param
         if self.parameters["start_date"].present?
           DateTime.parse(self.parameters["start_date"])
         else
-          self.parameters["start_date"] ||= 1.month.ago
+          self.parameters["start_date"] ||= 1.month.ago.to_s
```
polypixeldev, line 54:
```suggestion
          self.parameters["start_date"] ||= 1.month.ago.to_s
          1.month.ago
```
The `start_date_param` method should still return a `DateTime` to be consistent with the non-nils

polypixeldev (commented):
Thanks for fixing this! This should also be done for `end_date_param` in https://github.com/hackclub/hcb/blob/main/app/models/concerns/announcement/block/has_end_date.rb

app/models/announcement/block/donation_summary.rb
sampoder: good catch yes

sampoder:
looks like it's already done on that file?

polypixeldev:
Ah, it just needs to return `Time.now` in the else branch instead of the string (although maybe that should be `DateTime` instead of `Time`)

PR #11190, opened 2025-08-01
Add `is_reauthentication` to `logins`

davidcornu:
## Summary of the problem

Extracted from https://github.com/hackclub/hcb/pull/11162.

## Describe your changes

Adds a new `is_reauthentication` column which will eventually replace our use of `initial_login_id`.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # This file is auto-generated from the current state of the database. Instead
 # of editing this file, please use the migrations feature of Active Record to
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
```
sampoder: Let's please keep this in

sampoder (approved):
lgtm!

db/schema.rb
davidcornu: Oh whoops. Good catch

PR #11201, opened 2025-08-02
Fix sub-organizations not showing for auditors

Luke-Oldenburg:
## Summary of the problem

The policy is checking `admin_or_reader?` instead of `auditor_or_reader?`


sampoder (approved):
lgtm!

PR #11206, opened 2025-08-02
Fix comment policy

Mohamad-Mortada:
## Summary of the problem
Originally the v4 was checking `auditor_signed_in ` but method is undefined. Also the comment policy doesn't reference anything about checking if you can make an admin only comment. 

## Describe your changes
Moved policy logic to the policy file instead of in controller.


sampoder (approved):
lgtm!

PR #11211, opened 2025-08-04
Correctly handle validation errors for card grants

davidcornu:
## Summary of the problem

- https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/1486/samples/timestamp/2025-08-04T14:08:26Z
- https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/1502/samples/last

When validation fails for a card grant we
- Fire an AppSignal error regardless of what the error was
- Lose the form state by redirecting back to the transfers page instead of letting the user fix their errors

## Describe your changes

See individual commits

### DIFF
--- a/app/controllers/card_grants_controller.rb
+++ b/app/controllers/card_grants_controller.rb
@@ -16,8 +16,7 @@ class CardGrantsController < ApplicationController
 
     @prefill_email = params[:email]
 
-    @event.create_card_grant_setting unless @event.card_grant_setting.present?
-
+    @event.create_card_grant_setting! unless @event.card_grant_setting.present?
 
     @card_grant.amount_cents = params[:amount_cents] if params[:amount_cents]
   end
--- a/app/controllers/card_grants_controller.rb
+++ b/app/controllers/card_grants_controller.rb
@@ -18,7 +18,6 @@ class CardGrantsController < ApplicationController
 
     @event.create_card_grant_setting unless @event.card_grant_setting.present?
 
-    last_card_grant = @event.card_grants.order(created_at: :desc).first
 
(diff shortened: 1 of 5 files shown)


app/controllers/card_grants_controller.rb
```diff
@@ -29,14 +25,34 @@ def create
...
+      # number of other methods (e.g. `save!`) which either succeed or raise.
+      @card_grant.save!
+    rescue => e
+      case e
+      when ActiveRecord::RecordInvalid
+        # We expect to encounter validation errors from `CardGrant`, but anything
+        # else is the result of downstream logic which shouldn't fail.
+        raise e unless e.record.is_a?(CardGrant)
+
+        flash[:error] = @card_grant.errors.full_messages.to_sentence
```
sampoder, line 40: perfect -  thank you!

app/controllers/card_grants_controller.rb
```diff
@@ -29,14 +25,34 @@ def create
...
+      when ArgumentError
+        # `DisbursementService::Create` will raise `ArgumentError` if there are
+        # insufficient funds.
+        raise e unless e.message.start_with?("You don't have enough money")
```
sampoder, line 44: Is there a safer way to do this incase we change this message? Alright if not - just came to mind.

sampoder (approved):
code lgtm! tiny comments - not blocking

app/controllers/card_grants_controller.rb
davidcornu: yeah I don't like this. I think it's ok as a short term solution as we have tests to make sure we are catching these but I'll follow up with another PR to switch these to a custom exception class and use that to distinguish them here.

PR #11214, opened 2025-08-04
Decline cash withdrawals over $500

davidcornu:
## Summary of the problem

https://hackclub.slack.com/archives/C026RKHLPNJ/p1754325378783549

We are currently relying on the limit enforced by Stripe, which can change. As a precautionary measure we're doing to cap cash withdrawals to $500 (on cards that allow them) for now and introduce a more granular setting at a later date.

## Describe your changes

- Moves the `:stripe_authorization` factory to `spec/factories`, adds a `:cash_withdrawal` trait, and makes `pending_amount` configurable
- Adds test coverage for the current withdrawal behaviour (there wasn't any before)
- Updates the logic to cap the amount to $500


sampoder (approved):
Perfect! Thank you

PR #11216, opened 2025-08-04
Remove n+1 in `EventsController#index`

davidcornu:
## Summary of the problem

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/performance/incidents/7/samples/6596247683eb67648f30f807-126952945119088104601754329320

This endpoint can be a bit slow when loaded by an admin or auditor.

## Describe your changes

- Adds test coverage
- Improves `:organizer_position` factory so it doesn't create extra users and events
- Refactors `#index` to 
    - remove duplication in the serialization logic
    - preload the `config` relation 
    - add `strict_loading` to the loaded events to help prevent future regressions

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -54,8 +54,6 @@ class EventsController < ApplicationController
           )
         end
 
-        response.content_type = "text/json"
-
         render json: events
       end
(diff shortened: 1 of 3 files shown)


sampoder (approved):
thank you - appreciate this fix!

PR #11218, opened 2025-08-04
Set revision for AppSignal

davidcornu:
## Summary of the problem

We aren't currently reporting the code revision to AppSignal which makes historical stats harder to reason about.

## Describe your changes

Added `revision` key to `config/appsignal.yml` as per https://docs.appsignal.com/guides/deploy-markers.html#ruby.

### DIFF
--- a/config/appsignal.yml
+++ b/config/appsignal.yml
@@ -18,4 +18,4 @@ default: &defaults
 production:
   <<: *defaults
   active: true
-  revision: "<%= `git rev-parse --short HEAD`.strip %>"
+  revision: "<%= ENV["HATCHBOX_REVISION"] %>"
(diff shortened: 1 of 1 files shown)


sampoder (approved):

sampoder:
thank you!

davidcornu:
hmm I'm not sure this is going to work on hatchbox as I don't believe the app directory is a git repo

config/appsignal.yml
```diff
@@ -18,3 +18,4 @@ default: &defaults
 production:
   <<: *defaults
   active: true
+  revision: "<%= ENV["HATCHBOX_REVISION"] %>"
```
davidcornu, line 21: https://www.hatchbox.io/announcements/22 - confirmed this is available in production

PR #11222, opened 2025-08-04
Show whether a grant is one time use in grant details partial

polypixeldev:
## Summary of the problem

Users given a grant should be able to see if their grant is one time use - currently this is only shown to organizers in the manage grant overview.

## Describe your changes

Added one time use with a tooltip explaining it to the grant details partial

image

### DIFF
--- a/app/views/card_grants/_grant_details.html.erb
+++ b/app/views/card_grants/_grant_details.html.erb
@@ -20,9 +20,9 @@
           <p>
             <strong class="inline-flex items-center" style="gap: 4px">
               One time use
-              <%= link_to "#", class: "info tooltipped tooltipped--e", "aria-label": card_grant.one_time_use ? "This card will automatically freeze after you make a purchase" : "There is no limit on the number of purchases you can make using this grant", style: "display: flex;", data: { turbo: false } do %>
+              <span class="info tooltipped tooltipped--e" aria-label="This card will automatically freeze after you make a purchase"> 
                 <%= inline_icon "info", size: 18, style: "position: unset;" %>
-              <% end %>
+              </span>
             </strong>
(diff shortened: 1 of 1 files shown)


app/views/card_grants/_grant_details.html.erb
```diff
@@ -16,6 +16,16 @@
           <span><%= format_date card_grant.expires_on %></span>
         </p>
 
+        <p>
+          <strong class="inline-flex items-center" style="gap: 4px">
+            One time use?
+            <%= link_to "#", class: "info tooltipped tooltipped--e", "aria-label": card_grant.one_time_use ? "This card will automatically freeze after you make a purchase" : "There is no limit on the number of purchases you can make using this grant", style: "display: flex;", data: { turbo: false } do %>
```
davidcornu: - **concern**: Why are we using a `link_to` here? This isn't actually a link so a `<span>` should suffice
- **concern**: Why use `style: "display: flex;"` vs. the `flex` CSS class?

app/views/card_grants/_grant_details.html.erb
```diff
@@ -16,6 +16,16 @@
           <span><%= format_date card_grant.expires_on %></span>
         </p>
 
+        <p>
+          <strong class="inline-flex items-center" style="gap: 4px">
+            One time use?
```
Luke-Oldenburg: Personally I think we should just render this if it is one time use and drop the "?"

app/views/card_grants/_grant_details.html.erb
polypixeldev: good point! I had copied the tooltip from the other detail tooltips but just the span works

sampoder (approved):
lgtm!

PR #11228, opened 2025-08-05
Improve shipping logic on cards

Luke-Oldenburg:
## Summary of the problem

Sometimes cards are delivered before Stripe reports them as delivered. We should also be checking if the card has been activated to account for these cases and hide the shipping message.

## Describe your changes

Checks if the card has been initially activated.


app/models/stripe_card.rb
```diff
@@ -71,7 +71,7 @@ class StripeCard < ApplicationRecord
   scope :frozen, -> { where(stripe_status: "inactive", initially_activated: true) }
   scope :active, -> { where(stripe_status: "active") }
   scope :inactive, -> { where(stripe_status: "inactive", initially_activated: false) }
-  scope :physical_shipping, -> { physical.includes(:user, :event).reject { |c| c.stripe_obj[:shipping][:status] == "delivered" } }
+  scope :physical_shipping, -> { physical.includes(:user, :event).reject { |c| c.stripe_obj[:shipping][:status] == "delivered" || !c.initially_activated } }
```
sampoder: Why is the condition negated?

Luke-Oldenburg: It's a .reject

sampoder: Shouldn't we reject initially activated cards?

Luke-Oldenburg: Oh yea you're right.

app/models/stripe_card.rb
```diff
@@ -71,7 +71,7 @@ class StripeCard < ApplicationRecord
   scope :frozen, -> { where(stripe_status: "inactive", initially_activated: true) }
   scope :active, -> { where(stripe_status: "active") }
   scope :inactive, -> { where(stripe_status: "inactive", initially_activated: false) }
-  scope :physical_shipping, -> { physical.includes(:user, :event).reject { |c| c.stripe_obj[:shipping][:status] == "delivered" } }
+  scope :physical_shipping, -> { physical.includes(:user, :event).reject { |c| c.stripe_obj[:shipping][:status] == "delivered" || c.initially_activated || c.shipping_eta&.past? } }
```
davidcornu: **concern**: I would not consider this a `scope` as `reject` is causing query to be executed and subsequently filtered in-memory (and make HTTP requests to retrieve `stripe_obj`!)

Might I suggest


```suggestion
  def self.physical_shipping # <- needs a better name
    physical
      .includes(:user, :event) # <- not entirely sure why this is necessary
      .where(initially_activated: false) # <- exclude these at the DB level so we don't hit stripe more than necessary
      .reject { |c| c.stripe_obj[:shipping][:status] == "delivered" || c.shipping_eta&.past? }
  end
```

davidcornu (changes requested):

app/models/stripe_card.rb
davidcornu: Seems like we're also doing additional filtering in the controller (which AFAICT is the only call-site): https://github.com/hackclub/hcb/blob/58901bd6364df81046fd152717fc22e9f5cbc6e0/app/controllers/stripe_cards_controller.rb#L14-L16

Would it be worth moving that logic here?

app/controllers/stripe_cards_controller.rb
```diff
@@ -11,9 +11,12 @@ def index
...
   def shipping
     # Only show shipping for phyiscal cards if the eta is in the future (or 1 week after)
-    @stripe_cards = current_user.stripe_cards.where.not(stripe_status: "canceled").physical_shipping.filter do |sc|
-      sc.shipping_eta&.after?(1.week.ago)
-    end
+    @stripe_cards = current_user.stripe_cards.physical
+                                .where.not(stripe_status: "canceled")
+                                .where(initially_activated: false)
+                                .includes(:user, :event)
+                                .reject { |c| c.stripe_obj[:shipping][:status] == "delivered" || c.shipping_eta&.past? }
```
davidcornu: **concern**: As per https://github.com/hackclub/hcb/pull/11228#discussion_r2257844297, this contains a lot of implementation details that should ideally live in the `StripeCard` model (maybe as a class method as I had suggested)

Luke-Oldenburg: I figured it would be better to move it all to the controller since that is the only place we are calling it from (and that I anticipate us ever calling it from).

davidcornu: > I figured it would be better to move it all to the controller since that is the only place we are calling it from (and that I anticipate us ever calling it from).

All the more reason to put in the model where people might notice it if they're making changes. There are a lot of details about the schema and the structure of `stripe_obj` that ideally the controller shouldn't know about (see https://en.wikipedia.org/wiki/Law_of_Demeter).

davidcornu (approved):

PR #11229, opened 2025-08-05
Hide unpublished monthly announcements

Luke-Oldenburg:
## Summary of the problem

Currently, after editing a monthly announcement, it will appear in the announcement list. We don't want this to happen because it should only be linked from the callout at the top of the page until it is published.

## Describe your changes

Remove announcements from the saved scope if their template type is monthly and their published at is nil.

### DIFF
--- a/app/models/announcement.rb
+++ b/app/models/announcement.rb
@@ -58,7 +58,7 @@ class Announcement < ApplicationRecord
   scope :monthly_for, ->(date) { monthly.where("announcements.created_at BETWEEN ? AND ?", date.beginning_of_month, date.end_of_month) }
   validate :content_is_json
 
-  scope :saved, -> { where.not(aasm_state: :template_draft).where.not(content: {}) }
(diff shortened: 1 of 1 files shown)


app/models/announcement.rb
```diff
@@ -58,7 +58,7 @@ class Announcement < ApplicationRecord
   scope :monthly_for, ->(date) { monthly.where("announcements.created_at BETWEEN ? AND ?", date.beginning_of_month, date.end_of_month) }
   validate :content_is_json
 
-  scope :saved, -> { where.not(aasm_state: :template_draft).where.not(content: {}) }
+  scope :saved, -> { where.not(aasm_state: :template_draft).where.not(content: {}).where.not(template_type: Announcement::Templates::Monthly.name, published_at: nil) }
```
davidcornu, line 61: **concern**: This feels very specific but is probably fine for now/ We should revisit if there are other template types that require this behaviour.

davidcornu (approved):

PR #11230, opened 2025-08-05
Add test coverage for `OrganizerPositionInvitesController#create`

davidcornu:
## Summary of the problem

Related to https://github.com/hackclub/hcb/pull/10543/files#r2243532219

Ideally we'd like both the regular and API endpoints for invitations to share as much code as possible so they don't accidentally diverge. However, doing that kind of refactor is tricky if we don't have a test suite to lean on.

## Describe your changes

- Adds [WebMock](https://github.com/bblimke/webmock)
- Adds test coverage for `OrganizerPositionInvitesController#create`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -151,6 +151,7 @@ end
 group :test do
   gem "factory_bot_rails" # Test data
   gem "simplecov", require: false # Code coverage
+  gem "webmock"
 end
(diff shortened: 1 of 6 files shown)


sampoder (approved):
awesome!

PR #11237, opened 2025-08-06
Make sure unsuccessful HTTP calls with Faraday result in exceptions

davidcornu:
## Summary of the problem

In several places throughout the code base we make use of `Faraday` as an HTTP client to make external HTTP calls, but we either
- Fail to check the response code
- Don't configure the `:raise_errors` middleware (https://lostisland.github.io/faraday/#/middleware/included/raising-errors)

## Describe your changes

Audited every call to `Faraday.new` and made sure we were doing one of the above.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -1436,6 +1436,15 @@ class AdminController < ApplicationController
     9999 # return something invalidly high to get the ops team to report it
   end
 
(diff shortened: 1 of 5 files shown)


sampoder (approved):
code lgtm! though I'm fairly sure the Airtable tasks were raising errors already. @rluodev can you review the Loops code?

PR #11247, opened 2025-08-07
[Proposal] Add a `hcb_code_type` SQL function to make spelunking easier

davidcornu:
## Summary of the problem

Our transaction modelling relies quite heavily on the concept of an HCB code which encodes the type of transaction: https://github.com/hackclub/hcb/blob/5a5cd128c204caccaec8fecc898f91f61d687e77/app/services/transaction_grouping_engine/calculate/hcb_code.rb#L5-L34

When digging around in Blazer, it would be handy to have a quick way to split out the transaction type so we don't have to memorize these codes.

## Describe your changes

- Adds the `fx` [gem](https://github.com/teoljungberg/fx) which adds migration + schema support for PostgreSQL functions
- Adds a new `hcb_code_type` function
- Adds tests to make sure the Ruby and SQL versions don't accidentally diverge


Gemfile
```diff
@@ -11,6 +11,7 @@ gem "rails", git: "https://github.com/rails/rails.git", branch: "7-2-stable"
 gem "puma", "~> 6.6" # app server
 
 gem "pg", ">= 0.18", "< 2.0" # database
+gem "fx"
```
davidcornu, line 14: We could forgo this but it would require switching up the schema format (https://guides.rubyonrails.org/active_record_migrations.html#using-the-sql-schema-dumper).

sampoder (approved):
thank you - this is cool! only question is here do we add new transaction types, do we need a migration?

sampoder (approved):
thank you - this is cool! only question is here do we add new transaction types, do we need a migration?

davidcornu:
> thank you - this is cool! only question is here do we add new transaction types, do we need a migration?

Yep although

1. This is here to help with Blazer queries and such so it isn't critical to the proper functioning of the application
2. That's why I added a test that checks against the constants in `TransactionGroupingEngine::Calculate::HcbCode` so we'd get a failure in that scenario and would know to add the new type.

davidcornu:
> 2\. That's why I added a test that checks against the constants in `TransactionGroupingEngine::Calculate::HcbCode` so we'd get a failure in that scenario and would know to add the new type.

Case in point: https://github.com/hackclub/hcb/actions/runs/16833790146/job/47687974862?pr=11247#step:13:40

CleanShot 2025-08-08 at 11 34 22

Will add

PR #11251, opened 2025-08-07
[Annnouncement] Fix edit policy for announcements generated by HCB

polypixeldev:
## Summary of the problem

Announcements generated by HCB (like some prompted announcements or monthly announcements) currently cannot be edited by non-admins since the policy only allowed for the announcement author to edit.

## Describe your changes

Added a check that allows any manager to edit/update announcements generated by HCB, and updated the announcement show page to reference the correct policies on actions.


app/policies/announcement_policy.rb
```diff
@@ -14,7 +14,7 @@ def show?
   end
 
   def edit?
-    (manager? && record.author == user) || admin?
+    (manager? && (record.author == user || record.author == User.system_user)) || admin?
```
sampoder, line 19: Could you leave a comment to explain this?

app/views/announcements/show.html.erb
```diff
@@ -26,22 +26,26 @@
   </p>
 
   <div class="flex gap-2">
-    <% if ((current_user == @announcement.author && organizer_signed_in?(as: :manager)) || auditor_signed_in?) && !editing %>
-      <% if @announcement.draft? %>
+    <% if !editing %>
```
sampoder: can we use `unless`?

app/views/announcements/show.html.erb
```diff
@@ -26,22 +26,26 @@
...
+      <% if @announcement.draft? && (policy(@announcement).publish? || auditor_signed_in?) %>
```
sampoder, line 30: hmm - why are we showing this to auditors but not giving them permission?

polypixeldev: My thought process was that in most places on HCB, auditors can see all actions available but the buttons are disabled for them if they do not have permission, so I did the same for these buttons.

sampoder (approved):

PR #11252, opened 2025-08-07
Wise bug fixes part 1

davidcornu:
## Summary of the problem

Working through the bugs @leowilkin found

## Describe your changes

- Capitalize "Wise"
- Actually store the `wise_id` that's included in the form

### DIFF
--- a/app/models/wise_transfer.rb
+++ b/app/models/wise_transfer.rb
@@ -210,7 +210,10 @@ class WiseTransfer < ApplicationRecord
     update!(quoted_usd_amount_cents: estimated_usd_amount_cents) unless quoted_usd_amount_cents.present?
   end
 
-  WISE_ID_REGEXP = /wise.com\/transactions\/activities\/by-resource\/transfer\/(\d+)/i
+  WISE_ID_REGEXPS = [
+    /wise\.com\/transactions\/activities\/by-resource\/transfer\/(\d+)/i,
+    /wise\.com\/success\/transfer\/(\d+)/i
+  ].freeze
 
   def normalize_wise_id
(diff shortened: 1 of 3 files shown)


app/models/wise_transfer.rb
```diff
@@ -207,4 +210,20 @@ def generate_quote!
     update!(quoted_usd_amount_cents: estimated_usd_amount_cents) unless quoted_usd_amount_cents.present?
   end
 
+  WISE_ID_REGEXP = /wise.com\/transactions\/activities\/by-resource\/transfer\/(\d+)/i
+
+  def normalize_wise_id
```
YodaLightsabr, line 218: Aren't we doing this on the front end?

YodaLightsabr (approved):

app/models/wise_transfer.rb
leowilkin: the link i'd paste in follows this schema: https://wise.com/success/transfer/{wisetransferid} if that helps...

davidcornu: Ah yeah we are https://github.com/hackclub/hcb/blob/3fd965acd69eeb4d3268f2f1a14cfcedebf27e24/app/views/admin/wise_transfer_process.html.erb#L180-L186

but also, never trust the front-end alone

davidcornu: > the link i'd paste in follows this schema: [https://wise.com/success/transfer/{wisetransferid}](https://wise.com/success/transfer/%7Bwisetransferid%7D) if that helps...

@leowilkin oh I don't think we handle this one. Let me add

PR #11253, opened 2025-08-07
Fix stripe card defrost policy

### LINKED ISSUE #10682: [Stripe Cards] Improve papertrail
- [ ] determine why so many cards are missing a last frozen by
- [x] make column on DB for last frozen by
- [x] backfill this with the data from papertrail
- [x] fix records that are missing a last frozen by
- [x] ignore cards that are missing a last frozen by / last frozen by is nil in defrost policy

### LINKED ISSUE #10749: [StripeCardPolicy] Ignore cards that don't have a papertrail in defrost

Luke-Oldenburg:
## Summary of the problem

Cards that didn't have a papertrail set were not able to be unfrozen by the cardholder before.
Closes #10682 #10749

## Describe your changes

Check for a presence of a papertrail.

### DIFF
--- a/app/policies/stripe_card_policy.rb
+++ b/app/policies/stripe_card_policy.rb
@@ -6,7 +6,7 @@ class StripeCardPolicy < ApplicationPolicy
   end
 
   def shipping?
-    user&.auditor? || OrganizerPosition.role_at_least?(user, record&.event, :reader)
+    user&.auditor? || OrganizerPosition.role_at_least?(user, record.event, :reader)
   end
 
   def freeze?
@@ -14,8 +14,8 @@ class StripeCardPolicy < ApplicationPolicy
   end
 
   def defrost?
-    return false if record&.event&.financially_frozen?
-    return false if record&.last_frozen_by.present? && record&.last_frozen_by != user && !admin_or_manager?
+    return false if record.event&.financially_frozen?
+    return false if record.last_frozen_by.present? && record.last_frozen_by != user && !admin_or_manager?
 
     freeze?
   end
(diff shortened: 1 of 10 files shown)


app/policies/stripe_card_policy.rb
```diff
@@ -14,19 +14,18 @@ def freeze?
...
 
   def defrost?
-    unless record.last_frozen_by == user || admin_or_manager?
-      return false
-    end
+    return false if record&.event&.financially_frozen?
+    return false if record&.last_frozen_by.present? && record&.last_frozen_by != user && !admin_or_manager?
 
-    freeze? && !record.event.financially_frozen?
+    freeze?
```
sampoder: thank you! a lot cleaner

app/policies/stripe_card_policy.rb
```diff
@@ -68,11 +67,11 @@ def cardholder?
   end
 
   def admin_or_manager?
-    user&.admin? || OrganizerPosition.find_by(user:, event: record.event)&.manager?
+    user&.admin? || OrganizerPosition.find_by(user:, event: record&.event)&.manager?
   end
 
   def grantee?
-    cardholder? && record.card_grant
+    cardholder? && record&.card_grant
```
sampoder: when would record be `nil`?

Luke-Oldenburg: It might not be but I saw the & in some places so I standardized it throughout the file.

sampoder: Can we drop it in all places?

sampoder: Can we drop it in all places?

Luke-Oldenburg: Sure

sampoder (approved):
lgtm!

PR #11254, opened 2025-08-07
[Invoice] Add new `deposited` state to prepare moving to AASM

### LINKED ISSUE #4992: Invoice: Fix state management

rluodev:
## Summary of the problem

lots of random columns

## Describe your changes

this centralizes those into AASM. Next PR will be to migrate all AASM states with a migration and change any pages to use the correct state.


app/models/invoice.rb
```diff
@@ -152,7 +152,10 @@ class NoAssociatedStripeCharge < StandardError; end
   aasm timestamps: true do
     state :open_v2, initial: true
     state :paid_v2
+    state :deposited_v2
+    state :past_due_v2
```
sampoder: does it make sense for this to be a state? do we differ in our behaviour between past due invoices vs open invoices

rluodev: tbf not really

rluodev: fine with removing it if you are?

sampoder: Yes - please remove it

rluodev:
```suggestion
    state :deposited_v2
```

rluodev: done!

sampoder (approved):

sampoder (dismissed):
Actually I feel like archived should be separate state

rluodev:
@sampoder wdym?

garyhtou:
Note from call with Ruien: I believe what @sampoder meant here is that archived should not be an AASM state. This is because archived is an "addition to a state". An invoice would be archived in addition to open/paid/deposited, etc.

garyhtou (approved):

PR #11255, opened 2025-08-07
[Wise] Add a field to track the recipient ID

davidcornu:
## Summary of the problem

We want to keep track of the Wise recipient ID when marking transfers as complete.

## Describe your changes

- Adds a `wise_recipient_id` column to the `wise_transfers` table (nomenclature comes from https://docs.wise.com/api-docs/api-reference/recipient#get)
- Adds validation and normalization for the `wise_recipient_id` field to the `WiseTransfer` model (similar to #11252)
- Adds a new field to the completion form
    CleanShot 2025-08-07 at 15 28 00@2x
- Adds a link to the recipient page when the ID is populated
    CleanShot 2025-08-07 at 15 36 46@2x

### DIFF
--- a/app/views/hcb_codes/_wise_transfer.html.erb
+++ b/app/views/hcb_codes/_wise_transfer.html.erb
@@ -1,6 +1,6 @@
-<% if @hcb_code.wise_transfer.wise_id.present? %>
+<% if @hcb_code.wise_transfer.wise_url %>
   <% admin_tool("mt3") do %>
-    <%= link_to "https://wise.com/transactions/activities/by-resource/TRANSFER/#{@hcb_code.wise_transfer.wise_id}", class: "btn bg-accent", target: "_blank" do %>
+    <%= link_to @hcb_code.wise_transfer.wise_url, class: "btn bg-accent", target: "_blank" do %>
       <%= inline_icon "wise" %>
       View transfer on Wise
     <% end %>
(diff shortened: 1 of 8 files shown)


leowilkin:
ty <3

db/schema.rb
```diff
@@ -11,6 +11,7 @@
 # It's strongly recommended that you check this file into your version control system.
 
 ActiveRecord::Schema[7.2].define(version: 2025_08_08_033500) do
+
```
sampoder: Should this be here?

sampoder (approved):
lgtm!

app/views/hcb_codes/_wise_transfer.html.erb
```diff
@@ -1,6 +1,6 @@
-<% if @hcb_code.wise_transfer.wise_id.present? %>
+<% if @hcb_code.wise_transfer.wise_url %>
```
YodaLightsabr, line 1: nit: do we still want to call `.present?` here

db/schema.rb
YodaLightsabr: +1

YodaLightsabr (approved):
very clean!

db/schema.rb
davidcornu: It should definitely not. I must have added it when fixing a merge conflict.

app/views/hcb_codes/_wise_transfer.html.erb
davidcornu: We do already in the implementation of `wise_url` so didn't think it necessary. Happy to add it if you think it's clearer.

PR #11257, opened 2025-08-07
[Announcement] Cancel monthly announcements with no data

polypixeldev:
## Summary of the problem

Right now, monthly announcements are enabled for a lot of transparent organizations, including some that may not have any activity. We don't want to send out all these monthly announcements with empty data, like if the organization has no donations.

## Describe your changes

Any announcements that are still template drafts at the end of the month will not be published. At the 7 day mark, monthly announcements that have data in all of their blocks will be promoted to drafts, which will allow them to be sent. All other monthly announcements will be left as template drafts and a notification will be sent out to managers that if they do not edit the announcement, it will not be sent out. This is also displayed in a callout on the announcements page.

image


sampoder (approved):
thank you!

PR #11258, opened 2025-08-07
Render Wise transfers in the correct currency

davidcornu:
## Summary of the problem

We are incorrectly rendering Wise transfer amounts in non-USD currencies as USD.

## Describe your changes

Add special handling for Wise transfers:
1. Render the ledger amount (same as wires)
2. If the transfer isn't in USD, show the original amount in a tooltip
    CleanShot 2025-08-11 at 13 30 08@2x


app/views/events/transfers.html.erb
```diff
@@ -135,8 +135,27 @@
...
+                  <% else %>
+                    No USD amount
+                  <% end %>
+                <% end %>
+
+                <span class="tooltipped tooltipped--w inline-block" aria-label="<%= fx_tooltip %>">
+                  <%= d.amount.format(with_currency: true) %>
+                </span>
+              <% end %>
             <% else %>
```
sampoder: I feel like the pattern should match the wire pattern?

davidcornu: That was my first thought but I wasn't sure `d.local_hcb_code.amount_cents` would be populated.

1. The only non-nullable field on `WiseTransfer` is `amount_cents` which can be in any number of currencies and is populated when the transfer is created
    - `_cents` is a misnomer as it technically stores [subunits](https://en.wikipedia.org/wiki/Denomination_(currency)#Subunit_and_super_unit) (e.g. JPY 1000 would be stored as 1000 as it is non-fractional)
3. `quoted_usd_amount_cents` (nullable) is populated after a call to the Wise API (but might change?)
4.  `usd_amount_cents` (nullable) is populated by ops when the transfer is processed

As far as I can tell we don't have any validation on the presence of either (3) or (4). `quoted_usd_amount_cents` is populated on creation and synced to the canonical pending transaction  and subsequently replaced by `usd_amount_cents` (_if present_).

So _in theory_, we should be fine relying on the canonical pending transaction but

1. I would like to see stronger validations for all of the above
2. We would probably want to show the original currency (maybe in a tooltip) as that's what the user entered.

sampoder: +1 on validations but `d.local_hcb_code.amount_cents` should always be set. I'm cool with either product direction - @YodaLightsabr got thoughts? I'd just like us to be consistent with what we do for wires.

YodaLightsabr: I think I would prefer making the USD amount more prominent, but we should show the local amount in a tooltip. I will say that wherever we're rendering the USD amount, we should use the HCB code amount because that will always match the transaction amount on the ledger.

> _cents is a misnomer as it technically stores [subunits](https://en.wikipedia.org/wiki/Denomination_(currency)#Subunit_and_super_unit) (e.g. JPY 1000 would be stored as 1000 as it is non-fractional)

This is true, however I believe this is the terminology used by the `money-rails` gem.

davidcornu: Updated

sampoder (approved):

YodaLightsabr (approved):

YodaLightsabr (approved):

PR #11259, opened 2025-08-07
Give each CardGrant its own CardGrantSetting (Part 1)

Luke-Oldenburg:


db/schema.rb
davidcornu, line 1: **concern**: I'm a bit concerned about this approach as we now have

`event_id` | `card_grant_id` | &nbsp;
---- | ---- | ----
`NULL` | ✅ | Applies to card grant
✅ | `NULL` | Applies to event
`NULL` | `NULL` | ???
✅ | ✅ | ???

My recommendation here would be to keep `event_id` as `NOT NULL` (as card grants are specific to an event and can't be moved between events), which reduces the state space considerably 

`card_grant_id` | &nbsp;
---- | ----
 ✅ | Applies to card grant
`NULL` | Applies to event

app/jobs/one_time_jobs/backfill_card_grant_settings.rb
```diff
@@ -0,0 +1,26 @@
...
+                                       banned_categories: cg.banned_categories || default_card_grant_setting.banned_categories,
+                                       banned_merchants: cg.banned_merchants || default_card_grant_setting.banned_merchants,
+                                       category_lock: cg.category_lock || default_card_grant_setting.category_lock,
+                                       expiration_preference: default_card_grant_setting.expiration_preference,
+                                       invite_message: default_card_grant_setting.invite_message,
+                                       keyword_lock: cg.keyword_lock || default_card_grant_setting.keyword_lock,
+                                       merchant_lock: cg.merchant_lock || default_card_grant_setting.merchant_lock,
+                                       pre_authorization_required: cg.pre_authorization_required || default_card_grant_setting.pre_authorization_required,
+                                       reimbursement_conversions_enabled: default_card_grant_setting.reimbursement_conversions_enabled
+                                     })
```
davidcornu: - **nit**: If you put the opening `{` on a new line, our Rubocop config will let you shift all of these lines to the left
- **concern**: Do any of these methods return booleans (e.g. `pre_authorization_required`?). If so the fallback logic isn't going to work as if the card-grant-specific setting is `false` we'll always check the fallback.

app/models/card_grant.rb
```diff
@@ -243,31 +245,31 @@ def create_stripe_card(session)
...
+    (banned_categories || setting&.banned_categories || default_setting&.banned_categories || []).uniq
   end
 
   def allowed_category_names
     allowed_categories.map { |category| YellowPages::Category.lookup(key: category).name || "#{category}*" }.uniq
   end
 
   def keyword_lock
-    super || setting&.keyword_lock
+    super || setting&.keyword_lock || default_setting&.keyword_lock
```
davidcornu: **concern**: What is `super` here? `CardGrant` inherits from `application_record` which AFAICT doesn't have any notion of `keyword_lock`.

app/models/card_grant.rb
davidcornu, line 1: **question**: Is the plan to eventually drop all the attributes on `CardGrant` in favour of using `card_grant_setting`?

app/models/card_grant.rb
```diff
@@ -243,31 +245,31 @@ def create_stripe_card(session)
...
   end
```
davidcornu: **suggestion**: We seem to have very similar fallback logic in several places. It might be worth extracting this handling into a separate class that the model can delegate to.

Something like

```ruby
class CardGrantSettingLookup
  def initialize(card_grant, card_grant_setting: nil, event_card_grant_setting: nil)
    # ...
  end

  def allowed_categories
    # ...
  end
end
```

app/models/card_grant.rb
Luke-Oldenburg: I was confused over that as well but was afraid to change preexisting code.

app/models/card_grant.rb
Luke-Oldenburg: Any values that can have a default configured should be moved over to the setting.

app/jobs/one_time_jobs/backfill_card_grant_settings.rb
Luke-Oldenburg: Good point. I'll add hardcoded fallbacks.

db/schema.rb
Luke-Oldenburg: How would I reference an events card grant setting? I can't make it a unique index so there's only one per event if we require event id for all of them.

davidcornu: > I can't make it a unique index so there's only one per event if we require event id for all of them.

I believe you can create a unique index over `event_id, card_grant_id` with `NULLS NOT DISTINCT` (https://www.postgresql.org/docs/15/sql-createindex.html) which will ensure there's only one row with `card_grant_id = null` per `event_id`.

app/jobs/one_time_jobs/backfill_card_grant_settings.rb
davidcornu: > Good point. I'll add hardcoded fallbacks.

I'm not sure I follow.

To illustrate my point, imagine we `cg.pre_authorization_required` is `false` but `default_card_grant_setting.pre_authorization_required` is true, we'll _always_ returns `true` as `false || true` evaluates to `true`.

Luke-Oldenburg: Ohh I didn't follow. How can I get around that without a bunch of nested ternaries calling `.present?`?

davidcornu: 🤔 something like this could do

```ruby
banned_categories: [cg.banned_categories, default_card_grant_setting.banned_categories].first(&:present?)
```

app/models/card_grant.rb
Luke-Oldenburg:
```suggestion
    setting&.keyword_lock || default_setting&.keyword_lock
```

app/models/card_grant.rb
Luke-Oldenburg: The values will eventually be dropped from the model. Do you think this is still helpful with it down to 2 locations then (card grant setting and event card grant setting)?

app/models/event.rb
```diff
@@ -419,6 +419,10 @@ def ancestor_organizer_positions
     self.activated_at = Time.now
   end
 
+  before_save if: -> { plan.card_grants_enabled? } do
+    create_card_grant_setting!
```
Luke-Oldenburg, line 423: Is this function auto-generated? I confused myself somewhere since working on this last.

db/schema.rb
Luke-Oldenburg: I'm not sure that it's possible to write this in pure Ruby.

app/models/card_grant.rb
Luke-Oldenburg: I think this logic is different enough that we don't need this especially since a lot will be ripped out in subsequent PRs.

app/jobs/one_time_jobs/backfill_card_grant_settings.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module OneTimeJobs
+  class BackfillCardGrantSettings
+    def self.perform
+      CardGrant.find_each do |cg|
+        default_cg_setting = CardGrantSetting.find_by(event_id: cg.id)
```
Luke-Oldenburg:
```suggestion
        default_cg_setting = CardGrantSetting.where(event_id: cg.id, card_grant_id: nil).first
```

app/jobs/one_time_jobs/backfill_card_grant_settings.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module OneTimeJobs
+  class BackfillCardGrantSettings
+    def self.perform
+      CardGrant.find_each do |cg|
+        default_cg_setting = CardGrantSetting.where(event_id: cg.id, card_grant_id: nil).first
```
Luke-Oldenburg:
```suggestion
        default_cg_setting = CardGrantSetting.where(event_id: cg.event.id, card_grant_id: nil).first
```

db/schema.rb
sampoder: yes - you can have a `has_one` with a scope:

```
has_one :totp, -> { where(aasm_state: :verified) }, class_name: "User::Totp", inverse_of: :user
```

app/models/event.rb
sampoder: yes it is

app/controllers/api/v4/card_grants_controller.rb
```diff
@@ -25,6 +27,23 @@ def create
...
+              expiration_preference: default_cg_setting.expiration_preference,
+              invite_message: default_cg_setting.invite_message,
+              keyword_lock: cg_params.keyword_lock,
+              merchant_lock: cg_params.merchant_lock,
+              pre_authorization_required: cg_params.pre_authorization_required,
+              reimbursement_conversions_enabled: default_cg_setting.reimbursement_conversions_enabled,
+              card_grant_id: @card_grant.id,
+              event_id: @event.id
+            }
+          )
```
sampoder, line 46: could we do this cleaner - maybe cloning the object?

app/mailers/card_grant_mailer.rb
```diff
@@ -3,7 +3,7 @@
 class CardGrantMailer < ApplicationMailer
   def card_grant_notification
     @card_grant = params[:card_grant]
-    @custom_invite_message = @card_grant.setting.invite_message
+    @custom_invite_message = @card_grant.invite_message
```
sampoder, line 6: why was this change made?

sampoder: oh I see

app/views/events/edit.html.erb
```diff
@@ -21,7 +21,7 @@
           <%= link_to "Reimbursements", edit_event_path(@event, tab: "reimbursements"), data: { turbo: true, turbo_action: "advance" } %>
         <% end %>
       <% end %>
-      <% if @event.card_grant_setting.present? %>
+      <% if @event.plan.card_grants_enabled? %>
```
sampoder, line 24: hmm - what happens if card_grant_setting is nil?

Luke-Oldenburg: It shouldn't be with this new system - I guess we could keep the old logic here.

db/schema.rb
Luke-Oldenburg: Sorry I meant creating the migration to add the index. How would I do that?

Luke-Oldenburg:
I think card grant settings are fine for now with the smaller improvements I just introduced.
https://github.com/hackclub/hcb/pull/11632
https://github.com/hackclub/hcb/pull/11633
https://github.com/hackclub/hcb/pull/11634

PR #11260, opened 2025-08-07
Remove Venmo references from account numbers page

davidcornu:
## Summary of the problem

Our mentioning of Venmo is causing a lot of confusion as HCB accounts can't be used as payment method.

## Describe your changes

Remove references to Venmo (as requested by @leowilkin)

### DIFF
--- a/app/views/events/_nav.html.erb
+++ b/app/views/events/_nav.html.erb
@@ -63,7 +63,7 @@
       selected: local_assigns[:selected] == :invoices if policy(@event).invoices? %>
     <%= dock_item "Account numbers",
       account_number_event_path(@event),
-      tooltip: "Receive payouts from GoFundMe, Shopify, Venmo, and more",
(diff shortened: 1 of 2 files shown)


sampoder:
HCB account numbers can be used to receive money from Venmo

app/views/events/_nav.html.erb
```diff
@@ -63,7 +63,7 @@
       selected: local_assigns[:selected] == :invoices if policy(@event).invoices? %>
     <%= dock_item "Account numbers",
       account_number_event_path(@event),
-      tooltip: "Receive payouts from GoFundMe, Shopify, Venmo, and more",
```
sampoder, line 66: This copy is correct

sampoder (changes requested):
I think the confusion here is not caused by Venmo but by the fact that our numbers are deposit only

PR #11269, opened 2025-08-08
Use ActiveJob::TrafficControl to throttle mail jobs

### LINKED ISSUE #11187: Improve handling of SMTP throttling
There are cases where we run into `Net::SMTPServerBusy` because we exceed the rate limit ([recent example](https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2091/samples/timestamp/2025-08-01T00:00:11Z)), and depending on how how the code is written this can result in duplicate emails being sent.

For example, https://github.com/hackclub/hcb/blob/3dbf2fcc7e9318170d20d6c69d420909ed0d4bb3/app/jobs/monthly_donation_summary_job.rb synchronously sends a monthly email for every `Event` with donations in the last month, and if it runs into `Net::SMTPServerBusy` Sidekiq ends up retrying the whole thing which causes us to potentially send the same email twice for the records that were processed before the throttle limit was reached.

One option that has been discussed is putting individual email sends on a single queue, which we can then constrain via Sidekiq's capsules (https://github.com/sidekiq/sidekiq/wiki/Advanced-Options#capsules) or a more advanced solution. This would let retries only impact a single email and provide a rudimentary solution to globally throttling outbound email.

Internal discussion: https://hackclub.slack.com/archives/C026RKHLPNJ/p1754018188244529

polypixeldev:
## Summary of the problem

We currently have no code in place to handle mail delivery failing due to our AWS SES rate limit, which is currently 14 deliveries per second. We run into this especially when sending out large amount of mail at once, such as with `MonthlyDonationSummaryJob`.

Closes #11187 

## Describe your changes

Uses the `activejob-traffic_control` gem to throttle all mail sent using `deliver_later` to 12/second with a custom mail delivery job - this leaves room for mail sent using `deliver_now`. If more than that many jobs try to execute per second, `ActiveJob::TrafficControl` will automatically reschedule the job to execute later. This also moves current mass email send jobs to use `deliver_later` and remove the current soluton using the `Limiter` gem.


sampoder (approved):
this code lgtm! can I tag in @davidcornu or @garyhtou for a second review? 

when do you need this to ship by?

config/initializers/active_job_traffic_control.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+ActiveJob::TrafficControl.client = if Rails.env.production?
+                                     Redis.new
+                                   else
+                                     Dalli::Client.new
+                                   end
```
sampoder: Can you set this to none if Rails.env.testing?

polypixeldev: Yep, checking for test environment in the mail delivery job instead of here - I had thought `Dalli` was an in-memory store but they're both external servers, so I just have to disable throttling in test environments

polypixeldev:
Yeah, happy to get a second review on this to make sure email sending won't break in prod. Would like to get it shipped soon so that we can merge and send out #11086

PR #11272, opened 2025-08-08
[DX] Monkey-patch Rails' schema dumper to add `# frozen_string_literal: true`

davidcornu:
## Summary of the problem

1. The majority opinion seems to be that we should keep the `# frozen_string_literal: true` comment around in `db/schema.rb`
2. Rails doesn't add it by default so this means wasting brain cycles keeping it there

## Describe your changes

Patch the schema logic to add it by default.


polypixeldev:
This is great, tysm!

sampoder (approved):
bless

YodaLightsabr (approved):
Thank you! 🙏

PR #11273, opened 2025-08-08
[Contract] Redirect user to contract if it needs signing

### LINKED ISSUE #11030: [Organizer Position Contract] Redirect to contract if it needs signing
If an invite can't be accepted because it's waiting on contract signature, then:

1. Show a message saying "This invite can't be accepted until X signs the contract"
2. If the current user who still needs to sign, then add a button that redirects to the contact signing url.

Sarvesh-Mk:
## Summary of the problem

Currently if the user tries to sign the contract, it will say that "the contract needs to be signed by all parties before joining". This isn't a very helpful message.

## Describe your changes

If the user has not signed the contract yet, I have changed the button to join to a button that redirects to the docuseal contract. 
I do not check yet if the cosigner has signed the contract so if anyone could give advice on that, it'd be greatly helpful!

image
IN PROGRESS
Fixes #11030


app/views/organizer_position_invites/show.html.erb
```diff
@@ -14,10 +14,14 @@
...
-<% end %>
+<% end %> 
 
 <div class="flex justify-center items-center flex-wrap mt3">
   <%= link_to "Reject", organizer_position_invite_reject_path(@invite), method: :post, class: "btn bg-error m1" %>
-  <%= link_to "Accept invitation", organizer_position_invite_accept_path(@invite), method: :post, class: "btn bg-success h2 m1" %>
+  <% if @invite.accept %>
+    <%= link_to "Accept invitation", organizer_position_invite_accept_path(@invite), method: :post, class: "btn bg-success h2 m1" %>
+  <% else %>
+    <%= link_to "Sign Here", @invite.organizer_url, class: "btn bg-accent h2 m1" %>
```
Luke-Oldenburg:
```suggestion
    <%= link_to "Sign here", @invite.organizer_url, class: "btn bg-accent h2 m1" %>
```

app/views/organizer_position_invites/show.html.erb
```diff
@@ -14,10 +14,14 @@
       </div>
     <% end %>
   </div>
-<% end %>
+<% end %>
```
Luke-Oldenburg:
```suggestion
<% end %>
```

app/controllers/organizer_position_invites_controller.rb
```diff
@@ -67,7 +67,7 @@ def accept
     if @invite.accept
       redirect_to @invite.event
     else
-      flash[:error] = @invite.pending_signature? ? "Before accepting the invite, the associated contract needs to be signed by all parties." : "Failed to accept"
+      flash[:error] = @invite.pending_signature? ? "Your invite to #{@invite.event.name} can not be accepted till you have has signed the contract" : "Failed to accept"
```
Luke-Oldenburg:
```suggestion
      flash[:error] = @invite.pending_signature? ? "Your invite to #{@invite.event.name} cannot be accepted until you have has signed the contract" : "Failed to accept"
```

Luke-Oldenburg:
Doesn't this message show in other scenarios than just if the user hasn't signed the contract?

Sarvesh-Mk:
> Doesn't this message show in other scenarios than just if the user hasn't signed the contract?
It shouldn't, what other scenarios would it be shown in? I know it will show if the cosigner has not signed the contract so I am working on that but other than that it should be ok.

PR #11274, opened 2025-08-08
Fix for `ActiveRecord::PreparedStatementInvalid` when `search` contains placeholders

davidcornu:
## Summary of the problem

Internal thread: https://hackclub.slack.com/archives/C047Y01MHJQ/p1754599349421249
Exception: https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2117/samples/timestamp/2025-08-07T20:40:51Z

We are hitting a `ActiveRecord::PreparedStatementInvalid` error when `search` contains something that looks like [a placeholder](https://guides.rubyonrails.org/active_record_querying.html#placeholder-conditions) (e.g. `:pdf`).

This is because 
1. The value of `search` is first interpolated into an SQL fragment using `ActiveRecord::Base.sanitize_sql_array` in `search_modifier_for`: https://github.com/hackclub/hcb/blob/37873c40617cf6b30b3c9c621d38cc639d698bcc/app/services/transaction_grouping_engine/transaction/all.rb#L86-L96
2. This SQL fragment is then interpolated into a larger fragment in `canonical_transactions_grouped_sql`: https://github.com/hackclub/hcb/blob/37873c40617cf6b30b3c9c621d38cc639d698bcc/app/services/transaction_grouping_engine/transaction/all.rb#L231 https://github.com/hackclub/hcb/blob/37873c40617cf6b30b3c9c621d38cc639d698bcc/app/services/transaction_grouping_engine/transaction/all.rb#L257
3. (2), along with several other fragments are combined into a single query which is then passed through `ActiveRecord::Base.sanitize_sql_array` to interpolate in `@event_id`: https://github.com/hackclub/hcb/blob/37873c40617cf6b30b3c9c621d38cc639d698bcc/app/services/transaction_grouping_engine/transaction/all.rb#L319

This means that a `:foo` in `@search` is added in (2) but subsequently treated as a placeholder in (3)

Simplified example:
CleanShot 2025-08-08 at 13 37 26

## Describe your changes

- Adds some really basic test coverage to make sure 
    - the SQL query is at least run
    - the search functionality works
    - escaping is working correctly (or at least enough to solve this issue)
        Test is able to reproduce this failure on the existing code: 
        CleanShot 2025-08-08 at 13 20 05

    
- Replaced second call to `sanitize_sql_array` with interpolated calls to `sanitize_sql_for_conditions` for `@event_id`

### DIFF
--- a/app/services/transaction_grouping_engine/transaction/all.rb
+++ b/app/services/transaction_grouping_engine/transaction/all.rb
@@ -207,7 +207,7 @@ module TransactionGroupingEngine
               from
                 canonical_pending_event_mappings cpem
               where
-                cpem.event_id = :event_id
+                #{ActiveRecord::Base.sanitize_sql_for_conditions(["cpem.event_id = ?", @event_id])}
                 and cpem.subledger_id is null
               except ( -- hide pending transactions that have either settled or been declined.
                 select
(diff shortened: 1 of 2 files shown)


sampoder (approved):
thank you for the tests!

sampoder:
Merging due to ongoing errors

PR #11275, opened 2025-08-08
Fix the amount and date filters on the transactions page

### LINKED ISSUE #11263: [Bug] Filter by amount in ledger broken
### Describe your goals

Filter by amount in ledger

### Describe the bug

When I click filter with amount the styling entirely goes away :(

https://github.com/user-attachments/assets/82a95b64-e0b2-4c55-8b98-52f0ee34b9a6

davidcornu:
Fixes https://github.com/hackclub/hcb/issues/11263

### Context

- The transaction page uses a turbo frame to lazy load the ledger: https://github.com/hackclub/hcb/blob/bf8c2791596a5d502be1b8d12046490445f3c110/app/views/events/transactions.html.erb#L144-L148
- The contents of the turbo frame are rendered by the `.../ledger` endpoint (note that these contain the filters as well): https://github.com/hackclub/hcb/blob/6e1deab06a36ee580aedc5b5abd8b4ed4a64b7b7/app/views/events/ledger.html.erb#L2-L3
- The filters are a mix of links https://github.com/hackclub/hcb/blob/2abe39a5cde4869372dc8d4edbc302019753c6e5/app/views/events/_filter_menu.html.erb#L50-L52 and forms https://github.com/hackclub/hcb/blob/2abe39a5cde4869372dc8d4edbc302019753c6e5/app/views/events/_filter_menu.html.erb#L57-L61
- When the user clicks on the filter icon, these filter options are rendered inside of a menu https://github.com/hackclub/hcb/blob/d84dbbac2d29d75362b7172df841cb4458e763bc/app/views/events/_filter_menu.html.erb#L1

### Where this went wrong

1. Our menu component supports a `data-menu-append-to-value` attribute which determines where the DOM for the menu is rendered. This was configured in this case but pointed to an ID that did not exist (`turbo-frame#ledger`), which meant the menu was being appended outside of the `<turbo-frame>` element.
2. This meant that when one of the filter forms was submitted, turbo treated this as a full page visit to the `.../ledger` endpoint, and when the response didn't include the full page structure it bailed and issued a full page reload of that URL (likely because of the behaviour described in https://turbo.hotwired.dev/handbook/drive#reloading-when-assets-change)

## Describe your changes

- Fixes the `data-menu-append-to-value` attribute.
- Scopes the params passed to the ledger turbo frame to just `request.query_parameters` instead of all of `params` as that led to `action` and `controller` being set which caused havoc with all the `link_to(upsert_query_params...` calls we had
- Uses explicit paths in all `link_to`s

### DIFF
--- a/app/views/events/transactions.html.erb
+++ b/app/views/events/transactions.html.erb
@@ -141,7 +141,7 @@
 
   <%= render "pinned_transactions" %>
 
-  <%= turbo_frame_tag [dom_id(@event), :ledger], src: event_ledger_path(@event, **params.to_unsafe_h) do %>
+  <%= turbo_frame_tag [dom_id(@event), :ledger], src: event_ledger_path(@event, request.query_parameters) do %>
     <%= render partial: "filter" %>
 
     <%= render "application/loading_container", klass: "mt-8" %>
--- a/app/views/events/_filter.html.erb
+++ b/app/views/events/_filter.html.erb
@@ -101,7 +101,7 @@
           <%= avatar_for @user, size: 18 %>
           <%= @user.name %>
 
-          <%= link_to event_ledger_url(@event, upsert_query_params(user: nil)), class: "flex items-center", data: { turbo_prefetch: "false" } do %>
+          <%= link_to event_ledger_path(@event, upsert_query_params(user: nil)), class: "flex items-center", data: { turbo_prefetch: "false" } do %>
             <%= inline_icon "view-close", size: 20 %>
           <% end %>
         </div>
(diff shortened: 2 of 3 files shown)


app/views/events/transactions.html.erb
```diff
@@ -141,7 +141,7 @@
 
   <%= render "pinned_transactions" %>
 
-  <%= turbo_frame_tag [dom_id(@event), :ledger], src: event_ledger_path(@event, **params.to_unsafe_h) do %>
+  <%= turbo_frame_tag [dom_id(@event), :ledger], src: event_ledger_path(@event, request.query_parameters) do %>
```
davidcornu, line 144: This meant we were rendering the ledger endpoint with `action` and `controller` parameters which was really messing with the `link_to` calls in `app/views/events/_filter.html.erb` as we would get the transactions URL on initial load but the ledger URL on subsequent ones.

sampoder (approved):
thank you!

PR #11277, opened 2025-08-08
Handle locked accounts gracefully

davidcornu:
## Summary of the problem

When we try to create a session for a locked user https://github.com/hackclub/hcb/blob/21ec8d158f5d4ba52f4e7d780124dd01a26e611b/app/helpers/sessions_helper.rb#L55 we end up hitting a validation error https://github.com/hackclub/hcb/blob/a4b1cf68ea7847c7e8c6ed7df12c166c1f866f3c/app/models/user_session.rb#L115-L119 which goes unhandled (https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2036)

## Describe your changes

- Checks whether the user is locked in `SessionsHelper#sign_in` and throws `AccountLockedError`
- Adds handling for `AccountLockedError` in `LoginsController#complete`
- Adds tests to make sure
  - We correctly redirect users
  - We still allow admins to impersonate locked users

### DIFF
--- a/app/controllers/logins_controller.rb
+++ b/app/controllers/logins_controller.rb
@@ -171,6 +171,8 @@ class LoginsController < ApplicationController
         redirect_to choose_login_preference_login_path(@login, return_to: @return_to), status: :temporary_redirect
       end
     end
+  rescue SessionsHelper::AccountLockedError => e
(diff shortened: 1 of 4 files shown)


sampoder (approved):
lgtm! thank you!

PR #11278, opened 2025-08-09
[V4 API] Add expense payout transaction type

Luke-Oldenburg:
This adds the expense payout transaction type so we can link to the associated report from the mobile app.

> [!IMPORTANT]  
> Should we add other expense fields to the expense payout view created here or should we wait until there is a thorough reimbursement API and just return those fields from that?


sampoder (approved):
lgtm - no plans for a reimbursements API thus far so feel free to add more info

PR #11279, opened 2025-08-09
[My Reimbursements] Hide reports you can't review

Luke-Oldenburg:
## Summary of the problem

Before it was showing all pending reports for events you were in that did not have a review assigned.

> [!IMPORTANT]
> The reimbursements icon counter doesn't match up to the number of reimbursements on the my/reimbursements page; should it?

## Describe your changes

Now it only checks events you are a manager in becuase you must be a manager to review.

### DIFF
--- a/app/controllers/my_controller.rb
+++ b/app/controllers/my_controller.rb
@@ -129,7 +129,10 @@ class MyController < ApplicationController
 
   def reimbursements
     my_reports = current_user.reimbursement_reports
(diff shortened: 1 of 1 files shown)


sampoder:
> The reimbursements icon counter doesn't match up to the number of reimbursements on the my/reimbursements page; should it?

Can you please update this as well?

sampoder (approved):
lgtm!

PR #11280, opened 2025-08-10
[TUBS] Remove $5 friends banner

Sarvesh-Mk:
## Summary of the problem

I want to close the $5 friends tub by August 10th, so to do that I want to remove the code.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -109,25 +109,6 @@
         </p>
       </div>
     <% end %>
-    <% if Flipper.enabled?(:5_dollar_friends, current_user) %>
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.erb
```diff
@@ -109,23 +109,6 @@
         </p>
       </div>
     <% end %>
-    <% if organizer_signed_in? && @event && (@event.activated_at || @event.created_at) >= "2025-07-03" && !@event.demo_mode? && @event.organized_by_teenagers? && @event.active_teenagers <= 3 && !@hide_three_teens_banner %>
```
Luke-Oldenburg, line 112: `@hide_three_teens_banner` needs to be removed from the controller.

Luke-Oldenburg (changes requested):

PR #11285, opened 2025-08-11
Create an admin helper to swap disbursements events

### LINKED ISSUE #10641: [Admin] Button to swap source and destination on disbursements
### Describe the problem

Sometimes you open the disbursement page from an organization you are wishing to transfer into. This button would swap it from the source to the destination and vice versa.

Image

_No response_

sampoder:
Closes https://github.com/hackclub/hcb/issues/10641

Screenshot 2025-08-10 at 8 10 07 PM


sampoder:
I wish I could have done this client-side but we manage state in the select organisation partial and not the entire form so that made that too complex imo.

Issue #10691, opened 2025-06-24
[Reimbursements] Submit and return modal are swapped for admins


Luke-Oldenburg:
### Describe your goals

I was trying to return a reimbursement report that had been submitted to HCB ops for review.

### Describe the bug

The return button opened the modal to submit the reimbursement. After refreshing this goes away. The same occurs for the submit button, where it opens the return modal. I don't remember the exact steps to get here but the behavior was reproduced by @garyhtou.

PR #11287, opened 2025-08-11
Fix reimbursements buttons that open modals by giving them distinct IDs

sampoder:
## Summary of the problem

I think Turbo's morphing was doing something weird.


sampoder:
context: https://github.com/hackclub/hcb/issues/10691

PR #11290, opened 2025-08-11
[Receipt Bin] Use `Mail#recipients` instead of `.to.first`

davidcornu:
## Summary of the problem

- https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/539/samples/timestamp/2025-08-08T19:06:02Z
- https://hackclub.slack.com/archives/C096P73LEMT/p1754679998405029

`mail.to.first` isn't always present (e.g. when we are BCCed) which can cause errors in production.

## Describe your changes

Use `Mail#recipients` (https://api.rubyonrails.org/classes/Mail/Message.html#method-i-recipients) to handle more scenarios

### DIFF
--- a/app/mailboxes/receipt_bin_mailbox.rb
+++ b/app/mailboxes/receipt_bin_mailbox.rb
@@ -34,11 +34,19 @@ class ReceiptBinMailbox < ApplicationMailbox
 
   private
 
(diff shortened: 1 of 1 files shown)


sampoder (approved):
til! thank you

PR #11292, opened 2025-08-11
[Loops] Clean-up Faraday usage (take 2)

davidcornu:
## Summary of the problem

1. In https://github.com/hackclub/hcb/pull/11237 I centralized `SyncWithLoops`'s `Faraday` instance and added the `:raise_errors` middleware so we'd get exceptions when things went wrong
2. I subsequently noticed this was causing issues in production so undid my changes in #11291 
3. After digging a bit more I realized that 404s were to be expected from this endpoint and that this was previously handled by `return nil if resp.body.strip == "[]"`
    https://loops.so/docs/api-reference/find-contact
    CleanShot 2025-08-11 at 11 39 41@2x

## Describe your changes

Added a `rescue` block for `Faraday::ResourceNotFound` that just returns `nil`

### DIFF
--- a/app/services/user_service/sync_with_loops.rb
+++ b/app/services/user_service/sync_with_loops.rb
@@ -56,30 +56,18 @@ module UserService
 
     def contact_details
       @queue.shift
-      conn = Faraday.new(url: "https://app.loops.so/")
 
-      resp = conn.send(:get) do |req|
-        req.url("api/v1/contacts/find")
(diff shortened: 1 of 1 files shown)


davidcornu:
Approved via Slack: https://hackclub.slack.com/archives/C047Y01MHJQ/p1755638613925669?thread_ts=1755636753.938879&cid=C047Y01MHJQ

PR #11294, opened 2025-08-11
Discard `Receipt::SuggestPairingsJob` on `ActiveJob::DeserializationError`

davidcornu:
## Summary of the problem

We have a number of `Receipt::SuggestPairingsJob` instances that end up retrying in a loop because the underlying record has been deleted: https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/1717

## Describe your changes

Discard these jobs if they hit `ActiveJob::DeserializationError`

### DIFF
--- a/app/jobs/receipt/suggest_pairings_job.rb
+++ b/app/jobs/receipt/suggest_pairings_job.rb
@@ -3,6 +3,9 @@
 class Receipt
   class SuggestPairingsJob < ApplicationJob
     queue_as :low
+
+    discard_on ActiveJob::DeserializationError
+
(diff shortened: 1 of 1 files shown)


davidcornu:
Seems like https://github.com/hackclub/hcb/pull/11345 does a similar thing slightly differently. [Internal discussion](https://hackclub.slack.com/archives/C047Y01MHJQ/p1755531077004089).

sampoder:
Should we close this in favour of https://github.com/hackclub/hcb/pull/11353?

PR #11297, opened 2025-08-11
[Announcement] Ensure nil is not included in the monthly template

polypixeldev:
## Summary of the problem

New monthly announcements generated for non-transparent organizations currently have a bug where nil is included as a ProseMirror block, which caused the ProseMirror renderer to throw an error.

## Describe your changes

Added a filter to make sure that nil is not included as a block in the monthly template

### DIFF
--- a/app/models/announcement/templates/monthly.rb
+++ b/app/models/announcement/templates/monthly.rb
@@ -101,7 +101,7 @@ class Announcement
                 { type: "text", text: "The #{@event.name} team" },
               ],
             },
-          ].filter(&:present?),
+          ].compact,
         }
       end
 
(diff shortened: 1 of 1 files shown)


app/models/announcement/templates/monthly.rb
```diff
@@ -101,7 +101,7 @@ def json_content(donation_summary_block:, donation_goal_block:, top_categories_b
                 { type: "text", text: "The #{@event.name} team" },
               ],
             },
-          ],
+          ].filter(&:present?),
```
sampoder: Can you use `.compact`?

sampoder (approved):

PR #11298, opened 2025-08-11
Introduce transaction categories

davidcornu:
## Summary of the problem

In order to streamline some of the accounting work we have to do on a recurring basis, we want to be able to categorize transactions (automatically when possible, manually otherwise) so we can generate P&L statements on a per-`Event`, per-`Event` tree (e.g. including sub-events), and global level.

## Describe your changes

1. Adds a new `transaction_categories` table with an associated `transaction_category_mappings` table which connects the transaction tables to their assigned categories via a polymorphic association (`categorizable_{type,id}`)
    image
    - The mapping tables also stores a `assignment_strategy` (and potentially other metadata) so we can distinguish manually-assigned categories (which should never change) from those that were the result of mapping logic (which we might want to re-run if we find bugs or ship improvements)
    - For simplicity and consistency, the relations between these models use `category` and `category_mapping` nomenclature.
2. Adds a `TransactionCategory::Definition` class backed by `db/data/transaction_categories.json` which stores the canonical list of categories
    - Entries in the `transaction_categories` table are lazily created based on these definitions
    - This allows us to tie in additional data which doesn't need to be in the database (e.g. accounting codes, groupings)
3. Adds a `TransactionCategoryService` class which handles setting or clearing a category for a given `CanonicalTransaction` or `CanonicalPendingTransaction`
4. Adds an admin-visible `<select>` on the HCB code page (backed by corresponding controller actions for each transaction type) to set the transaction category.
    CleanShot 2025-08-18 at 11 19 50


sampoder (commented):
Design question - why one model for CPTs and one for CT?

davidcornu:
> Design question - why one model for CPTs and one for CT?

I'm asking myself the same question and I may change my mind. My initial reasoning was so I can avoid polymorphic relations and get foreign keys (without having one column for each transaction type and additional validation)

app/controllers/canonical_pending_transactions_controller.rb
```diff
@@ -31,6 +31,29 @@ def update
...
+    authorize @canonical_pending_transaction
+
+    name = params.dig(:canonical_pending_transaction, :category_name)
+
+    TransactionCategoryService
+      .new(model: @canonical_pending_transaction)
+      .set!(name:, assignment_strategy: "manual")
+
+    respond_to do |format|
+      format.turbo_stream {}
```
davidcornu: TODO: Set the flash here

sampoder:
> My initial reasoning was so I can avoid polymorphic relations and get foreign keys (without having one column for each transaction type and additional validation)

I'd be in favour of a polymorphic relationship

davidcornu:
> I'd be in favour of a polymorphic relationship

Can you explain why? It would remove a table and model which reduces some duplication but I'm not sure that's worth giving up foreign keys and an obvious database structure.

davidcornu:
After recent discussions I'm inclined to think storing category slugs might be more prudent. Refactor incoming.

db/data/transaction_categories.json
davidcornu, line 1: we can simplify this now that there's no more revenue categories

sampoder:
> Can you explain why? It would remove a table and model which reduces some duplication but I'm not sure that's worth giving up foreign keys and an obvious database structure.

The first two points and also because as we plan to rewrite the transaction engine, I'd like our models to be as flexible as possible.

> an obvious database structure

Polymorphic relationships are fairly common in our codebase and we've been working over the past few years to introduce more of them (as you've probably seen many of our old models don't use them). I think that we can expect familiarity with that structure.

app/models/canonical_pending_transaction_category_mapping.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: canonical_pending_transaction_category_mappings
+#
+#  id                               :bigint           not null, primary key
+#  assignment_strategy              :text             not null
```
sampoder: Any reason for this not being an integer like our other enums?

app/models/transaction_category.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: transaction_categories
+#
+#  id         :bigint           not null, primary key
+#  slug       :citext           not null
```
sampoder, line 8: What are we planning on using slugs for?

sampoder: I see they're similar to labels - what's the thought behind the name difference?

app/controllers/canonical_pending_transactions_controller.rb
```diff
@@ -31,6 +31,34 @@ def update
...
+    TransactionCategoryService
+      .new(model: @canonical_pending_transaction)
+      .set!(slug:, assignment_strategy: "manual")
+
+    message = "Transaction category was successfully updated."
+
+    respond_to do |format|
+      format.turbo_stream do
+        flash.now[:success] = message
+        render(turbo_stream: turbo_stream.replace("flash-container", partial: "application/flash"))
```
sampoder, line 51: thank you for using turbo streams!

app/models/transaction_category.rb
davidcornu: I need a human-readable way to cross-reference entries in the `transaction_categories` table with the additional metadata in `TransactionCategory::Definition` (loaded from `db/data/transaction_categories.json`). My original plan was to just use the label, however in recent discussions with Gary and Mel I've realized we likely don't want to couple what's being stored in the DB to what's being shown to users as the latter should be able to change without needing to backfill the former (e.g. if we want more detailed labels).

`slug` seemed like the right term for this https://en.wikipedia.org/wiki/Clean_URL#Slug

app/models/canonical_pending_transaction_category_mapping.rb
davidcornu: > Any reason for this not being an integer like our other enums?

The biggest reason is human-readability. If I'm digging through things in Blazer or looking at logs, it's _way_ easier to reason about what's going on (and not get things subtly wrong) rather than having to cross-reference integers from the code base.


&nbsp; | String enums | Integer enums | [Database enums](https://www.postgresql.org/docs/current/datatype-enum.html)
--- | --- | --- | ---
Pros | - Human-readable<br>- Easy to add to | - Space-efficient | - Space-efficient<br>- Human-readable (the DB does the mapping)
Cons | - Larger to store, transmit, and index | - Completely meaningless to humans<br>- Requires cross-referencing the code<br>- Easy to accidentally break | - You can only add new values to the end<br>- Requires DB migrations to change

davidcornu:
> Polymorphic relationships are fairly common in our codebase and we've been working over the past few years to introduce more of them (as you've probably seen many of our old models don't use them). I think that we can expect familiarity with that structure.

My general stance on polymorphic relations is they introduce more problems than they solve (referential integrity being the obvious one, but I've been bitten _many_ times by the strong coupling of the `_type` field to the class structure that backs it which requires tricky multi-step migrations to refactor) so I try to only use them when absolutely necessary.

> The first two points and also because as we plan to rewrite the transaction engine, I'd like our models to be as flexible as possible.

This is entirely opinion-based but my experience is that the opposite is true. I find stricter systems easier to reason about and evolve rather than trying to guess at what flexibility will be needed in the future.

----

Ultimately I'm not going to die on this hill so if we want it to be polymorphic so be it.

davidcornu:
Big rebase incoming to switch to a polymorphic relation

sampoder:
Thank you - appreciate it!

app/models/canonical_pending_transaction_category_mapping.rb
sampoder: Interesting - TIL. Thank you, this sounds good to me.

app/models/transaction_category.rb
sampoder: That feels good to me - thanks for the extra context

davidcornu:
Updated the description as well

sampoder (approved):
thank you!

PR #11299, opened 2025-08-11
[Announcement] Create and destroy monthly announcements when config is changed

polypixeldev:
## Summary of the problem

When a manager either enables or disables generating monthly announcements in event settings, nothing happens right now. When it is enabled, it should create a new monthly announcement, and when disabled, discard the current monthly draft.

Also, if an event had already disabled monthly announcements, the announcement is still in the database, so it currently would get sent out at the end of the month.

## Describe your changes

Added a save hook on `Event::Configuration` to create and destroy monthly announcements, a check in the monthly announcement job, and a warning on the event settings page about disabling generating monthly announcements.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -360,8 +360,12 @@ class EventsController < ApplicationController
           @event.plan.mark_inactive!(plan_param) # deactivate old plan and replace it
         end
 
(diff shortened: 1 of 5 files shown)


sampoder (approved):
lgtm - thank you!

PR #11301, opened 2025-08-11
Check destroy policy in followers modal

polypixeldev:
## Summary of the problem

The remove button currently shows up for all members of an organization even though only managers and the follower themselves can remove the follow.

## Describe your changes

Check the destroy policy to render the remove button in followers modal.


sampoder (approved):
lgtm!

PR #11305, opened 2025-08-11
[Document] Make rails aware that the `category` column will exist later

### LINKED ISSUE #11304: [Document] DB migrations fail from fresh start
DB migrations are failing from a fresh, nonexistent db since we added the new category enum. The enum isn't created initially in the model creation migration but is referenced, so rails detects that there's an error (category column not existing) and stops all following migrations.

fix: add `attribute :category, :integer, default: 0` above the enum to explicitly tell rails that this column will exist (i think)

rluodev:
## Summary of the problem

migrations from a fresh db fail because the category column does not exist when rails expects it to, since there is a category enum in the model.

## Describe your changes

add a line explicitly stating that the category column will exist later.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -49,6 +49,7 @@ class Document < ApplicationRecord
 
   scope :common, -> { where(event_id: nil) }
 
+  attribute :category, :integer, default: 0
   enum :category, {
     general: 0,
     nonprofit_status: 1,


sampoder:
this feels wrong - sorry :/ is there a better fix - what have we done with past enums?

rluodev:
@sampoder ik it feels janky to me too, but the alternative would seem to be adding the enum column in the original migration (which is bad and idt would work)

we've done what i did here before for the `:expense_number` attribute in `expense.rb`

sampoder:
> we've done what i did here before for the :expense_number attribute in expense.rb

expense number isn't an enum?

sampoder:
I'm going to close this out, I think David's comment here is the fix: https://github.com/hackclub/hcb/issues/11304#issuecomment-3179748679

PR #11306, opened 2025-08-12
[DX] Dependency clean-up

davidcornu:
## Summary of the problem

Did a quick pass through our `Gemfile` and did some housekeeping.

## Describe your changes

- I can't find any references to `dry-validation`
- We don't need `brakeman` in the default group (i.e. we only run it in development and CI, it isn't require to run the app)
- `barnes` seems to require a local StatsD server which AFAIK we do not have
- `recursive-open-struct` was used in one place and was trivially replaced by just passing an `OpenStruct` for the necessary property instead of a `Hash`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -128,8 +128,6 @@ gem "validates_zipcode" # validation for event's zip codes
 
 gem "rqrcode" # QR code generation
 
-gem "brakeman" # static security vulnerability scanner
-
 gem "awesome_print" # pretty print objects in console
 gem "byebug", platforms: [:windows]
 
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -128,27 +127,15 @@ gem "validates_zipcode" # validation for event's zip codes
 
 gem "rqrcode" # QR code generation
 
-gem "brakeman" # static security vulnerability scanner
-
 gem "awesome_print" # pretty print objects in console
 gem "byebug", platforms: [:windows]
-gem "dry-validation"
```
davidcornu, line 135: Originally introduced in fcdfe16049b35f3c1ba354822b17c5316bdebd98 and cleaned-up in c25a7a260bfba88caf4f0be498ec328b2de3f686.

sampoder (approved):

sampoder:
thank you!

PR #11289, opened 2025-08-11
Report `deprecation.rails` ActiveSupport::Notifications to AppSignal

albertchae:
## Summary of the problem

We don't have much visibility into Rails deprecations in prod or a test suite

## Describe your changes

HCB's setting for deprecations in production is notify
https://github.com/hackclub/hcb/blob/fd60bc13b0fe081f7e1da8cd4d0fa98ee2f21990/config/environments/production.rb#L115-L116

We can listen for these notifications and report them to AppSignal for
visibility. See https://www.fastruby.io/blog/rails/upgrades/deprecation-warnings-rails-guide.html for more details.

I think we should merge this soon so we can start collecting Rails 8
deprecations, then merge more of https://github.com/hackclub/hcb/pull/10836 piecemeal to see if the deprecations resolve.

To test locally, I changed the configuration to `:notify`, set a real AppSignal credential and ran this in a rails console
```
ActiveRecord::Base.connection_pool.connection
```
This reported the following warning to AppSignal

```
DEPRECATION WARNING: ActiveRecord::ConnectionAdapters::ConnectionPool#connection is deprecated
and will be removed in Rails 8.0. Use #lease_connection instead.
```

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2122

PR #11310, opened 2025-08-13
Remove deprecated usage of passing nil to argument model of form_with

albertchae:
## Summary of the problem

`form_with(model: nil)` is being deprecated in Rails 8

https://github.com/rails/rails/commit/0c150bacaaca7772121300ca124e3ed5392f6ab9

This is the most prominent deprecation warning we are seeing after adding https://github.com/hackclub/hcb/pull/11289

## Describe your changes

@KiKoS0 already had a commit addressing most of these in https://github.com/hackclub/hcb/pull/10836 so I pulled it out into a separate PR and rebased it off main to fix merge conflicts 

It looks like we can either omit the `model:` argument entirely or explicitly set `model: false`

### DIFF
--- a/app/views/admin/ach_start_approval.html.erb
+++ b/app/views/admin/ach_start_approval.html.erb
@@ -27,7 +27,7 @@
 <h3>Actions</h3>
 
 <% if @ach_transfer.pending? %>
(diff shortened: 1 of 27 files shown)


albertchae:
All of these are this warning
image

davidcornu (approved):

PR #11314, opened 2025-08-13
Fix validation error in `UserService::UpdateCardLocking`

davidcornu:
## Summary of the problem

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/1940

We have validation in place that prevents updates to users if they don't have 2fa enabled: https://github.com/hackclub/hcb/blob/8f0a8a89ce254306239f00e900bfed6aa0efffa5/app/models/user.rb#L178-L182 (introduced in https://github.com/hackclub/hcb/pull/9432)

This is causing failures in `UserService::UpdateCardLocking` because we get a validation error when trying to update a field on a user record: https://github.com/hackclub/hcb/blob/8f0a8a89ce254306239f00e900bfed6aa0efffa5/app/services/user_service/update_card_locking.rb#L23

## Describe your changes

- Switches the validation mechanism to only take effect when the `use_two_factor_authentication` field is updated.
- Shows a big angry banner if users need to enable 2FA
    CleanShot 2025-08-13 at 16 18 45@2x

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -175,8 +175,18 @@ class User < ApplicationRecord
 
   validate :profile_picture_format
 
+  def self.with_2fa_requirement_disabled(&)
+    previous_value = Fiber[:user_2fa_requirement_disabled]
+    Fiber[:user_2fa_requirement_disabled] = true
+    yield
+  ensure
+    Fiber[:user_2fa_requirement_disabled] = previous_value
+  end
+
   validate on: :update do
-    if Rails.env.production? && admin_override_pretend? && !use_two_factor_authentication?
+    next if Fiber[:user_2fa_requirement_disabled] || Rails.env.development?
+
+    if admin_override_pretend? && !use_two_factor_authentication?
       errors.add(:access_level, "two factor authentication is required for this access level")
     end
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -175,21 +175,7 @@ class User < ApplicationRecord
 
   validate :profile_picture_format
 
-  def self.with_2fa_requirement_disabled(&)
-    previous_value = Fiber[:user_2fa_requirement_disabled]
-    Fiber[:user_2fa_requirement_disabled] = true
-    yield
-  ensure
-    Fiber[:user_2fa_requirement_disabled] = previous_value
-  end
-
-  validate on: :update do
-    next if Fiber[:user_2fa_requirement_disabled] || Rails.env.development?
-
-    if admin_override_pretend? && !use_two_factor_authentication?
-      errors.add(:access_level, "two factor authentication is required for this access level")
-    end
-  end
+  validate(:admins_cannot_disable_2fa, on: :update)
 
   enum :comment_notifications, { all_threads: 0, my_threads: 1, no_threads: 2 }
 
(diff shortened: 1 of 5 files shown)


app/services/user_service/update_card_locking.rb
```diff
@@ -20,7 +20,9 @@ def run
         TwilioMessageService::Send.new(@user, message).run!
       end
 
-      @user.update!(cards_locked: cards_should_lock)
+      User.with_2fa_requirement_disabled do
+        @user.update!(cards_locked: cards_should_lock)
+      end
```
sampoder: Can you please send the user ID of the user without 2FA enabled to me?

davidcornu: https://hcb.hackclub.com/blazer/queries/942-users-with-elevated-privileges-and-no-2fa

Not sure who this triggered on through, just saw it crop up in the AppSignal errors.

---

In general, I'd recommend we handle enforcing 2FA for admins at the controller level rather than at the model layer as there are many cases (like this one) where the fact that 2FA isn't enabled isn't relevant and there's no way to report back to the user.

sampoder: lol - look who it was.

Screenshot 2025-08-13 at 3 26 25 PM

app/models/user.rb
```diff
@@ -175,8 +175,18 @@ class User < ApplicationRecord
...
+    yield
+  ensure
+    Fiber[:user_2fa_requirement_disabled] = previous_value
+  end
+
   validate on: :update do
-    if Rails.env.production? && admin_override_pretend? && !use_two_factor_authentication?
+    next if Fiber[:user_2fa_requirement_disabled] || Rails.env.development?
+
+    if admin_override_pretend? && !use_two_factor_authentication?
```
sampoder: I'm not super comfortable with this - I don't like selectively running validations. The reason we have this validation on user is so that users can't disable their 2FA requirement.

davidcornu: > I'm not super comfortable with this - I don't like selectively running validations.

Same. This is a duct tape solution that tries to solve the problem with minimal changes.

> The reason we have this validation on user is so that users can't disable their 2FA requirement.

Then I'd propose we do two things:

1. Keep the validation but only apply it if `use_two_factor_authentication` has changed, which will prevent users from disabling it and won't hinder other code paths that don't touch that attribute
2. Add a `before_filter` somewhere to force admins who don't have 2fa enabled to do so before doing anything else

davidcornu: I'll update this PR to do (1) as it would effectively solve the bug, and I'll follow up with (2)

davidcornu: Done

davidcornu: > 2\. Add a `before_filter` somewhere to force admins who don't have 2fa enabled to do so before doing anything else

This proved quite brittle in practice as we'd have to carefully exempt all the endpoints necessary to enable 2fa (including saving and confirming a phone number), so instead I chose to show a prominent banner.

sampoder (approved):
code lgtm - see prev convo: https://hackclub.slack.com/archives/C047Y01MHJQ/p1752671349803129

PR #11321, opened 2025-08-13
Don't do a policy check with `current_user` in receipts form

sampoder:


app/views/receipts/_form_v3.html.erb
```diff
@@ -65,7 +65,6 @@
             <% additional_classes = [].tap do |array|
                  array << "bg-success" if local_assigns[:success]
                  array << "bg-error" if local_assigns[:error]
-                 array << "disabled" if defined?(receiptable) && !ReceiptablePolicy.new(current_user, receiptable).upload?
```
sampoder, line 68: @Luke-Oldenburg I think the long term fix is to make this:

```
array << "disabled" if defined?(receiptable) && current_user && !ReceiptablePolicy.new(current_user, receiptable).upload?
```

rluodev (approved):
LGTM!

PR #11323, opened 2025-08-13
Add disabled to receipts form

Luke-Oldenburg:
## Summary of the problem

We had to remove the disabled because it was breaking the unauthenticated receipts page. This readds it with a check for the presence of `current_user`. See https://github.com/hackclub/hcb/pull/11321#discussion_r2274312989 for more context.


sampoder:
@rluodev can you take point on reviewing this?

sampoder (approved):

PR #11325, opened 2025-08-13
More deprecated usage of `form_with(model: nil...`

davidcornu:
## Summary of the problem

See https://github.com/hackclub/hcb/pull/11310 and https://github.com/hackclub/hcb/pull/11322

There was another one: https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2140

## Describe your changes

Use `model: false`

cc @albertchae

### DIFF
--- a/app/views/events/_pinned_transactions.html.erb
+++ b/app/views/events/_pinned_transactions.html.erb
@@ -12,7 +12,7 @@
       <% end %>
 
       <%= link_to organizer_signed_in? ? pin.hcb_code : "#", data: @show_popovers ? { turbo_frame: "_top", behavior: "modal_trigger", modal: "transaction_details_#{pin.__id__}" } : {}, class: "pinned__transaction relative card shadow-none border flex flex-col black" do %>
-        <%= form_with url: pin_hcb_code_path(pin.hcb_code, event: @event), model: @hcb_code, data: { turbo_method: :post, turbo_confirm: "This will unpin the transaction for all team members. Continue?" } do |form| %>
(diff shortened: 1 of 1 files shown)


davidcornu:
Tested this locally and works just fine without `@hcb_code`.

app/views/events/_pinned_transactions.html.erb
```diff
@@ -12,7 +12,7 @@
       <% end %>
 
       <%= link_to organizer_signed_in? ? pin.hcb_code : "#", data: @show_popovers ? { turbo_frame: "_top", behavior: "modal_trigger", modal: "transaction_details_#{pin.__id__}" } : {}, class: "pinned__transaction relative card shadow-none border flex flex-col black" do %>
-        <%= form_with url: pin_hcb_code_path(pin.hcb_code, event: @event), model: @hcb_code, data: { turbo_method: :post, turbo_confirm: "This will unpin the transaction for all team members. Continue?" } do |form| %>
+        <%= form_with url: pin_hcb_code_path(pin.hcb_code, event: @event), model: false, data: { turbo_method: :post, turbo_confirm: "This will unpin the transaction for all team members. Continue?" } do |form| %>
```
sampoder: why wouldn't we use `pin.hcb_code`?

davidcornu: Given that we're passing in `url` I'm not sure what difference it would make but I can test it out

davidcornu: Updated to use a `form_for`

albertchae (approved):

sampoder:
Merging to declutter AppSignal

PR #11326, opened 2025-08-14
Remove `domain` from error message in DeleteOrgUnit

sampoder:
@rluodev is there a better way to do this?

### DIFF
--- a/app/jobs/g_suite/delete_org_unit_job.rb
+++ b/app/jobs/g_suite/delete_org_unit_job.rb
@@ -7,7 +7,7 @@ class GSuite
     include Partners::Google::GSuite::Shared::DirectoryClient
 
     retry_on Google::Apis::ClientError, attempts: 3, wait: 2.minutes do |job, e|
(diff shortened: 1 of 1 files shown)


app/jobs/g_suite/delete_org_unit_job.rb
```diff
@@ -7,7 +7,7 @@ class DeleteOrgUnitJob < ApplicationJob
     include Partners::Google::GSuite::Shared::DirectoryClient
 
     retry_on Google::Apis::ClientError, attempts: 3, wait: 2.minutes do |job, e|
-      Rails.error.report("Failed to delete GSuite org unit for domain #{domain} after 3 attempts (pls do manually): #{e.message}\nBacktrace: #{e.backtrace}")
+      Rails.error.report("Failed to delete GSuite org unit for after 3 attempts (please do manually): #{e.message}\nBacktrace: #{e.backtrace}")
```
rluodev:
```suggestion
      Rails.error.report("Failed to delete GSuite org unit after 3 attempts (please do manually): #{e.message}\nBacktrace: #{e.backtrace}")
```

rluodev:
I think this might be the way to go with this tbh

PR #11327, opened 2025-08-14
[Receipts] Extract currency from textual content

### LINKED ISSUE #8207: [Receipt] Add `extracted_total_amount_currency` field

YodaLightsabr:
## Summary of the problem
HCB doesn't store the currency on uploaded receipts, and we render all extracted amounts as if they're USD.

## Describe your changes
Start extracting the currency on new receipt uploads.

Closes #8207

### DIFF
--- a/app/models/receipt.rb
+++ b/app/models/receipt.rb
@@ -39,6 +39,9 @@ class Receipt < ApplicationRecord
   blind_index :textual_content
   has_encrypted :extracted_card_last4
 
+  monetize :extracted_subtotal_amount_cents, as: "extracted_subtotal_amount", with_model_currency: :extracted_currency, allow_nil: true
+  monetize :extracted_total_amount_cents, as: "extracted_total_amount", with_model_currency: :extracted_currency, allow_nil: true
+
   include StripeAuthorizationsHelper
 
   include PublicIdentifiable
--- a/app/services/receipt_service/extract.rb
+++ b/app/services/receipt_service/extract.rb
@@ -25,6 +25,7 @@ module ReceiptService
 
         subtotal_amount_cents
         total_amount_cents // the amount likely to be charged to a credit card
+        currency // 3-digit ISO 4217 currency code
         card_last_four
         date // in the format of YYYY-MM-DD
         merchant_url // URL for merchant's primary website including https, if available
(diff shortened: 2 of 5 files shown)


app/services/receipt_service/extract.rb
```diff
@@ -77,6 +78,7 @@ def run!
         suggested_memo: data.transaction_memo,
         extracted_subtotal_amount_cents: data.subtotal_amount_cents&.to_i,
         extracted_total_amount_cents: data.total_amount_cents&.to_i,
+        extracted_currency: data.currency&.upcase,
```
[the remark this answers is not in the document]
YodaLightsabr: For now, we want to store exactly what the LLM outputs. If it isn't a valid currency, `Money` handles it gracefully and defaults to USD.

app/models/receipt.rb
```diff
@@ -39,6 +39,9 @@ class Receipt < ApplicationRecord
   blind_index :textual_content
   has_encrypted :extracted_card_last4
 
+  monetize :extracted_subtotal_amount_cents, as: "extracted_subtotal_amount", with_model_currency: :extracted_currency, allow_nil: true
+  monetize :extracted_total_amount_cents, as: "extracted_total_amount", with_model_currency: :extracted_currency, allow_nil: true
+
```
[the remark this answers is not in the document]
YodaLightsabr: This is already handled by the gem

db/migrate/20250814191925_add_extracted_currency_to_receipt.rb
```diff
@@ -0,0 +1,5 @@
+class AddExtractedCurrencyToReceipt < ActiveRecord::Migration[7.2]
+  def change
+    add_column :receipts, :extracted_currency, :string
```
davidcornu, line 3: Do we want a default value + `null: false` for this or are we making use of the fact that this could be null?

davidcornu: Ah ok I see https://github.com/hackclub/hcb/pull/11327/files#r2277629963

davidcornu (approved):

db/migrate/20250814191925_add_extracted_currency_to_receipt.rb
YodaLightsabr: Yeah, long-term I think we can backfill it and then add a default value + `null: false` to be more explicit with the default behavior, but for the time being this will help us identify what cases cause the LLM to _not_ output a valid currency.

PR #11328, opened 2025-08-14
Fix styling on card grant detail tooltips

polypixeldev:
## Summary of the problem

Some card grant details tooltips were links to `#` when they should have been `span`s, and one `span` tooltip wasn't centered properly.

## Describe your changes

Changed the `link_to "#"` tooltips to be spans that are properly centered with the text.

image


sampoder (approved):
thank you!

PR #11332, opened 2025-08-15
[Receipts] Calculate suggestions using local currency

YodaLightsabr:
## Summary of the problem
Currently receipt suggestions are calculated by the USD amount instead of the local merchant amount which causes issues for transactions in a different currency.

## Describe your changes
This PR calculates suggestions using the merchant's currency.


sampoder (approved):
thank you!

PR #11347, opened 2025-08-18
[Help Scout] Support Secure Mode for Beacons

YodaLightsabr:
## Summary of the problem
Emails can be tampered with

## Describe your changes
Stop tampering


sampoder (approved):
lgtm!

PR #11349, opened 2025-08-18
Run `User::UpdateCardLockingJob` when marking receipt as lost

sampoder:
May be causing issues.


app/models/concerns/receiptable.rb
```diff
@@ -34,6 +34,9 @@ def no_or_lost_receipt?
     def no_or_lost_receipt!
       self.marked_no_or_lost_receipt_at = Time.now
       self.save!
+      if user = try(:author) || try(:user) || try(:event)
```
garyhtou, line 37: I haven't fully looked through this code, but it surprises me that we're passing the value of `try(:event)` (which I assume is an Event), as the kwarg `user:`.

PR #11351, opened 2025-08-18
Override `deliver_mail` not `deliver_now`

sampoder:


sampoder:
third times a charm they say

PR #11355, opened 2025-08-18
Add missing annotation for `receipts.extracted_currency`

davidcornu:
## Summary of the problem

https://github.com/hackclub/hcb/pull/11327 added a new column but not the annotations

From https://hcb.hackclub.com/schema:

CleanShot 2025-08-18 at 14 01 45@2x

## Describe your changes

Commits the annotation


YodaLightsabr:
ahhh sorry about this. thanks for fixing

PR #11367, opened 2025-08-19
Handle malformed dates in the ReceiptService::Suggest

sampoder:


YodaLightsabr (approved):
This looks good. Are there any specific cases of this happening in App Signal? I'm wondering how common this is / if we need to update the prompt.

PR #11368, opened 2025-08-19
Redirect to my cards if nil event during card creation

polypixeldev:
## Summary of the problem

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/660
```
No route matches {action: "new", controller: "stripe_cards", event_id: nil}, possible unmatched constraints: [:event_id]
```

## Describe your changes

Confusing - could not reproduce this bug or find a reason why `event_id` would be nil when this form is submitted, but I added a check in case `event` is nil to just redirect to `/my/cards` instead of the event-specific cards page.


sampoder (approved):
lgtm! thank you

PR #11370, opened 2025-08-19
[CI] Pin PG to 15 and add Redis

davidcornu:
## Summary of the problem

- We're currently using the `postgres:latest` image which doesn't match what we're running in production
- We don't have Redis running on CI which prevents us from running some Sidekiq jobs

## Describe your changes

- Use the `15` tag for the `postgres` container
- Add a `redis` service

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:14
+        image: postgres:15
         env:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,7 +170,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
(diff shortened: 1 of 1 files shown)


polypixeldev (commented):
Looks like we have Postgres 15.12 in production -  `SELECT version();` returns
```
PostgreSQL 15.12 (Ubuntu 15.12-1.pgdg24.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0, 64-bit
```

davidcornu:
@polypixeldev oh! I should have known that but for some reason I was convinced it was 14 😅. Thanks!

.github/workflows/ci.yml
```diff
@@ -94,7 +94,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:latest
+        image: postgres:14
```
sampoder: Should this be 15?

davidcornu: Oh lol. I just realized we have multiple PG images in this file. Good catch

PR #11372, opened 2025-08-19
[Ledger] Fix ledger sorting showing unstyled page

polypixeldev:
## Summary of the problem

The ledger sorting buttons were showing an unstyled page since they were not in a turbo frame, causing the entire page content to be replaced by the ledger.

## Describe your changes

Puts the ledger sort menu in the ledger turbo frame so that it captures the links.

### DIFF
--- a/app/views/events/_filter.html.erb
+++ b/app/views/events/_filter.html.erb
@@ -30,7 +30,7 @@
   <% end %>
 
   <% admin_tool "overflow-hidden" do %>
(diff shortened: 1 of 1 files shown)


davidcornu:
Same issue as https://github.com/hackclub/hcb/pull/11275. Fun

davidcornu (approved):

polypixeldev:
Yep, thanks for fixing the filters on #11275! Made finding the fix for this really quick

PR #11373, opened 2025-08-19
[Reimbursement::Expense] Remove reimbursement expense emails

polypixeldev:
## Summary of the problem

Context: https://hackclub.slack.com/archives/C026RKHLPNJ/p1755264851531389
Emails for each reimbursement expense were a bit much, and not really necessary since an email is sent when the report is approved or a comment is created.

## Describe your changes

Removed the `expense_approved` and `expense_unapproved` emails and references to them.


sampoder (changes requested):
Sorry - I know I said we should drop this but I actually think we should refactor them to be batched and only send if the entire report isn't approved. So have a job that looks at all expenses approved in the past 20 minutes without an approved report and sends an email to those users

app/jobs/reimbursement/expenses_approved_job.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Reimbursement
+  class ExpensesApprovedJob < ApplicationJob
```
sampoder:
```suggestion
  class Expense
            class ApprovedNotificationJob < ApplicationJob
```

app/jobs/reimbursement/expenses_approved_job.rb
```diff
@@ -0,0 +1,32 @@
...
+    queue_as :default
+
+    def perform
+      approvals = PaperTrail::Version.where_object_changes_to(aasm_state: "approved")
+                                     .select { |version| version.item_type == "Reimbursement::Expense" && version.created_at > 20.minutes.ago }
```
sampoder: What about mileage reimbursements?

app/jobs/reimbursement/expenses_approved_job.rb
```diff
@@ -0,0 +1,32 @@
...
+        report = expense.report
+
+        next unless expense.approved? && report.submitted?
+
+        unless approved_expenses_by_report.key?(report)
+          approved_expenses_by_report[report] = []
+        end
+
+        approved_expenses_by_report[report].append(expense)
+      end
```
sampoder: Could we do a filter and then group by?

app/jobs/reimbursement/expenses_approved_job.rb
polypixeldev: Tested this in dev - the `item_type` is still `Reimbursement::Expense` even if it is a mileage expense, I assume since `has_paper_trail` is called in `Reimbursement::Expense` and `Reimbursement::Expense::Mileage` inherits from that.

app/jobs/reimbursement/expenses_approved_job.rb
polypixeldev: good idea! group by is awesome

polypixeldev:
Also added `.uniq` in case an expense is approved/unapprove/approved again so that it doesn't get listed twice

sampoder (approved):

PR #11374, opened 2025-08-19
[Reimbursement::Expense] Remove expense approval on update

polypixeldev:
## Summary of the problem

Currently, someone can issue a reimbursement report with an expense for $2, get it approved by a manager, but if the entire report is not yet approved they can still convert it to a draft and modify the $2 expense to have different details, such as for $4 instead of $2. When submitted back for review, it looks like the expense was approved, even though it could be entirely different.

## Describe your changes

Added a callback to `Reimbursement::Expense` that will mark the expense as pending if is updated with new details (memo/amount/category/description), so that it must be reviewed again before the report is paid out.

### DIFF
--- a/app/models/reimbursement/expense.rb
+++ b/app/models/reimbursement/expense.rb
@@ -117,6 +117,13 @@ module Reimbursement
       end
     end
 
+    before_update do
+      if approved? && (memo_changed? || amount_cents_changed? || category_changed? || description_changed?)
+        mark_pending!
+        self.approved_by = nil
+      end
+    end
+
     def receipt_required?
       true
     end


app/models/reimbursement/expense.rb
```diff
@@ -117,6 +117,13 @@ class Expense < ApplicationRecord
       end
     end
 
+    before_update do
+      if approved? && (memo_changed? || amount_cents_changed? || category_changed? || description_changed?)
```
sampoder, line 121: imo we should only do this if the amount changes

YodaLightsabr: I don't have a super strong opinion on this, but maybe we should also check if the receipts changed?

polypixeldev: I'm fine with making it just the amount, my thought was that any change being made and keeping it approved could be misleading, but amount is definitely the most important. I considered checking receipts as well, but I didn't see a good way of doing it since the expense modal isn't actually updated when a receipt is uploaded, it's just a polymorphic relationship on receipt to any receiptable.

sampoder: Ok - let's run with this and iterate on it production

garyhtou:
lol good find!

PR #11385, opened 2025-08-20
[Admin] Show previous POCs in event settings

### LINKED ISSUE #11384: [Admin] Include former POCs on event settings page
Right now it can be tricky to see who former POCs might have been. We should add a list of former POCs to HCB right below the point-of-contact field.

YodaLightsabr:
## Summary of the problem
It can be difficult to see previous POCs for an organization

## Describe your changes
Fixes #11384 by showing previous POCs below the POC field


app/models/event.rb
```diff
@@ -851,6 +851,10 @@ def organizer_contact_emails
     emails
   end
 
+  def point_of_contact_history
+    @point_of_contact_history ||= versions.map(&:changeset).select{|a| a["point_of_contact_id"].present?}.map { |c| User.find_by(id: c["point_of_contact_id"][0]) }
```
davidcornu: ~**concern**: Are we sure these user IDs will always exist?~

Nevermind - confused my finder methods. Do we want to `.compact` at the end to account for those?

YodaLightsabr: That's a good idea

davidcornu: **suggestion**: I think we can clean this up a little

```suggestion
    @point_of_contact_history ||= 
      versions
        .filter_map { |v| v.changeset["point_of_contact_id"].presence }
        .filter_map { |(old_id, _new_id)| User.find_by(id: old_id) }
```

sampoder (approved):
lgtm!

app/views/events/settings/_admin.html.erb
```diff
@@ -86,6 +86,9 @@
           Point of Contact <%= link_to "(Me!)", event_claim_point_of_contact_path(@event), method: :post %>
         <% } %>
         <%= form.collection_select :point_of_contact_id, User.admin.or(User.where(id: @event.point_of_contact_id)).order(:email), :id, :email, {}, { disabled: } %>
+        <% if event.point_of_contact_history.any? %>
+          (Previously <%= event.point_of_contact_history.map(&:email).join(", ") %>)
```
sampoder: I think you can use to_sentence here

davidcornu (approved):

PR #11371, opened 2025-08-19
[Categories] Import MCC mappings

davidcornu:
## Summary of the problem

Follow-up to #11298 

## Describe your changes

- Imports stripe merchant categories mappings from https://docs.google.com/spreadsheets/d/1I7TSqE6OxplCvS2YbJCBS5gNgY0bHIFQy4gVBWyqpWY/edit?gid=341089138#gid=341089138
- Adds test to make sure `db/data/transaction_categories.json` is valid

PR #11386, opened 2025-08-20
[Categories] Set automatically based on stripe merchant codes

davidcornu:
## Summary of the problem

In #11298 and #11371 we added a mechanism to associate transactions to categories. We would like to do that automatically for Stripe transactions.

## Describe your changes

1. Adds a `merchant_category` method to both the raw stripe transaction models
2. Adds `TransactionCategoryService#sync_from_stripe!` to handle the syncing logic
3. Runs (2) when `CanonicalPendingTransaction`s are created for `RawPendingStripeTransaction`s

I _think_ this is sufficient given the updated logic in `CanonicalPendingTransactionService::Settle` but I'm not sure if that's the only way Stripe transactions are created.

### DIFF
--- a/app/models/raw_pending_stripe_transaction.rb
+++ b/app/models/raw_pending_stripe_transaction.rb
@@ -38,7 +38,7 @@ class RawPendingStripeTransaction < ApplicationRecord
     stripe_transaction.dig("merchant_data", "name")
   end
 
-  def merchant_category_key
+  def merchant_category
     stripe_transaction&.dig("merchant_data", "category")
   end
 
--- a/app/models/raw_pending_stripe_transaction.rb
+++ b/app/models/raw_pending_stripe_transaction.rb
@@ -38,6 +38,10 @@ class RawPendingStripeTransaction < ApplicationRecord
     stripe_transaction.dig("merchant_data", "name")
   end
 
+  def merchant_category_key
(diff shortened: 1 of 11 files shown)


sampoder (approved):
code lgtm! only thing I'd say is I think it should be `merchant_category` or `merchant_category_code` and not `merchant_category_key`. when I was reading the code I was expecting `merchant_category_key` to refer to some sort of key / value mapping we had.

davidcornu:
> only thing I'd say is I think it should be `merchant_category` or `merchant_category_code` and not `merchant_category_key`. when I was reading the code I was expecting `merchant_category_key` to refer to some sort of key / value mapping we had.

My reasoning for `_key` is:
- We should match the nomenclature in `YellowPages`: https://github.com/hackclub/yellow_pages/blob/4bd51a8d596a0a65e4f2a7d202ff8224032ac117/lib/yellow_pages/category.rb#L19-L27
- `_code` could be confused with the MCC code which is an integer (https://docs.stripe.com/issuing/categories)

We could use `merchant_category`

davidcornu:
> I _think_ this is sufficient given the updated logic in `CanonicalPendingTransactionService::Settle` but I'm not sure if that's the only way Stripe transactions are created.

Investigation here: https://hackclub.slack.com/archives/C047Y01MHJQ/p1755716056250769

PR #11390, opened 2025-08-21
Fix nil event errors in ct/cpt partials

polypixeldev:
## Summary of the problem

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/1687/samples/timestamp/2025-08-20T15:06:42Z

Sometimes `@event` is not set and the canonical (pending) transaction being rendered is not yet mapped to an event - this needs to be handled in the canonical (pending) transaction partials.

## Describe your changes

Refactored the default event in both partials to a variable and does not render the create tag menu if there is no event set.


sampoder (approved):
thank you! I appreciate that you've also taken the time to clean this up (the `@event`)

sampoder:
Tested loading HQ locally

PR #11391, opened 2025-08-21
[Ledger] Allow auditors to set sorting

polypixeldev:
## Summary of the problem

Currently, changing sorting on the ledger doesn't work for auditors even though the admin tool displays since the controller checks for admin instead of auditor.

## Describe your changes

Change the events controller to allow auditors and above to set the sorting method


sampoder (changes requested):
good catch!

PR #11398, opened 2025-08-21
Use a toast for admin flash

davidcornu:
## Summary of the problem

The current admin flash is a bit rudimentary and isn't ideal when combined with turbo streams as it shifts the layout and doesn't go away.

This would normally be fine but I'm in the process of introducing several new interactions that would benefit.

## Describe your changes

- Moves the flash logic to a partial (so it can be refreshed from a turbo stream)
- Updates the DOM structure to rely on classes instead of inline styles
- Adds a slide down animation
- Adds a button to dismiss
- Adds a stimulus controller which handles the button clicks and will auto-dismiss after 5 seconds

https://github.com/user-attachments/assets/84019150-4389-4242-aa09-8f916d76215b

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -909,6 +909,74 @@ form:has(> #admin-logout-button) {
   margin: 0px;
 }
 
+@keyframes flash-slide-in {
+  from {
+    translate: 0 calc(-100% - var(--flash-offset, 0));
+  }
+
+  to {
+    translate: 0 0;
+  }
+}
+
+#flash-container {
+  position: absolute;
+  top: 0;
+  left: calc(50vw - 25ch);
+  right: calc(50vw - 25ch);
+  display: flex;
+  justify-content: center;
+
+  article.flash {
+    --flash-offset: 1rem;
+    border: 1px solid var(--hcb-bg-2);
+    border-top: 3px solid var(--hcb-link-1);
+    border-radius: 4px;
+    background: white;
+    padding: 1rem;
+    margin: var(--flash-offset) 0 0;
+    overflow: hidden;
+    width: 100%;
+    display: flex;
+    align-items: center;
+    gap: 1rem;
+    box-shadow: 0 1px 2px hsl(240 3.8% 46.1% / 12%);
+    animation-duration: 0.5s;
+    animation-name: flash-slide-in;
+    z-index: 999;
+
+    &.flash--success {
+      border-top-color: green;
+    }
+
+    &.flash--error {
+      border-top-color: red;
+    }
+
+    &.flash--dismissed {
+      translate: 0 calc(-100% - var(--flash-offset));
+      display: none;
+      transition-property: translate display;
+      transition-duration: 1s;
+      transition-behavior: allow-discrete;
+    }
+
+    .flash--content {
+      flex: 1;
+    }
+  }
+
+  .flash--close-button {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    background: transparent;
+    padding: 0;
+    margin: 0;
+    color: black;
+  }
+}
+
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -929,10 +929,10 @@ form:has(> #admin-logout-button) {
 
   article.flash {
     --flash-offset: 1rem;
-    border: 1px solid var(--hcb-bg-2);
+    border: 1px solid var(--hcb-bg-3);
     border-top: 3px solid var(--hcb-link-1);
     border-radius: 4px;
-    background: white;
+    background: var(--hcb-bg-1);
     padding: 1rem;
     margin: var(--flash-offset) 0 0;
     overflow: hidden;
@@ -973,7 +973,7 @@ form:has(> #admin-logout-button) {
     background: transparent;
     padding: 0;
     margin: 0;
-    color: black;
+    color: inherit;
   }
 }
 
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -930,7 +930,7 @@ form:has(> #admin-logout-button) {
   article.flash {
     --flash-offset: 1rem;
     border: 1px solid var(--hcb-bg-3);
-    border-top: 3px solid var(--hcb-link-1);
+    border-top: 3px solid var(--hcb-accent);
     border-radius: 4px;
     background: var(--hcb-bg-1);
     padding: 1rem;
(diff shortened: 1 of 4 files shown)


sampoder:
Will review when I get home! Could you attach a screen recording and screenshot?

davidcornu:
> Could you attach a screen recording and screenshot?

Yep was just recording one.

app/assets/stylesheets/admin.scss
```diff
@@ -909,6 +909,74 @@ form:has(> #admin-logout-button) {
...
+    &.flash--error {
+      border-top-color: red;
+    }
+
+    &.flash--dismissed {
+      translate: 0 calc(-100% - var(--flash-offset));
+      display: none;
+      transition-property: translate display;
+      transition-duration: 1s;
+      transition-behavior: allow-discrete;
```
davidcornu, line 961: **note**: This is relatively new functionality (https://caniuse.com/mdn-css_properties_display_is_transitionable) that lets us make the element `display: none` once the transition completes, which is necessary to allow users to click what was under the flash element. 

The alternative would be to add a JS listener but IMHO the fallback behaviour of having the element just disappear instantly is fine for the browsers that don't support that yet (mainly Firefox which I'm one of the few remaining users of 😅)

app/assets/stylesheets/admin.scss
```diff
@@ -909,6 +909,74 @@ form:has(> #admin-logout-button) {
   margin: 0px;
 }
 
+@keyframes flash-slide-in {
+  from {
+    translate: 0 calc(-100% - var(--flash-offset, 0));
```
davidcornu: **note**: This is to account for the element's top margin (using a custom property here to make sure it's consistent)

app/assets/stylesheets/admin.scss
```diff
@@ -909,6 +909,74 @@ form:has(> #admin-logout-button) {
...
+  to {
+    translate: 0 0;
+  }
+}
+
+#flash-container {
+  position: absolute;
+  top: 0;
+  left: calc(50vw - 25ch);
+  right: calc(50vw - 25ch);
```
davidcornu: **note**: This lets us have a absolutely-positioned element that is `50ch` wide at the centre of the screen. We could use other methods to centre the flash but they have the downside of covering more of the screen.

app/assets/stylesheets/admin.scss
```diff
@@ -909,6 +909,74 @@ form:has(> #admin-logout-button) {
...
+
+  .flash--close-button {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    background: transparent;
+    padding: 0;
+    margin: 0;
+    color: black;
+  }
```
davidcornu: **note**: This is required because
1. the close button should be a `<button>` in keeping with semantic HTML principles
2. we apply default styling to `<button>` which makes this use case (putting an icon in it) a bit rough

app/assets/stylesheets/admin.scss
```diff
@@ -909,6 +909,74 @@ form:has(> #admin-logout-button) {
...
+    display: flex;
+    align-items: center;
+    gap: 1rem;
+    box-shadow: 0 1px 2px hsl(240 3.8% 46.1% / 12%);
+    animation-duration: 0.5s;
+    animation-name: flash-slide-in;
+    z-index: 999;
+
+    &.flash--success {
+      border-top-color: green;
```
sampoder: I think we should be using our design system's green and red. It'd be nice to also support a `flash--info` with our blue colour.

app/assets/stylesheets/admin.scss
sampoder: 82.01% global usage seems good to me - thanks for the writeup!

app/assets/stylesheets/admin.scss
sampoder: Agreed. We should be using the default text colour in light mode and not the HTML `black` colour (we have a slight. `--hcb-tx-2` I think is the right CSS variable.

app/views/admin/_flash.html.erb
sampoder, line 1: this lgtm! just a note that this only the toast on admin pages, there's also a toast on normal pages - do we want to roll out this switch for that as well?

app/assets/stylesheets/admin.scss
sampoder: You can ignore the `flash-info` comment I can see that that is the default

sampoder:
Thoughts on putting the toast in the bottom right and having it rise up? We generally place a lot of content in the centre of the page (thinking about a future where we use this for non-admin pages).

app/assets/stylesheets/admin.scss
davidcornu: Good point. I'll update.

app/assets/stylesheets/admin.scss
davidcornu: > I think we should be using our design system's green and red.

Ah I just ported over the existing logic. I can switch to use the design system colours.

app/views/admin/_flash.html.erb
davidcornu: > just a note that this only the toast on admin pages

Yep that was intentional. The flash on the regular pages seemed fine.

> there's also a toast on normal pages - do we want to roll out this switch for that as well?

We could. I'd rather do that in two steps though once we're happy with the way this one is working.

davidcornu:
> Thoughts on putting the toast in the bottom right and having it rise up? We generally place a lot of content in the centre of the page (thinking about a future where we use this for non-admin pages).

We can definitely do that. My thinking was that we'd want to draw attention to it but it's not uncommon to have these pop up from the bottom (e.g. https://sonner.emilkowal.ski/)

app/assets/stylesheets/admin.scss
davidcornu: Made some additional tweaks in https://github.com/hackclub/hcb/pull/11398/commits/423b966f806abae74e0c49c1c7fb2fe8c79044eb to make dark mode look better.

Light | Dark
---- | ----
CleanShot 2025-08-22 at 10 59 49 | CleanShot 2025-08-22 at 10 59 55

app/assets/stylesheets/admin.scss
davidcornu: Updated in https://github.com/hackclub/hcb/pull/11398/commits/6db4d759038dc1b2e0b293cac583956115cac614

&nbsp; | Light | Dark
---- | ---- | ----
Success | CleanShot 2025-08-22 at 11 07 08 | CleanShot 2025-08-22 at 11 07 17
Error | CleanShot 2025-08-22 at 11 04 58 | CleanShot 2025-08-22 at 11 07 28
Info (default) | CleanShot 2025-08-22 at 11 07 48 | CleanShot 2025-08-22 at 11 07 41

davidcornu:
Moved to the bottom right in https://github.com/hackclub/hcb/pull/11398/commits/b74c4f7ce72d16000ac26dc664bbfe88062f9bcd

https://github.com/user-attachments/assets/90a83bde-632a-425b-839e-a6efc3d97c1d

sampoder (approved):

PR #11399, opened 2025-08-21
[Categories] Allow category to be set from the pending ledger table and transaction page

davidcornu:
## Summary of the problem

Adding more places for admins to set transaction categories.

## Describe your changes

- Removed some default margins in the `<form>` and `<select>` in `app/views/hcb_codes/_transaction_category_form.html.erb` so they take up less space
- Reworked the two `set_category` endpoints to leverage the same flash logic and use a `context` param to conditionally render the admin flash
- Rendered the transaction category form on the pending ledger table
- Flipped the transaction details table so there's more room to add additional properties
- Added a new row in the transaction details table for the category

Transaction | Pending ledger
---- | ----
image | image

⚠️ This builds upon the work in #11398 which makes the admin flash easier to target from turbo streams and better for these kinds of use cases.

### DIFF
--- a/app/controllers/canonical_transactions_controller.rb
+++ b/app/controllers/canonical_transactions_controller.rb
@@ -3,6 +3,7 @@
 require "csv"
 
 class CanonicalTransactionsController < ApplicationController
+  include TurboStreamFlash
   def show
     @canonical_transaction = CanonicalTransaction.find(params[:id])
 
@@ -49,7 +50,7 @@ class CanonicalTransactionsController < ApplicationController
     respond_to do |format|
       format.turbo_stream do
         flash.now[:success] = message
-        render(turbo_stream: turbo_stream.replace("flash-container", partial: "application/flash"))
+        update_flash_via_turbo_stream(use_admin_layout: params[:context] == "admin")
       end
       format.html do
         redirect_to(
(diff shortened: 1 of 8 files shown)


app/controllers/canonical_transactions_controller.rb
```diff
@@ -3,6 +3,7 @@
 require "csv"
 
 class CanonicalTransactionsController < ApplicationController
+  include TurboStreamFlash
```
sampoder, line 6: Thoughts on including this in `ApplicationController`?

sampoder (approved):
code lgtm! thank you

sampoder:
I feel like the toast should be on the bottom right and rise up - I'm worried that we generally put a lot of information in the centre of the page but if we put it on the bottom right we'd be covering less.

app/controllers/canonical_transactions_controller.rb
davidcornu: > Thoughts on including this in `ApplicationController`?

That was my original plan but then I realized `ApplicationController` is also used for the API where this functionality is useless 🤷. For now I think this is alright and we can reconsider if we see broader usage.

davidcornu:
Rebase incoming

PR #11400, opened 2025-08-21
[Categories] Add one-time jobs to backfill transaction categories

davidcornu:
## Summary of the problem

In https://github.com/hackclub/hcb/pull/11386 we started automatically assigning transaction categories based on stripe merchant codes, but that only applies to newly-created records.

## Describe your changes

Adds two iterable jobs that loop through the `canonical_pending_transactions` and `canonical_transactions` tables and run `TransactionCategoryService#sync_from_stripe!` on each record.


sampoder (approved):
lgtm! thank you

PR #11401, opened 2025-08-21
[Announcement] Exclude events with disabled monthly announcements in scope

polypixeldev:
## Summary of the problem

#11299 starte deleting monthly announcements when they were disabled in event configuration and added a check to `monthly_job`, but there were still places where the `monthly` scope was being used without checking if the event had disabled monthly announcements, and there are some monthly announcements in the database for events that have disabled it.

## Describe your changes

Moves the check for disabled monthly announcements to the `monthly` scope, ensuring no monthly announcement is listed for an event that has disabled them. Also removes the now-redundant checks for the configuration.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -418,9 +418,7 @@ class EventsController < ApplicationController
     end
     @announcements = @all_announcements.page(params[:page]).per(10)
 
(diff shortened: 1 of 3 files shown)


sampoder (approved):
lgtm!

PR #11405, opened 2025-08-22
Decline Stripe card authorizations for specific merchant categories

davidcornu:
## Summary of the problem

There are a subset of merchant categories that we'd rather block upfront than deal with later.

## Describe your changes

1. Adds `StripeAuthorizationService::FORBIDDEN_MERCHANT_CATEGORIES`
2. Adds `AdminMailer#blocked_authorizations`
    CleanShot 2025-08-22 at 15 40 47@2x
4. Adds logic to `...::HandleIssuingAuthorizationRequest` which checks against (1), sends (2), and declines.

### DIFF
--- a/app/services/stripe_authorization_service.rb
+++ b/app/services/stripe_authorization_service.rb
@@ -5,7 +5,6 @@ module StripeAuthorizationService
     Set.new(
       [
         "betting_casino_gambling",
-        "cigar_stores_and_stands",
         # This looks like a typo but matches Stripe's documentation
         # https://docs.stripe.com/issuing/categories
         "government_licensed_online_casions_online_gambling_us_region_only",
--- a/app/services/stripe_authorization_service.rb
+++ b/app/services/stripe_authorization_service.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module StripeAuthorizationService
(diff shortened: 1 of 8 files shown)


app/services/stripe_authorization_service.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module StripeAuthorizationService
+  FORBIDDEN_MERCHANT_CATEGORIES =
+    Set.new(
+      [
+        "betting_casino_gambling",
+        "cigar_stores_and_stands",
```
sampoder: Are we sure we want to block this? Cigar stores and tobacconists often sell electronics such as chargers or soft drinks, which we should allow.

davidcornu: Fair point! Got this list from Mel. Let me double-check

davidcornu: Did some internal digging and looks like we've seen legitimate transactions for `cigar_stores_and_stands` so will remove

davidcornu: https://github.com/hackclub/hcb/pull/11405/commits/10133016ddd7037962c25c53ca9746aade82b7c9

sampoder (approved):
looks good to me! thank you for this.

PR #11407, opened 2025-08-22
Refactor tooltips

manuthecoder:
This PR gives our tooltip components a long-overdue refresh. They’ve been rewritten in JavaScript with a cleaner, more maintainable approach, while keeping the code changes as minimal as possible. The result is tooltips that are easier to work with going forward!

image


sampoder (approved):
code lgtm! thank you.

PR #11413, opened 2025-08-23
[Admin] Add new users to referral programs page

Luke-Oldenburg:
## Summary of the problem

Before, we needed to use Blazer to access this number which painted an incorrect image of a referral's effectiveness when viewing the admin page.

## Describe your changes

- Add new users to referral programs page
- Add referral programs to command prompt

### DIFF
--- a/app/javascript/components/command_bar/actions.js
+++ b/app/javascript/components/command_bar/actions.js
@@ -508,6 +508,14 @@ export const adminActions = adminUrls => [
     icon: <Icon glyph="flag-fill" size={16} />,
     perform: navigate('/flipper/features'),
   },
+  {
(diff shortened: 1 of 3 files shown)


app/javascript/components/command_bar/actions.js
```diff
@@ -508,6 +508,14 @@ export const adminActions = adminUrls => [
     icon: <Icon glyph="flag-fill" size={16} />,
     perform: navigate('/flipper/features'),
   },
+  {
+    id: 'admin_tool_30',
```
sampoder:
```suggestion
    id: 'admin_tool_31',
```

sampoder (approved):
looks good to me!

PR #11417, opened 2025-08-24
[DX] Convert phone number link_to to span

Luke-Oldenburg:

### DIFF
--- a/app/views/stripe_cards/_details.html.erb
+++ b/app/views/stripe_cards/_details.html.erb
@@ -139,9 +139,9 @@
       <div class="fs-mask">
         <strong class="inline-flex items-center" style="gap: 4px">
           Phone number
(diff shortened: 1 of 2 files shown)


app/views/stripe_cards/_details.html.erb
```diff
@@ -139,9 +139,9 @@
       <div class="fs-mask">
         <strong class="inline-flex items-center" style="gap: 4px">
           Phone number
-          <%= link_to "#", class: "info tooltipped tooltipped--e", "aria-label": "Seem familiar? This is your number! Use this if asked during checkout :)", style: "display: flex;", data: { turbo: false } do %>
+          <span class="info tooltipped tooltipped--e h-[18px]" aria-label="Seem familiar? This is your number! Use this if asked during checkout :)">
```
sampoder:
```suggestion
          <span class="info tooltipped tooltipped--e h-[18px]" aria-label="This is your phone number - use this if asked during checkout.">
```
I think we should edit this copy a bit

Luke-Oldenburg: haha

sampoder (approved):

PR #11424, opened 2025-08-25
Small announcement tweaks

Luke-Oldenburg:
- change conditional rendering of remove follower button to a disabled
- show feed page always with messages if there are no announcements or followed events
- always show rss feed

### DIFF
--- a/app/views/events/announcement_overview.html.erb
+++ b/app/views/events/announcement_overview.html.erb
@@ -7,10 +7,8 @@
 <h1 class="flex heading items-center">
   <span class="flex-grow flex items-center flex-1">
     Announcements
-    <% if @event_follow.present? || organizer_signed_in?(as: :reader) %>
-      <%= link_to event_feed_path(@event, format: :atom), class: "muted tooltipped tooltipped--e ml-2", "aria-label": "you found an Atom feed!", data: { turbo: false } do %>
-        <%= inline_icon "rss", size: 30 %>
-      <% end %>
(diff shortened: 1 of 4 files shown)


app/views/events/announcement_overview.html.erb
```diff
@@ -7,10 +7,8 @@
 <h1 class="flex heading items-center">
   <span class="flex-grow flex items-center flex-1">
     Announcements
-    <% if @event_follow.present? || organizer_signed_in?(as: :reader) %>
-      <%= link_to event_feed_path(@event, format: :atom), class: "muted tooltipped tooltipped--e ml-2", "aria-label": "you found an Atom feed!", data: { turbo: false } do %>
-        <%= inline_icon "rss", size: 30 %>
-      <% end %>
+    <%= link_to event_feed_path(@event, format: :atom), class: "muted tooltipped tooltipped--e ml-2", "aria-label": "you found an Atom feed!", data: { turbo: false } do %>
```
davidcornu, line 10: Random question - do we include an RSS `<meta>` tag on these pages so they can be auto-discovered by feed readers?

davidcornu (approved):

app/views/events/announcement_overview.html.erb
polypixeldev: Yep! https://github.com/hackclub/hcb/blob/d46b6c61725515e7d30bad94483a0c39efc1a2d5/app/helpers/events_helper.rb#L123

PR #11427, opened 2025-08-25
Fix renaming transactions lag

### LINKED ISSUE #11205: freezing when editing the name of a transaction
### Describe your goals

trying to change the name of a transaction on the page thats like split screened with the transactions page (screenshot below)

### Describe the bug

is this a me problem? it might be, but it's been happening consistently. it freezes and doesnt let me edit when i click like in this screenshot, but i can type changes, and it updates when i hit enter after a sec. this only occurs on the split screen version. this isnt a huge deal but its been driving me nuts

Image

manuthecoder:
Currently, when a user finishes renaming a transaction, the UI appears laggy. This PR adds a loading indicator to give show visual feedback to the user 

https://github.com/user-attachments/assets/cec753e4-cf3e-44cb-814a-d4a1c64e53cf

Resolves #11205

### DIFF
--- a/app/views/hcb_codes/edit.html.erb
+++ b/app/views/hcb_codes/edit.html.erb
@@ -8,10 +8,13 @@
 
 <% if @frame %>
   <%= turbo_frame_tag @hcb_code do %>
-    <%= form_with model: @hcb_code, data: { turbo: true, controller: "form" }, html: { style: "flex-grow: 1" } do |form| %>
+    <%= form_with model: @hcb_code,
+                    data: { turbo: true, controller: "form" },
+                    html: { style: "flex-grow: 1", onsubmit: "$('#checkIcon').hide(); $('#loadingIcon').removeClass('hidden');" } do |form| %>
       <div class="flex items-center">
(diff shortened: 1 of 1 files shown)


app/views/hcb_codes/edit.html.erb
```diff
@@ -8,10 +8,13 @@
 
 <% if @frame %>
   <%= turbo_frame_tag @hcb_code do %>
-    <%= form_with model: @hcb_code, data: { turbo: true, controller: "form" }, html: { style: "flex-grow: 1" } do |form| %>
+    <%= form_with model: @hcb_code,
+                  data: { turbo: true, controller: "form" },
+                  html: { style: "flex-grow: 1", onsubmit: "$('#checkIcon').hide(); $('#loadingIcon').removeClass('hidden');" } do |form| %>
```
davidcornu, line 13: - **suggestion**: I think you can use `grow` from tailwind here (https://tailwindcss.com/docs/flex-grow) instead of inline styles
- **concern**: I'm not sure what our policy is but I generally wouldn't recommend adding inline scripts to new code. Do you mind moving this to a stimulus controller using actions and targets?
- **concern**: Could we standardize on using the `hidden` class rather than jQuery's `.hide()` which applies inline styles?

PR #11430, opened 2025-08-25
Revert "Decline Stripe card authorizations for specific merchant categories"

davidcornu:
Reverts hackclub/hcb#11405


davidcornu:
This was because of https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2226

PR #11431, opened 2025-08-25
Restore "Decline Stripe card authorizations for specific merchant categories"

davidcornu:
## Summary of the problem

1. #11405 added some logic to the Stripe Issuing webhook endpoint to ban gambling merchants
2. This inadvertently caused an issue (https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2226) as in test scenarios, `auth` is a `Hash` whereas in production `auth` is a `Stripe::Issuing::Authorization` object which doesn't have a `dig` method
3. #11430 reverted (1) to avoid causing more issues while the issue was fixed

## Describe your changes

1. Restores #11405
2. Updates the `:stripe_authorization` factory to construct a `Stripe::Issuing::Authorization` object
3. Updates usage of (2) to use `build` instead of `attributes_for`
4. Updates the `merchant_category` method to avoid `dig`

### DIFF
--- a/app/mailers/admin_mailer.rb
+++ b/app/mailers/admin_mailer.rb
@@ -89,4 +89,15 @@ class AdminMailer < ApplicationMailer
     )
   end
 
+  def blocked_authorization
+    @stripe_card = params.fetch(:stripe_card)
+    @event = @stripe_card.event
+    @merchant_category = params.fetch(:merchant_category)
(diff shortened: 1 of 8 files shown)


sampoder (approved):
Thank you for the quick turn around here

PR #11434, opened 2025-08-25
Add transaction category to Event's transaction CSV export

garyhtou:
## Summary of the problem

We want to make these categories easily exportable by users.

## Describe your changes

I'm adding it to the existing Event transaction CSV export, located here in the UI:

image

Here's an example of what the new CSV columns look like:
image

### DIFF
--- a/app/models/transaction_category.rb
+++ b/app/models/transaction_category.rb
@@ -38,7 +38,7 @@ class TransactionCategory < ApplicationRecord
   delegate :label, to: :definition
 
   def definition
-    TransactionCategory::Definition::ALL[slug]
+    TransactionCategory::Definition::ALL.fetch(slug)
   end
 
 end
--- a/app/models/transaction_category.rb
+++ b/app/models/transaction_category.rb
@@ -35,4 +35,10 @@ class TransactionCategory < ApplicationRecord
     inclusion: { in: TransactionCategory::Definition::ALL.keys }
   )
 
+  delegate :label, to: :definition
+
+  def definition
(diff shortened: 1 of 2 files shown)


app/models/transaction_category.rb
```diff
@@ -35,4 +35,10 @@ class TransactionCategory < ApplicationRecord
     inclusion: { in: TransactionCategory::Definition::ALL.keys }
   )
 
+  delegate :label, to: :definition
+
+  def definition
+    TransactionCategory::Definition::ALL[slug]
+  end
```
garyhtou: If, for some reason, the slug can't be looked up and `definition` is `nil`, then we will get a NoMethodError when calling `label`.

I'm not sure how likely it is for `definition` to return `nil`. If it's possible, then we should explicitly handle that.

Any thoughts @davidcornu?

davidcornu: > I'm not sure how likely it is for `definition` to return `nil`.

This should be impossible and if it does occur we should blow up as loudly as possible.

```suggestion
  delegate :label, to: :definition

  def definition
    TransactionCategory::Definition::ALL.fetch(slug)
  end
```

We should not be able to create a `TransactionCategory` that does not have a `Definition`: https://github.com/hackclub/hcb/blob/57b53eef68c660318f98865e726371a42ff6b4ed/app/models/transaction_category.rb#L31-L36

garyhtou: Okay, perfect, thank you!

davidcornu (approved):

PR #11437, opened 2025-08-26
[Event] Add Statement of Activity as an alpha feature

garyhtou:
This adds a very bare bones Statement of Activity using the new transaction categories we just added.

@MelSmith104 needs this for a demo of the new accounting/category feature for a meeting later today, so this new endpoint is simple and **admin only** for the time being. When we actually launch this feature, it'll be more feature-full and look better.

Anyways, here's what it looks like right now (in this PR).

image

You can change the date range via the form inputs, which autosubmit.

image

This page is accessible at http://localhost:4000/assemble/statement_of_activity (again, only for admins).

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -917,6 +917,20 @@ class EventsController < ApplicationController
     @end_date = Date.today.prev_month.beginning_of_month.to_date
   end
 
+  def statement_of_activity
+    authorize @event
+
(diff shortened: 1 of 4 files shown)


garyhtou:
arghhh 
image

garyhtou:
Checks passed on the [latest commit](https://github.com/hackclub/hcb/pull/11437/commits/11f5e9a01d357102d050dd9490752f299b78094c), so I force merged it.

PR #11442, opened 2025-08-26
Fix HCB code type capitalization in comments

YodaLightsabr:
## Summary of the problem
Transaction types are all lowercase even when they shouldn't be

## Describe your changes
Respect naming conventions that require uppercase letters (proper nouns, acronyms, etc.)


leowilkin:
ty ian

PR #11443, opened 2025-08-26
Refactor the admin nav logic so it isn't a big `Hash`

davidcornu:
## Summary of the problem

There's a whole bunch of logic embedded in `app/views/layouts/admin.html.erb` which handles constructing the admin nav (including making database queries for the various counters and figuring out which item is active).

Rather than add yet another item and move along I figured it'd be worth giving this part of the code some love before making my changes (https://x.com/KentBeck/status/250733358307500032).

## Describe your changes

1. Introduces `Admin::Nav` along with `Section` and `Item` classes
2. Moves the active item and section tracking logic into (1)
3. Moves the `task_sum` and `counter_sum` logic into (1)
4. Removes the notion of `flags`, replacing it with `count_type` (either `:tasks` or `:records`)
5. Adds basic tests

> [!TIP]
> Reviewing this refactor commit by commit should make it a lot easier to understand

### DIFF
--- a/app/models/admin/nav.rb
+++ b/app/models/admin/nav.rb
@@ -110,14 +110,54 @@ module Admin
       Section.new(
         name: "Misc",
         items: [
-          make_item(name: "Bank Accounts", path: bank_accounts_admin_index_path, count: BankAccount.failing.count, count_type: :records),
-          make_item(name: "HCB Fees", path: bank_fees_admin_index_path, count: BankFee.in_transit_or_pending.count, count_type: :records),
-          make_item(name: "Column Statements", path: admin_column_statements_path, count: Column::Statement.count, count_type: :records),
-          make_item(name: "Users", path: users_admin_index_path, count: User.count, count_type: :records),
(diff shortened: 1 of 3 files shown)


sampoder:
image

app/models/admin/nav.rb
```diff
@@ -0,0 +1,329 @@
...
+          make_item(
+            name: "Invoices",
+            path: invoices_admin_index_path,
+            count: 0
+          ),
+          make_item(
+            name: "Sponsors",
+            path: sponsors_admin_index_path,
+            count: 0
+          )
```
sampoder, line 239: Can we add counts to these as well?

sampoder (approved):
thank you, this looks good and I appreciate us cleaning this up.

app/models/admin/nav.rb
davidcornu: We can but I'm going to keep the scope of this PR to just the refactor.

PR #11444, opened 2025-08-26
[Wise] Make sure `wise_id` and `wise_recipient_id` are populated when marked as sent

davidcornu:
## Summary of the problem

At the moment
- We mark the transfer as sent before updating it based on the params so if a validation failure occurs, users no longer see the fields
- We don't validate that these fields are populated

## Describe your changes

- Adds presence validation on the fields that only takes effect when the transfer is sent
- Coalesce the operations in `WiseTransfersController#mark_sent` so that 
    1. The action is idempotent (as only one DB call is performed)
    2. We don't proceed if the record is invalid

### DIFF
--- a/app/controllers/wise_transfers_controller.rb
+++ b/app/controllers/wise_transfers_controller.rb
@@ -66,13 +66,16 @@ class WiseTransfersController < ApplicationController
   def mark_sent
     authorize @wise_transfer
 
-    @wise_transfer.mark_sent!
+    @wise_transfer.assign_attributes(wise_transfer_params)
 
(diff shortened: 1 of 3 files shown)


garyhtou (approved):
Note from discussion with David:

- Change validation to be for sent & deposited.
- Backfill the 29 Wise transfers in production: https://hcb.hackclub.com/blazer/queries/956-wise-transfers-sent-deposited-but-blank-wise_id-or-wise_recipent_id

davidcornu:
> Backfill the 29 Wise transfers in production

Done

PR #11447, opened 2025-08-27
Robotics 3D Printer Raffle

Luke-Oldenburg:
Screenshot From 2025-08-27 16-06-46
Screenshot From 2025-08-28 15-34-22

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -30,7 +30,7 @@ class ApplicationController < ActionController::Base
   end
 
   before_action do
-    @hide_robotics_banner = cookies[:hide_robotics_banner] == "1"
+    @hide_promotional_banner = cookies[:hide_promotional_banner] == "1"
   end
 
   before_action do
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -29,7 +29,6 @@ class ApplicationController < ActionController::Base
     cookies.permanent[:first_visit] = 1
   end
 
-  # This cookie is used for Safari PWA prompts
   before_action do
     @hide_robotics_banner = cookies[:hide_robotics_banner] == "1"
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -30,7 +30,7 @@ class ApplicationController < ActionController::Base
   end
 
   before_action do
-    @hide_promotional_banner = cookies[:hide_promotional_banner] == "1"
+    @hide_promotional_banner = cookies[:hide_robotics_raffle_banner] == "1"
   end
 
   before_action do
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -29,6 +29,11 @@ class ApplicationController < ActionController::Base
     cookies.permanent[:first_visit] = 1
   end
 
+  # This cookie is used for Safari PWA prompts
+  before_action do
+    @hide_robotics_banner = cookies[:hide_robotics_banner] == "1"
+  end
+
   before_action do
     # Disallow indexing
     response.set_header("X-Robots-Tag", "noindex")
--- a/app/controllers/my_controller.rb
+++ b/app/controllers/my_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class MyController < ApplicationController
-  skip_after_action :verify_authorized, only: [:activities, :toggle_admin_activities, :cards, :missing_receipts_list, :missing_receipts_icon, :inbox, :reimbursements, :reimbursements_icon, :tasks, :payroll, :feed, :toggle_robotics_banner] # do not force pundit
+  skip_after_action :verify_authorized, only: [:activities, :toggle_admin_activities, :cards, :missing_receipts_list, :missing_receipts_icon, :inbox, :reimbursements, :reimbursements_icon, :tasks, :payroll, :feed, :hide_promotional_banner] # do not force pundit
 
   def activities
     @before = params[:before] || Time.now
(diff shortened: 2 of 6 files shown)


anishanne:
@Luke-Oldenburg Lets go September 30th, will confirm w/ Mel

anishanne:
Is the banner dismissible? And can we add a shoutout to the fee waiver in effect

app/views/layouts/application.html.erb
```diff
@@ -119,6 +119,17 @@
       </div>
     <% end %>
 
+    <% if !@event.demo_mode? && @event.robotics_team? %>
```
davidcornu: `@event` isn't always present here as this is a layout (hence the CI failures)

Luke-Oldenburg:
> Is the banner dismissible? And can we add a shoutout to the fee waiver in effect

It's dismissible now but I don't want to make it any busier than it already is with more text. They should be checking the promotions page.

app/controllers/application_controller.rb
```diff
@@ -29,6 +29,11 @@ class ApplicationController < ActionController::Base
     cookies.permanent[:first_visit] = 1
   end
 
+  # This cookie is used for Safari PWA prompts
```
garyhtou: I think this comment might have been accidentally copied. We can remove it

```suggestion
```

app/views/events/promotions.html.erb
```diff
@@ -61,6 +61,40 @@
...
+          <div class="tooltipped tooltipped--w inline-block" aria-label="Must be tagged as a teen-led organization and robotics team">
+            <span class="btn disabled bg-clip-padding">Ineligible</span>
+          </div>
+        <% end %>
+      </div>
+
+      <p>
+        Is your organization a high-school robotics team? Get an entry into a raffle for a Bambu Lab A1 from now until September 30th for each active teenager in your organization!
+        <% if !(@event.robotics_team? && @event.organized_by_teenagers?) %>
+          Are you ineligible when you should be entered?
```
garyhtou: As an organizer, I think I might be slightly confused by this. How about:

```suggestion
          Are you a high-school robotics team, but it says you're ineligible?
```

app/controllers/application_controller.rb
```diff
@@ -29,6 +29,11 @@ class ApplicationController < ActionController::Base
...
+  before_action do
+    @hide_robotics_banner = cookies[:hide_robotics_banner] == "1"
+  end
```
garyhtou: This is okay for now, as long as we rip out when the promo ends. Ideally, we have a more abstracted system for hiding any banner. This way we don't need:
- Put very specific banner hiding code in ApplicationController (which is intended for very general, shared code)
- Add a new route for setting the cookie.

app/controllers/my_controller.rb
```diff
@@ -17,6 +17,11 @@ def toggle_admin_activities
     redirect_to my_activities_url
   end
 
+  def toggle_robotics_banner
+    cookies.permanent[:hide_robotics_banner] = 1
+    redirect_back_or_to root_path
+  end
+
```
garyhtou: FYI, the controller action name says "toggle", but the code logic does not allow for unhiding the banner. It only hides it.

I think you're okay with this, but just wanted to double check since the name "toggle" implies unhiding.

sampoder:
Can we make it dismissible with an icon? If this is becoming a recurring design I think we need something cleaner.

app/controllers/application_controller.rb
sampoder: +1 - I think we should do this now because we've started using this code repeatedly

app/views/events/promotions.html.erb
```diff
@@ -61,6 +61,40 @@
...
+
+      <p>
+        Is your organization a high-school robotics team? Get an entry into a raffle for a Bambu Lab A1 from now until September 30th for each active teenager in your organization!
+        <% if !(@event.robotics_team? && @event.organized_by_teenagers?) %>
+          Are you a high-school robotics team, but it says you're ineligible?
+          <%= link_to "Click here",
+            fillout_form("thC4xqWfTBus", { "hcbprojname" => @event.name, "hcburl" => event_url(@event), "name" => current_user.name, "email" => current_user.email }),
+            target: :_blank %>
+          to fix it.
+        <% end %>
```
sampoder: This feels weird? What causes someone to be ineligible

app/controllers/my_controller.rb
Luke-Oldenburg: I'll rename to hide.

app/views/events/promotions.html.erb
Luke-Oldenburg: If they're not tagged with robotics and organized by teens.

Luke-Oldenburg:
> Can we make it dismissible with an icon? If this is becoming a recurring design I think we need something cleaner.

Swapped it for an icon but I couldn't get the text to center. Not sure if that's an issue.

app/controllers/application_controller.rb
Luke-Oldenburg: Is it good now just renaming or is there something else we should update?

app/views/events/promotions.html.erb
sampoder: Maybe instead of Click here to fix it say Let us know so we can update our records - just something more human

app/controllers/my_controller.rb
```diff
@@ -17,6 +17,11 @@ def toggle_admin_activities
     redirect_to my_activities_url
   end
 
+  def hide_promotional_banner
+    cookies.permanent[:hide_promotional_banner] = 1
```
sampoder: How does this work when we have a new banner?

Luke-Oldenburg: We keep reusing this path for the new banner. I just realized the cookie will keep it closed though. Maybe the path name should be the same but the cookie changes.

sampoder (approved):
Approving to unblock - before the next banner is added can we please design a more maintainable solution?

PR #11449, opened 2025-08-27
Make `/admin` work again

davidcornu:
## Summary of the problem

Before https://github.com/hackclub/hcb/pull/11448 was merged, `/admin` worked (almost by accident) because `SetEvent#set_event` has special handling if an admin is logged in: https://github.com/hackclub/hcb/blob/6c41a4a917eb6515409b713431fb2946769a6c24/app/controllers/concerns/set_event.rb#L27-L29

This stopped working because we declared an explicit `admin` path to render the admin tools turbo frame for the home page.

## Describe your changes

- Rename `admin` to `admin_tools`
- Add an explicit handler for `/admin`


davidcornu:
fyi @Luke-Oldenburg

PR #11450, opened 2025-08-27
Fix admin tools navigation

davidcornu:
## Summary of the problem

In https://github.com/hackclub/hcb/pull/11448 (updating in https://github.com/hackclub/hcb/pull/11449) we moved the admin tools section of the home page into a lazily-loaded turbo frame. However we forgot to set `target="_top"` so all navigation stays within the frame (see https://turbo.hotwired.dev/handbook/frames#targeting-navigation-into-or-out-of-a-frame).

## Describe your changes

- Set the `target` attribute to promote all navigation to the top-level context
- Update the frame tag name to match the controller action and path


davidcornu:
fyi @Luke-Oldenburg

PR #11459, opened 2025-08-27
Use `persisted?` for all image display in event settings

sampoder:


app/views/events/settings/_details.html.erb
```diff
@@ -46,7 +46,7 @@
                 <%= form.label :logo, "Choose file", class: "field--fileupload__label" %>
                 <%= form.file_field :logo, accept: "image/png,image/jpeg", class: "field--fileupload__field", disabled: %>
               </div>
-              <% if @event.logo.attached? %>
+              <% if @event.logo. %>
```
sampoder:
```suggestion
              <% if @event.logo.attached? %>
```

PR #11226, opened 2025-08-05
Improve error handling in `Api::V4::CardGrantsController#create`

davidcornu:
Internal thread: https://hackclub.slack.com/archives/C047Y01MHJQ/p1754407379830219

## Summary of the problem

The API endpoint to create card grants has many of the same issues that were addressed in https://github.com/hackclub/hcb/pull/11211 for the non-API version (e.g. https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/1651/samples/last).

## Describe your changes

- Use `DisbursementService::Create::UserError` instead of a plain `ArgumentError` for exceptions that require specific handling
- Return an error response for validation and  `DisbursementService::Create` errors
- Return a `201 Created` response with a `Location` header
- Add tests

----

Internal thread making sure these changes are safe to roll out: https://hackclub.slack.com/archives/GBXBFEED9/p1754417640438699

PR #11460, opened 2025-08-27
Rescue on ArgumentErrors in the CardGrantsController

sampoder:


app/controllers/card_grants_controller.rb
```diff
@@ -152,6 +152,8 @@ def activate
     redirect_to @card_grant
   rescue Stripe::InvalidRequestError => e
     redirect_to @card_grant, flash: { error: "This card could not be activated: #{e.message}" }
+  rescue ArgumentError => e
+    redirect_to @card_grant, flash: { error: e.message }
```
davidcornu, line 156: **suggestion**: `ArgumentError` is generally something that should bubble up to reflect a _programmer_ error rather than a user one (which is why I added a `...::UserError` type in https://github.com/hackclub/hcb/pull/11226). I wouldn't assume `e.message` is safe to show end-users.

PR #11462, opened 2025-08-27
[Ledger] Add category filter

polypixeldev:
## Summary of the problem

It'd be great for users to be able to filter transactions by the new transaction categories!

## Describe your changes

Added the filter logic to both `TransactionGroupingEngine` and `PendingTransactionEngine` and implemented the filter on the frontend.

image


app/services/pending_transaction_engine/pending_transaction/all.rb
```diff
@@ -94,6 +95,12 @@ def canonical_pending_transactions
               cpts = cpts.where("canonical_pending_transactions.date <= cast(? as date)", @end_date)
             end
 
+            if @category
+              cpts = cpts.joins("LEFT JOIN transaction_category_mappings tcm on canonical_pending_transactions.id = tcm.categorizable_id AND tcm.categorizable_type = 'CanonicalPendingTransaction'")
+                         .joins("LEFT JOIN transaction_categories tc on tc.id = tcm.transaction_category_id")
+                         .where("tc.id = ?", @category.id)
+            end
```
davidcornu, line 101: **suggestion**: If we already have the category ID here we can skip a join

```suggestion
            if @category
              cpts = cpts.joins("LEFT JOIN transaction_category_mappings tcm on canonical_pending_transactions.id = tcm.categorizable_id AND tcm.categorizable_type = 'CanonicalPendingTransaction'")
                         .where("tcm.transaction_category_id = ?", @category.id)
            end
```

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -119,6 +126,14 @@ def stripe_joins_for(type)
         "left join raw_stripe_transactions on raw_stripe_transactions.id = raw_pending_stripe_transaction_id"
       end
 
+      def category_joins_for(type)
+        return "" unless @category.present?
+
+        ct_type = type == :ct ? "CanonicalTransaction" : "CanonicalPendingTransaction"
+
+        return "left join transaction_category_mappings tcm on #{type}.id = tcm.categorizable_id AND tcm.categorizable_type = '#{ct_type}' left join transaction_categories on transaction_categories.id = tcm.transaction_category_id"
```
davidcornu: **suggestion**: We can apply the same optimization here as https://github.com/hackclub/hcb/pull/11462/files#r2356735175

app/views/events/_filter_menu.html.erb
```diff
@@ -114,6 +115,13 @@
             </div>
           <% end %>
         </div>
+        <div data-tabs-target="tab" id="category">
+          <% TransactionCategory.find_each do |category| %>
```
davidcornu: **suggestion**: You can look up the definitions directly like in https://github.com/hackclub/hcb/blob/1841dea800d0f36bb491c569e657a659fe5500ee/app/views/hcb_codes/_transaction_category_form.html.erb#L17 rather than pinging the database if you like.

polypixeldev: I think it's better to select from the database here since `EventsController` expects the slug to be of a category in the database, so there could be an instance where there would be a category in the definition but not yet in the database, and trying to filter for that category would error.

davidcornu (approved):

davidcornu:
If I understand this correctly this would expose categories to event organizers (which I don't believe they have access to yet, maybe aside from CSV exports). Let's make sure this is ok before we ship this.

&rarr; https://hackclub.slack.com/archives/C026RKHLPNJ/p1758658385734359

garyhtou (approved):
Amazing! We'll merge it. although, we'll keep in mind that the data could be improved and there's currently no way for users to get categories (only admins can)— this is why there is a beta badge on it.

PR #11463, opened 2025-08-28
[v4] v4 scope for card grants

Mohamad-Mortada:
## Summary of the problem
Had a discussion with Gary regarding using this case for automation purposes.  

## Describe your changes

- I added a v4 scope that'll allow u to restrict scope to just this action.
- Added parameter to allow admins to adjust who sent the grant.


app/controllers/api/v4/card_grants_controller.rb
```diff
@@ -16,7 +16,19 @@ def index
...
+
+          if found_user.nil?
+            skip_authorization
+            return render json: { error: "invalid_user", messages: "User with email '#{params[:sent_by_email]}' not found" }, status: :bad_request
+          end
+
+          @current_user = found_user
+        end
+
+        @card_grant = @event.card_grants.build(params.permit(:amount_cents, :email, :merchant_lock, :category_lock, :keyword_lock, :purpose, :one_time_use, :pre_authorization_required, :instructions).merge(sent_by: @current_user))
```
garyhtou: I think it's a bit dangerous to muck with the `@current_user` variable. I understand we want to say that card grant was issued by `found_user`, however, changing `@current_user` would imply that the current user of this entire request is `found_user`, which isn't entirely true.

The reason is because the `current_user` method uses `@current_user` under the hood to store who the current user is.

```suggestion
        sent_by = current_user

        if current_user.admin? && params[:sent_by_email].present?
          found_user = User.find_by(email: params[:sent_by_email].strip.downcase)

          if found_user.nil?
            skip_authorization
            return render json: { error: "invalid_user", messages: "User with email '#{params[:sent_by_email]}' not found" }, status: :bad_request
          end

          sent_by = found_user
        end

        @card_grant = @event.card_grants.build(params.permit(:amount_cents, :email, :merchant_lock, :category_lock, :keyword_lock, :purpose, :one_time_use, :pre_authorization_required, :instructions).merge(sent_by:))
```

garyhtou (commented):
the general gist of this looks right to me. ideally, we get a third pair of eyes (maybe @24c02 if they're available?)

As a reminder, the goal is to create an API that is locked down with the permissions to _**ONLY**_ issue card grants. We reason why we wanna be this strict is because the owner of this API token will be an admin.

Oh yeah, and if they are an admin, we want to allow them to set the `sent_by` field.

app/controllers/api/v4/card_grants_controller.rb
Mohamad-Mortada: Ah I see, yea I agree.

garyhtou (approved):

PR #11465, opened 2025-08-28
Improve accuracy on user nav reimbursements icon

Luke-Oldenburg:
## Summary of the problem

The number was inaccurate before due to changes made on the reimbursements page.

## Describe your changes

I corrected the number to reflect the reimbursements page.

### DIFF
--- a/app/controllers/my_controller.rb
+++ b/app/controllers/my_controller.rb
@@ -139,7 +139,7 @@ class MyController < ApplicationController
     when "review"
       @reports = @reports_to_review
     else
-      @reports = @my_reports.or(@reports_to_review)
+      @reports = @my_reports.or(@reports_to_review).uniq
     end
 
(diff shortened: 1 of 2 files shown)


app/controllers/my_controller.rb
```diff
@@ -3,6 +3,8 @@
 class MyController < ApplicationController
   skip_after_action :verify_authorized, only: [:activities, :toggle_admin_activities, :cards, :missing_receipts_list, :missing_receipts_icon, :inbox, :reimbursements, :reimbursements_icon, :tasks, :payroll, :feed] # do not force pundit
 
+  before_action :load_reimbursements, only: [:reimbursements, :reimbursements_icon]
```
sampoder: This should be a private method and probably be `set_reimbursement_reports`

sampoder (approved):

PR #11472, opened 2025-08-28
[Filters] Improve amount filter with range slider

manuthecoder:
This pull request introduces a new dual range slider component for filtering invoice amounts, replacing the previous static amount range filter. 

**Invoice Filter Logic Updates:**

* Updated the invoice filter options in `invoices_controller.rb` to dynamically calculate the min and max invoice amounts and pass them to the amount range filter, enabling the slider to reflect the actual range of invoice amounts for the selected event.
* Removed the static `amount_range` filter from the default filter options to avoid duplication, as it is now dynamically added with computed values.

image


sampoder:
A couple of design things:

* I think the number inputs need labels (max/min)
* The inputs don't look like they're from our design system - specifically the padding at the bottom.
* We should display a dollar sign

Personally I would advocate for keeping the input stacked, like we do at the moment

Min
[                      ]

Max
[                      ]

The whitespace at the moment is a bit offputting

app/assets/stylesheets/components/_dual_range_slider.scss
sampoder, line 1: Can we use https://getbem.com/?

app/javascript/controllers/range_slider_controller.js
```diff
@@ -0,0 +1,175 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
+  static values = {
+    min: { type: Number, default: 0 },
+    max: { type: Number, default: 100 },
+    step: { type: Number, default: 1 },
+
+    lo: { type: Number, default: null },
+    hi: { type: Number, default: null },
```
sampoder: Can we use `low` and `high`?

app/javascript/controllers/range_slider_controller.js
```diff
@@ -0,0 +1,175 @@
...
+          <input type="number" class="rs-input" name="${this.keyBaseValue}_greater_than" />
+          <input type="number" class="rs-input" name="${this.keyBaseValue}_less_than" />
+        </div>
+        <div class="rs-track">
+          <div class="rs-bar"></div>
+          <div class="rs-range text-blue-600"></div>
+          <button type="button" class="rs-thumb text-blue-600"></button>
+          <button type="button" class="rs-thumb text-blue-600"></button>
+        </div>`
+
```
sampoder, line 152: Hmm - normally you would use targets with Stimulus: https://stimulus.hotwired.dev/reference/targets

manu-gurudath: resolved cuz of our slack conversation earlier

manu-gurudath:
done! let me know if anything else should be changed!

garyhtou (approved):
nice! merge once prettier is fixed

PR #11473, opened 2025-08-28
[Check Deposit] Allow member-roled depositor to view check img

garyhtou:
## Summary of the problem

Generally speaking, we hide check deposit images from readers and members because it contains the account & routing number of the donor (check sender).

However, we also allow members to upload check deposits. With our previous pundit policy, members would upload check deposits, but then wouldn't be able to see the image they uploaded.

## Describe your changes

This PR fixes that by allowing a user to view the check deposit images as long as they're an organizer (e.g. reader, member, etc.) AND also was the person who uploaded the check deposit.

This means the final pundit policy looks like:
```ruby
def view_image?
  # You can view the check deposit images (front & back) as long as you meet
  # at least one of the following conditions:
  # - You're an auditor (admin)
  # - You're a manager of the event
  # - You're an organizer of the event (e.g. reader, member, etc.), but ALSO
  #   was the person who uploaded the check deposit.
  auditor_or_manager? || (user? && record.created_by == user)
end
```

N.B. The "you're an organizer of the event" check is important because it prevents a user who uploaded a check deposit, but is later kicked from the organization, from accessing the check deposit images.

Although rare, this logic also intentionally allows for the following:
1. Member uploads a check deposit => they can view the check deposit images they uploaded.
2. The member gets demoted to reader => they can still view the check deposit images they uploaded.
3. The reader is removed from the org => they **lose access** to the images.

---

As a member, on the Check Deposit index page, I can only view image previews of check deposits that I uploaded.
image

As a member, I uploaded this check deposit, so I can view the images:
image

As a member, I did NOT upload this check deposit, so I CAN NOT view the images:
image

### DIFF
--- a/app/views/check_deposits/index.html.erb
+++ b/app/views/check_deposits/index.html.erb
@@ -34,7 +34,7 @@
   <div class="table-container">
     <table>
       <thead>
-        <% if @check_deposits.any? { |cd| policy(cd).view_image? } %>
+        <% if (can_view_any_check_images = @check_deposits.any? { |cd| policy(cd).view_image? }) %>
           <th></th>
         <% end %>
         <th>Status</th>
@@ -48,6 +48,8 @@
           <tr>
             <% if policy(check_deposit).view_image? %>
               <td style="width: 0%"><%= image_tag check_deposit.front.representation(resize_to_limit: [64, 64]), class: "rounded mr1", width: 40 %></td>
+            <% elsif can_view_any_check_images %>
+              <td></td>
             <% end %>
             <td><span class="badge bg-<%= check_deposit.state %> m0"><%= check_deposit.state_text %></span></td>
             <td><%= format_date check_deposit.created_at %></td>
--- a/app/policies/check_deposit_policy.rb
+++ b/app/policies/check_deposit_policy.rb
@@ -10,7 +10,13 @@ class CheckDepositPolicy < ApplicationPolicy
   end
 
   def view_image?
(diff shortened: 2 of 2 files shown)


app/views/check_deposits/index.html.erb
```diff
@@ -34,7 +34,7 @@
   <div class="table-container">
     <table>
       <thead>
-        <% if @check_deposits.any? { |cd| policy(cd).view_image? } %>
+        <% if (can_view_any_check_images = @check_deposits.any? { |cd| policy(cd).view_image? }) %>
```
Luke-Oldenburg, line 37: Where is this defined?

garyhtou: In this line, i'm defining a new variable called `can_view_any_check_images`. Here's a simplified version:

```ruby
if (a = (23 == 23))
  puts "yes"
else
  puts "no"
end

puts a
```

It results in:
```
yes
true
```

---

The result of `@check_deposits.any? { |cd| policy(cd).view_image? }` in stored in the variable called `can_view_any_check_images`, which is then used by the `if`. The variable `can_view_any_check_images` can then later be used in the code.

Luke-Oldenburg: Oh can you define a boolean and check its value in the same line?

garyhtou: yes, Ruby do be cool like that!

image

sampoder (approved):

PR #11474, opened 2025-08-28
[Accounting] Introduce the notion of event groups

davidcornu:
## Summary of the problem

As part of the work on categories (see #11298) and related reporting (see #11437) we want to be able to generate reports that combine several events (e.g. for Scrapyard) which aren't linked via sub-events.

## Describe your changes

- Introduces `Event::Group` and `Event::GroupMapping`
    - Groups are created by a user and are identified by a `name` that is globally unique
    - All admins can view and edit all groups
    - Group mappings connect events to groups
- Adds an index page for event groups where they can be created, deleted, and specific events can be removed
- Adds a new action to the admin events index that allows the groups the events are in to be edited

https://github.com/user-attachments/assets/eddf3f03-0673-4b72-bab5-b255f9181a18


app/controllers/admin/event_groups_controller.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+module Admin
+  class EventGroupsController < AdminController
```
garyhtou: wow, we aren't adding to the 1.5k line admin controller. this is lovely 🎉

db/migrate/20250827203528_add_event_groups.rb
```diff
@@ -0,0 +1,19 @@
+class AddEventGroups < ActiveRecord::Migration[7.2]
+  def change
+    create_table(:event_groups) do |t|
+      t.citext(:name, null: false)
+      t.references(:user, null: false, foreign_key: true, index: false)
+      t.timestamps
+
+      t.index([:user_id, :name], unique: true)
```
garyhtou: I have a feeling we want groups to be shared across all admins. There's a very likely case where Mel/Ops will create a group for Scrapyard, and then it's later shared with Zach, etc., who can easily grab reporting for that group.

I'd suggest we do a similar structure as Blazer queries. Any admin can create one. It says who created it. Anyone can edit. Anyone can view the list of queries.

We can enforce unique names across all event groups (not just scoped to user)

garyhtou (commented):
This is the right direction. What you've built out here will be very helpful.

When we introduce this to ops, we should remind them that:
- You should still use subevents and organization tags to organize your events
- This event groups feature is JUST for the accounting feature

To provide more context on why we're buildling this out, see:

image

app/controllers/admin/event_groups_controller.rb
davidcornu: Oh whoops `< AdminController` is actually bad. I autocompleted this thinking `AdminController` was a generic superclass but oh no. Will fix.

app/controllers/admin/event_group_memberships_controller.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Admin
+  class EventGroupMembershipsController < AdminController
```
davidcornu: TODO: This can't inherit from `AdminController`

db/migrate/20250827203528_add_event_groups.rb
davidcornu: > I'd suggest we do a similar structure as Blazer queries. Any admin can create one. It says who created it. Anyone can edit. Anyone can view the list of queries.
> 
> We can enforce unique names across all event groups (not just scoped to user)

Ah ok so we'll keep the `user_id` but that'll just be the creator.

app/controllers/admin/base_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Admin
+  class BaseController < ApplicationController
```
garyhtou, line 4: Thank you for this.

There are some other admin controllers in `/app/controllers/admin/` that should probably be re-inherited at some point (doesn't need to be in this PR)

app/models/admin/nav.rb
```diff
@@ -157,6 +157,12 @@ def misc
             path: referral_programs_admin_index_path,
             count: Referral::Program.count,
             count_type: :records
+          ),
+          make_item(
```
garyhtou, line 161: At some point, we should add this to the Command Bar (cmd+k), but that's low priority at the moment.

app/models/event/group.rb
```diff
@@ -0,0 +1,37 @@
...
+#
+#  index_event_groups_on_name     (name) UNIQUE
+#  index_event_groups_on_user_id  (user_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (user_id => users.id)
+#
+class Event
+  class Group < ApplicationRecord
```
garyhtou, line 23: Ty for namespacing this under Event

app/views/layouts/admin.html.erb
```diff
@@ -28,7 +28,8 @@
     <nav style="display: flex; flex-direction: column; background: var(--hcb-bg-2-5); gap: 0px">
       <div style="width: 100%; display: flex; gap: 8px; align-items: center; font-size: 22px; padding: 8px; border-bottom: 1px solid var(--hcb-bg-2);">
         <%= link_to root_path, class: "text-decoration-none", style: "height: 36px; margin-right: 8px;" do %>
-          <%= image_tag user_birthday? ? "logo-party.gif" : by_season("logo-#{Rails.env}.png", fall: "https://cloud-qmskqc293-hack-club-bot.vercel.app/0bank-pumpkin.png"),
+          <% logo_env = Rails.env.local? ? "development" : Rails.env %>
```
garyhtou, line 31: Ty

garyhtou (approved):

PR #11475, opened 2025-08-28
Remove empty tooltips

manuthecoder:
image

### DIFF
--- a/app/assets/javascripts/ui.js
+++ b/app/assets/javascripts/ui.js
@@ -135,8 +135,11 @@ window.attachTooltipListener = () => {
       const placement = [...trigger.classList].find(c => c.startsWith("tooltipped--"))?.split("--")[1] || "n";
       const offset = 5;
 
+      const label = trigger.getAttribute("aria-label");
+      if (!label.trim()) return;
+
       tooltip.className = "active";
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/ui.js
```diff
@@ -135,8 +135,11 @@ window.attachTooltipListener = () => {
       const placement = [...trigger.classList].find(c => c.startsWith("tooltipped--"))?.split("--")[1] || "n";
       const offset = 5;
 
+      const label = trigger.getAttribute("aria-label");
```
sampoder: Should this be:

```suggestion
      const label = trigger.getAttribute("aria-label").trim();
```

And then just:

```
if (!label) return;
```

app/assets/javascripts/ui.js
```diff
@@ -135,8 +135,11 @@ window.attachTooltipListener = () => {
       const placement = [...trigger.classList].find(c => c.startsWith("tooltipped--"))?.split("--")[1] || "n";
       const offset = 5;
 
+      const label = trigger.getAttribute("aria-label").trim();
+      if (!label.trim()) return;
```
sampoder:
```suggestion
      if (!label) return;
```

sampoder (approved):

PR #11477, opened 2025-08-28
[Ledger] Multiple filters

### LINKED ISSUE #8587: [Filter] Being able to add multiple filters of the same type

manu-gurudath:
This PR allows users to select the same filters multiple times for the ledger. This will resolve #8587

## Example
image


manu-gurudath:
double-check if the logic for filtering transactions is correct

app/controllers/events_controller.rb
```diff
@@ -1062,9 +1062,12 @@ def self.filter_transaction_type(type, settled_transactions:, pending_transactio
       }
     }
 
-    if type
-      filter = type_filters[type]
-      if filter
+    if type.present?
+      types = type.is_a?(Array) ? type : [type]
```
davidcornu: **nit**: You can leverage `Array()` here as well

```suggestion
      types = Array(type)
```

app/controllers/events_controller.rb
```diff
@@ -1219,7 +1222,9 @@ def set_transaction_filters
       @tag = Tag.find_by(event_id: @event.id, label: params[:tag])
     end
 
-    @user = @event.users.friendly.find(params[:user], allow_nil: true) if params[:user]
+    @user = if params[:user].present?
```
davidcornu: **suggestion**: Given that this is now an array, wouldn't `@users` make more sense?

```suggestion
    @users = if params[:user].present?
```

app/controllers/events_controller.rb
```diff
@@ -1219,7 +1222,9 @@ def set_transaction_filters
...
+              Array(params[:user]).map { |u| @event.users.friendly.find(u, allow_nil: true) }.compact
```
davidcornu: **concern** (minor): While I don't think this will end up being an issue in practice as I don't expect we'll ever have more than a handful, loading an array of records one at a time is a bit of a red flag. 

My understanding is that there isn't a "find many" version of FriendlyID so I guess this is the only option, I do wonder whether we should be using friendly IDs at all here given that these aren't really part of the URL. Going with regular IDs would let us simplify this to just

```ruby
@users = @event.users.where(id: Array(params[:user]))
```

app/controllers/events_controller.rb
```diff
@@ -1062,9 +1062,12 @@ def self.filter_transaction_type(type, settled_transactions:, pending_transactio
...
-      filter = type_filters[type]
-      if filter
+    if type.present?
+      types = Array(type)
+      types.each do |t|
+        filter = type_filters[t]
+        next unless filter
+
         settled_transactions = settled_transactions.select(&filter["settled"])
         pending_transactions = pending_transactions.select(&filter["pending"])
```
davidcornu, line 1026: **concern**: Is my understanding correct that this results in `type1 AND type2` rather than `type1 OR type2` in the presence multiple types? If so, does that make sense in this context?

app/services/transaction_grouping_engine/transaction/all.rb
```diff
@@ -98,9 +98,10 @@ def search_modifier_for(type)
       end
 
       def user_modifier
-        return "" unless @user&.stripe_cardholder&.stripe_id.present?
+        return "" unless @user.present? && @user.all? { |u| u&.stripe_cardholder&.stripe_id.present? }
```
davidcornu, line 108: **concern**: Given that `@user` is now an array I would rename it to `@users` (including the `user` keyword argument in `initialize`) and update the relevant call sites. Otherwise this is a bit surprising.

app/controllers/events_controller.rb
```diff
@@ -1219,7 +1222,9 @@ def set_transaction_filters
       @tag = Tag.find_by(event_id: @event.id, label: params[:tag])
     end
 
-    @user = @event.users.friendly.find(params[:user], allow_nil: true) if params[:user]
+    @users = if params[:user].present?
+              Array(params[:user]).map { |u| @event.users.friendly.find(u, allow_nil: true) }.compact
+            end
 
     @type = params[:type]
```
davidcornu, line 1214: **suggestion**: This should probably be `@types` as well

app/services/pending_transaction_engine/pending_transaction/all.rb
```diff
@@ -75,7 +75,8 @@ def canonical_pending_transactions
             end
 
             if @user
-              cpts = cpts.where("raw_pending_stripe_transactions.stripe_transaction->>'cardholder' = ?", @user&.stripe_cardholder&.stripe_id)
+              user_ids = Array(@user).map { |u| u.respond_to?(:stripe_cardholder) ? u.stripe_cardholder&.stripe_id : u }
```
davidcornu, line 80: **concern**: In what scenario would `u.respond_to?(:stripe_cardholder)` not return `true`?

app/helpers/application_helper.rb
```diff
@@ -4,8 +4,33 @@ module ApplicationHelper
...
+        if value.key?(:remove)
+          current -= Array(value[:remove]).compact
+        end
+
+        if value.key?(:add)
+          Array(value[:add]).compact.each do |v|
+            current << v unless current.include?(v)
+          end
+        end
+        current.empty? ? params_hash.delete(key) : params_hash[key] = current
```
davidcornu, line 28: **concern**: I'm not sure we should be adding this complexity to something as generic as `upsert_query_params` and I personally find the API unintuitive as we're mixing data that is merged in as-is with instructions on how to modify existing data.

I would instead recommend wrapping the param handling logic for the ledger filters in a separate class that can have methods for adding or removing users and types.

app/views/events/_filter_menu.html.erb
```diff
@@ -50,7 +50,7 @@
         <div data-tabs-target="tab" id="type">
           <% ["ach_transfer", "card_charge", "check_deposit", "donation", "fiscal_sponsorship_fee", "hcb_transfer", "invoice", "mailed_check", "paypal_transfer", "refund", "reimbursement", "wire"].each do |type| %>
             <div class="flex items-center">
-              <%= link_to(event_ledger_path(@event, upsert_query_params(type:)), class: "flex-auto menu__action #{"menu__action--active" if @type == type}", data: { turbo_prefetch: "false" }) do %>
+              <%= link_to(event_ledger_path(@event, upsert_query_params(type: { (@type&.include?(type) ? :remove : :add) => type })), class: "flex-auto menu__action #{"menu__action--active" if @type&.include?(type)}", data: { turbo_prefetch: "false" }) do %>
```
davidcornu: **question**: Given the introduction of safe navigation here (`@type&.include?(type)`), in which scenario could `@type` be `nil` here?

Luke-Oldenburg:
Closing this as we will just implement in the new engine.

PR #11478, opened 2025-08-28
Rescue invalid receipt files when creating transfers

sampoder:


app/controllers/ach_transfers_controller.rb
```diff
@@ -57,13 +57,16 @@ def create
...
-      end
+        begin
+          ::ReceiptService::Create.new(
+            uploader: current_user,
+            attachments: ach_transfer_params[:file],
+            upload_method: :transfer_create_page,
+            receiptable: @ach_transfer.local_hcb_code
+          ).run!
+        rescue ActiveRecord::RecordInvalid => e
+          return redirect_to event_transfers_path(@event), flash: { info: "ACH transfer successfully submitted, however, the receipt file was invalid." }
```
davidcornu, line 68: **concern**: How do we know that this exception relates to the receipt file?

sampoder: We can update this to invalid receipt

davidcornu: From a broader perspective, if every call to `ReceiptService::Create` needs to be wrapped in a generic `rescue` then I think we should reconsider the service's interface.

One pattern that I've used before is to include an errors object using https://api.rubyonrails.org/classes/ActiveModel/Errors.html, so that `.run` could return a boolean and it's up to the caller to check `.errors`.

davidcornu: See https://github.com/hackclub/hcb/blob/0b7710584cf72b0ff84ccd70caa2bc31d626208a/app/services/process_login_service.rb for an example.

sampoder: Hmm - I'm not sure what I think here. I don't really like that a transfer can be created with an invalid receipt in the first place. But because of the order of operations here that's the case

PR #11479, opened 2025-08-28
Rescue on argument errors in user controller

sampoder:
This should help with Stripe validations

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -336,7 +336,7 @@ class UsersController < ApplicationController
 
       render :edit, status: :unprocessable_entity
     end
-  rescue StripeInvalidNameError => e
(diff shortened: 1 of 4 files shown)


app/controllers/users_controller.rb
```diff
@@ -336,6 +336,8 @@ def update
 
       render :edit, status: :unprocessable_entity
     end
+  rescue ArgumentError => e
+    redirect_back_or_to edit_user_path(@user), flash: { error: e.message }
```
davidcornu: **concern**: see https://github.com/hackclub/hcb/pull/11460#discussion_r2307607551

sampoder: Going to use `Errors::StripeInvalidNameError`

PR #11480, opened 2025-08-28
Allow for updating feature previews of other users

Luke-Oldenburg:
## Summary of the problem

Before the page would show the feature previews for the current user instead of the selected user.

## Describe your changes

Now admins can toggle a user's feature previews from the UI.


sampoder (approved):
appreciate the work here!

PR #11483, opened 2025-08-28
Premium swag

alexbluo:
image

image


sampoder (changes requested):
Don't we already offer domains? I'm confused about this separate perk. We also should tone down the gradient. It doesn't look great repeated.

Luke-Oldenburg:
I don't really get the whole premium perks thing. We need an actual metric to measure it by because these perks don't require the same amount of teens.

Luke-Oldenburg:
> Don't we already offer domains? I'm confused about this separate perk.

We're making another domain perk.

> We also should tone down the gradient. It doesn't look great repeated.

I agree. If we're going to have repeated premium perks maybe we should put them in their own section that has the premium styling with normal cards inside of it.

alexbluo:
i was also confused by a second domains perk but mel and sean are trying a diff way to issue domains with card grants and this is their way of doing that

agree on gradients, will just remove border. maybe another section with premium styling sounds good but not priority

i think premium will just be any perks that require x active teens, only available for teen-led orgs. i also just realized robotics fee waiver was added and doesnt follow the same error message format and everything so ill update that too...

alexbluo:
image

need help making premium badge and ineligible button more opaque

will add bg images for domains and swag

think its better idea to put monthly active teens somewhere in team page or like anywhere else

Luke-Oldenburg:
Jukebox is kind of hard to read in the first one.

Luke-Oldenburg:
We could definitely use an overhaul of the promotions page in general but it's hard to do when it's constantly being updated.

sampoder:
Monthly active teens feels irrelevant on `/team` - at the end of the day we're a platform for managing finances and I'm not sure why that number would be relevant outside of the promotions page

alexbluo:
image

still need help making ineligible button and premium badge opaque.

active teens can be actually be counted by filtering for active teens on /team so removed the counter at bottom of cards

Luke-Oldenburg:
> active teens can be actually be counted by filtering for active teens on /team so removed the counter at bottom of cards

That's an admin tool. We should really put it inside of an `admin_tool`.

alexbluo:
ok where shld i move the active teen count for users then bc putting it at the bottom of every card like rn is weird

Luke-Oldenburg:
I think it's fine there.

alexbluo:
image

added active teen counts back. made more opaque. please take a look and merge asap

alexbluo:
@sampoder @Luke-Oldenburg bump

Luke-Oldenburg:
Can you add some shadow at the top? It's kind of hard to read the perk name. Also resolve merge conflicts.

alexbluo:
wym shadow at the top? like a gradient on the left of the wood background?

Luke-Oldenburg:
Not a gradient - just darken it.

alexbluo:
for which ones?

Luke-Oldenburg:
Also, run prettier.

Luke-Oldenburg:
> for which ones?

Definitely the last two. I think Jukebox is okay.

alexbluo:
image

Luke-Oldenburg:
I can't really notice the difference.

alexbluo:
pins and keychains is a completely diff background image.

domains is 25% darker

alexbluo:
image

40% darker now

Luke-Oldenburg:
wait im lowk colorblind can you have somebody else check. maybe it just looks bad to me

alexbluo:
image

realized heading was 10% transparent so fixed that
cc @sampoder

sampoder:
Red on purple is not enough contrast imo

Luke-Oldenburg:
for resolving conflicts just copy the three perks you have and accept incoming and then paste them inside of the teenagers check at the top

alexbluo:
image

i think this is def enough contrast. you can probably tell i rly like purple.

alexbluo:
image

Luke-Oldenburg:
I think this is good now. @sampoder can you give it a look please.

app/views/events/promotions.html.erb
```diff
@@ -7,43 +7,105 @@
...
-        <% else %>
-          <%= link_to "Apply",
-            fillout_form("briCUMzx5Sus", { "hcbprojname" => @event.name, "hcburl" => event_url(@event), "name" => current_user.name, "email" => current_user.email }),
-            class: "btn promo-premium-header bg-clip-padding",
-            target: :_blank %>
-        <% end %>
-      </div>
+<% if @event.organized_by_teenagers? || @active_teen_count > 0 %>
+  <% if !@event.plan.eligible_for_perks? %>
+    <% ineligible_reason = "Perks are not available for HQ organizations" %>
```
garyhtou:
```suggestion
    <% ineligible_reason = "Perks are not available for #{@event.plan.label} organizations" %>
```

app/views/events/promotions.html.erb
```diff
@@ -7,43 +7,105 @@
...
+  <% elsif !@perks_available %>
+    <% ineligible_reason = "You must be a manager and organization must not be on demo mode" %>
```
garyhtou:
```suggestion
    <% ineligible_reason = "You must be a manager and organization must not be in demo mode" %>
```

app/views/events/promotions.html.erb
```diff
@@ -7,43 +7,105 @@
...
+  </div>
+
+  <div class="card pb0 mb3 promo-pins-keychains" id="premium-pins">
+    <div class="card__banner card__banner--top flex justify-between items-center">
+      <h3 class="h1 py-4 my0 text-transparent promo-premium-header bg-clip-text inline-flex items-center flex-grow">
+        <span class="flex-grow">Pins & Keychains</span>
+        <span class="badge bg-muted mr-5" style="font-size: 15px; padding: 8px 14px;">Premium</span>
+      </h3>
+
+      <% ineligible_reason = "Organization must have 5 or more active teens" if @active_teen_count < 5 %>
```
garyhtou: By assigning and using the same variable name for all of these cards, you run the risk of them "bleeding" into each other's sections.

For example:
image

Here's what happened:
1. Org doesn't qualify for jukebox because they have less than 10 active teens.
2. `ineligible_reason = "Organization must have 10 or more active teens"` is ran
3. The `ineligible_reason` is never reset, and it still equals the string when it gets to the `if ineligible_reason` check for the other perks.

app/views/events/promotions.html.erb
```diff
@@ -7,43 +7,105 @@
...
+      </h3>
+
+      <% ineligible_reason = "Organization must have 10 or more active teens" if @active_teen_count < 10 %>
+      <% if ineligible_reason %>
+        <div class="tooltipped tooltipped--w inline-block" aria-label="<%= ineligible_reason %>">
+          <span class="btn disabled promo-premium-header bg-clip-padding">Ineligible</span>
+        </div>
+      <% else %>
+        <%= link_to "Apply",
+          fillout_form("briCUMzx5Sus", { "hcbprojname" => @event.name, "hcburl" => event_url(@event) + "/team", "name" => current_user.name, "email" => current_user.email }),
```
garyhtou:
```suggestion
          fillout_form("briCUMzx5Sus", { "hcbprojname" => @event.name, "hcburl" => event_team_url(@event), "name" => current_user.name, "email" => current_user.email }),
```

Same suggestion applies to the other cards too

app/assets/stylesheets/components/_promos.scss
```diff
@@ -187,18 +213,6 @@
   }
 }
 
-@keyframes premium-rotate-gradient {
```
garyhtou, line 190: You removed this, but it seems to still be referenced in the class definition for `promo-premium-border::before`.

image

The `promo-premium-border` class doesn't seem to be used, so maybe that's also good to remove?

alexbluo: thanks for catching this

sampoder (approved):

PR #11485, opened 2025-08-28
Add option for operations to charge fee for low-value wire transfers

YodaLightsabr:
## Summary of the problem
Currently, operations needs to manually make a disbursement if charging a fee for a low-value transfer is desired.

## Describe your changes
This PR adds a simple checkbox to accomplish the same thing:

image

### DIFF
--- a/app/controllers/wires_controller.rb
+++ b/app/controllers/wires_controller.rb
@@ -71,6 +71,19 @@ class WiresController < ApplicationController
 
     @wire.send_wire!
 
+    if params[:charge_fee] == "1"
+      disbursement = DisbursementService::Create.new(
+        name: "Low-value wire transfer fee",
+        destination_event_id: EventMappingEngine::EventIds::HACK_CLUB_BANK,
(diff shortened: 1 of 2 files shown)


app/controllers/wires_controller.rb
```diff
@@ -71,6 +71,19 @@ def send_wire
 
     @wire.send_wire!
 
+    if params[:charge_fee] == "1"
+      disbursement = DisbursementService::Create.new(
+        name: "Low-value wire transfer fee",
```
sampoder, line 76: Can we call it Under $500 wire...

sampoder (approved):
lgtm!

app/controllers/wires_controller.rb
YodaLightsabr: hmmmm... having an amount in the memo might be weird

sampoder: Ok let's stick with this

PR #11493, opened 2025-08-29
[Wise] Move recipient field to upper section

davidcornu:
## Summary of the problem

When processing Wise requests we first create the recipient and then send the transfer

## Describe your changes

Move the recipient ID field to the upper section

CleanShot 2025-08-29 at 11 20 29

### DIFF
--- a/app/views/admin/wise_transfer_process.html.erb
+++ b/app/views/admin/wise_transfer_process.html.erb
@@ -173,6 +173,15 @@
                   "@keyup": "amount = +$event.target.value * 100",
                   data: { controller: "truncate-decimal", action: "truncate-decimal#truncate blur->truncate-decimal#pad" } %>
               </div>
+              <div class="mb-2">
+                <%= form.label(:wise_recipient_id, "Wise recipient ID", class: "bold mb1") %> <br>
+                <%= form.text_field(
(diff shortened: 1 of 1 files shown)


app/views/admin/wise_transfer_process.html.erb
```diff
@@ -157,7 +157,7 @@
     <fieldset class="p-6" x-data="{ amount: null, balance: <%= @wise_transfer.event.balance_available_v2_cents %> }">
       <legend style="padding: 0px 8px">Send</legend>
           <%= form_with(model: @wise_transfer, local: true, url: wise_transfer_path(@wise_transfer), method: :patch) do |form| %>
-            <div class="field">
```
davidcornu: There is no `.field` CSS class in the Admin context. It comes from the form component https://github.com/hackclub/hcb/blob/0d57ec15300863ce33661db71c98a97852ab8304/app/assets/stylesheets/components/_forms.scss#L191-L193 which isn't included in https://github.com/hackclub/hcb/blob/da8ab875eb692070e38b4ba4c858d9bdfdebdc6a/app/assets/stylesheets/admin.scss.

Image

sampoder (approved):
lgtm - thank you!

PR #11496, opened 2025-08-29
[Accounting] Render a statement of activity for event groups

davidcornu:
## Summary of the problem

1. #11437 added the ability to render a statement of activity for a given `Event`
2. #11474 added the ability to group events together as an `Event::Group`

We wish to show (1) for (2) as we want to produce reports for related events (e.g. Scrapyard).

## Describe your changes

1. Extract common logic into `Event::StatementOfActivity`
2. Power the existing endpoint with (1)
3. Extract shared view logic from (2) into a partial
4. Use (1) and (3) to add an endpoint to render a statement of activity for a group
5. Minor styling tweaks to the groups index page
6. Minor improvements to the statement of activity template

https://github.com/user-attachments/assets/857cff8e-9719-47d9-b905-d1f5ca9ebba2


app/models/event/statement_of_activity.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+class Event
+  class StatementOfActivity
```
garyhtou, line 4: Thanks for extracting this out and making it reusable

app/views/events/_statement_of_activity_contents.html.erb
```diff
@@ -0,0 +1,72 @@
...
+    <th>Total Amount</th>
+  </tr>
+  </thead>
+  <tbody>
+  <% if statement_of_activity.transactions_by_category.empty? %>
+    <tr>
+      <td colspan="2" class="text-center p-14">
+        There were no transactions in this date range.
+      </td>
+    </tr>
```
garyhtou, line 61: Thanks for adding this!

garyhtou (approved):
Good stuff!

PR #11498, opened 2025-08-29
Report revision to AppSignal on frontend

polypixeldev:
## Summary of the problem

Some frontend bugs are hard to debug or know if they've been fixed because AppSignal doesn't report a revision on the frontend.

## Describe your changes

Exposes the `HATCHBOX_REVISION` environment variable to the frontend and passes it to AppSignal.


sampoder (approved):

sampoder:
I think you could have technically used a .js.erb file but let's avoid those

PR #11499, opened 2025-08-29
Add created by to referral program

Luke-Oldenburg:
## Summary of the problem

We need a way to track who creates which referral program. We will go back and manually backfill and then add a non null restraint.

### DIFF
--- a/db/migrate/20250829194238_add_creator_foreign_key_to_referral_programs.rb
+++ b/db/migrate/20250829194238_add_creator_foreign_key_to_referral_programs.rb
@@ -1,0 +1,13 @@
+class AddCreatorForeignKeyToReferralPrograms < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def up
+    add_foreign_key :referral_programs, :users, column: :creator_id, validate: false
+    add_index :referral_programs, :creator_id, algorithm: :concurrently
+  end
+
+  def down
+    remove_foreign_key :referral_programs, column: :creator_id
+    remove_index :referral_programs, :creator_id
+  end
+end
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -1388,7 +1388,7 @@ class AdminController < ApplicationController
   end
 
   def referral_program_create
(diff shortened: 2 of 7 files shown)


db/migrate/20250829194238_add_creator_foreign_key_to_referral_programs.rb
```diff
@@ -0,0 +1,13 @@
...
+  def up
+    add_foreign_key :referral_programs, :users, column: :creator_id, validate: false
+    add_index :referral_programs, :creator_id, algorithm: :concurrently
+  end
+
+  def down
+    remove_foreign_key :referral_programs, column: :creator_id
+    remove_index :referral_programs, :creator_id
+  end
+end
```
sampoder, line 13: I'm curious - why not use `change`?

Luke-Oldenburg: I think we had decided to try and use up / down going forwards after dealing with a lot of DB issues over summer.

sampoder: From my understanding, these migrations should be reversible in the `change` format as well. Any ways, either way is fine just was noticing it in a lot of PRs

sampoder (approved):

db/migrate/20250829194238_add_creator_foreign_key_to_referral_programs.rb
Luke-Oldenburg: There might actually be an issue with this migration. I'm going to test a bit more.

Luke-Oldenburg: I'm confused. Even after resetting my DB and rebuilding my codespace I'm getting errors that the constraint already exists.

Luke-Oldenburg: I don't think add_reference likes being reverted.

Luke-Oldenburg:
We're going to do this a different way.

PR #11502, opened 2025-08-30
[Filters] Date range picker

manuthecoder:
This PR significantly improves the UI for filtering date ranges. See preview below

image

https://github.com/user-attachments/assets/b0ee6722-fb11-4805-bb62-5dd9fc812b64


sampoder:
Looking really good! Just one piece of design feedback. I'm not sure the pills fit with our design system. I would personally vote for one continuous rectangle:

Screenshot 2025-08-30 at 11 43 00 PM

Though less rounded. Just slightly rounded.

app/helpers/events_helper.rb
```diff
@@ -136,8 +136,8 @@ def auto_discover_feed(event)
   private
 
   def validate_date_range(base, params)
-    less = params["#{base}_before"]
-    greater = params["#{base}_after"]
+    less = params["#{base}_after"]
+    greater = params["#{base}_before"]
```
garyhtou, line 140: lol. yeah now it's correct

garyhtou (commented):
Feedback with session with Manu:

We'll improve code readability and maintainbility before merging. Use existing method provided by browser or add a small JS package if it's worth it.

garyhtou (approved):

PR #11503, opened 2025-08-30
Remove show explore hack club column

Luke-Oldenburg:
## Summary of the problem

Screenshot From 2025-08-30 19-55-30


sampoder (changes requested):
this should be broken up into three PRs: ignoring the column, dropping the column and removing the ignore statement.

Luke-Oldenburg:
https://github.com/hackclub/hcb/pull/11511

sampoder (approved):

PR #11505, opened 2025-08-31
Small tweaks to referral programs page

Luke-Oldenburg:
- use form.label instead of label for
- remove click to copy from url
- fix user count


sampoder (approved):
lgtm! thank you

PR #11507, opened 2025-08-31
Tweak admin comments

Luke-Oldenburg:
- remove heading from user admin comments
- pass admin argument for user admin comments
- pass blankslate argument for event admin comments


sampoder (approved):
thank you!

PR #11511, opened 2025-08-31
Ignore show explore hack club column

Luke-Oldenburg:

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -1388,7 +1388,7 @@ class AdminController < ApplicationController
   end
 
   def referral_program_create
-    @referral_program = Referral::Program.new(name: params[:name], show_explore_hack_club: params[:show_explore_hack_club])
+    @referral_program = Referral::Program.new(name: params[:name], show_explore_hack_club: true)
 
     if @referral_program.save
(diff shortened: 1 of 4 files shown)


app/controllers/admin_controller.rb
```diff
@@ -1388,7 +1388,7 @@ def referral_programs
   end
 
   def referral_program_create
-    @referral_program = Referral::Program.new(name: params[:name], show_explore_hack_club: params[:show_explore_hack_club])
+    @referral_program = Referral::Program.new(name: params[:name], show_explore_hack_club: true)
```
sampoder:
```suggestion
    @referral_program = Referral::Program.new(name: params[:name])
```

sampoder: I'm fairly sure the first code wouldn't work as the column is ignored

Luke-Oldenburg: I was worried because there is a non null on the column.

sampoder: oh - ahhhhhhh can you remove that first?

Luke-Oldenburg: https://github.com/hackclub/hcb/pull/11526

sampoder (approved):

PR #11517, opened 2025-09-01
[DX] Replace present ternaries with presence

Luke-Oldenburg:

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -170,7 +170,7 @@ class AdminController < ApplicationController
   def bank_fees
     @page = params[:page] || 1
     @per = params[:per] || 100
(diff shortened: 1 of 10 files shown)


app/controllers/admin_controller.rb
```diff
@@ -1411,30 +1411,30 @@ def stream_data(content_type, filename, data, download = true)
   end
 
   def filtered_events(events: Event.all)
-    @q = params[:q].present? ? params[:q] : nil
-    @demo_mode = params[:demo_mode].present? ? params[:demo_mode] : "full" # full accounts only by default
+    @q = params[:q].presence || nil
```
Luke-Oldenburg:
```suggestion
    @q = params[:q].presence
```

sampoder (approved):

PR #11518, opened 2025-09-01
Remove almost from event card

Luke-Oldenburg:
## Summary of the problem

This felt cramped.
Screenshot From 2025-08-31 20-47-43


sampoder (approved):
https://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-distance_of_time_in_words includes some cleaner reasons to do this btw

PR #11524, opened 2025-09-01
Fix donation page policy

Luke-Oldenburg:
## Summary of the problem

We were forgetting to check if the donation page was available resulting in a 404 for some users. The else statement in the nav was also incorrect.

### DIFF
--- a/app/policies/event_policy.rb
+++ b/app/policies/event_policy.rb
@@ -169,7 +169,7 @@ class EventPolicy < ApplicationPolicy
   end
 
   def donation_overview?
-    show? && record.approved? && record.plan.donations_enabled?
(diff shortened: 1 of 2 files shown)


app/policies/event_policy.rb
```diff
@@ -169,7 +169,7 @@ def create_sub_organization?
   end
 
   def donation_overview?
-    show? && record.approved? && record.plan.donations_enabled?
+    show? && record.approved? && record.plan.donations_enabled? && record.donation_page_enabled?
```
Luke-Oldenburg, line 172: Should this have a ?

sampoder: Yes

Screenshot 2025-08-31 at 10 24 48 PM

sampoder (approved):

PR #11536, opened 2025-09-02
Default to receiving goods on ysws donations

Luke-Oldenburg:


sampoder (commented):
This doesn't feel great to me, it feels like something that should be in a plan. I'll defer final review to @garyhtou but this feels off to me

garyhtou:
Imma reply to your Slack thread about this!

PR #11541, opened 2025-09-02
[Accounting] Add XLSX download for Statement of Activity

garyhtou:
## Summary of the problem

We already have a web version of the Statement of Activity for Events and Event Groups (see https://github.com/hackclub/hcb/pull/11437 and https://github.com/hackclub/hcb/pull/11496).

Now we want an XLSX download for it. The advantage of XLSX over the web version is that XLSX allows you to expand into each category to view the transactions that make up said category. It uses the [Excel Outline feature](https://support.microsoft.com/en-us/office/outline-group-data-in-a-worksheet-08ce98c4-0063-4d42-8ac7-8278c49e9aff).

## Describe your changes

To make this happen, I added a new gem [`write_xlsx`](https://github.com/cxn03651/write_xlsx). I also needed to explicitly install Rubyzip (a dependency of `write_xlsx`). See https://github.com/cxn03651/write_xlsx/issues/127 for more info on why.

This PR adds a new "Download XLSX" button to the Statement of Activity pages.

_Screenshot from an Event Group's Statement of Activity:_
image

_Screenshot from an Event's Statement of Activity:_
image
<sup>sorry, i know it looks jank— that's a problem for later lol<sup>

Upon clicking the button, it downloads an XLSX that looks like:

image

Categories can be expanded by clicking the `+`:
image

Event groups have a bit more information:
- List of Events included in the event group
   image
   image

- Name of organization next to each transaction
   image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -73,6 +73,8 @@ gem "business_time"
 gem "poppler" # PDF parsing
 gem "wicked_pdf" # HTML to PDF conversion
 
+gem "write_xlsx" # Export Excel files
+gem "rubyzip", "< 3.0", ">= 2.3.0" # Force `write_xlsx` to use an older version of `rubyzip`. See https://github.com/cxn03651/write_xlsx/issues/127
 
 gem "rack-cors" # manage CORS
 gem "rack-attack" # rate limiting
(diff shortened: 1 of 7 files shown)


garyhtou:
TODO:
- Add option to have export async/emailed (maybe use Export model). Export can be slow with a lot of events
- Make it look better (mess with the font sizes)
- Pull statement of activity for subevents

PR #11542, opened 2025-09-02
[v4] Force SSL in redirect URI except localhost

YodaLightsabr:
## Summary of the problem
It's difficult to develop and publish HCB apps that run locally

## Describe your changes
Allow non-HTTPS localhost redirect URIs

### DIFF
--- a/config/initializers/doorkeeper.rb
+++ b/config/initializers/doorkeeper.rb
@@ -225,7 +225,7 @@ Doorkeeper.configure do
   #
   # force_ssl_in_redirect_uri !Rails.env.development?
   #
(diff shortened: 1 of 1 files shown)


config/initializers/doorkeeper.rb
```diff
@@ -225,7 +225,7 @@
   #
   # force_ssl_in_redirect_uri !Rails.env.development?
   #
-  # force_ssl_in_redirect_uri { |uri| uri.host != 'localhost' }
+  force_ssl_in_redirect_uri { |uri| uri.host != 'localhost' }
```
garyhtou: How does this affect URLs like `hcb://` which exist in production?

PR #11543, opened 2025-09-02
[GSuite] Use alias policy instead of account policy

Luke-Oldenburg:
## Summary of the problem

We were using account policy where we should have been using alias policy.

### DIFF
--- a/app/policies/g_suite_account_policy.rb
+++ b/app/policies/g_suite_account_policy.rb
@@ -26,7 +26,7 @@ class GSuiteAccountPolicy < ApplicationPolicy
   end
 
   def destroy?
-    user.admin?
+    admin_or_manager?
   end
 
   def reject?
--- a/app/policies/g_suite_account_policy.rb
+++ b/app/policies/g_suite_account_policy.rb
@@ -26,7 +26,7 @@ class GSuiteAccountPolicy < ApplicationPolicy
   end
 
   def destroy?
-    admin_or_manager?
+    user.admin?
   end
 
   def reject?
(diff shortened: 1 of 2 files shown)


app/policies/g_suite_account_policy.rb
```diff
@@ -26,7 +26,7 @@ def update?
   end
 
   def destroy?
-    user.admin?
+    admin_or_manager?
```
rluodev:
```suggestion
    user.admin?
```

This isn't the policy we should be editing - lowk my fault - Please edit the policy statement on `g_suite_accounts/_panel.html.erb` line 67 to use `!policy(al).destroy?` instead of `!policy(account).destroy?`

rluodev (changes requested):

app/views/g_suite_accounts/_panel.html.erb
```diff
@@ -75,7 +75,7 @@
             <% end %>
             <div class="flex flex-col justify-between">
               <div class="flex alias-creation">
-                <% if account.g_suite_aliases.count < 30 && policy(account).create? %>
+                <% if account.g_suite_aliases.count < 30 && policy(al).create? %>
```
rluodev:
```suggestion
                <% if account.g_suite_aliases.count < 30 && policy(account).create? %>
```

**context:** we check account policy on create because the alias doesn't exist at this point. Whether or not the account policy allows the current user to create is what we base the decision on (it's a bit janky). However, since our policy for destroy only allows admins to do it (for security reasons), and since the alias exists already when we check for destroy, we can authorize the action using the `GSuiteAliasPolicy`'s `destroy?` method instead.

rluodev (approved):
LGTM

sampoder (approved):

PR #11559, opened 2025-09-03
[Announcement] Fix the refresh block button

polypixeldev:
## Summary of the problem

For some reason, Turbo wasn't using the `data-turbo-method` attribute correctly on the refresh block button, so the button was sending a GET request instead of POST.

## Describe your changes

Changes the refresh block button to use `button_to` to send a POST request.


polypixeldev:
Found the root cause of the problem - see #11561

PR #11578, opened 2025-09-04
[Wise Reimbursements] Add quotes and show cost breakdown

YodaLightsabr:
## Summary of the problem
It's hard for organizers to see how much a transfer will cost

## Describe your changes

image

image


app/models/reimbursement/report.rb
```diff
@@ -335,7 +335,13 @@ def from_public_reimbursement_form?
     end
 
     def wise_transfer_quote_amount
-      @wise_transfer_quote_amount ||= WiseTransfer.generate_quote(amount_to_reimburse)
+      @wise_transfer_quote_amount ||= WiseTransfer.generate_quote(draft? ? amount : amount_to_reimburse)
```
sampoder: I  _**think**_ you can use `amount` directly:

```ruby
def amount_cents
      return amount_to_reimburse_cents if reimbursement_requested? || reimbursement_approved? || reimbursed?

      expenses.sum(:amount_cents)
end
```

sampoder (changes requested):
Code looks good but can you attach screenshots?

app/models/reimbursement/report.rb
YodaLightsabr: Oh good catch

sampoder (approved):

PR #11587, opened 2025-09-04
Handle string `charge` fields

sampoder:
For some reason this happens with voided invoices?


app/models/invoice.rb
```diff
@@ -277,13 +277,17 @@ def set_fields_from_stripe_invoice(inv = remote_invoice)
     self.starting_balance = inv.starting_balance
     self.statement_descriptor = inv.statement_descriptor
     self.status = inv.status
-    self.stripe_charge_id = inv&.charge&.id
+    if inv&.charge&.is_a?(String)
```
```suggestion
    if inv&.charge.is_a?(String)
```

PR #11590, opened 2025-09-04
[Announcement] Fix styling on donation goal block when there's 0 progress

polypixeldev:
## Summary of the problem

When there's zero progress made towards the goal, the donation goal block showed the "0%" to the far left of the progress bar, partially cut off.

## Describe your changes

Changed it to have whitespace instead of a percentage to keep the progress bar height.

### DIFF
--- a/app/views/announcements/blocks/_donation_goal.html.erb
+++ b/app/views/announcements/blocks/_donation_goal.html.erb
@@ -12,7 +12,9 @@
       </p>
       <div class="bg-gray-200 dark:bg-neutral-700 rounded-full w-full">
         <div class="h-full bg-primary rounded flex items-center justify-center" style="width: <%= percentage * 100 %>%">
-          <p class="text-sm text-black p-[1px] my-0"><%= number_with_precision(percentage * 100, precision: 1) %>%</p>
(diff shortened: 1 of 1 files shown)


app/views/announcements/blocks/_donation_goal.html.erb
```diff
@@ -12,7 +12,9 @@
       </p>
       <div class="bg-gray-200 dark:bg-neutral-700 rounded-full w-full">
         <div class="h-full bg-primary rounded flex items-center justify-center" style="width: <%= percentage * 100 %>%">
-          <p class="text-sm text-black p-[1px] my-0"><%= number_with_precision(percentage * 100, precision: 1) %>%</p>
+          <p class="text-sm text-black p-[1px] my-0">
+            <%= percentage > 0 ? "#{number_with_precision(percentage * 100, precision: 1)}%" : "&nbsp".html_safe %>
+          </p>
```
sampoder: What are you trying to achieve with the `&nbsp`?

polypixeldev: It was to add whitespace to keep the height of the progress bar, but I changed it to use just a fixed height instead

sampoder (approved):

PR #11596, opened 2025-09-05
Logout after inactivity

Luke-Oldenburg:
## Summary of the problem

Closes #7258
Supersedes #7303 and #7408
Screenshot From 2025-09-04 21-51-22


app/jobs/user_session/expiration_job.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class UserSession
+  class ExpirationJob < ApplicationJob
+    queue_as :low
+
+    def perform
+      UserSession.joins(:user)
+                 .where("user_sessions.last_seen_at + (users.session_validity_preference * interval '1 minute') < NOW()") # how does this handle nil?
```
Luke-Oldenburg:
```suggestion
                 .where("user_sessions.last_seen_at + (users.session_validity_preference * interval '1 minute') < NOW()")
```

db/schema.rb
```diff
@@ -2349,6 +2350,7 @@
     t.boolean "teenager"
     t.integer "creation_method"
     t.boolean "cards_locked", default: false, null: false
+    t.integer "session_validity_preference", default: 4320, null: false
```
sampoder: SESSION_DURATION_OPTIONS.values
=> [900, 3600, 21600, 86400, 259200, 604800, 1209600]

`4320` doesn't look like a valid option?

app/jobs/user_session/expiration_job.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class UserSession
+  class ExpirationJob < ApplicationJob
+    queue_as :low
+
+    def perform
+      UserSession.joins(:user)
+                 .where("user_sessions.last_seen_at + (users.session_validity_preference * interval '1 minute') < NOW()")
```
sampoder: Isn't `session_validity_preference` in seconds?

sampoder (changes requested):
I think we should be setting `expiration_at` and not bringing back peacefully expired. See https://github.com/hackclub/hcb/issues/8950

Luke-Oldenburg:
The job malted wrote wouldn't have worked well because of the timing of expiration so I refactored the logic to `signed_in_user` in `SessionsHelper`.

app/helpers/sessions_helper.rb
```diff
@@ -131,7 +132,9 @@ def current_session
   end
 
   def signed_in_user
-    unless signed_in?
+    if signed_in?
+      current_session.update!(expiration_at: Time.now + current_user.session_validity_preference)
```
sampoder: Maybe we do this every time we touch `last_active_at` instead?

app/controllers/users_controller.rb
```diff
@@ -300,6 +300,8 @@ def update
       end
     end
 
+    current_session.update!(expiration_at: Time.now + current_user.session_validity_preference) if params[:session_validity_preference]
+
```
sampoder: Could you add a comment with context here?

app/models/user_session.rb
```diff
@@ -72,14 +72,13 @@ def impersonated?
...
 
-  SESSION_DURATION = 2.weeks
-
   LAST_SEEN_AT_COOLDOWN = 5.minutes
 
   def touch_last_seen_at
     return if last_seen_at&.after? LAST_SEEN_AT_COOLDOWN.ago # prevent spamming writes
 
     update_columns(last_seen_at: Time.now)
+    current_session.update!(expiration_at: Time.now + current_user.session_validity_preference)
```
sampoder: this can just be
```suggestion
    update_columns(expiration_at: Time.now + user.session_validity_preference)
```
right?

app/controllers/users_controller.rb
```diff
@@ -300,6 +300,9 @@ def update
       end
     end
 
+    # extend the user session expiration to the new preference so you cant accidentally sign yourself out by updating it
```
sampoder:
```suggestion
    # extend the user session expiration to the new preference so you can't accidentally sign yourself out by updating it
```

app/helpers/sessions_helper.rb
```diff
@@ -34,7 +35,7 @@ def sign_in(user:, fingerprint_info: {}, impersonate: false, webauthn_credential
       if impersonate
         IMPERSONATED_SESSION_DURATION
       else
-        UserSession::SESSION_DURATION
```
sampoder, line 37: Why remove this?

app/views/users/edit_security.html.erb
```diff
@@ -133,6 +133,18 @@
       Sign out everywhere else
     <% end %>
   </div>
+
+  <%= form_with(model: @user, local: true, class: "pb2", data: { turbo_frame: "_top" }) do |form| %>
+    <div class="field">
+      <%= form.label :session_validity_preference, "Sign me out after" %>
+      <%= form.select :session_validity_preference, SessionsHelper::SESSION_DURATION_OPTIONS.to_a, selected: current_user.session_validity_preference || SessionsHelper::SESSION_DURATION_OPTIONS.fetch("3 days") %>
```
sampoder: Will `current_user.session_validity_preference` ever be null?

spec/support/session_support.rb
```diff
@@ -7,7 +7,7 @@ module SessionSupport
   # @param user [User]
   # @return [UserSession]
   def sign_in(user)
-    expiration_at = UserSession::SESSION_DURATION.seconds.from_now
+    expiration_at = Time.now + SessionsHelper::SESSION_DURATION_OPTIONS.fetch("2 weeks")
```
sampoder: Why change this?

app/controllers/users_controller.rb
```diff
@@ -300,6 +300,9 @@ def update
       end
     end
 
+    # extend the user session expiration to the new preference so you can't accidentally sign yourself out by updating it
+    current_session.update!(expiration_at: Time.now + current_user.session_validity_preference) if params[:session_validity_preference]
```
sampoder: Is this needed? How would `expiration_at` become in the past?

app/helpers/sessions_helper.rb
sampoder: Or should this be the user's preference?

Luke-Oldenburg: I just changed it from grabbing the constant from one location to another.

app/views/users/edit_security.html.erb
Luke-Oldenburg: https://github.com/hackclub/hcb/blob/9333c7fc0bb87f36098fdb57d048d6c1e4501668/app/models/user.rb#L25

spec/support/session_support.rb
Luke-Oldenburg: To consolidate all of the session related time constants to being grabbed from that object.

app/controllers/users_controller.rb
Luke-Oldenburg: I think you're right here.

app/views/users/edit_security.html.erb
Luke-Oldenburg: Oh I see what you mean by this comment now.

Luke-Oldenburg:
```suggestion
      <%= form.select :session_validity_preference, SessionsHelper::SESSION_DURATION_OPTIONS.to_a, selected: current_user.session_validity_preference %>
```

app/helpers/sessions_helper.rb
sampoder: If there is a default I think we want it defined in it's own variable but I'm a bit confused by we aren't using the user's preference here?

spec/support/session_support.rb
sampoder: https://github.com/hackclub/hcb/pull/11596/files#r2345770459

app/helpers/sessions_helper.rb
Luke-Oldenburg: This is for the hardcoded log out 2 weeks after the session was created.

sampoder: But wouldn't this be overridden the moment we update `last_active_at`?

app/models/user_session.rb
```diff
@@ -72,14 +72,13 @@ def impersonated?
...
 
-  SESSION_DURATION = 2.weeks
-
   LAST_SEEN_AT_COOLDOWN = 5.minutes
 
   def touch_last_seen_at
     return if last_seen_at&.after? LAST_SEEN_AT_COOLDOWN.ago # prevent spamming writes
 
     update_columns(last_seen_at: Time.now)
+    update_columns(expiration_at: Time.now + user.session_validity_preference)
```
sampoder: imo we should skip this for impersonated sessions

app/helpers/sessions_helper.rb
Luke-Oldenburg: Yea I think so. Do we want a hardcoded 2 week limit or do we want it to keep extending?

app/models/user_session.rb
Luke-Oldenburg:
```suggestion
    update_columns(expiration_at: Time.now + user.session_validity_preference) if !impersonated?
```

app/helpers/sessions_helper.rb
sampoder: I think we just keep extending it and drop this 2 week reference

app/models/user_session.rb
```diff
@@ -72,14 +72,13 @@ def impersonated?
...
 
-  SESSION_DURATION = 2.weeks
-
   LAST_SEEN_AT_COOLDOWN = 5.minutes
 
   def touch_last_seen_at
     return if last_seen_at&.after? LAST_SEEN_AT_COOLDOWN.ago # prevent spamming writes
 
     update_columns(last_seen_at: Time.now)
+    update_columns(expiration_at: Time.now + user.session_validity_preference) if !impersonated?
```
sampoder:
```suggestion
    update_columns(expiration_at: Time.now + user.session_validity_preference) unless impersonated?
```

sampoder (approved):
lgtm! @davidcornu can you take a quick look?

app/models/user_session.rb
```diff
@@ -72,14 +72,13 @@ def impersonated?
...
 
-  SESSION_DURATION = 2.weeks
-
   LAST_SEEN_AT_COOLDOWN = 5.minutes
 
   def touch_last_seen_at
     return if last_seen_at&.after? LAST_SEEN_AT_COOLDOWN.ago # prevent spamming writes
 
     update_columns(last_seen_at: Time.now)
+    update_columns(expiration_at: Time.now + user.session_validity_preference) unless impersonated?
```
davidcornu: **suggestion**: We can combine these into a single query


```suggestion
    updates = { last_seen_at: Time.now }
    updates[:expiration_at] = Time.now + user.session_validity_preference unless impersonated?
    update_columns(**updates) 
```

spec/controllers/ach_transfers_controller_spec.rb
```diff
@@ -19,7 +19,9 @@
     before do
       # janky way of signing in for tests, we basically override the call in sessions_helper that checks for current_user to return this test user
       user = create(:user)
+      session = create(:user_session)
       allow(controller).to receive(:current_user).and_return(user)
+      allow(controller).to receive(:current_session).and_return(session)
     end
```
davidcornu, line 22: **suggestion**: Probably out of scope for this PR but I've added code to do this in https://github.com/hackclub/hcb/blob/06025e92465cebeae990cbd434ea71080ef5aa0f/spec/support/session_support.rb

spec/controllers/users_controller_spec.rb
```diff
@@ -13,9 +13,6 @@
 
         initial_session = sign_in(admin_user)
 
-        # This is a normal session which should last for 2 weeks
-        expect(initial_session.expiration_at).to eq(2.weeks.from_now)
```
davidcornu, line 17: **concern**: Can we update this assertion rather than deleting it?

app/models/user_session.rb
davidcornu: **concern**: Do we have any test coverage for this behaviour (specifically that `touch_last_seen_at` bumps the expiry by the configured out)?

spec/controllers/ach_transfers_controller_spec.rb
Luke-Oldenburg: So this can be left alone?

app/models/user_session.rb
Luke-Oldenburg: No. How should I implement that?

spec/controllers/users_controller_spec.rb
```diff
@@ -13,8 +13,8 @@
...
+        # This is a normal session which should last for the admin user's session_validity_preference
```
Luke-Oldenburg:
```suggestion
        # This is a normal session which should last for the user's session_validity_preference
```

spec/controllers/ach_transfers_controller_spec.rb
davidcornu: Ideally we should replace this with `SessionSupport#sign_in` so we don't have to stub all these methods

davidcornu: Looking at this locally I believe you can just do

```suggestion
    before do
      user = create(:user)
      sign_in(user)
    end
```

app/models/user_session.rb
```diff
@@ -72,14 +72,14 @@ def impersonated?
     !impersonated_by.nil?
   end
 
-  SESSION_DURATION = 2.weeks
-
   LAST_SEEN_AT_COOLDOWN = 5.minutes
 
   def touch_last_seen_at
```
davidcornu: **concern**: This should probably be renamed

app/models/user_session.rb
davidcornu: See [`2ba3b1a` (#11596)](https://github.com/hackclub/hcb/pull/11596/commits/2ba3b1a3bdc6ad2061d5b2b0a0567a73e927374a)

Luke-Oldenburg: ty!

app/models/user_session.rb
```diff
@@ -74,7 +74,7 @@ def impersonated?
 
   LAST_SEEN_AT_COOLDOWN = 5.minutes
 
-  def touch_last_seen_at
+  def update_last_seen_at
```
davidcornu: My bad - I should have been a bit clearer in my comment. The reason I was suggesting this be renamed is that it longer just updates `last_seen_at`. Maybe something like `update_session_timestamps` or something?

davidcornu (approved):
Save for https://github.com/hackclub/hcb/pull/11596#discussion_r2370309333, LGTM.

garyhtou (commented):
Question: Does this mean that someone would keep a session forever alive? I feel like we should at least force users to sign in every X period (maybe at least every months; or 2 weeks).

Luke-Oldenburg:
> Question: Does this mean that someone would keep a session forever alive? I feel like we should at least force users to sign in every X period (maybe at least every months; or 2 weeks).


https://github.com/hackclub/hcb/pull/11596#discussion_r2353609588

garyhtou:
Thanks for surfacing that comment. I do strongly believe that there must be a hard time limit on sessions. We shouldn't allow sessions to exist forever.

Luke-Oldenburg:
I would agree with that.

garyhtou:
@sampoder, did you have any further thoughts here?

PR #11607, opened 2025-09-05
Add active teen count to admin organizations page

Luke-Oldenburg:
## Summary of the problem

As requested by Mel.


app/controllers/admin_controller.rb
```diff
@@ -1481,6 +1482,11 @@ def filtered_events(events: Event.all)
     states << "rejected" if @rejected
     relation = relation.where("events.aasm_state in (?)", states)
 
+    relation = relation.select do |event|
```
sampoder: Can you do this with SQL? Thanks!

sampoder (changes requested):

app/controllers/admin_controller.rb
Luke-Oldenburg: I spent like half an hour trying to figure that out. Could you help please.

sampoder:
```ruby
Event.joins(:users)
   .where(users: { teenager: true, id: User.active })
   .group('events.id')
   .having('COUNT(users.id) > 5')
```

I believe this works? My initial test in production suggested it did:

```
bank(prod)> Event.joins(:users)
bank(prod)>    .where(users: { teenager: true, id: User.active })
bank(prod)>    .group('events.id')
bank(prod)*    .having('COUNT(users.id) > 5').each do |e|
bank(prod)*      puts e.users.select { |u| u.active? && u.teenager? }.size
bank(prod)>    end
7
9
9
13
12
7
6
7
11
8
14
11
10
10
7
6
10
10
10
6
11
8
6
9
6
7
6
6
10
10
8
6
6
9
10
6
8
8
8
10
7
10
11
9
9
12
7
11
10
10
10
10
7
7
7
13
9
8
11
7
9
9
17
9
6
8
8
6
7
8
6
7
7
8
7
9
11
8
7
9
10
8
13
16
10
10
10
15
10
6
12
10
6
11
10
11
7
12
10
10
8
11
11
10
10
8
13
8
7
9
11
8
7
10
10
8
6
11
8
10
10
10
7
12
10
10
8
```

Luke-Oldenburg: Oh I was way overthinking that

Luke-Oldenburg: Not working for me in dev mode.

Luke-Oldenburg: We're going to go without the filter for now.

sampoder (approved):
Merging - this will slow down this page a fair bit so we may need to turbo frame it

Luke-Oldenburg:
> Merging - this will slow down this page a fair bit so we may need to turbo frame it

Load time doesn't seem to be too bad in production.

PR #11609, opened 2025-09-06
Always show admin tool in user_mention

Luke-Oldenburg:
## Summary of the problem

These always mess with the spacing on elements because they were able to change size. We also can't do transparency because it would leave a strange gap after the user mention while not hovered.


sampoder (changes requested):
Hmm. I feel like this clutters the user mentions though

garyhtou:
Ideally, we have better tooltips and move the admin tool stuff in there. This way, it can still be hidden by default until hovered, and with the tooltips, there's no layout shift.

Example with github:
image

PR #11613, opened 2025-09-08
Add last frozen by back to views

Luke-Oldenburg:
## Summary of the problem

This got accidentally dropped at some point.

### DIFF
--- a/app/views/card_grants/_card_details.html.erb
+++ b/app/views/card_grants/_card_details.html.erb
@@ -8,6 +8,21 @@
 
       <section class="details">
         <div class="fs-mask">
+          <strong>Status</strong>
+          <span class="flex items-baseline g-1">
+            <%= status_badge stripe_card.status_badge_type %>
+            <%= stripe_card.status_text.capitalize %>
+          </span>
+        </div>
+
+        <% if stripe_card.frozen? %>
+          <div class="fs-mask">
+            <strong>Frozen by</strong>
+            <%= user_mention stripe_card.last_frozen_by %>
+          </div>
+        <% end %>
+
+        <div class="fs-mask">
           <strong>Card number</strong>
           <div class="flex justify-between font-mono items-center">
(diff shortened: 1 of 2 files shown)


app/views/card_grants/_card_details.html.erb
```diff
@@ -7,6 +7,21 @@
...
+            <%= stripe_card.status_text.capitalize %>
+          </span>
+        </div>
+
+        <% if stripe_card.frozen? %>
+          <div class="fs-mask">
+            <strong>Frozen by</strong>
+            <%= user_mention stripe_card.last_frozen_by %>
+          </div>
+        <% end %>
```
sampoder: I think we should only show this if `last_frozen_by` is defined

Luke-Oldenburg: Good point. In theory it will always be defined now for all frozen cards.

app/views/card_grants/_card_details.html.erb
```diff
@@ -7,6 +7,21 @@
...
       <section class="details">
+        <div class="fs-mask">
+          <strong>Status</strong>
+          <span class="flex items-baseline g-1">
+            <%= status_badge stripe_card.status_badge_type %>
+            <%= stripe_card.status_text.capitalize %>
+          </span>
+        </div>
+
+        <% if stripe_card.frozen? %>
```
Luke-Oldenburg:
```suggestion
        <% if stripe_card.frozen? && stripe_card.last_frozen_by %>
```

app/views/stripe_cards/_details.html.erb
```diff
@@ -32,6 +32,13 @@
       </p>
     <% end %>
 
+    <% if stripe_card.frozen? %>
```
Luke-Oldenburg:
```suggestion
    <% if stripe_card.frozen? && stripe_card.last_frozen_by %>
```

sampoder (approved):

PR #11621, opened 2025-09-09
Create Stripe customers for donations

### LINKED ISSUE #11614: [Donations] Send Donor name & email to Stripe for improved fraud detection
We aren't doing this currently, which means stripe is missing some info for fraud detection

sampoder:
Closes https://github.com/hackclub/hcb/issues/11614

### DIFF
--- a/app/models/donation.rb
+++ b/app/models/donation.rb
@@ -375,9 +375,10 @@ class Donation < ApplicationRecord
     self.event.donations.succeeded.size == 1
   end
 
-  def create_payment_intent_attrs
+  def create_payment_intent_attrs(customer)
     {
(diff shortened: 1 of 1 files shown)


app/models/donation.rb
```diff
@@ -386,7 +387,8 @@ def create_payment_intent_attrs
   end
 
   def create_stripe_payment_intent
-    payment_intent = StripeService::PaymentIntent.create(create_payment_intent_attrs)
+    customer = StripeService::Customer.create(email:, name:)
```
garyhtou, line 390: Does this behave like a _find or create_? Or is it just a create?

PR #11629, opened 2025-09-09
Improve admin events csv export

Luke-Oldenburg:

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -447,7 +447,7 @@ class Event < ApplicationRecord
     country
     slug "URL" do |slug| "https://hcb.hackclub.com/#{slug}" end
     is_public "Transparent"
-    users "Active teenagers" do |users| users.count { |user| user.teenager? && user.active? } end
+    users "Active teenagers" do |users| users.active_teenager.distinct.count end
   end
 
   CUSTOM_SORT = Arel.sql(
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -443,8 +443,6 @@ class Event < ApplicationRecord
     id
     created_at
     name
-    balance
-    total_raised
     revenue_fee
     country
     slug "URL" do |slug| "https://hcb.hackclub.com/#{slug}" end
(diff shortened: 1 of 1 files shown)


app/models/event.rb
```diff
@@ -441,11 +441,15 @@ def ancestor_organizer_positions
...
     name
+    balance
+    total_raised
     revenue_fee
-    slug "url" do |slug| "https://hcb.hackclub.com/#{slug}" end
     country
-    is_public "transparent"
+    slug "Url" do |slug| "https://hcb.hackclub.com/#{slug}" end
+    is_public "Transparent"
+    users "Active teenagers" do |users| users.count { |user| user.teenager? && user.active? } end
```
sampoder: Can this be done in SQL please

app/models/event.rb
```diff
@@ -441,11 +441,15 @@ def ancestor_organizer_positions
...
     id
+    created_at
     name
+    balance
+    total_raised
     revenue_fee
-    slug "url" do |slug| "https://hcb.hackclub.com/#{slug}" end
     country
-    is_public "transparent"
+    slug "Url" do |slug| "https://hcb.hackclub.com/#{slug}" end
```
sampoder: Why was the capitalisation changed?

app/models/event.rb
```diff
@@ -441,11 +441,15 @@ def ancestor_organizer_positions
 
   comma do
     id
+    created_at
     name
+    balance
+    total_raised
```
sampoder: This should be done in the events balances export

app/models/event.rb
Luke-Oldenburg: To match the other columns.

app/models/event.rb
Luke-Oldenburg: So just leave this out?

app/models/event.rb
sampoder:
```suggestion
    slug "URL" do |slug| "https://hcb.hackclub.com/#{slug}" end
```

app/models/event.rb
sampoder: Yes

sampoder (approved):

PR #11631, opened 2025-09-10
Extract perks to partials

Luke-Oldenburg:


sampoder (approved):

sampoder:
thank you this is a lot cleaner!

PR #11654, opened 2025-09-11
Redirect non-admins to download page for common cocuments

anishanne:
## Summary of the problem
When sharing the link to a common doc with a user, I [often](https://hackclub.slack.com/archives/CN523HLKW/p1757272937977879?thread_ts=1757272008.806459&cid=CN523HLKW) click go into the documents tab of their org and click the document, then I share the url.

Ex: https://hcb.hackclub.com/documents/common-hack-foundation-990-2022

Unfortunately, this URL leads to a non-authorized page, when it could simply just redirect to the download page:

https://hcb.hackclub.com/documents/common-hack-foundation-990-2022/download

## Describe your changes
If a non admin user visits a common doc page, they are redirected directly to the download page.

### DIFF
--- a/app/views/donations/_donation_form.html.erb
+++ b/app/views/donations/_donation_form.html.erb
@@ -94,7 +94,7 @@
     <% if @event.config.anonymous_donations %>
       <div class="field field--checkbox mb0">
         <%= form.check_box :anonymous, style: "margin-left: 0;" %>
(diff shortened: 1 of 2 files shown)


app/views/donations/_donation_form.html.erb
sampoder, line 1: Can you remove unrelated changes?

anishanne: Sorry yah, just noticed them lol

PR #11660, opened 2025-09-12
Add new receipt input partial with support for receipt bin

polypixeldev:
## Summary of the problem

https://hackclub.slack.com/archives/C068U0JMV19/p1757243106361769 - it'd be great to be able to choose from the receipt bin when attaching a receipt to a form! Plus, the current input for receipts was already repeated a lot and should be extracted into a partial.

## Describe your changes

This extracts the receipt input into a new partial, which integrates the existing file upload with the receipt bin link form, which has also been extracted into a new partial. With the new `receipt-input` Stimulus controller, users can upload or link any number of receipts to the form. Receipt data extraction is supported for receipts linked from the bin.

On the controller side, each controller will parse the receipts parameter in addition to the file parameter and link them to the new receiptable if authorized.

https://github.com/user-attachments/assets/66ccd5c5-c02b-481f-b014-297288acb407


app/controllers/ach_transfers_controller.rb
```diff
@@ -56,6 +56,16 @@ def create
     end
 
     if @ach_transfer.save
+      if params[:ach_transfer][:receipts]
+        receipts = params[:ach_transfer][:receipts].split ","
```
sampoder: This is odd. Can you look into a way that doesn't involve treating the input like a string?

app/controllers/receipts_controller.rb
```diff
@@ -92,6 +92,12 @@ def link_modal
 
   end
 
+  def inline_link
+    skip_authorization
+
+    @receipts = Receipt.in_receipt_bin.with_attached_file.where(user: current_user).order(created_at: :desc)
+  end
```
sampoder: We should authorise this imo

sampoder: This can be as simple as `current_user.receipts.build`

app/controllers/receipts_controller.rb
```diff
@@ -92,6 +92,12 @@ def link_modal
 
   end
 
+  def inline_link
+    skip_authorization
+
+    @receipts = Receipt.in_receipt_bin.with_attached_file.where(user: current_user).order(created_at: :desc)
```
sampoder: How come we aren't doing `current_user.receipts`[...]

app/javascript/controllers/extraction_controller.js
```diff
@@ -26,12 +26,20 @@ export default class extends Controller {
     Turbo.navigator.delegate.adapter.progressBar.hide()
     document.querySelector('html').setAttribute('data-ai-loading', false)
 
+    this.pasteData(response)
+  }
+
+  pasteData(response) {
```
sampoder, line 32: by "pasting" do you mean filling the form?

app/javascript/controllers/receipt_input_controller.js
```diff
@@ -0,0 +1,75 @@
...
+    outlet.render()
+  }
+
+  addFile() {
+    const files = this.fileInputTarget.files
+    const newFile = files[files.length - 1]
+
+    const html = `<li>${newFile.name}</li>`
+    this.listTarget.innerHTML += html
+    this.clearButtonTarget.classList.remove('hidden')
```
sampoder: I think may want to work on the styling a bit more. The list feels a little out of place at the moment.

app/controllers/receipts_controller.rb
polypixeldev: I chose to skip authorization here since the the `/my/inbox` page doesn't authorize anything either - what would the policy be checking?

app/controllers/receipts_controller.rb
polypixeldev: Had copied this from `my_controller` as well - changed to use `current_user.receipts`!

app/javascript/controllers/extraction_controller.js
polypixeldev: Yes

app/javascript/controllers/receipt_input_controller.js
polypixeldev: worked on a new list that incorporates the file preview (using the payment-docs icon as fallback when there's no preview) - what do you think?
image

polypixeldev:
Closing as this PR is so far behind main, pre-UI3, and there are more changes coming to sending transfers. While this is still a feature I would like, it'll be easier to start from a new branch and use this as reference.

PR #11662, opened 2025-09-13
[v4] Pass in Stripe version to cards ephemeral key

Mohamad-Mortada:
## Summary of the problem
Ideally the stripe versions between the digital wallet mobile app client which is on version `2020-08-27` should be matching the server one. (Tested this manually before but didn't change issues I had with digital wallet issuing but it is something that is expected)

## Describe your changes
Added ability to pass in `stripe_version` as param

### DIFF
--- a/app/models/stripe_card.rb
+++ b/app/models/stripe_card.rb
@@ -387,8 +387,8 @@ class StripeCard < ApplicationRecord
     Time.now.utc > Time.new(stripe_exp_year, stripe_exp_month).end_of_month
   end
 
-  def ephemeral_key(nonce:)
-    Stripe::EphemeralKey.create({ nonce:, issuing_card: stripe_id }, { stripe_version: "2020-03-02" })
+  def ephemeral_key(nonce:, stripe_version: "2020-03-02")
+    Stripe::EphemeralKey.create({ nonce:, issuing_card: stripe_id }, { stripe_version: })
   end
 
   private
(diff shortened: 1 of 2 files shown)


app/models/stripe_card.rb
```diff
@@ -387,8 +387,8 @@ def expired?
     Time.now.utc > Time.new(stripe_exp_year, stripe_exp_month).end_of_month
   end
 
-  def ephemeral_key(nonce:)
-    Stripe::EphemeralKey.create({ nonce:, issuing_card: stripe_id }, { stripe_version: "2020-03-02" })
+  def ephemeral_key(nonce:, stripe_version: "2020-03-02")
```
sampoder, line 390: by default, should we not pass in `nil`?

Mohamad-Mortada: I believe the stripe version is a required value (I could be wrong) so it would be nice to have a default (2020-03-02 is what we've been using on server side for everything mostly)

sampoder: We have an app-wide Stripe version configured in `Partners::Stripe::Shared::Opts`

Mohamad-Mortada: 2020-03-02 is the minimum required version for ephemeral keys. the app wide version is older

https://docs.stripe.com/issuing/elements

sampoder (approved):

PR #11663, opened 2025-09-13
[v4] Add physical card shipping details

Mohamad-Mortada:
## Summary of the problem
Would be nice to show shipping estimate on mobile

## Describe your changes
Added shipping estimate, address, and status


sampoder (approved):
I wouldn't rely too much on this data - it's spotty at best

PR #11667, opened 2025-09-14
Display last active date to all members of org

anishanne:
## Summary of the problem
#11653 

## Describe your changes
Removed 4 lines to display last active date to all members of the org


Luke-Oldenburg:
This will display it to anybody looking at the team page including non-members.

sampoder (changes requested):
I don't think this information is relevant on this page, if we are trying to show people information about promotions can we please keep it to the promotions page? We're a banking tool, it is very odd to be reporting last active imo.

PR #11670, opened 2025-09-15
Don't consider impersonated sessions when calculating `last_seen_at` and `last_login_at` on `User`

sampoder:


app/jobs/user/seen_at_history/snapshot_job.rb
```diff
@@ -20,7 +20,7 @@ def records
       end
 
       def users
-        ::User.joins(:user_sessions).where("user_sessions.last_seen_at between ? and ?", period_start, period_end).distinct
+        ::User.joins(:user_sessions).where("user_sessions.last_seen_at between ? and ? AND user_sessions.impersonated_by_id is NULL", period_start, period_end).distinct
```
davidcornu: **suggestion** I think you can get away with


```suggestion
        ::User
          .joins(:user_sessions)
          .where(
          	user_sessions: { 
          		last_seen_at: period_start..period_end, 
          		impersonated_by_id: nil 
          	}
          )
          .distinct
```

davidcornu (approved):

PR #11675, opened 2025-09-15
Use role_at_least in `present_in_events?`

sampoder:


app/policies/hcb_code_policy.rb
```diff
@@ -62,7 +62,7 @@ def user_made_purchase?
   private
 
   def present_in_events?
-    record.events.select { |e| e.try(:users).try(:include?, user) }.present?
+     record.events.any? { |event| OrganizerPosition.role_at_least?(user, event, :reader) }
```
sampoder:
```suggestion
    record.events.any? { |event| OrganizerPosition.role_at_least?(user, event, :reader) }
```

PR #11683, opened 2025-09-16
Collect `account_holder` name for Wise reimbursements

### LINKED ISSUE #11682: [Admin] feature name here Wire reimbursements should allow user to enter a name
### Describe the problem

Wire reimbursements are failing because HCB assumes the user's name on HCB is the name of the bank account holder. A field to enter a name should be added. This happens in YSWS mostly because teens are paying out of pocket and then the organizer sends them a reimbursement report to open and fill out.

a field for account owner's name

sampoder:
Closes https://github.com/hackclub/hcb/issues/11682. Not sure how much I like accessing the class inside the concern.

### DIFF
--- a/app/models/concerns/has_wise_recipient.rb
+++ b/app/models/concerns/has_wise_recipient.rb
@@ -24,6 +24,8 @@ module HasWiseRecipient
     def self.information_required_for(currency)
       fields = []
 
+      fields << { type: :text_field, key: "account_holder", placeholder: "Fiona Hackworth", label: "Account holder's name" }
+
       if currency.in?(%w[AED BGN CHF CZK DKK EGP EUR GBP GEL HUF ILS NOK PKR PLN RON SEK TRY UAH])
         fields << { type: :text_field, key: "account_number", placeholder: "TR330006100519786457841326", label: "IBAN" }
       elsif currency.in?(%w[HKD NGN NPR NZD PHP SGD THB])
--- a/app/models/concerns/has_wise_recipient.rb
+++ b/app/models/concerns/has_wise_recipient.rb
@@ -24,7 +24,7 @@ module HasWiseRecipient
     def self.information_required_for(currency)
       fields = []
 
-      if self.is_a?(User::PayoutMethod::WiseTransfer)
+      if self.name.constantize == User::PayoutMethod::WiseTransfer
         fields << { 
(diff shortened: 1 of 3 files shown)


app/models/concerns/has_wise_recipient.rb
```diff
@@ -24,6 +24,15 @@ module HasWiseRecipient
     def self.information_required_for(currency)
       fields = []
 
+      if self.name.constantize == User::PayoutMethod::WiseTransfer
```
davidcornu: **question**: Do we need the indirection via `.name.constantize`?

```suggestion
      if self == User::PayoutMethod::WiseTransfer
```

sampoder: That works! I tried `.is_a?()` first but that didn't work

app/views/wise_transfers/_currency_specific_details.html.erb
```diff
@@ -1,9 +1,9 @@
-<%# locals: (form:, disabled:, excluded_fields: []) %>
+<%# locals: (form:, disabled:, excluded_fields: [], model: User::PayoutMethod::WiseTransfer) %>
```
davidcornu: **concern**: Does this default mean that calls to this partial that don't specific `model` will get the extra field? Do we want that?

davidcornu: e.g. here https://github.com/hackclub/hcb/blob/74a9364f3bbb4b3b0c6685834a092cbb460d082b/app/views/wise_transfers/new.html.erb#L81

sampoder: Oh good catch this should be reversed!
```suggestion
<%# locals: (form:, disabled:, excluded_fields: [], model: WiseTransfer) %>
```

app/models/concerns/has_wise_recipient.rb
davidcornu: > I tried `.is_a?()` first but that didn't work

Yeah `is_a?` works on instances but not on the class objects themselves (which is what we're dealing with in a `def self.*` context).

sampoder: Interesting - good to know

davidcornu (approved):

PR #11697, opened 2025-09-17
Update Rails to `7.2.2.2`

davidcornu:
## Summary of the problem

In #10720 I switch our Rails dependency to point to the `7-2-stable` branch of the git repo as there hadn't been a `7.2` release in some time and we wanted to take advantage of some of the changes that had been merged but not released. 

Since then `7.2.2.2` was released which contains everything we were already running plus some additional changes (https://github.com/rails/rails/compare/21ef9d1554eb1fa8f3ecf997fdf165add8df0917...v7.2.2.2) including a security fix (https://rubyonrails.org/2025/8/13/Rails-Versions-8-0-2-1-7-2-2-2-and-7-1-5-2-have-been-released).

## Describe your changes

- Switches the `rails` entry in the `Gemfile` to `~> 7.2`
- Includes the result of running `bundle update rails --conservative`


sampoder (approved):
thank you!

PR #11710, opened 2025-09-19
Make sure sub-events cannot be cyclical

davidcornu:
## Summary of the problem

In https://github.com/hackclub/hcb/pull/11705 I added a mechanism for admins to set the parent for a given event. This open up the possibility for cycles in the event &rarr; sub-event graph which would cause infinite loops in a few places.

## Describe your changes

Adds a validation to `Event` to ensure `parent_id` cannot be changed to a value that would create a cycle.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -161,6 +161,9 @@ class Event < ApplicationRecord
   belongs_to :parent, class_name: "Event", optional: true
   has_many :subevents, class_name: "Event", foreign_key: "parent_id"
 
+  MAX_PARENT_DEPTH = 50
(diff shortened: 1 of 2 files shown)


app/models/event.rb
```diff
@@ -952,4 +955,31 @@ def stripe_transaction_merchant(transaction)
...
+          break :halted
+        end
+      else
+        break :halted
+      end
+    end
+
+    if outcome != :halted
+      errors.add(:parent, "max depth exceeded")
+    end
```
garyhtou, line 982: good check!

garyhtou (approved):

PR #11716, opened 2025-09-20
Improve reimbursment editing

manuthecoder:
This PR improves reimbursement editing

| Before | After |
|--------|--------|
| image | image |


sampoder:
I think I prefer the previous design. I'm not sure why max reimbursable value is the largest field in the new design.

PR #11717, opened 2025-09-20
Improve dropdown menu styling

manuthecoder:
This PR improves dropdown menu styling so the UI remains consistent with other menus in HCB

## Before
image

## After

image


YodaLightsabr:
Thanks for doing this! I'd also suggest reevaluating our color choices for the selected option. It feels a little off / doesn't look like a color we use elsewhere in the codebase.

YodaLightsabr:
(Unless you've already done that—then update the screenshot and lmk)

YodaLightsabr (approved):

PR #11724, opened 2025-09-22
Fix and test `build_plan` behaviour of `before_validation` hook in `Event`

davidcornu:
## Summary of the problem

We have logic that sets an event's plan based on its parent (if present) which is currently broken.

## Describe your changes

- Extract the plan selection logic to a separate method (given the complexity of the conditional expression)
- Add tests

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -425,7 +425,7 @@ class Event < ApplicationRecord
   end
 
   before_validation do
-    build_plan(type: parent&.subevent_plan&.class || parent&.plan&.class || Event::Plan::Standard) if plan.nil?
+    build_plan(type: fallback_plan_class) if plan.nil?
   end
 
   # Explanation: https://github.com/norman/friendly_id/blob/0500b488c5f0066951c92726ee8c3dcef9f98813/lib/friendly_id/reserved.rb#L13-L28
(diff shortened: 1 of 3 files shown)


app/models/event.rb
```diff
@@ -425,7 +425,7 @@ def ancestor_organizer_positions
   end
 
   before_validation do
-    build_plan(type: parent&.subevent_plan&.class || parent&.plan&.class || Event::Plan::Standard) if plan.nil?
```
davidcornu, line 428: 1. `Event` does not have a `subevent_plan` method (it is only present on `Event::Configuration`) https://github.com/hackclub/hcb/blob/686c02d34b1d9c4de99a32afdf271c0630fec8be/app/models/event/configuration.rb#L12
2. `subevent_plan` returns a `String` so calling `.class` on it would have resulted in us trying to construct a `String` plan which would fail

spec/factories/event_factory.rb
```diff
@@ -9,7 +9,7 @@
     end
 
     after(:create) do |event, context|
-      event.plan.update(type: context.plan_type)
+      event.plan.update!(type: context.plan_type) if context.plan_type.present?
```
davidcornu, line 12: My first instinct was to disable the `plan_type` transient by default but we have a lot of places in the test suite that depend on `Event::Plan::FeeWaived` being the default rather than what the logic in `Event` chooses.

Instead, this allows `plan_type` to be selectively overridden (which isn't ideal)

sampoder (approved):

PR #11725, opened 2025-09-22
[Accounting] Add descendant support to statement of activity

davidcornu:
## Summary of the problem

1. In #11474 we added the ability for users to create arbitrary lists of events so that a single statement of activity can be generated from their combined transactions. 
2. (1) is a stopgap measure as our ideal scenario is to leverage sub-events (https://github.com/hackclub/hcb/pull/9739) for this kind of reporting, but we don't yet support sub-events when generating statements of activity.

## Describe your changes

- Restructure the page so the various filters appear under the heading
- Show a checkbox to include sub-events when rendering a statement of activity for an `Event` (as opposed to an `Event::Group`)
- Add support for sub-events to `Event::StatementOfActivity`

image

### DIFF
--- a/app/views/events/_statement_of_activity_contents.html.erb
+++ b/app/views/events/_statement_of_activity_contents.html.erb
@@ -1,23 +1,40 @@
 <%# locals: (statement_of_activity:) %>
 
-<div class="flex justify-between items-center gap-4">
-  <h1 class="border-b-0 flex-grow">Statement of Activity</h1>
-
-  <%= form_with(method: :get, data: { controller: "form" }, class: "mb-0") do |form| %>
-    <div class="flex gap-2 items-center w-fit text-nowrap">
-      <%= form.label(:start, "From") %>
-      <%= form.date_field :start, value: statement_of_activity.start_date, data: { action: "change->form#submit" }, class: "mb-0" %>
-      <%= form.label(:end, "To") %>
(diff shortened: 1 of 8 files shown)


app/views/events/_statement_of_activity_contents.html.erb
```diff
@@ -3,7 +3,7 @@
 <div class="flex justify-between items-center gap-4 my-4">
   <h1 class="border-b-0 flex-grow m-0 p-0">Statement of Activity</h1>
 
-  <%= link_to({ format: :xlsx }, class: "btn") do %>
+  <%= link_to({ format: :xlsx, **params.permit(:include_descendants) }, class: "btn") do %>
```
davidcornu: @garyhtou we're not passing the date filters at the moment so by default you get all activity. I assume that was intentional?

davidcornu:
Rebase incoming

app/views/events/_statement_of_activity_contents.html.erb
davidcornu: upon discussion: should also apply date params

davidcornu:
TODO: Move XLSX button to be inline with the filters as they should also apply to that

garyhtou (approved):
Looks good- just wrap up the remaining comments

PR #11728, opened 2025-09-22
[Wise] Move recipient ID field into its own section

davidcornu:
## Summary of the problem

From @leowilkin:

CleanShot 2025-09-22 at 16 12 35@2x

## Describe your changes

- Split out the recipient ID field into its own section so it can be updated independently
- Remove the `disabled` attribute so it is always active

CleanShot 2025-09-22 at 16 04 36


YodaLightsabr (approved):
Looks good. Thanks for moving applicable stuff into Tailwind as well.

PR #11730, opened 2025-09-22
Handle invalid preauth errors

Luke-Oldenburg:
## Summary of the problem

```
ActiveRecord::RecordInvalid: Validation failed: Product url is invalid
app/controllers/card_grant/pre_authorizations_controller.rb:62 in CardGrant::PreAuthorizationsController#update
```

## Describe your changes

Handle the error and report it to the user.

### DIFF
--- a/app/controllers/card_grant/pre_authorizations_controller.rb
+++ b/app/controllers/card_grant/pre_authorizations_controller.rb
@@ -70,6 +70,9 @@ class CardGrant
 
       flash[:success] = "Pre-authorization submitted"
       redirect_to card_grant_pre_authorizations_path(@card_grant)
+    rescue ActiveRecord::RecordInvalid => e
+      flash[:error] = e.message
+      redirect_to card_grant_pre_authorizations_path(@card_grant)
     end
 
     def clear_screenshots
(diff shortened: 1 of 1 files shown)


app/controllers/card_grant/pre_authorizations_controller.rb
```diff
@@ -70,6 +70,9 @@ def update
 
       flash[:success] = "Pre-authorization submitted"
       redirect_to card_grant_pre_authorizations_path(@card_grant)
+    rescue ActiveRecord::RecordInvalid => e
```
sampoder: A better way to handle this would be using the boolean returned by `.update` and then `pre_authorization.errors.to_sentence`

app/controllers/card_grant/pre_authorizations_controller.rb
```diff
@@ -59,7 +59,10 @@ def update
         end
       end
 
-      @pre_authorization.update!(product_url: pre_authorization_params[:product_url])
+      unless @pre_authorization.update(product_url: pre_authorization_params[:product_url])
+        flash[:error] = @pre_authorization.errors.to_sentence
```
davidcornu: **suggestion**: `full_messages` will include the attribute names which makes these a bit more readable in the context of a `flash`

```suggestion
        flash[:error] = @pre_authorization.errors.full_messages.to_sentence
```

davidcornu (approved):

PR #11736, opened 2025-09-23
[v4] add help scout user beacon signature endpoint

Mohamad-Mortada:
## Summary of the problem
I added a Help scout beacon to HCB mobile (surprisingly very easy), however due to Help Scout secure mode I need a signature which is only possible via an API

## Describe your changes
Added a simple endpoint `/api/v4/user/beacon_config` to return smth like

`{
    signature: "xxxxxxxxxxxxx"
}
`


garyhtou (changes requested):
It makes sense that the mobile app needs it. However, this API is also used for other reasons. From a security perspective, I'd assume we don't want this available to other oauth apps? What are your thoughts (give a glance at the helpscout docs to get a sense of how important this signature is)

app/controllers/api/v4/users_controller.rb
```diff
@@ -43,6 +43,12 @@ def available_icons
         render json: icons.compact_blank
       end
 
+      def beacon_config
+        render json: {
+          signature: OpenSSL::HMAC.hexdigest("sha256", Credentials.fetch(:HELPSCOUT, :BEACON_SECRET_KEY) || "", current_user&.email || "")
```
garyhtou: We should fail loudly here if we're missing the secret. and I think `current_user` should always be present

```suggestion
          signature: OpenSSL::HMAC.hexdigest("sha256", Credentials.fetch(:HELPSCOUT, :BEACON_SECRET_KEY), current_user.email)
```

Mohamad-Mortada:
Help Scout won't even work without the signature. We could hide this behind a scope if you want to. This is scrapable from the web tho

app/controllers/api/v4/users_controller.rb
Mohamad-Mortada: Makes sense

garyhtou:
I'm thinking we only make this endpoint accessible if the oauth app is trusted

https://github.com/hackclub/hcb/blob/6ebe1901c214a4d80be727b4fea71d1223bbfbbe/db/schema.rb#L1463

garyhtou:
> Help Scout won't even work without the signature.

Right, we will need to expose this one way or another

> We could hide this behind a scope if you want to.

Scopes are kinda more used to control what a specific API token has access to, and less of controlling what an applicaton has access to. It would work, but feels like the wrong tool for the job.

> This is scrapable from the web tho

Yep, that is also true. I think having this accessible in the API is a higher risk. For example, let's say you oauth HCB to a random app that claims to "analyze your finances". It then is able to use the API to get your helpscout signature and send in a helpscout chat on your behalf. It is true that the secret could be scraped from web at the moment, but that would require malicious javascript to run on the HCB site (such as via an extention).

Mohamad-Mortada:
I created `require_trusted_oauth_app!` method in the API application controller to handle this then have it be checked for beacon_config.

garyhtou (approved):

PR #11737, opened 2025-09-23
Improvements to robotics 3d printer raffle

Luke-Oldenburg:
## Summary of the problem

- show banner to affiliation orgs
- show number of entries on button

### DIFF
--- a/app/views/events/perks/_robotics_printer_raffle.html.erb
+++ b/app/views/events/perks/_robotics_printer_raffle.html.erb
@@ -8,7 +8,7 @@
         3D Printer Raffle
       </h3>
       <% if event.robotics_team? || event.affiliations.robotics.any? %>
-        <span class="btn disabled bg-clip-padding">Entered</span>
+        <span class="btn disabled bg-clip-padding"><%= active_teenagers_count %> <%= "entry".pluralize(active_teenagers_count) %></span>
       <% else %>
         <div class="tooltipped tooltipped--w inline-block" aria-label="Must be tagged as a robotics team or have a robotics affiliation.">
           <span class="btn disabled bg-clip-padding">Ineligible</span>
--- a/app/views/events/perks/_robotics_printer_raffle.html.erb
+++ b/app/views/events/perks/_robotics_printer_raffle.html.erb
@@ -8,7 +8,7 @@
         3D Printer Raffle
       </h3>
       <% if event.robotics_team? || event.affiliations.robotics.any? %>
-        <span class="btn disabled bg-clip-padding"><%= active_teenagers_count %> <%= "ticket".pluralize(active_teenagers_count) %></span>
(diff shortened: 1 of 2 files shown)


app/views/events/perks/_robotics_printer_raffle.html.erb
```diff
@@ -8,7 +8,7 @@
         3D Printer Raffle
       </h3>
       <% if event.robotics_team? || event.affiliations.robotics.any? %>
-        <span class="btn disabled bg-clip-padding">Entered</span>
+        <span class="btn disabled bg-clip-padding"><%= active_teenagers_count %> <%= "ticket".pluralize(active_teenagers_count) %></span>
```
davidcornu: **suggestion**: You can simplify this a bit with `ActionView::Helpers::TextHelper#pluralize`


```suggestion
        <span class="btn disabled bg-clip-padding"><%= pluralize(active_teenagers_count, "ticket") %></span>
```

app/views/layouts/application.html.erb
```diff
@@ -119,7 +119,7 @@
       </div>
     <% end %>
 
-    <% if @event && !@event&.demo_mode? && @event&.robotics_team? && organizer_signed_in? && !@hide_promotional_banner && Date.current.month < 10 %>
+    <% if @event && !@event&.demo_mode? && (@event&.robotics_team? || @event&.affiliations&.robotics&.any?) && organizer_signed_in? && !@hide_promotional_banner && Date.current.month < 10 %>
```
davidcornu: **suggestion**: We're already checking whether `@event` exists so we might be able to simplify this a bit

```suggestion
    <% if @event && !@event.demo_mode? && (@event.robotics_team? || @event.affiliations.robotics.any?) && organizer_signed_in? && !@hide_promotional_banner && Date.current.month < 10 %>
```

Luke-Oldenburg: CI doesn't like that. We should probably remove the event check entirely and only use safe navigation here.

davidcornu: > CI doesn't like that.

Say more.

> We should probably remove the event check entirely and only use safe navigation here.

I generally view safe navigation as a code smell as it introduces more branches and leads to overly cautious code (e.g. anyone reading this would assume we can't trust anything being present)

Luke-Oldenburg: It has complained at me in the past for leaving out safe navigation operators when it could potentially be nil. I can't find the specific run.

Luke-Oldenburg: Actually I just found it https://github.com/hackclub/hcb/actions/runs/17956994304/job/51070721265

davidcornu: Oof can't say I agree with that rule but so be it 👍

davidcornu (approved):

PR #11740, opened 2025-09-23
Cap session duration at 2 weeks

Luke-Oldenburg:

### DIFF
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -74,11 +74,13 @@ class UserSession < ApplicationRecord
 
   LAST_SEEN_AT_COOLDOWN = 5.minutes
 
+  MAX_SESSION_DURATION = 2.weeks
+
   def update_session_timestamps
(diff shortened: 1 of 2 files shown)


app/models/user_session.rb
```diff
@@ -78,7 +78,7 @@ def update_session_timestamps
     return if last_seen_at&.after? LAST_SEEN_AT_COOLDOWN.ago # prevent spamming writes
 
     updates = { last_seen_at: Time.now }
-    updates[:expiration_at] = user.session_validity_preference.seconds.from_now unless impersonated?
+    updates[:expiration_at] = [created_at + 2.weeks, user.session_validity_preference.seconds.from_now].min unless impersonated?
```
davidcornu: **suggestion**: `2.weeks` should probably be a named constant

davidcornu (approved):
Took the liberty of adding a test case

PR #11742, opened 2025-09-23
Teen fee waiver

Luke-Oldenburg:
## Summary of the problem

We're adding a monthly fee waiver that all teen orgs can opt in to and will be automated.
Screenshot From 2025-09-23 17-01-39

## Describe your changes

image


app/views/events/perks/_fee_waiver.html.erb
```diff
@@ -0,0 +1,45 @@
+<%# locals: (event:, perks_available:, active_teenagers_count:) %>
+
+<% if Date.current.month < 11 %>
```
davidcornu: **question**: Can you explain why we don't want to show this in December?

app/views/events/perks/_fee_waiver.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  <div class="card pb0 mb3 promo-robotics" id="fee-waiver">
```
davidcornu, line 4: **concern**: What is `promo-robotics`? If we actually need those styles we might consider renaming the class to something more generic.

app/views/events/perks/_fee_waiver.html.erb
```diff
@@ -0,0 +1,45 @@
...
+
+      <% if fee_waiver_ineligible_reason %>
+        <div class="tooltipped tooltipped--w inline-block" aria-label="<%= fee_waiver_ineligible_reason %>">
+          <span class="btn disabled promo-premium-header bg-clip-padding">Ineligible</span>
+        </div>
+      <% else %>
+        <%= link_to (event.fee_waiver_eligible ? "Opt-out" : "Opt-in").to_s,
+          event_toggle_fee_waiver_eligible_path,
+          method: :post,
+          class: "btn promo-premium-header bg-clip-padding" %>
```
davidcornu: **suggestion**: Rather than make this a `link_to` with `method: :post` which relies on `rails-ujs` (and has caused weird issues in the past, hence https://github.com/hackclub/hcb/pull/11678/commits/4a8470ec1607c5a49eed2439d2392b4745b815b1), we can make this a `button_to` which will make it a regular form submission which Turbo can handle.

app/views/events/perks/_fee_waiver.html.erb
```diff
@@ -0,0 +1,45 @@
...
+    <h3 class="mb0 mt0">Raise some money this month with zero fees on HCB!</h3>
+    <p>
+      Are you a teen-led organization? We’d love to waive all of your HCB fees till <%= Date.current.end_of_month.strftime("%B %e, %Y") %>! Fee waiver status updates nightly.
+    </p>
+    <table class="mb2">
+      <tr class="<%= "card__darker" if active_teenagers_count >= 0 && active_teenagers_count <= 4 %>">
+        <td>0-4 active teenagers</td>
+        <td>7.0% fee</td>
+      </tr>
+      <tr class="<%= "card__darker" if active_teenagers_count >= 5 && active_teenagers_count <= 9 %>">
```
davidcornu: **suggestion**: We can simplify this a little bit with `Comparable#between` ([docs](https://docs.ruby-lang.org/en/3.4/Comparable.html#method-i-between-3F))


```suggestion
      <tr class="<%= "card__darker" if active_teenagers_count.between?(0, 4) %>">
        <td>0-4 active teenagers</td>
        <td>7.0% fee</td>
      </tr>
      <tr class="<%= "card__darker" if active_teenagers_count.between?(5, 9) %>">
```

app/jobs/event/apply_fee_waiver_job.rb
```diff
@@ -0,0 +1,30 @@
...
+          event.plan.update!(type: Event::Plan::Standard::FeeWaived)
+          event.update!(fee_waiver_applied: true)
+        end
+
+        if (!event.fee_waiver_eligible || active_teen_count < 5 || Date.current.month >= month_lock) && event.fee_waiver_applied
+          event.plan.update!(type: Event::Plan::Standard)
+          event.update!(fee_waiver_applied: false)
+        end
+      end
+    end
```
davidcornu, line 44: **concern**: I find this boolean logic really hard to follow so took a stab at what a refactor might look like

```suggestion
    MONTH_LOCK = 11 # 👈 Move this to a constant as it doesn't change

    def perform
      Event.find_each do |event|
        process_event(event) # 👈 Move this logic to a method so we can use return
      end
    end

    private

    def process_event(event)
      active_teen_count = event.users.active_teenager.count

      # 👇 Combine two of the branches as they share a lot of logic
      if active_teen_count >= 5 && event.fee_waiver_eligible && Date.current.month < MONTH_LOCK
        plan_type = # 👈 This is the only thing that changes between the branches so let's make that obvious
          if active_teen_count >= 10
            Event::Plan::Standard::FeeWaived
          else
            Event::Plan::Standard::ThreePointFive
          end

        # 👇 These two operations need to happen together so let's wrap them in a transaction
        ActiveRecord::Base.transaction do
          event.plan.update!(type: plan_type)
          event.update!(fee_waiver_applied: true)
        end

        return # 👈 Make it clear there's nothing left to do
      end

      # 👇 Move this conditional into a guard
      return unless event.fee_waiver_applied

      # 👇 These two operations need to happen together so let's wrap them in a transaction
      ActiveRecord::Base.transaction do
        event.plan.update!(type: Event::Plan::Standard)
        event.update!(fee_waiver_applied: false)
      end
    end

```

db/schema.rb
```diff
@@ -1015,6 +1015,8 @@
     t.boolean "financially_frozen", default: false, null: false
     t.boolean "donation_tiers_enabled", default: false, null: false
     t.bigint "parent_id"
+    t.boolean "fee_waiver_eligible", default: false, null: false
+    t.boolean "fee_waiver_applied", default: false, null: false
```
davidcornu, line 1023: **question**: Is it work making these timestamps? https://changelog.com/posts/you-might-as-well-timestamp-it

app/views/events/perks/_fee_waiver.html.erb
Luke-Oldenburg: We're extending it on a month-by-month basis starting with October.

app/views/events/perks/_fee_waiver.html.erb
Luke-Oldenburg: This still has the background image from the old fee waiver which was for robotics.

app/views/events/perks/_fee_waiver.html.erb
davidcornu: Oh 🤦 `< 11` excludes November and December.

> We're extending it on a month-by-month basis starting with October.

I find this a bit confusing as 

- `Date.current.month < 11` suggest that the promo will become available again in January
- Having a magic `11` in multiple places doesn't suggest this is something we're likely to change

Instead, might I suggest we define a constant somewhere like

```ruby
TEEN_LED_ORG_FEE_WAIVER_CUTOFF = Date.new(2025, 10, 31)
```

and then update the relevant call sides to something like

```ruby
Date.current <= TEEN_LED_ORG_FEE_WAIVER_CUTOFF
```

Luke-Oldenburg: Yea we've been doing this for a fair bit of things and I was planning to fix them all before we hit the new year and it accidentally rolled over.

db/schema.rb
Luke-Oldenburg: Do we need to with papertrail?

app/jobs/event/apply_fee_waiver_job.rb
Luke-Oldenburg: This is great!

db/schema.rb
davidcornu: Probably not. Always worth asking when adding new fields though.

app/jobs/event/apply_fee_waiver_job.rb
davidcornu, line 1: You can definitely remove all my comments here - the goal was to provide context behind my giant suggestion in https://github.com/hackclub/hcb/pull/11742/files#r2380152942

app/views/events/perks/_robotics_printer_raffle.html.erb
```diff
@@ -1,26 +1,28 @@
...
-      </h3>
-      <% if event.robotics_team? || event.affiliations.robotics.any? %>
-        <span class="btn disabled bg-clip-padding"><%= pluralize(active_teenagers_count, "ticket") %></span>
-      <% else %>
-        <div class="tooltipped tooltipped--w inline-block" aria-label="Must be tagged as a robotics team or have a robotics affiliation.">
-          <span class="btn disabled bg-clip-padding">Ineligible</span>
-        </div>
-      <% end %>
-    </div>
+<% if Date.current < Date.new(2025, 12, 1) %>
```
davidcornu, line 3: **concern**: This is better than just checking that the month is less than 11 but we're still duplicating this date in a bunch of places. I assume you'll constantize this in when you do https://github.com/hackclub/hcb/pull/11742#discussion_r2380268609.

davidcornu: e.g. we could add `app/model/promotions.rb` with something like

```ruby
module Promotions
  CUTOFFS = {
    robotics_printer_raffle: Date.new(2025, 12, 1)
  }.freeze
end
```

and reference this as `Promotions::CUTOFFS.fetch(:robotics_printer_raffle)`

Luke-Oldenburg: Will do in a subsequent PR.

davidcornu (approved):

app/jobs/event/apply_fee_waiver_job.rb
```diff
@@ -0,0 +1,47 @@
...
+      if active_teen_count >= 5 && event.fee_waiver_eligible && Date.current < DATE_LOCK
+        plan_type =
+          if active_teen_count >= 10
+            Event::Plan::Standard::FeeWaived
+          else
+            Event::Plan::Standard::ThreePointFive
+          end
+
+        ActiveRecord::Base.transaction do
+          event.plan.update!(type: plan_type)
```
garyhtou: There's a slightly more complicated process for updating an event's plan. Please see how it's done when updated via the UI.

https://github.com/hackclub/hcb/blob/ca3cecee49ca389b1d652e7613c204804bfaecc2/app/controllers/events_controller.rb#L363-L365

garyhtou (changes requested):
Just once piece of feedback, and then this can be merged

app/jobs/event/apply_fee_waiver_job.rb
```diff
@@ -0,0 +1,49 @@
...
+          else
+            Event::Plan::Standard::ThreePointFive
+          end
+
+        unless event.plan.type == plan_type
+          ActiveRecord::Base.transaction do
+            event.plan.mark_inactive!(plan_type)
+            event.update!(fee_waiver_applied: true)
+          end
+        end
```
Luke-Oldenburg: @garyhtou does this seem correct to you? I wanted to prevent updating the plan if it remained the same.

app/jobs/event/apply_fee_waiver_job.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      if active_teen_count >= 5 && event.fee_waiver_eligible && Date.current < DATE_LOCK
+        plan_type =
+          if active_teen_count >= 10
+            Event::Plan::Standard::FeeWaived
+          else
+            Event::Plan::Standard::ThreePointFive
+          end
+
+        unless event.plan.type == plan_type
```
garyhtou:
```suggestion
        unless event.plan.instance_of?(plan_type)
```

`event.plan.type` will return a string such as
```ruby
"Event::Plan::FeeWaived"
```
However, `plan_type` will be a class such as
```ruby
Event::Plan::FeeWaived
```

This means `"Event::Plan::FeeWaived" != Event::Plan::FeeWaived`. So, we need to check to compare either class to class or string to string. I suggested using `instance_of?`, which will compare class to class, but any of the following also works:

```ruby
event.plan.type == plan_type.name # compares strings

event.plan.class == plan_type # compares classes, and is the same as instance_of?
```

It's common for `instance_of?` to be confused with `is_a?`. `is_a?(Event::Plan::FeeWaived)` allows for any of `Event::Plan::FeeWaived`'s subclasses, which we do NOT want in this case. We want to check for specifically that class (and none of its subclasses).

app/jobs/event/apply_fee_waiver_job.rb
garyhtou: Just left a comment: https://github.com/hackclub/hcb/pull/11742/files#r2400821036

app/jobs/event/apply_fee_waiver_job.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def process_event(event)
+      active_teen_count = event.users.active_teenager.count
+
+      if active_teen_count >= 5 && event.fee_waiver_eligible && Date.current < DATE_LOCK
+        plan_type =
+          if active_teen_count >= 10
+            Event::Plan::Standard::FeeWaived
+          else
+            Event::Plan::Standard::ThreePointFive
```
garyhtou:
```suggestion
            Event::Plan::FeeWaived
          else
            Event::Plan::ThreePointFive
```
The correct names for these classes don't have Standard in them.

Image

app/jobs/event/apply_fee_waiver_job.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def perform
+      Event.find_each do |event|
+        process_event(event)
+      end
+    end
+
+    private
+
+    def process_event(event)
```
garyhtou, line 17: This is okay, we need to be mindful of Standard organizations with a fee waiver being later changed to a "permanent" custom plan (e.g. HackClubHQ). This job would kick them back to Standard without the fee waiver

garyhtou (approved):

PR #11751, opened 2025-09-24
[DX] Make the test suite work without credentials or external calls

davidcornu:
## Summary of the problem

**Prior art**
- https://github.com/hackclub/hcb/issues/10755
- https://github.com/hackclub/hcb/issues/8945
- https://github.com/hackclub/hcb/pull/10829
- https://github.com/hackclub/hcb/pull/11061
- #11230 (which added WebMock)

Because our test suite makes external API calls and relies on certain credentials being available,
- It does not run successfully for third-party (including Dependabot) PRs as we don't provision all the environment variables
- It is a pain to get running locally as you need credentials for a bunch of services
- It does not run without a network connection

## Describe your changes

I commented out my local `.env` file, configured WebMock to block all requests, and chipped away at all the errors until the whole thing ran successfully. This ended up involving:

1. Adding fallback values for `TEST_URL_HOST` (this seems like an odd thing to require in the first place)
2. Generating a fake Stripe API key when one isn't configured (only in the test env)
3. Generating an ephemeral encryption key for `Lockbox` if one isn't configured (only in the test env)
4. Generating ephemeral encryption keys for `ActoveRecord::Encryption` if they aren't configured (only in the test env)
5. Added a cached version of the `EuCentralBank` values for use in the test env
6. Added stubs and/or performed additional refactors to impacted tests

When (2), (3), and (4) take effect we log the ephemeral values to STDERR:

CleanShot 2025-09-24 at 16 40 37

I've also removed `WebMock.allow_net_connect!` so any newly-introduced network calls will fail.

## Notes

- Some of the changes in this PR were extracted into https://github.com/hackclub/hcb/pull/11750
- This was purposefully opened from a fork to see if we can get a clean CI run from an outside contribution

### DIFF
--- a/spec/models/donation_spec.rb
+++ b/spec/models/donation_spec.rb
@@ -4,31 +4,10 @@ require "rails_helper"
 
 RSpec.describe Donation, type: :model do
   include ActiveJob::TestHelper
+  include DonationSupport
 
   before do
-    expect(StripeService::Customer).to(
-      receive(:create)
-        .and_return(
-          Stripe::Customer.construct_from(
-            id: "cus_#{SecureRandom.alphanumeric(10)}"
-          )
-        )
-        .at_least(:once)
-    )
-
-    expect(StripeService::PaymentIntent).to(
-      receive(:create)
-        .and_return(
-          Stripe::PaymentIntent.construct_from(
-            id: "pi_#{SecureRandom.alphanumeric(10)}",
-            amount: 12_34,
-            amount_received: 0,
-            status: "processing",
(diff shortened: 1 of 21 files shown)


spec/models/donation_spec.rb
```diff
@@ -2,8 +2,13 @@
 
 require "rails_helper"
 
-RSpec.describe Donation, type: :modal do
+RSpec.describe Donation, type: :model do
```
garyhtou, line 5: good catch

garyhtou (approved):
this is very nice to have!

PR #11752, opened 2025-09-24
Explain the lack of pagination for Column reports

davidcornu:
## Summary of the problem

https://hackclub.slack.com/archives/C047Y01MHJQ/p1758733559361789

This code is a bit scary to the uninitiated as on the surface it looks like we could be skipping data by not checking the `has_more` property of the response object.

## Describe your changes

- Add a guard to make sure we never exceed 100 days
- Add a comment explaining why we don't need to paginate
- Remove dead code


garyhtou (approved):
Thank you!

PR #11753, opened 2025-09-26
Log the response body for OpenAI errors

davidcornu:
## Summary of the problem

We're getting spikes of these errors https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/1953 but we don't have enough information to figure out what's going on.

## Describe your changes

Add some error handling logic which re-raises the same exception but appends the response body to the message.

### DIFF
--- a/app/models/card_grant/pre_authorization.rb
+++ b/app/models/card_grant/pre_authorization.rb
@@ -183,6 +183,13 @@ class CardGrant
       end
 
       broadcast_refresh_to self
+    rescue Faraday::Error => e
+      # Modify the original exception to append the response body to the message
+      # so these are easier to debug
+      raise(e.exception(<<~MSG))
+        #{e.message}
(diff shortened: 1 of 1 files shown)


app/models/card_grant/pre_authorization.rb
```diff
@@ -183,6 +183,13 @@ def analyze!
       end
 
       broadcast_refresh_to self
+    rescue Faraday::Error => e
+      # Modify the original exception to append the response body to the message
+      # so these are easier to debug
+      raise(e.exception(<<~MSG))
```
davidcornu, line 189: `Exception#exception` creates a clone with a different message: https://docs.ruby-lang.org/en/3.4/Exception.html#method-i-exception

Image

YodaLightsabr (approved):

PR #11757, opened 2025-09-26
[Accounting] Disbursement transaction categories

davidcornu:
## Summary of the problem

We want to be able to tag the two transactions associated with a disbursement when the disbursement is created (by admins or via the API) or reviewed.

## Describe your changes

1. Adds `source_transaction_category_id` and `destination_transaction_category_id` columns to the `disbursements` table
2. Adds fields for (1) on the new disbursement form that are visible to admins
    CleanShot 2025-09-29 at 15 46 23
3. Updates `PendingTransactionEngine::CanonicalPendingTransactionService::ImportSingle::IncomingDisbursement` and `PendingTransactionEngine::CanonicalPendingTransactionService::ImportSingle::OutgoingDisbursement` to apply (1) to the canonical pending transactions they are responsible for creating
    - This has the benefit of handling both immediate and scheduled disbursements as in the latter case we end up calling `...::OutgoingDisbursement` to process them https://github.com/hackclub/hcb/blob/9d308531227888e63cb5afe4ffbe405bc6b1b225/app/jobs/disbursement/daily_job.rb#L13-L14
4. Adds a link to the HCB code page (as it's a convenient place to edit transaction categories)
    1. On the disbursement page
        CleanShot 2025-09-29 at 15 52 07
    2. On the disbursement processing page
        CleanShot 2025-09-29 at 15 54 28
5. Updates the disbursement processing page to show related transactions so their categories can be updated (using the same table as the HCB code page)
    Immediate | Scheduled
    ---- | ----
    CleanShot 2025-09-29 at 15 57 57 | CleanShot 2025-09-29 at 15 58 14
6. Updates the admin disbursements index to always link to the processing page (as it now has a use case beyond approving/rejecting)
    CleanShot 2025-09-29 at 16 00 14

### DIFF
--- a/app/models/disbursement.rb
+++ b/app/models/disbursement.rb
@@ -226,6 +226,10 @@ class Disbursement < ApplicationRecord
     @canonical_pending_transactions ||= ::CanonicalPendingTransaction.where(hcb_code:)
   end
 
+  def transactions_helper
+    @transactions_helper ||= Disbursement::TransactionsHelper.new(self)
+  end
+
   def processed?
     in_transit? || deposited?
   end
--- a/app/models/disbursement.rb
+++ b/app/models/disbursement.rb
@@ -4,41 +4,47 @@
 #
 # Table name: disbursements
 #
-#  id                       :bigint           not null, primary key
-#  aasm_state               :string
-#  amount                   :integer
-#  deposited_at             :datetime
-#  errored_at               :datetime
-#  in_transit_at            :datetime
-#  name                     :string
-#  pending_at               :datetime
-#  rejected_at              :datetime
-#  scheduled_on             :date
-#  should_charge_fee        :boolean          default(FALSE)
-#  created_at               :datetime         not null
-#  updated_at               :datetime         not null
-#  destination_subledger_id :bigint
-#  event_id                 :bigint
-#  fulfilled_by_id          :bigint
-#  requested_by_id          :bigint
-#  source_event_id          :bigint
-#  source_subledger_id      :bigint
+#  id                                  :bigint           not null, primary key
+#  aasm_state                          :string
+#  amount                              :integer
+#  deposited_at                        :datetime
+#  errored_at                          :datetime
+#  in_transit_at                       :datetime
+#  name                                :string
+#  pending_at                          :datetime
+#  rejected_at                         :datetime
+#  scheduled_on                        :date
+#  should_charge_fee                   :boolean          default(FALSE)
+#  created_at                          :datetime         not null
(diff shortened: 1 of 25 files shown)


app/models/disbursement.rb
```diff
@@ -217,6 +226,10 @@ def canonical_pending_transactions
     @canonical_pending_transactions ||= ::CanonicalPendingTransaction.where(hcb_code:)
   end
 
+  def transactions_helper
```
garyhtou, line 229: nice!

app/services/disbursement_service/create.rb
```diff
@@ -39,35 +53,37 @@ def run
...
-      # 2. Canonize the newly added raw pending transactions
-      o_cpt = ::PendingTransactionEngine::CanonicalPendingTransactionService::ImportSingle::OutgoingDisbursement.new(raw_pending_outgoing_disbursement_transaction: rpodt).run
-      # 3. Map to event
-      ::PendingEventMappingEngine::Map::Single::OutgoingDisbursement.new(canonical_pending_transaction: o_cpt).run
-      # 4. Front if required
-      o_cpt.update(fronted: @fronted)
-
-      if disbursement.scheduled_on.nil?
-        # We only want to import Incoming Disbursements AFTER the scheduled date
+      ActiveRecord::Base.transaction do
```
garyhtou, line 56: Good call

app/views/admin/disbursement_process.html.erb
```diff
@@ -1,86 +1,149 @@
...
+      locals: {
+        canonical_transactions: @disbursement.transactions_helper.settled_source,
+        canonical_pending_transactions: @disbursement.transactions_helper.pending_source,
+      }
+    ) %>
+
+<h4>Destination transactions</h4>
+
+<% if @disbursement.transactions_helper.settled_destination.empty? && @disbursement.transactions_helper.pending_destination.empty? %>
+  <p>Destination transactions for this disbursement have not been created yet. Once created, they will have the following properties:</p>
```
garyhtou, line 81: It's good to have this disclaimer

app/views/disbursements/_form.html.erb
```diff
@@ -37,6 +37,28 @@
     <span class="muted">This is to help HCB keep record of our transactions.</span>
   </div>
 
+  <% admin_tool do %>
+      <% category_select = ->(field) {
```
garyhtou, line 41: nice!

garyhtou (approved):
Works well! Good to merge. thanks!

PR #11759, opened 2025-09-26
Use the correct currency in the expenses email

ImShyMike:
## Summary of the problem
The expenses aproved email always uses USD regardless of the currency the report is set to.

image
(should be EUR, not USD)

## Describe your changes
Added the unit parameter to the `render_money` call (taken from the expense's currency)

(im _pretty_ sure this works but i have no clue how i would even test this :sob: )

### DIFF
--- a/app/views/reimbursement_mailer/expenses_approved.html.erb
+++ b/app/views/reimbursement_mailer/expenses_approved.html.erb
@@ -8,7 +8,7 @@
 
 <ul>
   <% @expenses.each do |expense| %>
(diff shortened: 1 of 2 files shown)


davidcornu:
Thanks for the fix @ImShyMike. Took the liberty of tweaking the implementation and adding a test.

davidcornu (approved):

garyhtou:
Thank you @ImShyMike and @davidcornu!

PR #11763, opened 2025-09-29
[User Payout] Move presence validation from frontend to backend

garyhtou:
Hot fix to get the form working again

## Summary of the problem

image

The issue is that the Wise transfer form has required field which are irrlevant when you aren't using wise for your payout. This causes the form to be unsubmittable since JS tried to validate that the required fields are filled.

## Describe your changes

Move presence validation from frontend to backend

### DIFF
--- a/app/models/concerns/has_wise_recipient.rb
+++ b/app/models/concerns/has_wise_recipient.rb
@@ -8,7 +8,7 @@ module HasWiseRecipient
     has_country_enum(field: :recipient_country)
 
     validate do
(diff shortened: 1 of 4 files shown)


garyhtou:
This likely is NOT a complete fix since a similar fix probably should also be applied to the currency specific fields too
https://github.com/hackclub/hcb/blob/233e37864dc566250784bc6c3eae27b54186e31b/app/views/users/_payout_form.html.erb#L107

PR #11766, opened 2025-09-29
Refactor `first_login` email to `onboarded`

### LINKED ISSUE #11708: [User] First login email is sent without the user's actual name
The [first login email](https://github.com/hackclub/hcb/blob/f8f33a48f0e4a7c92f878f7d018294e022800186/app/views/user_session_mailer/first_login.html.erb#L5) is sent before the user gets a chance to fill in their name. In an ideal world, we have their name before this email is sent.
Image

The `gary+cookie-leak` is just pulled from the [handle](https://github.com/hackclub/hcb/blob/1841dea800d0f36bb491c569e657a659fe5500ee/app/models/user.rb#L492-L494) of their email.

The new solution we come up with should still send this email (with the email handle as name) if they don't complete the user sign up flow

sampoder:
Closes https://github.com/hackclub/hcb/issues/11708 by factoring this email to only go out when a user is "onboarded" which we define as the user having a full name.


app/models/user.rb
```diff
@@ -160,6 +160,8 @@ class User < ApplicationRecord
 
   after_update :update_stripe_cardholder, if: -> { phone_number_previously_changed? || email_previously_changed? }
 
+  after_update_commit :send_onboarded_email, if: -> { full_name_previously_changed? && full_name_before_last_save.blank? && full_name.present? }
```
garyhtou, line 163: This is a bit brittle in the case that we change what it means to be onboarded.

Can we abstract some of this logic out into a method that plays well with our existing `onboarding?` method?

https://github.com/hackclub/hcb/blob/ab1d1c494062e1d5fa9a636de1b179708489a880/app/models/user.rb#L315-L318

app/models/user_session.rb
```diff
@@ -55,9 +55,7 @@ class UserSession < ApplicationRecord
   scope :recently_expired_within, ->(date) { expired.where("expiration_at >= ?", date) }
 
   after_create_commit do
-    if user.user_sessions.size == 1
-      UserSessionMailer.first_login(user:).deliver_later
-    elsif fingerprint.present? && user.user_sessions.excluding(self).where(fingerprint:).none?
+    if user.user_sessions.size > 1 && fingerprint.present? && user.user_sessions.excluding(self).where(fingerprint:).none?
```
garyhtou, line 58: This comment is not directly related to this PR. However, maybe we should explicitly wrap this in a `!impersonated?`.

This isn't completely necessary since impersonated sessions do not have fingerprints. However, I do still believe it's good to have this condition explicitly documented in the code.

garyhtou (commented):
Looks good! Almost there. Do you have a plan on how to handle users who don't finish onboarding? See https://github.com/hackclub/hcb/issues/11708#:~:text=The%20new%20solution%20we%20come%20up%20with%20should%20still%20send%20this%20email%20(with%20the%20email%20handle%20as%20name)%20if%20they%20don%27t%20complete%20the%20user%20sign%20up%20flow

sampoder:
>  Do you have a plan on how to handle users who don't finish onboarding?

No, honestly, it doesn't make sense to me why we'd email them this email. We can have a job that does that every 10 minutes but just feels weird to send someone an email they don't seem to want.

PR #11767, opened 2025-09-29
Default to `financially_frozen` and remove that freeze after a contract is signed

sampoder:
Closes https://github.com/hackclub/hcb/issues/11760, see Slack thread for context behind decision: https://hackclub.slack.com/archives/C026RKHLPNJ/p1758917526179979.

This is a common frustration:

image (44)

And an equally pressing issue is how an event can spend money without a contract?

Visual change:

Screenshot 2025-10-01 at 11 00 33 PM

### DIFF
--- a/app/controllers/docuseal_controller.rb
+++ b/app/controllers/docuseal_controller.rb
@@ -34,12 +34,6 @@ class DocusealController < ActionController::Base
         document.save!
         contract.update(document:)
         contract.mark_signed!
-
-        # Unfreeze the event if this is the first signed contract
-        event = contract.organizer_position_invite.event
-        if event.organizer_position_contracts.signed.count == 1
-          event.update!(financially_frozen: false)
-        end
       elsif params[:event_type] == "form.declined"
         contract.mark_voided!
       elsif cosigner.present? && cosigner["status"] != "completed"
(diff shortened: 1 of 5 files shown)


app/controllers/docuseal_controller.rb
```diff
@@ -34,6 +34,12 @@ def webhook
         document.save!
         contract.update(document:)
         contract.mark_signed!
+
+        # Unfreeze the event if this is the first signed contract
+        event = contract.organizer_position_invite.event
+        if event.organizer_position_contracts.signed.count == 1
+          event.update!(financially_frozen: false)
+        end
```
garyhtou: Should this be a callback of `mark_signed!`? It feels a tad bit cleaner that way

garyhtou (changes requested):
Thanks for this. Can you please add a PR description that provides further context beyond what the linked issue provides?

There are some nuances and context behind this decision that we should document. And a screenshot of the visual changes would be nice too!

See this [Slack message](https://hackclub.slack.com/archives/C047Y01MHJQ/p1758034135901299) for context around why we'd like more detailed PR descriptions

app/controllers/docuseal_controller.rb
sampoder: nice - yeah - moved!

sampoder:
Noted & updated re: PR description; sorry - rushed this together when I had time.

PR #11770, opened 2025-09-30
[Wise transfer] Temporarily disable for maintenance

garyhtou:
## Summary of the problem

At the moment, operations is overwhelmed with the number of Wise transfers, and there are a few serious bugs that are making it even harder for them. We've [decided to temporarily disable Wise transfers](https://hackclub.slack.com/archives/C026RKHLPNJ/p1759263275754389?thread_ts=1759261355.207749&cid=C026RKHLPNJ) until we get these bugs fixed up.

This PR disables Wise transfers for both reimbursements (as a payout method) and also for event transfers.

## Describe your changes

I've decided to add more tooling around disabling payout methods. It's typically a pain to disable a payout method since there are so many locations you need to add checks and messaging. Now, you can simply update the `User::PayoutMethod::UNSUPPORTED_METHODS` constant to declaratively disable a payout method.

`User::PayoutMethod::UNSUPPORTED_METHODS` does not affect the transfer options that events have access to. The process for disabling that is still fairly manual, although it's [not as bad](https://github.com/hackclub/hcb/commit/275b46906a7587633081743084e4211002dde98e) as reimbursement payout methods.

### DIFF
--- a/app/models/reimbursement/report.rb
+++ b/app/models/reimbursement/report.rb
@@ -109,8 +109,8 @@ module Reimbursement
         transitions from: [:draft, :reimbursement_requested], to: :submitted do
           guard do
             user.payout_method.present? && event && !exceeds_maximum_amount? && !below_minimum_amount? &&
-              expenses.any? && !missing_receipts? && user.payout_method.class != User::PayoutMethod::PaypalTransfer &&
-              !event.financially_frozen? && expenses.none? { |e| e.amount.zero? } && !mismatched_currency?
+              expenses.any? && !missing_receipts? && !event.financially_frozen? && expenses.none? { |e| e.amount.zero? } &&
+              !mismatched_currency? && payout_method_allowed?
           end
         end
         after do
@@ -374,6 +374,10 @@ module Reimbursement
       )
 
       mark_reimbursed!
+    end
+
+    def payout_method_allowed?
+      user.payout_method.present? && !user.payout_method.unsupported?
     end
 
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -525,8 +525,15 @@ class User < ApplicationRecord
   end
 
   def valid_payout_method
-    unless payout_method_type.nil? || payout_method.is_a?(User::PayoutMethod::Check) || payout_method.is_a?(User::PayoutMethod::AchTransfer) || payout_method.is_a?(User::PayoutMethod::PaypalTransfer) || payout_method.is_a?(User::PayoutMethod::Wire) || payout_method.is_a?(User::PayoutMethod::WiseTransfer)
-      errors.add(:payout_method, "is an invalid method, must be check, PayPal, wire, Wise transfer, or ACH transfer")
+    if payout_method_type.present? && User::PayoutMethod::SUPPORTED_METHODS.none? { |method| payout_method.is_a?(method) }
+      # I'm using `try` here in the slim chance that `payout_method` is some
+      # random model and doesn't include `User::PayoutMethod::Shared`.
+      if payout_method.try(:unsupported?)
+        reason = payout_method.unsupported_details[:reason]
+        errors.add(:payout_method, "is invalid. #{reason} Please choose another option.")
+      else
+        errors.add(:payout_method, "is invalid. Please choose another option.")
+      end
     end
   end
 
--- a/app/models/user/payout_method/shared.rb
+++ b/app/models/user/payout_method/shared.rb
@@ -21,12 +21,13 @@ class User
         }
 
         validate do
-          if User::PayoutMethod::DISABLED_METHODS.include?(self.class)
-            errors.add(:base, "#{title_kind} is currently unavailable. Please choose another method.")
+          if User::PayoutMethod::UNSUPPORTED_METHODS.include?(self.class)
+            errors.add(:base, "#{self.unsupported_details[:reason]} Please choose another method.")
           end
         end
 
         delegate :unsupported?, to: :class
+        delegate :unsupported_details, to: :class
       end
 
       class_methods do
--- a/app/models/user/payout_method/shared.rb
+++ b/app/models/user/payout_method/shared.rb
@@ -32,7 +32,7 @@ class User
 
       class_methods do
         def unsupported?
-          self.in?(User::PayoutMethod::UNSUPPORTED_METHODS.keys)
+          User::PayoutMethod::UNSUPPORTED_METHODS.key?(self)
         end
 
         def unsupported_details
--- a/app/models/user/payout_method/shared.rb
+++ b/app/models/user/payout_method/shared.rb
@@ -19,6 +19,24 @@ class User
           Reimbursement::PayoutHolding.where(report: user.reimbursement_reports).failed.each(&:mark_settled!)
           Employee::Payment.where(employee: user.jobs).failed.each(&:mark_approved!)
         }
+
+        validate do
+          if User::PayoutMethod::DISABLED_METHODS.include?(self.class)
+            errors.add(:base, "#{title_kind} is currently unavailable. Please choose another method.")
+          end
+        end
+
+        delegate :unsupported?, to: :class
+      end
+
+      class_methods do
+        def unsupported?
+          self.in?(User::PayoutMethod::UNSUPPORTED_METHODS.keys)
+        end
+
+        def unsupported_details
+          User::PayoutMethod::UNSUPPORTED_METHODS[self]
+        end
       end
     end
   end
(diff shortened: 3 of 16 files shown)


app/models/user/payout_method/wise_transfer.rb
```diff
@@ -21,8 +21,9 @@
 class User
   module PayoutMethod
     class WiseTransfer < ApplicationRecord
+      include Shared
```
garyhtou, line 24: I expected all payout method models to `include Shared`. However, `User::PayoutMethod::Wire` and `User::PayoutMethod::WiseTransfer` were missing it.

I added it, but would like a sanity check to make sure there wasn't a good reason for leaving it out.

app/models/user.rb
```diff
@@ -525,8 +525,15 @@ def on_phone_number_update
   end
 
   def valid_payout_method
-    unless payout_method_type.nil? || payout_method.is_a?(User::PayoutMethod::Check) || payout_method.is_a?(User::PayoutMethod::AchTransfer) || payout_method.is_a?(User::PayoutMethod::PaypalTransfer) || payout_method.is_a?(User::PayoutMethod::Wire) || payout_method.is_a?(User::PayoutMethod::WiseTransfer)
-      errors.add(:payout_method, "is an invalid method, must be check, PayPal, wire, Wise transfer, or ACH transfer")
+    if payout_method_type.present? && User::PayoutMethod::SUPPORTED_METHODS.none? { |method| payout_method.is_a?(method) }
```
garyhtou, line 528: Realistically, I don't think we need `is_a?` here. A simple:
```ruby
User::PayoutMethod::SUPPORTED_METHODS.exclude?(payout_method.class)
```
probably should be fine. This is because we aren't dealing with inheritance for payout method classes.

I kept it, but if we agree that `is_a?` is not needed, then let's get rid of it.

spec/controllers/users_controller_spec.rb
```diff
@@ -114,5 +114,37 @@
...
+            payout_method_type: "User::PayoutMethod::PaypalTransfer",
+            payout_method_attributes: {
+              recipient_email: "gary@hackclub.com"
+            }
+          }
+        }
+      )
+
+      expect(response).to have_http_status(:unprocessable_entity)
+      expect(response.body).to include("Payout method is invalid. Due to integration issues, transfers via PayPal are currently unavailable. Please choose another option.")
```
garyhtou: @davidcornu, I'd love some help here. Because of [this validation](https://github.com/hackclub/hcb/blob/a622bdd1db791ae03ab7a545260a4fe034ddc0ac/app/models/user.rb#L527-L538), I would expect to see the following string flashed:
```
Payout method is invalid. Due to integration issues, transfers via PayPal are currently unavailable. Please choose another option.
```

However, upon running the test, I could only locate
```
Due to integration issues, transfers via PayPal are currently unavailable.
```
within the response HTML.

app/models/reimbursement/report.rb
```diff
@@ -109,8 +109,8 @@ class Report < ApplicationRecord
         transitions from: [:draft, :reimbursement_requested], to: :submitted do
           guard do
             user.payout_method.present? && event && !exceeds_maximum_amount? && !below_minimum_amount? &&
-              expenses.any? && !missing_receipts? && user.payout_method.class != User::PayoutMethod::PaypalTransfer &&
-              !event.financially_frozen? && expenses.none? { |e| e.amount.zero? } && !mismatched_currency?
+              expenses.any? && !missing_receipts? && !event.financially_frozen? && expenses.none? { |e| e.amount.zero? } &&
+              !mismatched_currency? && payout_method_allowed?
```
davidcornu, line 113: - **concern**: This `guard` block is really getting out of hand. Might I suggest we at least split the conditions into separate lines
- **note**: I couldn't make heads or tails of this diff so here's my manual audit: 
	Predicate | Change
	---- | ----
	`user.payout_method.present?` | Kept
	`event` | Kept
	`!exceeds_maximum_amount?` | Kept
	`!below_minimum_amount?` | Kept
	`expenses.any?` | Kept
	`!missing_receipts?` | Kept
	`user.payout_method.class != User::PayoutMethod::PaypalTransfer` | Replaced with `payment_method_allowed?`
	`!event.financially_frozen?` | Kept
	`expenses.none? { \|e\| e.amount.zero? }` | Kept
	`!mismatched_currency?` | Kept

app/models/user/payout_method/shared.rb
```diff
@@ -19,6 +19,25 @@ module Shared
...
+          end
+        end
+
+        delegate :unsupported?, to: :class
+        delegate :unsupported_details, to: :class
+      end
+
+      class_methods do
+        def unsupported?
+          self.in?(User::PayoutMethod::UNSUPPORTED_METHODS.keys)
```
davidcornu: **nit**: Save ourselves allocating an array for `keys`
```suggestion
          User::PayoutMethod::UNSUPPORTED_METHODS.key?(self)
```

app/models/user.rb
```diff
@@ -525,8 +525,15 @@ def on_phone_number_update
...
+      # I'm using `try` here in the slim chance that `payout_method` is some
+      # random model and doesn't include `User::PayoutMethod::Shared`.
```
davidcornu, line 530: **concern**: Rather than add this workaround let's just make sure every payout method class includes `User::PayoutMethod::Shared` and fail loudly if it does not. One option might be to make `Shared` a superclass instead (with `self.abstract_class = true` so ActiveRecord doesn't think it has a table) and reject everything that isn't a subclass when assigning a payout method.

app/views/reimbursement/reports/_actions.html.erb
```diff
@@ -86,7 +86,7 @@
       <% end %>
 
       <div style="flex-grow: 1;" class="flex justify-end">
-        <% if !user.payout_method.present? || report.payout_holding&.failed? || (user.payout_method.present? && user.payout_method.class == User::PayoutMethod::PaypalTransfer) || report.mismatched_currency? %>
+        <% if !user.payout_method.present? || report.payout_holding&.failed? || (user.payout_method.present? && user.payout_method.unsupported?) || report.mismatched_currency? %>
```
davidcornu, line 89: **concern**: Can we make this a method on `report` and crucially name it? This many predicates are hard to reason about 😅

app/views/users/payout_form/_option.html.erb
```diff
@@ -0,0 +1,16 @@
...
+
+<%= form.radio_button :payout_method_type, model.name, disabled: model.unsupported? %>
+<%= form.label :payout_method_type, value: model.name do %>
+  <%= inline_icon icon, size: 28 %>
+  <strong><%= title %></strong>
+
+  <%= yield %><%# in case you wanna add more flare to a specific option %>
+
+  <% if model.unsupported? %>
+    <div class="badge bg-warning m1 mt0" style="width: fit-content;"><%= model.unsupported_details[:status_badge] %></div>
```
davidcornu: **concern**: We don't need inline styles here (and should avoid adding more)

```suggestion
    <div class="badge bg-warning m1 mt0 w-fit"><%= model.unsupported_details[:status_badge] %></div>
```

https://tailwindcss.com/docs/width

spec/controllers/users_controller_spec.rb
davidcornu: See [`0702e11` (#11770)](https://github.com/hackclub/hcb/pull/11770/commits/0702e112fed38ae5ee856ad441c52ba2e61a5d95)

app/models/reimbursement/report.rb
garyhtou: Yeah, this definitely is getting out of hand. I suggest we tackle that in another PR?

app/models/user/payout_method/shared.rb
garyhtou: Great suggestion! [`be8a4b7` (#11770)](https://github.com/hackclub/hcb/pull/11770/commits/be8a4b7cab223d4c224b2849ccf8d71530d0916a)

spec/controllers/users_controller_spec.rb
garyhtou: Thank you!

app/models/user.rb
garyhtou: I agree, will tackle in another PR

garyhtou: Although with how [this is written](https://github.com/hackclub/hcb/blob/a622bdd1db791ae03ab7a545260a4fe034ddc0ac/app/models/user.rb#L347-L351), maybe there's a chance it could literally be a random model/constant 🤔. Might be concerning; will have to investigate further.

app/views/reimbursement/reports/_actions.html.erb
garyhtou: Yeah, will create a ticket for this

PR #11773, opened 2025-10-01
[Admin][Referral] Make users unique

garyhtou:
## Summary of the problem

On the admin referrals page, we have a stat that shows how many new users each referral is responsible for. We use this to see which of our initiatives are working and which ones need improvement. This isn't used for goal reporting purposes.

Anyways, I found that it joins attributions with users but didn't `uniq` the users. This is needed because a single user may have multiple attributions per referral program.

## Describe your changes

add 5 characters lol


Luke-Oldenburg (approved):
lgtm!

PR #11781, opened 2025-10-02
Conditionally render country emojis

sampoder:
Context:

Screenshot 2025-10-01 at 10 45 00 PM

### DIFF
--- a/app/views/hcb_codes/_stripe_card.html.erb
+++ b/app/views/hcb_codes/_stripe_card.html.erb
@@ -114,7 +114,7 @@
         <% if defined?(@frame) && @frame %>
           <details>
             <summary class="inline-flex no-select">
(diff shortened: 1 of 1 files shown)


app/views/hcb_codes/_stripe_card.html.erb
```diff
@@ -317,7 +319,7 @@
         </p>
         <p>
           <strong>Country</strong>
-          <span><%= country_to_emoji @hcb_code.stripe_merchant["country"] %></span>
+          <span><%= @hcb_code.stripe_merchant["country"] ? country_to_emoji(@hcb_code.stripe_merchant["country"]) : "Unknown" %></span>
```
garyhtou, line 322: A cleaner solution would be to have `country_to_emoji` return `nil` if `nil` is passed in.

PR #11783, opened 2025-10-02
[Wise] Add "Convert to Wise transfer" button on reimbursement reports

YodaLightsabr:
## Summary of the problem
Off-by-one errors are causing issues with our current Wise reimbursement flow.

## Describe your changes
As a temporary workaround, this PR allows HCB admins to convert an existing Wise reimbursement report into a standard Wise transfer on the relevant organization. This way ops can still process reimbursements through Wise without having to worry about potentially fluctuating costs.

image

image


app/models/reimbursement/report.rb
```diff
@@ -152,8 +152,8 @@ class Report < ApplicationRecord
 
       event :mark_rejected do
         transitions from: [:draft, :submitted, :reimbursement_requested], to: :rejected
-        after do
-          ReimbursementMailer.with(report: self).rejected.deliver_later
+        after do |skip_mailer = false|
```
davidcornu: **suggestion**: This is hard to tell from the aasm docs but our own code suggests we can use a keyword arg here, which would make the call sites a lot more descriptive.

```suggestion
        after do |skip_mailer: false|
```

app/views/reimbursement/reports/_actions.html.erb
```diff
@@ -243,10 +243,36 @@
               <% turbo_confirm = "⚠️ WARNING: Approving this reimbursement will make the event balance negative. This organization does not have sufficient funds to fulfill this request. Are you sure you want to proceed?" %>
             <% end %>
             <% if report.user.payout_method.is_a?(User::PayoutMethod::WiseTransfer) %>
-              <%= link_to "#", class: "btn bg-info", data: { behavior: "modal_trigger", modal: "wise_approval" }, id: "submit_button" do %>
-                <%= inline_icon "thumbsup" %>
-                Approve & send Wise reimbursement
+              <%= link_to "#", class: "btn bg-info", data: { behavior: "modal_trigger", modal: "convert_to_wise_transfer" }, id: "submit_button" do %>
```
davidcornu: **suggestion**: Does this need to be a `<a>`? Semantically, if this is triggering an action then it should be a `<button>` (https://niquette.ca/articles/button-or-anchor/#buttons).

app/views/reimbursement/reports/_actions.html.erb
```diff
@@ -243,10 +243,36 @@
...
+                  <li>Close this reimbursement report without marking it as reimbursed</li>
+                </ul>
+
+                <div class="flex flex-row items-center justify-between mt-6">
+                  <%= link_to reimbursement_report_convert_to_wise_transfer_path(report), class: "btn btn-large bg-info", method: :post do %>
+                    <%= inline_icon "wise" %>
+                    Convert to Wise transfer
+                  <% end %>
+
+                  <%= link_to "#", class: "btn bg-error tooltipped tooltipped--n", "aria-label": "Don't do this without checking with engineering first!", data: { behavior: "modal_trigger", modal: "wise_approval" }, id: "submit_button" do %>
```
davidcornu: Same comment as https://github.com/hackclub/hcb/pull/11783/files#r2399002536.

app/views/reimbursement/reports/_actions.html.erb
```diff
@@ -311,6 +337,7 @@
                   </div>
 
                   <%= form.submit "Approve report" %>
+                  <p class="muted" style="font-size: 0.75rem;">🚧 Warning: this is risky—do not do this without checking with engineering first.</p>
```
davidcornu: **suggestion**: Let's avoid adding more inline styles to the code base

See https://tailwindcss.com/docs/font-size

```suggestion
                  <p class="muted text-xs">🚧 Warning: this is risky—do not do this without checking with engineering first.</p>
```

app/models/reimbursement/report.rb
```diff
@@ -350,6 +350,48 @@ def wise_transfer_may_exceed_balance?
       !::Shared::AmpleBalance.ample_balance?(wise_transfer_quote_amount.cents, event)
     end
 
+    def convert_to_wise_transfer!(as: User.system_user)
```
davidcornu, line 353: **suggestion**: This would be a good piece of functionality to add a test for

app/models/reimbursement/report.rb
YodaLightsabr: I had thought about this but I hadn't seen it used anywhere before. Are we doing it in our codebase?

app/views/reimbursement/reports/_actions.html.erb
YodaLightsabr: Not really, I was mostly just following the existing convention in the file. I agree that a button makes more sense.

app/models/reimbursement/report.rb
YodaLightsabr: I don't think we currently have any tests for reimbursement reports. I definitely agree that we should go in and add tests for the reimbursement flow, but I don't want it to be a blocker right now—we have 41 pending reimbursement reports blocking on this PR.

davidcornu (approved):

YodaLightsabr:
Thank you!

PR #11787, opened 2025-10-02
Refactor flavor text service to fix spooky lambda bug

YodaLightsabr:
## Summary of the problem
https://hackclub.slack.com/archives/C026RKHLPNJ/p1759420027929249

## Describe your changes
Refactor `generate` into two methods

### DIFF
--- a/app/services/flavor_text_service.rb
+++ b/app/services/flavor_text_service.rb
@@ -11,7 +11,7 @@ class FlavorTextService
     @random = Random.new(@seed)
   end
 
-  def generate
+  def sample
     return development_flavor_texts.sample(random: @random) if @env == "development"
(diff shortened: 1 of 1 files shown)


app/services/flavor_text_service.rb
```diff
@@ -11,7 +11,7 @@ def initialize(user: nil, env: Rails.env, deterministic: true)
     @random = Random.new(@seed)
   end
 
-  def generate
+  def sample
```
davidcornu: **suggestion**: We can probably make this `private` now

davidcornu (approved):

app/services/flavor_text_service.rb
YodaLightsabr: oop. good catch

PR #11792, opened 2025-10-03
Remove partial cent expenses

sampoder:
Cents should be whole numbers, never partial this code allowed for that. We will floor this and then pick up the reminder in the fee expense.

### DIFF
--- a/app/controllers/reimbursement/reports_controller.rb
+++ b/app/controllers/reimbursement/reports_controller.rb
@@ -225,7 +225,7 @@ module Reimbursement
 
             conversion_rate = (wise_total_without_fees_cents / @report.amount_to_reimburse_cents).round(4)
             @report.update(conversion_rate:)
-            approved_amount_usd_cents = @report.expenses.approved.sum { |expense| expense.amount_cents * expense.conversion_rate }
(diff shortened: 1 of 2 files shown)


app/controllers/reimbursement/reports_controller.rb
```diff
@@ -225,7 +225,7 @@ def admin_approve
 
             conversion_rate = (wise_total_without_fees_cents / @report.amount_to_reimburse_cents).round(4)
             @report.update(conversion_rate:)
-            approved_amount_usd_cents = @report.expenses.approved.sum { |expense| expense.amount_cents * expense.conversion_rate }
+            approved_amount_usd_cents = Reimbursement::Report.find("RnSKLM").expenses.approved.sum { |expense| (expense.amount_cents * expense.conversion_rate).floor }
```
davidcornu: **concern**: I assume this was debugging code 😅 

```suggestion
            approved_amount_usd_cents = @report.expenses.approved.sum { |expense| (expense.amount_cents * expense.conversion_rate).floor }
```

sampoder:
```suggestion
            approved_amount_usd_cents = @report.expenses.approved.sum { |expense| (expense.amount_cents * expense.conversion_rate).floor }
```

sampoder: oops - yes

PR #11794, opened 2025-10-03
[Wise] Honor payout method account holder when converting to transfer

davidcornu:
## Summary of the problem

https://hackclub.slack.com/archives/C026RKHLPNJ/p1759503377571179

When we are creating a `WiseTransfer` from a `ReimbursementReport` (see #11783), we are using `user.name` as the recipient when the user might have configured a different value in `User::PayoutMethod::WiseTransfer#account_holder`.

## Describe your changes

Check whether the payout method has a `account_holder` and use it instead of `user.name` if present.


YodaLightsabr (approved):
Thank you!

PR #11797, opened 2025-10-03
Handle card grant top-ups in `Disbursement::TransactionsHelper`

davidcornu:
## Summary of the problem

In #11757 I added a some logic to render the source and destination transactions for disbursements but failed to account for sub-ledgers, which meant card grant top-ups (where both the source and destination event are the same) were rendered in a very chaotic way.

## Describe your changes

- Instead of grouping transactions by `event`, use `[event_id, subledger_id]`
- Remove `.strict_loading` (see reasoning in the commit message)

**Before**

CleanShot 2025-10-03 at 15 17 33

**After**

CleanShot 2025-10-03 at 15 17 47

### DIFF
--- a/app/models/disbursement/transactions_helper.rb
+++ b/app/models/disbursement/transactions_helper.rb
@@ -7,39 +7,57 @@ class Disbursement
     end
 
     def settled_source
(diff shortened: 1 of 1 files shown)


sampoder (approved):
thank you!

PR #11807, opened 2025-10-07
[HcbCode] Refactor receipt ignore code to be more maintainable

Luke-Oldenburg:
## Summary of the problem

These transactions didn't exist pre-2024 so it can be moved to the first clause of the if.

### DIFF
--- a/app/models/hcb_code.rb
+++ b/app/models/hcb_code.rb
@@ -502,7 +502,7 @@ class HcbCode < ApplicationRecord
 
     return false unless event&.plan&.receipts_required?
 
-    ([:card_charge, :wise_transfer].include?(type)) ||
+    [:card_charge, :wise_transfer].include?(type) ||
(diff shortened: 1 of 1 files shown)


garyhtou (changes requested):
Good find!

**Feedback:** This code works. However, I would recommend restructuring the logic so that:
- By default, receipts are required for all types
- Then, we grant exceptions for specific types created before Feb 2nd, 2024.

This way, when we add more types (e.g. `wise_transfer`), this code does not need to be updated to explicitly require receipts for them. This is because all new types should have receipts required by default.

This may look like:
```ruby
# Before Feb. 2024, receipts were not required for ACHs, checks, PayPal transfers, and Wires
return false if ([:ach, :check, :paypal_transfer, :wire].include?(type) && created_at <= Time.utc(2024, 2, 1))

true
```

YodaLightsabr (approved):

garyhtou (approved):

PR #11813, opened 2025-10-07
[Donation Goal] Don't double / triple send goal reached emails

sampoder:
Screenshot 2025-10-07 at 12 04 54 PM

### DIFF
--- a/app/models/donation.rb
+++ b/app/models/donation.rb
@@ -366,13 +366,22 @@ class Donation < ApplicationRecord
       DonationMailer.with(donation: self).notification.deliver_later
     end
 
-    if event.donation_goal.present? && (event.donation_goal.progress_amount_cents >= event.donation_goal.amount_cents)
+    if reached_donation_goal?
       EventMailer.with(event:).donation_goal_reached.deliver_later
     end
   end
 
   def first_donation?
     self.event.donations.succeeded.size == 1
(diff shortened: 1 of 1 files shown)


app/models/donation.rb
```diff
@@ -375,6 +375,15 @@ def first_donation?
     self.event.donations.succeeded.size == 1
   end
 
+  def reached_donation_goal?
+    return false unless event.donation_goal.present?
+    return false unless event.donation_goal.progress_amount_cents >= event.donation_goal.amount_cents
+    # this prevents us from sending the email after the organization has already reached the goal
+    return false if (event.donation_goal.progress_amount_cents - amount_cents) >= event.donation_goal.amount_cents
```
garyhtou, line 382: Makes sense, thanks for the comment!

app/models/donation.rb
```diff
@@ -366,7 +366,7 @@ def send_notification
       DonationMailer.with(donation: self).notification.deliver_later
     end
 
-    if event.donation_goal.present? && (event.donation_goal.progress_amount_cents >= event.donation_goal.amount_cents)
+    if reached_donation_goal?
```
garyhtou, line 369: I very much appreciate you extracting it into a method!

garyhtou (approved):

PR #11817, opened 2025-10-07
Update allowance copy in OPI accepted mailer

### LINKED ISSUE #11811: [OPI Mailer] Update copy about $0 initial spending allowance
Image

Saying
> They have an initial spending allowance of $0.00.

feels a bit silly. Let's figure out if there's a reason why the copy is rendered like this. If not, let's update it!

Luke-Oldenburg:
## Summary of the problem

Closes #11811


sampoder (approved):

garyhtou:
Thank you!

PR #11830, opened 2025-10-08
[Wrapped] Abstract year into the Metric class

YodaLightsabr:
## Summary of the problem
Most `Metric`s are all based on data from the current or previous year. It can be a pain to update the year in every place, and it makes it hard to iterate on data for a new year.

## Describe your changes
This PR removes all hardcoded instances of `2024` in any metric and instead calls `Metric.year`.

### DIFF
--- a/app/models/metric.rb
+++ b/app/models/metric.rb
@@ -78,4 +78,8 @@ class Metric < ApplicationRecord
     metric
   end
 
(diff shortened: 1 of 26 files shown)


sampoder (approved):

sampoder:
thank you!

PR #11831, opened 2025-10-08
[Donation] Use `amount` not `amount_cents`

sampoder:
Screenshot 2025-10-08 at 12 44 39 PM

ctx ^

### DIFF
--- a/app/models/donation.rb
+++ b/app/models/donation.rb
@@ -379,7 +379,7 @@ class Donation < ApplicationRecord
     return false unless event.donation_goal.present?
     return false unless event.donation_goal.progress_amount_cents >= event.donation_goal.amount_cents
     # this prevents us from sending the email after the organization has already reached the goal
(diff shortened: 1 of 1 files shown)


app/models/donation.rb
```diff
@@ -379,7 +379,7 @@ def reached_donation_goal?
     return false unless event.donation_goal.present?
     return false unless event.donation_goal.progress_amount_cents >= event.donation_goal.amount_cents
     # this prevents us from sending the email after the organization has already reached the goal
-    return false if (event.donation_goal.progress_amount_cents - amount_cents) >= event.donation_goal.amount_cents
+    return false if (event.donation_goal.progress_amount_cents - amount) >= event.donation_goal.amount_cents
```
garyhtou, line 382: We should consider renaming the column at some point

garyhtou (approved):

app/models/donation.rb
sampoder: yes i agree, standards are good

PR #11837, opened 2025-10-09
[v4] Add donation availability to event

Mohamad-Mortada:
## Summary of the problem
@polypixeldev brought up a very good point. If an organization doesn't have donations enabled (or stuff like frozen) we should disable the Tap to Pay functionality for it on mobile. 

## Describe your changes
simple boolean `donation_available` to let yk if its supported


YodaLightsabr (dismissed):
Could you update the relevant test case here: https://github.com/hackclub/hcb/blob/ea00579ba5f3fbab3372b89f1e82ce9409219abf/spec/controllers/api/v4/card_grants_controller_spec.rb#L42

YodaLightsabr (approved):

PR #11838, opened 2025-10-09
[GSuite::Revocation] Copy edits to mailers

garyhtou:
## Summary of the problem

We're about the send out revocations, and I wanted to give the mailers a copy pass

## Describe your changes

Made the first warning mailer more friendly and less scary.


rluodev (approved):
LGTM!

PR #11842, opened 2025-10-10
Fix short session expiry bug

Luke-Oldenburg:
## Summary of the problem

Users were reporting being logged out before their session was supposed to be expired.


sampoder (approved):

garyhtou:
Amazing! @Luke-Oldenburg, can you please follow up in #hcb-team and messages in #hcb to let people know this has been patched?

PR #11846, opened 2025-10-12
Changed text that said HCB operates at a loss, which isn't true

pizzalover125:
## Summary of the problem
A perk stated that HCB operates at a loss, which isn't true anymore.

## Describe your changes
Changed the text.

### DIFF
--- a/app/views/events/perks/_hcb_stickers.html.erb
+++ b/app/views/events/perks/_hcb_stickers.html.erb
@@ -16,7 +16,7 @@
   <div class="mb0">
     <p><strong>Want some HCB stickers?</strong> You can request an envelope full of them for yourself or your teammates! Every shipment request is manually sent out by the HCB team, and every envelope is personally prepared, packed, and mailed, with the address handwritten.</p>
     <p><em><strong>With great power comes great responsibility...</strong></em></p>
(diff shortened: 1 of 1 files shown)


app/views/events/perks/_hcb_stickers.html.erb
```diff
@@ -16,7 +16,7 @@
   <div class="mb0">
     <p><strong>Want some HCB stickers?</strong> You can request an envelope full of them for yourself or your teammates! Every shipment request is manually sent out by the HCB team, and every envelope is personally prepared, packed, and mailed, with the address handwritten.</p>
     <p><em><strong>With great power comes great responsibility...</strong></em></p>
-    <p>Please enjoy responsibly. HCB operates at a loss that donors pay for, and stickers are not cheap—every package costs real donor money.</p>
+    <p>Please enjoy responsibly. Stickers are not cheap—every package costs real money.</p>
```
Luke-Oldenburg:
```suggestion
    <p>Please enjoy responsibly. Stickers are not cheap and every package costs real money.</p>
```

sampoder:
I think the original copy is fine and I'd like to preserve it - the spirit of the copy is better communicated in the original version. We can debate HCB's profitability separately as it depends on how you measure things afaik

PR #11847, opened 2025-10-12
[Invoices] Move deposited invoices into a state in aasm

### LINKED ISSUE #4992: Invoice: Fix state management

rluodev:
## Summary of the problem

Centralizing the columns!

## Describe your changes

Migrating data from current "deposited?" method into "aasm_state"'s "deposited_v2" state.


sampoder (commented):
Is this ready for review?

garyhtou (changes requested):
Please use a one time job for data modifications instead of using a migration

garyhtou (changes requested):
We also need to update the code first to:
1. Transition newly deposited invoices to `deposited_v2`
2. Change all reads to use a method that considers both `deposited?` method and the `deposited_v2` AASM state

After, we can then do this data migration.

db/migrate/20250918021841_migrate_invoice_states_to_aasm.rb
Luke-Oldenburg, line 1: Should this file be deleted?

rluodev: u are so right

app/jobs/one_time_jobs/backfill_deposited_state_for_invoices.rb
```diff
@@ -0,0 +1,18 @@
...
+    queue_as :default
+
+    def perform
+      Invoice.find_in_batches(batch_size: 100) do |batch|
+        batch.each do |invoice|
+          if invoice.deposited?
+            invoice.update_column(:aasm_state, :deposited_v2)
+          end
+        end
+      end
```
garyhtou:
```suggestion
      Invoice.find_each do |invoice|
        if invoice.deposited?
          invoice.mark_deposited!
        end
      end
```

app/models/invoice.rb
```diff
@@ -226,12 +226,11 @@ def archived?
...
 
   def deposited? # TODO move to aasm
-    canonical_transactions.count >= 2 || manually_marked_as_paid? || completed_deprecated?
+    deposited_v2? || canonical_transactions.count >= 2 || manually_marked_as_paid? || completed_deprecated?
   end
 
   def state
-    return :success if paid_v2? && deposited?
-    return :success if paid_v2? && event.can_front_balance?
+    return :success if (paid_v2? && event.can_front_balance?) || deposited?
```
garyhtou:
```suggestion
    return :success if (paid_v2? && event.can_front_balance?) # paid, and will be deposited soon
    return :success if deposited?
```

app/models/invoice.rb
```diff
@@ -244,7 +243,7 @@ def state
   end
 
   def state_text
-    return "Deposited" if paid_v2? && (event.can_front_balance? || deposited?)
+    return "Deposited" if (paid_v2? && event.can_front_balance?) || deposited?
```
garyhtou:
```suggestion
    return "Deposited" if paid_v2? && event.can_front_balance? # paid, and will be deposited soon
    return "Deposited" if deposited?
```

app/jobs/invoice/mark_depositeds_job.rb
```diff
@@ -0,0 +1,18 @@
...
+  class MarkDepositedsJob < ApplicationJob
+    queue_as :default
+    def perform
+      Invoice.where(aasm_state: :paid_v2).find_in_batches(batch_size: 100) do |batch|
+        batch.each do |invoice|
+          if invoice.canonical_transactions.count >= 2 || invoice.manually_marked_as_paid? || invoice.completed_deprecated?
+            invoice.mark_deposited!
+          end
+        end
+      end
```
garyhtou:
```suggestion
      Invoice.where(aasm_state: :paid_v2).find_each do |invoice|
        if invoice.canonical_transactions.count >= 2 || invoice.manually_marked_as_paid? || invoice.completed_deprecated?
          invoice.mark_deposited!
        end
      end
```

garyhtou (approved):

garyhtou (approved):

garyhtou (changes requested):
Actually, we're going to hold off from merging this because there's some odd prod state data.

garyhtou:
https://github.com/hackclub/hcb/issues/12319

PR #11848, opened 2025-10-13
[Reimbursements] Make it clear that the report cap is in USD

YodaLightsabr:
## Summary of the problem
It can be unclear that the report cap is in USD after switching to a different currency.

## Describe your changes
Explicitly show "USD" when the report amount is in a non-USD currency.

### DIFF
--- a/app/views/reimbursement/reports/show.html.erb
+++ b/app/views/reimbursement/reports/show.html.erb
@@ -15,7 +15,7 @@
       <%= "is requesting " unless @report.closed? %>
       <%= render "total", report: @report %>
       <% if @report.maximum_amount_cents %>
-        (<%= render_money @report.maximum_amount_cents %> <%= "USD" if @report.currency != "USD" %> cap)
+        (<%= render_money @report.maximum_amount_cents %> <%= "USD" unless @report.currency == "USD" %> cap)
       <% end %>
       <%= "from #{@event.name}" if @event && @use_user_nav %> for
     </span>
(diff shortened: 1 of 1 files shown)


app/views/reimbursement/reports/show.html.erb
```diff
@@ -15,7 +15,7 @@
       <%= "is requesting " unless @report.closed? %>
       <%= render "total", report: @report %>
       <% if @report.maximum_amount_cents %>
-        (<%= render_money @report.maximum_amount_cents %> cap)
+        (<%= "#{render_money(@report.maximum_amount_cents)}#{" USD" if @report.currency != "USD"}" %> cap)
```
YodaLightsabr:
```suggestion
        (<%= render_money @report.maximum_amount_cents %> <%= "USD" if @report.currency != "USD" %> cap)
```

app/views/reimbursement/reports/show.html.erb
```diff
@@ -15,7 +15,7 @@
       <%= "is requesting " unless @report.closed? %>
       <%= render "total", report: @report %>
       <% if @report.maximum_amount_cents %>
-        (<%= render_money @report.maximum_amount_cents %> cap)
+        (<%= render_money @report.maximum_amount_cents %> <%= "USD" if @report.currency != "USD" %> cap)
```
sampoder:
```suggestion
        (<%= render_money @report.maximum_amount_cents %> <%= "USD" unless @report.currency == "USD" %> cap)
```
tiny thing

sampoder (approved):

app/views/reimbursement/reports/show.html.erb
YodaLightsabr: lol I completely forgot.

PR #11850, opened 2025-10-13
Add a search bar animation

moodynooby:
## Summary of the problem

based on the [issue](https://github.com/hackclub/hcb/issues/7688) ,  I am trying to add a animation while searching
\

Closes #7688

## Describe your changes

I aimed to create a animation that is very non invasive to the code base and simple

### DIFF
--- a/app/javascript/components/command_bar/results.js
+++ b/app/javascript/components/command_bar/results.js
@@ -3,7 +3,6 @@
 import React from 'react'
 import { KBarResults, useMatches } from 'kbar'
 import { Guide } from './search/guide'
-
 export function RenderResults() {
   const { results, rootActionId } = useMatches()
   return (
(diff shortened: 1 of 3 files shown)


moodynooby:
@Luke-Oldenburg please let me know where I am wrong as this is still rough prototype

db/schema.rb
Luke-Oldenburg, line 1: Can you discard these changes please?

app/javascript/components/command_bar/results.js
```diff
@@ -3,7 +3,6 @@
 import React from 'react'
 import { KBarResults, useMatches } from 'kbar'
 import { Guide } from './search/guide'
-
```
Luke-Oldenburg, line 6: Can you add this newline back please?

moodynooby:
@Luke-Oldenburg  any update?

Luke-Oldenburg:
> @Luke-Oldenburg any update?

I believe the intention of this issue is for search bars such as on the reimbursements page rather than the command bar because there is no visual indication there. The "Loading..." text serves as a visual indication that the command bar is running the action although it may be nice to replace the text with an animation there. I could not get a search to run for long enough to see what it would look like though. Could you please attach a video?

moodynooby:
Oh I thought the component would be reused but anywasys let me get back to you with a video

PR #11855, opened 2025-10-13
Store Discord message ID on activities after they're sent to Discord

YodaLightsabr:
## Summary of the problem

We eventually want to handle replies to certain types of activity messages. However, I don't want to rely on parsing message content for us to understand the context behind a Discord message; it would be much better for us to store a reference to the message and look it up in our systems when we need to.

## Describe your changes

Adds `discord_message_id` to `activities`

### DIFF
--- a/app/jobs/discord/process_notification_job.rb
+++ b/app/jobs/discord/process_notification_job.rb
@@ -29,12 +29,14 @@ module Discord
 
       text = ReverseMarkdown.convert(html)[0..4000]
 
-      Discord::Bot.bot.send_message(@event.discord_channel_id, nil, false, {
-                                      description: text,
-                                      timestamp: @activity.created_at.iso8601,
-                                      author: { name: @user.name, icon_url: profile_picture_for(@activity.owner) },
-                                      color:
-                                    })
+      message = Discord::Bot.bot.send_message(@event.discord_channel_id, nil, false, {
+                                                description: text,
+                                                timestamp: @activity.created_at.iso8601,
+                                                author: { name: @user.name, icon_url: profile_picture_for(@activity.owner) },
+                                                color:
+                                              })
+
+      @activity.update(discord_message_id: message.id.to_s)
     end
 
     private
--- a/db/migrate/20251013183626_add_discord_message_id_to_activity.rb
+++ b/db/migrate/20251013183626_add_discord_message_id_to_activity.rb
@@ -1,0 +1,5 @@
+class AddDiscordMessageIdToActivity < ActiveRecord::Migration[7.2]
+  def change
+    add_column :activities, :discord_message_id, :string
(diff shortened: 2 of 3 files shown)


sampoder:
What if we send multiple messages for an activity in the future?

YodaLightsabr:
Right now activities are designed to fit into one single Discord embed. Even if we needed to add attachments or additional embeds, or even components, Discord allows for all of those to fit into a single message.

Even if we, at some point, decided to send multiple messages for an activity (I can't think of a reason we'd need to), we can still use this value as a representation of the first / primary message and have the bot send all follow-up messages as "replies" to keep the association on Discord's end.

db/migrate/20251013183626_add_discord_message_id_to_activity.rb
```diff
@@ -0,0 +1,5 @@
+class AddDiscordMessageIdToActivity < ActiveRecord::Migration[7.2]
+  def change
+    add_column :activities, :discord_message_id, :string
```
garyhtou, line 3: Can we please make this a separate model such as Discord::Message (with belongs_to :activity) instead of modifying activities?

This keeps each model single/clearly purposed and organized. And, also allow for easily adding future delivery channels such as Slack

app/jobs/discord/process_notification_job.rb
```diff
@@ -29,12 +29,14 @@ def perform(public_activity_id)
 
       text = ReverseMarkdown.convert(html)[0..4000]
 
-      Discord::Bot.bot.send_message(@event.discord_channel_id, nil, false, {
-                                      description: text,
-                                      timestamp: @activity.created_at.iso8601,
-                                      author: { name: @user.name, icon_url: profile_picture_for(@activity.owner) },
-                                      color:
-                                    })
+      message = Discord::Bot.bot.send_message(@event.discord_channel_id, nil, false, {
```
garyhtou, line 32: At some point, we need to handle the case where this fails

db/migrate/20251013183626_add_discord_message_id_to_activity.rb
YodaLightsabr: I was hoping we wouldn't have to but I can see how that would be cleaner. Good point about potentially supporting other delivery mechanisms in the future.

sampoder:
>Right now activities are designed to fit into one single Discord embed. Even if we needed to add attachments or additional embeds, or even components, Discord allows for all of those to fit into a single message.

Oh I meant if we sent the same message but to different channels

YodaLightsabr:
> Oh I meant if we sent the same message but to different channels

Only one channel ID can be set per organization.

YodaLightsabr:
Closing in favor of #11871

PR #11856, opened 2025-10-13
Allow admins to inspect objects on HCB

YodaLightsabr:
## Summary of the problem

HCB engineers and operations staff frequently need to access information about a record that isn't show in the interface. Sometimes records can get into broken states, and you can't access the intended admin page. Or, sometimes we just don't show enough about an object, requiring staff to use Blazer or use the production console.

## Describe your changes

This PR proposes adding an admin-only inspection toolbar. Inspired by Stripe and Cockpit, this toolbar intelligently detects the primary object of the current page you're on (using some Rails routing magic) and renders a text-based representation of the object to assist with troubleshooting and debugging. It also allows you to follow any direct associations or look up an object manually by and of its primary fields.

image


app/controllers/admin_controller.rb
```diff
@@ -1404,6 +1404,36 @@ def referral_program_create
   def active_teenagers_leaderboard
   end
 
+  def inspect
```
garyhtou, line 1440: I believe this overrides a Ruby method

app/controllers/admin_controller.rb
```diff
@@ -1404,6 +1404,36 @@ def referral_program_create
...
+    redirect_to inspect_resource_admin_index_path(resource: params[:resource_type], id: params[:resource_id])
+  end
+
+  def inspect_resource
+    @resource_type = params[:resource]
+    @resource_id = params[:id]
+
+    Zeitwerk::Loader.eager_load_all
```
garyhtou, line 1448: There are likely some columns that we shouldn't even allow admins inspect. I'm specifically thinking of encrypted columns such as session tokens, plaid tokens, etc.

At the moment, encrypted columns are not available to admins via Blazer bc blazer doesn't decrypt.

app/controllers/admin_controller.rb
```diff
@@ -1404,6 +1404,36 @@ def referral_program_create
...
+
+    # get all named classes extending ApplicationRecord
+    @all_resource_types = ObjectSpace.each_object(Class).select { |c| c < ApplicationRecord }.select(&:name).map(&:name)
```
garyhtou, line 1451: Is this different than calling `descendants` on ApplicationRecord?

Also, maybe we should do ActiveRecord::Base to capture more models such as those created by gems. See which classes it would add.

app/models/inspector.rb
```diff
@@ -0,0 +1,57 @@
...
+    Zeitwerk::Loader.eager_load_all
+
+    descendants = ObjectSpace.each_object(Class).select { |c| c < ApplicationRecord }
+
+    descendants.filter_map(&:name)
+  end
+
+  def self.find_relations(object)
+    if object.class < ApplicationRecord
+      object = object.attributes
```
garyhtou, line 14: I believe AR has a method something along the lines of reflect_on_associations.

garyhtou (commented):
first pass!

app/controllers/admin_controller.rb
YodaLightsabr: ah, good catch. maybe I'll restructure these to be one route. there was previously a need for them to be separate but I can combine them now since it's all being done in turbo.

app/controllers/admin_controller.rb
YodaLightsabr: I had considered making a concern, say `inspectable`, that could help us standardize configuration keep track of which columns should be accessible on the model itself. Does that sound like a good path forward?

> At the moment, encrypted columns are not available to admins via Blazer bc blazer doesn't decrypt.

Do we want to hide encrypted columns in the inspection toolbar? I can see a case for either being made; on one hand, it allows for more destructive powers as an admin, but on the other hand, it reduces our reliance on the production Rails console.

app/controllers/admin_controller.rb
YodaLightsabr: I'll look into this. I think I had that approach initially and it caused some issues, but I don't recall exactly why I switched away from this. I'll get back to you.

app/models/inspector.rb
YodaLightsabr: It does—this was a quick and dirty way of adding a shortcut for `[anything]_id`, but it would be nice to work with other association types too.

PR #11772, opened 2025-10-01
[v4] Activate Grant Cards

Mohamad-Mortada:
## Summary of the problem
I wanna have the ability to activate grant cards on mobile (kinda how we have org invites). 

## Describe your changes

I copied most of the logical from the web controller and tweaked it for the API. Then you can activate via `/api/v4/card_grants/xxxxxxx/activate`
https://github.com/hackclub/hcb/blob/841c9782c608aeda7ca226638299a3a19595a76b/app/controllers/card_grants_controller.rb#L147-L157

PR #11858, opened 2025-10-14
Replace session object with just IP address in StripeCardService::Create

Mohamad-Mortada:
## Summary of the problem
For #11772 I can't use the `current_session` object as it is not defined in the API and we are only passing in current session for the IP address in the service. 

## Describe your changes
I just changed the parameters to just take in the IP address instead of the whole session object.

### DIFF
--- a/app/controllers/card_grants_controller.rb
+++ b/app/controllers/card_grants_controller.rb
@@ -147,7 +147,7 @@ class CardGrantsController < ApplicationController
   def activate
     authorize @card_grant
 
-    @card_grant.create_stripe_card(current_session.ip)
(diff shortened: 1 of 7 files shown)


app/controllers/card_grants_controller.rb
```diff
@@ -147,7 +147,7 @@ def spending
   def activate
     authorize @card_grant
 
-    @card_grant.create_stripe_card(current_session)
+    @card_grant.create_stripe_card(current_session.ip)
```
garyhtou:
```suggestion
    @card_grant.create_stripe_card(request.remote_ip)
```

garyhtou (approved):

PR #11873, opened 2025-10-17
Don't require a signed in user for receipt status image

polypixeldev:
## Summary of the problem

https://hackclub.slack.com/archives/GBXBFEED9/p1759940792433679
In Gmail, a broken image is displayed where it should be a button to upload the receipt.
Originally, we thought this was due to Gmail not rendering SVGs correctly. However, it seems that we were actually requiring a signed in user to fetch the image, which would not be sent when viewing from any email client, but it would exist when viewing from hcb.hackclub.com, letter opener, or previews in development.

It's still possible that SVGs could not work in Gmail, but I'll test to see if it works after this PR is merged and deployed.

## Describe your changes

Bypass requiring a signed in user for the receipt status action.

### DIFF
--- a/app/controllers/hcb_codes_controller.rb
+++ b/app/controllers/hcb_codes_controller.rb
@@ -205,7 +205,7 @@ class HcbCodesController < ApplicationController
   def receipt_status
     @secret = params[:s]
     @hcb_code = HcbCode.find_signed(@secret, purpose: :receipt_status)
-    
+
     raise Pundit::NotAuthorizedError if @hcb_code.nil?
   end
 
--- a/app/controllers/hcb_codes_controller.rb
+++ b/app/controllers/hcb_codes_controller.rb
@@ -4,7 +4,7 @@ class HcbCodesController < ApplicationController
   include TagsHelper
 
   skip_before_action :signed_in_user, only: [:receipt, :attach_receipt, :receipt_status, :show]
-  skip_after_action :verify_authorized, only: [:receipt]
+  skip_after_action :verify_authorized, only: [:receipt, :receipt_status]
 
   def show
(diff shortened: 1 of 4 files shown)


sampoder (dismissed):
Should we always look receipt up by secret?

```
HcbCode.find_signed(@secret, purpose: :receipt_status) == @hcb_code
```

sampoder:
Or I guess my question is when would we not be be looking up by secret>

polypixeldev:
Good point - we are always looking up by secret since most likely, there will be no signed in user associated with the request sent by the email client. I've moved the `receipt_status` route to be on the collection instead of member since I don't think we need to have both the signed ID and an unsigned ID in the URL

app/controllers/hcb_codes_controller.rb
```diff
@@ -203,13 +203,8 @@ def dispute
...
 
   def receipt_status
-    @hcb_code = HcbCode.find(params[:id])
     @secret = params[:s]
-
-    authorize @hcb_code
-
-  rescue Pundit::NotAuthorizedError
-    raise unless HcbCode.find_signed(@secret, purpose: :receipt_status) == @hcb_code
+    @hcb_code = HcbCode.find_signed(@secret, purpose: :receipt_status)
```
garyhtou, line 207:
```suggestion
    @hcb_code = HcbCode.find_signed(@secret, purpose: :receipt_status)
    
    raise Pundit::NotAuthorizedError if @hcb_code.nil?
```

garyhtou (approved):

garyhtou:
thought: With how this PR is written, existing images in emails will continue not to work, even though we theoretically could fix them. This is because we change how the URL is generated.

PR #11881, opened 2025-10-17
[v4 API] Prevent admins from obtaining stripe card ephemeral keys

garyhtou:
## Summary of the problem

It was reported by @thedev132 that admins can add other people's cards to Apple Wallet via the mobile app

## Describe your changes

Disallow admins from obtaining someone else's stripe card ephemeral key


Mohamad-Mortada (approved):
LGTM

PR #11883, opened 2025-10-17
Update eligibility criteria text for premium pins

alexdevz:
Added eligibility note for US-based organizations.

## Summary of the problem

Costs for intl shipping are too high- decision has been made to be US-only for the time being until we can find a solution.

## Describe your changes

Added text that mentions it is for US-based organizations only.

### DIFF
--- a/app/views/events/perks/_premium_pins.html.erb
+++ b/app/views/events/perks/_premium_pins.html.erb
@@ -28,7 +28,7 @@
 
   <p>
     Ever wish you were just a little more dripped out in HCB swag?
-    We'll ship you some super cool custom HCB pins and keychains for free if you have 5 or more active teenagers!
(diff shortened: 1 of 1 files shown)


alexdevz:
let me know if i didnt do this right 😅

Luke-Oldenburg (dismissed):
We should probably disable it using `!event.country_US?` and add a corresponding tooltip.

alexdevz:
> We should probably disable it using `!event.country_US?` and add a corresponding tooltip.

Yep that would be smart, I just had no idea if we had something to flag for that.

garyhtou (approved):

garyhtou:
Good idea @Luke-Oldenburg, we can add that in another PR

PR #11885, opened 2025-10-18
[Admin] Add transfer approval limits

garyhtou:
## Summary of the problem

For security, we want to implement transfer approval limits for admins. All users (including admins) should still be able to request transfers of any amount. However, only certain individuals can approve larger amounts.

## Describe your changes

- Admins can approve <= $5k
- Superadmins can approve <= $200k
- If anyone needs >= $200k, talk to me! (@garyhtou)

This applies to all external transfers that require human review.

image

### DIFF
--- a/app/controllers/reimbursement/reports_controller.rb
+++ b/app/controllers/reimbursement/reports_controller.rb
@@ -201,6 +201,8 @@ module Reimbursement
 
     def admin_approve
       authorize @report
+      # TODO: This does NOT consider currency
+      Governance::Admin.ensure_may_approve_transfer!(current_user, @report.amount_to_reimburse_cents)
 
       begin
         @report.with_lock do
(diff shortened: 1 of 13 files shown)


app/controllers/reimbursement/reports_controller.rb
```diff
@@ -201,6 +201,8 @@ def request_reimbursement
 
     def admin_approve
       authorize @report
+      # TODO: This does NOT consider currency
+      Governance::Admin.ensure_may_approve_transfer!(current_user, @report.amount_to_reimburse_cents)
```
garyhtou, line 205: From my understanding, there's no easy way to get a USD amount cents for a reimbursement report. Wise and Wire have this functionality, so I guess we could add logic to look into those models. However, this is a method we should probably add to `Reimbursement::Report`

YodaLightsabr (approved):
Looks good! I think this is a pretty clean implementation.

PR #11895, opened 2025-10-20
Improve edit user UI for auditors

Luke-Oldenburg:
## Summary of the problem

Buttons and form elements weren't disabled for auditors.


garyhtou (approved):
very nice!

PR #11897, opened 2025-10-20
[Docs] Updated DNS provider from DNS Simple to Cloudflare

afonsoingles:
## Summary of the problem

We no longer use DNS Simple as a DNS provider, so the dev docs were outdated.

## Describe your changes

Updated the DNS provider in the docs from DNS Simple to Cloudflare.


garyhtou (approved):
ty

PR #11900, opened 2025-10-21
[Auditor] Disable edit security UI (Part 2)

Luke-Oldenburg:
## Summary of the problem

Buttons and form elements weren't disabled for auditors.

### DIFF
--- a/app/views/users/_oauth_authorization.erb
+++ b/app/views/users/_oauth_authorization.erb
@@ -38,7 +38,7 @@
               <% if token.expires_in %>
                 <%= link_to make_authorization_eternal_users_path(id: token.id), method: :post,
                             class: "btn bg-success py1 px2 h4",
-                            disabled: !admin_signed_in? do %>
+                            disabled: !policy(token).make_eternal? do %>
                   <%= inline_icon "clock", size: home_action_size %>
                   Make eternal
                 <% end %>
--- a/app/views/users/_oauth_authorization.erb
+++ b/app/views/users/_oauth_authorization.erb
@@ -3,7 +3,7 @@
     <span class="bold flex items-center">
       <span><%= authorization.application.name %></span> <span class="badge bg-info ml1">App</span>
     </span>
-    <%= link_to revoke_oauth_application_users_path(authorization.application.id), method: :delete, class: "muted tooltipped tooltipped--w z5", 'aria-label': "Sign out of this app" do %>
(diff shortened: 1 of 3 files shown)


app/views/webauthn_credentials/_webauthn_credential.html.erb
```diff
@@ -13,7 +13,7 @@
         <span class="secondary">No name</span>
       <% end %>
     </span>
-    <%= link_to user_webauthn_credential_path(webauthn_credential.user, webauthn_credential), method: :delete, class: "muted tooltipped tooltipped--w z5", 'aria-label': "Delete this security key" do %>
+    <%= link_to user_webauthn_credential_path(webauthn_credential.user, webauthn_credential), method: :delete, class: "muted tooltipped tooltipped--w z5", 'aria-label': "Delete this security key", disabled: do %>
```
garyhtou: Instead of taking in `disabled`, which is `!policy(@user).update?`, we should check `!policy(webauthn_credentials).destroy?`

app/views/users/_oauth_authorization.erb
```diff
@@ -3,7 +3,7 @@
     <span class="bold flex items-center">
       <span><%= authorization.application.name %></span> <span class="badge bg-info ml1">App</span>
     </span>
-    <%= link_to revoke_oauth_application_users_path(authorization.application.id), method: :delete, class: "muted tooltipped tooltipped--w z5", 'aria-label': "Sign out of this app" do %>
+    <%= link_to revoke_oauth_application_users_path(authorization.application.id), method: :delete, class: "muted tooltipped tooltipped--w z5", 'aria-label': "Sign out of this app", disabled: do %>
```
garyhtou, line 6: This one is good to keep since there's no explicit policy for it

app/views/users/_oauth_authorization.erb
```diff
@@ -37,7 +37,8 @@
             <td>
               <% if token.expires_in %>
                 <%= link_to make_authorization_eternal_users_path(id: token.id), method: :post,
-                            class: "btn bg-success py1 px2 h4" do %>
+                            class: "btn bg-success py1 px2 h4",
+                            disabled: !admin_signed_in? do %>
```
garyhtou: This has its own policy. `ApiTokenPolicy#make_eternal?`

garyhtou (approved):

PR #11902, opened 2025-10-21
[Perks] Disable Premium Pins for non-US orgs

Luke-Oldenburg:
## Summary of the problem

https://github.com/hackclub/hcb/pull/11883#issuecomment-3423884774


garyhtou (approved):
nice! ty

PR #11906, opened 2025-10-21
[CPT Mailer] Switch to PNGs for receipt status image

polypixeldev:
## Summary of the problem

#11873 tried to fix an issue where receipt status images did not display in Gmail. While that PR did fix an issue affecting all email clients, it seems that Gmail still doesn't display the image as it is an SVG.

## Describe your changes

Re-exported the receipt buttons from Figma as a PNG and modified the controller to send the appropriate PNG from the assets directory.


app/controllers/hcb_codes_controller.rb
```diff
@@ -206,7 +206,13 @@ def receipt_status
     @secret = params[:s]
     @hcb_code = HcbCode.find_signed(@secret, purpose: :receipt_status)
 
-    raise Pundit::NotAuthorizedError if @hcb_code.nil?
+    if @hcb_code.nil? && HcbCode.find_signed(@secret, purpose: :receipt_status) != @hcb_code
```
garyhtou: I'm a bit confused by the `&& HcbCode.find_signed(@secret, purpose: :receipt_status) != @hcb_code`.

From my understanding, if `@hcb_code` is `nil`, then `HcbCode.find_signed(@secret, purpose: :receipt_status)` must also be nil in this scenario.

polypixeldev: Yep, missed this while resolving a merge conflict here - fixed!

garyhtou (approved):
Imma merge, but can we please get higher-quality screenshots? These appear a bit blurry.

PR #11907, opened 2025-10-21
[Subevents] Add scoped user-facing tags

### LINKED ISSUE #11899: [Subevent] Add user facing tags
When an org has many subevents, it can be difficult to organize/sort them. It would be nice to subevent tags available for organizers to set and filter by.

@polypixeldev is going to add subevent tags.
- Tags are scoped to the parent event
- Tags can be added/removed from the subevent page
- On the subevent page, you can filter by tag
- On the subevent page, you'll see a subevent balance based on the subevents visible (what's allowed by the filters)

We aim to have something presentable by Wednesday night.

Nice to have:
- Color for tags
- Add tags to CSV exports on subevent page
- Maybe add the ability to add/remove tags from the subevent's event settings page

polypixeldev:
## Summary of the problem
Subevents should be able to be tagged with user-facing tags to help filter them!

Closes #11899

## Describe your changes
Adds a new `Event::ScopedTag` model, which is the new user-facing tag model that is scoped to a parent event. This model has a many-to-many relationship with `Event` through `Event::ScopedTagsEvent`.

Managers of the parent event can create, destroy, and add tags to a new subevent they are creating within the form using a combobox implemented with Alpine, similar to the `PaymentRecipient` combobox. Tags can also be created, deleted, and toggled from the add tag menu on each event card on the sub-organizations page.

Sub-organizations can be filtered by tag, which will also update the balance shown at the top of the sub-organizations page.

https://github.com/user-attachments/assets/4d853ca3-ceac-4a3c-b647-9db92285f596

**note:** I've disabled ERB lint for `app/views/events/sub_organizations.html.erb` and `app/views/events/_scoped_tag_option.html.erb` as that is what was done for the other files where Alpine is used, since erb_lint doesn't like HTML attributes that start with `@`. I'll make a PR after this to change all `@click` attributes to use `x-on:click` to avoid needing to disable ERB lint for the whole file. Other than issues related to the `@click` attribute, these files pass the lints.


app/models/event/scoped_tag.rb
```diff
@@ -0,0 +1,19 @@
...
+#
+# Table name: event_scoped_tags
+#
+#  id         :bigint           not null, primary key
+#  name       :string
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+class Event
+  class ScopedTag < ApplicationRecord
```
garyhtou, line 22: Should ScopeTag belong to a parent event? This way, we know how to get a list of scope tags. Without it, I don't think we're actually able to scope them to a parent event.

db/migrate/20251021013000_create_event_scoped_tags.rb
```diff
@@ -0,0 +1,9 @@
+class CreateEventScopedTags < ActiveRecord::Migration[7.2]
+  def change
+    create_table :event_scoped_tags do |t|
+      t.string :name
```
garyhtou: should be not null

db/migrate/20251021013806_create_event_scoped_tags_events.rb
```diff
@@ -0,0 +1,12 @@
+class CreateEventScopedTagsEvents < ActiveRecord::Migration[7.2]
+  def change
+    create_table :event_scoped_tags_events, id: false do |t|
+      t.belongs_to :event, null: false, foreign_key: true
+      t.belongs_to :event_scoped_tag, null: false, foreign_key: true
+
+      t.index [:event_scoped_tag_id, :event_id], unique: true
```
garyhtou, line 7: Good!

app/models/event/scoped_tag.rb
polypixeldev: My thinking was that we wouldn't have any `ScopedTag`s that don't have at least one event, so from there we would be able to add a validation that whenever an `EventScopedTagsEvent` is created, the associated `Event` must have the same `parent_id` as the other events on the associated `ScopedTag`. Making `ScopedTag` belong to a parent event would definitely make this clearer, though, and as long as we have the proper validation it should stay in sync.

app/models/event/scoped_tag.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    belongs_to :parent_event, class_name: "Event"
+
+    validate :name_is_unique_within_parent_event
+
+    def removal_confirmation_message
+      message = "Are you sure you'd like to delete this tag?"
+
+      if events.length > 0
+        message + " It will be removed from #{pluralize(events.length, 'event')}."
```
garyhtou:
```suggestion
        message + " It will be removed from #{pluralize(events.length, 'organization')}."
```

app/models/event/scoped_tag.rb
```diff
@@ -0,0 +1,52 @@
...
+      else
+        message
+      end
+    end
+
+    private
+
+    def name_is_unique_within_parent_event
+      if parent_event.subevent_scoped_tags.where(name:).exists?
+        errors.add(:name, "is not unique within the parent event")
```
garyhtou:
```suggestion
        errors.add(:name, "is not unique within the parent organization")
```

app/controllers/events_controller.rb
```diff
@@ -837,11 +837,18 @@ def sub_organizations
...
         search = params[:q] || params[:search]
+        scoped_tag = Event::ScopedTag.find_by(name: params[:tag])
 
         relation = @event.subevents
         relation = relation.where("name ILIKE ?", "%#{search}%") if search.present?
+        relation = relation.joins(:scoped_tags).where("event_scoped_tags.id = #{scoped_tag.id}") if scoped_tag.present?
         relation = relation.order(created_at: :desc)
 
+        @filter_options = [
+          { key: "tag", label: "Tag", type: "select", options: @event.subevent_scoped_tags.map { |scoped_tag| scoped_tag.name } }
```
```suggestion
          { key: "tag", label: "Tag", type: "select", options: @event.subevent_scoped_tags.map(&:name) }
```

garyhtou (approved):
Very nice! Imma merge this so that it's ready for tmrw

PR #11910, opened 2025-10-21
Fix invoice min of $1.00

anishanne:
## Summary of the problem
HelpScout Ticket #49269

When flagging a reimbursement as a personal transaction, for a charge of $1.00 (⁠-100), the condition ⁠-100 >= -100 is ⁠true, so we incorrectly show an error.


Luke-Oldenburg (approved):
lgtm!

garyhtou (approved):

PR #11911, opened 2025-10-22
[OrganizerPositionInvite] Hint when accepting invite with wrong user

garyhtou:
## Summary of the problem

A user ran into a confusing moment when they attempted to accept an invite with another email address. They were unaware they were using the wrong account because all HCB says is "You are not authorized to perform this action."

## Describe your changes

🥁 Introducing.... a helpful hint!

image


YodaLightsabr (approved):
LGTM

Issue #11912, opened 2025-10-22
[User] Deter fake ages


garyhtou:
<img width="422" height="482" alt="Image" src="https://github.com/user-attachments/assets/78db3ca7-ab1e-4afb-9502-01bd7ae6cd63" />

I'm not saying you can't be 103 years old and using HCB, but i'm fairly certain there are some unreasonable ages in there (espeically given the peak at 100). In fact, there are 6 people who aren't born yet! oh!

We need this: https://www.tiktok.com/@thematchapost/video/7407859609320410410
and strong birthday/age validations in the User model.

PR #11913, opened 2025-10-22
[User] Add some age helper methods

garyhtou:
These aren't currently used within the app, but are handy in the Rails Console when running queries ([example](https://hcb.hackclub.com/console/sessions/1211) and [another example](https://hackclub.slack.com/archives/C02BD6CEW65/p1761116205388589)).

It will also come in handy for https://github.com/hackclub/hcb/issues/11912


manuthecoder (approved):

garyhtou:
ty!

PR #11920, opened 2025-10-23
🧠 Remember login methods from previous logins

YodaLightsabr:
## Summary of the problem

HCB PMO... it doesn't remember what login methods I use across devices, and I spend probably a combined hour each year just clicking "Sign in another way" when an email code gets sent to my inbox.

## Describe your changes

HCB will infer the login method you'd like to use based on the last successful login attempt. It will automatically select the two methods you've previously used, so there's no need to click "Sign in another way" and feel bad for the tireless and pointless work done by `::LoginCodeService::Request`.


sampoder (approved):
this looks good to me! thank you for splitting up email & sms

PR #11921, opened 2025-10-23
[Admin] Time-based, customizable admin transfer approval limits

garyhtou:
## Summary of the problem

To further lock down our security and improve upon https://github.com/hackclub/hcb/pull/11885, we're implementing time-based transfer limits for admins, with a clear, easily auditable, paper trail of approval and attempts.

## Describe your changes

**TLDR:** Each admin now has a user-specific daily transfer limit. For example, Bob can approve $5k worth of transfers in the span of 1 day. Alice can approve $50k worth of transfers in the span of 1 day.

There really are no visual changes. However, there are some functionality changes:
- Each user has a daily transfer limit, currently only settable via the Rails console; although I plan to add a UI for superadmins.
  image

- We block transfer approvals from an impersonated session
   image
   <sub>btw, mel you don't smell lol</sub>

- All approval attempts (whether approved or denied) are logged. And when a request is denied (such as when an admin hits their transfer limits), it is reported to Gary & Mel so they know when people's transfer limits need to be increased.
  image

- Transfer limits need to be explicitly configured
   image

## What's next?

In future PRs:
- Multi-step approvals. Requiring two approvals for huge transfers
- UI for superadmins to configure transfer limits


app/models/governance/admin/transfer/approval_attempt.rb
```diff
@@ -0,0 +1,102 @@
...
+        belongs_to :user
+        belongs_to :transfer, polymorphic: true
+
+        monetize :attempted_amount_cents
+        validates :attempted_amount_cents, numericality: { greater_than: 0 }
+        validate :one_successful_approval_per_transfer, on: :create
+
+        enum :result, {
+          approved: "approved",
+          denied: "denied"
```
garyhtou, line 65: send email to mel/gary if denied

garyhtou: [`31c3259` (#11921)](https://github.com/hackclub/hcb/pull/11921/commits/31c325982d57412028dca8ef28571c49db20e602)

app/models/governance/admin.rb
```diff
@@ -2,57 +2,8 @@
 
 module Governance
   module Admin
-    TRANSFER_APPROVAL_LIMITS = {
```
garyhtou, line 5: This removes the first scrappy approval system I created last week

app/models/user.rb
```diff
@@ -231,10 +231,13 @@ def auditor?
     ["auditor", "admin", "superadmin"].include?(self.access_level) && !self.pretend_is_not_admin
   end
 
-  # admin? takes into account an admin user's preference
+  # admin? by default, takes into account an admin user's preference
   # to pretend to be a non-admin, normal user
-  def admin?
-    ["admin", "superadmin"].include?(self.access_level) && !self.pretend_is_not_admin
+  def admin?(override_pretend: false)
```
garyhtou, line 236: I chose to add an argument to the existing `admin?` method instead of using `admin_override_pretend?` because `admin_override_pretend?` has a different behavior since it considers auditors.

We should get rid of `admin_override_pretend?` and just use the `admin?`, `auditor?`, and `superadmin?` method with an extra `override_pretend` argument.

app/mailers/governance/admin/transfer/approval_attempt_mailer.rb
```diff
@@ -0,0 +1,36 @@
...
+
+          mail to: report_recipients,
+               subject: "[HCB] Admin Transfer Denied: #{impersonation_snippet}#{@approval_attempt.user.name} for #{@approval_attempt.attempted_amount.format}"
+        end
+
+        private
+
+        def report_recipients
+          REPORT_RECIPIENTS.filter_map do |public_id|
+            User.find_by_public_id(public_id).email_address_with_name
```
garyhtou:
```suggestion
            User.find_by_public_id(public_id)&.email_address_with_name
```

PR #11923, opened 2025-10-23
[Event] Hot-fix creation by adding safe nav to scoped_tags

garyhtou:
## Summary of the problem

UI passed in `nil` for `scoped_tags`, which raised "undefined method 'map' for nil"

## Describe your changes

Add safe nav, and make default value `n`.il


garyhtou:
https://github.com/hackclub/hcb/pull/11922

PR #11926, opened 2025-10-24
[v4] Preload only the paginated event transactions

Mohamad-Mortada:
## Summary of the problem
API times to load big orgs like HQ or HCB Ops is wayyyyyy longer than ledger. Why? The code is preloading associations for ALL transactions, and we're only showing 25 of them.

## Describe your changes
This moves the association preloader logic after we have paginated them, so only preloads the 25 or whatever transactions instead of all.

### DIFF
--- a/app/controllers/api/v4/events_controller.rb
+++ b/app/controllers/api/v4/events_controller.rb
@@ -31,7 +31,7 @@ module Api
 
         if @transactions.any?
 
-          page_settled = @transactions.select { |tx| tx.is_a?(::CanonicalTransactionGrouped) }
+          page_settled = @transactions.select { |tx| tx.is_a?(CanonicalTransactionGrouped) }
           page_pending = @transactions.select { |tx| tx.is_a?(CanonicalPendingTransaction) }
 
           if page_settled.any?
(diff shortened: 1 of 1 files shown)


app/controllers/api/v4/events_controller.rb
```diff
@@ -20,17 +20,28 @@ def transactions
...
         type_results = ::EventsController.filter_transaction_type(params[:type], settled_transactions: @settled_transactions, pending_transactions: @pending_transactions)
         @settled_transactions = type_results[:settled_transactions]
         @pending_transactions = type_results[:pending_transactions]
 
         @total_count = @pending_transactions.count + @settled_transactions.count
         @transactions = paginate_transactions(@pending_transactions + @settled_transactions)
+
+        if @transactions.any?
+
+          page_settled = @transactions.select { |tx| tx.is_a?(::CanonicalTransactionGrouped) }
```
YodaLightsabr:
```suggestion
          page_settled = @transactions.select { |tx| tx.is_a?(CanonicalTransactionGrouped) }
```

YodaLightsabr (approved):

PR #11927, opened 2025-10-24
[Wise] Show more descriptive error message & fix CAD payout settings

YodaLightsabr:
## Summary of the problem

We don't want to re-process payout holdings that are no longer pending, however we're currently filtering out pending payout holdings with a scope, even though there's already code to filter them out in the job.

We're also not storing the Interac email, which can be ambiguous on the operations side and mess up our data if their HCB email address changes.

## Describe your changes

This PR removes the `.pending` scope (which prevents `ActiveRecord::RecordNotFound` errors causing a 404) and does away with the "Use the same email for Interac?" checkbox.

image


Mohamad-Mortada (approved):

Luke-Oldenburg (approved):

garyhtou (commented):
Great, thank you

PR #11931, opened 2025-10-24
[Suborganization] Turboframe balance to improve performance

### LINKED ISSUE #11918: [Subevent] Improve performance of subevents page
This page can take forever to load once you have 400 orgs

Image

YodaLightsabr:
## Summary of the problem

The sub-organizations page is slow

## Describe your changes

Preload associations and make balance async

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -562,6 +562,8 @@ class EventsController < ApplicationController
   def async_sub_organization_balance
     authorize @event
 
+    @sub_organizations = filtered_sub_organizations
+
     render :async_balance, layout: false
   end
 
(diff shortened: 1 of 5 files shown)


app/controllers/events_controller.rb
```diff
@@ -559,6 +559,12 @@ def async_balance
     render :async_balance, layout: false
   end
 
+  def async_sub_organization_balance
+    authorize @event
```
garyhtou, line 563: Unfortunately, this is a regression. The previous balance calcuation considered filters since it used `@sub_organizations` which is after filters are applied.

I believe this new async frame doesn't consider filters.

garyhtou:
I pushed [`0c0cb94` (#11931)](https://github.com/hackclub/hcb/pull/11931/commits/0c0cb948d15dac53a6648a0d3c868a2fd0ceb138) because I found that the balance was often out of sync with the displayed list of filtered events.

There were a few reasons why:
1. We were caching the frame (without consideration of filter parameters), but the balance can take sooo long to load that it creates confusion. This style of caching (stale while revalidate) can be great when the filter parameters are taken into consideration.
2. We had [`data-turbo-permanent`](https://turbo.hotwired.dev/reference/attributes#:~:text=data%2Dturbo%2Dpermanent) set which prevented the balance from updating the DOM in certain cases.

PR #11933, opened 2025-10-26
[Organization] Home page tweaks

manuthecoder:
Since we're planning to release the home page publicly, it's essential we make some minor changes to the UI to ensure it looks great!

Here's what changed: 
* Renamed Graphs → Insights
* Added links and border to Total raised/Total spent heading to make it look less empty
    * image
* Removed inconsistent background from "All time" button on light mode
* Adjusted padding/margin in card titles
* Fixed center adjustment for top tags/top merchants empty states
* Fixed line breaks for empty states so it looks even

### DIFF
--- a/app/views/events/home/_categories_chart.html.erb
+++ b/app/views/events/home/_categories_chart.html.erb
@@ -11,7 +11,7 @@
         It's quiet here...
       </h3>
       <p class="text-gray-500 my-1">
-        Categories will appear here<br/>as you spend more money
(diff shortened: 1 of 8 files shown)


Luke-Oldenburg:
Could we add see all buttons to money in and money out using the new direction filters?

garyhtou (approved):
we're gonna merge, but will need to

- Move "See all" on recent transactions to the top right
- Luke's comment, but also do it for total raised/spent.

PR #11944, opened 2025-10-27
[Announcements] Update notice email to clarify monthly announcements

polypixeldev:
## Summary of the problem

On October 25th, emails were sent to many managers of organizations with monthly announcements enabled with the subject "Your scheduled monthly announcement has been canceled". This was confusing and surprising since this was the first time most of these organizations had heard of monthly announcements.

## Describe your changes

The announcement notice mailer was never sent out, so it has been updated to include a clarification for the events that got the cancellation email. For events with monthly announcements enabled that did not get the cancellation email, the notice mailer will just contain an overview of announcements and monthly announcements.

`OneTimeJobs::SendAnnouncementNotice.perform_now` should be run in the console to send this email after this PR has been merged and deployed.

For events that got the cancellation email:
image

For events that did not get the cancellation email:
image

PR #11945, opened 2025-10-27
[Announcements] Unlink transparency mode and monthly announcements

polypixeldev:
## Summary of the problem

See #11944 for context - for now, we've decided to not automatically link transparency mode to enabling monthly announcements as it doesn't give the event managers any chance to get context on what monthly announcements are.

## Describe your changes

Remove the callbacks on `Event::Configuration` and `Event` that automatically enabled monthly announcements if the event was/became transparent.


Luke-Oldenburg (approved):
lgtm!

garyhtou (approved):

PR #11947, opened 2025-10-27
[Discord] Don't rely on HCB code being passed through `custom_id`

YodaLightsabr:
## Summary of the problem

We're requiring that the HCB code exists at the time of activity sending (which isn't always the case), and we're also relying on the HCB passed through Discord.

## Describe your changes

By computing the HCB code from the Discord message ID, we're able to rely less on the custom ID given by Discord. And, by removing this dependency, we can skip passing in an HCB code when sending the public activity notification, and we're less susceptible to race conditions where the HCB code doesn't yet exist.

### DIFF
--- a/app/jobs/discord/handle_interaction_job.rb
+++ b/app/jobs/discord/handle_interaction_job.rb
@@ -99,9 +99,9 @@ module Discord
     end
 
     def attach_receipt_component
-      hcb_code = HcbCode.find_by_hashid!(@params)
       discord_message = Discord::Message.find_by(discord_message_id: @interaction.dig(:message, :id))
       activity = PublicActivity::Activity.find_by(id: discord_message&.activity_id)
+      hcb_code = activity.trackable&.canonical_pending_transaction&.local_hcb_code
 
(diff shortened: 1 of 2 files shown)


app/jobs/discord/handle_interaction_job.rb
```diff
@@ -99,9 +99,9 @@ def ping_component
     end
 
     def attach_receipt_component
-      hcb_code = HcbCode.find_by_hashid!(@params)
       discord_message = Discord::Message.find_by(discord_message_id: @interaction.dig(:message, :id))
       activity = PublicActivity::Activity.find_by(id: discord_message&.activity_id)
+      hcb_code = activity.trackable&.canonical_pending_transaction&.local_hcb_code
```
garyhtou: Based on how this is written, it appears `discord_message` and `activity` may be `nil`. Should we explicitly handle that case?

At the moment, it may error when calling `trackable` on `nil`.

garyhtou (approved):
Ty!

app/jobs/discord/handle_interaction_job.rb
YodaLightsabr:
```suggestion
      hcb_code = activity&.trackable&.canonical_pending_transaction&.local_hcb_code
```

PR #11952, opened 2025-10-28
[Admin] Improve performance of admin email list

garyhtou:
## Summary of the problem

This page was getting unusable. The biggest culprit was generating a select field with EVERY user in the database!

And then, with a bunch of iframes on the page, browser extensions were going crazy trying to load into every iframe.

## Describe your changes

Turboframe the iframes... yes, I know... crazy. See https://github.com/hackclub/hcb/blob/c8b67027858888deb0084ffb15a81ae4f7a30607/app/views/admin/email.html.erb for an explanation.

Turn the User select dropdown into a simple `user_id` text field. Yeah, it sucks now, but at least the page will load.

### DIFF
--- a/app/views/admin/emails.html.erb
+++ b/app/views/admin/emails.html.erb
@@ -55,7 +55,7 @@
   <section class="modal bg-snow" style="height: 60vh;" data-behavior="modal" role="dialog" id="message_<%= msg.id %>">
     <%= modal_header "Email to #{msg.user&.name || "unknown user"}" %>
     <%= turbo_frame_tag dom_id(msg), src: url_for(controller: "admin", action: "email", message_id: msg.id), loading: "lazy" do %>
-      loading...
+      Loading...
     <% end %>
   </section>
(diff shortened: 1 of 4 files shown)


app/views/admin/emails.html.erb
```diff
@@ -56,4 +51,13 @@
   </tbody>
 </table>
 
+<% @messages.each do |msg| %>
+  <section class="modal bg-snow" style="height: 60vh;" data-behavior="modal" role="dialog" id="message_<%= msg.id %>">
+    <%= modal_header "Email to #{msg.user&.name || "unknown user"}" %>
+    <%= turbo_frame_tag dom_id(msg), src: url_for(controller: "admin", action: "email", message_id: msg.id), loading: "lazy" do %>
+      loading...
```
YodaLightsabr:
```suggestion
      Loading...
```

YodaLightsabr (approved):
Looks good, thank you!

PR #11953, opened 2025-10-28
[Governance ApprovalAttempt] Handle redundant approvals

garyhtou:
## Summary of the problem

At the moment, when the admin attempts to approve a transfer a second time, it errors on them, blocking the transfer from being processed.

See `one_successful_approval_per_transfer` validation for explanation on why redundant approvals happen.

## Describe your changes

Instead of erroring when an admin attempts to approve a second time, we record the redundant approval.


garyhtou:
Merging in the interest of unblocking the ops team

PR #11959, opened 2025-10-28
[Wise] Fix Wise reimbursement expense payout inaccuracy

YodaLightsabr:
## Summary of the problem

Wise reimbursements are breaking because the expense payouts and payout holdings are not the correct amounts. This is because the following lines of code do _very_ different things:

```diff
expense_payouts << Reimbursement::ExpensePayout.create!(amount_cents: (-expense.amount_cents * expense.conversion_rate).floor, event: expense.report.event, expense:)

expense_payouts << Reimbursement::ExpensePayout.create!(amount_cents: -(expense.amount_cents * expense.conversion_rate).floor, event: expense.report.event, expense:)
```

The one above takes the floor after the amount is negated (effectively taking the ceiling before negating). This causes issues when we're using the floor value for the positive end of the USD calculation (the way we calculate the Wise fee expense).

## Describe your changes

Moves the negative sign outside of the parentheses to floor the amount before negating it (the same process used in other places where we convert expense currencies).

### DIFF
--- a/app/models/reimbursement/report.rb
+++ b/app/models/reimbursement/report.rb
@@ -409,7 +409,7 @@ module Reimbursement
       expense_payouts = []
 
       expenses.approved.each do |expense|
-        expense_payouts << Reimbursement::ExpensePayout.create!(amount_cents: (-expense.amount_cents * expense.conversion_rate).floor, event: expense.report.event, expense:)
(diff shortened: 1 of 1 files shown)


sampoder (approved):
maths is not maths

Image

garyhtou:
Crazy bug! I'm so glad we found it. Great work here

PR #11970, opened 2025-10-30
[Team] Invite flow redesign

### LINKED ISSUE #11916: [Invite links] Combine email-based invites and invite links into one modal
### Problem
The current invite link  UX could use some rework as part of our larger goal to make HCB more intuitive. 

Image

### Proposed solution

Let's redesign the invite link UX with an interface similar to Notion's page sharing experience. While we'll draw inspiration from their pattern, our implementation will differ since HCB supports creating multiple invite links rather than just one.

We should also add dedicated copy and share buttons for each link. Currently, users need to right-click and select "Copy link address" to copy an invite link, which isn't intuitive or mobile-friendly.

https://github.com/user-attachments/assets/213a3cad-a022-494d-9fc4-6c1089189043

manuthecoder:
### Objective
The goal of this PR is to combine email-based invites and invite links into one modal, similar to Google Docs. See #11916 

### Changes made
they say a picture is worth a thousand words, so here's million with a video

https://github.com/user-attachments/assets/6587cf72-d3a4-4855-8774-682a3d7bc794


garyhtou (commented):
image
spending controls should be only shown for members. and hidden for readers and managers

garyhtou (approved):

PR #11983, opened 2025-11-01
[v4] Card Designs

Mohamad-Mortada:
## Summary of the problem
@Luke-Oldenburg already made a create cards endpoint that takes in a stripe personalization ID but we dont know what they are. This gives users more flexibility as they can choose their design. 

## Describe your changes
Create a card designs method to fetch designs for either specifically an event or in general if event id not provided. 
Endpoint would be /api/v4/card/card_designs

### DIFF
--- a/app/controllers/api/v4/stripe_cards_controller.rb
+++ b/app/controllers/api/v4/stripe_cards_controller.rb
@@ -159,7 +159,7 @@ module Api
       def card_designs
         if params[:event_id].present?
           set_api_event
-          authorize @event, :create_stripe_card?
+          authorize @event, :create_stripe_card?, policy_class: EventPolicy
 
           @designs = [event.stripe_card_personalization_designs&.available, StripeCard::PersonalizationDesign.common.available].flatten.compact
         else
--- a/app/controllers/api/v4/stripe_cards_controller.rb
+++ b/app/controllers/api/v4/stripe_cards_controller.rb
@@ -156,6 +156,20 @@ module Api
 
       end
 
+      def card_designs
+        if params[:event_id].present?
+          set_api_event
+          authorize @event, :create_stripe_card?
+
+          @designs = [event.stripe_card_personalization_designs&.available, StripeCard::PersonalizationDesign.common.available].flatten.compact
+        else
+          skip_authorization
+          @designs = StripeCard::PersonalizationDesign.common.available
+        end
+
+        @designs += StripeCard::PersonalizationDesign.unlisted.available if current_user.auditor?
+      end
+
     end
   end
 end
--- a/app/views/api/v4/stripe_cards/card_designs.json.jbuilder
+++ b/app/views/api/v4/stripe_cards/card_designs.json.jbuilder
@@ -5,6 +5,7 @@ json.array! @designs do |design|
   json.name design.name_without_id
   json.color design.color
   json.status design.stripe_status
+  json.unlisted design.unlisted?
   json.common design.common
   json.logo_url rails_blob_url(design.logo) if design.logo.attached?
 end
--- a/app/views/api/v4/stripe_cards/card_designs.json.jbuilder
+++ b/app/views/api/v4/stripe_cards/card_designs.json.jbuilder
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 json.array! @designs do |design|
-  json.id design.id
   json.name design.name_without_id
   json.color design.color
   json.status design.stripe_status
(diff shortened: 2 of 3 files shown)


app/controllers/api/v4/stripe_cards_controller.rb
```diff
@@ -156,6 +156,20 @@ def ephemeral_keys
 
       end
 
+      def card_designs
+        if params[:event_id].present?
+          set_api_event
+          authorize @event, :create_stripe_card?
```
Luke-Oldenburg:
```suggestion
          authorize @event, :create_stripe_card?, policy_class: EventPolicy
```
We also probably want to have a different policy for this.

app/controllers/api/v4/stripe_cards_controller.rb
```diff
@@ -156,6 +156,20 @@ def ephemeral_keys
...
+
+          @designs = [event.stripe_card_personalization_designs&.available, StripeCard::PersonalizationDesign.common.available].flatten.compact
+        else
+          skip_authorization
+          @designs = StripeCard::PersonalizationDesign.common.available
+        end
+
+        @designs += StripeCard::PersonalizationDesign.unlisted.available if current_user.auditor?
```
Luke-Oldenburg:
```suggestion
        @designs += StripeCard::PersonalizationDesign.unlisted.available if auditor_signed_in?
```

app/views/api/v4/stripe_cards/card_designs.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+json.array! @designs do |design|
+  json.id design.id
```
Luke-Oldenburg: I see designs don't currently have a hashid but we may want to consider using that if we are going to add this to the API.

app/controllers/api/v4/stripe_cards_controller.rb
Luke-Oldenburg, line 170:
```suggestion
        @designs += StripeCard::PersonalizationDesign.unlisted.available if auditor_signed_in?
```

Mohamad-Mortada: auditor signed in doesn't exist in the API. I made the same mistake and it errored

app/views/api/v4/stripe_cards/card_designs.json.jbuilder
Mohamad-Mortada: So do you suggest just removing for now? Not a big problem imo

app/controllers/api/v4/stripe_cards_controller.rb
Mohamad-Mortada: imo the policy is fine as designs should only be accessible by people who are actually using it to create a card. is ur worry abt it being too restrictive?

Luke-Oldenburg: It would be an identical policy most likely but named differently.

app/controllers/api/v4/stripe_cards_controller.rb
Luke-Oldenburg: Should we make a separate PR to add the helpers for that to the API?

app/views/api/v4/stripe_cards/card_designs.json.jbuilder
Luke-Oldenburg: I would recommend either adding hashid or public id. Whichever one the other APIs use.

app/controllers/api/v4/stripe_cards_controller.rb
Mohamad-Mortada: yea we prob should

app/views/api/v4/stripe_cards/card_designs.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+json.array! @designs do |design|
+  json.name design.name_without_id
+  json.color design.color
+  json.status design.stripe_status
+  json.unlisted design.unlisted?
+  json.common design.common
+  json.logo_url rails_blob_url(design.logo) if design.logo.attached?
```
garyhtou:
```suggestion
  json.logo_url design.logo.attached? ? rails_blob_url(design.logo) : nil
```

garyhtou (approved):
At some point, let's move this to its own controller. Probably `Api::V4::StripeCards::PersonalizationDesignsController#index`.

PR #11984, opened 2025-11-01
Implement currency conversion and update total display in reimbursments

Pegoku:
## Summary of the problem

Many reimbursements are in non-USD currencies, and it can be hard to know if the amount requested is over the limit.

## Describe your changes

Introduce a method to convert amounts to USD and update the total display to show the USD equivalent in the report.

### Example:
image

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -37,6 +37,22 @@ module ApplicationHelper
     render_money(amount, opts)
   end
 
+  # Convert an amount in a given currency to USD
+  # @param amount_cents [Integer] The amount in cents
+  # @param currency [String] The currency code (e.g., "EUR", "GBP")
+  # @return [Money] The equivalent amount in USD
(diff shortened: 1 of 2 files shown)


garyhtou:
@YodaLightsabr can you please look at this

Luke-Oldenburg:
I don't think we want to be making the heading longer. Maybe we could tooltip this?

Pegoku:
Did the change. Do you think it's better now? 
image

Luke-Oldenburg:
@garyhtou what do you think? Should we do something similar for the cap with an estimation of it in the reimbursement currency?

YodaLightsabr:
We want to make it clear how much a non-USD reimbursement report will cost in USD, but I'm not sure the title is the right place for that. We've recently added this information to the action bar at the bottom of the report, and it's more accurate because it uses the Wise API to get the mid-market rate and factor in fees.

PR #11985, opened 2025-11-02
Use safe navigation on `ancestor_organizer_positions`

Luke-Oldenburg:
## Summary of the problem

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/1388/samples/timestamp/2025-11-02T21:14:00Z

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -302,7 +302,7 @@ class EventsController < ApplicationController
     @positions = Kaminari.paginate_array(@all_positions).page(params[:page]).per(params[:per] || @view == "list" ? 20 : 10)
 
     if @event.parent
-      ops = @event.ancestor_organizer_positions.includes(:user)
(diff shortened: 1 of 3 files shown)


garyhtou (commented):
We're a little confused as to how `@event.ancestor_organizer_positions` can be `nil` since `ancestor_organizer_positions` should always return a `OrganizerPosition::ActiveRecord_Relation` even when no records are found.

I'll also mention that @Luke-Oldenburg experiences a few bugs on his computer (running Fedora 42) on both Chromium and Firefox.
- When visiting the admin stripe card personalization design page, it instantly redirects to a random personalization design
- When visiting any personalization design show page, it 500s

I was unable to reproduce either of these bugs on my computer (MacOS; Chrome).

Based on AppSignal, I can see that @thedev132 also ran into this error (500 on show page), so @Luke-Oldenburg is going to try to reproduce this with @thedev132 and find the root cause.

PR #12014, opened 2025-11-03
[IntraFi Transactions] Auto-map after importing

garyhtou:
## Summary of the problem

There is code to auto-map IntraFi transactions, however, it's ran once a day: https://github.com/hackclub/hcb/blob/8444870912091589304cce75a733c926e0145585/app/jobs/transaction_engine/nightly_job.rb#L10-L9

## Describe your changes

To speed things, we explicitly run the mapping engine after we import new IntraFi transactions.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -313,6 +313,9 @@ class AdminController < Admin::BaseController
       end
     end
 
+    # Auto mapp the transactions
+    ::EventMappingEngine::Nightly.new.run
+
     duplicates = transactions.count - raw_intrafi_transactions.count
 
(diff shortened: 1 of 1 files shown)


app/controllers/admin_controller.rb
```diff
@@ -313,6 +313,9 @@ def raw_intrafi_transactions_import
       end
     end
 
+    # Auto mapp the transactions
+    ::EventMappingEngine::Nightly.new.run
```
garyhtou, line 317: Ideally, this happens in a background job, but it's fine. This endpoint is hit once a month and by an engineer.

YodaLightsabr (approved):

PR #12016, opened 2025-11-04
[UI3] Redirect requests to UI3

garyhtou:
## Summary of the problem

We want certain people to use UI3. However, the application may accidentally redirect users away from UI3. So we're adding logic in the main prod app (located a hcb.hackclub.com) to redirect the user back to UI3.

## Describe your changes

WIP

TODO @manuthecoder:
- ensure it works for POSTs and other non-GET requests
- See if we have any infinite redirects


garyhtou:
One thought: It may be fine if we only redirect GET requests. This is because UI3 only contains UI changes. There are no backend changes. This means that it's fine with POST requests originating from UI3 hit the main prod deployment (unless CORS blocks it or something).

Luke-Oldenburg:
concern: will this stop people with the UI3 flipper set from accessing the normal site?

garyhtou:
> concern: will this stop people with the UI3 flipper set from accessing the normal site?

Correct. It will keep them on UI3 until the flipper is disabled for them. The goal is to dogfood UI3 and ensure we get bugs reported and fixed.

manuthecoder (approved):

manuthecoder (approved):

PR #12018, opened 2025-11-04
[Announcements] Add job to automatically enable monthly announcements

polypixeldev:
## Summary of the problem

Monthly announcements have been in production for a few months now, but we haven't really rolled them out well yet, and there's been some confusion about them for events that got them automatically enabled without any notice.

## Describe your changes

We've decided to only automatically enable monthly announcements for transparent teen-led events. This PR adds a new one time job, `OneTimeJobs::EnableMonthlyAnnouncements` to do this. It finds all eligible events that do not already have monthly announcements enabled, enables it, and sends a notice email that explains the feature and links to the new monthly announcement that will be sent out on December 1st.

A parameter `exclude_ids` is added to the job to make sure we don't automatically enable the feature for any events that we just sent the clarification email in #11944 to - it would be weird if someone disabled monthly announcements after receiving that email, and then we automatically enabled it for them _again_.

To run in production after merging:
```ruby
exclude_ids = [...] # we have saved the list of IDs that we sent the clarification email to
OneTimeJobs::EnableMonthlyAnnouncements.perform_now(exclude_ids:)
```

image


app/jobs/one_time_jobs/enable_monthly_announcements.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module OneTimeJobs
+  class EnableMonthlyAnnouncements < ApplicationJob
+    def perform(exclude_ids:)
+      eligible_events = Event.transparent.organized_by_teenagers.where.not(id: exclude_ids).includes(:config).where(config: { generate_monthly_announcement: false })
```
garyhtou: Only include orgs that have people on them. otherwise, there's no one for the announcement mailer to email.

```suggestion
      eligible_events = Event.transparent.organized_by_teenagers.where.not(id: exclude_ids).includes(:config).where(config: { generate_monthly_announcement: false }, id: OrganizerPosition.pluck(:event_id))
```

garyhtou (approved):

PR #12019, opened 2025-11-04
Replace is_signee with a new owner role

### LINKED ISSUE #12012: [Organizer Position] Add `Owner` role
At the moment, we have managers, but they can be either signees or not. That's confusing.

So, we're adding a new role called Owner. Owners are basically managers who are contract signees. All Owners must be a contract signee.

polypixeldev:
## Summary of the problem

The logic was starting to get a little confusing with signees being a different column than role, but signees must be a manager and there must be at least 1 signee per organization. Since roles are now fully shipped, it makes sense to make signees be an owner role that is above all other roles.

Closes #12012 

## Describe your changes

This PR does a few things:
- Adds the owner role to the role enum
- Adds the owner role to the invite form, but only to admins
- Replaces all references to the `is_signee` column on `OrganizerPosition` and `OrganizerPositionInvite` to reference the owner role
- Updates copy mentioning the manager role to say owners and managers
- Adds maintenance tasks to backfill the role

**When this PR is merged:**
Run the `BackfillOpOwnersTask` and `BackfillOpiOwnersTask` maintenance tasks

**Next PRs:**
1. Drop the `is_signee` column on `OrganizerPosition` and `OrganizerPositionInvite`
2. Update policies to only require an OPDR when the acting user is below manager or when removing an owner

**Note:** While users can be promoted to the Owner role by an admin after accepting their invite to an organization, this will not automatically send them the contract. The best way for this to work might be to store organizer position roles as manager at first, and then promote to owner once the contract is signed. This should be considered when revamping contracts in the new onboarding flow.

**Invite form:**
image


app/models/organizer_position_invite.rb
```diff
@@ -223,19 +216,21 @@ def on_contract_signed(contract)
...
-      end
+      ActiveRecord::Base.transaction do
+        # Unfreeze the event if this is the first signed contract
+        if event.contracts.signed.count == 1
+          event.update!(financially_frozen: false)
+        end
 
-      organizer_position&.update!(fiscal_sponsorship_contract: contract)
+        organizer_position&.update!(role: :owner, fiscal_sponsorship_contract: contract)
+        update!(role: :owner)
```
garyhtou: Disallow sending contracts to manager. Instead, the UI should allow you to change their role to owner; and as a result, a contract is sent.

PR #12025, opened 2025-11-04
[Wise] Show reimbursements with incomplete processing at the top of admin page

YodaLightsabr:
## Summary of the problem

It's bad when Wise reimbursement reports aren't fully processed, but it can be tricky to keep track of which ones need to be followed up on.

## Describe your changes

This PR highlights unfinished Wise reimbursement reports by placing them at the top of the admin page in orange. In theory, reports are never in this `settled` payout holding state for more than a few minutes while being processed. Any reports shown at the top should be looked into immediately.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -487,15 +487,15 @@ class AdminController < Admin::BaseController
 
     relation = relation.reimbursement_requested if @pending
 
-    @unprocessed_wise_reports = Reimbursement::Report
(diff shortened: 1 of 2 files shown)


app/controllers/admin_controller.rb
```diff
@@ -487,9 +487,16 @@ def reimbursements
 
     relation = relation.reimbursement_requested if @pending
 
+    @unprocessed_wise_reports = Reimbursement::Report
```
```suggestion
    @unprocessed_wise_report_ids = Reimbursement::Report
```

garyhtou (approved):

PR #12032, opened 2025-11-05
[Wise] Fix breakdown for reimbursements that have been sent

YodaLightsabr:
## Summary of the problem

The breakdown doesn't make much sense for reimbursements that have already been sent.

## Describe your changes

Change the wording for reimbursements with a final USD amount (payout holding amount) and hide created fee expenses from the table.


garyhtou (approved):
problem & fix makes sense to me. please give it another thorough test!

PR #12040, opened 2025-11-06
[Wise] Update admin UI for approving reimbursements

YodaLightsabr:
## Summary of the problem

The previous UI was meant to only be temporary. Now that the normal reimbursement flow is working, we want to emphasize that it's the preferred method for processing reimbursements.

## Describe your changes

image


YodaLightsabr:
Merging this to prevent future conflicts because it's only admin copy changes

PR #12044, opened 2025-11-07
[Announcement::Block] Remove unused html columns and add explainer callout

polypixeldev:
## Summary of the problem

Announcement blocks still had the `rendered_html` and `rendered_email_html` columns. These were added so that blocks would not always update with new data after publishing - however, now that blocks now have parameters, this is no longer as much of a concern as they can have a defined end date for data that they use.

Blocks are also not explained to users, which can be confusing as they don't behave like normal text in an announcement, and other parts of HCB such as monthly announcements depend on logic using blocks.

## Describe your changes

First, ignores the `rendered_html` and `rendered_email_html` columns, and removes references to them including the ability to refresh a block - blocks are always rendered on-demand with the latest data.

**#12991 should be merged after this PR is merged and deployed to drop the columns.**

Secondly, adds an info callout when viewing an unpublished announcement that explains what blocks are and how to use them:
image


Luke-Oldenburg (commented):
It appears the migration was ran and then deleted without rerunning migrations resulting in the model headers still missing the columns.

polypixeldev:
> It appears the migration was ran and then deleted without rerunning migrations resulting in the model headers still missing the columns.

Annotations respect the ignored columns setting, so that's why they're missing. The only thing this PR doesn't do is actually delete the data so that ActiveRecord isn't caching columns that don't exist in the database.

garyhtou (approved):

PR #12045, opened 2025-11-07
Direct users to links using event slugs instead of IDs

polypixeldev:
## Summary of the problem

See https://hackclub.slack.com/archives/CN523HLKW/p1762382193901539 - non-admin users couldn't create invite links since the form and links controller were using links that included the event ID instead of the slug in the path. `SetEvent` restricts using event IDs to only auditors+.

## Describe your changes

Replaced the url helper calls to just use `event:` instead of the ID, as the url helper automatically handles using the slug.
Tested in development.


manuthecoder (approved):
LGTM!

Luke-Oldenburg (approved):
lgtm!

Mohamad-Mortada (approved):
LGTM

YodaLightsabr (approved):

garyhtou (approved):

PR #12047, opened 2025-11-07
Fix follower/following user list borders

manuthecoder:
This PR fixes the ugly borders for follower/following lists when in dark mode, as reported by @Luke-Oldenburg

### Before
image

### After
image


sampoder (approved):
amazing

PR #12068, opened 2025-11-10
Block members from deleting receipts on expenses they don't own

### LINKED ISSUE #12065: [Receipt] Don't allow members to modify reimbursements they don't own
Do we want members to be able to modify receipts if it's for a reimbursement report that doesn't belong to them? We allow for it currently but I think it should be disabled.

_Originally posted by @Luke-Oldenburg in https://github.com/hackclub/hcb/pull/10665#discussion_r2485069542_

Luke-Oldenburg:
## Summary of the problem

closes #12065

### DIFF
--- a/app/policies/receipt_policy.rb
+++ b/app/policies/receipt_policy.rb
@@ -10,14 +10,14 @@ class ReceiptPolicy < ApplicationPolicy
       return record.user == user
     end
 
+    # the receipt is on a reimbursement report. people making reports may not be in the organization.
+    if record.receiptable.instance_of?(Reimbursement::Expense)
+      return (record.receiptable.report.user == user || OrganizerPosition.role_at_least?(user, record.receiptable.event, :manager)) && unlocked?
+    end
+
     # any members of events should be able to modify receipts.
     if record.receiptable.event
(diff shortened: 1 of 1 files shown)


app/policies/receipt_policy.rb
```diff
@@ -10,16 +10,16 @@ def destroy?
       return record.user == user
     end
 
+    # the receipt is on a reimbursement report. people making reports may not be in the organization.
+    if record.receiptable.instance_of?(Reimbursement::Expense)
+      return (record.receiptable.report.user == user || OrganizerPosition.role_at_least?(user, record.receiptable.event, :manager)) && unlocked?
```
Luke-Oldenburg, line 15: Could event be nil here if the reimbursement was started via email? Would that be an issue?

garyhtou:
See question here: https://github.com/hackclub/hcb/issues/12065#issuecomment-3530407181

garyhtou (approved):

app/policies/receipt_policy.rb
garyhtou: I believe all receiptables have an event

PR #11976, opened 2025-10-31
Upgrade to Rails 8 without turning framework defaults on

albertchae:
## Describe your changes

This is an update of https://github.com/hackclub/hcb/pull/10836 but with another pass at merging the generated configs from `bin/rails app:update` and also not turning on the values in `config/initializers/new_framework_defaults_8_0.rb` so we can deploy without those first.

The general philosophy for merging the generated configs is to make the config look like as if we had started HCB with rails new on 8.0 and then applied the changes we did over HCB's lifetime to get it to the current state. This meant deleting default values (especially commented out ones) if they were removed in newer versions of rails and copying over comment blocks to match rails 8.0 generators.

We also have to explicitly edit `config/environments/staging.rb`. This isn't a standard rails environment so `bin/rails app:update` doesn't generate a new template for it. Instead I diff-ed it against `production.rb` and moved over new defaults that seemed to make sense while keeping staging specific changes as is

These files have the biggest diffs with the fresh configs from `bin/rails app:update`, you can see the HTML output of vimdiff to double check
- [development.rb](https://html-preview.github.io/?url=https://gist.githubusercontent.com/albertchae/db2f8e5bfd6f67ebdfc955f88ee5190b/raw/1a48e776ab993e84f747f1a6e3527995024a75bf/development.rb.html)
- [production.rb](https://html-preview.github.io/?url=https://gist.githubusercontent.com/albertchae/06e8bf914a4c04f2da304f831efb836c/raw/6b1a7bd9308aed3b43ff51f456a9924a6c8b441b/production.rb.html)
- [test.rb](https://html-preview.github.io/?url=https://gist.githubusercontent.com/albertchae/12a9a3b9e0a09ef8d7240b8249b5657d/raw/35fcc2df9b200609c7e0cef89121fe40fb170902/test.rb.html) 
- ~~[puma.rb](https://html-preview.github.io/?url=https://gist.githubusercontent.com/albertchae/afdd05f52a14157e9ad19284a1522b01/raw/a490121afa16f3cf87a3dd0bfaf5e1a4bb2c7d01/puma.rb.html)~~ This is very close to the default now, the only difference is the appsignal plugin

PR #12071, opened 2025-11-11
Upgrade strong_migrations to v2

polypixeldev:
## Summary of the problem

After our migration to Rails 8 started (#11976), we started seeing that `rails db:migrate` was no longer working as expected - it did not detect new migrations to run. This was caused by our old version of the `strong_migrations` gem, which had been held back to major version 1 because we still used Postgres 11 in development. Now that we use Postgres 15 in development, we can safely upgrade `strong_migrations` to the latest version.

## Describe your changes

Upgrades `strong_migrations` to version 2.5.1.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -108,9 +108,7 @@ gem "reverse_markdown" # public activity to discord
 gem "namae" # multi-cultural human name parser
 gem "premailer-rails" # css to inline styles for emails
 gem "safely_block"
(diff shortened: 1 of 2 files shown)


garyhtou (approved):
Based on https://github.com/ankane/strong_migrations/blob/master/CHANGELOG.md#200-2024-06-28, we should be safe to upgrade since 2.0 only dropped support for old ruby, activerecord, and postgres versions

PR #12073, opened 2025-11-11
[Code Style] Remove ? from private policies

Luke-Oldenburg:
## Summary of the problem

We weren't following typical policy convention.


sampoder:
tbh i think i preferred what we had before ¯\\\_(ツ)_/¯

Luke-Oldenburg:
> tbh i think i preferred what we had before ¯\_(ツ)_/¯

we should standardize whatever we choose to go with; this syntax was used in some files already but not all of them

PR #12075, opened 2025-11-11
[Home] Remove latest HCB announcement

Luke-Oldenburg:
## Summary of the problem

I think this is confusing people more than it is driving users to the announcements feature due to frequently having outdated info and irrelevant info to adults.


manuthecoder (approved):
LGTM!

garyhtou (approved):

garyhtou:
We're keeping the view code but removing the instance variable because that has the same effect. In the future, if we want to add this back, we just need to set the instance variable.

PR #12080, opened 2025-11-11
Fix currency code reference in Wise transfer message

YodaLightsabr:
## Summary of the problem

We were calling a `Currency` method on a currency string

## Describe your changes

Fix the currency code reference


garyhtou (approved):
nice!

PR #12094, opened 2025-11-12
Move contracts to a new `Contract` model

polypixeldev:
## Summary of the problem

Currently, contracts are in an `OrganizerPosition::Contract` model and tightly coupled to `OrganizerPositionInvite`. With the new application flow I'm working on, we'll need contracts to be abstracted more to support a source from either an `Event::Application` or `OrganizerPositionInvite`, since `OrganizerPositionInvite` will not exist yet when the applicant is signing the contract during the application flow.

## Describe your changes

Primarily, this PR adds a new `Contract` model that is mostly the same as `OrganizerPosition::Contract`, with a couple differences:
- Has a polymorphic relationship `Contractable` instead of a relationship directly to `OrganizerPositionInvite`. Only `OrganizerPositionInvite` implements this right now, but in the applications PR `Event::Application` will also implement it.
- A `type` column instead of a `purpose` column. The `purpose` column was added in the original implementation of `OrganizerPosition::Contract`, but it is not currently being used. This PR adds this back as STI, and a future PR will implement better automation for termination contracts

All code that references `OrganizerPosition::Contract` has been changed to reference `Contract` instead, including the relations on `User` and `Event`. A future PR will drop the `OrganizerPosition::Contract` model and associated table.

In the database migration that creates this new table, steps have been added that copy over current data from the `organizer_position_contracts` table and renames item types for versions stored in the PaperTrail `versions` table.

I've tested this PR using my own DocuSeal account in test mode, but it should be tested with our production DocuSeal account and a real contract before merging into production.


app/controllers/contracts_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  private
+
+  def set_opc
+    @contract = Contract.find(params[:id])
+  end
+
+  def opc_params
+    params.require(:contract).permit(:contractable_id, :contractable_type, :cosigner_email, :include_videos)
```
garyhtou: Rename opc to contract

app/models/contract.rb
```diff
@@ -0,0 +1,240 @@
...
+
+  def archive_on_docuseal!
+    docuseal_client.delete("/submissions/#{external_id}")
+  end
+
+  def one_non_void_contract
+    if contractable.contracts.where.not(aasm_state: :voided).excluding(self).any?
+      self.errors.add(:base, "source already has a contract!")
+    end
+  end
```
garyhtou, line 137: this could be private i think

app/models/organizer_position_invite.rb
```diff
@@ -206,6 +206,39 @@ def signee?
     is_signee
   end
 
+  def send_contract(cosigner_email: nil, include_videos: false)
+    Contract.create!(contractable: self, cosigner_email:, include_videos:)
+    event.set_airtable_status("Documents sent")
+    update!(is_signee: true)
+    organizer_position&.update(is_signee: true)
```
garyhtou: Wrap in transaction and move set airtable status to be last

app/models/organizer_position_invite.rb
```diff
@@ -206,6 +206,39 @@ def signee?
...
+    event.plan.contract_docuseal_template_id
+  end
+
+  def contract_user
+    user
+  end
+
+  def contract_event
+    event
+  end
```
garyhtou: I wonder if it's more performant to make this an association that shares the same FK as the existing user/event association.

maybe:
```ruby
belongs_to :contract_user, foreign_key: :user_id, class_name: "User"
```

db/migrate/20251111025217_create_contracts.rb
```diff
@@ -0,0 +1,52 @@
...
+          contractable_id: row.organizer_position_invite_id,
+          contractable_type: "OrganizerPositionInvite",
+          document_id: row.document_id,
+        }
+      end
+
+      ActiveRecord::Base.transaction { Contract.insert_all(new_rows) }
+    end
+
+    ActiveRecord::Base.connection.reset_pk_sequence!('contracts')
```
garyhtou, line 49: let's add a comment about this

app/models/organizer_position/contract.rb
sampoder, line 1: probably best to remove this fyi

polypixeldev: Leaving this so that we can use this model and the activerecord methods in the database migration script - a separate PR will remove this

sampoder: sgtm

app/models/organizer_position.rb
```diff
@@ -35,6 +38,7 @@ class OrganizerPosition < ApplicationRecord
 
   belongs_to :user
   belongs_to :event
+  belongs_to :contract, optional: true
```
sampoder: maybe clarify the type of contract? because we might want to put termination on this

sampoder (approved):
lgtm! minus comment on type of contract

garyhtou (commented):
Good schema!

app/models/organizer_position.rb
garyhtou: +1 on this

garyhtou (approved):

PR #12103, opened 2025-11-13
Try fixing puma reporting to appsignal by loading credentials earlier

albertchae:
## Summary of the problem

HCB is not reporting puma to appsignal

## Describe your changes

This might be lighter weight than a full preload_app or require "rails"


albertchae:
We might be able to remove this after upgrading to Puma 7

garyhtou (approved):

PR #12106, opened 2025-11-13
[Flipper] Always render unlisted features for admins

Luke-Oldenburg:
## Summary of the problem

This way admins don't need to go into the Flipper UI to enable them.

### DIFF
--- a/app/views/events/settings/_features.html.erb
+++ b/app/views/events/settings/_features.html.erb
@@ -1,13 +1,15 @@
 <%# locals: (event:) %>
 
-<%= render partial: "features/preview", locals: {
-      id: "home-page-redesign",
-      classes: ["feature-home-page-redesign"],
-      feature_flag: :event_home_page_redesign_2024_09_21,
-      name: "Home Page Redesign",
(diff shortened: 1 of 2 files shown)


manuthecoder (approved):
lgtm!

app/views/events/settings/_features.html.erb
```diff
@@ -1,13 +1,15 @@
...
 
-<%= render partial: "features/preview", locals: {
-      id: "home-page-redesign",
-      classes: ["feature-home-page-redesign"],
-      feature_flag: :event_home_page_redesign_2024_09_21,
-      name: "Home Page Redesign",
-      description: "A fresh new look for your event's home page!",
-      event:
-    } if Flipper.enabled?(:event_home_page_redesign_2024_09_21, event) %>
+<% admin_tool("mb3") do %>
```
sampoder: use `admin_tool_if` so we still show it if it is enabled. so `admin_tool_if` if it is not enabled.

sampoder (approved):
amazing!

PR #12107, opened 2025-11-14
Try loading credentials before appsignal plugin for puma.rb

albertchae:
This is contrary to the order suggested in https://docs.appsignal.com/ruby/integrations/puma.html#secrets
but seems logical. Also pre Rails 7, `preload_app!` happened before adding `plugin :appsignal`


garyhtou (approved):

albertchae:
This broke prod, so the order does matter

PR #12113, opened 2025-11-15
[Card Grant] Customizable invite message

### LINKED ISSUE #12112: [Card Grant] Allow for per-card grant `invite_message`
At the moment, `invite_message` is only set on the Card Grant Settings. This means that all card grants in that org share the same invite message.

However, some API Card Grant Issuers would like to send custom `invite_message`s per card grant.

So, let's add an `invite_message` to card grants and make sure we accept it via the v4 API.

Luke-Oldenburg:
## Summary of the problem

Tested in UI and API.
closes #12112

### DIFF
--- a/app/views/card_grants/_create_form.html.erb
+++ b/app/views/card_grants/_create_form.html.erb
@@ -24,7 +24,7 @@
       This message is included in emails sent to card grant recipients when invited to activate a card grant.
     </p>
 
-    <%= form.text_area :invite_message, placeholder: "It's pizza time!", value: @event.card_grant_setting.invite_message.presence, disabled: %>
+    <%= form.text_area :invite_message, placeholder: "It's pizza time!", value: @event&.card_grant_setting&.invite_message, disabled: %>
     <p class="h5 muted mt0 mb1">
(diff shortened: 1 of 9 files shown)


app/views/card_grants/_create_form.html.erb
```diff
@@ -18,6 +18,20 @@
     </div>
   </div>
 
+  <div class="field">
+    <%= form.label :invite_message, "Invitation message" %>
+    <p class="h5 muted m-0">
+      This message is included in emails sent to card grant recipients when invited to activate a card grant.
```
garyhtou: I'm editing this because this field is for the ONE card grant they're currently creating, rather than a generic setting.
```suggestion
      This message is included in the card grant invitation email sent to the recipient.
```

app/views/card_grants/_create_form.html.erb
```diff
@@ -18,6 +18,20 @@
...
+    </p>
+
+    <%= form.text_area :invite_message, placeholder: "It's pizza time!", value: @event&.card_grant_setting&.invite_message, disabled: %>
```
garyhtou, line 27: Very nice pre-filling!

garyhtou (approved):

garyhtou:
hmmm just sitting here waiting for eslint...
image

PR #12131, opened 2025-11-16
[Donation Form] Enhance email field description

anishanne:
Updated the email field description to include event name and communication details.

### DIFF
--- a/app/views/donations/_donation_form.html.erb
+++ b/app/views/donations/_donation_form.html.erb
@@ -24,7 +24,7 @@
 
     <div class="field mb2">
       <%= form.label :email, "Your email" %>
-      <span class="muted block h5">We’ll send <%= @monthly ? "receipts" : "a receipt" %> here for your tax records.</span>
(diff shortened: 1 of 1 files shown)


app/views/donations/_donation_form.html.erb
```diff
@@ -24,7 +24,7 @@
 
     <div class="field mb2">
       <%= form.label :email, "Your email" %>
-      <span class="muted block h5">We’ll send <%= @monthly ? "receipts" : "a receipt" %> here for your tax records.</span>
+      <span class="muted block h5"><% @event.name %> will send <%= @monthly ? "receipts" : "a receipt" %> here for your tax records, and may use it to communicate with you.</span>
```
garyhtou:
```suggestion
      <span class="muted block h5"><%= @event.name %> will send <%= @monthly ? "receipts" : "a receipt" %> here for your tax records, and may use it to communicate with you.</span>
```

garyhtou (approved):

PR #12132, opened 2025-11-16
Use StrongMigrations.alphabetize_schema to minimize git churn

albertchae:
This will be available from Rails once we are on 8.1, so we can remove it then

https://github.com/ankane/strong_migrations?tab=readme-ov-file#schema-sanity

There are valid reasons to use non alphabetical schema ordering, but
it shouldn't matter enough in development for HCB


Luke-Oldenburg (approved):
lgtm!

PR #12140, opened 2025-11-17
[Seedfile] Add more data and improve generation process

Luke-Oldenburg:
## Summary of the problem

The user did not work by default anymore.


sampoder (approved):
lgtm

PR #12141, opened 2025-11-17
[v4] Add IP Address to API Access Tokens

Mohamad-Mortada:
## Summary of the problem
Currently in security page we don't show where that token was created and in context of Mobile app we would like to see where we logged in (or where token was refreshed) 

## Describe your changes
I added an IP address field to the API Token model then doorkeeper will update the token object after a successful response with the remote IP. Added some model functions to get lat and long for security page map


config/initializers/doorkeeper.rb
```diff
@@ -366,9 +366,12 @@
   #   puts "BEFORE HOOK FIRED! #{request}"
   # end
   #
-  # after_successful_strategy_response do |request, response|
-  #   puts "AFTER HOOK FIRED! #{request}, #{response}"
-  # end
+  after_successful_strategy_response do |request, response|
```
sampoder: woah this is cool

manuthecoder (approved):

manuthecoder (approved):

config/initializers/doorkeeper.rb
```diff
@@ -458,3 +462,10 @@
   #
   # realm "Doorkeeper"
 end
+
+# Captures request IP on token creation
+Rails.application.config.to_prepare do
+  Doorkeeper::TokensController.prepend_before_action(
+    -> { Current.request_ip = request.remote_ip }, only: :create
+  )
+end
```
garyhtou: Might be best to make this a `before_action` in `ApplicationController` so we get `Current.request_ip` set for all requests. By doing so, Doorkeeper should also see it set because it [inherits from ApplicationController](https://github.com/hackclub/hcb/blob/6ebe1901c214a4d80be727b4fea71d1223bbfbbe/config/initializers/doorkeeper.rb#L6-L5).

app/views/users/_oauth_authorization.erb
```diff
@@ -49,7 +49,22 @@
       </tbody>
     </table>
   <% end %>
-  <p class="regular italic muted text-sm m-0 mt-1">
-    <%= authorization.authorization_count == 1 ? "Authorized" : "Last authorized" %> <%= local_time_ago authorization.created_at %>
-  </p>
+  <% latest_token_with_ip = authorization.tokens.select { |t| t.ip_address.present? }.max_by(&:created_at) %>
```
garyhtou: Let's keep it simple and just grab the latest token regardless of whether it has an IP. This is because in the near future (once currently issued tokens expire), all tokens will have an IP.

garyhtou (approved):

garyhtou:
Reversible migration is failing due to a migration added by another PR

PR #12142, opened 2025-11-17
[Transfers] Update quiz to include Wise

### LINKED ISSUE #12129: [Transfers] Update quiz to include Wise

scooterthedev:
## Summary of the problem

closes https://github.com/hackclub/hcb/issues/12129 Since Wise was introducted into HCB the quiz has not been update for it. This PR fixes that.

## Describe your changes

Added a question to the quiz that if: Your recipient in the USA, and you transferring under $500 you should use wise else use int wire.


garyhtou:
@manuthecoder, would you like to review this PR?

app/javascript/controllers/transfer_form_controller.js
```diff
@@ -71,6 +79,11 @@ export default class extends Controller {
 
       this.yesClickHandler = () => this.renderQuestion(question.yes)
       this.noClickHandler = () => this.renderQuestion(question.no)
+    } else if (payload.question) {
+      this.questionTarget.innerHTML = payload.question
+
+      this.yesClickHandler = () => this.renderQuestion(payload.yes)
+      this.noClickHandler = () => this.renderQuestion(payload.no)
```
Luke-Oldenburg, line 86: What's this for?

Luke-Oldenburg (commented):
This looks pretty good although I'm hesitant to have the question be regarding an amount. There are still scenarios where the transaction is over $500 and it makes sense to use Wise and also some countries where Wise is not an option such as India.

app/javascript/controllers/transfer_form_controller.js
scooterthedev: it's to be able to set the text in the question UI. I tried it without these fews lines and everything broke.

scooterthedev: oh wait.. you may be right lol. Let me test locally. It prob shouldn't be here 3 times.

scooterthedev: Testing locally without it, and it just dies at undefined with this missing.

YodaLightsabr:
Before suggesting Wise or wire transfers to users, we should also factor in the list of supported currencies for wires and the list of supported currencies for Wise.

manuthecoder (approved):

manuthecoder (changes requested):
see comments above

manu-gurudath:
Closing this in favor of #12729 since this PR is not active

PR #12144, opened 2025-11-17
Add locks to `Reimbursement::PayoutHoldingService`

sampoder:
Recently we've been double-sending these. This adds a lock + checks the status for each item in the loop (after obtaining the lock) to prevent double sending.

### DIFF
--- a/app/services/reimbursement/payout_holding_service/nightly.rb
+++ b/app/services/reimbursement/payout_holding_service/nightly.rb
@@ -6,122 +6,125 @@ module Reimbursement
       def run
         clearinghouse = Event.find_by(id: EventMappingEngine::EventIds::REIMBURSEMENT_CLEARING)
         Reimbursement::PayoutHolding.settled.find_each(batch_size: 100) do |payout_holding|
+          payout_holding.with_lock.do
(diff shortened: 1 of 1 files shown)


garyhtou (approved):

garyhtou:
Fix lint and then we're good to go!

app/services/reimbursement/payout_holding_service/nightly.rb
```diff
@@ -6,125 +6,127 @@ class Nightly
       def run
         clearinghouse = Event.find_by(id: EventMappingEngine::EventIds::REIMBURSEMENT_CLEARING)
         Reimbursement::PayoutHolding.settled.find_each(batch_size: 100) do |payout_holding|
+          payout_holding.with_lock.do
```
garyhtou:
```suggestion
          payout_holding.with_lock do
```

garyhtou: You may have a mismatched `end` after this fix. Feel free to dismiss my change request and merge after fixing

garyhtou (changes requested):

PR #12152, opened 2025-11-18
Disable send grant button according to card grant policy

polypixeldev:
## Summary of the problem

Currently, the send grant button on the new card grants overview page checks the transfer create policy for if it is disabled. However, this policy differs from the card grant create policy because it does not check if the event plan has card grants enabled.

## Describe your changes

Changes the disabled attribute on the send grant button to check the card grant policy instead of the transfer create policy.


Luke-Oldenburg (approved):
lgtm!

sampoder (approved):
cool

PR #12160, opened 2025-11-19
Only pass-through query parameters for signed in users

sampoder:
This is clogging up our requests because these are not cached. Non-signed in users can still filter through the UI but not through query parameters. We never document / advertise filtering through query params so this is largely bots. I think this regression is fine.

### DIFF
--- a/app/views/events/transactions.html.erb
+++ b/app/views/events/transactions.html.erb
@@ -141,7 +141,7 @@
 
   <%= render "pinned_transactions" %>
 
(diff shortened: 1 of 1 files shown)


sampoder:
@claude can you review?

sampoder:
Turns out people are directly requesting to `/ledger` so this might not help :/

Luke-Oldenburg (commented):
Do we have a helper function to check if the user is signed in?

PR #12162, opened 2025-11-19
Only show user avatars if `organizer_signed_in?`

sampoder:
I think this will reduce the amount of requests we get as each image is a new request


sampoder:
@claude can you review?

Luke-Oldenburg (commented):
I think this is good to improve performance for now.

PR #12168, opened 2025-11-21
Use Turbo Stream to update comments list on deletion

### LINKED ISSUE #11601: Deleting the last comment doesn't render the blankslate
https://github.com/user-attachments/assets/64358750-11b1-4ab1-a1cc-97fd70792c21

gerardoacuna:
When a comment is destroyed, replace the entire comments section with the updated list partial instead of just removing the DOM element. This ensures the UI properly reflects the current state and handles edge cases like empty state rendering.

Fixes #11601

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -48,14 +48,21 @@ class CommentsController < ApplicationController
 
   def destroy
     authorize @comment
+    commentable = @comment.commentable
     @comment.destroy!
 
     respond_to do |format|
       format.turbo_stream do
-        render turbo_stream: turbo_stream.remove(@comment)
+        render turbo_stream: turbo_stream.replace(
+          "comments",
+          partial: "comments/list",
+          locals: {
+            comments: commentable.comments,
+            show_blankslate: commentable.comments.empty? }
+        )
       end
 
-      format.any { redirect_back_or_to @comment.commentable }
+      format.any { redirect_back_or_to commentable }
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/views/comments/_list.html.erb
```diff
@@ -1,8 +1,7 @@
-<section class="comments">
```
Luke-Oldenburg, line 1: Are we losing styling by switching this from class to id or was it a typo?

Luke-Oldenburg:
Thank you for this!

app/views/comments/_list.html.erb
gerardoacuna: The ID is needed because it's referenced by the controller and I didn't notice any style changes. I'll leave both just to be sure.

YodaLightsabr (approved):

Luke-Oldenburg (approved):
lgtm!

app/controllers/comments_controller.rb
```diff
@@ -48,14 +48,21 @@ def update
...
 
     respond_to do |format|
       format.turbo_stream do
-        render turbo_stream: turbo_stream.remove(@comment)
+        render turbo_stream: turbo_stream.replace(
+          "comments",
+          partial: "comments/list",
+          locals: {
+            comments: commentable.comments,
+            show_blankslate: commentable.comments.empty? }
```
Luke-Oldenburg:
```suggestion
            show_blankslate: commentable.comments.empty?
          }
```

PR #12178, opened 2025-11-22
Pass in `event` to `receipt_optional?` optionally

sampoder:
Reduces the SQL queries!

### DIFF
--- a/app/models/hcb_code.rb
+++ b/app/models/hcb_code.rb
@@ -497,7 +497,10 @@ class HcbCode < ApplicationRecord
               ")
   }
 
-  def receipt_required?
+  # we optionally take an event parameter here. this
+  # is a performance optimisation because it allows us to
+  # load the event once on the ledger and never again
+  def receipt_required?(event = event)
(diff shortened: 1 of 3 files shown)


sampoder:
@claude please review

app/models/hcb_code.rb
```diff
@@ -497,7 +497,10 @@ def ct2
               ")
   }
 
-  def receipt_required?
+  # we optionally take an event parameter here. this
+  # is a performance optimisation because it allows us to
+  # load the event once on the ledger and never again
+  def receipt_required?(event = event)
```
```suggestion
  def receipt_required?(event = self.event)
```

app/models/hcb_code.rb
```diff
@@ -511,8 +514,8 @@ def receipt_required?
     false
   end
 
-  def receipt_optional?
-    !receipt_required?
+  def receipt_optional?(event = event)
```
```suggestion
  def receipt_optional?(event = self.event)
```

Luke-Oldenburg (approved):
lgtm!

PR #12183, opened 2025-11-22
bugfix: renders are missing %=

Luke-Oldenburg:
## Summary of the problem

These weren't actually rendering.


app/views/disbursements/show.html.erb
```diff
@@ -1,6 +1,6 @@
 <% title "Transfer to #{@disbursement.event.name}" %>
 <% page_sm %>
-<% render "events/nav" %>
+<%= render "events/nav" %>
```
garyhtou, line 3: it seems like we already see event nav in prod

sampoder (approved):
lets try this - and have fun!

PR #12184, opened 2025-11-22
[Performance] Restructure `gte_member_in_events?`

Luke-Oldenburg:
- return early if not signed in
- move admin check out of or
- just rely on `role_at_least?`


sampoder (approved):
lgtm!

PR #12185, opened 2025-11-22
[DX] Improve comment style in `transactions.html.erb`

Luke-Oldenburg:
## Summary of the problem

These were causing syntax highlighting errors.


sampoder (approved):
awesome

PR #12186, opened 2025-11-22
[DX] Remove nil safety operator from event show

Luke-Oldenburg:
## Summary of the problem

This page can't render without an event.


sampoder (approved):
yes? i wonder why this was the case

PR #12187, opened 2025-11-22
[Performance] Don't render inaccessible HTML

Luke-Oldenburg:
## Summary of the problem

This slows down rendering time.


sampoder (approved):
sweet

PR #12189, opened 2025-11-23
[V4] Improve donation API

Luke-Oldenburg:
## Summary of the problem

We were incorrectly marking all donations as tax-deductible and missing other fields.

### DIFF
--- a/app/controllers/api/v4/donations_controller.rb
+++ b/app/controllers/api/v4/donations_controller.rb
@@ -10,7 +10,7 @@ module Api
       def create
         amount = params[:amount_cents]
         if params[:fee_covered] && @event.config.cover_donation_fees
-          amount = (params[:amount_cents] / (1 - @event.revenue_fee)).ceil
+          amount /= (1 - @event.revenue_fee)).ceil
         end
 
         @donation = Donation.new({
--- a/app/controllers/api/v4/donations_controller.rb
+++ b/app/controllers/api/v4/donations_controller.rb
@@ -8,8 +8,13 @@ module Api
       before_action :set_api_event, only: [:create]
 
       def create
+        amount = params[:amount_cents]
+        if params[:fee_covered] && @event.config.cover_donation_fees
+          amount = (params[:amount_cents] / (1 - @event.revenue_fee)).ceil
+        end
+
         @donation = Donation.new({
-                                   amount: params[:amount_cents],
+                                   amount:,
                                    event_id: @event.id,
                                    collected_by_id: current_user.id,
                                    in_person: true,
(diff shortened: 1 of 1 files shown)


app/controllers/api/v4/donations_controller.rb
```diff
@@ -8,14 +8,21 @@ class DonationsController < ApplicationController
...
-                                   amount: params[:amount_cents],
+                                   amount:,
                                    event_id: @event.id,
                                    collected_by_id: current_user.id,
                                    in_person: true,
                                    name: params[:name].presence,
                                    email: params[:email].presence,
-                                   tax_deductible: params[:tax_deductible] || true
+                                   anonymous: params[:anonymous].presence,
+                                   tax_deductible: params[:tax_deductible].presence,
```
sampoder: we generally assume donations are `tax_deductible`

app/controllers/api/v4/donations_controller.rb
```diff
@@ -8,14 +8,21 @@ class DonationsController < ApplicationController
       before_action :set_api_event, only: [:create]
 
       def create
+        amount = params[:amount_cents]
+        if params[:fee_covered] && @event.config.cover_donation_fees
+          amount = (params[:amount_cents] / (1 - @event.revenue_fee)).ceil
+        end
```
sampoder, line 14: maybe put this in a helper so we don't do it multiple times

Luke-Oldenburg: This code is a little different from the other controller-- I think it's fine to leave as is for now.

app/controllers/api/v4/donations_controller.rb
```diff
@@ -8,14 +8,21 @@ class DonationsController < ApplicationController
       before_action :set_api_event, only: [:create]
 
       def create
+        amount = params[:amount_cents]
+        if params[:fee_covered] && @event.config.cover_donation_fees
+          amount = (params[:amount_cents] / (1 - @event.revenue_fee)).ceil
```
YodaLightsabr:
```suggestion
          amount /= (1 - @event.revenue_fee)).ceil
```

YodaLightsabr (changes requested):

YodaLightsabr (approved):

app/controllers/api/v4/donations_controller.rb
```diff
@@ -8,14 +8,21 @@ class DonationsController < ApplicationController
       before_action :set_api_event, only: [:create]
 
       def create
+        amount = params[:amount_cents]
+        if params[:fee_covered] && @event.config.cover_donation_fees
+          amount /= (1 - @event.revenue_fee)).ceil
```
Luke-Oldenburg:
```suggestion
          amount /= (1 - @event.revenue_fee).ceil
```

PR #12206, opened 2025-11-23
[v4] Do not force filter params on org transactions

Mohamad-Mortada:
## Summary of the problem
Filter params are required and we dont properly pass in filters

## Describe your changes
Make em optional


sampoder (approved):
oh this is cool

PR #12213, opened 2025-11-23
Fix admin mentions

YodaLightsabr:
## Summary of the problem

Admin mentions create `<a>` elements, which messes with the page layouts if you have nested links.

## Describe your changes

Change `<a>` to `<div>` and use a context menu event handler


manuthecoder (approved):
LGTM!

PR #12214, opened 2025-11-23
Fix login flash position

### LINKED ISSUE #12078: [Login] Left align flash
The flash on the login page appears misaligned because it's centered and leaves a small gap on the sides.

Image

Maybe we should consider left-aligning when it's split screen like this?

Luke-Oldenburg:
## Summary of the problem

closes #12078

### DIFF
--- a/app/views/application/_flash.html.erb
+++ b/app/views/application/_flash.html.erb
@@ -1,10 +1,10 @@
 <%# locals: (klass: "") -%>
 
 <div id="flash-container">
-  <% flash.to_h.except("confetti", "confetti_emojis", "popover").each do |key, value| %>
+  <% flash.to_h.except("confetti", "confetti_emojis", "popover").each_with_index do |(key, value), index| %>
     <article
       role="alertdialog"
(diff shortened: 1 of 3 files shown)


app/views/application/_flash.html.erb
```diff
@@ -1,10 +1,10 @@
 <%# locals: (klass: "") -%>
 
 <div id="flash-container">
-  <% flash.to_h.except("confetti", "confetti_emojis", "popover").each do |key, value| %>
+  <% flash.to_h.except("confetti", "confetti_emojis", "popover").each_with_index do |(key, value), index| %>
     <article
       role="alertdialog"
-      class="flash flex items-center mt3 mb2 mx-auto shrink-none <%= key %> <%= klass %>"
+      class="flash flex items-center <%= "mt3" unless index == 0 %> mb2 mx-auto shrink-none <%= key %> <%= klass %>"
```
Luke-Oldenburg, line 7: There is a rare scenario where this could get the mt3 class twice but wouldn't result in any visual change.

sampoder (approved):
cool nice!

PR #12215, opened 2025-11-23
[Schema] Uniquely index `event_configurations` by `event_id`

### LINKED ISSUE #11960: [Event::Configuration] Add unique index on `event_id`
We don't want multiple `Event::Configuration`s per Event. So, let's add a unique index on `event_configurations.event_id`. A Rails validation would also be nice

```ruby
class Event::Configuration
  validates :event, uniqueness: true
end
```

https://hcb.hackclub.com/blazer/queries/998-duplicate-event-configuration

Luke-Oldenburg:
## Summary of the problem

We need to remove duplicates from production before merging. At first glance, they all appear to be identical so we could just keep the record rails is using.
closes #11960


sampoder (approved):

sampoder (approved):
fixed in prod

PR #12239, opened 2025-11-25
[Wise] Prevent changing payout method to Wise with a processing report

YodaLightsabr:
## Summary of the problem

We currently allow people to switch from international wire → Wise if they have a report being processed. This causes problems with fee expenses.

## Describe your changes

Add a validation to prevent this from happening.

Fixes #12043


sampoder (approved):

app/models/user.rb
```diff
@@ -591,6 +591,10 @@ def valid_payout_method
         errors.add(:payout_method, "is invalid. Please choose another option.")
       end
     end
+
+    if payout_method_type_changed? && payout_method.is_a?(User::PayoutMethod::WiseTransfer) && reimbursement_reports.where(aasm_state: %i[submitted reimbursement_requested reimbursement_approved]).any?
+      errors.add(:payout_method, "cannot be changed to Wise transfer with reports that are being processed. Please cancel those reports first.")
```
sampoder: they may have to reach out to ops in this case? i don't think users can cancel their own reports

garyhtou (approved):

PR #12240, opened 2025-11-25
Remove playground mode & mock data

sampoder:
Twas a good run; we can't maintain this much longer.


sampoder:
@claude please review

PR #12244, opened 2025-11-25
[Privacy Policy] Add redirect

Mohamad-Mortada:
## Summary of the problem
Add redirect to our url shortened Privacy Policy

## Describe your changes
simple redirect

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -580,6 +580,7 @@ Rails.application.routes.draw do
   get "brand_guidelines", to: redirect("branding")
   get "branding", to: "static_pages#branding"
   get "security", to: "static_pages#security"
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -580,6 +580,7 @@
   get "brand_guidelines", to: redirect("branding")
   get "branding", to: "static_pages#branding"
   get "security", to: "static_pages#security"
+  get "privacy", to: redirect("https://hack.club/hcb-privacy-policy")
```
Luke-Oldenburg, line 583: Do we want to make this a page on HCB inside of `static_pages`?

garyhtou: Not at the moment— this is temporary until we get the actual privacy policy

garyhtou (approved):

PR #12228, opened 2025-11-23
Fix admin tools

manuthecoder:
image

Fixes the admin tool menu open functionality

PR #12245, opened 2025-11-25
Fix nil aria label error

polypixeldev:
## Summary of the problem

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/462
We're getting errors because aria-label is nil

## Describe your changes

Added .present?

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -100,6 +100,8 @@ module UsersHelper
                    "#{user.name} is an admin"
                  elsif user.auditor?
                    "#{user.name} is an auditor"
+                 else
+                   user.name
                  end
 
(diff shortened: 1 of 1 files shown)


garyhtou (approved):

garyhtou (approved):

garyhtou:
based on the recent changes in https://github.com/hackclub/hcb/pull/12228, i think we want to only `+=` if `aria_label` is a string (or is present).

garyhtou:
Merging https://github.com/hackclub/hcb/pull/12246 instead. However, it would still be good to add `.present?` in case `options[:aria_label]` is an empty string— although `.present?` wouldn't fix this bug.

PR #12290, opened 2025-12-02
[Mobile] Download redirect route

Luke-Oldenburg:
This detects the current user agent to either redirect to the respective app store or the general HCB page to download.


app/controllers/static_pages_controller.rb
```diff
@@ -52,6 +52,16 @@ def admin_tools
     @transaction_volume = CanonicalTransaction.included_in_stats.sum("abs(amount_cents)")
   end
 
+  def mobile
+    if request.user_agent.match("iPhone")
+      redirect_to "https://apps.apple.com/us/app/hcb-by-hack-club/id6465424810", allow_other_host: true
+    elsif request.user_agent.match("Android")
```
Luke-Oldenburg:
```suggestion
    if request.user_agent&.match("iPhone")
      redirect_to "https://apps.apple.com/us/app/hcb-by-hack-club/id6465424810", allow_other_host: true
    elsif request.user_agent&.match("Android")
```

app/controllers/static_pages_controller.rb
```diff
@@ -52,6 +52,16 @@ def admin_tools
     @transaction_volume = CanonicalTransaction.included_in_stats.sum("abs(amount_cents)")
   end
 
+  def mobile
```
garyhtou, line 55: add `skip_before_action :signed_in_user`

garyhtou (approved):

PR #12294, opened 2025-12-02
[Namespacing] Refactor `UserSession` to `User::Session`

Luke-Oldenburg:


Luke-Oldenburg:
The solution to this error is to update all rows in `activities` to have the correct class name but I'm not sure what we can do as a temporary fix to prevent downtime.
Screenshot From 2025-12-08 00-51-39

app/models/login.rb
```diff
@@ -28,7 +28,7 @@ class Login < ApplicationRecord
   include Hashid::Rails
 
   belongs_to :user
-  belongs_to :user_session, optional: true
+  belongs_to :user_session, class_name: "User::Session", optional: true
```
garyhtou, line 31: nice catch!

garyhtou (approved):
very nice!

PR #12315, opened 2025-12-04
[V4] Add API endpoint for sub-organizations

### LINKED ISSUE #12272: [V4] Add endpoint for creating a sub-organization
It would be great to add v4 API support for creating sub-organizations. Ideally, the route is nested underneath an existing parent organization and takes in the following:

- Name of new organization
- The email of the contract signee
- (Optional) Co-signer email
- (Optional) Tags

We should return the object for the newly created organization.

YodaLightsabr:
## Summary of the problem

It's hard to automate creating sub-organizations

#12272 

## Describe your changes

Adds `/organization/:id/organizations` (`GET` and `POST`) for listing and creating sub-organizations

Fixes #12272


Mohamad-Mortada (approved):
A bit more validation in the create method would be better but lgtm.

app/controllers/api/v4/events_controller.rb
```diff
@@ -3,11 +3,51 @@
 module Api
   module V4
     class EventsController < ApplicationController
-      before_action :set_event, except: [:index]
+      before_action :set_event, except: [:index, :create]
       skip_after_action :verify_authorized, only: [:index]
 
       def index
-        @events = current_user.events.not_hidden.includes(:users).order("organizer_positions.created_at DESC")
+        if params[:event_id]
```
polypixeldev: To me it'd make more sense to have a separate action for listing subevents of a given event, instead of combining that and listing the current user's events into the index action. That would also allow you to use the `require_oauth2_scope` method for verifying scopes for listing subevents, which includes a little more logic than what's implemented here.

polypixeldev (commented):
just one thing!

app/controllers/api/v4/events_controller.rb
Luke-Oldenburg: I think it's okay to combine. It follows similar patterns such as using the same method for indexing an org's vs. a user's cards. Is it possible to update the `require_oauth2_scope` method to work better here?

app/controllers/api/v4/events_controller.rb
```diff
@@ -3,11 +3,51 @@
...
+          is_signee: true,
+          country: params[:country],
+          point_of_contact_id: parent_event.point_of_contact_id,
+          invited_by: current_user,
+          is_public: parent_event.is_public,
+          plan: parent_event.config.subevent_plan.presence,
+          risk_level: parent_event.risk_level,
+          parent_event: parent_event,
+          scoped_tags: params[:scoped_tags]
+        ).run
```
Luke-Oldenburg, line 45: Could we refactor `EventService::Create` to automatically set the other properties if `parent_event` is passed? I think that would DRY up the code some.

YodaLightsabr: I think this is fine for now—conditional defaults are a weird pattern to have, and we're only repeating it once.

app/controllers/api/v4/events_controller.rb
YodaLightsabr: The reasoning here was that it's still event index, just under a different context.

PR #12316, opened 2025-12-04
[Donation Tiers] Fix donations only being charged once

Luke-Oldenburg:
## Summary of the problem

> Hi! We've been having an issue reported with the donation tiers that have been set up where folks are not being signed up on a recurring basis, but rather as one-time donors.

## Describe your changes

Check for presence of tier id.

### DIFF
--- a/app/controllers/donations_controller.rb
+++ b/app/controllers/donations_controller.rb
@@ -80,7 +80,7 @@ class DonationsController < ApplicationController
 
     authorize @donation
 
-    @monthly = params[:monthly].present?
+    @monthly = params[:monthly].present? || params[:tier_id].present?
 
     if @monthly
       @recurring_donation = @event.recurring_donations.build(


garyhtou (approved):

garyhtou (changes requested):
actually, @manuthecoder can you look at this? i'm actualy not sure if this works lol

manuthecoder (approved):
LGTM!

app/controllers/donations_controller.rb
```diff
@@ -80,7 +80,7 @@ def start_donation
 
     authorize @donation
 
-    @monthly = params[:monthly].present?
+    @monthly = params[:monthly].present? || params[:tier_id].present?
```
garyhtou, line 83: it's prob better for check for @tier.present? instead.

garyhtou:
i'm gonna merge it as is!

PR #12318, opened 2025-12-04
Fix earmuffed mail error and include cc and bcc recipients

polypixeldev:
## Summary of the problem

We've been seeing a ton of https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2520/samples/timestamp/2025-12-04T21:24:48Z lately, which is caused by the new earmuffed mail feature. In short, the `after_action` method was constructing the mail object to see if it contained recipients, but this sometimes caused errors when `mail` is not called in the mailer and the default mail object would not be valid.

## Describe your changes

Changes earmuffed recipients to work by overriding the `mail` method itself to remove earmuffed recipients from the to, cc, and bcc headers of the mail object (currently, this only acts on the to header, but comment emails only use bcc). This prevents constructing a mail object when `mail` is never called.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -43,7 +43,7 @@ class ApplicationMailer < ActionMailer::Base
 
       all_recipients = new_to + new_cc + new_bcc
 
-      unless all_recipients.empty?
+      unless all_recipients.empty? || Rails.env.development?
         msg.to = new_to
         msg.cc = new_cc
         msg.bcc = new_bcc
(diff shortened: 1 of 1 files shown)


app/mailers/application_mailer.rb
```diff
@@ -37,6 +35,22 @@ def self.earmuffed_recipients
     end
   end
 
+  def mail(headers = {}, &block)
+    super(headers, &block).tap do |msg|
```
garyhtou:
```suggestion
  def mail(...)
    super(...).tap do |msg|
```

garyhtou (approved):
seems good to me. can u test and then merge?

garyhtou:
tysm for this

PR #12320, opened 2025-12-05
Don't require receipts on USPS operating account

24c02:


app/models/event/plan/usps_debit.rb
```diff
@@ -0,0 +1,48 @@
...
+#
+class Event
+  class Plan
+    class USPSDebit < Internal
+      def label = "HQ USPS debit"
+
+      validate do
+        unless event_id == EventMappingEngine::EventIds::HQ_USPS_OPS
+          errors.add(:event, "must be Theseus USPS Operating Account")
+        end
```
garyhtou, line 31: NICE

app/models/event/plan/usps_debit.rb
```diff
@@ -0,0 +1,48 @@
...
+#
+#  index_event_plans_on_event_id  (event_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (event_id => events.id)
+#
+class Event
+  class Plan
+    class USPSDebit < Internal
```
garyhtou: Can u actually inherit from `HackClubHQ`?

```suggestion
    class USPSDebit < HackClubHQ
```

`Internal` is for more transaction engine dark magic — ugly stuff like https://hcb.hackclub.com/999

app/models/event/plan/usps_debit.rb
```diff
@@ -0,0 +1,48 @@
...
+        unless event_id == EventMappingEngine::EventIds::HQ_USPS_OPS
+          errors.add(:event, "must be Theseus USPS Operating Account")
+        end
+      end
+
+      def description
+        "account for HQ's USPS integration - no receipts (if it gets card-locked we're federally cooked)"
+      end
+
+      def features = %w[cards account_number transfers documentation unrestricted_disbursements front_disbursements]
```
garyhtou: Since we're inheriting from `HackClubHQ`, we likely don't need to override features and just leave its behavior to whatever `HackClubHQ` does.

config/initializers/inflections.rb
```diff
@@ -22,4 +22,5 @@
   inflect.irregular "has", "have"
   inflect.irregular "hasn't", "haven't"
   inflect.acronym "HQ"
+  inflect.acronym "USPS"
```
garyhtou, line 25: nice!

app/models/event/plan/usps_debit.rb
24c02: hmm, but 7893 is on `Internal` now?

app/models/event/plan/usps_debit.rb
24c02: i'm knocking a few features off here, but i guess it doesn't really matter

sampoder (changes requested):
Hmm. This feels strange to me. I don't think we should be adding specific logic for one account in our codebase.

24c02:
closing this, i managed to get us back on ACH debit

PR #12323, opened 2025-12-05
[Referrals] Create referral links (part 1)

YodaLightsabr:
## Summary of the problem

Referral programs aren't grouped by anything, and it's difficult to see different sources of traffic for one referral program.

## Describe your changes

Add referral links to track multiple sources of traffic for a single referral program


app/models/referral/program.rb
```diff
@@ -19,9 +19,12 @@ class Program < ApplicationRecord
 
     validates :name, presence: true
 
+    belongs_to :creator, class_name: "User"
```
Luke-Oldenburg, line 31: This needs to be backed by a database column.

app/models/disbursement.rb
Luke-Oldenburg, line 1: Discard these changes please.

db/schema.rb
Luke-Oldenburg, line 1: Discard the unrelated changes please.

app/models/referral/attribution.rb
```diff
@@ -23,6 +23,7 @@
 module Referral
   class Attribution < ApplicationRecord
     belongs_to :program, class_name: "Referral::Program", foreign_key: "referral_program_id", inverse_of: :attributions
+    belongs_to :link, class_name: "Referral::Link", foreign_key: "referral_link_id", inverse_of: :attributions
```
Luke-Oldenburg, line 29: This needs to be backed by a database column.

app/models/login.rb
```diff
@@ -11,12 +11,14 @@
 #  is_reauthentication      :boolean          default(FALSE), not null
 #  created_at               :datetime         not null
 #  updated_at               :datetime         not null
+#  referral_link_id         :bigint
 #  referral_program_id      :bigint
```
Luke-Oldenburg, line 15: We should figure out why we need to store these on login.

app/models/referral/link.rb
```diff
@@ -0,0 +1,42 @@
...
+    include Hashid::Rails
+
+    belongs_to :program, class_name: "Referral::Program"
+    belongs_to :creator, class_name: "User"
+
+    has_many :attributions, dependent: :destroy, foreign_key: :referral_link_id, inverse_of: :link
+    has_many :users, -> { distinct }, through: :attributions, source: :user
+    has_many :logins, foreign_key: :referral_link_id, class_name: "Login", inverse_of: :referral_link
+
+    def value
```
Luke-Oldenburg, line 37: Are we using this function anywhere?

app/models/referral/program.rb
```diff
@@ -12,16 +12,24 @@
...
     include Hashid::Rails
 
     validates :name, presence: true
 
+    belongs_to :creator, class_name: "User"
+
     has_many :attributions, dependent: :destroy, foreign_key: :referral_program_id, inverse_of: :program
     has_many :users, -> { distinct }, through: :attributions, source: :user
     has_many :logins, foreign_key: :referral_program_id, class_name: "Login", inverse_of: :referral_program
+    has_many :links, class_name: "Link", inverse_of: :program
```
Luke-Oldenburg:
```suggestion
    has_many :links, class_name: "Referral::Link", inverse_of: :program
```

Luke-Oldenburg (approved):
lgtm

db/migrate/20251205023624_add_referral_link_id_to_logins.rb
garyhtou, line 1: Would it be better to have a Referral::Usage that has a polymorphic relationship to a `source`? This way we don't need a FK on every model

garyhtou: Or, a `Referral::Attribution` should have a polymorphic source on it directly

PR #12336, opened 2025-12-06
[CT] Make account verification deposit detection more laxed

garyhtou:
## Summary of the problem

`cirrus3` on Hacker News [reported](https://news.ycombinator.com/item?id=46168821) that we weren't hiding a account verification deposit.

We currently have logic for the typical "ACCTVERIFY" transactions. However, it appears that Expensify and a few other companies don't follow that standard.

image

^ How it should look

## Describe your changes

I'm allowing these additional substrings to be detected as an account verification transaction:
- `verify`
- `validation`
- `sdv-vrfy`
- (in addition to `acctverify`)

### DIFF
--- a/app/models/canonical_transaction.rb
+++ b/app/models/canonical_transaction.rb
@@ -350,7 +350,12 @@ class CanonicalTransaction < ApplicationRecord
   end
 
   def likely_account_verification_related?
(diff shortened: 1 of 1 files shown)


YodaLightsabr (approved):
Looks good!

PR #12352, opened 2025-12-07
[Referral] Add support for redirecting to a custom URL

### LINKED ISSUE #12329: [Referral] Add `redirect_to` to `Referral::Program`

Luke-Oldenburg:
## Summary of the problem

We want people to be able to share links that will redirect to pages such as hackclub.com/hcb or hcb.hackclub.com/mobile
closes #12329


app/controllers/referral/links_controller.rb
```diff
@@ -17,11 +17,13 @@ def show
         Rails.error.handle do
           Referral::Attribution.create!(user: current_user, program: @link.program, link: @link)
         end
+
+        redirect_to @link.program.redirect_to.presence || root_path, allow_other_host: true
```
polypixeldev, line 22: Since this is only settable by admins this should be fine - I would leave a quick comment here explaining that

app/views/admin/referral_programs.html.erb
```diff
@@ -8,6 +8,7 @@
           <div class="flex flex-col">
             <h5 class="m-0 p-0"><%= program.name %></h5>
             <small class="text-muted"><%= program.created_at.strftime("%Y-%m-%d") %> • Created by <%= program.creator.name %></small>
+            <small class="text-muted">Redirects to: <%= link_to(program.redirect_to.presence || root_url, program.redirect_to.presence || root_url) %></small>
```
polypixeldev:
```suggestion
            <small class="text-muted">Redirects to: <%= link_to(program.redirect_to.presence || root_url) %></small>
```
`link_to` will default to just the link if you only pass it one argument

app/controllers/referral/links_controller.rb
Luke-Oldenburg:
```suggestion
        # This is only configurable by admins
        redirect_to @link.program.redirect_to.presence || root_path, allow_other_host: true
```

app/models/referral/program.rb
```diff
@@ -10,6 +10,7 @@
 #  login_header_text    :string
 #  login_text_color     :string
 #  name                 :string           not null
+#  redirect_to          :string
```
garyhtou, line 13: Validate this is a real URL that's either HTTP or HTTPS

app/views/admin/referral_programs.html.erb
```diff
@@ -8,6 +8,7 @@
           <div class="flex flex-col">
             <h5 class="m-0 p-0"><%= program.name %></h5>
             <small class="text-muted"><%= program.created_at.strftime("%Y-%m-%d") %> • Created by <%= program.creator.name %></small>
+            <small class="text-muted">Redirects to: <%= link_to(program.redirect_to.presence || root_url) %></small>
```
garyhtou, line 11:
```suggestion
            <small class="text-muted">Redirects to: <%= program.redirect_to.present ? link_to(program.redirect_to) : "none" %></small>
```

Luke-Oldenburg: I would prefer to show this will bring them to home.

garyhtou: okay, that makes sense

garyhtou (approved):

PR #12353, opened 2025-12-07
[Referral] Small improvements to links

Luke-Oldenburg:
- fix capitalization
- add rails uniqueness validation
- mention slug is optional in ui
- make slug nullable (default it to the hashid in an `after_create_commit` callback). This allows for us to totally eliminate the `value` method and searching by slug || hashid now that the hashid is stored in the slug
- fix bug where custom slugs dont get passed in the redirect for non-signed in users


app/models/referral/link.rb
```diff
@@ -27,15 +27,20 @@ module Referral
...
     has_many :attributions, dependent: :destroy, foreign_key: :referral_link_id, inverse_of: :link
     has_many :users, -> { distinct }, through: :attributions, source: :user
     has_many :logins, foreign_key: :referral_link_id, class_name: "Login", inverse_of: :referral_link
 
-    def value
-      slug.presence || hashid
+    private
+
+    def set_default_slug!
+      update_column(:slug, self.hashid)
```
garyhtou:
```suggestion
      update!(slug: self.hashid)
```

app/models/referral/link.rb
```diff
@@ -27,15 +27,20 @@ module Referral
...
     has_many :attributions, dependent: :destroy, foreign_key: :referral_link_id, inverse_of: :link
     has_many :users, -> { distinct }, through: :attributions, source: :user
     has_many :logins, foreign_key: :referral_link_id, class_name: "Login", inverse_of: :referral_link
 
-    def value
-      slug.presence || hashid
+    private
+
+    def set_default_slug!
+      update!(slug: self.hashid)
```
garyhtou:
```suggestion
      update!(slug: self.hashid) unless self.slug.present?
```

garyhtou (approved):

PR #12356, opened 2025-12-07
[V4 API] Refactor `EventsController#transactions` to `TransactionsController#index`

Luke-Oldenburg:
This does not affect routes. This is to make it easier to add filtering by transactions just for a specific user regardless of event easier in the future. This has been tested.
closes #12317


Mohamad-Mortada (approved):
Lgtm

PR #12358, opened 2025-12-07
[Transfers] Improve "in transit" filter

### LINKED ISSUE #12348: [Transfers] In transit filter only shows PayPal
https://hcb.hackclub.com/reimbursement-clearinghouse/transfers?filter=in_transit

Luke-Oldenburg:
The issue with checks was because the scope didn't include column statuses.

- add wise to stats
- update check enums and calculations

closes #12348


app/controllers/events_controller.rb
```diff
@@ -627,9 +627,9 @@ def transfers
...
+      deposited: @ach_transfers.deposited.sum(:amount) + @checks.deposited.sum(:amount) + @increase_checks.increase_deposited.or(@increase_checks.in_transit).sum(:amount) + @disbursements.fulfilled.pluck(:amount).sum + @paypal_transfers.deposited.sum(:amount_cents) + @wires.deposited.sum(&:usd_amount_cents) + @wise_transfers.deposited,
+      in_transit: @ach_transfers.in_transit.sum(:amount) + @checks.in_transit_or_in_transit_and_processed.sum(:amount) + @increase_checks.in_transit.sum(:amount) + @disbursements.reviewing_or_processing.sum(:amount) + @paypal_transfers.approved.or(@paypal_transfers.pending).sum(:amount_cents) + @wires.approved.or(@wires.pending).sum(&:usd_amount_cents) + @wise_transfers.approved.or(@wise_transfers.pending).or(@wise_transfers.sent),
+      canceled: @ach_transfers.rejected.sum(:amount) + @checks.canceled.sum(:amount) + @increase_checks.canceled.sum(:amount) + @disbursements.rejected.sum(:amount) + @paypal_transfers.rejected.sum(:amount_cents) + @wires.rejected.sum(&:usd_amount_cents) + @wise_transfers.rejected.or(@wise_transfers.failed)
```
sampoder: I think these needs to be arrays of amounts, not objects.

garyhtou: +1, and we'll need to juggle both `usd_amount_cents` and `quoted_usd_amount_cents` since `usd_amount_cents` is not set until it settles, i believe.

garyhtou (commented):
Notes from discussion with Luke:

- We want to add a `deposited` scope too since it seems like the transfer page is currently using `in_transit` for deposited; which is incorrect
- Refer to https://column.com/docs/checks/notifications-and-states#states-and-events to get a full list of states for our scopes.

app/models/increase_check.rb
```diff
@@ -156,8 +157,38 @@ def payment_recipient_attributes
     requires_attention: "requires_attention"
   }, prefix: :increase
 
-  enum :column_status, %w(initiated issued manual_review rejected pending_deposit pending_stop deposited stopped pending_first_return pending_second_return first_return pending_reclear recleared second_return settled returned pending_user_initiated_return user_initiated_return_submitted user_initiated_returned pending_user_initiated_return_dishonored).index_with(&:itself), prefix: :column
-  enum :column_delivery_status, %w(created mailed rendered_pdf in_transit in_local_area processed_for_delivery delivered failed rerouted returned_to_sender).index_with(&:itself), prefix: :column_delivery
+  # https://column.com/docs/checks/notifications-and-states#states-and-events
+  enum :column_status, %w[
```
Luke-Oldenburg, line 164: For some reason we had incoming check statuses in here too before even though those are handled by a different model. Am I missing something?

app/models/wise_transfer.rb
```diff
@@ -238,6 +238,10 @@ def estimated_usd_amount_cents
     @estimated_usd_amount_cents ||= WiseTransfer.generate_quote(Money.from_cents(amount_cents, currency)).cents
   end
 
+  def usd_amount_cents_or_quoted
+    @usd_amount_cents_or_quoted = usd_amount_cents || quoted_usd_amount_cents
```
Luke-Oldenburg: **note**: this can still return nil

app/models/increase_check.rb
sampoder: We just copied this from Column

app/models/wise_transfer.rb
sampoder: when we sum we should fallback to 0

Luke-Oldenburg: Is it fine to add an || 0 here?

sampoder: hmm I think I would prefer not to and just do the fallback in the sum

app/models/wise_transfer.rb
```diff
@@ -239,6 +239,10 @@ def estimated_usd_amount_cents
...
```
sampoder:
```suggestion
    @usd_amount_cents_or_quoted ||= usd_amount_cents || quoted_usd_amount_cents
```
are you trying to do memoisation?

garyhtou: I think we can go without memoization since they're both columns. we can get rid of the instance variable too

Luke-Oldenburg:
```suggestion
    usd_amount_cents || quoted_usd_amount_cents
```

garyhtou (approved):
This looks good to me. There are some very minor performance improvements we could make:

```ruby
# what we have at the moment
WiseTransfer.deposited.map(&:usd_amount_cents_or_quoted).compact.sum

# selecting ony the columns we need
WiseTransfer.deposited.select(:usd_amount_cents, :quoted_usd_amount_cents).map(&:usd_amount_cents_or_quoted).compact.sum

# computing sum in SQL
WiseTransfer.deposited.sum(“COALESE(usd_amount_cents, quoted_usd_amount_cents)”)

# using a virtual column: https://blog.saeloun.com/2022/01/25/rails-7-postgres-support-for-generated-columns/
# it's cool, but maybe slightly overkill
```

@sampoder , can you please double-check the changes made to `column_status` and `column_delivery_status` in the IncreaseCheck model?

sampoder (approved):
column changes lgtm!

PR #12371, opened 2025-12-08
[User] Disallow SMS enrollment for fresh users

garyhtou:
## Summary of the problem

We're seeing increased traffic of new spam accounts attempting SMS verification

## Describe your changes

Disallow SMS auth for brand new users


garyhtou:
@sentry review

PR #12381, opened 2025-12-10
Disable Rails session store

polypixeldev:
## Summary of the problem

Rails' session store was adding a new cookie to every request that we don't use or need.

## Describe your changes

Configures the session store to be disabled so that we don't have an unnecessary cookie

WIP: testing

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -1,0 +1,3 @@
+# frozen_string_literal: true
+
+Rails.application.config.session_store :disabled


sampoder:
I think we use this in some places

garyhtou:
> I think we use this in some places

Can you please share where? Based on research @polypixeldev and I did, we believe we don't use Rails's default session store. Instead, we've rolled our own for our auth system.

polypixeldev:
Sam's right here! We do use the session store for a few login-related things, like preference and sudo mode.

I think it makes sense to use both our custom `cookies.encrypted` store and the session store to customize the cookie expiry. With `cookies.encrypted`, we can dynamically set the expiry of the `session_token` cookie based on the user's settings and have everything else use Rails' cookie store to expire with the browser session (which can be forever since browsers usually restore sessions).

PR #12384, opened 2025-12-11
[Mobile] Add Android App Links

Mohamad-Mortada:
Set up two-way association between the Android app and the website same way we currently have /.well-know/apple-app-site-association

https://docs.expo.dev/linking/android-app-links/


Luke-Oldenburg (approved):
lgtm

PR #12387, opened 2025-12-11
[Referral] Refactor creation code

Luke-Oldenburg:
## Summary of the problem

closes #12355


app/policies/referral/link_policy.rb
```diff
@@ -2,9 +2,13 @@
 
 module Referral
   class LinkPolicy < ApplicationPolicy
-    def show
+    def show?
```
Luke-Oldenburg, line 5: Not sure how this was working before without a ? because the lack of that was causing the other policies to fail.

app/policies/referral/program_policy.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Referral
+  class ProgramPolicy < ApplicationPolicy
+    def create?
+      true
```
Luke-Oldenburg:
```suggestion
      user.admin?
```

app/policies/referral/program_policy.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Referral
+  class ProgramPolicy < ApplicationPolicy
+    def create?
+      user.admin?
```
Luke-Oldenburg:
```suggestion
      user.auditor?
```

app/controllers/referral/links_controller.rb
```diff
@@ -2,10 +2,11 @@
...
   class LinksController < ApplicationController
-    before_action :set_link, only: :show
     skip_before_action :signed_in_user, only: :show
 
     def show
+      @link = Referral::Link.find_by(slug: params[:id]).presence
+
       unless signed_in?
         skip_authorization
         return redirect_to auth_users_path(referral: @link.slug)
```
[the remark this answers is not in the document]
Luke-Oldenburg: This is fixed in #12352

manuthecoder (approved):

sampoder (approved):

app/policies/referral/program_policy.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Referral
+  class ProgramPolicy < ApplicationPolicy
+    def create?
+      user.auditor?
+    end
```
sampoder, line 7: not blocking but this doesn't make sense with the context of the role. should be admin?

PR #12391, opened 2025-12-12
Expose `source_organization` on API v3 disbursements

YodaLightsabr:
## Summary of the problem

Right now, it's not exactly clear what direction money moves on API v3 `Transfer` objects. This PR aims to make it more clear by indicating where the money comes from.

Additional context: https://hackclub.slack.com/archives/C068U0JMV19/p1765293704716499

## Describe your changes

Adds `source_organization` to the linked object type `Transfer`

### DIFF
--- a/app/api/api/entities/transfer.rb
+++ b/app/api/api/entities/transfer.rb
@@ -18,6 +18,10 @@ module Api
           ]
         }
 
+        expose_associated Organization, as: :source_organization, hide: [API_LINKED_OBJECT_TYPE, Transaction] do |obj, options|
+          obj.source_event
+        end
+
       end
 
     end


app/api/api/entities/transfer.rb
```diff
@@ -18,6 +18,10 @@ class Transfer < LinkedObjectBase
           ]
         }
 
+        expose_associated Organization, as: :source_organization, hide: [API_LINKED_OBJECT_TYPE, Transaction] do |obj, options|
```
garyhtou, line 21: Does it make sense to also expose the destination org?

Because imagine when a Transfer appears on the source org's ledger. It may be nice to get info on the destination org.

This isn't necessary for the problem we're solving here since with just the source, you can determine the direction.

garyhtou (approved):
it passes the smell test 👃

app/api/api/entities/transfer.rb
YodaLightsabr: The existing implementation already shows the destination organization as `organization`.

garyhtou: ah cause it's exposed by the parent class. nice!

PR #12397, opened 2025-12-14
Add role to Invitations API

transcental:
## Summary of the problem

Currently the only way to see what role you've been invited to in an org is through the email. If you have a v4 API app, you will be able to use `/api/v4/user/invitations` to get the role on invitations with this change

## Describe your changes

Just added `invitation.role` as a `role` key in the response :)

image


Luke-Oldenburg (approved):
lgtm

Mohamad-Mortada (approved):

PR #12400, opened 2025-12-16
[Metric] Make Sidekiq throttled queue concurrency configurable

garyhtou:
## Summary of the problem

metrics was only running with 1 thread— which means it's taking forever.

We had rails max threads at 5 (5/3 == 1; if u ask integer math)

## Describe your changes

make it decoupled from rails threads


YodaLightsabr (approved):
zoom zoom

PR #12410, opened 2025-12-18
[v4] fix render action for card grant creation

24c02:
https://github.com/hackclub/hcb/commit/166a8e163380fd3e1de82aebae5fba61300b34d8 was a breaking change - previously it implicitly rendered [create](https://github.com/hackclub/hcb/blob/main/app/views/api/v4/card_grants/create.json.jbuilder), now it renders show, which doesn't expand disbursements by default!

### DIFF
--- a/app/controllers/api/v4/card_grants_controller.rb
+++ b/app/controllers/api/v4/card_grants_controller.rb
@@ -65,7 +65,7 @@ module Api
           return
         end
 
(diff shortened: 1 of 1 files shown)


manuthecoder (approved):
LGTM!

YodaLightsabr (approved):

PR #12415, opened 2025-12-18
[v4] Add card grant transactions

Mohamad-Mortada:
## Summary of the problem
Currently on the mobile app we are only showing the stripe card transactions of grant cards. This doesn't include critical information such as top ups, withdraws, cancels and other grant card specific transactions

## Describe your changes
Add a transaction endpoint to our card grants (/api/v4/card_grants/xxxxxx/transactions) with very similar implementation as stripe card transactions


Luke-Oldenburg (approved):
lgtm!

PR #12417, opened 2025-12-18
[Org home] Clarify copy for transparency mode

garyhtou:
## Summary of the problem

The org home can be very confusing in transparency mode when there is not enough data to be displayed.

## Describe your changes

Make the copy a bit conditional based on whether you're viewing in transparency mode.

When signed in (not much has changed):

image

In transparency:
image

Also, in transparency, the "show all" button for activities is now hidden because transparency mode users don't have access.
image

Before:
image


manuthecoder (approved):

garyhtou:
We can likely do another copy pass. I'm not completely satisfied lol

PR #12423, opened 2025-12-19
Associate signed contracts with OPs on accept

polypixeldev:
## Summary of the problem

We're currently only linking signed contracts to the organizer position in the `on_contract_signed` callback on invite, but this doesn't work before accepting an invite since the organizer position doesn't exist yet. Most contracts are signed before accepting the invite, since you can't accept an invite with a pending contract.

## Describe your changes

Added linking contracts to OP in the accept method on invite, and fixed the `send_contract` method which was re-using the contract variable when it's already defined as a method on invite. This wasn't breaking anything so far, but it was performing an extra, unnecessary query.


app/models/organizer_position_invite.rb
```diff
@@ -209,18 +210,20 @@ def signee?
...
   def send_contract(cosigner_email: nil, include_videos: false)
+    fs_contract = nil
+
     ActiveRecord::Base.transaction do
-      contract = Contract::FiscalSponsorship.create!(contractable: self, include_videos:, external_template_id: event.plan.contract_docuseal_template_id, prefills: { "public_id" => event.public_id, "name" => event.name, "description" => event.airtable_record&.[]("Tell us about your event") })
-      contract.parties.create!(user:, role: :signee)
-      contract.parties.create!(external_email: cosigner_email, role: :cosigner) if cosigner_email.present?
+      fs_contract = Contract::FiscalSponsorship.create!(contractable: self, include_videos:, external_template_id: event.plan.contract_docuseal_template_id, prefills: { "public_id" => event.public_id, "name" => event.name, "description" => event.airtable_record&.[]("Tell us about your event") })
+      fs_contract.contract.parties.create!(user:, role: :signee)
+      fs_contract.contract.parties.create!(external_email: cosigner_email, role: :cosigner) if cosigner_email.present?
```
Luke-Oldenburg: `fs_contract` does not have a `contract`.

YodaLightsabr (approved):

PR #12424, opened 2025-12-19
UI3 improvements (part 3)

manuthecoder:
This PR cleans up a few UX rough edges and fixes some reliability issues across the frontend. The main focus is smoother blog menu animations, more reliable JS loading, and small UI improvements for organizer flows.

**What changed**

* Added smoother dropdown animations for the blog menu on desktop and mobile, updated the widget to use them, and slightly tweaked icon and badge sizing so everything looks cleaner
* Fixed `Intl` polyfill load order by moving it to `application.js` and removing duplicate imports, which prevents runtime errors
* Made home page components load explicitly to avoid Turbo Frame race conditions and “home is not defined” issues
* Polished organizer invite banners and forms with better border colors, especially for dark mode
* Improved spacing and icon sizing in the role selection form for better readability and accessibility


app/views/organizer_position_invites/show.html.erb
```diff
@@ -1,4 +1,4 @@
-<% title "You’ve been invited to #{@invite.event.name}" %>
+<% title "You've been invited to #{@invite.event.name}" %>
```
garyhtou:
```suggestion
<% title "You’ve been invited to #{@invite.event.name}" %>
```

app/views/organizer_position_invites/show.html.erb
```diff
@@ -13,31 +13,30 @@
...
-  <% if @invite.contract && !@invite.contract.signed? %>
-    <% if @invite.contract.party(:signee).present? %>
-      <%= link_to "View pending contract", contract_party_path(@invite.contract.party(:signee)), class: "btn bg-accent h2 m1" %>
+  <div class="flex justify-center items-center flex-wrap mt-3 gap-2">
+    <%= link_to "Reject", organizer_position_invite_reject_path(@invite), method: :post, class: "btn bg-error" %>
+    <% if @invite.contract && !@invite.contract.signed? %>
+      <% if @invite.contract.party(:signee).present? %>
+        <%= link_to "View pending contract", contract_party_path(@invite.contract.party(:signee)), class: "btn bg-accent" %>
+      <% else %>
+        <%= link_to "View pending contract", @invite.contract.signee_docuseal_url, class: "btn bg-accent" %>
```
garyhtou, line 36: This is unrelated to this PR, but @polypixeldev, can you double check whether we should be directly linking users to `signee_docuseal_url`?

garyhtou (approved):
Two things to look into:
- We don't want Account numbers and Check deposits in event nav, although many users are having a hard time finding it on the Home page. So we're temporarily adding it back. We need to find a long-term solution that's not in the nav bar
- Determine whether we need to import `intl` and home components on _**every**_ page—it slows down hcb.

app/views/organizer_position_invites/show.html.erb
polypixeldev: This was added to not break this page if parties were not backfilled for that invite - the `if` here will use the party path if it exists

PR #12432, opened 2025-12-20
Handle nil signee party on the home page

polypixeldev:
## Summary of the problem

While we shouldn't have any sent contracts that don't have an associated signee party, we seem to have missed some in the backfill (see https://hcb.hackclub.com/blazer/queries/1051-sent-contracts-without-a-signee-role). This is causing the HCB home page to 500 for anyone who has one of these contracts.

## Describe your changes

While we work on sorting out this data, this PR adds a check to make sure we don't try to render any contracts that don't have a signee party.

### DIFF
--- a/app/views/static_pages/index.html.erb
+++ b/app/views/static_pages/index.html.erb
@@ -174,11 +174,12 @@
 
     <%= render "static_pages/index/explore" %>
 
-    <% if current_user.contracts.sent.any? %>
+    <% valid_contracts = current_user.contracts.sent.select { |contract| contract.party(:signee).present? } %>
+    <% if valid_contracts.any? %>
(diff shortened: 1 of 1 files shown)


Mohamad-Mortada (approved):
lgtm

app/views/static_pages/index.html.erb
```diff
@@ -174,11 +174,12 @@
 
     <%= render "static_pages/index/explore" %>
 
-    <% if current_user.contracts.sent.any? %>
+    <% valid_contracts = current_user.contracts.sent.select { |contract| contract.party(:signee).present? } %>
```
Luke-Oldenburg, line 177: Please make this an instance variable.

PR #12434, opened 2025-12-20
[v4] Fix bad missing receipt logic

Mohamad-Mortada:
## Summary of the problem
Accidentally tried to include events when we can already do hcb_code.events


Luke-Oldenburg (approved):
lgtm

PR #12439, opened 2025-12-21
Fix nil party error on invite show page

polypixeldev:
## Summary of the problem

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2286 - another error related to some contracts still missing a signee party.

## Describe your changes

Added a check to link directly to docuseal if the signee party does not exist yet.


sampoder (approved):
cool fix merging

PR #12441, opened 2025-12-22
Require a second factor to enable 2FA

### LINKED ISSUE #12419: [User] Prevent `use_two_factor_authentication` when only 1 factor is available
In [this helpscout ticket](https://secure.helpscout.net/conversation/3150880940/50814?viewId=4555814), admins updated someone's phone number, which had put it into an unverified state.

However, since this user had 2FA enabled, it locked them out of their account because there was no second factor available for authentication.

sampoder:
Closes https://github.com/hackclub/hcb/issues/12419. Also adds a test for this!

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -50,13 +50,10 @@ RSpec.describe UsersController do
     render_views
 
     it "requires sudo mode in order to change 2fa settings" do
-      user = create(
-        :user,
-        use_two_factor_authentication: true,
-        phone_number: "+18556254225",
-        phone_number_verified: true,
-        use_sms_auth: true
-      )
+      user = create(:user, phone_number: "+18556254225")
+      user.update!(phone_number_verified: true)
+      user.update!(use_sms_auth: true)
+      user.update!(use_two_factor_authentication: true)
       Flipper.enable(:sudo_mode_2015_07_21, user)
       stub_twilio_sms_verification(phone_number: user.phone_number, code: "123456")
       sign_in(user)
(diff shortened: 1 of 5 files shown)


YodaLightsabr (approved):
Code looks good, please fix tests though

spec/controllers/users_controller_spec.rb
```diff
@@ -50,12 +50,10 @@
...
     it "requires sudo mode in order to change 2fa settings" do
-      user = create(
-        :user,
-        use_two_factor_authentication: true,
-        phone_number: "+18556254225",
-      )
+      user = create(:user, phone_number: "+18556254225")
       user.update!(phone_number_verified: true)
+      user.update!(use_sms_auth: true)
+      user.update!(use_two_factor_authentication: true)
```
Luke-Oldenburg, line 56: Can we pass these in create?

spec/controllers/users_controller_spec.rb
```diff
@@ -92,12 +90,10 @@
...
     it "does not require sudo mode unless the feature flag is enabled" do
-      user = create(
-        :user,
-        use_two_factor_authentication: true,
-        phone_number: "+18556254225",
-      )
+      user = create(:user, phone_number: "+18556254225")
       user.update!(phone_number_verified: true)
+      user.update!(use_sms_auth: true)
+      user.update!(use_two_factor_authentication: true)
```
Luke-Oldenburg, line 96: Can we pass these in create?

spec/models/user_spec.rb
```diff
@@ -232,23 +232,38 @@
...
     it "can be disabled by regular users" do
-      user = create(:user, use_two_factor_authentication: true)
+      user = create(:user, use_two_factor_authentication: true, phone_number: "+18556254225", phone_number_verified: true, use_sms_auth: true)
 
       expect(user.update(use_two_factor_authentication: false)).to eq(true)
     end
+
+    it "can be enabled with SMS auth" do
+      user = create(:user, phone_number: "+18556254225", phone_number_verified: true)
+      user.update!(use_sms_auth: true)
```
Luke-Oldenburg, line 256: Can we pass this in create?

sampoder: No because of callbacks in the `User` model, any time the phone number is changed, `use_sms_auth` becomes false.

PR #12442, opened 2025-12-22
Add a 24 hour buffer for card locking

### LINKED ISSUE #12421: [Card Locking] Add 24hr grace period
We need to add a time-based grace period to card locking.

Example problem of why the current solution doesn't work: You're trying to get 50 students onto a subway. You need to tap your card per student. It would be nice to get all 50 taps done and then worry about receipts later.

Possible solutions:
- New transactions don't count as missing a receipt (for card locking purposes) for the first 24 hrs.
- Or, when your card is supposed to lock. You get a serious warning notification that says "your cards will lock in 24 hrs. Upload your receipts by XYZ date/time". This means you have a 24 grace period AFTER you're in violation.

sampoder:
This means that cards won't be locked until 24 hours after the 10 transactions have been made. Plus we'll still send a warning. Closes https://github.com/hackclub/hcb/issues/12421.

### DIFF
--- a/app/controllers/stripe_controller.rb
+++ b/app/controllers/stripe_controller.rb
@@ -38,8 +38,9 @@ class StripeController < ActionController::Base
 
     if approved
       user = service.card.user
-      ::User::UpdateCardLockingJob.perform_later(user:)
+      ::User::UpdateCardLockingJob.set(wait: 24.hours).perform_later(user:)
       ::User::SendCardLockingNotificationJob.perform_later(user:, event: service.card.event)
+      ::User::SendCardLockingNotificationJob.set(wait: 24.hours).perform_later(user:, event: service.card.event)
     end
 
     response.set_header "Stripe-Version", "2022-08-01"
--- a/app/controllers/stripe_controller.rb
+++ b/app/controllers/stripe_controller.rb
@@ -38,7 +38,7 @@ class StripeController < ActionController::Base
 
     if approved
       user = service.card.user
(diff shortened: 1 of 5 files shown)


app/services/user_service/send_card_locking_notification.rb
```diff
@@ -9,17 +9,24 @@ def initialize(user:)
     def run
       return unless Flipper.enabled?(:card_locking_2025_06_09, @user)
 
-      count = @user.transactions_missing_receipt(since: Receipt::CARD_LOCKING_START_DATE).count
+      current_count = @user.transactions_missing_receipt(from: Receipt::CARD_LOCKING_START_DATE, to: 24.hours.ago).count
```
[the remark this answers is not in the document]
sampoder: thank you, fixed

app/controllers/stripe_controller.rb
```diff
@@ -38,8 +38,9 @@ def handle_issuing_authorization_request(event)
 
     if approved
       user = service.card.user
-      ::User::UpdateCardLockingJob.perform_later(user:)
+      ::User::UpdateCardLockingJob.set(wait: 24.hours).perform_later(user:)
```
sampoder: Might make sense to make this 24.hours + 1 minute? Not sure how quickly HCB codes get created

garyhtou: Yes, let's do that

```suggestion
      ::User::UpdateCardLockingJob.set(wait: 24.hours + 1.minute).perform_later(user:)
```

garyhtou (approved):

PR #12443, opened 2025-12-22
Make sure current user is always defined with Discord

sampoder:
This fixes those annoying messages in hcb-engr-alerts


YodaLightsabr (approved):
Thank you for fixing this

PR #12444, opened 2025-12-22
Don't allow `params[:return_to]` to return to the logins controller

### LINKED ISSUE #12418: [Login] Do not allow `return_to` to login controller to prevent infinite login loop
I have a user who's stuck in a login loop. I suspect this is because they have a `LoginController#show` url in their `return_to` param.

https://secure.helpscout.net/conversation/3150880940/50814?viewId=4555814

sampoder:
closes #12418


Luke-Oldenburg (approved):
lgtm

PR #12449, opened 2025-12-22
[Wise] Add $50,000 maximum amount

### LINKED ISSUE #12394: Wise ACH Capped at $50k
### Describe your goals

Hi! We need to abort this Wise reimbursement report. Wise has a daily $50k limit for ACH. We should also set a capped amount for Wise transfers on HCB as well and suggest them to send a Wire instead. 

### Describe the bug

Here's the link: https://ui3.hcb.hackclub.com/reimbursement/reports/B0Snv9

YodaLightsabr:
## Summary of the problem

Wise direct debits are capped at $50,000, as reported in #12394.

## Describe your changes

Fixes #12394

Adds $50,000 cap to Wise

### DIFF
--- a/app/controllers/reimbursement/reports_controller.rb
+++ b/app/controllers/reimbursement/reports_controller.rb
@@ -234,6 +234,11 @@ module Reimbursement
               return redirect_to @report
             end
 
+            unless params[:wise_total_including_fees].to_f <= 50_000.00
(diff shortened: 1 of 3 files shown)


garyhtou (approved):
We should add a warning on the submitter side too

PR #12452, opened 2025-12-22
[Wrapped] Update banner to use 2025

YodaLightsabr:
## Summary of the problem

The banner wasn't working

## Describe your changes

Now the banner works!


Luke-Oldenburg (commented):
Why don't we want this nested under `window` anymore?

YodaLightsabr:
It didn't work and required us passing the year from Ruby into JS. It's cleaner to just leave it as it was.

Luke-Oldenburg (approved):
lgtm

Mohamad-Mortada (approved):

PR #12460, opened 2025-12-23
[v4] Add reversed field to transactions

Mohamad-Mortada:
## Summary of the problem
The app currently is showing reversed transactions as declined so we need someway to check if the transaction is reversed

## Describe your changes
added a reversed field

### DIFF
--- a/app/views/api/v4/transactions/_transaction.json.jbuilder
+++ b/app/views/api/v4/transactions/_transaction.json.jbuilder
@@ -3,6 +3,7 @@
 # This partial is capable of rendering `HcbCode`, `CanonicalPendingTransaction`, and `CanonicalTransactionGrouped` instances
 
 hcb_code = tx.is_a?(HcbCode) ? tx : tx.local_hcb_code
+is_cpt = tx.is_a?(CanonicalPendingTransaction)
 amount = transaction_amount(tx, event: @event)
 
 json.id hcb_code.public_id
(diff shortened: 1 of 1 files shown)


app/views/api/v4/transactions/_transaction.json.jbuilder
```diff
@@ -3,15 +3,17 @@
 # This partial is capable of rendering `HcbCode`, `CanonicalPendingTransaction`, and `CanonicalTransactionGrouped` instances
 
 hcb_code = tx.is_a?(HcbCode) ? tx : tx.local_hcb_code
```
Luke-Oldenburg, line 5: **unrelated**: does this ternary mean that `hcb_code` is not always a `HcbCode`? Is there a better name for this?

app/views/api/v4/transactions/_transaction.json.jbuilder
```diff
@@ -3,15 +3,17 @@
...
+is_cpt = tx.is_a?(CanonicalPendingTransaction)
```
Luke-Oldenburg, line 6: This doesn't save many characters but it makes the code easier to read by having it all follow the same pattern imo.
```suggestion
is_cpt = tx.is_a?(CanonicalPendingTransaction)
is_hcb_code = tx.is_a?(HcbCode)
```

app/views/api/v4/transactions/_transaction.json.jbuilder
```diff
@@ -3,15 +3,17 @@
...
 json.id hcb_code.public_id
 json.date tx.date
 json.amount_cents amount
 json.memo hcb_code.memo(event: @event)
 json.has_custom_memo hcb_code.custom_memo.present?
-json.pending (tx.is_a?(CanonicalPendingTransaction) && tx.unsettled?) || (tx.is_a?(HcbCode) && !tx.pt&.fronted? && tx.pt&.unsettled?)
-json.declined (tx.is_a?(CanonicalPendingTransaction) && tx.declined?) || (tx.is_a?(HcbCode) && tx.pt&.declined?)
+json.pending (is_cpt && tx.unsettled?) || (tx.is_a?(HcbCode) && !tx.pt&.fronted? && tx.pt&.unsettled?)
+json.declined (is_cpt && tx.declined?) || (tx.is_a?(HcbCode) && tx.pt&.declined?)
+json.reversed (is_cpt && tx.raw_pending_stripe_transaction&.stripe_transaction&.dig("status") == "reversed") || (tx.is_a?(HcbCode) && tx.stripe_reversed_by_merchant?)
```
Luke-Oldenburg:
```suggestion
json.pending (is_cpt && tx.unsettled?) || (is_hcb_code && !tx.pt&.fronted? && tx.pt&.unsettled?)
json.declined (is_cpt && tx.declined?) || (is_hcb_code && tx.pt&.declined?)
json.reversed (is_cpt && tx.raw_pending_stripe_transaction&.stripe_transaction&.dig("status") == "reversed") || (is_hcb_code && tx.stripe_reversed_by_merchant?)
```

app/views/api/v4/transactions/_transaction.json.jbuilder
Mohamad-Mortada: > # This partial is capable of rendering `HcbCode`, `CanonicalPendingTransaction`, and `CanonicalTransactionGrouped` instances

The comment on this code explains that we aren't always going to be rendering just HCB Codes but also CPTs but from my understanding `hcb_code` is always an `HcbCode` because we are first checking if `tx` is an HCB code itself, if not and smth like a CPT then lets grab the HcbCode related to that CPT. So either way `hcb_code` is an `HcbCode`

Luke-Oldenburg: Ah I read the ternary wrong.

Luke-Oldenburg (approved):
lgtm

PR #12466, opened 2025-12-24
[v4] Index tags for organizations

Mohamad-Mortada:
## Summary of the problem
If we wanna filter by tags on mobile then we need to know what those tags are. 

## Describe your changes
added GET /api/v4/organizations/xxxxxx/tags to return a list of tags


app/controllers/api/v4/tags_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Api
+  module V4
+    class TagsController < ApplicationController
+      include SetEvent
+
+      before_action :set_api_event, only: [:index]
```
Luke-Oldenburg:
```suggestion
      before_action :set_api_event
```

app/controllers/api/v4/tags_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+      def index
+        authorize @event, :show_in_v4?
```
Luke-Oldenburg: I don't think we need to be authorizing the event here. `index?` and `show_in_v4?` do the same thing.
```suggestion
```

Luke-Oldenburg (approved):
lgtm

app/views/api/v4/tags/index.json.jbuilder
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+expand @event do
```
garyhtou: `expand @event` might not be needed

garyhtou (approved):

PR #12467, opened 2025-12-24
Revert "[V4 API] Fix undefined method 'color' for nil in `_stripe_card.json.jbuilder`"

Mohamad-Mortada:
Causing issues on mobile

Reverts hackclub/hcb#12365


Luke-Oldenburg (approved):
lgtm

PR #12468, opened 2025-12-24
Add tappable tooltips for mobile

manuthecoder:
Closes #12302 by adding clickable tooltips


app/assets/javascripts/ui.js
```diff
@@ -150,40 +149,77 @@ window.attachTooltipListener = () => {
...
+      showTooltip(trigger);
+    },
 
-      // Sync size classes
-      ["tooltipped--lg", "tooltipped--xl"].forEach(cls => {
-        if (trigger.classList.contains(cls)) tooltip.classList.add(cls);
-      });
+    mouseleave(event) {
+      const trigger = event.currentTarget;
+      // Don't hide on mouseleave if it's a clickable tooltip that's been clicked
```
Luke-Oldenburg: I think we should do this. I don't see a need for this behavior on desktop.

garyhtou: Let's make this mobile only

garyhtou (approved):

PR #12469, opened 2025-12-24
[v4] Improve invitations API

Mohamad-Mortada:
## Summary of the problem
Create doesn't work due to not being exempted from `set_invitation` and we cant index org invites and cant delete org invites

## Describe your changes
Fixed create
added destroy method
added ability to index event invites


app/controllers/api/v4/invitations_controller.rb
```diff
@@ -6,11 +6,17 @@ class InvitationsController < ApplicationController
       include SetEvent
 
       skip_after_action :verify_authorized, only: [:index]
-      before_action :set_invitation, except: [:index]
+      before_action :set_invitation, except: [:index, :create]
```
Luke-Oldenburg: I think it might make a bit more sense to use an only here.
```suggestion
      before_action :set_invitation, only: [:show, :destroy]
```

app/controllers/api/v4/invitations_controller.rb
```diff
@@ -6,11 +6,17 @@ class InvitationsController < ApplicationController
...
       before_action :set_api_event, only: [:create]
 
       def index
-        @invitations = current_user.organizer_position_invites.pending
+        if params[:event_id]
```
Luke-Oldenburg:
```suggestion
        if params[:organization_id]
```

app/controllers/api/v4/invitations_controller.rb
Luke-Oldenburg, line 1: **unrelated**: this should probably be renamed to `organizer_position_invites_controller.rb`.

app/controllers/api/v4/invitations_controller.rb
```diff
@@ -28,7 +34,7 @@ def create
           return render json: { error: "User is already an organizer" }, status: :unprocessable_entity
         end
 
-        if @event.organizer_position_invites.pending.exists?(email: params[:email])
+        if @event.organizer_position_invites.pending.joins(:user).where(users: { email: params[:email] }).exists?
           return render json: { error: "User already has a pending invitation" }, status: :unprocessable_entity
         end
```
Luke-Oldenburg: The model already has validations for `not_already_organizer` and `not_already_invited`.

app/controllers/api/v4/invitations_controller.rb
```diff
@@ -61,12 +67,22 @@ def reject
...
+        render json: { message: "Invitation successfully deleted" }, status: :ok
+      end
+
       private
 
       def set_invitation
         @invitation = authorize OrganizerPositionInvite.find_by_public_id(params[:id]) || OrganizerPositionInvite.friendly.find(params[:id])
 
-        if @invitation.cancelled? || @invitation.rejected? || @invitation.user != current_user
+        if @invitation.cancelled? || @invitation.rejected? || (@invitation.user != current_user && action_name != "destroy")
```
Luke-Oldenburg: This if statement doesn't feel right.

Luke-Oldenburg (dismissed):

app/controllers/api/v4/invitations_controller.rb
Mohamad-Mortada: I believe `event_id` is our more standard usage in the codebase

app/controllers/api/v4/invitations_controller.rb
Mohamad-Mortada: imo this provides cleaner error responses

app/controllers/api/v4/invitations_controller.rb
Luke-Oldenburg: We are migrating to `organization_id` moving forwards. Old APIs have support for `event_id` as backwards compatibility but all APIs should support `organization_id`.

app/controllers/api/v4/invitations_controller.rb
Luke-Oldenburg: I don't think it is worth the added complexity. We should be returning the model errors.

Mohamad-Mortada: Model errors don't return a JSON error message tho right? I believe it would just return a 500 or 400

Luke-Oldenburg: Model errors have a method to get the text from them.

app/controllers/api/v4/invitations_controller.rb
garyhtou: `@invitation.cancelled? || @invitation.rejected?` can stay, but the logic behind `@invitation.user != current_user && action_name != "destroy"` should be moved into Pundit

Mohamad-Mortada:
```suggestion
        if @invitation.cancelled? || @invitation.rejected?
```

app/controllers/api/v4/invitations_controller.rb
```diff
@@ -6,19 +6,24 @@ class InvitationsController < ApplicationController
...
       skip_after_action :verify_authorized, only: [:index]
-      before_action :set_invitation, except: [:index]
+      before_action :set_invitation, only: [:show, :destroy]
       before_action :set_api_event, only: [:create]
 
       def index
-        @invitations = current_user.organizer_position_invites.pending
+        if params[:organization_id]
+          set_api_event
+          authorize @event
```
Luke-Oldenburg: This will call `EventPolicy#index?`

app/controllers/api/v4/invitations_controller.rb
```diff
@@ -46,12 +51,22 @@ def reject
...
+          raise ActiveRecord::RecordInvalid.new(@invitation)
+        end
+
+        render json: { message: "Invitation successfully deleted" }, status: :ok
+      end
+
       private
 
       def set_invitation
         @invitation = authorize OrganizerPositionInvite.find_by_public_id(params[:id]) || OrganizerPositionInvite.friendly.find(params[:id])
```
Luke-Oldenburg:
```suggestion
        @invitation = OrganizerPositionInvite.find_by_public_id(params[:id]) || OrganizerPositionInvite.friendly.find(params[:id])
        authorize @invitation
```

app/controllers/api/v4/invitations_controller.rb
```diff
@@ -46,12 +51,22 @@ def reject
...
 
-        if @invitation.cancelled? || @invitation.rejected? || @invitation.user != current_user
+        if @invitation.cancelled? || @invitation.rejected?
```
Luke-Oldenburg, line 69: What's the reasoning for this change?

YodaLightsabr (approved):

app/controllers/api/v4/invitations_controller.rb
Luke-Oldenburg: For posterity: this was removed because `ActiveRecord::RecordNotFound` was the incorrect error to be raising.

PR #12477, opened 2025-12-27
Create `Column:: OperatingBalanceJob`

sampoder:
Closes https://github.com/hackclub/hcb/issues/12175

Screenshot 2025-12-28 at 12 01 55 AM

And don't worry (like me) - this is not prod data!

### DIFF
--- a/app/jobs/column/operating_balance_job.rb
+++ b/app/jobs/column/operating_balance_job.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Column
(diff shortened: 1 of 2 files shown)


Luke-Oldenburg (commented):
We should make this account for the chance that there is a transfer in progress.

sampoder:
> We should make this account for the chance that there is a transfer in progress.

Haven't been able to think of a clean way of doing this but if it starts to be a regular edge case, I'll find a way.

PR #12479, opened 2025-12-28
[v4] Add card grant expiration field

Mohamad-Mortada:
Added 'expires_on' to the card grant JSON response cuz why not?


YodaLightsabr (approved):
Looks good!

PR #12482, opened 2025-12-29
[v4] Added check creation to API

Mohamad-Mortada:
## Summary of the problem
Can't issue checks on mobile

## Describe your changes
Added index, show, and create routes for checks


sampoder (approved):
lgtm!

app/controllers/api/v4/checks_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+        )
+
+        authorize @check
+
+        if @check.amount > SudoModeHandler::THRESHOLD_CENTS
+          return render json: {
+            error: "invalid_operation",
+            messages: ["Checks above the sudo mode threshold of #{ApplicationController.helpers.render_money(SudoModeHandler::THRESHOLD_CENTS)} are not allowed via API."]
+          }, status: :bad_request
+        end
```
Luke-Oldenburg, line 49: This should be a model validation.

app/models/increase_check.rb
```diff
@@ -63,6 +63,9 @@ class IncreaseCheck < ApplicationRecord
   include PublicActivity::Model
   tracked owner: proc{ |controller, record| controller&.current_user }, event_id: proc { |controller, record| record.event.id }, only: [:create]
 
+  include PublicIdentifiable
+  set_public_id_prefix :icheck
```
Luke-Oldenburg:
```suggestion
  set_public_id_prefix :ick
```

app/controllers/api/v4/checks_controller.rb
Mohamad-Mortada: No it shouldnt, its an API only thing

PR #12483, opened 2025-12-29
[V4 API] Revoken' some tokens

24c02:
sometimes people leak their tokens! 
spending all their money shouldn't be the only thing a finder can do with them...
this PR adds a `POST /api/v4/user/revoke` endpoint for [hackclub/revoker](https://github.com/hackclub/revoker) that nukes the current token from orbit.


Mohamad-Mortada (approved):
lgtm!

PR #12490, opened 2025-12-29
Fix application error and maintenance mode icons

afonsoingles:
## Summary of the problem

The icons in application-error and maintenance mode were using the old CDN that broke :(

## Describe your changes

I changed these icons to the new cdn


garyhtou (approved):
thank you!

PR #12495, opened 2025-12-30
[Policy] Improve defaults for safety

Luke-Oldenburg:
## Summary of the problem

We should default all undefined policies to false.

### DIFF
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -29,7 +29,7 @@ class ApplicationPolicy
   end
 
   def edit?
-    false
+    update?
   end
 
   def destroy?
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -21,7 +21,7 @@ class ApplicationPolicy
   end
 
   def new?
-    false
+    create?
   end
 
   def update?
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -13,7 +13,7 @@ class ApplicationPolicy
   end
 
   def show?
-    scope.where(id: record.id).exists?
+    false
   end
 
   def create?
@@ -21,7 +21,7 @@ class ApplicationPolicy
   end
 
   def new?
-    create?
+    false
   end
 
   def update?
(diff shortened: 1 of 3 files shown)


app/policies/application_policy.rb
```diff
@@ -13,23 +13,23 @@ def index?
...
   end
 
   def update?
     false
   end
 
   def edit?
-    update?
+    false
   end
```
sampoder, line 33: We should keep these the same as they were before. They are already false.

app/policies/application_policy.rb
```diff
@@ -13,23 +13,23 @@ def index?
...
 
   def show?
-    scope.where(id: record.id).exists?
+    false
   end
```
sampoder, line 17: should at least still show it for admins

garyhtou: I believe we should keep it false by default. We should ideally require that all policies are explicitly defined and not choose to fallback to ApplicationPolicy.

There are some cases where we don't want to show records to all admins, so false is the safest default.

app/policies/application_policy.rb
garyhtou: I'm cool with keeping the old values. The old values keep the new <> create and edit <> update pages together, which I believe is largely fairly safe.

app/policies/application_policy.rb
```diff
@@ -13,23 +13,23 @@ def index?
...
 
   def update?
     false
   end
 
   def edit?
-    update?
+    false
```
Luke-Oldenburg:
```suggestion
    update?
```

app/policies/application_policy.rb
```diff
@@ -13,23 +13,23 @@ def index?
...
   end
 
   def create?
     false
   end
 
   def new?
-    create?
+    false
```
Luke-Oldenburg:
```suggestion
    create?
```

garyhtou (approved):

app/policies/application_policy.rb
sampoder: should we throw an error then? my original comment was because I was worried that we may have some of these left in production but I guess by returning false, we'll catch those pretty quickly.

PR #12500, opened 2025-12-30
[Stripe Personalization] Fix Stripe overwriting backfilled personalization IDs

Mohamad-Mortada:
## Summary of the problem
Old cards before personalization was a thing are getting their backfilled stripe personalization ID overwritten because anytime we sync from stripe (e.g freeze) it changes it to stripe's value (which is nil cuz it wasnt a thing then) 

## Describe your changes
If its before the specific date that we found and its nil we should set it to our default design which has an ID of 311

### DIFF
--- a/app/models/stripe_card.rb
+++ b/app/models/stripe_card.rb
@@ -303,7 +303,14 @@ class StripeCard < ApplicationRecord
     self.last4 = stripe_obj[:last4]
     self.stripe_status = stripe_obj[:status]
     self.card_type = stripe_obj[:type]
-    self.stripe_card_personalization_design_id = StripeCard::PersonalizationDesign.find_by(stripe_id: stripe_obj[:personalization_design])&.id
+    # On ~2024-03-26, Stripe introduced personalization designs for physical cards
+    # This resulted in older cards not having a personalization design ID.
+    # This fix checks if its an old card without a personalization design ID and sets it to the default black design ID (311).
+    if self.created_at < Time.utc(2024, 3, 26) && stripe_obj[:personalization_design].nil?
+      self.stripe_card_personalization_design_id = StripeCard::PersonalizationDesign.find_by(stripe_id: 311)&.id
+    else
+      self.stripe_card_personalization_design_id = StripeCard::PersonalizationDesign.find_by(stripe_id: stripe_obj[:personalization_design])&.id
+    end
 
     if stripe_obj[:status] == "active"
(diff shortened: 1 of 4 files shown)


app/models/stripe_card.rb
```diff
@@ -303,7 +303,14 @@ def sync_from_stripe!
     self.last4 = stripe_obj[:last4]
     self.stripe_status = stripe_obj[:status]
     self.card_type = stripe_obj[:type]
-    self.stripe_card_personalization_design_id = StripeCard::PersonalizationDesign.find_by(stripe_id: stripe_obj[:personalization_design])&.id
+    # On ~2024-03-26, Stripe introduced personalization designs for physical cards
+    # This resulted in older cards not having a personalization design ID.
+    # This fix checks if its an old card without a personalization design ID and sets it to the default black design ID (311).
+    if self.created_at < Time.utc(2024, 3, 26) && stripe_obj[:personalization_design].nil?
```
garyhtou: The first Stripe Card we found with a null personalization design (according to Stripe) was created at `2024-03-26 05:02:30.268839000 UTC +00:00`, which means this date likely needs to be `< 2024-3-27`

app/models/stripe_card.rb
```diff
@@ -303,7 +303,14 @@ def sync_from_stripe!
...
+      self.stripe_card_personalization_design_id = StripeCard::PersonalizationDesign.find_by(stripe_id: 311)&.id
```
garyhtou: Can we please store 311 as a constant?

garyhtou (changes requested):

Luke-Oldenburg (commented):
I don't think we should be updating the syncing code. The stripe card object should be identical to the data stored on Stripe including nil values. Allowing those to diverge feels inappropriate imo. I think it would be better to update the rendering code to handle nil design values.

sampoder:
Omg. So this is why this was happening.

sampoder:
> I don't think we should be updating the syncing code. The stripe card object should be identical to the data stored on Stripe including nil values. Allowing those to diverge feels inappropriate imo. I think it would be better to update the rendering code to handle nil design values.

I think this is fine because this field's content never matches Stripe because its our internal ID not Stripe's ID

sampoder (approved):

garyhtou:
Bug: This attempts to set personalization design for virtual cards too

PR #12507, opened 2025-12-31
[StripeCard] Create expiration job

Luke-Oldenburg:
## Summary of the problem

closes #12322

### DIFF
--- a/app/jobs/stripe_card/expiration_job.rb
+++ b/app/jobs/stripe_card/expiration_job.rb
@@ -4,7 +4,7 @@ class StripeCard
   class ExpirationJob < ApplicationJob
     queue_as :low
     def perform
-      StripeCard.active.or(StripeCard.frozen).find_each do |sc|
+      StripeCard.where.not(stripe_status: "canceled").find_each do |sc|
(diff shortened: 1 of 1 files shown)


app/jobs/stripe_card/expiration_job.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class StripeCard
+  class ExpirationJob < ApplicationJob
+    queue_as :low
+    def perform
+      StripeCard.active.or(StripeCard.frozen).find_each do |sc|
```
garyhtou: This will then include `inactive` (physicals that were never activated).
```suggestion
      StripeCard.where.not(stripe_status: "canceled").find_each do |sc|
```

garyhtou (commented):
Code looks good. However, before we merge this, we'll want to:
- Add UI/cards on what to do when your card is expiring
	- Issue yourself a new card (with a warning that it takes 5-7 business days to ship?)
	- Transfer subscriptions
- Nice to have: Send an email 1 month before your card expires. Containing same information as above. Give them an easy button to issue a replacement card.
- Explore benefits of replacement cards (`replacement_for`) and implement it

PR #12508, opened 2025-12-31
Optimize Transfers Page (Part 1)

Mohamad-Mortada:
## Summary of the problem
Currently we are doing very inefficient and complex logic to render transfers. https://github.com/hackclub/hcb/blob/84719b00a400e6cf37481f7a3d3283d977af07d1/app/controllers/events_controller.rb#L612-L671

## Describe your changes
This first part of the fix will create a new table called `TransferRecord` which will be a collection of all transfer types allowing easy searching, filters, and computing stats.

Before we can move onto merging the second part we will need to backfill all past transfers then Part 2 will consist of a job to monitor that they are properly synced and the actual usage of this new table in the events controller

### DIFF
--- a/app/models/ach_transfer.rb
+++ b/app/models/ach_transfer.rb
@@ -67,6 +67,8 @@ class AchTransfer < ApplicationRecord
   include Payoutable
   include Payment
   include Freezable
+  include HasTransferRecord
+  transfer_record_statuses deposited: :deposited, canceled: [:rejected, :failed]
 
   def payment_recipient_attributes
     %i[bank_name account_number routing_number]
--- a/app/models/concerns/has_transfer_record.rb
+++ b/app/models/concerns/has_transfer_record.rb
@@ -1,0 +1,120 @@
+# frozen_string_literal: true
+
+module HasTransferRecord
+  extend ActiveSupport::Concern
+
+  included do
+    class_attribute :transfer_record_config, default: {
+      deposited_states: [],
+      canceled_states: [],
+      event_method: :event,
+      name_method: nil,
+      email_method: :recipient_email,
+      amount_method: :amount
+    }
+
+    has_one :transfer_record, as: :transferable, dependent: :destroy
+
+    after_create :sync_transfer_record!
+    after_save :sync_transfer_record!, if: :needs_sync?
+  end
+
+  class_methods do
+    def transfer_record_statuses(deposited: [], canceled: [], event: :event, name: nil, email: :recipient_email, amount: :amount)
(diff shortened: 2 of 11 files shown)


sampoder (changes requested):
I don't think we should be creating a new database table for one page. If anything we can essentially use HCB code for this?

Mohamad-Mortada:
I was talking with @garyhtou about this and he suggested this path. Isn't HCB Code also inefficient too however as to access information like status you would need to access the linked transaction? The goal of this is to also use this logic in the API later on

sampoder:
Yes, but I don't think we should introduce a new model with a parallel purpose to HCB code. The idea of a HCB code is to be a general purpose model that can represent multiple types of transactions.

sampoder:
An alternative way to do this would be to actually load it as a filtered set of HCB codes and paginate that

sampoder:
An example of why I don't think we need a new `Transfer` model, 50% of the SQL queries are caused by this chunk of code:

```erb
<% if (Flipper.enabled?(:hcb_code_popovers_2023_06_16, current_user) || !organizer_signed_in?) && !d.is_a?(CardGrant) %>
  <section class="modal modal--scroll modal--popover bg-snow" data-behavior="modal" role="dialog" id="transfer_modal_<%= d.hcb_code %>" data-state-url="<%= hcb_code_url(d.hcb_code) %>" data-state-title="Transfer of <%= render_money(d.local_hcb_code.amount_cents_by_event(@event).abs) %>">
    <%= modal_header("Transfer of #{render_money(d.local_hcb_code.amount_cents_by_event(@event).abs)}", external_link: hcb_code_url(d.hcb_code)) %>
    <%= turbo_frame_tag d.local_hcb_code.public_id, src: d.local_hcb_code.popover_path, loading: :lazy do %>
      <%= render partial: "application/loading_container" %>
    <% end %>
  </section>
<% end %>
```

If you cut out any calls to `local_hcb_code` (which is fairly do-able, just use the transfer's amount and the actual HCB code stored on the model), you would be able to cut out all of those SQL queries.

sampoder:
The other 40% come from the fact that we call `.usd_amount_cents` when calculating the stats. When I cut out this and what I mentioned above, I got HQ down to ~50 SQL queries. 

We can get around the `usd_amount_cents` one by either doing some smart queries using CT and CPT or by caching these numbers. 

I would encourage us to not overcomplicate the transaction engine any more with new models and instead build on what we have (I have a proposal to add `amount_cents` to HCB code) or smartly work around issues with our current set-up.

PR #12509, opened 2026-01-01
[v4] Move error handling to application level

Mohamad-Mortada:
## Summary of the problem
Its very poor code if we need to always rescue for each operation. Its much cleaner if we handle most cases on the application level and make it rescue all errors. 

## Describe your changes
Created a concern called `ErrorHandling` which consists of many rescues statements which will handle majority of errors safely and keep accurate error messages.

### DIFF
--- a/app/services/event_service/create.rb
+++ b/app/services/event_service/create.rb
@@ -41,7 +41,8 @@ module EventService
     end
 
     def run
-      raise ArgumentError, "name required" unless @name.present?
+      raise ArgumentError, "organization name is required" unless @name.present?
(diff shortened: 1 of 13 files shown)


app/services/event_service/create.rb
```diff
@@ -41,7 +41,8 @@ def initialize(name:,
     end
 
     def run
-      raise ArgumentError, "name required" unless @name.present?
+      raise ArgumentError, "organization name is required" unless @name.present?
+      raise ArgumentError, "at least one organizer email is required" if @emails.empty?
```
Mohamad-Mortada:
```suggestion
```

garyhtou (approved):

PR #12520, opened 2026-01-03
Restore comments on OPDRs

sampoder:
Closes https://github.com/hackclub/hcb/issues/12463

### DIFF
--- a/app/views/organizer_position_deletion_requests/show.html.erb
+++ b/app/views/organizer_position_deletion_requests/show.html.erb
@@ -163,14 +163,6 @@
       (<%= signee_count_after_removal %> <%= "signee".pluralize(signee_count_after_removal) %> will remain)
     </p>
   <% end %>
-
-  <% if @opdr.organizer_position.deleted_at.nil? %>
-    <%= link_to "Delete organizer ⚠️☣☠",
-      organizer_path(@opdr.organizer_position.id),
-      method: :delete,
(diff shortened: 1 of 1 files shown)


app/views/organizer_position_deletion_requests/show.html.erb
```diff
@@ -163,14 +163,6 @@
...
-
-  <% if @opdr.organizer_position.deleted_at.nil? %>
-    <%= link_to "Delete organizer ⚠️☣☠",
-      organizer_path(@opdr.organizer_position.id),
-      method: :delete,
-      data: { confirm: "You are about to remove #{@opdr.organizer_position.user.initial_name} from #{@opdr.event.name}. This has PERMANENT consequences (e.g. canceled cards). #{"⚠️ They are the last signee for this organization! This organization will be left with NO SIGNEES! ⚠️ " if @opdr.organizer_last_signee?}Are you sure you want to do this?" },
-      class: "btn bg-error mr1" %>
-  <% else %>
-    <p><strong>Organizer is marked as deleted</strong></p>
-  <% end %>
```
garyhtou, line 175: Can you help me understand why this was removed?

sampoder: We're double rendering this button and it seemed like we didn't need the second Delete. I think this a remnant of a merge conflict

garyhtou: Cool ty!

PR #12534, opened 2026-01-04
Use locks for creating user sessions

sampoder:
Closes https://github.com/hackclub/hcb/issues/11951


sampoder:
can confirm, tested locally

PR #12535, opened 2026-01-04
Add "disabled" to "phone_raw" input

### LINKED ISSUE #11940: [User Settings] Disable phone number field if user can't edit
Same intentions as this PR (https://github.com/hackclub/hcb/pull/11895). However, disabling the phone number field is a bit more complicated bc it's a React component.

Image

sampoder:
Closes https://github.com/hackclub/hcb/issues/11940


sampoder:
the package just respects this value so its chill

PR #12541, opened 2026-01-04
Add transaction information to card locking mailers

sampoder:
Closes https://github.com/hackclub/hcb/issues/11835


sampoder:
Screenshot 2026-01-04 at 11 54 24 PM

PR #12550, opened 2026-01-05
[CardGrant] Fix accept policy

Luke-Oldenburg:
## Summary of the problem

https://hackclub.slack.com/archives/CN523HLKW/p1767538091470279

We were allowing people to accept card grants that were previously cancelled. There were 4 of these in prod

## Describe your changes

Only allow accepting a card grant (and creating the Stripe Card) if the card grant is in an active status.

### DIFF
--- a/app/policies/card_grant_policy.rb
+++ b/app/policies/card_grant_policy.rb
@@ -46,7 +46,7 @@ class CardGrantPolicy < ApplicationPolicy
   end
 
   def activate?
-    user&.admin? || (record.user == user && authorized_to_activate?)
+    user&.admin? || (record.user == user && authorized_to_activate?) && record.status == active
   end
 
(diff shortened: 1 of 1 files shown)


garyhtou (approved):

app/policies/card_grant_policy.rb
```diff
@@ -50,8 +50,7 @@ def edit_withdraw?
   end
 
   def activate?
-    user&.admin? || (cardholder? && authorized_to_activate?)
-  end
+    (user&.admin? || (cardholder? && authorized_to_activate?)) && record.active?
```
sampoder, line 53:
```suggestion
    (user&.admin? || (cardholder? && authorized_to_activate?)) && record.active?
  end
```

PR #12552, opened 2026-01-05
Handle check errors cleaner in reimbursements

sampoder:
We're over creating checks and its bad. For all other transfers, if there's an error, we reject the transfer. Not for checks though until now. Also gives better errors for engineers to debug (aka me oof).

### DIFF
--- a/app/services/reimbursement/payout_holding_service/nightly.rb
+++ b/app/services/reimbursement/payout_holding_service/nightly.rb
@@ -68,10 +68,15 @@ module Reimbursement
                   user: User.system_user
                 )
                 check.save!
-                check.send_check!
-                payout_holding.increase_check = check
-                payout_holding.save!
-                payout_holding.mark_sent!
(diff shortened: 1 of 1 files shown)


garyhtou (approved):

app/services/reimbursement/payout_holding_service/nightly.rb
```diff
@@ -68,10 +68,16 @@ def run
...
-                payout_holding.increase_check = check
-                payout_holding.save!
-                payout_holding.mark_sent!
+                begin
+                  check.send_check!
+                  payout_holding.increase_check = check
+                  payout_holding.save!
+                  payout_holding.mark_sent!
+                rescue Faraday::Error => e
+                  check.mark_rejected!
```
garyhtou, line 77: Should we also mark the payout holding as failed? (and maybe send the failed mailer)

PR #12566, opened 2026-01-06
Flag manual Column transfers being imported

### LINKED ISSUE #12564: [Column] Always flag `admt_` transfers with engineering
`admt` transfers are manually created by the Column team. HCB engineering should be always notified when these get imported because sometimes these transfers are created under the WRONG account numbers— resulting in the transaction showing up under the wrong organization

sampoder:
Closes https://github.com/hackclub/hcb/issues/12564


garyhtou (commented):
Thank you!

PR #12573, opened 2026-01-07
[Reimbursements] Account for non-USD reports in totals

YodaLightsabr:
## Summary of the problem

The current totals include non-USD values

## Describe your changes

Excludes non-USD values from the totals

image

Also adds a tooltip for applicable totals

### DIFF
--- a/app/models/reimbursement/report.rb
+++ b/app/models/reimbursement/report.rb
@@ -344,6 +344,12 @@ module Reimbursement
       Money.from_cents(0)
     end
 
+    def cached_wise_transfer_quote_amount
+      Rails.cache.fetch("cached_wise_transfer_quote_amount_#{id}", expires_in: 3.days) do
+        wise_transfer_quote_amount
+      end
+    end
+
     def wise_transfer_quote_without_fees_amount
       @wise_transfer_quote_without_fees_amount ||= WiseTransfer.generate_detailed_quote(amount)[:without_fees_usd_amount]
     rescue
--- a/app/views/events/reimbursements.html.erb
+++ b/app/views/events/reimbursements.html.erb
@@ -35,16 +35,21 @@
 
 <div class="statset mb-2">
   <div class="stat">
-    <span class="stat__label">Total</span>
-    <span class="stat__value" style="text-wrap: nowrap"><%= render_money_amount(@event.reimbursement_reports.to_calculate_total.sum(&:amount_cents)) %></span>
+    <span class="stat__label">
+      Total
+      <a class="tooltipped tooltipped--w link--ignore flex" aria-label='This figure excludes foreign currency reimbursement reports that have not yet settled to USD.' tabindex="0">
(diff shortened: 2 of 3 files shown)


app/views/events/reimbursements.html.erb
```diff
@@ -35,16 +35,21 @@
...
 <div class="statset mb-2">
   <div class="stat">
-    <span class="stat__label">Total</span>
-    <span class="stat__value" style="text-wrap: nowrap"><%= render_money_amount(@event.reimbursement_reports.to_calculate_total.sum(&:amount_cents)) %></span>
+    <span class="stat__label">
+      Total
+      <a class="tooltipped tooltipped--w link--ignore flex" aria-label='This figure excludes foreign currency reimbursement reports that have not yet settled to USD.' tabindex="0">
+        <%= inline_icon "info", width: 24 %>
+      </a>
+    </span>
```
sampoder, line 40: Are we not able to give an estimate?

YodaLightsabr: An estimate for an organization with 10 pending GBP reports, for example, would require 10 hits to the Wise API each time the reimbursements page is loaded

sampoder: hmm - maybe we cache them?

YodaLightsabr: I'm open to ideas though

sampoder: because it's just an estimate we can cache it for a long time

garyhtou (commented):
I think it's important that we include non-USD amounts too. Let's cache the Wise quotes in some way— I'm happy to help think through solutions with you.

We might need a general concept/model in the codebase to handle currency conversions (that could be used by multiple models in the future); rather than rebuildling this code for each model. That general model would also handle caching.

For the time being, this is an okay stop-gap solution!

YodaLightsabr:
Yeah, excluding non-USD amounts is suboptimal. For now though, I think this implementation is better than what we currently do, so I'm going to merge it. I'll brainstorm some caching implementations and let you know.

sampoder:
We should just do it now - it's fairly easy. Will push in 5 minutes.

sampoder:
Pushed a basic version of it; UI still needs updating and I'm not sure if I got the scopes perfectly. (also not sure why `@pending` is `@event.reimbursement_reports.pending.reimbursed` lol, I think thats an old typo we never cached)

app/models/reimbursement/report.rb
```diff
@@ -344,6 +344,12 @@ def wise_transfer_quote_amount
       Money.from_cents(0)
     end
 
+    def cached_wise_transfer_quote_amount
+      Rails.cache.fetch("cached_wise_transfer_quote_amount_#{id}", expires_in: 3.days) do
```
YodaLightsabr, line 348: @sampoder is there a way to invalidate the cached amount when the report's amount changes?

PR #12575, opened 2026-01-07
Use `resize_to_limit` to reduce the size of logos in emails

### LINKED ISSUE #12011: [Mailer] Fix massive HCB Logo using Active Storage Variants
It seems like we've been having issues with the HCB logo being massive on email clients. Zach reports, and it might have been fixed for Superhuman: https://github.com/hackclub/hcb/pull/11685

However, Helpscout still has the same issue:

Image

I think we can truly fix this by using Active Storage Variants to resize (lower the resolution) the image to the desired pixel width/height.

sampoder:
Closes https://github.com/hackclub/hcb/issues/12011

### DIFF
--- a/app/helpers/logo_helper.rb
+++ b/app/helpers/logo_helper.rb
@@ -17,7 +17,7 @@ module LogoHelper
   end
 
   def create_hcb_logo_blob
-    logo_path = Rails.root.join("public", "brand", "hcb-icon-icon-original.png")
+    logo_path = Rails.root.join("public/brand/hcb-icon-icon-original.png")
 
     ActiveStorage::Blob.create_and_upload!(
       key: HCB_LOGO_BLOB_KEY,
--- a/app/helpers/logo_helper.rb
+++ b/app/helpers/logo_helper.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module LogoHelper
+  HCB_LOGO_BLOB_KEY = "hcb-email-logo"
+
+  # Returns a URL for the HCB logo resized to the specified height using Active Storage variants.
+  def hcb_logo_variant_url(height: 80)
+    blob = find_or_create_hcb_logo_blob
+    variant = blob.variant(resize_to_limit: [nil, height])
+    Rails.application.routes.url_helpers.url_for(variant)
+  end
+
+  private
+
+  def find_or_create_hcb_logo_blob
+    ActiveStorage::Blob.find_by(key: HCB_LOGO_BLOB_KEY) || create_hcb_logo_blob
+  end
+
+  def create_hcb_logo_blob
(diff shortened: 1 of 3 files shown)


app/helpers/logo_helper.rb
```diff
@@ -0,0 +1,29 @@
...
+  def hcb_logo_variant_url(height: 80)
+    blob = find_or_create_hcb_logo_blob
+    variant = blob.variant(resize_to_limit: [nil, height])
+    Rails.application.routes.url_helpers.url_for(variant)
+  end
+
+  private
+
+  def find_or_create_hcb_logo_blob
+    ActiveStorage::Blob.find_by(key: HCB_LOGO_BLOB_KEY) || create_hcb_logo_blob
```
manuthecoder: if multiple emails are rendered in a batch, this could cause N+1 queries. maybe memoize or cache this?

app/views/layouts/mailer/_header.html.erb
```diff
@@ -9,16 +9,22 @@
...
-        style="height: 2.5rem; max-width: 80%" height="40">
+      <% if @announcement.event.logo.present? && @announcement.event.logo.variable? %>
+        <img
+          src="<%= Rails.application.routes.url_helpers.url_for(@announcement.event.logo.variant(resize_to_limit: [nil, 80])) %>"
+          style="height: 2.5rem; max-width: 80%" height="40">
+      <% else %>
+        <img
+          src="<%= hcb_logo_variant_url %>"
+          style="height: 2.5rem; max-width: 80%" height="40">
+      <% end %>
```
manu-gurudath, line 20:
```
height="40"
```

This is inconsistent with the default value of `80` in line 7 of `app/helpers/logo_helper.rb`

app/helpers/logo_helper.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def create_hcb_logo_blob
+    logo_path = Rails.root.join("public/brand/hcb-icon-icon-original.png")
+
+    ActiveStorage::Blob.create_and_upload!(
```
manu-gurudath, line 22: this function should probably handle instances where `create_and_upload!` fails

app/views/layouts/mailer/_header.html.erb
sampoder: yeah it's not perfect but this is because on retina displays a 40px image would look bad: https://sleeplessmedia.com/articles/optimizing-website-images-and-graphics-for-retina-displays

app/helpers/logo_helper.rb
sampoder: hmm - in emails, we would just retry because its a job?

manuthecoder (approved):
lgtm!

PR #12582, opened 2026-01-08
Fix mismatching Ruby version in codespaces

polypixeldev:
## Summary of the problem

Currently, we're using the Ruby devcontainer image provided by Microsoft in our devcontainer configuration for codespaces. However, patch versioned images are not provided, so when containers are rebuilt they will automatically install the latest patch version instead of what's in `.ruby-version`, causing the build to fail.

## Describe your changes

Change the codespace configuration to use the base Debian Bookworm image instead of the provided Ruby image, and install Ruby using rvm according to the version specified in `docker-compose.yml`. This does increase the initial build time as rvm compiles Ruby, which took a few minutes on a base-level codespace.

Going forward, updating the Ruby version will just require updating `.ruby-version` and `docker-compose.yml`.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,11 +1,19 @@
-ARG VARIANT="3.4-bookworm"
-FROM mcr.microsoft.com/devcontainers/ruby:${VARIANT}
+FROM mcr.microsoft.com/devcontainers/base:bookworm
+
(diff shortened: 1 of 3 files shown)


manuthecoder (approved):

sampoder (approved):
based thank you

PR #12583, opened 2026-01-08
[Stripe Card] Add safe navigation operator to `_freeze.html.erb:27`

rluodev:
## Summary of the problem

Loading this partial would fail if the card did not exist on a card grant, causing the page to break.

## Describe your changes

Add a safe navigation operator which evaluates to nil and prevents erroring from `NoMethodError` on `nil` on subsequent method calls.

### DIFF
--- a/app/views/stripe_cards/actions/_freeze.html.erb
+++ b/app/views/stripe_cards/actions/_freeze.html.erb
@@ -24,7 +24,7 @@
       <p class="text-sm opacity-60"><%= policy(stripe_card).freeze? ? freeze_tooltip : "You don't have permission to perform this action" %></p>
     </div>
   <% end %>
-<% elsif !stripe_card.canceled? %>
+<% elsif !stripe_card&.canceled? %>
   <%= link_to "#",
               method: :post,
               class: "btn btn--list-item",


Mohamad-Mortada (approved):

sampoder:
why are we rendering this partial without a stripe card?

app/views/stripe_cards/actions/_freeze.html.erb
```diff
@@ -24,7 +24,7 @@
       <p class="text-sm opacity-60"><%= policy(stripe_card).freeze? ? freeze_tooltip : "You don't have permission to perform this action" %></p>
     </div>
   <% end %>
-<% elsif !stripe_card.canceled? %>
+<% elsif !stripe_card&.canceled? %>
```
garyhtou, line 27: If `stripe_card` is `nil`, this statement would result in `true`, which renders the Freeze button for a stripe card that does not exist.

rluodev: Yeah, but the freeze button is disabled (see lines 28-36) which I think makes sense in this context

rluodev:
> why are we rendering this partial without a stripe card?

This gets rendered as part of `_actions.html.erb` for card grants, which can have a nonexistent stripe card if the grant has preauth

garyhtou:
I believe a better solution is to add a `if card_grant.present?` before rendering this partial here:

https://github.com/hackclub/hcb/blob/4db3da3006fea83f562fb435797ce6f7106e2baa/app/views/card_grants/edit_actions.html.erb#L5

Examples:

https://github.com/hackclub/hcb/blob/4db3da3006fea83f562fb435797ce6f7106e2baa/app/views/card_grants/_actions.html.erb#L6

https://github.com/hackclub/hcb/blob/4db3da3006fea83f562fb435797ce6f7106e2baa/app/views/card_grants/_actions.html.erb#L14

PR #12587, opened 2026-01-08
[Stripe Card] Remove empty card when no actions are available

### LINKED ISSUE #12586: [Stripe Card] Canceled cards have small white blank box because the `_freeze.html.erb` partial does not render any content into it.
Image

rluodev:
## Summary of the problem

Previously there was a blank area if a card was canceled

## Describe your changes

Add a blankslate so the box is not empty.

Previous:

Screenshot 2026-01-08 at 1 22 31 PM

New:

Screenshot 2026-01-08 at 1 49 21 PM

---

EDIT: we decided to just remove the card entirely if no actions are available.

### DIFF
--- a/app/views/stripe_cards/_actions.html.erb
+++ b/app/views/stripe_cards/_actions.html.erb
@@ -1,34 +1,39 @@
 <%# locals: (stripe_card:) %>
+<% if !stripe_card.canceled? %>
+  <div class="mt-3 mb-2 flex flex-col justify-center flex-wrap card card--homepage p-2">
+    <% if !stripe_card.initially_activated? %>
+      <%= render "stripe_cards/actions/activate", stripe_card: %>
+    <% else %>
+      <%= render "stripe_cards/actions/rename", stripe_card: %>
+      <%= render "stripe_cards/actions/freeze", stripe_card: %>
+    <% end %>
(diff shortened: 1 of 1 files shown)


app/views/stripe_cards/_actions.html.erb
```diff
@@ -1,34 +1,39 @@
...
+        data-cc-address-postal-code="<%= stripe_card.stripe_cardholder.stripe_billing_address_postal_code %>"
+        data-cc-address-country="<%= stripe_card.stripe_cardholder.stripe_billing_address_country %>"
+        data-organization-name="<%= stripe_card.event.name %>">
+      </onepassword-save-button>
+    <% end %>
+  </div>
+<% else %>
+  <div class="my-3 mx-0 flex flex-col justify-center flex-wrap px-2">
+    <%= blankslate "This card is canceled. No actions are available." %>
+  </div>
```
garyhtou: After chatting with @rluodev, we're dropping the blankslate in favor of not showing anything at all. This is because blankslates are helpful if the user is _**expecting**_ to see something there, but it's missing.

In this case, since the card is visible cancelled, i'm not sure that users would be expecting or looking for any actions.

```suggestion
```

garyhtou (approved):

PR #12597, opened 2026-01-10
[Receipts] Respect inline attachments in receipt mailbox

YodaLightsabr:
## Summary of the problem

#12595

> The receipt mailbox treats any attachment (inline or not) as the entire receipt, causing a simple logo to break the entire email forward.

## Describe your changes

Fixes #12595

> If someone adds an attachment, we don't want the presence of an email signature to cause the non-empty body to be converted to a PDF—we want to take the attachment. We should only take attachments with a Content-Disposition header of attachment, though, not inline.


rluodev (approved):

sampoder:
yay

PR #12598, opened 2026-01-10
[Receipts] Lazy load PDF receipts

### LINKED ISSUE #10870: Switch pdf receipts partial to use pdf.js or g viewer
When loading any reimbursement report on iOS mobile, it attempts to download every single pdf in the report.

Ian suggested we switch the receipts partial to use one of these packages.

YodaLightsabr:
## Summary of the problem

PDF receipts will auto-download in some browsers, even if put into an iframe and properly tagged with `Content-Disposition: inline`.

## Describe your changes

This PR doesn't set the `src` on the `<iframe>` until opened/expanded. This PR also fixes #10870.

### DIFF
--- a/app/assets/javascripts/ui.js
+++ b/app/assets/javascripts/ui.js
@@ -20,6 +20,9 @@ const loadModals = element => {
         ? 'turbo-frame-modal'
         : undefined,
     })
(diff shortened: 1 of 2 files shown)


manuthecoder (approved):
lgtm

PR #12599, opened 2026-01-11
Fix transparency mode bugs

manuthecoder:
This PR provides various fixes to button visibility on transparency mode. 

## Changes
* Account numbers, check deposit, payroll, and perks are now hidden from sidebar
* Hidden organization position contact emails are not clickable anymore
* Card grant actions (frost/defrost, cancel) are now hidden
* Removes duplicate `.truncate` and `.text-center` class names
* Makes cursor default on homepage transactions (instead of previously `pointer`)
* Hides "See transactions" button, since clicking on this redirects to an errored page saying "Invalid parameters. Please try again"

### DIFF
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -44,13 +44,15 @@ module EventsHelper
       selected: selected == :transactions,
     }
 
-    items << {
-      name: "Account numbers",
-      path: account_number_event_path(@event),
-      tooltip: "View account numbers",
-      icon: "hashtag",
-      selected: selected == :account_number,
-    }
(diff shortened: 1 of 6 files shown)


app/helpers/events_helper.rb
Luke-Oldenburg, line 1: I have a more comprehensive cleanup of this file in #12604. Could we use that instead please?

garyhtou: Looking at this PR, it looks like it no longer touches events helper! so we should be good here.

garyhtou (approved):

PR #12602, opened 2026-01-12
[Card Overview] Only show 18 cards

Luke-Oldenburg:
## Summary of the problem

20 / 3 is not evenly divisible resulting in a missing card at the bottom of the page when tiling 3 across and there is >= 20 cards on the org.


sampoder (approved):
lol sure, good idea

PR #12603, opened 2026-01-12
[Home] Make card grant invites rows

Luke-Oldenburg:
## Summary of the problem

There is too much content to fit 3 per row.

## Describe your changes

Use `grid-cols-1`


sampoder (approved):
okay! sgtm!

PR #12607, opened 2026-01-12
Add `joined_as_teenager` to `User`

Luke-Oldenburg:
This allows us to get a more accurate count on referrals


sampoder:
making birthday encrypted strikes again smh

app/models/user.rb
```diff
@@ -424,6 +425,10 @@ def teenager?
     age&.<=(18)
   end
 
+  def joined_as_teenager?
+    (created_at - birthday).years <= 18
```
```suggestion
    age_on(created_at) <= 18
```

app/jobs/user/update_teenager_column_job.rb
```diff
@@ -10,6 +10,11 @@ def perform
         groups.each do |value, users|
           User.where(id: users.pluck(:id)).update_all(teenager: value)
         end
+
+        groups = users_relation.select(:id, :birthday_ciphertext).group_by(&:joined_as_teenager?)
+        groups.each do |value, users|
+          User.where(id: users.pluck(:id)).update_all(joined_as_teenager?: value)
+        end
```
garyhtou: We don't need this in the job because once the value is set, it's done. It's not like `teenager,` which can change over time.
```suggestion
```

app/models/user.rb
```diff
@@ -424,6 +425,10 @@ def teenager?
     age&.<=(18)
   end
 
+  def joined_as_teenager?
+    age_on(created_at) <= 18
```
```suggestion
    age_on(created_at)&.<=(18)
```

garyhtou (approved):

garyhtou:
Once the other PR merges, there will be a conflict here. Fix that and then this will be good to merge!

garyhtou:
We'll need to backfill

app/jobs/one_time_jobs/backfill_joined_as_teenager.rb
Luke-Oldenburg, line 1: This is structured different than the `UpdateTeenagerColumnJob` but I think it should still work.

app/models/user.rb
```diff
@@ -161,6 +162,7 @@ class User < ApplicationRecord
   include HasTasks
 
   before_update { self.teenager = teenager? }
+  before_update { self.joined_as_teenager = joined_as_teenager? }
```
YodaLightsabr, line 165:
```suggestion
  before_update(if: :will_save_change_to_birthday?) { self.joined_as_teenager = joined_as_teenager? }
```
Double check my code, but this may be what you want.

app/models/user.rb
```diff
@@ -430,6 +432,10 @@ def teenager?
     age&.<=(18)
   end
 
+  def joined_as_teenager?
+    age_on(created_at)&.<=(18)
+  end
+
```
YodaLightsabr, line 438: This will end up clashing with the `joined_as_teenager?` defined by Rails pulling from the database. Should we call this something else so that we pull from the DB when possible?

Luke-Oldenburg: Gary and I discussed this and I think we're going to loop back to fix this as it also affects `teenager?`

garyhtou (approved):
@YodaLightsabr has two good points that @Luke-Oldenburg and I also discussed last time. We'll come back with another PR to fix up both `teenager?` and `joined_as_teenager` since they share the same implementation

PR #12611, opened 2026-01-12
Fix grammar on transaction decline troubleshooter

YodaLightsabr:
## Summary of the problem

We're missing a "because"

## Describe your changes

I added said "because"

Thank you to @alexdevz for reporting


sampoder (approved):
because this a good idea

PR #12612, opened 2026-01-12
Test reversible migrations in reverse order

polypixeldev:
## Summary of the problem

In https://github.com/hackclub/hcb/actions/runs/20934979891/job/60155176895?pr=12188, there are 2 new migrations created, with the 2nd depending on the 1st. The reversible migrations CI currently tests migrations in ascending order. This causes an error as when the first migration (creates table) is tested, it removes any changes to that table done by future migrations, so reverting those migrations fails.

## Describe your changes

Reverse the order in which migrations are tested. No prior migration can depend on a future migration, so this should fix this issue.


sampoder (approved):
genius - thx!

PR #12613, opened 2026-01-12
Fix yarn install in codespaces

polypixeldev:
## Summary of the problem

#12582 changed the devcontainer Dockerfile to pull from the debian base and install node manually. However, this caused an error with the yarn installation, as corepack needed to be enabled.

## Describe your changes

Enable corepack and disable the interactive download prompt when running yarn for the first time.


sampoder (approved):
interesting lgtm

PR #12614, opened 2026-01-12
Make it clear when you are updating someone else's settings

sampoder:
Screenshot 2026-01-12 at 10 52 54 PM


sampoder:
s/o to ian for the designs

PR #12616, opened 2026-01-12
Revert "Revert "[Reimbursements] Account for non-USD reports in totals""

sampoder:
Reverts hackclub/hcb#12606


app/models/reimbursement/report.rb
sampoder, line 1: fyi @YodaLightsabr

PR #12627, opened 2026-01-13
Refactor to use `ActiveSupport::CurrentAttributes`

### LINKED ISSUE #11928: [SessionHelper] Refactor to use ActiveSupport::CurrentAttributes
`ActiveSupport::CurrentAttributes` has some advantages over the methods we have now: https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html

We should really only have a couple attributes in `Current`:

- `current_session`
- `current_user`

sampoder:
Closes https://github.com/hackclub/hcb/issues/11928; very conservative approach to this.

### DIFF
--- a/app/helpers/sessions_helper.rb
+++ b/app/helpers/sessions_helper.rb
@@ -58,7 +58,7 @@ module SessionsHelper
     end
 
     user_session.save!
-    self.current_user = user
+    Current.session = user_session
 
     user_session
   end
@@ -79,10 +79,6 @@ module SessionsHelper
     signed_in? &&
       current_user&.superadmin? &&
       !current_session&.impersonated?
-  end
-
-  def current_user=(user)
-    @current_user = user
   end
 
   def organizer_signed_in?(event = @event, as: :reader)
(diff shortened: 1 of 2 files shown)


app/helpers/sessions_helper.rb
```diff
@@ -121,14 +117,13 @@ def current_user
   end
 
   def current_session
-    return @current_session if defined?(@current_session)
-
-    session_token = cookies.encrypted[:session_token]
+    Current.session ||= begin
```
garyhtou, line 120: This is a chance in favor compared to
```
return @current_session if defined?(@current_session)
```

since `||=` will reattempt on a falsey value such as `nil`.

garyhtou (approved):
I think ideally, we remove the `current_user` and `current_session` methods in favor of `Current.user` and `Current.session`. However, we can make that a separate PR if you'd like!

app/helpers/sessions_helper.rb
sampoder: This felt safer imo - for the reattempt reason

sampoder:
> I think ideally, we remove the `current_user` and `current_session` methods in favor of `Current.user` and `Current.session`.

Yeah I want to do this in stages. We also use `@current_user` in places which we need to fix

app/helpers/sessions_helper.rb
garyhtou: The potential issue might be that if there's no `current_session`, we'll reattempt the `find_by` tens of times per page load (each time we call `current_session`). It may or may not be cached by ActiveRecord, but it is a performance concern to consider.

Reattempt isn't helpful here because `cookies.encrypted[:session_token]` doesn't change within a given request, so we should only run this method once.

If we're confident that this doesn't have performance impacts, then I'm cool with keeping it.

sampoder: I think ActiveRecord would catch these and cache them but I haven't confirmed that. To replicate this situation, I'd want an expired session token and to load a public page right?

garyhtou: Yep! (or set any random value for the `session_token` cookie)

PR #12629, opened 2026-01-13
Calculate the `Current.session` once in a before action

sampoder:
Previously `current_session` was doing this calculation a bunch if `Current.session` was `nil`. Also this will let us remove `#current_session` eventually.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,7 +8,7 @@ class ApplicationController < ActionController::Base
   include SetGovernanceRequestContext
 
   protect_from_forgery
-  
+
   # set Current.session
   before_action do
     Current.session ||= begin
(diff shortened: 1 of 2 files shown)


sampoder:
Also makes https://github.com/hackclub/hcb/pull/12628 safe to merge

app/controllers/application_controller.rb
```diff
@@ -8,6 +8,17 @@ class ApplicationController < ActionController::Base
   include SetGovernanceRequestContext
 
   protect_from_forgery
+  
+  # set Current.session
+  before_action do
+    Current.session ||= begin
```
garyhtou: We can drop the ||= for a simple =

garyhtou (commented):
I think this probably should go above all the Includes since some of them may depend on Current.session

app/controllers/application_controller.rb
sampoder:
```suggestion
    Current.session = begin
```

sampoder:
> I think this probably should go above all the Includes since some of them may depend on Current.session

sgtm - fwiw - clicking around Bank everything was working as expected

PR #12641, opened 2026-01-14
Fix mobile margins

manuthecoder:
Fixes ugly mobile margin bug

image


app/views/application/_docs_header.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="bg-smoke dark:bg-darkless mb3">
+<div class="bg-smoke dark:bg-darkless mb3 px-6" style="padding-right: 0">
```
sampoder: why not do this with tailwind?

manu-gurudath: fixed!

PR #12644, opened 2026-01-15
[Card Grants] Inherit pre-authorization setting by default

YodaLightsabr:
## Summary of the problem

Previously, card grants wouldn't automatically inherit pre-authorization settings (you needed to check the box each time)

## Describe your changes

This PR now properly inherits the value when creating a new card grant


garyhtou:
can you please check Slack? lol

sampoder (changes requested):
Why not use `set_defaults`?

YodaLightsabr:
@sampoder `set_defaults` runs `before_create`, and we need these attributes to be set so that they can be configured per-grant on the create form

sampoder:
Maybe we should do something like this: https://stackoverflow.com/a/10061330? It feels odd for this to be in the controller imo

YodaLightsabr:
Should we move the invite message there, too? That way instead of a blank message defaulting to the org-level message, it prefills the org-level message and lets you customize it

sampoder:
> Should we move the invite message there, too? That way instead of a blank message defaulting to the org-level message, it prefills the org-level message and lets you customize it

yes imo - good idea

sampoder (approved):
lgtm - thank you!

PR #12658, opened 2026-01-16
[v4] Add frozen by for card grants

Mohamad-Mortada:


sampoder (approved):

Luke-Oldenburg (commented):
If this is already in the stripe card API then I think it would be better to leave it there and fetch it through the card.

Mohamad-Mortada:
thats true ill move it to stripe card

PR #12662, opened 2026-01-16
Introduce error references

### LINKED ISSUE #12570: [AppSignal] Provide error reference number to user so it's easier for us to debug
Some APM/error trackers will provide the end user with an error reference number so they can provide it to support.

It would be handy if we had that.

sampoder:
Closes https://github.com/hackclub/hcb/issues/12570. Managed to test everything except the AppSignal stuff locally.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,6 +9,8 @@ class ApplicationController < ActionController::Base
 
   protect_from_forgery
 
+  around_action :attach_error_reference
+
   # Ensure users are signed in. Create one-off exceptions to this on routes
   # that you want to be unauthenticated with skip_before_action.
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -125,4 +127,10 @@ def confetti!(emojis: nil)
     flash[:confetti_emojis] = emojis.join(",") if emojis
   end
 
+  def attach_error_reference
+    error_reference = ErrorReference.from_request_id(request.uuid)
+    Appsignal.add_tags(error_reference:) if defined?(Appsignal) && Appsignal.active?
+    yield
```
garyhtou, line 133: Since we're not doing anything after the yield, this could be rewritten as a `before_action`.

PR #12664, opened 2026-01-17
[v4] allow freezing on stripe cards as grantees

Mohamad-Mortada:
## Summary of the problem
This is causing grantees to not be able to freeze their cards as `authorize` fails on policy of `update`. So we authorize each action independently

## Describe your changes
remove inital authorize

### DIFF
--- a/app/controllers/api/v4/stripe_cards_controller.rb
+++ b/app/controllers/api/v4/stripe_cards_controller.rb
@@ -72,7 +72,7 @@ module Api
       end
 
       def update
-        @stripe_card = authorize StripeCard.find_by_public_id!(params[:id])
+        @stripe_card = StripeCard.find_by_public_id!(params[:id])
 
         if params[:status] == "frozen"
           return render json: { error: "not_authorized" }, status: :forbidden unless policy(@stripe_card).freeze?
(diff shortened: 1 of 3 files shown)


app/controllers/api/v4/stripe_cards_controller.rb
```diff
@@ -72,7 +72,9 @@ def create
       end
 
       def update
-        @stripe_card = authorize StripeCard.find_by_public_id!(params[:id])
+        @stripe_card = StripeCard.find_by_public_id!(params[:id])
+        # we authorize by show as actions are handled by their own policies (update interferes with this)
+        authorize @stripe_card, :show?
```
sampoder: this feels really unsafe.  also, where are these policy calls being made?

sampoder (changes requested):
we shouldn't be using the show? policy for updating.

app/controllers/api/v4/stripe_cards_controller.rb
Mohamad-Mortada: Check lines 80, 90, 100. Its pretty hard to fix this in a clean way if we're using one method to handle multiple actions :/

sampoder (approved):

PR #12673, opened 2026-01-19
[v4] Deprecate stripe card update method in favor of individual methods

Mohamad-Mortada:
## Summary of the problem
Start the process in which we migrate our current applications over from the old update method in favor of using specific actions such as `/api/v4/cards/crd_xxxxxx/freeze` instead of a generic `PATCH /api/v4/cards/crd_xxxxxx`. 

## Describe your changes
Moved each action's functionality to its own method and update now references that method for backwards compatibility


sampoder (approved):
thank you

PR #12678, opened 2026-01-19
[v4] Prevent ephemeral key generation on unauthorized card grants

Mohamad-Mortada:
## Summary of the problem
Currently, if card grants have preauth and they fail it they can still use the card grant by getting an ephemeral key to access stripe card details

## Describe your changes
Prevents it via policy change


manuthecoder (approved):
lgtm!

sampoder (approved):

PR #12704, opened 2026-01-22
Add file upload component

manuthecoder:
## Description
This PR introduces a stimulus-based file input component that replaces the default browser file input with a custom interface. The component displays file names alongside their corresponding icons, creating a cleaner, more intuitive user experience. It includes dynamic tooltip support for additional context and a convenient close button for removing selected files.

## Preview 
they say a picture is worth a thousand words, so here's a billion with a video

https://github.com/user-attachments/assets/5f461a7e-9790-442c-a647-a6a4275d7abc


app/views/events/settings/_details.html.erb
manu-gurudath, line 1: this is intentional because menus have weird behavior with the DOM

garyhtou (approved):

PR #12705, opened 2026-01-22
[v4] Standardize error message field

Mohamad-Mortada:
## Summary of the problem
This is the only field where we use `message` instead of an array of `messages`. We should be consistent

## Describe your changes
Changed it


Luke-Oldenburg (approved):
lgtm

PR #12188, opened 2025-11-23
📝 In-platform applications for HCB

YodaLightsabr:
## Summary of the problem

There are a number of improvements we could make to the application process. A lot of them depend on the form living in the same codebase as the rest of HCB.

## Describe your changes

We brought applications into the core of HCB, and we've added a number of improvements, including but not limited to:
* Autosave
* Follow-up emails for incomplete applications
* Integrated contract signing
* Streamlined rejection messages
* Associating applications to users and events

PR #12709, opened 2026-01-23
Add polymorphic relationship `affiliable` to affiliations

polypixeldev:
## Summary of the problem

As part of work on #12188, we're going to be adding affiliations to applications. However, since an event is not yet created at this time, we need to make affiliations have a polymorphic relationship to either `Event` or `Event::Application`. Once the application is approved and the event created, all we'll have to do is copy and change `affiliable`.

## Describe your changes

Adds a new `affiliable` polymorphic relationship on `Event::Affiliation`. The migration automatically copies over the current linked events and makes the relationship non-null.

Controllers and views that work with affiliations have been updated to use the new polymorphic relationship, and the routes for affiliations are now in the root scope instead of being nested inside the events resource.

After this PR is deployed, I'll make a new one to delete the old `event` relationship.


db/migrate/20260123011001_validate_add_affiliable_to_affiliations.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def down
+    add_check_constraint :event_affiliations, "affiliable_type IS NOT NULL", name: "event_affiliations_affiliable_type_null", validate: false
+    change_column_null :event_affiliations, :affiliable_type, true
+
+    add_check_constraint :event_affiliations, "affiliable_id IS NOT NULL", name: "event_affiliations_affiliable_id_null", validate: false
+    change_column_null :event_affiliations, :affiliable_id, true
+  end
+end
```
Luke-Oldenburg:
```suggestion
end

```

garyhtou (approved):

garyhtou:
When we create an `Event` based on an `Event::Application`, we'll create a copy of the `Event::Application`'s `Affiliation` and associate the new copy to `Event`. This way we preserve the affiliation that the org applied as.

PR #12711, opened 2026-01-23
[v4] Add billing address to user

Mohamad-Mortada:
## Summary of the problem
People wanna see their billing address in app

## Describe your changes
And so the people shall!


Luke-Oldenburg (approved):
lgtm

PR #12712, opened 2026-01-23
[Emburse Cards] Fix random </main> tag

manuthecoder:
why do we have a random `</main>` tag. this removes the random container added as well so it fixes this bug

image


sampoder (approved):
LOL cool

PR #12722, opened 2026-01-24
Add option to skip Open Graph tags in layout

manuthecoder:
## Problem
bug with donation pages (duplicate og tags)
image


sampoder (approved):
amazing

PR #12728, opened 2026-01-25
[Discord] Add support for PDF receipts

YodaLightsabr:
## Summary of the problem

#12568

PDF receipts don't currently  work.

## Describe your changes

Fixes #12568

- Accepts PDF receipts
- Now provides a message for invalid file types
- Refactors modal handling code to be cleaner

### DIFF
--- a/app/jobs/discord/handle_interaction_job.rb
+++ b/app/jobs/discord/handle_interaction_job.rb
@@ -118,7 +118,7 @@ module Discord
 
       attachments = @interaction.dig(:data, :resolved, :attachments)
       file = attachments&.values&.first
-      content_type = file[:content_type] if file.present?
+      content_type = file&.[](:content_type)
 
       unless file.present? && (content_type == "application/pdf" || content_type&.start_with?("image/"))
         return respond(embeds: [{
--- a/app/jobs/discord/handle_interaction_job.rb
+++ b/app/jobs/discord/handle_interaction_job.rb
@@ -117,13 +117,22 @@ module Discord
       return respond(content: "This Discord server is not currently linked to the same HCB organization") unless hcb_code.event.id == @current_event&.id
 
       attachments = @interaction.dig(:data, :resolved, :attachments)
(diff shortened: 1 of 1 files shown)


app/jobs/discord/handle_interaction_job.rb
```diff
@@ -117,13 +117,20 @@ def attach_receipt_modal
...
+      unless file.nil? || content_type == "application/pdf" || content_type&.start_with?("image/")
+        return respond(embeds: [{
+                         title: "There was a problem with your receipt",
+                         description: "Only images and PDF files are supported. Please try again.",
+                         color:
+                       }])
+      end
+
+      filename = file[:filename]
+      content_type = file[:content_type]
```
manu-gurudath: content_type is set twice

app/jobs/discord/handle_interaction_job.rb
```diff
@@ -117,13 +117,20 @@ def attach_receipt_modal
...
-      io = URI(url).open if url.present?
+      unless file.nil? || content_type == "application/pdf" || content_type&.start_with?("image/")
+        return respond(embeds: [{
+                         title: "There was a problem with your receipt",
+                         description: "Only images and PDF files are supported. Please try again.",
+                         color:
+                       }])
+      end
+
+      filename = file[:filename]
```
manu-gurudath, line 131: make sure file is not nil, this will throw NoMethodError

YodaLightsabr: The `unless` should do this but I had it the other way around. Fixed now

app/jobs/discord/handle_interaction_job.rb
YodaLightsabr: Good catch, thanks

garyhtou (approved):
We'll try this first and see if it works. If we continue to get IO errors, then we'll try the StringIO solution that Sam suggested in the linked GH issue

garyhtou:
Based on Ian's testing in development, this version seems to work fine

PR #12729, opened 2026-01-25
Add Wise transfer option to transfer wizard 🪄

manuthecoder:
This PR adds the Wise transfer option to the transfer form and controller.

## Changes
- Added Wise as a transfer option in the transfer wizard
- Updated transfer form to include Wise option
- Updated controller to handle Wise transfers

## Screenshots
See commit for implementation details.

### DIFF
--- a/app/javascript/controllers/transfer_form_controller.js
+++ b/app/javascript/controllers/transfer_form_controller.js
@@ -14,6 +14,7 @@ export default class extends Controller {
     'answerText',
     'answerCTA',
     'learnMore',
+    'wiseAnswerNote',
   ]
 
   static values = {
@@ -88,6 +89,8 @@ export default class extends Controller {
 
       this.answerTarget.hidden = false
       this.wizardTarget.hidden = true
+
+      this.wiseAnswerNoteTarget.hidden = payload.type !== 'Wise transfer'
     }
   }
 
@@ -98,6 +101,7 @@ export default class extends Controller {
     if (answer == 'Mailed check') value = 'check'
     if (answer == 'International wire') value = 'wire'
     if (answer == 'Wise transfer') value = 'wise'
+
     window.Turbo.visit(this[`${value}Value`])
   }
 }
--- a/app/javascript/controllers/transfer_form_controller.js
+++ b/app/javascript/controllers/transfer_form_controller.js
@@ -20,6 +20,7 @@ export default class extends Controller {
     ach: String,
     check: String,
     wire: String,
+    wise: String,
   }
 
   static questions = [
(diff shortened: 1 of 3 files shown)


Luke-Oldenburg:
Please see #12142

manu-gurudath:
**(just for reference)**

looking at the list, these are the 5 countries which are supported by wires which are not supported by wise:

1. HRK (Croatian Kuna)
2. RUB (Russian Ruble)
3. ISK (Icelandic Króna)
4. INR (Indian Rupee)
5. UGX (Ugandan Shilling)


however, all countries supported by wise are supported by wires

app/views/events/_transfer_form.html.erb
```diff
@@ -73,6 +74,18 @@
       </a>
       <p class="eyebrow pt-0 mb-0">Our recommendation</p>
       <h3 class="mt-2 font-bold mb-2 capitalize" data-transfer-form-target="answerText"></h3>
+      <div style="font-weight: normal" class="mb-4" data-transfer-form-target="wiseAnswerNote">
+        <strong>Note:</strong>
+        Wise does <strong>not</strong> support currencies in:
```
YodaLightsabr:
```suggestion
        Wise does <strong>not</strong> support these currencies:
```

app/views/events/_transfer_form.html.erb
```diff
@@ -73,6 +74,18 @@
...
+        Wise does <strong>not</strong> support these currencies:
+        <ul>
+          <li>HRK (Croatian Kuna)</li>
+          <li>RUB (Russian Ruble)</li>
+          <li>ISK (Icelandic Króna)</li>
+          <li>INR (Indian Rupee)</li>
+          <li>UGX (Ugandan Shilling)</li>
+        </ul>
+        If you need to send to one of these currencies, <strong>please select "International wire" instead.</strong>
+      </div>
```
Luke-Oldenburg, line 86: Should we make this a question?

app/javascript/controllers/transfer_form_controller.js
```diff
@@ -44,6 +43,18 @@ export default class extends Controller {
         link: 'https://help.hcb.hackclub.com/article/25-what-are-money-transfers',
       },
     },
+    {
+      id: 3,
+      question: 'Is your transfer amount over $500?',
```
Luke-Oldenburg, line 48: Would it be better to make this $50,000?

garyhtou: Let's ask ops. We allow wire transfers for free starting at $500+. However, Ops will have a better idea of what they prefer and which is cheaper/easier.

garyhtou (approved):

PR #12733, opened 2026-01-25
[Column::AccountNumber] Add unique validation for `account_number_bidx`

Luke-Oldenburg:
## Summary of the problem

We were missing a uniqueness validation in rails from when we added the DB validation.


garyhtou (approved):
Nice!

PR #12734, opened 2026-01-25
[User Settings] Only render billing address tab if stripe cardholder exists

Luke-Oldenburg:
## Summary of the problem

This page can't load if there is no stripe cardholder and there isn't a reason to create one from opening this page as the user has no cards yet.


manuthecoder (approved):
LGTM!

garyhtou (approved):
Code looks good. I also like that we're hiding this until they need it. As we add more features/settings, it can get overwhelming for users.

PR #12736, opened 2026-01-25
[Ledger] Add declined filter and paginate PTs with CTs

### LINKED ISSUE #12735: [Ledger] Add filter for declined transactions
## Describe the feature

It would be great to have a toggle to show declined transactions on the ledger. This would especially come in handy for the card grants ledger.

YodaLightsabr:
## Summary of the problem

#12735

Right now, we exclude declined transactions from the ledger. However, there are some cases where it would be nice to have declined transactions visible, like card grants, for example.

## Describe your changes

Implements #12735

This PR adds a new filter to show declined and reversed transactions (CPTs and CTs). As a side effect, it also combines the transaction pagination to include CPTs, too. This is because HCB assumes that there are only a few PTs and puts them at the very top on page 1. However, when filtering by declined transactions, it's easy to get over 100 CPTs included, necessitating pagination.

<table>
<tr>
 <td>image<br>The filter menu</td>
 <td>image<br>The ledger with the filter (blurred for good measure—this data is all public but I'm not sure if it's been aggregated this way before)<td>
</table>

### DIFF
--- a/app/views/events/_filter_menu.html.erb
+++ b/app/views/events/_filter_menu.html.erb
@@ -22,6 +22,7 @@
         <button id="category" data-tabs-target="btn" data-action="click->tabs#select">Category</button>
         <button id="merchant" data-tabs-target="btn" data-action="click->tabs#select">Merchant</button>
         <button id="receipts" data-tabs-target="btn" data-action="click->tabs#select">Receipts</button>
+        <button id="declined" data-tabs-target="btn" data-action="click->tabs#select">Declined</button>
       </div>
 
       <div class="col-span-2 p-2 overflow-y-scroll">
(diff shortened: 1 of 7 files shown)


app/views/events/_filter_menu.html.erb
```diff
@@ -134,6 +135,14 @@
...
           </div>
         </div>
+        <div data-tabs-target="tab" id="declined">
+          <div class="flex items-center">
+            <%= link_to("Declined or reversed", event_ledger_path(@event, upsert_query_params(declined_or_reversed: true, page: 1)), class: "flex-auto menu__action #{"menu__action--active" if @declined_or_reversed}", data: { turbo_prefetch: "false" }) %>
+          </div>
+          <div>
+            <%= link_to("All other transactions", event_ledger_path(@event, upsert_query_params(declined_or_reversed: nil, page: 1)), class: "flex-auto menu__action #{"menu__action--active" unless @declined_or_reversed}", data: { turbo_prefetch: "false" }) %>
+          </div>
+        </div>
```
Luke-Oldenburg, line 145: I think we should have 3 options:
- all
- declined or reversed
- successful

Luke-Oldenburg, line 145: Could we put this inside the amount tab or make the tab name more generic?

garyhtou (changes requested):
I talked to Ian and we're gonna hold off on this until the tx engine rewrite

app/views/events/_filter_menu.html.erb
garyhtou: +1 on having an All option

YodaLightsabr:
Because we're waiting until the transaction engine rewrite, this will be quite easy with the new `Ledger::Query` system (#12834) and look nothing like this code. As a result, I'm closing this PR.

PR #12741, opened 2026-01-25
[Card Grants] Add support message and URL to card grant setting

YodaLightsabr:
## Summary of the problem

#12740

Card grant issuers have requested a way to provide grant recipients with a custom support message. This will also help reduce the support workload on our team.

## Describe your changes

Implements #12740

* Adds `support_message` and `support_url` to `CardGrantSetting`
* Adds a new action to card grants to view the support message / navigate to the support URL
* Adds card grants to the settings dropdown even if a `CardGrantSetting` doesn't yet exist

image

image

image


app/views/events/_settings_dropdown.html.erb
```diff
@@ -18,7 +18,7 @@
                     edit_event_path(@event, tab: "reimbursements"),
                     selected: @settings_tab == "reimbursements",
                     data: { turbo: true, turbo_action: "advance" } %>
-      <% if @event.card_grant_setting.present? %>
+      <% if @event.plan.card_grants_enabled? %>
```
Luke-Oldenburg, line 21: This line was a safety in case the event doesn't have a `CardGrantSetting` as that will cause the form to error. There should always be a `CardGrantSetting` if card grants are enabled on the plan. If you have an example of bad prod data please link it so we can determine why it isn't being created.

Luke-Oldenburg (commented):
Could you provide an example of a support message / support URL please? I want to make sure this isn't something that should just be a generic support article or individually customizable on grants.

app/views/events/_settings_dropdown.html.erb
YodaLightsabr: I don't think we create the `CardGrantSetting` if the plan changes. I'm fine dropping the change from this PR, but this shouldn't break anything.

YodaLightsabr: Since we do `CardGrantSetting.find_or_create_by!(event: @event) if @event.plan.card_grants_enabled? && @settings_tab == "card_grants"` on https://github.com/hackclub/hcb/blob/3a1422e4faaff8ba237dc4b92497896e68bdfd79/app/controllers/events_controller.rb#L321, we should be fine rendering the tab without an existing `CardGrantSetting`.

garyhtou (approved):

PR #12742, opened 2026-01-26
[Card Grants] Require agreeing to purpose and CoC to activate

YodaLightsabr:
## Summary of the problem

Users might not be informed about restrictions on their grant. This makes it hard to enforce a consistent fraud policy.

## Describe your changes

This PR adds a Code of Conduct checkbox and a conditional purpose checkbox that only renders if the grant's purpose is set.

image


app/views/card_grants/_invitation.html.erb
```diff
@@ -2,14 +2,32 @@
...
+              <%= form.check_box :purpose, required: true %>
+              <%= form.label :purpose do %>
+                I will only use this card for the provided purpose: <strong><%= card_grant.purpose %></strong>
+              <% end %>
+            </div>
+          <% end %>
+          <div class="field field--checkbox mb2">
+            <%= form.check_box :policy, required: true %>
+            <%= form.label :policy do %>
+              I will abide by the <%= link_to "Code of Conduct", "https://hackclub.com/conduct", target: :_blank %> and understand that fraud will result in a ban
```
Luke-Oldenburg: What in the Code of Conduct applies to spending from a grant? I don't disagree with making them agree to spending terms but I don't think this policy covers our bases appropriately.

garyhtou: It looks like we have the following line in the CoC.

> Defrauding Hack Club, including HCB, by collecting funds or resources under false information, identity, or pretenses. This is treated as a third offense and could result in a call to your school admin.

I'm cool with keeping the CoC in the list of things you agree to. Primarily, since ppl who receive card grants are Hack Clubbers.

In the future, we'll change this to a dedicated card grant policy

app/views/card_grants/_invitation.html.erb
```diff
@@ -2,14 +2,32 @@
...
+        <p>
+          Before you can start spending, you need to agree to the Card Issuing Terms and pledge to use the card responsibly.
+        </p>
         <%= form_with url: activate_card_grant_path(card_grant) do |form| %>
-          <div class="field field--checkbox mb2 justify-center">
+          <% if card_grant.purpose.present? %>
+            <div class="field field--checkbox mb2">
+              <%= form.check_box :purpose, required: true %>
+              <%= form.label :purpose do %>
+                I will only use this card for the provided purpose: <strong><%= card_grant.purpose %></strong>
```
garyhtou: Include that they agree to the instructions too

app/views/card_grants/_invitation.html.erb
YodaLightsabr: image

garyhtou (approved):

PR #12743, opened 2026-01-26
Add external_id to Contract::Party

polypixeldev:
## Summary of the problem

We currently have to fetch the DocuSeal document every time we need the slug for a party submission. Since this slug doesn't change once we create the document, we can just store it in the database to avoid all of these requests.

## Describe your changes

Adds `external_id` to `Contract::Party`, use it instead of looking up the slug, and backfill and set it for all contracts sent with DocuSeal.

After deploy, make sure to run `OneTimeJobs::BackfillPartyExternalIds.perform_now`, which will backfill slugs in most recent order.

### DIFF
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -168,6 +168,14 @@ class Contract < ApplicationRecord
     end
 
     update(external_service: :docuseal, external_id: response.body.first["submission_id"])
+
+    submitters = docuseal_document["submitters"]
+
+    parties.each do |party|
+      slug = submitters.select { |s| s["role"] == party.docuseal_role }&.[](0)&.[]("slug")
+
+      party.update!(external_id: slug) if slug.present?
+    end
   end
 
   def archive_on_docuseal!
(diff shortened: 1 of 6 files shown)


app/models/contract.rb
```diff
@@ -168,6 +168,14 @@ def send_using_docuseal!
     end
 
     update(external_service: :docuseal, external_id: response.body.first["submission_id"])
+
+    submitters = docuseal_document["submitters"]
+
+    parties.each do |party|
+      slug = submitters.select { |s| s["role"] == party.docuseal_role }&.[](0)&.[]("slug")
+
+      party.update!(external_id: slug) if slug.present?
```
garyhtou:
```suggestion
      slug = submitters.select { |s| s["role"] == party.docuseal_role }&.[](0)&.[]("slug")

      if slug.present?
        party.update!(external_id: slug)
      else
        Rails.error.unexpected("Contract Party (#{party.id}) role and/or slug missing in DocuSeal.")
      end
```

garyhtou (approved):

garyhtou:
@polypixeldev is fixing tests, and then we're good to go

PR #12744, opened 2026-01-26
Fix layout alignment in HCB code transaction headers

manuthecoder:
### problem

image

### solution
image


app/views/hcb_codes/_disbursement.html.erb
```diff
@@ -19,7 +19,7 @@
       <span class="regular muted">to</span>
       <%= @destination.name %>
     </span>
-    <span class="badge h4 bg-<%= @hcb_code.disbursement.state %>">
+    <span class="badge h4 bg-<%= @hcb_code.disbursement.state %>" style="order: 1">
```
sampoder: should we not be using tailwind for this?

sampoder (approved):

PR #12746, opened 2026-01-26
Remove old contract model

polypixeldev:
## Summary of the problem

We've fully switched to `Contract`, and are no longer using `OrganizerPosition::Contract`!

## Describe your changes

Delete the old `OrganizerPosition::Contract` model file


db/migrate/20260126203310_drop_organizer_position_contracts.rb
garyhtou, line 1: Nice! listing the columns here make it reversible

garyhtou (approved):

PR #12747, opened 2026-01-26
Add not_voided scope to contracts

polypixeldev:
## Summary of the problem

`contracts.where.not(aasm_state: :voided)` isn't great, let's clean that up and make it a scope

## Describe your changes

Factor it out into the `not_voided` scope!


Mohamad-Mortada (approved):

manuthecoder (approved):
lgtm

PR #12750, opened 2026-01-26
Improve blankslates

manuthecoder:
This PR improves the experience for new users by adding full-page blankslates


garyhtou (commented):
From discussion:
- We're going to call this page style a "one-pager". It serves as an educational component for organizers when the page has no data. The goal is to educate the organizer on how the feature works.
- Transparency mode will stay the same—they'll continue seeing the existing blank slates

**Open question:**
- Do we want to allow organizers to close the one-pager and view the existing blank slates? I imagine the one-pager having an Apple Liquid Glass-like background with an `X` on the top right.

garyhtou:
> Do we want to allow organizers to close the one-pager and view the existing blank slates? I imagine the one-pager having an Apple Liquid Glass-like background with an X on the top right.

Based on discussions with engineers: no need

garyhtou (approved):

PR #12753, opened 2026-01-27
Write an additional HCBCode during disbursement create to represent the incoming transaction

albertchae:
The ultimate goal is to be able to enforce an event_id foreign key on HCBCode to make fetching an event's transactions easier. Disbursements are a blocker to this because the single associated HCBCode with a disbursement belongs to 2 separate events.

So we want to split this HCBCode up into 2, one for outgoing and one for incoming. We originally thought we'd create 2 new ones entirely, keep the existing one temporarily, then migrate over to the 2 new ones. But instead we decided to reuse the existing disbursement HCBCode to represent "outgoing" disbursements. This makes it a lot easier to keep links to old disbursement hcb codes live as well as associations like comments/receipts.

This PR only introduces an additional write to create an incoming disbursement HCBCode. After we verify behavior in production, the plan will be to backfill all existing disbursements to have an incoming side hcb code and then we can start reading from it properly and then we can start reading from it properly


app/models/disbursement.rb
```diff
@@ -220,16 +220,28 @@ def transferred_at
...
   end
 
-  def hcb_code
-    "HCB-#{TransactionGroupingEngine::Calculate::HcbCode::DISBURSEMENT_CODE}-#{id}"
+  def outgoing_hcb_code
+    "HCB-#{TransactionGroupingEngine::Calculate::HcbCode::OUTGOING_DISBURSEMENT_CODE}-#{id}"
   end
 
+  # Legacy alias - use outgoing_hcb_code instead
+  alias_method :hcb_code, :outgoing_hcb_code
```
albertchae, line 228: we will delete this in the near future

spec/services/disbursement_service/create_spec.rb
```diff
@@ -36,28 +36,28 @@
 
     pending_outgoing = disbursement.raw_pending_outgoing_disbursement_transaction
     expect(pending_outgoing.amount_cents).to eq(-123_45)
-    expect(pending_outgoing.date_posted).to eq(Date.today)
+    expect(pending_outgoing.date_posted).to eq(Date.current)
```
albertchae, line 39: https://railsdrop.com/2023/06/23/understanding-the-difference-between-date-current-and-date-today-in-ruby/

Use rails setting timezone for Date
This is more in line with how `date_posted` is being set to `@disbursement.created_at`

https://github.com/hackclub/hcb/blob/f7cfd61650a2c66640fd4c317b63ec730bba2d24/app/services/pending_transaction_engine/raw_pending_outgoing_disbursement_transaction_service/disbursement/import_single.rb#L21

garyhtou (approved):

PR #12755, opened 2026-01-27
Store local object on `RawColumnTransaction`

YodaLightsabr:
## Summary of the problem

The complete data associated with a Column transaction isn't currently stored on `RawColumnTransaction`

## Describe your changes

This PR creates a new column to store the Column data


app/models/raw_column_transaction.rb
```diff
@@ -9,6 +9,7 @@
 #  column_transaction :jsonb
 #  date_posted        :date
 #  description        :text
+#  local_object       :jsonb
```
garyhtou: Let's try something more descriptive. Maybe `column_transfer`?

app/jobs/one_time_jobs/backfill_local_object_on_raw_column_transactions.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module OneTimeJobs
+  class BackfillLocalObjectOnRawColumnTransactions
+    def self.perform
+      raw_column_transactions = RawColumnTransaction.where(id: CanonicalTransaction.where(transaction_source_type: "RawColumnTransaction", hcb_code: HcbCode.where("hcb_code ILIKE 'HCB-000%'").select(:hcb_code)).select(:transaction_source_id))
```
garyhtou: where the column is currently null

garyhtou (approved):

PR #12757, opened 2026-01-27
Add ScGoogleGrant plan

polypixeldev:
## Summary of the problem

New grant program with a new contract template, so we're adding a new plan!

## Describe your changes

Adds new plan that inherits from standard and overrides contract template ID.

This PR also adds a new plan method to skip certain prefilled fields on the fiscal sponsorship contract, and implements it in `Contract::FiscalSponsorship`. While this is not the best way to do this, it'll work for now since we need to get these contracts out. In the future, we can improve this by creating `Contract::Template` and storing prefills along with those.


garyhtou (approved):
imma trust that this works lol

PR #12774, opened 2026-01-29
[User Settings] Fix billing address sometimes being hidden

Luke-Oldenburg:
## Summary of the problem

When viewing another user's settings that didn't have a stripe cardholder, your own billing address tab would not render too.

## Describe your changes

Use user local instead of instance variable. I plan on removing the instance variables from all the nav partials and replacing with strict locals in a future PR.


manuthecoder (approved):

sampoder (approved):
sure sgtm

PR #12775, opened 2026-01-29
[User] Rename teenager related functions and prevent excessive writes

Luke-Oldenburg:
## Summary of the problem

Cleanup from review of #12607

## Describe your changes

- Rename `teenager?` and `joined_as_teenager?`
- Add `(if: :will_save_change_to_birthday?)` to model callbacks


sampoder (approved):
lgtm thx 🫡

PR #12785, opened 2026-01-30
Fix `OneTimeJobs::BackfillColumnTransferOnRawColumnTransactions`

YodaLightsabr:
## Summary of the problem

These should be instance methods, not class methods. We also want to be limiting the process method, not the perform method.


Luke-Oldenburg (approved):
lgtm

PR #12787, opened 2026-01-30
Create Disbursement wrapper classes for incoming/outgoing directions

albertchae:
## Summary of the problem

Step 4 of https://github.com/hackclub/hcb/issues/12725


app/models/disbursement/incoming.rb
```diff
@@ -0,0 +1,27 @@
...
+    def hcb_code
+      disbursement.incoming_hcb_code
+    end
+
+    def event
+      disbursement.destination_event
+    end
+
+    def amount
+      disbursement.amount.abs
```
garyhtou: I'm generally not a fan of `abs` because it can hide issues, such as when the amount is unexpectedly negative in this case. I prefer `* -1` instead.

However, amount here should already be positive, so let's drop it.
```suggestion
      disbursement.amount
```

app/models/disbursement/outgoing.rb
```diff
@@ -0,0 +1,27 @@
...
+    def hcb_code
+      disbursement.outgoing_hcb_code
+    end
+
+    def event
+      disbursement.source_event
+    end
+
+    def amount
+      -disbursement.amount.abs
```
garyhtou:
```suggestion
      -disbursement.amount
```

app/models/disbursement/incoming.rb
garyhtou, line 1: During the migration period, I believe the `canonical_transactions`, `canonical_pending_transactions`, and `pending_expired?` methods would return inaccurate results because the Incoming HcbCodes would contain no CT nor CPT until we complete the data migration (which involves moving CT/CPT from Outgoing to Incoming HcbCodes).

So, I think we might want to override these method definitions for the time being so they also look for CT/CPT located on the outgoing side.

thoughts?

```ruby
def canonical_transactions
  @canonical_transactions ||= disbursement.local_hcb_code.canonical_transactions.where("amount_cents > 0")
end

def canonical_pending_transactions
  @canonical_pending_transactions ||= disbursement.local_hcb_code.canonical_pending_transactions.where("amount_cents > 0")
end

def pending_expired?
  canonical_pending_transactions.pending_expired.any?
end
```

garyhtou (approved):
Thank you!

PR #12789, opened 2026-01-31
Fix transaction memo bug 

manuthecoder:
bug 
image


YodaLightsabr:
Which browser was the screenshot taken on?

YodaLightsabr (approved):

PR #12790, opened 2026-01-31
[Command bar] Add empty state

manuthecoder:
creates a beautiful empty state instead of a weird gap when you type something which returns no results

image


sampoder (approved):
looks hella good to me!

PR #12792, opened 2026-01-31
Use new incoming / outgoing disbursements

sampoder:


app/api/api/entities/transaction.rb
```diff
@@ -27,12 +27,8 @@ class Transaction < Base
...
           # Canonical Transactions. Therefore, for the API, we are overriding the
           # default amount_cents exposure defined in the LinkedObjectBase.
-          if hcb_code.disbursement? && org == hcb_code.disbursement.source_event &&
-             !(hcb_code.disbursement.source_subledger_id && hcb_code.disbursement.destination_subledger_id.nil?) # disbursements with a source_subledger_id and no destination_subledger_id are returned card grants
-            next -hcb_code.disbursement.amount
-          elsif hcb_code.disbursement?
-            next hcb_code.disbursement.amount
-          end
+          next hcb_code.incoming_disbursement.amount if hcb_code.incoming_disbursement?
+          next hcb_code.outgoing_disbursement.amount if hcb_code.outgoing_disbursement?
```
YodaLightsabr, line 40: Might these still need to be negative? `outgoing_disbursement.amount` returns a positive number

sampoder: checked, we don't need to

garyhtou: https://github.com/hackclub/hcb/blob/3c74cf0ed9df543cad1f2f14048227d087dc7de6/app/models/disbursement/outgoing.rb#L15-L17

app/models/disbursement.rb
```diff
@@ -259,6 +259,10 @@ def fulfilled?
     deposited?
   end
 
+  def plain_transfer?
+    !source_subledger_id && !destination_subledger_id
+  end
+
```
garyhtou: Should we rename to `inter_event_transfer?` "plain" feels a bit generic.

app/models/hcb_code.rb
```diff
@@ -103,7 +103,7 @@ def type
     return :ach if ach_transfer?
     return :check if check? || increase_check?
     return :card_grant if card_grant?
-    return :disbursement if disbursement?
+    return :disbursement if outgoing_disbursement? || incoming_disbursement?
```
YodaLightsabr, line 106: Should we be returning `:outgoing_disbursement` and `:incoming_disbursement` here instead of `:disbursement` if we're treating those two as different types?

app/controllers/events_controller.rb
```diff
@@ -971,8 +971,8 @@ def self.filter_transaction_type(type, settled_transactions:, pending_transactio
         "pending" => ->(t) { t.raw_pending_outgoing_check_transaction_id || t.increase_check_id }
       },
       "hcb_transfer"           => {
-        "settled" => ->(t) { t.local_hcb_code.disbursement? && !t.local_hcb_code.disbursement.destination_subledger_id && !t.local_hcb_code.disbursement.source_subledger_id },
-        "pending" => ->(t) { t.local_hcb_code.disbursement? && !t.local_hcb_code.disbursement.destination_subledger_id && !t.local_hcb_code.disbursement.source_subledger_id }
+        "settled" => ->(t) { t.local_hcb_code.outgoing_disbursement? || t.local_hcb_code.incoming_disbursement? },
+        "pending" => ->(t) { t.local_hcb_code.outgoing_disbursement? || t.local_hcb_code.incoming_disbursement?  }
```
garyhtou: Is this a change in behavior since we're no longer excluding card grant disbursements? We may need to check that `Disbursement#plain_transfer?` is `true`.

```suggestion
        "settled" => ->(t) { t.local_hcb_code.outgoing_disbursement&.plain_transfer? || t.local_hcb_code.incoming_disbursement&.plain_transfer? },
        "pending" => ->(t) { t.local_hcb_code.outgoing_disbursement&.plain_transfer? || t.local_hcb_code.incoming_disbursement&.plain_transfer? }
```

P.S. I want to rename `plain_transfer?` to `inter_event_transfer?`

app/models/canonical_pending_transaction.rb
```diff
@@ -252,15 +252,28 @@ def linked_object
     return raw_pending_donation_transaction.donation if raw_pending_donation_transaction
     return raw_pending_invoice_transaction.invoice if raw_pending_invoice_transaction
     return raw_pending_bank_fee_transaction.bank_fee if raw_pending_bank_fee_transaction
-    return raw_pending_incoming_disbursement_transaction.disbursement if raw_pending_incoming_disbursement_transaction
-    return raw_pending_outgoing_disbursement_transaction.disbursement if raw_pending_outgoing_disbursement_transaction
+    return Disbursement::Incoming.new(raw_pending_incoming_disbursement_transaction.disbursement) if raw_pending_incoming_disbursement_transaction
+    return Disbursement::Outgoing.new(raw_pending_outgoing_disbursement_transaction.disbursement) if raw_pending_outgoing_disbursement_transaction
```
garyhtou: Let's move the wrapping into the raw models.

```ruby
class RawPendingIncomingDisbursementTransaction
  def incoming_disbursement
    Disbursement::Incoming.new(disbursement)
  end
end
```
and the same thing for RawPendingOutgoingDisbursementTransaction.

app/models/canonical_pending_transaction.rb
```diff
@@ -252,15 +252,28 @@ def linked_object
...
 
     nil
   end
 
   def disbursement
-    return linked_object if linked_object.is_a?(Disbursement)
+    Rails.error.unexpected "CanonicalPendingTransaction#disbursement accessed"
+    return nil unless raw_pending_outgoing_disbursement_transaction || raw_pending_incoming_disbursement_transaction
 
+    linked_object
```
garyhtou: I have a feeling that in order to maintain existing behavior, we need to return an object of the `Disbursement` class, rather than our wrapped classes. This is because any existing code that checks `is_a?(Disbursement)` would fail if `object.class == Disbursement::Incoming/Outgoing`.

```suggestion
    # outgoing_disbursement and incoming_disbursement are derived from `linked_object`
		(outgoing_disbursement || incoming_disbursement)&.disbursement
```

app/models/canonical_transaction.rb
```diff
@@ -380,8 +380,14 @@ def emburse_transfer
     nil # TODO
   end
 
-  def disbursement
-    return linked_object if linked_object.is_a?(Disbursement)
```
garyhtou, line 384: Are we actually ready to remove this method? I'm concerned we still have some usages litered around. Maybe we should deprecate it like what we did for CPT.

app/models/hcb_code.rb
```diff
@@ -347,7 +350,7 @@ def incoming_disbursement?
   end
 
   def card_grant?
-    disbursement? && disbursement&.card_grant.present?
+    outgoing_disbursement? && outgoing_disbursement&.card_grant.present?
```
garyhtou, line 354:
```suggestion
    # Is this the issuing of a card grant? This method should return false on the receiving end (never true for Disbursement::Incoming)
    outgoing_disbursement? && outgoing_disbursement&.card_grant.present?
```

garyhtou (commented):
TODO: Maybe remove `TransactionGroupingEngine::Calculate::HcbCode::DISBURSEMENT_CODE` constant.

image

FYI, doing so will modify our tests: https://github.com/hackclub/hcb/blob/6b8354c448346c5dc4264abeb41415ef96f55401/spec/misc/hcb_code_type_spec.rb#L24-L29

sampoder:
> TODO: Maybe remove TransactionGroupingEngine::Calculate::HcbCode::DISBURSEMENT_CODE constant.

Removed

app/services/transaction_engine/syntax_sugar_service/linked_object.rb
```diff
@@ -32,7 +32,9 @@ def run
 
           return likely_donation if donation?
 
-          return likely_disbursement if disbursement?
+          # return outgoing_disbursement if outgoing_disbursement?
+
+          # return incoming_disbursement if incoming_disbursement?
```
garyhtou, line 37: I've confirmed these methods are not used

db/functions/hcb_code_type_v03.sql
```diff
@@ -0,0 +1,29 @@
...
+    WHEN '200' THEN 'donation'
+    WHEN '201' THEN 'partner_donation'
+    WHEN '300' THEN 'ach_transfer'
+    WHEN '310' THEN 'wire'
+    WHEN '350' THEN 'paypal_transfer'
+    WHEN '360' THEN 'wise_transfer'
+    WHEN '400' THEN 'check'
+    WHEN '401' THEN 'increase_check'
+    WHEN '402' THEN 'check_deposit'
+    WHEN '500' THEN 'disbursement'
```
garyhtou: Do we want this to now say `outgoing_disbursement`? Or do we save that for later?

garyhtou (approved):

PR #12794, opened 2026-01-31
[Event::Plan] Clarify Internal is for transaction engine use only

garyhtou:
## Summary of the problem

I'm finding HCB Ops' orgs in `Internal `, when it should only be used for transaction engine stuff.

## Describe your changes

Clarify in the Event::Plan description.

### DIFF
--- a/app/models/event/plan/internal.rb
+++ b/app/models/event/plan/internal.rb
@@ -28,7 +28,7 @@ class Event
       end
 
       def description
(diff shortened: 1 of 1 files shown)


garyhtou:
We'll need to have a convo around whether we want all our weird test orgs (e.g. DevHacks, etc.) to be under Internal; or create a new Event::Plan for them. Although sometimes we want to have test orgs with different event plans. My rec is:
- most test orgs should be the standard plan or hack club HQ (with we have 0% fees)
- and tag it with org tags or a boolean on event configuration

PR #12795, opened 2026-01-31
Redirect form unauthorized outgoing disbursements to incoming disbursements

YodaLightsabr:
Note: this needs to merge after #12792

## Summary of the problem

Once we split disbursement HCB codes, we'll need to handle the case where someone accesses the source HCB code but only has permission to view the destination.

## Describe your changes

When a policy check on HCB code fails, but it's an outgoing disbursement and the user has access to the incoming disbursement, redirect there.

### DIFF
--- a/app/controllers/hcb_codes_controller.rb
+++ b/app/controllers/hcb_codes_controller.rb
@@ -54,7 +54,7 @@ class HcbCodesController < ApplicationController
       render :show
     end
   rescue Pundit::NotAuthorizedError => e
-    if @hcb_code.stripe_card.card_grant.present? && current_user == @hcb_code.stripe_card.card_grant.user
+    if @hcb_code.stripe_card&.card_grant.present? && current_user == @hcb_code.stripe_card.card_grant.user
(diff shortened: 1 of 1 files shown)


app/controllers/hcb_codes_controller.rb
```diff
@@ -57,6 +57,13 @@ def show
     if @hcb_code.stripe_card.card_grant.present? && current_user == @hcb_code.stripe_card.card_grant.user
       redirect_to card_grant_path(@hcb_code.stripe_card.card_grant, frame: params[:frame])
     else
+      if @hcb_code.outgoing_disbursement?
+        incoming_disbursement = @hcb_code.outgoing_disbursement.disbursement.incoming_disbursement
+        if incoming_disbursement.event.users.include?(current_user)
```
garyhtou: This should be a policy check: "Are they able to `show?` the HcbCode of the incoming disbursement?"

garyhtou (approved):
Great! We'll test this once the other PR merges, but this logic looks right.

This PR depends on the other PR adding `HcbCode#outgoing_disbursement`.

sampoder (approved):

garyhtou (approved):

PR #12798, opened 2026-01-31
Remap newly mapped CTs to incoming hcb codes

sampoder:


app/services/event_mapping_engine/map/hcb_codes/short.rb
```diff
@@ -42,6 +42,10 @@ def run
                 subledger_id: guess_subledger_id(hcb_code, ct)
               }
               ::CanonicalEventMapping.create!(attrs)
```
garyhtou: Are there callbacks on CEM that we should be aware of? (that depend on HcbCode/Event)

garyhtou: if so, maybe we should modify `hcb_code` before commiting to the DB

sampoder: Yep, will switch the order

PR #12799, opened 2026-01-31
Use `OrganizerPosition.role_at_least?` in `CardGrantPolicy`

YodaLightsabr:
## Summary of the problem

`user_in_event?` currently needs the user to explicitly be in the event and doesn't factor in parent events

## Describe your changes

Using `OrganizerPosition.role_at_least?` fixes this


sampoder (approved):
yes please yay

PR #12801, opened 2026-01-31
Add event and subledger to HcbCode

sampoder:


app/models/canonical_event_mapping.rb
```diff
@@ -38,11 +38,19 @@ class CanonicalEventMapping < ApplicationRecord
 
   scope :on_main_ledger, -> { where(subledger_id: nil) }
 
-  after_create { canonical_transaction.write_hcb_code }
+  after_create do
+    canonical_transaction.write_hcb_code
```
garyhtou, line 42: Is there a need for this line? because CT has `after_create :write_hcb_code`

garyhtou (approved):

app/models/canonical_event_mapping.rb
garyhtou: We likely dont, but let's address this later since this code is existing behavior

PR #12802, opened 2026-01-31
Improve minimalistic ledger

manuthecoder:
one of the main complaints of the minimalistic ledger is that positive/negative transactions are hard to read. 

this little experiment improves this flipper flag

image


sampoder (approved):
so fire

PR #12809, opened 2026-02-01
Only load the policy if we have a user

sampoder:


Luke-Oldenburg (commented):
Should we use signed_in here?

sampoder:
sure

app/controllers/hcb_codes_controller.rb
```diff
@@ -58,7 +58,7 @@ def show
       redirect_to card_grant_path(@hcb_code.stripe_card.card_grant, frame: params[:frame])
     elsif @hcb_code.outgoing_disbursement?
       incoming_hcb_code = @hcb_code.outgoing_disbursement.disbursement.incoming_disbursement.local_hcb_code
-      if HcbCodePolicy.new(current_user, incoming_hcb_code).show?
+      if current_user && HcbCodePolicy.new(current_user, incoming_hcb_code).show?
```
sampoder:
```suggestion
      if signed_in? && HcbCodePolicy.new(current_user, incoming_hcb_code).show?
```

PR #12815, opened 2026-02-01
[v3] Pass Disbursement instead of incoming/outgoing

garyhtou:
## Summary of the problem

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/649/samples/last

V3 was erroring anytime we tried to render a disbursement because we were passing it incoming/outgoing.

## Describe your changes

pass in disbursement


sampoder (approved):
lgtm!

PR #12824, opened 2026-02-01
Create `Ledger` model

garyhtou:
We're creating a new `Ledger` model to represent a collection of transactions (`Ledger::Item`).

There are two different types of `Ledgers`: Primary and non-primary. Every `Ledger::Item` should eventually (once mapped) belong to exactly one primary ledger. This is where it affects balances. For example, you calculate an Event's balance by going to the Event's (one) primary ledger and calling `Ledger#balance_cents`. Another use case of primary ledgers is Card Grant, since it tracks its balance separately from Event. `Ledger::Item` can belong to only one primary ledger.

Non-primary ledgers serve as a collection of transactions for _any_ reason. For example:
- A list of transactions representing a subscription
- A list of transactions missing receipts
	- This means ledgers may contain transactions that belong to _any_ Event. Non-primary ledgers, in fact, _**do not**_ have the concept of belonging to an Event.
- A bit of a stretch, but pinned transactions could be represented as a non-primary ledger.

`Ledger` `has_many` `Ledger::Item` through `Ledger::Mapping`. This means that `Ledger::Mapping` is similar to `CanonicalEventMapping`. `Ledger::Item` is similar to `HcbCode`.


sampoder (approved):

garyhtou:
I'm now realizing there's a bug in https://github.com/hackclub/hcb/pull/12833. We'll merge and deal with that later.
image

PR #12828, opened 2026-02-01
Revert "Fix `CommentPolicy` bug"

sampoder:
Reverts hackclub/hcb#12813


Luke-Oldenburg (commented):
I think we still want this? If anything it would be good to add nil safety to the other policies in this file.

garyhtou (approved):

PR #12834, opened 2026-02-02
Add `Ledger::Query`

YodaLightsabr:
## Summary of the problem

We need an intermediate query layer for our new `Ledger::Item`s

## Describe your changes

Example usage: 

```
Ledger::Query.new({
  "memo": "Initial transaction",
  "$or": [
    { "amount_cents": 1 },
    { "amount_cents": { "$eq": 10000} }
  ]
}).execute
```

This PR also implements `Ledger::Query` in `EventsController` including compatibility with the existing filter UI.

image


app/models/ledger/query.rb
```diff
@@ -0,0 +1,136 @@
+# frozen_string_literal: true
+
+class Ledger
+  class Query
+    def initialize(query_hash)
+      @query_hash = self.class.sanitize_query(query_hash)
```
garyhtou, line 34: We need to validate the query first:
- Doesn't have extra keys
- `query_hash.is_a?(Hash)`

app/models/ledger/query.rb
```diff
@@ -0,0 +1,136 @@
...
+
+      # TODO: handle authorization
+    end
+
+    # Expected to return an ActiveRecord::Relation of Ledger::Item
+    def execute(ledger_id: nil)
+      # TODO: (WARNING) We're currently using AR `where` syntax, but WILL change
+      apply_query(ledger_id.present? ? Ledger::Item.where(ledger_id: ledger_id) : Ledger::Item, @query_hash)
+    end
```
garyhtou: `Query` needs to be able to support querying multiple ledgers at the same time.

Imagine: You want to search for transactions in all subevents of an org.

```suggestion
    def execute(ledgers: [])
      results = apply_query(Ledger::Item.all, @query_hash)
      results.merge(ledger: ledgers)
    end
```
**Question:** Is doing a `where` or `merge` at the end more secure? Rather than having it be the base of queries.

I added `.all` to `Ledger::Item` because it returns into an `ActiveRecord::Collection`, which is what you pass in other times you call `apply_query`.

app/models/ledger/query.rb
```diff
@@ -0,0 +1,136 @@
...
+        else
+          relation = relation.or(apply_partial_predicate(Ledger::Item, "$eq", key, value))
+        end
+      end
+
+      relation
+    end
+
+    def sanitize_column(column_name)
+      safe_column = Ledger::Item.column_names.find { |col| col == column_name.to_s }
```
garyhtou: Please make this a constant instead of dynamically using `column_names`. We don't want people searching by `Ledger::Item`'s `id`, `marked_no_or_lost_receipt_at`, nor `short_code`.

app/models/ledger/query.rb
```diff
@@ -0,0 +1,136 @@
...
+          end
+        else
+          relation = relation.or(apply_partial_predicate(Ledger::Item, "$eq", key, value))
+        end
+      end
+
+      relation
+    end
+
+    def sanitize_column(column_name)
```
garyhtou: This is unused, and also I'd expect it to return `safe_column` at the end.

app/models/ledger/query.rb
```diff
@@ -0,0 +1,136 @@
...
+      elsif operand.is_a?(String)
+        case operator.to_s
+        when "$in"
+          return relation.where(key => operand)
+        when "$nin"
+          return relation.where.not(key => operand)
+        when "$ilike"
+          return relation.where("#{key} ILIKE ?", operand)
+        when "$like"
+          return relation.where("#{key} ILIKE ?", operand)
```
garyhtou: Let's not implement these for now. They're _**slowwww**_

app/models/ledger/query.rb
```diff
@@ -0,0 +1,141 @@
...
+      relation
+    end
+
+    def sanitize_column(column_name)
+      safe_column = Ledger::Item.column_names.find { |col| col == column_name.to_s }
+
+      raise Ledger::Query::QueryError.new("Invalid column name: #{column_name}") unless safe_column.present?
+    end
+
+    def apply_partial_predicate(relation, operator, key, operand)
```
garyhtou: Key needs to be sanitized

app/models/ledger/query.rb
```diff
@@ -0,0 +1,141 @@
...
+    end
+
+    def self.sanitize_query(query_hash)
+      # TODO: Implement query sanitization logic
+      query_hash
+    end
+
+    private
+
+    def apply_query(raw_relation, query_hash, context = "and")
```
garyhtou: Use keyword args here

app/controllers/ledgers_controller.rb
```diff
@@ -5,8 +5,7 @@ def show
     @ledger = Ledger.find_by_hashid!(params[:id])
     authorize @ledger
 
-    # TODO: Replace with Ledger::Query
-    @items = @ledger.items.order(date: :desc, created_at: :desc).page(params[:page])
+    @items = Ledger::Query.new({}).execute(ledgers: [@ledger]).order(date: :desc, created_at: :desc).page(params[:page])
```
garyhtou:
```suggestion
    @items = Ledger::Query.new({}).execute(ledgers: [@ledger]).order(date: :desc, id: :desc).page(params[:page])
```

app/models/ledger/query.rb
```diff
@@ -0,0 +1,148 @@
...
+    def column_index(column_name)
+      safe_column = PERMITTED_COLUMNS.index(column_name.to_s)
+
+      raise Ledger::Query::Error.new("Invalid column name: #{column_name}") unless safe_column.present?
+
+      safe_column
+    end
+
+    def apply_partial_predicate(relation, operator, raw_key, operand)
+      key = PERMITTED_COLUMNS[column_index(raw_key)]
```
garyhtou: You can also do:

```ruby
PERMITTED_COLUMNS = %w[memo amount_cents date].index_by(&:itself).freeze

key = PERMITTED_COLUMNS[raw_key]
```

This is a bit cleaner and will still make brakeman happy

spec/models/ledger/query_spec.rb
```diff
@@ -0,0 +1,385 @@
...
+                                 "$and" => [
+                                   { amount_cents: { "$gt" => 100 } },
+                                   { amount_cents: { "$lt" => 300 } }
+                                 ]
+                               })
+
+        expect(result.to_sql).to match(/amount_cents > 100/)
+        expect(result.to_sql).to match(/amount_cents < 300/)
+        expect(result.pluck(:id)).to match_array(ids_of(item_c, item_d))
+      end
```
garyhtou, line 241: I think we need a test to verify that the SQL generated is an `AND` of the two clauses

spec/models/ledger/query_spec.rb
```diff
@@ -0,0 +1,385 @@
...
+},
+                                   { "$and" => [
+                                     { amount_cents: { "$gte" => 300 } },
+                                     { amount_cents: { "$lte" => 300 } }
+                                   ]
+}
+                                 ]
+                               })
+
+        expect(result.to_sql).to match(/OR/)
```
garyhtou: I think we need to test a few more things in the generated SQL make us confident that this is correct

garyhtou: maybe something like
```ruby
[45] pry(main)> q.match?(/amount_cents.*AND.*amount_cents.*OR.*amount_cents.*AND.*amount_cents/)
```

app/models/ledger/query.rb
```diff
@@ -0,0 +1,141 @@
...
+        end
+      end
+
+      case operator.to_s
+      when "$eq"
+        relation.where(key => operand)
+      when "$ne"
+        relation.where.not(key => operand)
+      else
+        raise Ledger::Query::Error.new("Unsupported operator: #{operator}")
```
garyhtou:
```suggestion
        raise Ledger::Query::Error.new("Unsupported comparison operator: #{operator}")
```

spec/models/ledger/query_spec.rb
```diff
@@ -0,0 +1,385 @@
...
+    end
+
+    it "raises on non-hash query" do
+      expect { described_class.new("invalid") }.to raise_error(Ledger::Query::Error, /must be a Hash/)
+    end
+
+    it "raises on invalid column name" do
+      query = { invalid_column: 100 }
+      expect { described_class.new(query).execute }.to raise_error(Ledger::Query::Error, /Invalid column name/)
+    end
```
garyhtou: Let's use `field`

Luke-Oldenburg (approved):
lgtm

garyhtou:
> 🤖 Review by [Claude Code](https://claude.com/claude-code), curated by @garyhtou. Multi-expert pass (correctness, security, filter parity, MongoDB conformance, tests) treating `Ledger::Query` as a public API — the query hash will be client-generated, so the load-bearing question per finding is *"would fixing this after callers depend on it be a breaking change?"* Findings were adversarially verified and several reproduced at runtime. Fixes are linked to their commits below.

## Fixed in this PR

**Cross-org data leak (critical).** `execute` skipped ledger scoping whenever the given collection was empty, treating "no ledgers" as "every ledger" — reachable via `CardGrantsController#transaction_index` for any event with zero card grants (an anonymously-reachable path). Now scopes unconditionally; querying all ledgers requires an explicit `all_ledgers:` opt-in, which fails closed on any non-`true` value (e.g. the string `"false"` previously leaked). f7a74f0, bfc618f

**`maximum_amount` filter 500'd the ledger page (high).** `EventsController#ledger` compiled the filter to `{ "$and": { … } }` — a Hash where `$and` requires an array — which raised `Ledger::Query::Error` past the action's Pundit-only rescue. Now emits a single same-field range predicate using `Money#cents` (also fixes a fractional-cent boundary drift), and `$and`/`$or` reject a non-array operand cleanly. b73330a

**MongoDB null/array semantics (breaking to fix later).** `$ne`/`$nin` compiled to bare `!=`/`NOT IN`, silently dropping NULL rows on nullable columns (`linked_object_type`, `receipt_required`, `marked_no_or_lost_receipt_at`) — the opposite of Mongo, which matches missing/null. And `$eq`/implicit equality with an array operand silently became `IN`. Now `$ne`/`$nin` re-include NULLs, and array operands outside `$in`/`$nin` are rejected. d982615

**Operator-first dispatch + `$not` redesign + DoS bounds.** c9f266e
- Dispatch was operand-Ruby-type-first, so an operator's meaning depended on operand type: string `$gt` and non-array `$in` raised misleading "unsupported operator" errors, and `$eq [array]` silently meant `IN`. Now dispatches on the operator, then validates the operand — string ranges work, wrong-type operands get clear errors, nothing is silently reinterpreted.
- The DSL invented a non-Mongo top-level `$not` while rejecting the only `$not` Mongo defines (field-level). Replaced with `$nor` (real top-level negation) + field-level `$not` (`{amount_cents: {$not: {$gt: 100}}}`), both NULL-matching and scope-safe.
- `sanitize_query` was a no-op; since query hashes are untrusted client input, `validate_complexity!` now caps nesting depth (20), predicate-group count (200), and array length (1000) at construction.
- `$search` documented as an HCB extension (not a Mongo operator; `$text` is).

## Blocker for widening the `new_ledger_2026_06_30` flag — fixed in a separate PR

**Account-verification amounts exposed to transparency viewers.** The ledger row rendered micro-deposit amounts (a shared secret proving ownership of a linked external account) in the clear, whereas the legacy page redacts them for non-organizers. This is a pre-existing exposure on `main` (independent of this PR's query work), so it's fixed off `main` in **#14268**. This PR's `minimum_amount`/`maximum_amount`/`$search` filters additionally make those amounts *enumerable* — so once those filters are wired to real behavior, the legacy "drop account-verification rows from amount-filtered results for non-organizers" guard should be ported here too.

## Filter parity: `/:event_id/transaction_list` → `/:event_id/ledger`

The query engine is ready, but `EventsController#ledger` only translates a subset of filters — several controls render an active "filter applied" chip while the query ignores them (the literal `# To-do` in the action). Not yet wired:

| Filter | Status | Notes |
|---|---|---|
| direction, min/max amount, missing_receipts, date, search | wired | search diverges: `$search` is pg_search word-match on `memo` only, vs the old `ILIKE '%q%'` substring across memo/friendly_memo/custom_memo |
| tag, user, type, category, merchant, order_by | **no-op** | chip renders, query ignores it; merchant tab also links to the old page |
| subledger | **ignored** | always resolves the primary `@event.ledger` |
| running balance, acct-verification suppression | **absent** | |

This is the remaining "full parity" work; I paused it here pending your go-ahead (it's a much larger diff than the fixes above). `missing_receipts` also diverges semantically (the ledger adds a plan-aware `receipt_required` that the old `amount <= 0` rule doesn't) — worth a product decision, not a silent difference.

## MongoDB conformance — remaining, deferred for a design decision

The DSL is now "compliant or cleanly-rejecting" for the reachable operators, with two deliberate, load-bearing extensions kept: empty `$and`=all / `$or`=none (the unfiltered ledger view sends `{$and: []}`), and `$search`. Still open, each cheap now and breaking later:

- **Field vocabulary** exposes raw DB column names (`linked_object_type`, `marked_no_or_lost_receipt_at`); renames are breaking, so alias to friendlier public names now or consciously accept them.
- **`Query::Error`** is a single class; split into client-vs-internal subclasses while nothing pattern-matches on it (backward-compatible under the parent). Note a type-mismatched operand can still raise `ActiveRecord::StatementInvalid` at realization, *outside* `LedgersController#show`'s rescue — validate operand-vs-column type at build time.
- `$nor`/`$exists`/`$regex`/etc. beyond the current set remain unsupported (raise cleanly; safe to add later).

## Test coverage

All fixes above are TDD'd — 58 examples in `spec/models/ledger/query_spec.rb` (operators, logical ops, null semantics, scoping, complexity bounds) plus controller regression specs. The remaining gap is the `EventsController#ledger` filter-translation layer, which has no request/controller spec exercising the query-hash it builds — worth adding alongside the parity work.

Luke-Oldenburg:
@claude

YodaLightsabr:
@claude Review this PR

PR #12855, opened 2026-02-02
[Ledger::Item] Index `amount_cents` and `date` for faster filtering

garyhtou:
Index `amount_cents` and `date` for faster filtering


sampoder (approved):
rubber stamp

PR #12856, opened 2026-02-02
[Ledger] Immediately create Ledger on Owners after create

garyhtou:
With this and the backfill, we can get rid of `find_or_create_ledger` and just do a find.

### DIFF
--- a/app/models/card_grant.rb
+++ b/app/models/card_grant.rb
@@ -56,6 +56,7 @@ class CardGrant < ApplicationRecord
 
   belongs_to :event
   has_one :ledger, -> { where(primary: true) }, inverse_of: :card_grant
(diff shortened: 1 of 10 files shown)


app/models/card_grant.rb
```diff
@@ -56,6 +56,7 @@ class CardGrant < ApplicationRecord
 
   belongs_to :event
   has_one :ledger, -> { where(primary: true) }, inverse_of: :card_grant
+  after_create :create_ledger
```
garyhtou, line 59: ik this syntax looks mysterious, but it does accurately create it as a primary ledger

sampoder (approved):

YodaLightsabr (approved):
looks good

PR #12862, opened 2026-02-02
Create `HcbCode#datetime`

sampoder:


garyhtou (approved):

garyhtou (commented):
We also need to update the creation of new `Ledger::Item` to use similar logic. Then afterwards, we can use this method for the backfill job

YodaLightsabr (approved):

Luke-Oldenburg (approved):
lgtm

PR #12863, opened 2026-02-03
Ignore event_id on affiliations before removing

polypixeldev:
## Summary of the problem

We're now using the `affiliable` polymorphic relation, and have backfilled it. Following `strong_migrations`, we should ignore the column before dropping it.

## Describe your changes

Ignore the `event_id` column and remove references to it.

**Merge #12884 first, then this PR, and lastly #12864**

### DIFF
--- a/app/controllers/event/affiliations_controller.rb
+++ b/app/controllers/event/affiliations_controller.rb
@@ -13,7 +13,6 @@ class Event
           name: params[:type],
           metadata: @metadata,
           affiliable: @affiliable,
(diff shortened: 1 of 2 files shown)


garyhtou:
> Merge https://github.com/hackclub/hcb/pull/12864 immediately after this is deployed - creating new affiliations will be broken until this column is removed due to the database non-null constraint on event_id.

Maybe we should remove the not null first

garyhtou (approved):

PR #12864, opened 2026-02-03
Remove event_id from affiliations

polypixeldev:
## Summary of the problem

We're no longer using this column - see #12863.

## Describe your changes

Removes the column

**don't merge before #12863!**


Luke-Oldenburg (commented):
I think this needs to be 3 PRs to ignore it and remove usages, remove the column, and unignore it.

app/models/event/affiliation.rb
```diff
@@ -11,22 +11,15 @@
...
-#
-# Foreign Keys
-#
-#  fk_rails_...  (event_id => events.id)
 #
 class Event
   class Affiliation < ApplicationRecord
     include Hashid::Rails
 
-    belongs_to :event
```
garyhtou: This technically was already removed in https://github.com/hackclub/hcb/pull/12863, but i'm sure the diff will sort itself out

garyhtou (approved):

PR #12889, opened 2026-02-04
Wrap application event creation in a transaction

polypixeldev:
## Summary of the problem

Something could go wrong when creating the event, and it'll be easier to fix that if we roll it all back if it fails. This is what `EventService::Create` does.

## Describe your changes

Wrap the event activation/creation logic in a transaction


app/models/event/application.rb
```diff
@@ -317,28 +317,30 @@ def check_cosigner_update
...
-        affiliation_copy.save!
-      end
+        affiliations.each do |affiliation|
+          affiliation_copy = affiliation.dup
+          affiliation_copy.affiliable = event
+          affiliation_copy.save!
+        end
 
-      Event::ApplicationMailer.with(application: self).activated.deliver_later
+        Event::ApplicationMailer.with(application: self).activated.deliver_later
```
manu-gurudath: move `deliver_later` outside the transaction block, if the transaction subsequently rolls back or the job executes before the transaction commits the mailer job will reference data that doesnt exist or is an inconsistent state

for example see lines 139-154 in `organizer_position_invite.rb`, the `accept` method places `deliver_later` outside the transaction block

manuthecoder (changes requested):

manuthecoder (approved):
lgtm!

app/models/event/application.rb
```diff
@@ -322,25 +322,27 @@ def check_cosigner_update
     def activate_event!
       raise "Contract must be signed before activation" unless contract.signed?
 
-      poc = contract.party(:hcb).user
+      ActiveRecord::Base.transaction do
```
garyhtou:
```suggestion
      self.with_lock do
        raise ArgumentError.new("Event was already created") if event.present?
```

polypixeldev:
Tested lock change, works!

manuthecoder (approved):

garyhtou (approved):

PR #12900, opened 2026-02-05
Enable Rails 8 Regexp.timeout default for ReDoS

albertchae:
## Summary of the problem

Protect against ReDoS with a timeout

The second of 3 PRs to flip rails 8 framework defaults

### DIFF
--- a/config/initializers/new_framework_defaults_8_0.rb
+++ b/config/initializers/new_framework_defaults_8_0.rb
@@ -29,4 +29,4 @@ Rails.application.config.action_dispatch.strict_freshness = true
 ###
 # Set `Regexp.timeout` to `1`s by default to improve security over Regexp Denial-of-Service attacks.
 #++
-# Regexp.timeout = 1
+Regexp.timeout = 1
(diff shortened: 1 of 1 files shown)


config/initializers/new_framework_defaults_8_0.rb
```diff
@@ -29,4 +29,4 @@
 ###
 # Set `Regexp.timeout` to `1`s by default to improve security over Regexp Denial-of-Service attacks.
 #++
-# Regexp.timeout = 1
+Regexp.timeout = 1
```
albertchae, line 32: copying comment from https://github.com/hackclub/hcb/pull/10836/changes#r2190781041, but hcb does execute user provided regexes in some places, so a timeout is good https://github.com/hackclub/hcb/blob/d293e57763729f69a1612606e22095761650eccd/app/services/stripe_authorization_service/webhook/handle_issuing_authorization_request.rb#L96

garyhtou (approved):

garyhtou:
1 second time out is very reasonable. The only place where we're taking user input for REGEX is for approving card grant transactions. And if they're already taking > 1 sec, then the webhook would likely timeout anyways

garyhtou:
According to the docs, it will raise a `RuntimeError`, so we'll just keep and eye on it

PR #12907, opened 2026-02-06
[Migration] Fix 20260204193727_allow_null_in_affiliation_event.rb

garyhtou:
## Summary of the problem

Intended migration order:
1. 20260204193727_allow_null_in_affiliation_event.rb
2. 20260203031519_remove_event_id_from_affiliations.rb

In production, the migrations _did_ run in the intended order due to PR merge order.

However, due to timestamp order, 20260204193727_allow_null_in_affiliation_event is now causing errors because it attempts to change a column that was previously removed.

## Describe your changes

Only run migration is the column still exists


garyhtou:
After merging, let's enforce the reversible migration CI check. I think it's good enough now to require before PR merge

PR #12908, opened 2026-02-06
[v4] Block locked users from authorizing oauth applications

### LINKED ISSUE #12904: [v4] Prevent API authorizations for locked users
Locked users should not be able to authorize an OAuth app. Currently locked users can use the mobile app

Mohamad-Mortada:
Block locked users from authorizing oauth applications

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -358,7 +358,7 @@ class User < ApplicationRecord
     # Invalidate all sessions
     user_sessions.destroy_all
     # Invalidate all API tokens
-    api_tokens.update_all(revoked_at: Time.now)
+    api_tokens.accessible.update_all(revoked_at: Time.current)
   end
 
   def unlock!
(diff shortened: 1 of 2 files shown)


garyhtou (commented):
Can you also add this:

https://github.com/hackclub/hcb/issues/12904#issuecomment-3857371032

app/models/user.rb
```diff
@@ -357,6 +357,8 @@ def lock!
 
     # Invalidate all sessions
     user_sessions.destroy_all
+    # Invalidate all API tokens
+    api_tokens.update_all(revoked_at: Time.now)
```
garyhtou: We don't wanna override old data or modify already unusable tokens.
```suggestion
    api_tokens.accessible.update_all(revoked_at: Time.current)
```

garyhtou (approved):

app/models/user.rb
Mohamad-Mortada: ahhhhhh

PR #12909, opened 2026-02-06
[Ledger] Allow signed in users to filter

garyhtou:
## Summary of the problem

Shipwrecked is running into an issue where they would allow a non-organizer to filter their ledger.

## Describe your changes

Relax restrictions to allow signed in users to filter. Previously, it was only organizers who could filter ledgers.


Luke-Oldenburg (approved):
lgtm

Mohamad-Mortada (approved):

PR #12910, opened 2026-02-06
[Intercom] Use Public IDs to match mobile app

Mohamad-Mortada:
## Summary of the problem
Currently, mobile app registers user with public ID and rails does with ID causing a mismatch and duplicate user

## Describe your changes
Switches to use ID
Why I did it this way: 
https://github.com/intercom/intercom-rails/issues/106


config/initializers/intercom.rb
```diff
@@ -1,5 +1,15 @@
 # frozen_string_literal: true
 
+class IntercomUser
+  def self.from_authenticated_user(user)
+    {
+      id: user.public_id,
+      email: user.email,
+      name: user.name,
+    }
+  end
```
garyhtou, line 8: Can we double check that intercom's compatible with a hash? I'd assume they're likely calling `id` as a method, rather than accessing it as `[:id]`.

If it does call it as a method, then maybe consider using the `Data` class

Mohamad-Mortada: They are just calling `id` as a method. Switched to using `Data` class

garyhtou (approved):
This smells about right 👃

PR #12912, opened 2026-02-06
[Admin] Add Public ID lookup tool

garyhtou:
## Summary of the problem

As we use Public IDs more and more, including in Intercom, it will become more common for Ops to need to look up these user ids. It's also useful for engineers, too, so we don't need to open the Rails console if we want to convert a public ID to a numeric ID, such as for use in a blazer query

## Describe your changes

Add a Public ID look-up page. Adds better tests for PublicIdentifiable

### DIFF
--- a/app/models/concerns/public_identifiable.rb
+++ b/app/models/concerns/public_identifiable.rb
@@ -4,6 +4,42 @@
 module PublicIdentifiable
   extend ActiveSupport::Concern
 
+  # Central registry for public ID lookup
+  # Models still call set_public_id_prefix, this is just for admin lookup tool
+  # Lazy-loaded to avoid circular dependencies during Rails initialization
+  def self.models
+    @models ||= {
+      ach: AchTransfer,
+      act: PublicActivity::Activity,
+      apl: Event::Application,
+      bfe: BankFee,
+      cdg: CardGrant,
+      cdp: CheckDeposit,
+      chg: Api::Models::CardCharge,
+      chk: Check,
+      cmt: Comment,
+      crd: StripeCard,
+      don: Donation,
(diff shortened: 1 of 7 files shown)


app/models/concerns/public_identifiable.rb
```diff
@@ -4,6 +4,40 @@
 module PublicIdentifiable
   extend ActiveSupport::Concern
 
+  # Central registry for public ID for admin lookup tool
+  def self.models
```
Mohamad-Mortada, line 8: Are we able to programmatically to pull a list of all models that use the Public ID? The tests seem to be doing that but it may also be more expensive im assuming. I just don't like that every time we make a new public ID in the model we also gotta update it here

Mohamad-Mortada (approved):
Besides the comment above it LGTM :)

24c02:
[this](https://github.com/hackclub/theseus/blob/2e8cb65768cde267fe91632ebfb51918724a4d6d/app/controllers/public_ids_controller.rb#L41-L52) is how i avoid maintaining a list of models - it's hacky but might be worth stealing pieces of?

garyhtou:
> [this](https://github.com/hackclub/theseus/blob/2e8cb65768cde267fe91632ebfb51918724a4d6d/app/controllers/public_ids_controller.rb#L41-L52) is how i avoid maintaining a list of models - it's hacky but might be worth stealing pieces of?

that was my intuition on how to build this out. however, from my understanding, it requires preloading the app in order for it to work in development since classes are lazy loaded

garyhtou:
One of my goals for this PR was to remove the need for `set_public_id_prefix`, but it looks like i haven't done that yet

PR #12913, opened 2026-02-06
[v4 API] Block locked users

24c02:
Right now, if you have an active API token and your account gets locked - #12908 blocks new authorizations, but still allows existing tokens to function (because tokens aren't tied to the sessions that get destroyed).
This fixes that by erroring if the user is locked when we authorize against the token.


24c02:
nvm - i missed that #12908 invalidates tokens >_<

PR #12914, opened 2026-02-06
[Schema] Add unique indexes to `card_grants`

Luke-Oldenburg:
Prod data is okay:
https://hcb.hackclub.com/blazer/queries/1096-duplicate-disbursement-on-cardgrant
https://hcb.hackclub.com/blazer/queries/1095-duplicate-stripecard-on-cardgrant
https://hcb.hackclub.com/blazer/queries/1094-duplicate-subledger-on-cardgrant


app/models/card_grant.rb
```diff
@@ -84,6 +84,10 @@ class CardGrant < ApplicationRecord
     self.expiration_at ||= CardGrantSetting.expiration_preferences[card_grant_setting.expiration_preference].days.from_now
   end
 
+  validates :disbursement, uniqueness: true
+  validates :stripe_card, uniqueness: true, allow_nil: true
+  validates :subledger, uniqueness: true
```
Luke-Oldenburg: The `belongs_to` has all of these as optional but all of the rows have a value for `disbursement` and `subledger`. Does this mean that:
A) the optional should be removed for those two.
B) there should be an `allow_nil: true` here
C) do nothing

garyhtou: and the answer is 🥁🥁🥁🥁🥁 `A`!! ding ding ding 🔔

garyhtou (approved):

garyhtou (approved):

app/models/card_grant.rb
Luke-Oldenburg: Gary and I discussed more and discovered it was B because of the order model hooks run in.

garyhtou (commented):
We added `allow_nil` because the `belong_to`s have an `optional: true`. In practice, all three associations are created in either a `before_create` or `after_create`, which means they always exist in the DB.

But during the validation step, these associations do not exist.

PR #12920, opened 2026-02-06
[CardGrant] Fix expiry notification mailer

Luke-Oldenburg:
## Summary of the problem

It was matching the exact timestamp instead of the date so these emails never sent.

## Describe your changes

Use Postgres `DATE` function and also add missing preview.


manuthecoder (approved):
lgtm!

garyhtou (approved):
good find!

PR #12925, opened 2026-02-07
[Code Quality] Introduce HasHcbCode Concern

manuthecoder:
This PR improves the code quality by consolidating repeated code into a HasHcbCode Concern


Luke-Oldenburg (commented):
Can you split this into multiple PRs please.

app/models/ach_transfer.rb
```diff
@@ -342,13 +339,8 @@ def canonical_transactions
...
 
-  def hcb_code
-    "HCB-#{TransactionGroupingEngine::Calculate::HcbCode::ACH_TRANSFER_CODE}-#{id}"
-  end
-
-  def local_hcb_code
-    @local_hcb_code ||= HcbCode.find_or_create_by(hcb_code:)
-  end
+  include HasHcbCode
+  has_hcb_code TransactionGroupingEngine::Calculate::HcbCode::ACH_TRANSFER_CODE, eager_create: true
```
garyhtou, line 343: We should move this to where the `after_create :local_hcb_code` so that we don't change the order of the callbacks

garyhtou (approved):

PR #12928, opened 2026-02-07
[Applications] Bug fixes and improvements based on feedback

YodaLightsabr:
## Describe your changes

* Fix issues arising from lack of safe navigation
* Add archiving to applications
* Create new `active_applications` scoped association on `User`
* Fix various UI bugs
* Ensure every apply layout page has a proper footer
* Remove redundant `if` statement on applications index


app/models/user.rb
```diff
@@ -85,6 +85,7 @@ class User < ApplicationRecord
 
   has_many :logins
   has_many :applications, class_name: "Event::Application", inverse_of: :user
+  has_many :active_applications, ->(user) { where(event: user.events, archived_at: nil) }, class_name: "Event::Application", inverse_of: :user
```
polypixeldev: This is only including applications that have been activated, but this relation is used for applications that are in progress. Could modify the relation here to use
```suggestion
  has_many :active_applications, ->(user) { where(archived_at: nil, event_id: nil) }, class_name: "Event::Application", inverse_of: :user
```
, or I think it would be better to use the `in_progress` scope on `Event::Application` (although it needs to be modified to not include archived applications.

app/views/static_pages/index.html.erb
```diff
@@ -138,7 +138,7 @@
             <li class="overflow-visible card card--hover flex flex-row justify-between items-center gap-2">
               <span>
                 <%= status_badge "info" %>
-                <strong><%= invite.event.name %></strong>
```
polypixeldev, line 141: How is `invite.event` potentially `nil` since `event_id` is a non-null column?

app/models/user.rb
YodaLightsabr: Good catch, I was doing the opposite of what I should've been doing. I've switched this to `user.applications.active` since we don't need any user data.

app/views/static_pages/index.html.erb
YodaLightsabr: Sorry. Shouldn't have pushed this. Bad local data.

app/models/event/application.rb
```diff
@@ -195,6 +199,13 @@ def next_step
       return "" if rejected?
     end
 
+    def archive_action_name
+      return "Archive application" if rejected? || approved?
+      return "Withdraw application" if submitted? || under_review?
```
garyhtou: Since `withdrew` is not a separate state in AASM, we'll need to somehow communicate to Airtable/Ops that this application was withdrawn.

I think we should keep that state separate from `rejected` to have more accurate stats and to prevent ops confusion ("why was this application rejected? I don't see anything wrong with it.")

garyhtou (approved):

app/controllers/event/applications_controller.rb
```diff
@@ -241,7 +241,7 @@ def user_params
     end
 
     def record_pageview
-      if Event::Application.last_page_vieweds.keys.include?(action_name.to_s)
+      if Event::Application.last_page_vieweds.keys.include?(action_name.to_s) && @application.user == current_user
```
garyhtou, line 244: good catch

app/jobs/event/application_sync_to_airtable_job.rb
```diff
@@ -0,0 +1,44 @@
...
+      app["City"] = application.address_city
+      app["State"] = application.address_state
+      app["Address Country"] = application.address_country
+      app["Event Location"] = application.address_country
+      app["How did you hear about HCB?"] = application.referrer
+      app["Accommodations"] = application.accessibility_notes
+      app["(Adults) Political Activity"] = application.political_description
+      app["Referral Code"] = application.referral_code
+      app["HCB Status"] = application.aasm_state.humanize unless application.draft?
+      app["Synced from HCB at"] = Time.current
```
garyhtou, line 35: nice!

app/jobs/event/application_sync_to_airtable_job.rb
```diff
@@ -0,0 +1,44 @@
...
+    queue_as :low
+
+    def perform(application)
+      return if application.draft?
+
+      app = ApplicationsTable.all(filter: "{recordID} = \"#{application.airtable_record_id}\"").first if application.airtable_record_id.present?
+      app ||= ApplicationsTable.all(filter: "{HCB Application ID} = \"#{application.hashid}\"").first
+      app ||= ApplicationsTable.new("HCB Application ID" => application.hashid)
+
+      app["First Name"] = application.user.first_name
```
garyhtou, line 14: We need to inform ops that changes should be made in HCB; otherwise it will be overriden

app/views/events/settings/_affiliation_form.html.erb
```diff
@@ -67,13 +67,16 @@
...
-        <%= form.number_field :size, placeholder: "10", step: 1, required: true, class: "!p-0 !border-none !shadow-none min-h-auto", disabled:, "x-model": "people" %>
+        <%= form.number_field :size, placeholder: "10", step: 1, min: 0, required: true, class: "!p-0 !border-none !shadow-none min-h-auto", disabled:, "x-model": "people" %>
         <span class="bold muted" x-text="people == 1 ? 'person' : 'people'">people</span>
       </div>
     </div>
   </div>
 </template>
 
+<template x-if="type == 'hack_club'">
+  <p class="p-0 text-sm">Select this only if you're a local Hack Club chapter. If you're running an official project for Hack Club HQ, don't use this application form.</p>
```
garyhtou, line 78: low priority: instead of "Don't use this application form", we should tell me explicitly where they should go

garyhtou (approved):

PR #12941, opened 2026-02-09
[Card Grants] Add padding below pagination in table

manuthecoder:
Fixes the bottom padding 

image


app/views/card_grants/card_index.html.erb
```diff
@@ -49,6 +49,7 @@
       </table>
 
       <%= paginate @paginated_card_grants %>
+      <div class="mb-5"></div>
```
Luke-Oldenburg: Can we do this without an empty div?

garyhtou (approved):

PR #12950, opened 2026-02-11
[Policy] Use `OrganizerPosition#role_at_least?` universally

Luke-Oldenburg:
## Summary of the problem

`OrganizerPosition#find_by` and `record.users.include?` do not account for sub-organizations properly.
closes #12933

### DIFF
--- a/app/policies/event_policy.rb
+++ b/app/policies/event_policy.rb
@@ -220,7 +220,7 @@ class EventPolicy < ApplicationPolicy
   end
 
   def receive_grant?
-    record.users.include?(user)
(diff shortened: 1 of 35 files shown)


app/policies/event_policy.rb
```diff
@@ -220,7 +220,7 @@ def toggle_event_tag?
   end
 
   def receive_grant?
-    record.users.include?(user)
+    OrganizerPosition.role_at_least?(user, record, :reader)
   end
```
garyhtou, line 224: This is an unused policy, let's get rid of it in a separate PR

garyhtou (approved):
nice!

PR #12954, opened 2026-02-11
Fix email validation to require TLD

### LINKED ISSUE #12905: [Donations] Validate email address to include TLD before passing to Stripe
Stripe requires email addresses to have a TLD, but the email spec doesn't require email addresses to have one. As a result, front-end out of the box email validaiton allows incomplete emails to go through. This results in a Stripe::InvalidRequestError and a 500 response.

manuthecoder:
Closes #12905 
> Stripe requires email addresses to have a TLD, but the email spec doesn't require email addresses to have one. As a result, front-end out of the box email validaiton allows incomplete emails to go through. This results in a Stripe::InvalidRequestError and a 500 response.


garyhtou (changes requested):
Let's use `validates_email_format_of` since we already have that gem installed

garyhtou (changes requested):
What if we already have data in the database that does not meet `validates_email_format_of`'s requirements? It would likely cause the record to fail to save even when updating unrelated (non-email) columns.

Approaches we can take:
1. Only validate when email is changed
2. Only validate on create, then scan the prod DB, fix all bad prod data (might not always be possible), then always validate (not just on create)

manuthecoder:
`#1` is the simpler solution. We can use the rails dirty tracking method to do this

garyhtou (approved):

PR #12959, opened 2026-02-12
Remove ignore column for affiliation event_id

polypixeldev:
## Summary of the problem

The `event_id` column has been dropped and code deployed, so we no longer need this line.

## Describe your changes

Remove old ignore column line.


Luke-Oldenburg (approved):
lgtm

garyhtou (approved):

PR #12961, opened 2026-02-12
Disallow age < 13

kimorv:
## Summary of the problem

Changes made to disallow age entered from being less than 13

## Describe your changes

Had to check if age is less than 13 and declare an error if it is


manuthecoder (approved):

Luke-Oldenburg:
See also #12780

PR #12973, opened 2026-02-13
Revert "Revert "[Intercom] Use Public IDs to match mobile app""

garyhtou:
Reverts hackclub/hcb#12972

Originally reverted because we found that users should not start new conversations after merging orignal PR.

However, after reverting (in hackclub/hcb#12972), we found it was still the case, and the setting in Intercom for allowing customers to start new convos was actually turned off.

So, we're now bringing back the original PR.

### DIFF
--- a/config/initializers/intercom.rb
+++ b/config/initializers/intercom.rb
@@ -1,5 +1,14 @@
 # frozen_string_literal: true
 
+# Proxied keys pulled from here:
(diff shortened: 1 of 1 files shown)


Mohamad-Mortada (approved):
I approve of reverting a revert PR which was meant to revert a reverted PR which shouldn't have been reverted

PR #12976, opened 2026-02-13
Allow admins to set event tags and risk level when activating an application

polypixeldev:
## Summary of the problem

Previously, when activating an application from Airtable, we automatically set the organized by teens and robotics team tags, but admins had to manually go set any others after activating. With the new applications flow, we are currently not setting any tags automatically.

## Describe your changes

Add a form to the HCB contract signing page for an application that has a multiple select menu to choose tags. Organized by teens is selected by default based on `teen_led`, and robotics team and hack club are selected by default based on affiliations.

image
image

### DIFF
--- a/app/assets/stylesheets/components/_forms.scss
+++ b/app/assets/stylesheets/components/_forms.scss
@@ -74,8 +74,13 @@ textarea {
   }
 }
 
-[data-dark='true'] option {
+[data-dark='true'] select:not([multiple]) option {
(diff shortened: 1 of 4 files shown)


garyhtou (commented):
Add risk level and merge.

In the future, we'll add a page before HCB party signing similar to the existing "Activate" page.

garyhtou (approved):

PR #12987, opened 2026-02-16
Add user_id to AppSignal errors

YodaLightsabr:
## Summary of the problem

We're being hit with some abuse, but this should make it easier to track down.

## Describe your changes

Start tracking error user ID in AppSignal


Luke-Oldenburg (approved):
lgtm

PR #12988, opened 2026-02-16
Prevent 3+ updates to user phone number in <24 hours

YodaLightsabr:
## Summary of the problem

We're getting hit with a lot of SMS abuse

## Describe your changes

Prevent frequent user phone number changes

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -300,6 +300,11 @@ class UsersController < ApplicationController
       end
     end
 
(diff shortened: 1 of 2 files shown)


app/controllers/users_controller.rb
```diff
@@ -300,6 +300,11 @@ def update
       end
     end
 
+    if @user.phone_number_changed? && @user.phone_number_update_count(since: 24.hours.ago) >= 2 && !admin_signed_in?
```
Luke-Oldenburg, line 303:
```suggestion
    if @user.phone_number_changed? && @user.phone_number_update_count(since: 24.hours.ago) > 2 && !auditor_signed_in?
```

PR #12992, opened 2026-02-17
Fix tour

polypixeldev:
## Summary of the problem

UI3 seems to have broken the tour for a couple reasons
- The tour starts on the transactions page instead of the home page, which is now the default
- We weren't properly passing through the `data` hash to the dock items, which tell the tour where to anchor the steps.

## Describe your changes

Move the tour rendering to the event home page and properly pass through the data hash.

The tour is still a little broken on mobile, since the cards can't align to the tab bar items like before UI3, but it still renders in the top.


Luke-Oldenburg (approved):
lgtm

manuthecoder (approved):

garyhtou (approved):

PR #12994, opened 2026-02-17
[Login] Prompt TOTP after successful webauthn auth

Lazylllama:
## Summary of the problem
Having to manually press the button to change to another auth method then having to press the TOTP button and finally being able to enter the code you had from the beginning. No one wants to have to open their email to get a code and have to wait for a slow email delivery. Most people have their TOTP codes at hand much faster, especially those already using security keys...

## Describe your changes

Prompt for TOTP rather than email code after a successful webauthn authentication since this is probably a more common flow.


Lazylllama:
_(This could also work by just prioritizing TOTP always, but I didnt want to make that big of a change since that impacts many, many more users.)_

Lazylllama:
275 commits and 2 months later, someone has already fixed this. How did no-one even review this in all that time?

garyhtou:
Hi @Lazylllama! Sorry about that. It looks like this PR and a few others slipped through the cracks in the sea of notifications we receive. 2 months is definitely too long of a time to be waiting for a review. We'll keep a closer eye on PRs awaiting review- and feel free to remind us if you'd been waiting for a while.

Thank you for your willingness to contribute

PR #12998, opened 2026-02-18
Fix transfer flash

manuthecoder:
Fixes flash being hidden by navbar

image

### DIFF
--- a/app/views/layouts/transfer.html.erb
+++ b/app/views/layouts/transfer.html.erb
@@ -1,5 +1,6 @@
 <% page_full %>
 <% @no_app_shell = true %>
+<% @hide_flash = true %>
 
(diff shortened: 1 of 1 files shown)


app/views/layouts/transfer.html.erb
```diff
@@ -1,5 +1,6 @@
 <% page_full %>
 <% @no_app_shell = true %>
+<% @hide_flash = true %>
```
Luke-Oldenburg, line 3: Where is this used?

manu-gurudath: `app/views/layouts/application.html.erb`

Luke-Oldenburg: We should phase out using instance variables like this in the future but it's fine for now.

Luke-Oldenburg (approved):
lgtm

garyhtou (approved):

PR #13000, opened 2026-02-18
[OPI] Fix accepted mailer for reader invites

Luke-Oldenburg:
## Summary of the problem

We have to compare as a string and not a symbol.


hanaeatsplanes:
13k prs

garyhtou (approved):

PR #13001, opened 2026-02-19
Fix issues with teenager checks in applications

polypixeldev:
## Summary of the problem

- We were keeping the cosigner email, even if the user updated their birthday to not be a minor anymore, causing a contract to be created with a cosigner for an adult.
- We were checking for `teenager?` (inclusive of 18) instead of minor (exclusive of 18) when rendering the cosigner email in a couple places
- We were mixing checking application `teen_led?` (set when they create the application) with the user's birthday (can be updated at any time), causing potential issues where the code doesn't agree on what flow they should go through

## Describe your changes

- Set the cosigner email to `nil` on contract create if the user is not a minor
- Add a `is_minor?` method to the user model that checks for _under_ 18 and use it when checking cosigner email
- Update `teen_led` at application submission and use that instead of `user.teenager?`


app/models/event/application.rb
```diff
@@ -117,7 +117,9 @@ class Application < ApplicationRecord
       event :mark_submitted do
         transitions from: :draft, to: :submitted
         after do
-          if user.teenager?
+          update!(teen_led: user.is_teenager?)
```
Luke-Oldenburg, line 120: Do we want this to use `user#is_minor?`. Is it okay to make teen led applications that have an adult signee?

polypixeldev: Our current definiton of teenager is 18 or under, so this should use `is_teenager?`. The only difference for 18 year olds is that they don't need a cosigner since they're no longer a minor.

Luke-Oldenburg (approved):
lgtm

garyhtou (approved):

PR #13002, opened 2026-02-19
Require database lock when updating contract AASM state

polypixeldev:
## Summary of the problem

There's a small risk of a race condition where we end up running AASM callbacks for the contract being signed twice - once with `Contract::Party#sync_with_docuseal`, and another via the DocuSeal webhook.

## Describe your changes

Require locking when updating AASM state for contracts (https://github.com/aasm/aasm#pessimistic-locking)


manuthecoder (approved):
lgtm!

Luke-Oldenburg (approved):
lgtm

garyhtou (approved):

PR #13004, opened 2026-02-19
Sync application org type in Airtable

polypixeldev:
## Summary of the problem

The old application form was setting the "Org Type" field in Airtable for robotics orgs - we can do this for robotics and hack club orgs.

## Describe your changes

Set application org type in Airtable based on application affiliations.


Luke-Oldenburg (approved):
lgtm

garyhtou (approved):

PR #13005, opened 2026-02-19
Void contract on application rejection

polypixeldev:
## Summary of the problem

Contracts would be left as sent when we reject - let's keep the data tidy

## Describe your changes

Void the contract (if present, which would only be for teen-led applications) when we reject an application.


manuthecoder (approved):

Luke-Oldenburg (approved):
lgtm

garyhtou (approved):

PR #13016, opened 2026-02-20
[IncreaseCheck] Add stop and reissue buttons to UI

Luke-Oldenburg:
## Summary of the problem

Before Ops would have to either go into Column to stop a check or ask the engineering team to reissue.

## Describe your changes

Adds buttons to the UI to perform these tasks.


polypixeldev (approved):

app/models/increase_check.rb
```diff
@@ -260,22 +262,24 @@ def send_check!
     mark_approved!
   end
 
-  def reissue!
-    return unless column_id.present? && (column_issued? || column_stopped?)
+  def stop!
+    # https://column.com/docs/api/#check-transfer/stop
+    return if column_id.nil? || !can_stop?
```
garyhtou: We should raise an error

app/models/increase_check.rb
```diff
@@ -260,22 +262,24 @@ def send_check!
...
-      column_object: nil,
-      check_number: nil,
-      column_status: nil,
-      column_delivery_status: nil,
+      column_id: column_check["id"],
+      column_object: column_check,
+      check_number: column_check["check_number"],
+      column_status: column_check["status"],
+      column_delivery_status: column_check["delivery_status"],
     )
```
```suggestion
      column_object: column_check,
      column_status: column_check["status"],
      column_delivery_status: column_check["delivery_status"],
)
```

app/models/increase_check.rb
```diff
@@ -260,22 +262,24 @@ def send_check!
...
+  end
 
-    send_column!("reissue_#{stopped_id}")
+  def reissue!
```
garyhtou, line 342:
```suggestion
  def reissue!
    raise NotImplementedError
```

garyhtou (approved):

app/models/increase_check.rb
```diff
@@ -262,24 +311,49 @@ def send_check!
...
+      user_id:,
+      memo:,
+      amount:,
+      payment_for:,
+      recipient_name:,
+      address_line1:,
+      address_line2:,
+      address_city:,
+      recipient_email:,
+      send_email_notification:, # TODO: disable email notification and send reissue mailer
```
garyhtou: Keep `send_email_notification` as is, but add a `reissued_for` to handle this logic

app/models/increase_check.rb
```diff
@@ -262,24 +311,49 @@ def send_check!
...
+      address_zip:,
+      payment_recipient_id:,
+    )
+
+    if reissued_check.save
+      # TODO: copy receipts
+    end
```
garyhtou: Let's bang this so we don't continue on if it fails to save

app/models/increase_check.rb
```diff
@@ -262,24 +311,49 @@ def send_check!
...
-      column_id: column_check["id"],
       column_object: column_check,
-      check_number: column_check["check_number"],
       column_status: column_check["status"],
       column_delivery_status: column_check["delivery_status"],
     )
   end
 
   def reissue!
+    stop!
```
garyhtou: We should probably allow calling `reissue!` even if the check was previously stopped.

So, maybe say `stop! unless stopped?`

app/models/increase_check.rb
```diff
@@ -260,11 +265,9 @@ def increase?
...
   def state
-    if column?
-      :info
-    elsif pending?
+    if pending?
       :muted
-    elsif rejected? || increase_canceled? || increase_stopped? || increase_returned? || increase_rejected?
+    elsif rejected? || increase_canceled? || increase_stopped? || increase_returned? || increase_rejected? || column_pending_stop? || column_stopped? || column_rejected?
       :error
     elsif increase_deposited?
```
Luke-Oldenburg:
```suggestion
    elsif increase_deposited? || column_settled?
```

app/models/increase_check.rb
```diff
@@ -316,22 +319,50 @@ def send_check!
...
+      recipient_name:,
+      address_line1:,
+      address_line2:,
+      address_city:,
+      recipient_email:,
+      send_email_notification:,
+      address_zip:,
+      payment_recipient_id:,
+      reissued_for_id: id,
     )
```
garyhtou, line 360: Write a test:

When reissuing, the test compares a list of derived columns between the new and old checks and ensure they match.

The list of columns should be derived by:

`original_check.attributes.keys - %w[id reissued_for]`

The goal is to define a list of columns that won't match.

This test should catch the case where we add a new column to IncreaseCheck, but forget to update `reissue` to copy the column.

app/models/receipt.rb
```diff
@@ -212,6 +212,17 @@ def extracted_incorrect_merchant?
     false
   end
 
+  def self.reupload(old_receiptable:, new_receiptable:)
+    where(receiptable: old_receiptable).map do |receipt|
```
```suggestion
    where(receiptable: old_receiptable).find_each do |receipt|
```

garyhtou (approved):

garyhtou:
Write the test and then merge

PR #13017, opened 2026-02-21
Add reissue contract button to HCB's contract signing page

polypixeldev:
## Summary of the problem

Sometimes, signees or cosigners will fill out a contract wrong (such as uploading an invalid form of ID). In these cases, it'd be nice for ops to have an easy way to reissue a new contract to them and tell them what they need to fix.

## Describe your changes

Adds a new "Reissue" button to the sidebar on HCB's signing page. This will open a modal where they can add messages to include in the emails sent to the signee and cosigner, if present. Reissuing will void the current contract and create a new identical contract.

image


app/views/contract/parties/show.html.erb
```diff
@@ -7,33 +7,70 @@
...
+
+    <%= form_with url: reissue_contract_path(@contract) do |form| %>
+      <div class="field">
+        <%= form.label :signee_message, "Signee message" %>
+        <%= form.text_area :signee_message, placeholder: "Leave blank if the signee does not need to make any changes", rows: 2, required: @contract.party(:cosigner).nil?, data: { behavior: "no_autosize" } %>
+      </div>
+
+      <% if @contract.party(:cosigner).present? %>
+        <div class="field">
+          <%= form.label :cosigner_message, "Cosigner message" %>
```
garyhtou, line 73: Make it clear this supports MD

app/controllers/contracts_controller.rb
```diff
@@ -1,13 +1,60 @@
...
+        external_service: @contract.external_service,
+        include_videos: @contract.include_videos,
+        prefills: @contract.prefills,
+        type: @contract.type,
+        external_template_id: @contract.external_template_id
+      )
+
+      @contract.parties.not_hcb.each do |party|
+        new_contract.parties.create!(external_email: party.external_email, role: party.role, user: party.user)
+      end
```
garyhtou: Instead of copying from the old contract (since source data from contractable may change), let's:

1. Create a standard interface for contractable: `send_contract`
2. Call it here

Ensure `send_contract` is idempotent. It doesn't send another contract unless the previous one doesn't exist or is voided.

garyhtou (approved):

PR #13025, opened 2026-02-21
[Performance] Optimize `active_teenagers`

manuthecoder:
This pull request updates the `active_teenagers` method in the `Event` model to improve its efficiency and accuracy. The method now uses a direct query on the `users` association instead of iterating through `organizer_positions`, which should result in faster and more reliable results.

User query optimization:

* The `active_teenagers` method now queries `users` with `teenager: true` and `active` scope, using `distinct.count` for improved performance and accuracy.


app/models/event.rb
```diff
@@ -872,7 +872,7 @@ def set_airtable_status(status)
   end
 
   def active_teenagers
-    organizer_positions.joins(:user).count { |op| op.user.is_teenager? && op.user.active? }
+    users.where(teenager: true).active.distinct.count
```
```suggestion
    users.active_teenager.count
```

app/models/event.rb
```diff
@@ -872,7 +872,7 @@ def set_airtable_status(status)
   end
 
   def active_teenagers
```
garyhtou, line 874: unrelated: We should rename this method to make it clear that this is a count (integer) and not a list of users

garyhtou (approved):

PR #13028, opened 2026-02-21
[Code Quality] Improve condition checks

manuthecoder:
`.size == 0` / `.size >= 1` / `.count == 0 `→ `.none?` / `.any?` / `.empty?`


app/views/user/backup_code_mailer/code_used.html.erb
```diff
@@ -4,7 +4,7 @@
 
 <p>
   You've just used one of your backup codes to login to
-  <% if @user.backup_codes.active.size == 0 %>
+  <% if @user.backup_codes.active.empty? %>
     HCB and have none left. We strongly urge you to regenerate more codes at your earliest convenience in your <%= link_to "user settings", security_user_url(@user) %> to ensure you can always access your account.
   <% elsif @user.backup_codes.active.size <= 3 %>
```
garyhtou, line 9: This here is an unintuitive case where `size == 0` or `size.zero?` is more performant than `empty?`.

First, some context: `empty?` (when the collection is not loaded into memory yet) will an SQL statement similar to:

```sql
SELECT 1 AS one FROM "user_backup_codes" WHERE "user_backup_codes"."user_id" = 1803 AND "user_backup_codes"."aasm_state" = 'active'
```

Notice how it does a `SELECT 1` to check for existence, rather than a `COUNT`? This is more performance when **_all_** you want to do is check for existence. But, we want more than just an existence check because on line 9, we call `count` to get a number (not just "do any exist?").

This means this section of code (line 7 and 9) triggers two SQL statements:

```sql
# existence check from line 7
SELECT 1 AS one FROM "user_backup_codes" WHERE "user_backup_codes"."user_id" = 1803 AND "user_backup_codes"."aasm_state" = 'active'

# count from line 9
SELECT COUNT(*) FROM "user_backup_codes" WHERE "user_backup_codes"."user_id" = 1803 AND "user_backup_codes"."aasm_state" = 'active'
```

Since we're already gonna need the count on line 9, if we use `size` (to get the count) on line 7, we save ourselves one SQL statement by not doing the existence check.

---

You can test this in the Rails console by doing

```ruby
# log SQL statements so they're visible in the rails console
ActiveRecord::Base.logger = Logger.new($stdout)

User.find(1803).backup_codes.active.empty?

User.find(1803).backup_codes.active.size.zero?
```

Luke-Oldenburg (commented):
Could you please also search for `size > 0`? I'm not sure if we use that anywhere. Also, what is the difference between `none?` and `empty?` in the context here? To my knowledge they should be the same thing when being performed on an ActiveRecord.

app/views/user/backup_code_mailer/code_used.html.erb
```diff
@@ -4,7 +4,7 @@
 
 <p>
   You've just used one of your backup codes to login to
-  <% if @user.backup_codes.active.size == 0 %>
+  <% if @user.backup_codes.active.empty? %>
```
Luke-Oldenburg:
```suggestion
  <% if @user.backup_codes.active.size == 0 %>
```

Luke-Oldenburg (approved):
lgtm

PR #13039, opened 2026-02-25
[Schema] Make timestamps on `CardGrantSetting` nonnull

Luke-Oldenburg:

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -12,7 +12,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -488,7 +488,7 @@
     t.bigint "disbursement_id"
     t.string "email", null: false
     t.bigint "event_id", null: false
-    t.datetime "expiration_at"
+    t.datetime "expiration_at", precision: nil
```
garyhtou:
```suggestion
    t.datetime "expiration_at"
```

garyhtou (approved):

PR #13043, opened 2026-02-25
[Code Quality] Introduce `PubliclyTrackable` concern

manuthecoder:
Introduces a standard PublicActivity tracking configuration:
   - Improves AchTransfer, CheckDeposit, Employee, IncreaseCheck, Invoice, PaypalTransfer, Wire, WiseTransfer
   - Replaces boilerplate `include PublicActivity::Model` + `tracked` declarations with a single include


garyhtou (changes requested):
Ideally, the concern has the flexibility to support all `tracked` method calls.

I think it adds complexity when we have to decide whether or not to use `PubliclyTrackable` when trying to use `PublicActivity::Model`.

PR #13045, opened 2026-02-25
[AASM] Make `ach_transfers` table's `aasm_state` column not null

rluodev:
## Summary of the problem

Previously, was null.

## Describe your changes

Now, not null. Confirmed via Blazer on 2/25 that there are no null entries in this column on prod.


manuthecoder (approved):
LGTM!

Luke-Oldenburg (approved):
lgtm

garyhtou (approved):

PR #13050, opened 2026-02-25
[AASM] Make `announcements` table's `aasm_state` column not null

rluodev:
## Summary of the problem

previously, was null

## Describe your changes

now, not null. Confirmed via Blazer on 2/25 that there are no null entries in this column on prod.


db/migrate/20260225215720_make_announcement_aasm_state_not_null.rb
```diff
@@ -0,0 +1,10 @@
+class MakeAnnouncementAasmStateNotNull < ActiveRecord::Migration[8.0]
+  def up
+    safety_assured do
+      change_column_null :announcements, :aasm_state, false
+    end
```
garyhtou: Let's follow the instructions that Strong Migrations gives:

```
=== Dangerous operation detected #strong_migrations ===

Setting NOT NULL on an existing column blocks reads and writes while every row is checked.
Instead, add a check constraint and validate it in a separate migration.

class MakeAnnouncementAasmStateNotNull < ActiveRecord::Migration[8.0]
  def change
    add_check_constraint :announcements, "aasm_state IS NOT NULL", name: "announcements_aasm_state_null", validate: false
  end
end

class ValidateMakeAnnouncementAasmStateNotNull < ActiveRecord::Migration[8.0]
  def up
    validate_check_constraint :announcements, name: "announcements_aasm_state_null"
    change_column_null :announcements, :aasm_state, false
    remove_check_constraint :announcements, name: "announcements_aasm_state_null"
  end

  def down
    add_check_constraint :announcements, "aasm_state IS NOT NULL", name: "announcements_aasm_state_null", validate: false
    change_column_null :announcements, :aasm_state, true
  end
end
```

garyhtou (changes requested):

Luke-Oldenburg (approved):
lgtm

garyhtou (approved):

PR #13052, opened 2026-02-26
Revert "[StripeCard] Validate presence of subledger and card_grant"

garyhtou:
Reverts hackclub/hcb#12956


garyhtou:
This broke activation of card grants

PR #13053, opened 2026-02-26
Use `url_from` to sanitize redirect URLs in controllers

manuthecoder:
image

There are some instances where open redirects are possible

### DIFF
--- a/app/controllers/receipts_controller.rb
+++ b/app/controllers/receipts_controller.rb
@@ -59,7 +59,7 @@ class ReceiptsController < ApplicationController
         end
 
         if params[:redirect_url]
-          redirect_to params[:redirect_url]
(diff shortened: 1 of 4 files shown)


app/controllers/receipts_controller.rb
```diff
@@ -59,7 +59,7 @@ def link
         end
 
         if params[:redirect_url]
-          redirect_to params[:redirect_url]
+          redirect_to url_from(params[:redirect_url]) || @receiptable.try(:hcb_code) || @receiptable
```
garyhtou: Shouldn't the `else` handle this case?

If necessary, we could add a `present?` check on the `if`

garyhtou (approved):

PR #13062, opened 2026-02-26
[Attach Receipt Page] Flash missing top margin

manuthecoder:
Closes #13054 by fixing margins and moving application flash to under heading for consistency 

image


Luke-Oldenburg:
Could you please attach a screenshot of where the flash appears? Also could it use our application font?

YodaLightsabr (approved):

PR #13068, opened 2026-02-26
[Donation Receipt] Use "payment" when non-tax deductible

garyhtou:
## Summary of the problem

We already subbed out the word "donation" for "payment" when it's non-tax-deductible. But we're still saying "gift".

## Describe your changes

I'm subbing out the word "gift" for "payment". cc @MelSmith104 

Mel suggested "purchase". I chose "payment" since it feels more flexible than "purchase", but can make it "purchase" if we want.

PR also includes formatting for the file. Review each individual commit for an easier diff


YodaLightsabr (commented):
LGTM

PR #13070, opened 2026-02-27
[Applications] Add country restrictions

YodaLightsabr:
## Summary of the problem

We want to restrict the countries that users can apply for HCB from.

## Describe your changes

Add a warning when a user enters an unsupported country.


app/models/event/application.rb
```diff
@@ -158,6 +158,8 @@ class Application < ApplicationRecord
 
     scope :in_progress, -> { where.not(aasm_state: ["approved", "rejected"]) }
 
+    DISALLOWED_COUNTRIES = %w[IN NG RU CU IR KP SY BY VE SD SS MM AF YE SO PK]
```
garyhtou: Should we add a rails validation?

app/views/event/applications/personal_info.html.erb
```diff
@@ -99,10 +99,22 @@
         <%= form.label :address_country do %>
           Country <span class="text-primary">*</span>
         <% end %>
-        <%= form.country_select :address_country, { priority_countries: ["US"], priority_countries_divider: "", include_blank: "Select a country", required: true }, { class: "!min-h-[36px]" } %>
+        <%= form.country_select :address_country, { priority_countries: ["US"], priority_countries_divider: "", include_blank: "Select a country", required: true }, { class: "!min-h-[36px]", "x-model": "country", "x-on:change": "$el.setCustomValidity(disallowed_countries.includes(country) ? 'HCB is not supported in this country' : '')" } %>
+        <% if Event::Application::DISALLOWED_COUNTRIES.include?(@application.address_country) %>
+          <script>
+            event_application_address_country.setCustomValidity('HCB is not supported in this country')
+          </script>
```
polypixeldev: Could this be a `x-init` directive, to not mix Alpine and custom script tags?

app/views/event/applications/personal_info.html.erb
```diff
@@ -14,7 +14,7 @@
 <% end %>
 
 <%= form_with model: @application, url: application_path(@application), method: :patch, id: "edit_application_#{@application.id}", class: "mt-7", data: { action: "input->application-autosave#save", "application-autosave-target": "form" } do |form| %>
-  <div class="card mb-6" x-data="{ minor: <%= @application.user.is_minor? || false %> }">
+  <div class="card mb-6" x-data="{ minor: <%= @application.user.is_minor? || false %>, country: '<%= @application.address_country.presence %>', disallowed_countries: '<%= Event::Application::DISALLOWED_COUNTRIES.join(",") %>'.split(',') }">
```
polypixeldev, line 17:
```suggestion
  <div class="card mb-6" x-data="{ minor: <%= @application.user.is_minor? || false %>, country: '<%= @application.address_country.presence %>', disallowed_countries: <%= Event::Application::DISALLOWED_COUNTRIES %> }">
```
Tested this locally - Rails will serialize it to a JSON array, so we shouldn't need this join/split

polypixeldev: I understand now - leaving Rails to serialize it will cause it to use double quotes, which while the object is still parsed by my browser, isn't valid HTML

app/models/event/application.rb
YodaLightsabr: Maybe we could add a validation only on submission? Ideally we don't want to error on autosave.

app/views/event/applications/personal_info.html.erb
YodaLightsabr: Oh that's great, I didn't know `x-init` existed

garyhtou (approved):

PR #13071, opened 2026-02-27
Fix Blazer current user

polypixeldev:
## Summary of the problem

Blazer's user functionality has been missing since #12627, since `current_user` is `nil` without `ApplicationController` setting it.

## Describe your changes

Adds new `find_current_session` and `blazer_current_user` methods, and set Blazer to use `blazer_current_user`.


manuthecoder (approved):
lgtm!

garyhtou (approved):

PR #13076, opened 2026-02-27
[Popovers] Improve performance

manuthecoder:
This PR cleans up how popover modals work across the app. Instead of each view managing its own modal markup and logic, everything now flows through a single shared popover component.

**What changed**

- Added a reusable `_shared_popover.html.erb` component and updated `ui.js` to handle dynamic content loading and cleanup — replacing one-off modal sections scattered across transaction, card grant, and stripe card views
- Introduced a `PopoverHelper` module that standardizes how popover trigger data attributes are built for different entity types (HCB codes, card grants, stripe cards, employees)
- Updated `canonical_transactions`, `canonical_pending_transactions`, `event_cards_table`, and `ledger_audits/tasks/index` to use the new helper, and removed all the old duplicated modal markup
- Popover behavior remains gated behind the `:hcb_code_popovers_2023_06_16` feature flag for gradual rollout


garyhtou (approved):
I'm very happy with this. I makes for a better DX too because when you don't need to worry about adding code for each individual popover modal

app/helpers/popover_helper.rb
```diff
@@ -0,0 +1,68 @@
...
+      modal: "shared_popover",
+      popover_title: title,
+      popover_src: src,
+      popover_frame_id: frame_id,
+      popover_state_url: state_url,
+      popover_external_link: external_link,
+      popover_state_title: state_title,
+      popover_size: size
+    }.compact
+  end
```
YodaLightsabr, line 25: If we never call this directly from the view and don't see a need to in the future, I would make this method private.

YodaLightsabr (approved):

PR #13083, opened 2026-02-28
[v3 API] expose commenter on activity

3kh0:
## Summary of the problem

`comment.create` activity events in the API do not disclose the commenter, leaving it to be nil. The activity entity only checked `activity.owner` for a user, but for comments, the creator is stored on the trackable (comment) itself via its user association.

## Describe your changes

Added a fallback so that if `activity.owner` isn't a User, it checks `activity.trackable.user` instead. This makes it so `comment.create` events (and similar) correctly expose the associated user in the API.


garyhtou (commented):
Ah interesting. Does this mean the proper fix is also to set `Activity#owner` to be whoever created the comment?

If so, I'd much rather us fix the data

manuthecoder (approved):
lgtm!

db/migrate/20260301052816_backfill_comment_activity_owners.rb
garyhtou, line 1: If we're backfilling, I think we need to update the `tracked` statement in the model to ensure new data is saved correctly.

app/api/api/entities/activity.rb
```diff
@@ -14,7 +14,7 @@ class Activity < Base
       end
 
       expose_associated User do |activity, options|
-        activity.owner.is_a?(User) ? activity.owner : activity.user
```
garyhtou, line 17: I have a feeling we probably still want to keep this existing code since other activities may depend on it.

garyhtou (changes requested):

PR #13085, opened 2026-02-28
[Admin] Sort reimbursement reports by admin review requested instead of creation date

manuthecoder:
image


garyhtou (approved):
okay, couple things!
- I know Lucy said sort by `submitted_at`, i think her intention is to have Ops process older reports first since they've been waiting the longest. So, I changed the sort to be `reimbursement_requested_at ASC`. I'm using `reimbursement_requested_at` since that's been it falls into Ops' hands and the timer starts ticking.
- For reports that don't have a `reimbursement_requested_at`, continue to sort by `created_at DESC`.

garyhtou:
And on admin page, show how long a report has been waiting for review

PR #13092, opened 2026-03-01
Fix showing invites without contract parties

polypixeldev:
## Summary of the problem

I added this extra check, but forgot to put a safe navigation operator. This is necessary because while we backfilled most contract parties, there were some that we did not have information for, so they do not have an associated signee party.

## Describe your changes

Add safe nav operator


manuthecoder (approved):
beautiful.

garyhtou (approved):

PR #13095, opened 2026-03-01
[Applications] Update the correct application user when editing as admin

polypixeldev:
## Summary of the problem

We were using `current_user` instead of `@application.user` when updating profile fields on the personal info or edit pages, causing the admin's profile information to get overwritten when they made a change.

## Describe your changes

Use `@application.user` instead of `current_user` when updating or reading profile data in application forms.

### DIFF
--- a/app/controllers/event/applications_controller.rb
+++ b/app/controllers/event/applications_controller.rb
@@ -176,9 +176,9 @@ class Event
       @application.save!
 
       if user_params.present?
-        success = current_user.update(user_params)
+        success = @application.user.update(user_params)
         if params[:autosave] != "true" && !success
-          render turbo_stream: turbo_stream.replace(:user_errors, partial: "event/applications/error", locals: { user: current_user })
+          render turbo_stream: turbo_stream.replace(:user_errors, partial: "event/applications/error", locals: { user: @application.user })
           return
         end
       end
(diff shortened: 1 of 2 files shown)


manuthecoder (approved):
LGTM!

app/views/event/applications/personal_info.html.erb
```diff
@@ -20,11 +20,11 @@
         <%= form.label :full_name do %>
           Full name <span class="text-primary">*</span>
         <% end %>
-        <%= form.text_field :full_name, value: current_user.full_name.presence, required: true, placeholder: "Fiona Hackworth III", style: "max-width:100%", pattern: '^\S+.+' %>
+        <%= form.text_field :full_name, value: @application.user.full_name.presence, required: true, placeholder: "Fiona Hackworth III", style: "max-width:100%", pattern: '^\S+.+' %>
```
garyhtou, line 23: LOLL that's funny

app/controllers/event/applications_controller.rb
```diff
@@ -176,9 +176,9 @@ def update
       @application.save!
 
       if user_params.present?
-        success = current_user.update(user_params)
+        success = @application.user.update(user_params)
```
garyhtou, line 179: We should investigate whether this bypasses any important checks/logic located in `UserController#update`

garyhtou (approved):

garyhtou:
Thank you!

PR #13102, opened 2026-03-01
[Applications] Add country rejection message

polypixeldev:
## Summary of the problem

While we restrict what countries users can submit in the form, they may still change the country to something else.

## Describe your changes

Add a country rejection message.

### DIFF
--- a/app/models/event/application.rb
+++ b/app/models/event/application.rb
@@ -199,7 +199,7 @@ class Event
       country = <<~MSG.strip
         Hi #{user.first_name},
 
-        Thank you for expressing interest in using HCB for your project, #{name}. Unfortunately, a lot of our banking infrastructure does not work in countries outside of the U.S. and Canada. We really want to support projects from all around the world, but due to different banking systems, we are not able to partner with organizations from certain countries, including #{ISO3166::Country.new(address_country).common_name}.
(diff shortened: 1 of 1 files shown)


Mohamad-Mortada (approved):

app/models/event/application.rb
```diff
@@ -196,10 +196,22 @@ def rejection_messages
         The HCB Team
       MSG
 
+      country = <<~MSG.strip
+        Hi #{user.first_name},
+
+        Thank you for expressing interest in using HCB for your project, #{name}. Unfortunately, a lot of our banking infrastructure does not work in countries outside of the U.S. and Canada. We really want to support projects from all around the world, but due to different banking systems, we are not able to partner with organizations from certain countries, including #{ISO3166::Country.new(address_country).common_name}.
```
garyhtou: Imma leave this more generic since sometimes we reject applications that apply with `country = 'US'` since they clearly operate in an unsupport country. Including `ISO3166::Country.new(address_country).common_name` is nice, but would be confusing if we said we don't operate in the US.

If you'd like, you can conditionally render it if country != US nor canada.

```suggestion
        Thank you for expressing interest in using HCB for your project, #{name}. We really want to support projects from all around the world. However, due to regulatory restrictions and incompatible financial systems, we are unable to partner with organizations that operate in certain countries.
```

garyhtou (approved):
Good to merge as-is with the suggestion, or feel free to add the conditional and merge without further review.

PR #13106, opened 2026-03-01
[Development] Fix Blazer not connecting to DB

Luke-Oldenburg:
## Summary of the problem

We needed to explicitly specify the DB name.


manuthecoder (approved):

garyhtou (approved):
People should add
```env
BLAZER_DATABASE_URL=postgres://postgres:postgres@db:5432/bank_development
```
to their `.env`/`.env.development` file in order to get blazer working.

If you're running Rails outside of Docker, then try
```env
BLAZER_DATABASE_URL=postgres://postgres:postgres@127.0.0.1:5432/bank_development
```

PR #13109, opened 2026-03-01
[User] Store first and last name separately in DB

manuthecoder:
Closes #12727


Luke-Oldenburg:
> Dropped NOT NULL on first_name for smooth rollout

Shouldn't we just delete the migration?

PR #13110, opened 2026-03-01
[BusinessTime] Configure gem to support holidays

### LINKED ISSUE #13041: [BusinessTime] Configure gem to support Holidays
We'll need to define our own holidays

Luke-Oldenburg:
## Summary of the problem

closes #13041

### DIFF
--- a/config/business_time.yml
+++ b/config/business_time.yml
@@ -37,7 +37,7 @@ business_time:
     - November 10, 2028
     - November 23, 2028
     - December 25, 2028
-    - January 1, 2029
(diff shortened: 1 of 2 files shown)


manuthecoder (approved):
lgtm! although i do wish there was a better way than listing all them out until 2030

manu-gurudath:
also i noticed some dates use leading zeroes while some don't, you might want to fix it for consistency

garyhtou (approved):
nice!

Image
(time is in UTC)

PR #13113, opened 2026-03-02
Add unauthorized callout for transfer action based on event policy

manu-gurudath:
image

Disabling the fields alone is confusing for users, this was seen in #hq-hcb.


app/views/layouts/transfer.html.erb
```diff
@@ -30,6 +30,11 @@
     </div>
   </div>
   <div class="flex-1 md:pt-3 sm:mb-24">
+    <% if @event && !policy(@event).create_transfer? %>
```
garyhtou: hide this for disbursements

app/views/layouts/transfer.html.erb
```diff
@@ -35,6 +35,11 @@
     </div>
   </div>
   <div class="flex-1 md:pt-3 sm:mb-24">
+    <% if @event && !policy(@event).create_transfer? && controller_name != "disbursements" %>
```
Luke-Oldenburg: Check if disbursement is defined?

manu-gurudath:
Closing this PR in favor of the new new payments system, which already has this callout

PR #13114, opened 2026-03-02
[User] Fix erroneous "Welcome to HCB" email sends

garyhtou:
## Summary of the problem

Users are getting a "Welcome to HCB" email when updating their user (e.g. uploading a new profile picture).

Problem: `full_name_before_last_save` is `nil` if `full_name` wasn't changed

## Describe your changes

Check to see that full name was changed AND it was blank before save.

I haven't tested this—I would appreciate if someone tests it lol

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -373,7 +373,7 @@ class User < ApplicationRecord
   end
 
   def was_onboarding?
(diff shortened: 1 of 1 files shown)


garyhtou:
Okay, i did a quick test and it seems to work

PR #13115, opened 2026-03-02
Add HtmlCompressor middleware for HTML minification in production

manuthecoder:
HTML minification strips out unnecessary whitespace, comments, and redundant attributes from ERB templates’ output, shrinking file sizes by 20-50% and speeding up page load!


sampoder:
this is sick, nice idea

garyhtou (approved):

PR #13116, opened 2026-03-02
[Event::Applications] Use `parent_layout`

manuthecoder:
There's a lot of repeated code, and it and can be fixed with the `parent_layout` helper


garyhtou (approved):
We can likely remove the `fullstory` partial at the end since it might already be rendered by the layout

PR #13117, opened 2026-03-02
[Reimbursement::Report] Add processing time

Luke-Oldenburg:
I think this could still be more pronounced on the page so I'm open to suggestions. I also think this file could use to be refactored as it has gotten pretty hard to read.


manuthecoder (approved):
lgtm!

Mohamad-Mortada (approved):

PR #13119, opened 2026-03-02
[Reimbursements] Fix visual bugs

manuthecoder:
### Problem
Various users reported visual bugs in reimbursements through Slack. 
image

### Preview
image
image
image
image


YodaLightsabr (commented):
Could you add some more screenshots of different states, and one from Wise reimbursements? I think we already have a bar at the bottom in some cases. I'm worried this redesign might make things too busy.

YodaLightsabr:
It looks like there's a lot of padding in the new design. Are we able to keep the existing layout / padding but still have the stat take up two lines (number below the header) on narrow screens?

garyhtou (approved):
I'm cool with how this looks! Happy to continue iterating if you've got thoughts, Ian.

Image

Image

garyhtou:
(Screenshots in PR description are tad outdated due to fix in https://github.com/hackclub/hcb/pull/13119/commits/74937d569b4a330b80d9eb3f3f8ff4b948c06648)

PR #13121, opened 2026-03-02
[v3 API] Publicly log bank fees

3kh0:
## Summary of the problem

The current activity feed is great, however bank fees are missing. That means if you are trying to build [Sankey charts](https://en.wikipedia.org/wiki/Sankey_diagram) based on the public feed, you are likely to miss data!

## Describe your changes

Updated the `bank_fee` modal with PublicActivity and includes create_activity callbacks for `in_transit` and `settled`.

The /api/v3/activities endpoint returns the modal with the correct organization and transaction data.

Do note that none of these have an `event_id` tied to them; this is on purpose. Having these fill up the sidebar and Discord feeds does not feel right, but it's fine for a low-level API.


sampoder:
The expense payouts should not be in here - the logs are handled by the other reimbursement-related models

app/models/bank_fee.rb
```diff
@@ -31,6 +31,9 @@ class BankFee < ApplicationRecord
   include PublicIdentifiable
   set_public_id_prefix :bfe
 
+  include PublicActivity::Model
+  tracked owner: proc{ |controller, record| controller&.current_user }, event_id: proc { |controller, record| record.event.id }, only: [:create]
```
manu-gurudath: this stores `event_id` in `activity_custom_fields_global` (via `assign_custom_fields` in the gem's `Tracked` module). this value is used as a default for all `create_activity` calls on the model, including the manual ones in the aasm callbacks (`bank_fee.in_transit`, `bank_fee.settled`), via `prepare_custom_fields`

all three activities will have `event_id` set and will show up in the sidebar and Discord feed. the stated design intent won't be achieved with this implementation.

if the intent is to only expose these via the api (with no event_id), the `tracked` line needs to either:

-   not include `event_id`
-   use `only: []` and pass `event_id` explicitly only where desired

manuthecoder (changes requested):
the description still references expense payout changes ("Updated the bank_fee modal and expense_payout with PublicActivity..."), but the diff only contains bank_fee.rb. see sam's comment, the expense payout changes were presumably dropped, but the description wasn't updated

app/models/bank_fee.rb
3kh0: Resolved in a7abd151f0ebb34d6024aaa662cf05b56b03a3ca

app/models/bank_fee.rb
```diff
@@ -47,10 +50,16 @@ class BankFee < ApplicationRecord
     state :settled
 
     event :mark_in_transit do
+      after do
+        create_activity(key: "bank_fee.in_transit", owner: nil, recipient: event)
+      end
```
garyhtou: Is it necessary to expose this? This is more related to the internal workings of HCB

garyhtou (commented):
We should probably add views to handle the rendering of these activities.

See this folder: `app/views/public_activity/`

app/models/bank_fee.rb
3kh0: 8f96efb83a52fcede656101ddc79240932739261

PR #13127, opened 2026-03-03
Use `minify_html` instead of `htmlcompressor`

manuthecoder:
## Summary of the problem
`htmlcompressor` has bugs. This PR uses `minify_html` which is more modern and faster


garyhtou (approved):
Ideally, we handle this at the Rack level, however this is good enough for now

PR #13131, opened 2026-03-03
[Affiliations] Hide bad data and add metadata validation

polypixeldev:
## Summary of the problem

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2761
Our frontend validations seem to have been bypassed, and affiliations with metadata `{}` were created.

## Describe your changes

- Adds a `nonempty` scope to `Event::Affiliation` to hide any existing affiliations with empty metadata
- Adds a `metadata_contains_required_fields` validation to ensure that all affiliation types have backend validations that match their frontend validations

### DIFF
--- a/app/models/event/affiliation.rb
+++ b/app/models/event/affiliation.rb
@@ -27,6 +27,7 @@ class Event
     store_accessor :metadata, :league, :team_number, :size, :venue_name
 
     scope :robotics, -> { where(name: %w[first vex]) }
+    scope :nonempty, -> { where.not(metadata: {}) }
 
     def display_name
       case name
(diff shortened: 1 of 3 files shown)


garyhtou (approved):
Do you think we should remove the `nonempty` scope after deleting the bad data? So does it have a purpose even if all records are valid

polypixeldev:
Yeah, we shouldn't need that scope anymore once we delete bad data. When we do go to delete that data, let's make sure to just check is_valid in case there's affiliations that aren't empty but don't have some required fields

PR #13134, opened 2026-03-04
[Reimbursements] Fix sub-organization peer-review bug

YodaLightsabr:
## Summary of the problem

Managers of parent organizations that aren't managers on child organizations should still have manager permissions, even if they're not directly on the organization.

## Describe your changes

Remove redundant line of code that causes a permissions bug.


garyhtou (approved):
the `event.users.size` might also not work as intended here.

We should probably include users from parent orgs in that check.

cool with merging as is

PR #13135, opened 2026-03-04
[Admin] Fix disbursements performance

manuthecoder:
this PR improves the disbursements admin interface by moving event dropdown data into a preloaded `@events_for_select` variable in the controller

* it selects only necessary fields and applying a custom order — instead of querying the database directly from the view. 
* eagerly loads both the `event` and `source_event` associations on disbursement queries to eliminate N+1 issues
* tightens the `pending_expired?` model method to check for expired canonical pending transactions directly


manu-gurudath:
this is not a valid fix to this issue

PR #13144, opened 2026-03-05
Fix parent events retrieval to include all ancestors

manuthecoder:
Closes #13140

image

the issue was that [0...-1] is unreliable because Event.where doesn't guarantee a specific order. however .reject makes sure that the current event is hidden (a previous bug with the subevent picker)


garyhtou (commented):
`ancestors` appears to include self. I wonder if that's why we had the `[0...-1]`

PR #13146, opened 2026-03-05
[v3 API] Expose additional details on card charges

3kh0:
## Summary of the problem

The HCB v3 API exposes very little data for card charges, most of which can be see in the actual dashboard. Some fun things of mention are the fact you can not determine if a transaction was declined or not, you can't see the real merchant (memos do not always mean actual merchant), and merchant_ids are nowhere to be seen.

## Describe your changes

Added new fields to the card charge object in the API:

- `status`: `pending`, `closed`, `reversed`, or `declined`
- `declined`: yay or nay
- `decline_reason`: `merchant_not_allowed` or other reasons (null when not declined)
- `merchant`: object with `name`, `category`, `country`, and`network_id` (some really cool things can be done here)
- `authorization_method`: `online`, `chip`, `contactless`

No existing fields were changed or removed.


app/api/api/entities/card_charge.rb
```diff
@@ -11,6 +11,51 @@ class CardCharge < LinkedObjectBase
...
+          expose :merchant_network_id, as: :network_id, documentation: { type: "string" } do |hcb_code|
+            hcb_code.stripe_merchant&.dig("network_id")
+          end
+        end
+
+        expose :authorization_method, documentation: {
+          type: "string",
+          values: %w[keyed_in swipe chip contactless online]
+        } do |hcb_code|
+          hcb_code.pt&.raw_pending_stripe_transaction&.stripe_transaction&.dig("authorization_method")
```
manu-gurudath: potential n+1, use eager loading instead

manuthecoder (changes requested):

app/api/api/entities/card_charge.rb
3kh0: 825cef21d1dd678a1c92a2ed06ab332556a361c6

manuthecoder (approved):
LGTM!

app/api/api/entities/card_charge.rb
```diff
@@ -11,6 +11,52 @@ class CardCharge < LinkedObjectBase
...
+            "closed"
+          else
+            "pending"
+          end
+        end
+
+        expose :declined, documentation: { type: "boolean" } do |hcb_code|
+          hcb_code.pt&.declined? || false
+        end
+
```
garyhtou: I think just having the `status` key is sufficient. No need to duplicate the data here.

app/api/api/entities/card_charge.rb
```diff
@@ -11,6 +11,52 @@ class CardCharge < LinkedObjectBase
...
+            hcb_code.stripe_merchant&.dig("network_id")
+          end
+        end
+
+        expose :authorization_method, documentation: {
+          type: "string",
+          values: %w[keyed_in swipe chip contactless online]
+        } do |hcb_code|
+          hcb_code.pt&.raw_pending_stripe_transaction&.stripe_transaction&.dig("authorization_method") ||
+            hcb_code.raw_stripe_transaction&.stripe_transaction&.dig("authorization_method")
```
garyhtou: There is a `stripe_merchant` method on HcbCode! That might come in handy here

app/api/api/v3.rb
```diff
@@ -59,11 +59,16 @@ def card_charges
             settled = TransactionGroupingEngine::Transaction::All.new(event_id: org.id).run
 
             combined = pending + settled
-            combined.select! { |t| t.local_hcb_code.type == :card_charge }
+            prefix = "HCB-#{::TransactionGroupingEngine::Calculate::HcbCode::STRIPE_CARD_CODE}-"
```
garyhtou, line 79: Can you tell me more about the changes to using `start_with?`. Was `local_hcb_code.type` not sufficient?

app/api/api/entities/card_charge.rb
```diff
@@ -11,6 +11,52 @@ class CardCharge < LinkedObjectBase
...
+        expose :status, documentation: {
+          type: "string",
+          values: %w[pending closed reversed declined]
+        } do |hcb_code|
+          if hcb_code.pt&.declined?
+            "declined"
+          elsif hcb_code.stripe_reversed_by_merchant?
+            "reversed"
+          elsif hcb_code.canonical_transactions.any?
+            "closed"
```
garyhtou: I believe we generally say "approved"

3kh0: Gotcha, normally in the backend you close out transactions, (hence the verbage) but I can make that change to clear up confusion!

app/api/api/entities/card_charge.rb
3kh0: c9620c04cf6a1ec884917edcd661e2a09733a4aa

app/api/api/entities/card_charge.rb
3kh0: 1b21f3d4743e062c56dcc22d9ccba3b0047b04b9

app/api/api/v3.rb
3kh0: Both produce the same result, but I used start_with? as it is slightly more performant, it has zero additional queries (if I were to use `local_hcb_code.type == :card_charge` that could spring up a n+1 unless we preload it)

app/api/api/entities/card_charge.rb
```diff
@@ -11,6 +11,47 @@ class CardCharge < LinkedObjectBase
...
+          elsif hcb_code.canonical_transactions.any?
+            "approved"
+          else
+            "pending"
+          end
+        end
+
+        expose :decline_reason, documentation: { type: "string" } do |hcb_code|
+          hcb_code.pt&.decline_reason&.to_s
+        end
```
garyhtou: Can we make this key only appear when it is declined?

app/api/api/v3.rb
garyhtou: I believe we'd also need to include `STRIPE_FORCE_CAPTURE_CODE`, otherwise we'd be missing some transactions.

I think this might also be a bug in `HcbCode#type`: https://github.com/hackclub/hcb/issues/13188

app/api/api/entities/card_charge.rb
3kh0: d5d6d70c17cc4d9e78d226d9329a7593a120d398

garyhtou (commented):
Thanks for this PR! We're gonna release the new transaction engine first and then merge this PR because the new transaction engine will require a partial rewrite of this card charge entity and potentially cause some breaking changes. To avoid breaking changes in the API, we'll release this after the new transaction engine is fully

PR #13148, opened 2026-03-06
[Nav] Make mobile use desktop's nav in a sidebar

manuthecoder:
This PR fixes at least a dozen bugs with the current mobile navigation system (for example: you currently can’t access another user’s settings except for their admin settings page if you’re on mobile, the weird screen resizing bugs, etc)

here's a video of these changes

https://github.com/user-attachments/assets/ed9ce1ac-44cc-45a5-af65-23dec63dbc4a

this pr creates a seamless experience between mobile and desktop since it uses the exact same navigation, a point which many people complained about 

it also improves performance (before, we rendered both desktop and mobile navigation regardless of the platform u were on, later hidden through CSS display: none)

also the new navigation lets you see recent activity and open the command palette on mobile as well!


app/javascript/controllers/sidebar_controller.js
```diff
@@ -0,0 +1,61 @@
...
+    document.addEventListener('keydown', this.handleKeydown)
+    window.addEventListener('resize', this.handleResize)
+    this.handleResize()
+    this.sidebarTarget
+      .querySelectorAll(
+        '.dock__item:not(summary.dock__item,.user-menu-trigger)'
+      )
+      .forEach(item => {
+        item.addEventListener('click', this.close.bind(this))
+      })
```
garyhtou, line 18: Another potential solution is to listen to `window.location`. If `window.location` changes, and the sidebar is open, then close it.

There's a channce this is less buggy

garyhtou (commented):
In another PR, let's hide the command bar trigger bar on mobile, and replace it with a simple magnifying glass icon on the top right of the side bar that triggers the command bar.

Image

garyhtou (approved):

PR #13149, opened 2026-03-06
[Payout Method] Allow user to set name for all methods

### LINKED ISSUE #13142: [Payout Method] Allow user to set name for all methods
For example, allow them to send a name for the check. Sometimes the name on the bank account is different than your name on HCB

manuthecoder:
Resolves #13142 by adding the ability to set the recipient name for payouts


app/models/user/payout_method/wise_transfer.rb
```diff
@@ -14,6 +14,7 @@
 #  currency                         :string
 #  recipient_country                :integer
 #  recipient_information_ciphertext :text
+#  recipient_name                   :string
```
garyhtou, line 17: We already store this inside `recipient_information` as `account_holder`. Let's maybe migrate `account_holder` into `recipient_name` 🤷‍♂️

PR #13152, opened 2026-03-07
Fix reimbursement report page performance (flickering, listener leaks, redundant queries)

anishanne:
This pr was written 100% by claude code and is for the purposes of Ian's entertainment. Don't merge it!
                                                           
  ## Summary

  The reimbursement report show page (`app/views/reimbursement/reports/show.html.erb`) is laggy and glitchy when creating, editing, and deleting expenses. Users experience flickering, slow responses, and sluggish interactions.

  Root causes: (1) Turbo broadcast triggers a full-page morph on top of already-applied turbo_stream updates, (2) JavaScript event listener leaks and heavy DOM manipulation, (3) redundant SQL queries in the _actions partial re-rendered on
  every CRUD operation, (4) per-expense duplicate queries for move modals.

  ### Phase 1: Remove redundant Turbo broadcast from Expense model

  **`app/models/reimbursement/expense.rb`**

  Removed `broadcasts_refreshes_to` and its `include TouchHistory` guard. Every expense save was broadcasting a refresh to the report channel. The show page subscribes via `turbo_stream_from @report` + `turbo_refreshes_with method: :morph`,
  causing a full page re-fetch + morph AFTER the controller's targeted turbo_stream updates had already applied. This double-update reset Stimulus controller state (edit mode, tooltips, focus) and caused flickering.

  The controller already returns comprehensive turbo_stream responses for all CRUD ops. The Report model's own broadcast (guarded by `was_touched?`) still provides real-time updates for direct report changes.

  ### Phase 2: Fix JavaScript event listener leaks and DOM manipulation

  **`app/javascript/controllers/expense_form_controller.js`**

  - **2a: Fix duplicate keydown listeners.** Lines 30-35 registered `document.addEventListener('keydown', ...)` inside the `for (const field of this.fieldTargets)` loop. N fields = N duplicate global listeners, never removed. Fixed by moving
   to a single bound listener registered once in `connect()`, removed in `disconnect()`.

  - **2b: Fix lightbox click listener accumulation.** `#lightbox()` added a new click listener every time `edit()` was called. Fixed by creating the bound handler once in `connect()` and adding/removing it in `#lightbox()` based on state.

  - **2c: Replace DOM-restructuring tooltips with class toggling.** `#addTooltip()` wrapped fields in new DOM nodes via `insertBefore`/`appendChild`; `#removeTooltip()` unwrapped them. This restructured the DOM every toggle and called
  `window.attachTooltipListener()` globally. Fixed by toggling `tooltipped` and `tooltipped--n` classes + `aria-label` on existing parent `<div>`s marked with `data-tooltip-wrapper` in `_form.html.erb`.

  - **2d: Added `disconnect()` method** to clean up keydown and lightbox listeners.

  **`app/views/reimbursement/expenses/_form.html.erb`**

  Added `data-tooltip-wrapper` attributes to the memo, value, category, and description parent `<div>`s so the tooltip class toggling can find them.

  ### Phase 3: Reduce SQL queries in _actions.html.erb

  - **3a: Eager-load expenses in the controller.** `set_report_user_and_event` now uses `Reimbursement::Report.includes(expenses: [:approved_by, :receipts]).find(...)`.

  - **3b: Pre-compute collections at top of `_actions.html.erb`.** Added `loaded_expenses`, `expenses_to_sum`, `approved_to_sum`, `complete_to_sum` computed once from the loaded association. Replaced all scoped queries throughout the partial
   (`report.expenses.count > 0`, `report.expenses.none?`, `report.expenses.any?`, `report.expenses.approved.none?`, `report.expenses.approved.to_sum.count`, `report.expenses.to_sum.count`, `report.expenses.to_sum.complete.count`) with
  in-memory operations on these collections.

  - **3c: Reload association in `actions_turbo_stream`.** Added `@expense.report.expenses.reload` so the partial gets fresh data after create/update/delete.

  ### Phase 4: Deduplicate per-expense move modal queries

  **`app/views/reimbursement/reports/_expenses.html.erb`**

  Each expense rendered a move modal querying `current_user.reimbursement_reports` and `current_user.events` — identical queries running N times. Pre-computed `move_report_options` and `move_event_options` once before the loop and passed as
  locals.

  **`app/views/reimbursement/expenses/_expense.html.erb`**

  Added `move_report_options: nil, move_event_options: nil` to strict locals. When nil (turbo_stream rendering of a single expense), falls back to computing inline. When pre-computed values are passed from the loop, uses those directly.

  ### Phase 5: Fix inline expense query in show.html.erb

  **`app/views/reimbursement/reports/show.html.erb`**, line 221

  Replaced `@report.expenses.select { |e| e.type == Reimbursement::Expense::Fee.name }.empty?` with `@report.expenses.none?(&:is_fee?)`. The `is_fee?` method already exists on the Expense model, and with eager-loaded expenses from Phase 3a
  this operates in memory.

### DIFF
--- a/app/controllers/reimbursement/expenses_controller.rb
+++ b/app/controllers/reimbursement/expenses_controller.rb
@@ -120,6 +120,7 @@ module Reimbursement
     end
 
     def actions_turbo_stream
+      @expense.report.expenses.reload
       turbo_stream.replace("action-wrapper", partial: "reimbursement/reports/actions", locals: { report: @expense.report, user: @expense.report.user })
     end
 
--- a/app/controllers/reimbursement/reports_controller.rb
+++ b/app/controllers/reimbursement/reports_controller.rb
@@ -430,7 +430,7 @@ module Reimbursement
     private
 
     def set_report_user_and_event
-      @report = Reimbursement::Report.find(params[:report_id] || params[:id])
+      @report = Reimbursement::Report.includes(expenses: [:approved_by, :receipts]).find(params[:report_id] || params[:id])
       @event = @report.event
       @user = @report.user
     rescue ActiveRecord::RecordNotFound
--- a/app/javascript/controllers/expense_form_controller.js
+++ b/app/javascript/controllers/expense_form_controller.js
@@ -18,6 +18,19 @@ export default class extends Controller {
   }
 
   connect() {
+    this._handleKeydown = (e) => {
+      if (e.key === 'Escape' && this.enabledValue) {
+        this.formTarget.reset()
+        this.close()
+      }
+    }
+    document.addEventListener('keydown', this._handleKeydown)
+
+    this._handleLightboxClick = (e) => {
+      e.preventDefault()
+      this.formTarget.requestSubmit()
+    }
+
     for (const field of this.fieldTargets) {
       if (field.nodeName == 'SELECT') {
         field.disabled = !this.enabledValue
@@ -26,13 +39,6 @@ export default class extends Controller {
         field.addEventListener('dblclick', () => this.edit())
         this.#addTooltip(field, 'Double-click to edit...')
       }
-
-      document.addEventListener('keydown', e => {
-        if (e.key === 'Escape' && this.enabledValue) {
-          this.formTarget.reset()
-          this.close()
-        }
-      })
     }
 
     // we don't render the button if the report is reimbursed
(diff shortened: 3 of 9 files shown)


sampoder:
Can you include before and after screenshots for the four issues described? There's context behind many of the Turbo streams choices so I'd like actually see what the issues are vs. a text-based description

sampoder:
Also let's break this into 4 PRs. Too many interconnected things make this hard to review as one.

PR #13157, opened 2026-03-09
[Contracts] Properly handle contract deletion

polypixeldev:
## Summary of the problem

We've been getting some errors such as https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2769 and https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2586 due to contracts getting deleted.

## Describe your changes

The only case in which contracts should be (soft) deleted is if the associated contractable is deleted. This PR adds cascade (soft) deletion to contract parties and handles a deleted party in the reminder job.

I think we should go ahead and delete existing contracts that link to a nonexistent OPI (https://hcb.hackclub.com/blazer/queries/1129-contracts-with-deleted-opis) - we make the assumption in code that all contracts have a contractable and should avoid related errors by deleting these contracts.

### DIFF
--- a/app/models/concerns/contractable.rb
+++ b/app/models/concerns/contractable.rb
@@ -4,7 +4,11 @@ module Contractable
   extend ActiveSupport::Concern
 
   included do
-    has_many :contracts, as: :contractable, dependent: :destroy
+    has_many :contracts, as: :contractable
+
+    after_destroy_commit do
+      contracts.where(aasm_state: [:pending, :sent]).each(&:mark_voided!)
+    end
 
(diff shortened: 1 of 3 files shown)


app/models/concerns/contractable.rb
```diff
@@ -6,6 +6,10 @@ module Contractable
   included do
     has_many :contracts, as: :contractable, dependent: :destroy
 
+    before_destroy do
+      contracts.where(aasm_state: [:pending, :sent]).find_each(&:mark_voided!)
+    end
+
```
garyhtou: Let's move to a `before_destroy` in Contract

garyhtou (approved):

PR #13181, opened 2026-03-11
[Applications] Add unarchiving

polypixeldev:
## Summary of the problem

People might accidentally archive their application, so they (and admins) should be able to unarchive that application.

## Describe your changes

Add action to unarchive applications and option to include archived applications on the applications admin page.


app/models/event/application.rb
```diff
@@ -367,6 +367,12 @@ def archive!
       update!(archived_at: Time.current)
     end
 
+    def unarchive!
+      send_contract if contract.nil? && ((teen_led && (!draft? && !rejected?)) || (!teen_led && approved?))
```
YodaLightsabr: It would be great if this conditional logic could be abstracted at some point. I've seen it in a lot of places and I think it could be reused. In a different PR, though.

YodaLightsabr (approved):

YodaLightsabr:
LGTM. Please fix checks and then we're good to merge.

PR #13199, opened 2026-03-13
[Event::Application] Fix flash bugs

manuthecoder:
This PR fixes the way flashes are rendered for `Event::Application`s


polypixeldev (commented):
This seems to hide flashes entirely, not just getting rid of the duplicate flash

app/views/layouts/apply.html.erb
```diff
@@ -2,6 +2,7 @@
 <% @home_size = 50 %>
 <% content_for(:container_class, "container--full !px-0") %>
 <% no_app_shell %>
+<% @hide_flash = true %>
```
polypixeldev:
```suggestion
```

app/views/layouts/apply.html.erb
```diff
@@ -32,6 +33,7 @@
 <% end %>
 
 <%= yield(:header) %>
+    <%= render "application/flash" %>
```
polypixeldev:
```suggestion
```

app/views/layouts/apply.html.erb
polypixeldev, line 122:
```suggestion
<% @hide_flash = true %>
<% parent_layout "application" %>
```

polypixeldev (changes requested):
This adds a new flash to the top of the screen, above the header. Existing code in this layout positions the flash in the center of the header on wide screens and below it on small screens. These suggestions fix showing these flashes and not the parent layout flashes while still respecting the hide_flash instance variable.

polypixeldev (approved):

garyhtou (approved):

PR #13204, opened 2026-03-14
[Disbursements] Remove reliance on hash IDs for org dropdown

YodaLightsabr:
## Summary of the problem

This page is very slow! ([Flamegraph](https://hackclub.enterprise.slack.com/files/U022XFD2TML/F0ALL6H1P9Q/frame_1_-_2026-03-13t231946.241.png)) Hash IDs take about 3 seconds per page load for admins. Numeric IDs don't have this problem, but we don't want to expose numeric IDs to non-admin users. 

## Describe your changes

* Remove reliance on public IDs for admins loading the page by using numeric ID instead.
* Use friendly ID instead of public ID as the identifier for Turbo

### DIFF
--- a/app/controllers/disbursements_controller.rb
+++ b/app/controllers/disbursements_controller.rb
@@ -71,10 +71,10 @@ class DisbursementsController < ApplicationController
   def create
     if admin_signed_in?
       @source_event = Event.find(disbursement_params[:source_event_id])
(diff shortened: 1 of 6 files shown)


garyhtou (changes requested):
The long-term fix we want here is to move searching of long select options to the backend.

PR #13206, opened 2026-03-15
[v4][Tap to Pay] add payment intent route

Mohamad-Mortada:
#12971


app/controllers/api/v4/donations_controller.rb
```diff
@@ -32,6 +32,26 @@ def create
         render "show", status: :created
       end
 
+      def payment_intent
```
garyhtou, line 36: Only trusted apps should be able to do this

garyhtou (approved):

PR #13207, opened 2026-03-15
[v4] Limit Event Balance graph data to one year

Mohamad-Mortada:
We want to mimic the organization homepage but currently it will show all data from when the org was created which is inefficient if we only want one year


YodaLightsabr (approved):
LGTM. Please add description before merging.

PR #13209, opened 2026-03-15
[Logins] Keep track of `state` and refactor messy login code

YodaLightsabr:
## Summary of the problem

`LoginsController` has gotten pretty messy over the years because of our reluctance to touch it more than we have to. It's resulted in some pretty funky stuff, like:
* Stimulus targets that aren't bound to a controller
* Email and SMS logins sharing the same view, only differentiated by a hidden `sms` parameter
* Repetition of code multiple times across views

We also rely on some janky hacks like checking if the `return_to` URL contains `"application"` to determine what layout we should render.

## Describe your changes

This PR adds the `state` column to `Login`, allowing us to remove a lot of repeated code in favor of a robust DB-backed solution. This PR also (finally) rips apart email and SMS login code auth, adds a new page for security code auth, and defaults to the user's last-used login methods. I've also refactored and renamed some things to be more clear and readable.

image


app/models/login.rb
```diff
@@ -9,6 +9,7 @@
 #  authentication_factors   :jsonb
 #  browser_token_ciphertext :text
 #  is_reauthentication      :boolean          default(FALSE), not null
+#  state                    :jsonb
```
garyhtou, line 12: Validate the size of `state`. Limit to some KB

app/models/user.rb
```diff
@@ -580,6 +580,10 @@ def discord_account
     @discord_account ||= @discord_bot.user(discord_id)
   end
 
+  def preferred_login_methods
+    logins.complete.last&.authentication_factors&.filter_map { |key, value| key if value } || []
+  end
```
garyhtou, line 590: Let's define an explicit order

app/controllers/logins_controller.rb
```diff
@@ -205,17 +232,25 @@ def complete
...
-    path = URI(params[:return_to] || "").path
+    @for_application = @login&.state&.[]("for_application") || ActiveModel::Type::Boolean.new.cast(params[:for_application])
+  end
 
-    @for_application = path.starts_with?("/applications")
-  rescue URI::InvalidURIError
-    @for_application = false
+  def set_return_to
+    @return_to = @login&.state&.[]("return_to") || url_from(params[:return_to])
   end
```
garyhtou: Let's enforce that if you're using `return_to` and `for_application`, you can only read from the state from the Login model. We should never read from `param`. Instead, `param` needs to get written into `state`. This give us a clear paper trail of what states were used, and also where the data is coming from.

app/controllers/logins_controller.rb
```diff
@@ -29,27 +28,29 @@ def new
 
   # when you submit your email
   def create
-    user = User.create_with(creation_method: @for_application ? :application_form : :login).find_or_create_by!(email: params[:email])
+    user = User.create_with(creation_method: params[:for_application] ? :application_form : :login).find_or_create_by!(email: params[:email])
 
+    state = { return_to: params[:return_to], for_application: params[:for_application] }
```
garyhtou: Let's sanitize using `url_form` before saving to the DB, and also before redirecting the user

app/controllers/logins_controller.rb
```diff
@@ -169,11 +192,11 @@ def complete
       if @referral_link.present?
         redirect_to referral_link_path(@referral_link)
       elsif (@user.full_name.blank? || @user.phone_number.blank?) && !@for_application
-        redirect_to edit_user_path(@user.slug, return_to: params[:return_to])
+        redirect_to edit_user_path(@user.slug, return_to: @return_to)
```
garyhtou: We should grab the `return_to` from the login

app/views/logins/backup_code.html.erb
```diff
@@ -9,30 +9,31 @@
...
-    <%= hidden_field_tag :device_info %>
-    <%= hidden_field_tag :os_info %>
-    <%= hidden_field_tag :timezone %>
-    <%= hidden_field_tag :return_to, @return_to if @return_to %>
+    <%= render partial: "logins/fingerprint_info" %>
+
     <div class="flex flex-row justify-between items-center mt-4 gap-2">
-      <% if @webauthn_available || @totp_available || @email_available || @sms_available %>
-        <%= link_to "Sign in another way", choose_login_preference_login_path(@login, return_to: @return_to), class: "block mt-0 no-underline" %>
+      <% if @login.available_factors.excluding(:backup_code).any? %>
```
garyhtou, line 27: Very nice!!

app/views/logins/sms.html.erb
```diff
@@ -0,0 +1,76 @@
...
+            "",
+            placeholder: "Enter your login code",
+            class: "!max-w-full w-max",
+            autocomplete: "one-time-code",
+            required: true,
+            autofocus: true,
+            inputmode: "numeric",
+            data: { "op-ignore": true }
+          ) %>
+    <% end %>
```
garyhtou, line 49: I think we should graduate/kill this flipper flag soon

Luke-Oldenburg (approved):
lgtm

manuthecoder (approved):
thank you.

app/controllers/logins_controller.rb
```diff
@@ -18,39 +16,29 @@ class LoginsController < ApplicationController
...
-    user = User.create_with(creation_method: @for_application ? :application_form : :login).find_or_create_by!(email: params[:email])
+    @user = User.create_with(creation_method: login_params[:purpose] == "application" ? :application_form : :login).find_or_create_by!(email: params[:email])
 
-    if params[:referral_link_id].present?
-      referral_link = Referral::Link.find_by(slug: params[:referral_link_id]).presence
-      login = user.logins.create(referral_link:)
-    else
-      login = user.logins.create
-    end
+    @login = @user.logins.create(login_params)
```
garyhtou:
```suggestion
    @login = @user.logins.new(login_params)
    @user = User.create_with(creation_method: @login.for_application? ? :application_form : :login).find_or_create_by!(email: params[:email])

    @login.save!
```

app/controllers/logins_controller.rb
```diff
@@ -205,17 +200,37 @@ def complete
...
+    end
+  end
 
-    @for_application = path.starts_with?("/applications")
-  rescue URI::InvalidURIError
-    @for_application = false
+  def login_params
+    params.require(:login).permit(:return_to, :purpose, :referral_link_id)
+  rescue ActionController::ParameterMissing
+    ActionController::Parameters.new
```
garyhtou: This shouldn't be possible. So it'd prefer for it to error and for us to know about it.

```suggestion
```

app/controllers/logins_controller.rb
```diff
@@ -60,31 +43,39 @@ def create
...
-      redirect_to login_code_login_path(@login), status: :temporary_redirect
-    when "totp"
-      session[:login_preference] = "totp" if remember
-      redirect_to totp_login_path(@login), status: :temporary_redirect
-    when "webauthn"
-      # This should never happen, because WebAuthn auth is handled on the frontend
-      redirect_to choose_login_preference_login_path(@login)
+  # post to request email login code
+  def email
+    resp = LoginCodeService::Request.new(email: @email, ip_address: request.remote_ip, user_agent: request.user_agent).run
```
garyhtou, line 66: For the future, we should get rid of `@email` and only depend on `@login.email`.

Related: https://github.com/hackclub/hcb-private/issues/110

app/controllers/logins_controller.rb
```diff
@@ -60,31 +43,39 @@ def create
...
+
+    if resp[:error].present?
+      flash[:error] = resp[:error]
+      return redirect_to auth_users_path
     end
+
+    render status: :unprocessable_entity
+  rescue ActionController::ParameterMissing
+    flash[:error] = "Please enter an email address."
+    redirect_to auth_users_path
```
garyhtou: I think this should no longer be possible because we're pulling email from `Login` and not params. Let's drop this

garyhtou: Please make this change to all the other factor actions as well

app/controllers/logins_controller.rb
```diff
@@ -168,12 +162,12 @@ def complete
     if @login.complete? && @login.user_session.present?
       if @referral_link.present?
         redirect_to referral_link_path(@referral_link)
-      elsif (@user.full_name.blank? || @user.phone_number.blank?) && !@for_application
-        redirect_to edit_user_path(@user.slug, return_to: params[:return_to])
+      elsif (@user.full_name.blank? || @user.phone_number.blank?) && !@login.for_application?
```
garyhtou, line 160: In the future, we should swap this out for `@user.onboarding?` We may need to update `User#onboarding?` to cover more checks.

garyhtou (approved):

PR #13210, opened 2026-03-16
[Admin] Add Combo Box for async searching of dropdowns

manuthecoder:
Currently, the admin dashboard is extremely slow due to the way filters are being rendered. This PR fixes performance by using turbo frames and also adds extra functionality to the filter menu!


app/controllers/admin_controller.rb
```diff
@@ -684,6 +704,10 @@ def paypal_transfers
     @q = params[:q].presence
     @event_id = params[:event_id].presence
 
+    if @event_id
+      @event = Event.find(@event_id)
+    end
```
```suggestion
  @event = Event.find_by(params[:event_id) if params[:event_id].present?
```

And then switch `@event_id` usage to `@event.id`

Let's do this for other as well.

garyhtou (commented):
Let's split the unrelated CSS changes into a different PR and we'll re-review

app/controllers/admin_controller.rb
```diff
@@ -386,6 +386,26 @@ def ledger
     @canonical_transactions = relation.page(@page).per(@per).order(date: :desc)
   end
 
+  def event_search
+    @q = params[:q].presence
+    @events = if @q.present?
+                Event.where("name ILIKE ? OR CAST(id AS TEXT) ILIKE ?", "%#{Event.sanitize_sql_like(@q)}%", "%#{Event.sanitize_sql_like(@q)}%").order(:name).limit(20).select(:id, :name)
```
garyhtou: Should we use the `pg_search_scope` `search_name`?

app/controllers/admin_controller.rb
```diff
@@ -386,6 +386,26 @@ def ledger
...
+              else
+                Event.none
+              end
+    render turbo_stream: helpers.async_combobox_options(@events)
+  end
+
+  def user_search
+    @q = params[:q].presence
+    @users = if @q.present?
+               User.where("full_name ILIKE ? OR email ILIKE ? OR CAST(id AS TEXT) ILIKE ?", "%#{User.sanitize_sql_like(@q)}%", "%#{User.sanitize_sql_like(@q)}%", "%#{User.sanitize_sql_like(@q)}%").order(:full_name).limit(20).select(:id, :full_name, :email)
```
garyhtou: Should we use the `pg_search_scope` `search_name`?

app/controllers/admin_controller.rb
```diff
@@ -386,6 +386,26 @@ def ledger
     @canonical_transactions = relation.page(@page).per(@per).order(date: :desc)
   end
 
+  def event_search
+    @q = params[:q].presence
+    @events = if @q.present?
+                Event.where("name ILIKE ? OR CAST(id AS TEXT) ILIKE ?", "%#{Event.sanitize_sql_like(@q)}%", "%#{Event.sanitize_sql_like(@q)}%").order(:name).limit(20).select(:id, :name)
+              else
+                Event.none
```
garyhtou: Use `Event::CUSTOM_SORT` and list the first 20

app/controllers/admin_controller.rb
```diff
@@ -386,6 +386,26 @@ def ledger
...
+              end
+    render turbo_stream: helpers.async_combobox_options(@events)
+  end
+
+  def user_search
+    @q = params[:q].presence
+    @users = if @q.present?
+               User.where("full_name ILIKE ? OR email ILIKE ? OR CAST(id AS TEXT) ILIKE ?", "%#{User.sanitize_sql_like(@q)}%", "%#{User.sanitize_sql_like(@q)}%", "%#{User.sanitize_sql_like(@q)}%").order(:full_name).limit(20).select(:id, :full_name, :email)
+             else
+               User.none
```
garyhtou: Let's return something instead of none. Choose a reasonable sort and grab the first 20

app/controllers/admin_controller.rb
```diff
@@ -714,6 +734,8 @@ def wires
     @q = params[:q].presence
     @event_id = params[:event_id].presence
```
garyhtou: Should we get rid of this instance variable now that we have `@event` defined?

app/views/layouts/admin.html.erb
```diff
@@ -63,7 +63,7 @@
 
     <%= render(partial: "admin/flash") %>
 
-    <div id="main" class="overflow-x-auto overflow-y-auto" style="padding: 2rem; margin: 0 auto">
+    <div id="main" class="min-h-[100vh] overflow-x-auto overflow-y-auto" style="padding: 2rem; margin: 0 auto">
```
garyhtou: This causes guarantee scroll on the page. Manu said it's needed otherwise the combobox menu gets cut off. Let's see if we can find the root cause and a more suitable solution.

YodaLightsabr (approved):
Code looks good! Please address the UI bugs I DMed you about, but we're in a good spot!

app/controllers/admin_controller.rb
```diff
@@ -386,6 +386,26 @@ def ledger
...
   end
 
+  def event_search
+    @q = params[:q].presence
+    @events = if @q.present?
+                Event.search_name(@q).order(Event::CUSTOM_SORT).limit(20).select(:id, :name)
+              else
+                Event.order(Event::CUSTOM_SORT).limit(20).select(:id, :name)
+              end
+    render turbo_stream: helpers.async_combobox_options(@events)
```
YodaLightsabr, line 396:
```suggestion
    @events = if @q.present?
                Event.search_name(@q)
              else
                Event
              end.order(Event::CUSTOM_SORT).limit(20).select(:id, :name)
    render turbo_stream: helpers.async_combobox_options(@events)
```

app/controllers/admin_controller.rb
```diff
@@ -386,6 +386,26 @@ def ledger
...
+  end
+
+  def user_search
+    @q = params[:q].presence
+    @users = if @q.present?
+               User.search_name(@q).limit(20).select(:id, :full_name, :email)
+             else
+               User.order(:full_name).limit(20).select(:id, :full_name, :email)
+             end
+    render turbo_stream: helpers.async_combobox_options(@users)
```
YodaLightsabr, line 406:
```suggestion
    @users = if @q.present?
               User.search_name(@q)
             else
               User.order(:full_name)
             end.limit(20).select(:id, :full_name, :email)
    render turbo_stream: helpers.async_combobox_options(@users)
```

app/models/event.rb
```diff
@@ -934,6 +934,15 @@ def valid_scoped_tags
     scoped_tags.where(parent_event_id: parent_id)
   end
 
+  # both to_combobox_display and value are used for comboboxes
+  def to_combobox_display
+    "#{name} (#{id})"
+  end
+
+  def value
```
YodaLightsabr: This is a pretty generic method name—is this required by the combobox gem? Is there an alternative method name we could use? It's worth looking into.

app/models/event.rb
```diff
@@ -934,6 +934,15 @@ def valid_scoped_tags
     scoped_tags.where(parent_event_id: parent_id)
   end
 
+  # both to_combobox_display and value are used for comboboxes
+  def to_combobox_display
+    "#{name} (#{id})"
```
YodaLightsabr, line 939: Is there a world where we want to change the combobox display based on the user's permissions? (For example, admin vs. regular user) — if so, are we prepared to handle that?

YodaLightsabr (commented):
Nice work here!

app/models/event.rb
manuthecoder: unfortunately, yes

app/models/event.rb
manuthecoder: we can do pass params to the controller in the gem

app/models/event.rb
YodaLightsabr: yes to which?

app/models/event.rb
YodaLightsabr: How would we handle that at the model level, though?

manuthecoder: actually it somehow works without the `id` function, strange i was running into issues earlier

manuthecoder: for authorization, we can either use either pundit authorization or `render_in`

app/models/event.rb
manuthecoder: actually the combobox works without this interesting

YodaLightsabr (approved):

garyhtou (approved):

garyhtou:
^ deferring approval to ian. approving to allow merge due to gemfile codeownership

PR #13226, opened 2026-03-17
[User] Simplify `accessible_events`

garyhtou:
This PR does two things:

- Simplify `accessible_events`. No need for recursion if we use `role_by_event_id` as the source of our data
- Returns an Active Record Relation instead of an Array; Allows for changing more `where` filters.

@YodaLightsabr, can you please double check this provides the same results?


YodaLightsabr (approved):
LGTM. Ty

PR #13239, opened 2026-03-20
[CardGrant] Add large banner when transactions decline

manuthecoder:
Closes #13215

### Preview 
image

### DIFF
--- a/app/views/card_grants/_transactions.html.erb
+++ b/app/views/card_grants/_transactions.html.erb
@@ -3,6 +3,12 @@
 <% if hcb_codes&.any? %>
   <h2 class="heading h2 leading-[1.5] mt2 ml0 pt1 pb1 pr2">Transactions on this card</h2>
 
+  <% if hcb_codes.any? { |hc| hc.pt&.hcb_decline_reason == :merchant_not_allowed } %>
+    <% event_name = hcb_codes.first.event.name %>
+    <%= render "application/callout", type: "warning", title: "Transaction declined by #{event_name}" do %>
+      One or more transactions were declined because the merchant isn't enabled for this card. Please reach out to the <%= event_name %> team to request access.
+    <% end %>
+  <% end %>
   <div style="overflow-x: auto;" class="table-container">
     <table>
       <tbody data-behavior="transactions">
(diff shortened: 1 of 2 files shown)


app/views/card_grants/_transactions.html.erb
```diff
@@ -3,6 +3,13 @@
 <% if hcb_codes&.any? %>
   <h2 class="heading h2 leading-[1.5] mt2 ml0 pt1 pb1 pr2">Transactions on this card</h2>
 
+  <% if hcb_codes.any? { |hc| hc.canonical_pending_transactions.any? { |pt| pt.hcb_decline_reason == :merchant_not_allowed } } %>
```
garyhtou: Should we only show this if the most recent transaction was declined?

It may be annoying for this to continue showing after the merchant was enabled, and the charge successfully went through

manu-gurudath: done!

app/views/card_grants/_transactions.html.erb
```diff
@@ -3,6 +3,13 @@
...
+    <% declined = hcb_codes.find { |hc| hc.pt&.hcb_decline_reason == :merchant_not_allowed } %>
+    <% event_name = declined&.event&.name || "this card" %>
+    <%= render "application/callout", type: "warning", title: "Transaction declined for #{event_name}" do %>
+      One or more transactions were declined because the merchant isn't enabled for this card. Please reach out to the <%= event_name %> team to request access.
```
YodaLightsabr: It would be great if we could pluralize this automatically. Either "One of your transactions" or "Multiple of your transactions". If it's one, we're able to go into more detail and say which merchants.

Let's also slightly reword the second sentence:

```suggestion
      One or more transactions were declined because the merchant isn't enabled for this card. Please reach out to the <%= event_name %> team if you need support.
```

app/views/card_grants/_transactions.html.erb
YodaLightsabr: I think it might be better to move this logic to the controller and include this is its own partial directly on the show page.

YodaLightsabr (dismissed):

garyhtou (approved):

PR #13244, opened 2026-03-21
[Admin] Add admin homepage

YodaLightsabr:
## Summary of the problem

Right now, the only way to access the admin navigation is to a) go to your homepage (and wait for your homepage to load) to access all of your admin cards, or b) type in the address for a random admin page. This is annoying.

## Describe your changes

This PR adds `/admin` (an admin index page). It has all of the admin cards laid out in a nice wide grid.

image


manuthecoder (approved):
i love this

Luke-Oldenburg (commented):
We have a lot of places where each admin page is defined now. Is there a way we can consolidate these?

- this new admin page
- admin cards on home page
- command bar
- admin navigation

Luke-Oldenburg:
See https://github.com/hackclub/hcb/issues/13918

PR #13248, opened 2026-03-22
Add shared comments for disbursements

albertchae:
## Summary of the problem

https://hackclub.slack.com/archives/C026RKHLPNJ/p1773163261567279

Previously, since we had a single HcbCode for disbursements, comments were always shared. Now they can be on either side of the disbursement, but we still want the ability to share comments selectively.

## Describe your changes

Since both sides of a disbursement still have a connection to the Disbursement, we are adding Commentable to that model.

It was unavoidable to add shared comment specific methods to HcbCode but we tried to delegate them to models that make more sense as much as possible (e.g. `shared_comment_recipient_label` lives on ingoing/outgoing disbursements)

- Switch from hardcoded redirect to `redirect_back_or_to` since it is exactly for the case we want to handle. Redirect to the original commentable as a fallback because we don't want to redirect to a Disbursement on a shared comment vs the HcbCode that the comment is originating from.
- Add `all_comments` to Commentable that aggregates comments from both the primary commentable and its shared_commentable. Currently it doesn't own ordering because that's already handled in the view, but we should probably push that to the model
- UI shows all comments on your side and shared comments, interleaved by created_at (post time). For the checkbox we show the name of the other. Added some styling to shared comments similar to admin comments. side, but for created comments we can't since they are only stored on the shared Disbursement model and it's possible the author has access to both sides of the disbursement.
- Minor refactor of `HcbCode#disbursement?` to use existing predicate methods

### DIFF
--- a/app/models/disbursement/outgoing.rb
+++ b/app/models/disbursement/outgoing.rb
@@ -4,6 +4,10 @@ class Disbursement
   class Outgoing
     include Base
 
+    def amount
+      -disbursement.amount
+    end
+
     def hcb_code
       disbursement.outgoing_hcb_code
     end
@@ -12,12 +16,20 @@ class Disbursement
       disbursement.source_event
     end
 
-    def amount
-      -disbursement.amount
+    def counterparty_event
+      disbursement.destination_event
     end
 
     def subledger
       disbursement.source_subledger
+    end
+
+    def counterparty_subledger
+      disbursement.destination_subledger
+    end
+
+    def counterparty
+      disbursement.incoming_disbursement
     end
 
     def transaction_category
--- a/app/policies/comment_policy.rb
+++ b/app/policies/comment_policy.rb
@@ -56,6 +56,9 @@ class CommentPolicy < ApplicationPolicy
         user_list += record.commentable.event&.users || [] # event&.users can be nil (event-less reports)
         user_list += record.commentable.event&.ancestor_users || []
       end
+    elsif record.commentable.is_a?(Disbursement)
+      # TODO: possibly replace this by adding #events to Disbursement?
+      user_list = record.commentable.source_event.users + record.commentable.destination_event.users
     elsif record.commentable.is_a?(Event)
       user_list = []
     else
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -17,7 +17,9 @@ class CommentsController < ApplicationController
 
     if @comment.save
       flash[:success] = "Comment created."
-      redirect_to @commentable.is_a?(Event) ? edit_event_path(@commentable, tab: :admin) : @commentable
+      # Use return_to param if provided, otherwise fall back to the commentable
+      # url_from validates the URL is internal to prevent open redirect vulnerabilities
+      redirect_back_or_to url_from(params[:comment][:return_to]) || @commentable
     else
       render :new, status: :unprocessable_entity
     end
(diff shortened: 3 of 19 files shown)


app/policies/comment_policy.rb
```diff
@@ -56,6 +56,9 @@ def users
         user_list += record.commentable.event&.users || [] # event&.users can be nil (event-less reports)
         user_list += record.commentable.event&.ancestor_users || []
       end
+    elsif record.commentable.is_a?(Disbursement)
+      # TODO: possibly replace this by adding #events to Disbursement?
+      user_list = record.commentable.source_event.users + record.commentable.destination_event.users
```
garyhtou, line 61: Yeah, let's implement `Disbursement#events` so that it uses the first case in this `if` statement and includes `ancestor_users`

app/policies/comment_policy.rb
garyhtou, line 51: This code is buggy

garyhtou: Double check whether `ancestor_users` already includes `users`

app/models/disbursement/outgoing.rb
```diff
@@ -24,6 +24,12 @@ def transaction_category
       disbursement.source_transaction_category
     end
 
+    # Label for sharing comments with the other side (destination)
+    def shared_comment_recipient_label
+      card_grant = destination_subledger&.card_grant
+      card_grant ? "Grant to #{card_grant.user.name}" : destination_event.name
```
garyhtou:
```suggestion
      card_grant ? "Grant recipient #{card_grant.user.name}" : destination_event.name
```

garyhtou (commented):
Here's another approach we can try:
- Add a select dropdown on the frontend.
	- Can be implemented as a dropdown button (GitHub style)
- Depending on the selected option, it will set a few form fields:
	- `commentable_type` and `commentable_id`
	- `return_to` (this is where the controller redirects after the POST)
		- In practice, this will just be an HcbCode URL (the url of the page which the form is being rendered)
- controller would only need to be updated to handle `return_to`
	- `redirect_back_or_to url_from(return_to) || @commentable`

garyhtou (commented):
Feedback from call:
- Remove red styling for "shared comments". Shared comments shouldn't affect the comment card border.
- Keep the "shared with X" on the top right of the comment card, but polish up the styling
- On the Disbursement show page, list comments from all three commentables (Disbursement, Outgoing Disbursement, Incoming Disbursement)
- On the Disbursement show page, the "Add comment" form adds the comment to the Disbursement commentable. The "Add comment" button should make it extremely clear that you're adding to _**both**_ sides.
	- We will not add the option to select a commentable right now (no choice between Disbursement, Outgoing Disbursement, Incoming Disbursement). The only option for commentable is the Disbursement.
- On each transaction, the _**other**_ transaction contains comments, which means there are comments not visible on the current page. Add a link to view all comments.
	- Example: Outgoing disbursement has 2 comments, Disbursement has 3 comments, Incoming disbursement has 4 comments. On the outgoing disbursement show page it says "View 4 other comments" and links to the Disbursement show page. On the incoming disbursement, it says "View 2 other comments" and links to the disbursement show page.

albertchae:
- fix showing the shared with label across incoming/outgoing vs the main disbursement
- move view other side link to below bottom comment and above text box

app/policies/comment_policy.rb
albertchae: will do this in a follow up

app/policies/comment_policy.rb
albertchae: it does because ancestor_ids includes id https://github.com/hackclub/hcb/blame/647326a0c27c89b972eb7c20560c6418769f8644/app/models/event.rb#L132-L133

will remove the confusing line above and rename the method in a follow up

albertchae:
Follow ups:
- rename ancestor_ methods or make it more obvious they include self
- [delete line 50 in comment_policy](https://github.com/hackclub/hcb/pull/13248#discussion_r3190129359)
    - ✅ https://github.com/hackclub/hcb/pull/13697/
- [add Disbursement#events and remove the branch in the comment policy](https://github.com/hackclub/hcb/pull/13248#discussion_r2997165705)
    - ✅ https://github.com/hackclub/hcb/pull/13698
- push comment ordering logic down into model (currently it's in the view)
    - ✅ https://github.com/hackclub/hcb/pull/13699
- try to redo styling for the comment box header overall

garyhtou (approved):

app/policies/comment_policy.rb
albertchae: https://github.com/hackclub/hcb/pull/13697/

PR #13249, opened 2026-03-22
[Admin] Turbo frame expensive navigation bar load

YodaLightsabr:
## Summary of the problem

The admin navigation bar takes about 1 second to load.

## Describe your changes

This PR puts it into a turbo frame while leaving a rough structure in its place, allowing the rest of the page to load in 1 second earlier.


manuthecoder (approved):
🔥

garyhtou (approved):

PR #13251, opened 2026-03-22
Experiment: Add event mentions

YodaLightsabr:
This is an experiment! It's probably not ready to merge, but I'd like feedback from people. Is this something that would be useful? If we added it, what information would you want it to have?

## Summary of the problem

Whenever we reference organizations in the admin UI, it's ugly and requires extra clicks to get more information.

## Describe your changes

Adds a new `event_mention` helper (currently only used on some admin pages). When hovered, it renders a tooltip, and when clicked, it takes you to the organization.

image
image


manuthecoder (approved):
LGTM!

PR #13263, opened 2026-03-24
[SpendingItem] Add link to HcbCode

Luke-Oldenburg:
## Summary of the problem

closes #13238


Luke-Oldenburg (commented):
I couldn't get the styling right on it. For some reason it's still grayed with underline instead of white and underline on hover.

garyhtou (approved):

garyhtou:
this is good. we can improve it in the future

PR #13267, opened 2026-03-24
[Reimbursements] Show currency on index page when necessary

YodaLightsabr:
## Summary of the problem

When you have USD and CAD reports, for example, it can be difficult to differentiate which dollar is which.

## Describe your changes

When an organization has any reimbursement reports in a non-USD currency, we'll always show the currency symbol.


manuthecoder (approved):
lgtm!

Mohamad-Mortada (approved):

PR #13274, opened 2026-03-25
Add soft delete to organizer position invites

manuthecoder:
Add deleted_at column to organizer_position_invites and create a concurrent index (migration disables DDL transaction). Enable acts_as_paranoid in OrganizerPositionInvite model.

Closes #13160

### DIFF
--- a/app/models/organizer_position_invite.rb
+++ b/app/models/organizer_position_invite.rb
@@ -55,6 +55,7 @@
 #     creation.
 #
 class OrganizerPositionInvite < ApplicationRecord
+  acts_as_paranoid
   has_paper_trail
 
   include PublicIdentifiable
--- a/db/migrate/20260325000000_add_deleted_at_to_organizer_position_invites.rb
+++ b/db/migrate/20260325000000_add_deleted_at_to_organizer_position_invites.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddDeletedAtToOrganizerPositionInvites < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_column :organizer_position_invites, :deleted_at, :datetime
+    add_index :organizer_position_invites, :deleted_at, algorithm: :concurrently
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -12,7 +12,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2026_03_12_205901) do
+ActiveRecord::Schema[8.0].define(version: 2026_03_25_000000) do
   create_schema "google_sheets"
 
   # These are extensions that must be enabled in order to support this database
@@ -1769,6 +1769,7 @@ ActiveRecord::Schema[8.0].define(version: 2026_03_12_205901) do
     t.datetime "accepted_at", precision: nil
     t.datetime "cancelled_at", precision: nil
     t.datetime "created_at", precision: nil, null: false
+    t.datetime "deleted_at"
     t.bigint "event_id", null: false
     t.boolean "initial", default: false
     t.integer "initial_control_allowance_amount_cents"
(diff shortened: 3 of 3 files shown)


app/models/organizer_position_invite.rb
```diff
@@ -55,6 +55,7 @@
 #     creation.
 #
 class OrganizerPositionInvite < ApplicationRecord
+  acts_as_paranoid
```
manu-gurudath, line 58: The model already has `cancelled_at` for soft-cancellation — what's the intended difference between calling `destroy` (sets `deleted_at`) vs `cancel` (sets `cancelled_at`)? The PR description doesn't explain the business motivation. Looking at the code, the primary driver seems to be cascading soft-deletes from `Event` (which is already paranoid and has `dependent: :destroy` on this association). If that's the only use case, it's worth calling out explicitly.

manu-gurudath, line 58: **Slug uniqueness collision risk**: The `slug` column has a `UNIQUE` DB constraint. When an invite is soft-deleted, its slug stays in the table. The `slug_candidates` method queries `OrganizerPositionInvite.where("slug LIKE ?", ...)` which, under paranoid's default scope, **won't see soft-deleted records**. So if you try to create a new invite for the same event+user after the previous one was soft-deleted (e.g. event restored after soft-delete), slug generation could produce a slug that already exists for a deleted record, causing a DB unique constraint violation.

You'd need to either:
- Scope that `where` with `with_deleted`, or
- Clear slugs on soft-delete, or  
- Confirm that re-inviting after a soft-deleted invite is not a supported workflow.

app/models/organizer_position_invite.rb
manu-gurudath, line 1: **`User` dependent destroy behavior change**: `User` is not paranoid, but has `has_many :organizer_position_invites, dependent: :destroy`. Previously hard-deleting a user would hard-delete their invites. Now it will soft-delete them instead. Is that intentional?

Also, the `pending_assign` scope (`where(accepted_at: nil, rejected_at: nil)`) has a comment: _"we need to include canceled invites so that we can assign users to them"_. With paranoid active, soft-deleted invites are silently excluded from `event.organizer_position_invites` entirely — confirm the session-assignment flow still works correctly and doesn't need `with_deleted` here.

manuthecoder (commented):
The migration is well-formed (concurrent index + `disable_ddl_transaction!` is the right pattern). A few concerns before merging:

**1. Slug uniqueness collision (concrete bug)**
The `slug` column has a DB-level `UNIQUE` constraint. After soft-deleting an invite its slug stays in the table. `slug_candidates` queries `OrganizerPositionInvite.where("slug LIKE ?", ...)` which, under paranoid's default scope, excludes soft-deleted records — so it won't account for "in-use" slugs belonging to deleted invites. Re-inviting the same user to the same event after a soft-delete (e.g. after restoring a soft-deleted event) would produce a slug collision and a DB unique constraint violation. The `not_already_invited` validation would also pass cleanly since it won't see the soft-deleted record. Options: include `with_deleted` in that query, clear the slug on soft-delete, or explicitly document that this workflow is unsupported.

**2. Semantic overlap with `cancelled_at`**
The model already has three lifecycle timestamps (`accepted_at`, `rejected_at`, `cancelled_at`). Please clarify what it means for an invite to be `deleted_at` vs `cancelled_at` — and what code path would call `destroy` rather than `cancel`. Based on the codebase the likely driver is the cascade from `Event` (which is already paranoid with `dependent: :destroy`) — if so, that's worth stating in the PR description.

**3. `User` dependent destroy behavior change**
`User` is not paranoid but has `dependent: :destroy` on invites. Hard-deleting a user previously hard-deleted their invites; now it will soft-delete them. Intentional?

**4. No tests**
Given the existing lifecycle complexity (`pending`, `pending_assign`, `cancel`, `slug_candidates`, `not_already_invited`) it would be good to have at least a regression spec confirming paranoid exclusions don't break existing flows.

garyhtou (approved):

PR #13276, opened 2026-03-26
[AASM] Make `bank_fees` table's `aasm_state` column not null

rluodev:
## Summary of the problem

column was nullable before. Confirmed via blazer 3/25 no entries in the column were null.

## Describe your changes

now, not nullable.


Luke-Oldenburg (approved):
lgtm

garyhtou (approved):

PR #13277, opened 2026-03-26
[AASM] Make `checks` table's `aasm_state` column not null

rluodev:
## Summary of the problem

previously, was nullable

## Describe your changes

now, not nullable. Confirmed via Blazer on 3/25 that there are no null entries in this column on prod


Luke-Oldenburg (approved):
lgtm

garyhtou (approved):

PR #13286, opened 2026-03-26
Safe navigate email attachment content disposition

polypixeldev:
## Summary of the problem

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/539 - while I haven't verified this, this may just be caused by the `mail` gem, as it was upgraded to a new minor version a few weeks before this issue started. Either way, it seems to just be that `content_disposition` can now be `nil`.

## Describe your changes

Add safe navigation operator when accessing `content_disposition`.


garyhtou (approved):

garyhtou (commented):
FYI @YodaLightsabr

PR #13290, opened 2026-03-26
[Event] Remove deprecated demo mode endpoint

garyhtou:
## Summary of the problem

We just moved HCB applications to this Rails application. This means that the demo mode endpoint used by the old React-based HCB application is no longer needed.

## Describe your changes

Remove the endpoint and its associated service


garyhtou:
I also created this ticket: https://github.com/hackclub/site/issues/1942

PR #13295, opened 2026-03-27
[v4] Return client secret on Payment Intent creation

Mohamad-Mortada:
## Summary of the problem
Needed to process the payment intent on the mobile app

## Describe your changes
returns client secret of that payment intent


app/controllers/api/v4/donations_controller.rb
```diff
@@ -35,6 +35,7 @@ def create
 
       def payment_intent
         amount = params[:amount_cents]
+        donation_id = params[:donation_id]
         if params[:fee_covered] && @event.config.cover_donation_fees
           amount /= (1 - @event.revenue_fee).ceil
```
garyhtou: Instead of taking in these parameters here, we should pull them off the donation that was already created

app/controllers/api/v4/donations_controller.rb
```diff
@@ -35,6 +35,7 @@ def create
 
       def payment_intent
```
garyhtou, line 37: Make this a `member` route instead of `collection`

garyhtou (changes requested):

garyhtou (approved):

PR #13296, opened 2026-03-27
[v4] Add Organizer Position Invite index & show views

Mohamad-Mortada:
## Summary of the problem
Right now it errors if we try to render an invitation or a list as we are missing templates

## Describe your changes
fixed it by adding the templates


Luke-Oldenburg (approved):
Should we rename these routes at some point to match the controller?

Mohamad-Mortada:
yes it'll need to be a breaking change tho so we need to have 2 routes then brown out the existing.

YodaLightsabr (approved):

PR #13321, opened 2026-03-28
[AASM] Make `documents` table's `aasm_state` column not null

rluodev:
## Summary of the problem

previously was nullable

## Describe your changes

now, not nullable. 6 documents currently have the column as null due to being previously deleted. https://hcb.hackclub.com/blazer/queries/1151-documents-where-aasm_state-is-null - **a command needs to be run to make these entries not null before merge**


garyhtou (changes requested):
We'll come back to this, I need to fix prod data first

garyhtou (approved):
fixed prod data

PR #13322, opened 2026-03-29
[Stripe Cardholder] Only pass verified phone number to Stripe

manuthecoder:
Closes #13288


garyhtou (approved):

garyhtou (changes requested):
Ah, actually, a few more things come to mind:
1. There are some countries HCB is unable to send SMS codes to. How do we handle this case?
2. There are a large number of existing users with stripe cards that don't have their phone number verified. Merging this PR will cause their phone number to be removed from stripe the next time their cardholder is updated. Are we okay with that? If so, should we remove their phone numbers in an OTJ now so that the paper trail is more clear?

app/jobs/one_time_jobs/clear_unverified_phone_numbers_from_stripe_cardholders.rb
polypixeldev, line 1: Would be nice if this was a maintenance task

spec/models/user_spec.rb
polypixeldev, line 1: Revert deleting these unrelated tests

spec/services/stripe_cardholder_service/create_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    it "raises when phone number is verified but event is unapproved" do
+      user = create(:user, phone_number: "+18556254225", phone_number_verified: true)
+      unapproved_event = create(:event)
+      unapproved_event.update_column(:aasm_state, "unapproved")
+
+      service = described_class.new(current_user: user, ip_address:, event_id: unapproved_event.id)
+
+      expect { service.run }.to raise_error(ArgumentError, /not permitted under spend only plan/)
+    end
```
garyhtou: this is not completely accurate
```suggestion
```

garyhtou (approved):

PR #13324, opened 2026-03-30
[Applications] Add onboarding videos to application flow

polypixeldev:
## Summary of the problem

In the contract, we mention onboarding videos. We used to optionally send these along with the contract email, but with the new application flow, we don't send a contract notification email to the signee.

## Describe your changes

Added a videos page that signees visit before signing the agreement and embeds the two YouTube onboarding videos.

image


garyhtou (approved):

garyhtou (commented):
In another PR, let's add a timestamp for when they saw the videos. This is nice info to have for compliance/legal, but also will simplify our code.

At the moment, we need to explicitly remember to redirect them to the videos if we want them to sign the agreement. However, it'd be simpler if we always sent them to the agreement show page, and then that controller determines whether you need to see the videos or not (based on whether you've already seen it)

PR #13328, opened 2026-03-30
[Organizations] Add category filtering for /api/v3/organizations

BananaJeanss:
## Summary of the problem
The /api/v3/organizations endpoint returns all organizations with no way to filter by category, meaning you'd have to paginate and do a lot more requests with data you don't actually need to find orgs with the category you need to find.

## Describe your changes
Adds an optional ?category= parameter, with valid value constraints, which only returns organizations that match the category.


app/api/api/v3.rb
```diff
@@ -310,6 +310,7 @@ def type_expansion(expand: [], hide: [])
     params do
       use :pagination, per_page: 50, max_per_page: 100
       use :expand
+      optional :category, type: String, desc: "Filter organizations by category", values: ["hack_club_hq", "robotics_team", "hackathon", "hack_club", "climate"]
```
Luke-Oldenburg, line 328:
```suggestion
      optional :category, type: String, desc: "Filter organizations by category", values: EventTag::Tags::ALL.map.entries
```

app/models/event.rb
```diff
@@ -123,6 +123,28 @@ class Event < ApplicationRecord
...
   scope :not_organized_by_teenagers, -> { includes(:event_tags).where.not(event_tags: { name: [EventTag::Tags::ORGANIZED_BY_TEENAGERS, EventTag::Tags::ORGANIZED_BY_HACK_CLUBBERS] }).or(includes(:event_tags).where(event_tags: { name: nil })) }
   scope :robotics_team, -> { includes(:event_tags).where(event_tags: { name: EventTag::Tags::ROBOTICS_TEAM }) }
+
+  CATEGORY_TAGS = {
+    "climate"       => EventTag::Tags::CLIMATE,
+    "hack_club"     => EventTag::Tags::HACK_CLUB,
+    "hackathon"     => EventTag::Tags::HACKATHON,
+    "robotics_team" => EventTag::Tags::ROBOTICS_TEAM
+  }.freeze
+
```
Luke-Oldenburg, line 135:
```suggestion
```

app/models/event.rb
```diff
@@ -123,6 +123,28 @@ class Event < ApplicationRecord
...
+    # check if tag is in hash
+    if (tag = CATEGORY_TAGS[category])
+      includes(:event_tags).where(event_tags: { name: tag })
+    elsif category == "hack_club_hq"
+      # hack_club_hq is stored as a plan not a tag
+      includes(:plan).where(event_plans: { type: [Event::Plan::HackClubAffiliate.sti_name, Event::Plan::HackClubHQ.sti_name] })
+    else
+      # unrecognized category, return nothing, though this should never make it this far since values: constraint
+      none
+    end
```
Luke-Oldenburg, line 149:
```suggestion
    includes(:event_tags).where(event_tags: { name: category })
```

Luke-Oldenburg (commented):
Thank you for making this PR! I'm not sure if we want to allow for filtering by plan. If so, I think that likely should be a separate filter. That said, the code here is mostly good for filtering by tag.

app/models/event.rb
```diff
@@ -123,6 +123,28 @@ class Event < ApplicationRecord
...
+  }.freeze
+
+  # filter organizations by category
+  scope :by_category, ->(category) {
+    # check if tag is in hash
+    if (tag = CATEGORY_TAGS[category])
+      includes(:event_tags).where(event_tags: { name: tag })
+    elsif category == "hack_club_hq"
+      # hack_club_hq is stored as a plan not a tag
+      includes(:plan).where(event_plans: { type: [Event::Plan::HackClubAffiliate.sti_name, Event::Plan::HackClubHQ.sti_name] })
```
Luke-Oldenburg, line 145: See https://github.com/hackclub/hcb/blob/main/app/api/api/entities/organization.rb#L18. I think it would be okay to add the `HackClubHQ` plan but we should do it in both places to maintain consistency.
```suggestion
      includes(:plan).where(event_plans: { type: Event::Plan::HackClubAffiliate.sti_name })
```

Luke-Oldenburg (commented):
lgtm! ignore my last review. i havent worked with this code in a while.

app/models/event.rb
```diff
@@ -123,6 +123,28 @@ class Event < ApplicationRecord
...
+
+  CATEGORY_TAGS = {
+    "climate"       => EventTag::Tags::CLIMATE,
+    "hack_club"     => EventTag::Tags::HACK_CLUB,
+    "hackathon"     => EventTag::Tags::HACKATHON,
+    "robotics_team" => EventTag::Tags::ROBOTICS_TEAM
+  }.freeze
+
+  # filter organizations by category
+  scope :by_category, ->(category) {
```
manu-gurudath, line 137: The endpoint would return zero results for all existing API consumers not using the new filter.



```suggestion
  scope :by_category, ->(category) {
  	return all if category.nil?
```

BananaJeanss: no idea how this happened cause when I tested it, it worked fine, fixed now

app/models/event.rb
```diff
@@ -123,6 +123,30 @@ class Event < ApplicationRecord
   scope :organized_by_teenagers, -> { includes(:event_tags).where(event_tags: { name: [EventTag::Tags::ORGANIZED_BY_TEENAGERS, EventTag::Tags::ORGANIZED_BY_HACK_CLUBBERS] }) }
   scope :not_organized_by_teenagers, -> { includes(:event_tags).where.not(event_tags: { name: [EventTag::Tags::ORGANIZED_BY_TEENAGERS, EventTag::Tags::ORGANIZED_BY_HACK_CLUBBERS] }).or(includes(:event_tags).where(event_tags: { name: nil })) }
   scope :robotics_team, -> { includes(:event_tags).where(event_tags: { name: EventTag::Tags::ROBOTICS_TEAM }) }
+
+  CATEGORY_TAGS = {
+    "climate"       => EventTag::Tags::CLIMATE,
+    "hack_club"     => EventTag::Tags::HACK_CLUB,
+    "hackathon"     => EventTag::Tags::HACKATHON,
+    "robotics_team" => EventTag::Tags::ROBOTICS_TEAM
+  }.freeze
```
manu-gurudath, line 134: This doesn't include all of the tags in `app/api/api/entities/organization.rb:11`

BananaJeanss: should be good now?

Luke-Oldenburg: It's missing `hack_club_hq`

Luke-Oldenburg (approved):
lgtm

PR #13337, opened 2026-03-31
[Applications] Check if videos were watched before signing agreement

polypixeldev:
## Summary of the problem

#13324 added videos to the application flow, but it's still possible for someone to get to the agreement page without watching the videos.

## Describe your changes

Add a column to applications that's set to true once videos are watched (user presses continue on the videos page, which is disabled until both videos have started playing).


app/javascript/controllers/onboarding_videos_controller.js
```diff
@@ -0,0 +1,54 @@
...
+      this.checkAllStarted()
+    }
+  }
+
+  checkAllStarted() {
+    const allStarted = Array.from(this.startedMap.values()).every(v => v)
+
+    if (allStarted) {
+      document.getElementById(this.buttonIdValue).disabled = false
+    }
```
garyhtou, line 52: Very nice!

app/javascript/controllers/onboarding_videos_controller.js
garyhtou, line 1: I'm loving how clean this stimulus controller is

db/migrate/20260331024739_add_videos_watched_to_event_applications.rb
```diff
@@ -0,0 +1,5 @@
+class AddVideosWatchedToEventApplications < ActiveRecord::Migration[8.0]
+  def change
+    add_column :event_applications, :videos_watched, :boolean
```
garyhtou, line 3: Should we add column with `default: true` and `null: false`, then change column default to `false`

app/views/event/applications/videos.html.erb
```diff
@@ -38,12 +38,12 @@
...
-      <iframe class="flex-grow aspect-[560/315] min-h-0 max-w-full" src="https://www.youtube.com/embed/LMh9FCm8iIE?si=qYV_l5IVGZMhhVvW" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen></iframe>
+      <div data-onboarding-videos-target="player" data-video-id="LMh9FCm8iIE" class="flex-grow aspect-[560/315] min-h-0 max-w-full"></div>
     </div>
   </div>
 </div>
 
 <% content_for :footer do %>
   <a class="btn bg-muted" href="<%= application_path(@application) %>">← Back</a>
-  <a class="btn bg-blue" href="<%= agreement_application_path(@application) %>">Continue →</a>
+  <%= button_to "Continue →", application_path(@application, return_to: agreement_application_path(@application)), class: "btn bg-blue", method: :patch, params: { event_application: { videos_watched: true } }, id: :continue_button, disabled: true %>
```
garyhtou, line 48: Clever to use a patch with a return_to

app/javascript/controllers/onboarding_videos_controller.js
```diff
@@ -0,0 +1,54 @@
...
+      const player = new YT.Player(el, {
+        videoId: videoId,
+        height: '100%',
+        events: {
+          onStateChange: event => this.handleStateChange(index, event),
+        },
+      })
+
+      this.players.push(player)
+      this.startedMap.set(index, false)
```
garyhtou: I wonder if it's slightly more "accurate" if we used `videoId` as the key instead of `index`.

app/controllers/event/applications_controller.rb
```diff
@@ -165,6 +165,11 @@ def videos
     def agreement
       authorize @application
 
+      unless @application.videos_watched
+        redirect_to videos_application_path(@application)
+        return
+      end
```
garyhtou, line 171: In the previous PR, we changed all the application show links to video links. I think we should change them back to the application show links and let the redirect decide where they should go.

That can either happen in this PR or another PR.

db/migrate/20260331024739_add_videos_watched_to_event_applications.rb
polypixeldev: I think we should allow nil here since it'd be confusing to set this to true for applications where the applicants were never even shown the onboarding videos

garyhtou, line 3:
```suggestion
    add_column :event_applications, :videos_watched, :boolean
    change_column_default :event_applications, :videos_watched, false
```

garyhtou (approved):

PR #13341, opened 2026-03-31
[Admin] Add transaction category dropdowns to process pages

YodaLightsabr:
## Summary of the problem

Right now, admins can't set transaction categories on transactions while they're processing. (They technically can, but it would require them manually looking up the HCB code and applying the change there—this would be a really awful workflow.)

## Describe your changes

This PR adds the transaction category dropdown to wires, Wise transfers, ACH transfers, and checks, and it makes most admin process pages side-by-side with separate fieldsets for each type of processing operation.


app/views/admin/ach_start_approval.html.erb
```diff
@@ -1,197 +1,227 @@
...
-  <p>
-    Submitted by:
-    <%= "ADMIN" if @ach_transfer.creator.auditor? %> <%= @ach_transfer.creator.name %>
-    (<%= link_to @ach_transfer.creator.phone_number, "tel:#{@ach_transfer.creator.phone_number}" %>, <%= mail_to @ach_transfer.creator.email %>) of <%= link_to @ach_transfer.event.name, @ach_transfer.event %>
-  </p>
+  <div class="flex-grow"></div>
 
-  <% if @ach_transfer.event.financially_frozen? %>
-    <p class="font-bold">(❄️ ⚠️ EVENT CURRENTLY FINANCIALLY FROZEN)</p>
+  <%= link_to("https://dashboard.column.com/app/transfers/edit/ach/#{@ach_transfer.column_id}", class: "btn btn-small bg-none !bg-[#4575CD] hover:saturate-[125%] transition-all !text-white flex-shrink-0", target: "_blank") do %>
```
garyhtou, line 14: In a separate PR, we can add a `HasColumnDashboardUrl` concern, similar to https://github.com/hackclub/hcb/blob/d6a1fc7403cf394b81ee9552e03063f23643b3f7/app/models/concerns/has_stripe_dashboard_url.rb#L3

garyhtou (approved):
Fix up the table bug and then we're good to merge

PR #13344, opened 2026-03-31
[Admin] Show individual transactions on statement of activity

YodaLightsabr:
## Summary of the problem

The statement of activity page currently doesn't show individual transactions.

## Describe your changes

This PR lets admins see individual transactions on the statement of activity page.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -870,7 +870,7 @@ class EventsController < ApplicationController
       @event,
       start_date_param: params[:start],
       end_date_param: params[:end],
-      include_descendants: ActiveRecord::Type::Boolean.new.cast(params[:include_descendants]),
+      include_descendants: ActiveRecord::Type::Boolean.new.cast(params[:include_descendants] || true),
     )
 
     respond_to do |format|
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -866,11 +866,17 @@ class EventsController < ApplicationController
   def statement_of_activity
     authorize @event
 
+    include_descendants = if params.key?(:include_descendants)
+                            ActiveRecord::Type::Boolean.new.cast(params[:include_descendants])
+                          else
+                            true
+                          end
+
     @statement_of_activity = Event::StatementOfActivity.new(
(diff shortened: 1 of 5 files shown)


app/controllers/events_controller.rb
```diff
@@ -870,7 +870,7 @@ def statement_of_activity
       @event,
       start_date_param: params[:start],
       end_date_param: params[:end],
-      include_descendants: ActiveRecord::Type::Boolean.new.cast(params[:include_descendants]),
+      include_descendants: ActiveRecord::Type::Boolean.new.cast(params[:include_descendants] || true),
```
garyhtou, line 873: I think you wanna drop this change, right?

app/views/events/_statement_of_activity_contents.html.erb
```diff
@@ -1,5 +1,14 @@
 <%# locals: (statement_of_activity:) %>
 
+<% included_events = statement_of_activity.events
+   category_totals = statement_of_activity.category_totals
+   transactions_by_category = statement_of_activity.transactions_by_category
+   transactions = transactions_by_category.values.flatten
+   total_revenue = transactions.sum { |transaction| transaction.amount_cents.positive? ? transaction.amount_cents : 0 }
+   total_expense = transactions.sum { |transaction| transaction.amount_cents.negative? ? transaction.amount_cents : 0 }
+   net_asset_change = transactions.sum(&:amount_cents)
```
garyhtou, line 9: Some of these are probably better defined as methods on `StatementOfActivity` — but it's fine to leave them here right now

garyhtou (approved):
Please drop changes to app/controllers/events_controller.rb, then good to merge

app/controllers/events_controller.rb
YodaLightsabr: This defaults to true if it's not present, but it won't override an explicit false value (since `params[:include_descendants]` is always a string). Do you think we should include or exclude sub-orgs by default?

garyhtou: ah, it's inside the parens. hm, i don't have a preference. this is good to go

PR #13354, opened 2026-04-01
[v4] Indexing of donations 

Mohamad-Mortada:
## Summary of the problem
I want to recreate the donations page that we have on web in mobile.

## Describe your changes
Added a new index method to donations controller which paginates past donations and returns data in a similar form as web where we have recurring donations as a standalone list, stats, and past donations


app/views/api/v4/donations/index.json.jbuilder
```diff
@@ -0,0 +1,29 @@
...
+  json.name rd.name(show_anonymous: false)
+  json.email rd.email
+  json.started_on rd.created_at.to_date
+  json.amount_cents rd.amount
+  json.total_donated_cents rd.total_donated
+end
+
+pagination_metadata(json)
+
+json.past_donagio @past_donations do |donation|
```
manu-gurudath:
```suggestion
json. past_donations @past_donations do |donation|
```

app/controllers/api/v4/donations_controller.rb
```diff
@@ -4,10 +4,28 @@ module Api
   module V4
     class DonationsController < ApplicationController
       include SetEvent
+      include ApplicationHelper
```
manu-gurudath: do we need this?

Mohamad-Mortada: we pull the `paginate` method from helper

app/controllers/api/v4/donations_controller.rb
```diff
@@ -4,11 +4,30 @@ module Api
...
+        authorize @event, :show_in_v4?
+
+        @recurring_donations = @event.recurring_donations.order(created_at: :asc)
+
+        all_past_donations = @event.donations
+                                   .where(aasm_state: params[:status] || [:in_transit, :deposited, :refunded])
+                                   .order(created_at: :desc)
+                                   .to_a
+
+        @past_donations = paginate(all_past_donations, &:public_id)
```
Mohamad-Mortada:
```suggestion
        @past_donations = paginate_cursor(all_past_donations, &:public_id)
```

app/controllers/api/v4/donations_controller.rb
```diff
@@ -4,11 +4,28 @@ module Api
...
-      before_action :set_api_event, only: [:create]
+      before_action :set_api_event, only: [:index, :create]
       before_action :set_donation, only: [:payment_intent]
       before_action :require_trusted_oauth_app!, only: [:payment_intent]
 
+      def index
+        authorize @event, :show_in_v4?
+
+        all_past_donations = @event.donations
+                                   .where(aasm_state: params[:status] || [:in_transit, :deposited, :refunded])
```
garyhtou: Return 400 if invalid param status.

Don't list all states here bc it's brittle when we add a new state. Instead, only call `where` when `params[:status].present?`.

example: https://github.com/hackclub/hcb/blob/6d5ccaeb04ed498ca1ac1ed0df2894daae598751/app/controllers/events_controller.rb#L463

garyhtou: From a user perspective, they never see the `in_transit` state in the UI bc we have instant deposits.

app/controllers/api/v4/donations_controller.rb
```diff
@@ -4,11 +4,32 @@ module Api
...
 
-      before_action :set_api_event, only: [:create]
+      before_action :set_api_event, only: [:index, :create]
       before_action :set_donation, only: [:payment_intent]
       before_action :require_trusted_oauth_app!, only: [:create, :payment_intent]
 
+      def index
+        authorize @event, :show_in_v4?
+
+        donations = @event.donations.order(created_at: :desc)
```
garyhtou:
```suggestion
        donations = @event.donations.not_pending.order(created_at: :desc)
```

app/controllers/api/v4/donations_controller.rb
```diff
@@ -4,11 +4,32 @@ module Api
...
+
+        if params[:status].present?
+          valid_statuses = Donation.aasm.states.map { |s| s.name.to_s }
+          return render json: { error: "invalid_operation", messages: ["'#{params[:status]}' is not a valid status."] }, status: :bad_request unless valid_statuses.include?(params[:status])
+
+          donations = donations.where(aasm_state: params[:status])
+        end
+
+        @past_donations = paginate_cursor(donations.to_a, &:public_id)
```
garyhtou:
```suggestion
        @donations = paginate_cursor(donations.to_a, &:public_id)
```
and rename the view too

app/views/api/v4/donations/index.json.jbuilder
```diff
@@ -0,0 +1,14 @@
...
+if expand?(:stats)
+  json.stats do
+    json.total_cents @total_cents
+    json.monthly_cents @monthly_cents
+  end
+end
+
+pagination_metadata(json)
+
+json.data @past_donations do |donation|
```
garyhtou:
```suggestion
json.data @donations do |donation|
```

app/controllers/api/v4/donations_controller.rb
```diff
@@ -4,11 +4,32 @@ module Api
...
+
+        if expand?(:stats)
+          @total_cents = @event.donations.succeeded_and_not_refunded.sum(:amount)
```
garyhtou:
```suggestion
          @total_cents = @event.donations.not_pending.succeeded_and_not_refunded.sum(:amount)
```

app/controllers/api/v4/donations_controller.rb
```diff
@@ -4,11 +4,32 @@ module Api
...
+          @monthly_cents = @event.recurring_donations.active.sum(:amount)
```
garyhtou:
```suggestion
```

app/views/api/v4/donations/index.json.jbuilder
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+if expand?(:stats)
+  json.stats do
+    json.total_cents @total_cents
+    json.monthly_cents @monthly_cents
```
garyhtou:
```suggestion
```

app/controllers/api/v4/donations_controller.rb
```diff
@@ -4,11 +4,32 @@ module Api
...
+      def index
+        authorize @event, :show_in_v4?
+
+        donations = @event.donations.order(created_at: :desc)
+
+        if params[:status].present?
+          valid_statuses = Donation.aasm.states.map { |s| s.name.to_s }
+          return render json: { error: "invalid_operation", messages: ["'#{params[:status]}' is not a valid status."] }, status: :bad_request unless valid_statuses.include?(params[:status])
+
+          donations = donations.where(aasm_state: params[:status])
```
garyhtou: We want to hide the "in_transit" state from the user. I'd recommend adding a new concept of a "visible_state" (give it a better name). Here's some very rough and buggy psuedo code

```ruby
class Donation
	# VISIBLE_STATE_MAPPING = {
	# 	in_transit: :deposited
	# }

	include VisibleStatable
	set_visible_state_mapping({
		in_transit: :deposited
	})

end

class VisibleStatable
	include Concern

	included do
		def visible_state
			@visible_state_mapping[aasm_state] || aasm_state
		end

		def filter_by_visible_state(state)
			value = @visible_state_mapping.invert[state] || state
			unless self.aasm.states.map { |s| s.name }.include?(value)
				raise ArgumentError, "invalid state"
			end
			self.where(aasm_state: value)
		end
	end

	class_methods do
		def set_visible_state_mapping(mapping)
			@visible_state_mapping = mapping
		end
	end
end
```

app/controllers/api/v4/donations_controller.rb
polypixeldev: Not anymore - it's a controller concern now, so we should be able to remove this

app/models/donation.rb
```diff
@@ -66,6 +66,9 @@ class Donation < ApplicationRecord
   set_public_id_prefix :don
 
   include AASM
+  include VisibleStatable
+  set_visible_state_mapping(in_transit: :deposited)
```
polypixeldev: I don't think this is technically correct - if an event can't front balance, we don't say it's deposited

Mohamad-Mortada: I spoke with Gary, this is expected behavior as we render donations as instantly deposited

app/controllers/api/v4/donations_controller.rb
```diff
@@ -24,7 +24,7 @@ def index
 
         @donations = paginate_cursor(donations.to_a, &:public_id)
 
-        if expand?(:stats)
+        if @expand.include?(:stats)
```
polypixeldev, line 21: fine with doing this now but it would be nice to have the expand? helper available in controllers as a concern or smth in the future

app/models/concerns/visible_statable.rb
```diff
@@ -0,0 +1,41 @@
...
+      state_sym = state.to_sym
+
+      internal_states = visible_state_mapping.keys.select { |internal| resolve_visible_state(internal, event) == state_sym }
+
+      if aasm.states.map(&:name).include?(state_sym)
+        masked = visible_state_mapping.key?(state_sym) && resolve_visible_state(state_sym, event) != state_sym
+        internal_states |= [state_sym] unless masked
+      end
+
+      raise ArgumentError, "invalid state" if internal_states.empty?
```
polypixeldev: We should not raise an error, as the state may or may not be valid depending on the event

app/models/concerns/visible_statable.rb
polypixeldev, line 1: Since this concern is now coupled with event, it'd make sense to include event somewhere in the name

polypixeldev (approved):

PR #13355, opened 2026-04-01
[v4][Breaking Change] Shallow route indexing of invoices and sponsors

Mohamad-Mortada:
## Summary of the problem
Our vision is to move most API endpoints to be shallow routed rather than nesting. Currently, i dont believe anyone is using this API route and is the easiest to migrate. 

## Describe your changes
Shallow routed invoices and sponsors


garyhtou (approved):
I confirmed today via AppSignal logs that these two endpoints were NOT being used.

PR #13357, opened 2026-04-01
[v4] Add deletion requests of team members

Mohamad-Mortada:
## Summary of the problem
Can't delete team members on HCB Mobile

## Describe your changes
I extracted out organizer positions into its own controller and added its own index method which we will later use to replace pulling the team members from `/api/v4/organizations/xxx`. Then added POST `/api/v4/organization_positions/xxx/removal_request` to request removal of a team member


app/views/api/v4/events/_event.json.jbuilder
```diff
@@ -31,6 +31,7 @@ if policy(event).account_number? && expand?(:account_number)
   json.swift_bic_code event.bic_code
 end
 
+# TODO: Remove users field once migration to /api/v4/organizations/:id/team is done
```
Mohamad-Mortada:
```suggestion
# TODO: Remove users field once migration to /api/v4/organizer_positions is done
```

app/models/organizer_position.rb
```diff
@@ -34,6 +34,12 @@ class OrganizerPosition < ApplicationRecord
   include OrganizerPosition::HasRole
   include OrganizerPosition::HasSpending
 
+  include Hashid::Rails
+  hashid_config salt: ""
```
manuthecoder: why is this an empty string?

app/controllers/api/v4/organizer_positions_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+        authorize @event, :show_in_v4?
+        @organizer_positions = @event.organizer_positions.includes(:user).order(created_at: :desc)
+      end
+
+      def removal_request
+        authorize @organizer_position, :can_request_removal?
+
+        @deletion_request = @organizer_position.organizer_position_deletion_requests.build(
+          submitted_by: current_user,
+          reason: params.require(:reason)
```
manuthecoder, line 22: `reason` can be blank, is this intentional

app/views/api/v4/organizer_positions/index.json.jbuilder
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+json.organizers @organizer_positions do |op|
+  json.id op.public_id
+  json.role op.role
+  json.joined_at op.created_at
+  json.is_signee op.is_signee
+  json.user do
+    json.partial! "api/v4/users/user", user: op.user
```
manuthecoder: missing show_email

config/routes.rb
```diff
@@ -719,6 +719,12 @@
           end
         end
 
+        resources :organizer_positions, only: [:index] do
+          member do
+            post "removal_request"
+          end
+        end
```
manuthecoder, line 769: the new resource is registered at the top level unlike other collection resources in the v4 API (which are nested)

app/models/organizer_position.rb
Mohamad-Mortada: All are like this

config/routes.rb
Mohamad-Mortada: Intentional

app/controllers/api/v4/organizer_positions_controller.rb
Mohamad-Mortada: Model will handle error messages

app/views/api/v4/organizer_positions/index.json.jbuilder
Mohamad-Mortada: fixed

app/views/api/v4/organizer_positions/_organizer_position.json.jbuilder
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# locals: (json:, organizer_position:)
+
+object_shape(json, organizer_position) do
+  json.role organizer_position.role
+  json.signee organizer_position.is_signee
+  json.user do
```
Mohamad-Mortada: Would u recommend using `expand_association` here. Where I debate here is that user is an important value in an OP and without it, the partial seems a bit too empty

app/controllers/api/v4/organizer_positions_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+module Api
+  module V4
+    class OrganizerPositionsController < ApplicationController
+      include SetEvent
+
+      before_action :set_api_event, only: [:index]
+      before_action :set_organizer_position, only: [:removal_request]
+
+      def index
+        authorize @event, :index_in_v4?
```
polypixeldev: index_in_v4? is a little confusing - that reads like authorizing getting a list of events, not a list of OPs

app/models/organizer_position.rb
Mohamad-Mortada:
```suggestion
```

polypixeldev: Remove this line - there's no existing links using the IDs with no salt

app/views/api/v4/organizer_positions/_organizer_position.json.jbuilder
polypixeldev: I would use it since everywhere else does

app/controllers/api/v4/organizer_positions_controller.rb
Mohamad-Mortada:
```suggestion
        authorize @event, :show_in_v4?
```

polypixeldev (approved):

PR #13366, opened 2026-04-01
[v4] expose user emails if user is part of the same org

Mohamad-Mortada:
## Summary of the problem
Currently if I query for a list of my team members on an org I can't see their emails which may be useful 

## Describe your changes
Added that emails in a user partial are visible ONLY if that user shares an org with the current user


app/helpers/api/v4/application_helper.rb
```diff
@@ -60,6 +60,15 @@ def expand(*keys)
         @expand = before
       end
 
+      def shares_org_with?(user)
+        return false if current_user.nil?
```
garyhtou:
```suggestion
        return false if current_user.nil? || user.nil?
```

garyhtou (changes requested):
An alternative method is to have the user partial take in a `show_email` param which is set to `true` when rendered from the `OrganizerPositionsController#index`.

This prevents confusion about when `email` is accessible. With this implementation, the `email` key may appear and disappear in a confusing manner depending on whether you share an org with someone or not (e.g. when we support listing team members of transparent orgs)

app/helpers/api/v4/application_helper.rb
```diff
@@ -60,6 +60,15 @@ def expand(*keys)
         @expand = before
       end
 
+      def shares_org_with?(user)
+        return false if current_user.nil? || user.nil?
+
+        @current_user_event_ids ||= OrganizerPosition.where(user: current_user).pluck(:event_id)
```
garyhtou:
```suggestion
        @current_user_event_ids ||= current_user.readable_events.pluck(:id)
```

app/helpers/api/v4/application_helper.rb
```diff
@@ -60,6 +60,15 @@ def expand(*keys)
...
+        return false if @current_user_event_ids.empty?
+
+        OrganizerPosition.where(user:, event_id: @current_user_event_ids).exists?
```
garyhtou:
```suggestion
        @current_user_event_ids.to_set.intersect?(User.find(7727).readable_events.pluck(:id))
```

garyhtou (commented):
Write some tests, then good to merge

spec/controllers/api/v4/events_controller_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+      request.headers["Authorization"] = "Bearer #{token.token}"
+
+      get :show, params: { id: event.friendly_id, expand: "users" }, as: :json
+
+      expect(response).to have_http_status(:ok)
+      me = response.parsed_body["users"].find { |u| u["id"] == current_user.public_id }
+      expect(me).to include("email" => "me@example.com")
+    end
+
+    it "always includes users in the show response" do
```
garyhtou: Do we want this test case? I think with agreed we don't want to expand by default in the future (which would contradict with this test)

spec/controllers/api/v4/events_controller_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+      shared_event = create(:event)
+      create(:organizer_position, user: current_user, event: shared_event)
+      create(:organizer_position, user: other_user,   event: shared_event)
+
+      helper = helper_instance.call(current_user)
+      expect(current_user).to receive(:readable_events).once.and_call_original
+
+      helper.shares_org_with?(other_user)
+      helper.shares_org_with?(other_user)
+    end
```
garyhtou, line 170: We should test sub events too. LMK if you have questions on what those tests would look like.

garyhtou (approved):

PR #13367, opened 2026-04-01
[v4][Breaking Change] Shallow route ACH transfer

Mohamad-Mortada:
## Summary of the problem
Another API I doubt any person uses right now which needs to be switched to use our new API standard


garyhtou (approved):
I verified this is not being used via AppSignal logs (no hits in the past 30 days)

PR #13369, opened 2026-04-02
[BREAKING CHANGE][v4] Shallow route tags + CRUD

Mohamad-Mortada:
## Summary of the problem
Follow new API standard of shallow routing + allow creation and deletion of org tags

## Describe your changes
Added create, show, and delete methods and shallow routed the index route


garyhtou (approved):
I've confirmed this endpoint wasn't used. so breaking change should affect no one

Image

PR #13374, opened 2026-04-02
[v4] Create Wire transfer

Mohamad-Mortada:
## Summary of the problem
Wires are always fun :) 

## Describe your changes
Added wire controller with a create method and added support of wire transfer types for v4 transaction view


app/views/api/v4/transactions/_wire_transfer.json.jbuilder
```diff
@@ -0,0 +1,25 @@
...
+json.address_line2 wire_transfer.address_line2
+json.address_city wire_transfer.address_city
+json.address_state wire_transfer.address_state
+json.address_postal_code wire_transfer.address_postal_code
+
+json.created_at wire_transfer.created_at
+
+json.sender do
+  json.partial! "api/v4/users/user", user: wire_transfer.user
+end
```
Luke-Oldenburg:
```suggestion
end

```

app/views/api/v4/wires/show.json.jbuilder
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+json.partial! "api/v4/transactions/wire_transfer", wire_transfer: @wire
```
Luke-Oldenburg:
```suggestion
json.partial! "api/v4/transactions/wire_transfer", wire_transfer: @wire

```

YodaLightsabr (approved):

Luke-Oldenburg (commented):
I don't think this handles receipts.

app/controllers/api/v4/wires_controller.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module Api
+  module V4
+    class WiresController < ApplicationController
+      include SetEvent
+
+      before_action :set_api_event, only: [:create]
+
+      def create
```
Luke-Oldenburg, line 21: Is it possible to extract this to its own method identical to https://github.com/hackclub/hcb/blob/c2ed2bd69bf4a90fcd04d1b2758347ef404465a2/app/controllers/wires_controller.rb#L110-L130?

config/routes.rb
```diff
@@ -758,6 +758,7 @@
         resources :checks, only: [:index, :create, :show]
         resources :sponsors, only: [:index, :show, :create]
         resources :check_deposits, only: [:index, :show, :create]
+        resources :wires, only: [:create]
```
polypixeldev: add index and show

in another pr, add it to ach as well

app/controllers/api/v4/wires_controller.rb
polypixeldev: +1

app/controllers/api/v4/wires_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+          :address_postal_code,
+          :send_email_notification,
+          :file,
+          *Wire.recipient_information_accessors
+        )
+        @wire = @event.wires.build(wire_params.merge(user: current_user))
+
+        authorize @wire
+
+        if @wire.usd_amount_cents > SudoModeHandler::THRESHOLD_CENTS
```
polypixeldev, line 26: worth noting that wires controller uses amount_cents, not usd_amount_cents. I think wires_controller should be using usd_amount_cents

PR #13382, opened 2026-04-02
[Disbursements] Show logo when `id` is not set in URL

manuthecoder:
The disbursements page rendered without any navigation when accessed without an event ID — no back button, no logo, leaving users with no way out.

Now the HCB logo is shown in that state, giving users a consistent anchor point to navigate back.


garyhtou (approved):
Before:

Image

After:
Image

Other transfer methods (remains unchanged):
Image

PR #13387, opened 2026-04-02
[DocuSeal] Return 401 instead of 200 with `success: false`

garyhtou:
## Summary of the problem

At the moment, when the webhook is unauthorized, we return a 200 with `success: false`. However, I can't find anywhere in the DocuSeal docs where it asks for this.

## Describe your changes

It's a lot more standard to return a 4XX when unauthorized. So I'm doing that here — returning a 401 when unauthorized.

Either way, HCB has been and continues to be secure from forged webhooks here.

### DIFF
--- a/app/controllers/docuseal_controller.rb
+++ b/app/controllers/docuseal_controller.rb
@@ -4,11 +4,11 @@ class DocusealController < ActionController::Base
   protect_from_forgery except: :webhook
 
   def webhook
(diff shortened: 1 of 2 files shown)


polypixeldev (approved):
LGTM and tested locally with DocuSeal!

garyhtou:
TYSM!

PR #13389, opened 2026-04-02
[Accounting] Auto-categorize card grant disbursements

YodaLightsabr:
## Summary of the problem

Disbursements that are automatically generated from card grants are currently uncategorized. However, we know that they fall under the "Grants / stipends" category, so we can assign it automatically.

## Describe your changes

This PR automatically categorizes card grant disbursements as "Grants / stipends".

### DIFF
--- a/app/jobs/one_time_jobs/backfill_category_on_card_grant_disbursements.rb
+++ b/app/jobs/one_time_jobs/backfill_category_on_card_grant_disbursements.rb
@@ -7,7 +7,8 @@ module OneTimeJobs
       slug = "grants-stipends"
       category = TransactionCategory.find_or_create_by!(slug:)
 
-      disbursements.update(source_transaction_category: category, destination_transaction_category: category)
+      disbursements.where(source_transaction_category: nil).update(source_transaction_category: category)
+      disbursements.where(destination_transaction_category: nil).update(destination_transaction_category: category)
 
       hcb_codes = []
       disbursements.find_each(batch_size: 100) do |disbursement|
@@ -16,6 +17,8 @@ module OneTimeJobs
       end
 
       CanonicalTransaction.where(hcb_code: hcb_codes).find_each(batch_size: 100) do |ct|
+        next unless ct.category.nil?
+
         TransactionCategoryService
           .new(model: ct)
           .set!(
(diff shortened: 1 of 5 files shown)


app/models/card_grant.rb
garyhtou, line 1: Does this properly set `assignment_strategy` as "automatic"?

app/jobs/one_time_jobs/backfill_category_on_card_grant_disbursements.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module OneTimeJobs
+  class BackfillCategoryOnCardGrantDisbursements < ApplicationJob
+    def perform
+      disbursements = Disbursement.where.not(source_subledger_id: nil, destination_subledger_id: nil)
+      slug = "grants-stipends"
+      category = TransactionCategory.find_or_create_by!(slug:)
+
+      disbursements.update(source_transaction_category: category, destination_transaction_category: category)
```
garyhtou: Only do this if category was NOT already set

garyhtou: Wait.... the concept of storing `source_transaction_category` and `destination_transaction_category` might be obsolete now that we've split disbursement HcbCodes... Let me know what you think

garyhtou: It's just a thought that's cross my mind, but I haven't investigated deeper

YodaLightsabr: good catch

YodaLightsabr: I think we still need it to create the canonical pending transactions

YodaLightsabr: fixed

app/models/card_grant.rb
YodaLightsabr: It does now!

garyhtou (approved):

PR #13392, opened 2026-04-03
Rename Interac to Interac e-Transfer

Badbird5907:
## Summary of the problem

Most canadians call an Interac transfer an "e-Transfer". Interac refers to the overall payment network (they also handle debit cards), while e-Transfer is the particular service used for sending money.

## Describe your changes

Renamed "Interac" to "Interac e-Transfer"


YodaLightsabr (approved):
Thanks for submitting this!

PR #13399, opened 2026-04-03
[Accounting] Split inter- and intra- internal disbursements

YodaLightsabr:
## Summary of the problem

Intra-organization transfers always total $0, so they should be in their own category.

## Describe your changes

This PR splits the categories up.

### DIFF
--- a/app/models/event/statement_of_activity.rb
+++ b/app/models/event/statement_of_activity.rb
@@ -69,6 +69,7 @@ class Event
     memo_wise def category_totals
       totals = transactions.includes(:category).group("category.slug").sum(:amount_cents)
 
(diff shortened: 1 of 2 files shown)


app/models/event/statement_of_activity.rb
```diff
@@ -51,15 +59,26 @@ def supports_descendants?
...
-        category_totals[category.slug] # I'm using SQL calculated totals since it is faster than Array's sum(&:amount_cents)
+        category_totals[category.slug]
       end.to_h
     end
 
     memo_wise def category_totals
-      transactions.includes(:category).group("category.slug").sum(:amount_cents)
+      totals = transactions.includes(:category).group("category.slug").sum(:amount_cents)
+
+      if totals.key?(INTERNAL_TRANSFER_SLUG)
```
garyhtou, line 73:
```suggestion
      # Ideally, we'd use the preprocessed transactions categories provided by `grouped_transactions`. However, in this case, it's more efficient to use the raw `transactions` so that we can compute the majority of these category totals in SQL. Then, we use `grouped_transactions` to correct/replace the category totals for Internal Transfers (which get split into Intra/Inter org transfers)
      if totals.key?(INTERNAL_TRANSFER_SLUG)
```

garyhtou (approved):

PR #13401, opened 2026-04-03
[v4][dev-docs] Document v4 API standards

Mohamad-Mortada:
## Summary of the problem
The v4 API isn't very consistent with its implementation. We need some sort of structure before we continue to expand 

## Describe your changes
I've written up a draft on what HCB's standards are for creating v4 API routes and what we require each route to have


dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,307 @@
...
+  "type": "transaction",
+  "created_at": "2025-03-15T12:00:00Z",
+  "amount_cents": -4500,
+  "memo": "Amazon Web Services"
+}
+```
+
+---
+
+## IDs & Object References
```
Luke-Oldenburg, line 166: When creating a new public ID prefix it should be 3 letters.

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,307 @@
...
+---
+
+## Object Shape
+
+Every top-level API object **must** include the following fields:
+
+| Field        | Type     | Description                                      |
+|--------------|----------|--------------------------------------------------|
+| `id`         | `string` | The public ID of the object (e.g. `txn_abc123`). |
+| `type`       | `string` | A machine-readable type label (e.g. `transaction`, `ach_transfer`). |
```
Luke-Oldenburg: Let's rename this to object.

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,307 @@
...
+}
+```
+
+| Field         | Type      | Description                                              |
+|---------------|-----------|----------------------------------------------------------|
+| `total_count` | `integer` | Total number of results matching the query.              |
+| `has_more`    | `boolean` | Whether more results exist beyond this page.             |
+| `data`        | `array`   | The page of results.                                     |
+
+### Query Parameters
```
Luke-Oldenburg: pagination?

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,307 @@
...
+      "id": "org_h1izp",
+      "object": "organization",
+      "name": "Hack Club HQ"
+    }
+  }
+  ```
+
+### Guidelines
+
+- Use the `expand?(:symbol)` helper in Jbuilder views to conditionally render expanded objects.
```
Luke-Oldenburg:
```suggestion
- Use the `expand?(:symbol)` helper in jbuilder views to conditionally render expanded objects.
```

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,307 @@
...
+Check each endpoint's documentation for its supported expansions. Common ones include:
+
+| Expansion         | Available On                             |
+|-------------------|------------------------------------------|
+| `organization`    | transactions, cards, card grants         |
+| `user`            | cards, card grants                       |
+| `balance_cents`   | organizations                            |
+| `account_number`  | organizations (requires permission)      |
+| `users`           | organizations                            |
+| `total_spent_cents` | cards                                  |
```
Luke-Oldenburg: Let's standardize the formatting on the tables in here. Should it be lined up or just one space after the word?

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,307 @@
...
+
+---
+
+## Partials & Reuse
+
+Every API-representable model **must** have a single canonical Jbuilder partial (e.g. `_ach_transfer.json.jbuilder`). All endpoints that render that object must use the partial.
```
Luke-Oldenburg:
```suggestion
Every API-representable model **must** have a single canonical jbuilder partial (e.g. `_ach_transfer.json.jbuilder`). All endpoints that render that object must use the partial.
```

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,307 @@
...
+
+### Why?
+
+- A single source of truth for each object's shape.
+- Developers rely on the same fields appearing whether the object is returned from a show, list, or nested context.
+- Reduces drift and copy-paste bugs.
```
Luke-Oldenburg: This sounds a bit like AI.

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,307 @@
...
+
+### Rules
+
+1. **One partial per model.** Located at `app/views/api/v4/<resource>/_<resource>.json.jbuilder`.
+2. **Nest, don't duplicate.** If a transaction includes an ACH transfer, render `partial: "api/v4/transactions/ach_transfer"`.
+3. **Show endpoints are thin.** A `show.json.jbuilder` should be essentially:
+   ```ruby
+   json.partial! @ach_transfer
+   ```
+4. **List endpoints wrap in the pagination envelope**, then render partials for each item.
```
Luke-Oldenburg: This also sounds a bit like AI.

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,307 @@
...
+
+### Example
+
+```ruby
+# app/views/api/v4/ach_transfers/_ach_transfer.json.jbuilder
+
+json.id ach_transfer.public_id
+json.type "ach_transfer"
```
Luke-Oldenburg: Is there a way that we can make this automatically be added to all jbuilders using Rails functions based on the model?

Luke-Oldenburg (commented):
Looks pretty good. Few minor things. We should also link to this from the code style guide and/or contributing guide.

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,332 @@
...
+
+```ruby
+def json_object(json, object)
+  json.id object.public_id
+  json.object object.model_name.element
+  json.created_at object.created_at
+end
+```
+
+This helper derives the `object` field automatically from the model's class name (e.g. an `AchTransfer` record produces `"ach_transfer"`), so you never need to hardcode it.
```
YodaLightsabr, line 44: Is there ever a world where the object name will differ from the active record model name? What should we do in those (hopefully unlikely) cases?

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,332 @@
...
+
+Use the `json_object` helper to set these three fields consistently across all partials:
+
+```ruby
+def json_object(json, object)
+  json.id object.public_id
+  json.object object.model_name.element
+  json.created_at object.created_at
+end
+```
```
garyhtou, line 42:
```suggestion
```ruby
def object_shape(json, object, &block)
  json.id object.public_id
  json.object object.model_name.element

  block.call

  json.created_at object.created_at
end
```
```

or call it `api_object`

The advantages of this is:
- `created_at` is at the end of the hash
- It's clear what's contained within that object (e.g. everything inside the block is a part of the object)

File: dev-docs/v4-api/standards.md (hunk 10 above)
Comment (garyhtou) [inline dev-docs/v4-api/standards.md:44]: We need to have an "override" list. In case we want to rename `Event` to `organization`

File: dev-docs/v4-api/standards.md [hunk 12]
```diff
@@ -0,0 +1,332 @@
...
+
+This helper derives the `object` field automatically from the model's class name (e.g. an `AchTransfer` record produces `"ach_transfer"`), so you never need to hardcode it.
+
+Every canonical partial should begin with a call to `json_object`:
```
garyhtou:
```suggestion
Every object's partial should begin with a call to `json_object`:
```

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,332 @@
...
+      "id": "org_h1izp",
+      "object": "organization",
+      "name": "Hack Club HQ"
+    }
+  }
+  ```
+
+### Guidelines
+
+- Use the `expand?(:symbol)` helper in jbuilder views to conditionally render expanded objects.
```
garyhtou, line 212: Ensure that `expand?(:symbol)` knows to only expand only level

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,332 @@
...
+
+To reduce payload size and unnecessary database work, related objects are **not** included by default. Developers opt in using the `expand` query parameter.
+
+```
+GET /api/v4/cards/crd_x9f3k?expand=user,organization
+```
+
+### How It Works
+
+- Without expansion, a related object appears as an ID string:
```
garyhtou, line 194: Clarify how expansion works for arrays

dev-docs/v4-api/standards.md
garyhtou: api/v4//user/1?expand=organization,transactions

```ruby
{
	id: "usr_123",
	object: "user",
	transactions: [
		{
			id: "txn_123",
			object: "transactions"
		},
		{
			id: "txn_234",
			object: "transactions"
		},
	],
	organizations: [
		{
			id: "txn_123"
			object: "organization",
			# we don't want `expand?` to expand a `transactions` key here.
		},
	],
}
```

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,332 @@
...
+
+### Implementation Notes
+
+- Cursors should be the `public_id` of the last item on the current page.
+
+---
+
+## Expanding Related Objects
+
+To reduce payload size and unnecessary database work, related objects are **not** included by default. Developers opt in using the `expand` query parameter.
```
garyhtou: Clarify that the `expand` query takes in the `key` in the object space you want expanded. It's not the type of the association.

The reason why this is important is because of the case of disbursements where we return a `source_organization` and `destination_organization`

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,332 @@
...
+
+### Standard Error Codes
+
+| Code                 | HTTP Status | Meaning                                        |
+|----------------------|-------------|------------------------------------------------|
+| `invalid_auth`       | `401`       | Missing or invalid API token.                  |
+| `not_authorized`     | `403`       | Token is valid but lacks permission.           |
+| `not_found`          | `404`       | Resource does not exist.                       |
+| `invalid_operation`  | `400`/`422` | Request is well-formed but violates a business rule. |
+| `validation_error`   | `422`       | One or more fields failed validation.          |
```
garyhtou: This is somewhat of a breaking change, but maybe we consider using Rail's error names

https://gist.github.com/mlanett/a31c340b132ddefa9cca#file-rails-http-status-codes-L16

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,332 @@
...
+All monetary values are represented in **cents** (the smallest currency unit) as integers. Never use floats for money.
+
+```json
+{
+  "amount_cents": 4200,
+  "balance_cents": 150000
+}
+```
+
+---
```
garyhtou, line 350: Add: How does the v4 API operate for admins?

By default, the API behaves as if you have "pretend not be an admin" on. Only when the API token has an `admin` scope, will you gain admin permissions.

There should be multiple admin scopes:
- `admin:read`
- `admin:write`

YodaLightsabr: +1 on this behavior for admins. Is this current behavior? Or what we want to see in an ideal world.

dev-docs/v4-api/standards.md
YodaLightsabr: +1

dev-docs/v4-api/standards.md
```diff
@@ -0,0 +1,357 @@
...
+## Table of Contents
+
+- [Object Shape](#object-shape)
+- [IDs & Object References](#ids--object-references)
+- [Shallow Routing](#shallow-routing)
+- [Pagination](#pagination)
+- [Expanding Related Objects](#expanding-related-objects)
+- [Partials & Reuse](#partials--reuse)
+- [Error Responses](#error-responses)
+- [Admin Access](#admin-access)
```
garyhtou, line 16: Let's add:
- Basic
	- We only take and respond with JSON
- Authentication
	- How to create a new oauth app
	- The fact that you need to refresh your tokens
- Rate limits

garyhtou (commented):
Let's add a section on how we handle deprecations. and also a "fast-tracked deprecation" for things we believe no one's using and will be deprecated quicker or with minimal warning.

dev-docs/v4-api/standards.md
```diff
@@ -24,18 +25,47 @@ Every top-level API object **must** include the following fields:
...
+
+```ruby
+# app/views/api/v4/ach_transfers/_ach_transfer.json.jbuilder
+
+json_object(json, ach_transfer)
+
+json.recipient_name ach_transfer.recipient_name
+json.amount_cents ach_transfer.amount
+# ...
+```
```
garyhtou, line 66: Let's use the block syntax we chatted about last time

dev-docs/v4-api/standards.md
```diff
@@ -24,18 +25,47 @@ Every top-level API object **must** include the following fields:
...
+
+This produces:
 
 ```json
 {
-  "id": "txn_h1izp",
-  "type": "transaction",
+  "id": "ach_x9f3k",
+  "object": "ach_transfer",
   "created_at": "2025-03-15T12:00:00Z",
```
garyhtou, line 74: With the block syntax, `created_at` would be at the end.

dev-docs/v4-api/standards.md
```diff
@@ -156,6 +187,7 @@ GET /api/v4/cards/crd_x9f3k?expand=user,organization
   ```json
   { "organization_id": "org_h1izp" }
   ```
+  - An exception to this would be object arrays as those can't be represented as an ID string
```
garyhtou, line 198: Let's explicitly say what happens with object arrays. Possibilities include:

- Key is `organization_ids`. An array of string IDs
- An array of objects that only includes an `organization_id`
- Or, we disallow arrays of objects to be in a response of another object. This is because an array of objects should only be provided via an index endpoint of said object.
	- e.g. User object should never include a list of organizations. Instead, you should call `/organizations?user_id=usr_123`.
	- The only exception is if the array does not contain API objects. This should be very rare.

I'm leaning towards the last option.

dev-docs/v4-api/standards.md
```diff
@@ -255,18 +285,37 @@ Errors follow a consistent shape:
...
 ### Standard Error Codes
 
-| Code                 | HTTP Status | Meaning                                        |
-|----------------------|-------------|------------------------------------------------|
-| `invalid_auth`       | `401`       | Missing or invalid API token.                  |
-| `not_authorized`     | `403`       | Token is valid but lacks permission.           |
-| `not_found`          | `404`       | Resource does not exist.                       |
-| `invalid_operation`  | `400`/`422` | Request is well-formed but violates a business rule. |
-| `validation_error`   | `422`       | One or more fields failed validation.          |
+| Code                 | HTTP Status | Meaning                                              |
```
garyhtou, line 296: Add a commet saying that this should mirror Rails' HTTP status code names

https://kapeli.com/cheat_sheets/HTTP_Status_Codes_Rails.docset/Contents/Resources/Documents/index

dev-docs/v4-api/standards.md
```diff
@@ -255,18 +285,37 @@ Errors follow a consistent shape:
...
+|----------------|----------------------------------------------------------------------|
+| `admin:read`   | Grants read-only access to admin-level data (e.g. all organizations, internal fields). |
+| `admin:write`  | Grants the ability to perform write actions reserved for admins.     |
+
+`admin:write` does **not** imply `admin:read` — both scopes must be granted independently if both are needed.
+
+### Implementation Notes
+
+- Always check for the admin scope explicitly. Do not fall back to checking if the authenticated user is an admin.
+- Endpoints that expose admin-only data or actions must document which scope they require.
```
garyhtou, line 323: In order to gain admin permissions, you must have BOTH the scope on the API token and the `access_level` on the user

garyhtou (commented):
Add a requirement to use strict locals
https://blog.appsignal.com/2024/09/11/ruby-on-rails-7-1-partial-strict-locals-and-their-gotchas.html

dev-docs/v4-api/standards.md
```diff
@@ -242,31 +342,36 @@ Every API-representable model **must** have a single canonical jbuilder partial
...
+  json.status ach_transfer.aasm_state
+
+  json.organization ach_transfer.event, partial: "api/v4/events/event", as: :event if expand?(:organization)
+  json.sender do
+    if ach_transfer.creator.present?
+      json.partial! "api/v4/users/user", user: ach_transfer.creator
+    else
+      json.nil!
+    end
   end
```
garyhtou:
```suggestion
  if expand?(:organization)
    json.organization ach_transfer.event, partial: "api/v4/events/event", as: :event
  else
    json.organization_id ach_transfer.event.public_id
  end
  
  if expand?(:sender)
    json.sender do
      if ach_transfer.creator.present?
        json.partial! "api/v4/users/user", user: ach_transfer.creator
      else
        json.nil!
      end
    end
  else
    json.sender_id ach_transfer.creator&.public_id
  end
```

garyhtou: Additionally, we should consider adding a helper method like

```ruby
def expand_association(name, associated_object)
  if expand?(name)
    json.send(name) do
      if associated_object.present?
        json.partial! "TODO need dynamic path to associated_object's partial", user: associated_object
      else
        json.nil!
      end
    end
  else
    json.send(:"#{name}_id", associated_object&.public_id)
  end
end
```

garyhtou (approved):

PR #13415, opened 2026-04-06
Force utf8 comment encodings

polypixeldev:
## Summary of the problem

https://hackclub.slack.com/archives/C047Y01MHJQ/p1759419037175179 & https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/790/samples/timestamp/2026-04-06T19:18:30Z - somehow, we got an email with `ISO-8859-1` encoding! When decrypted though, Rails assumes that it's utf8 encoding, throwing an error.

## Describe your changes

This seems to only be happening in CommentMailbox due to various encodings coming from emails and then us encrypting the content, so I added `encode` to ensure that we always get `utf8` before encrypting and saving to the DB. If it's already utf8 (majority of cases), this won't do anything. I've also already fixed the encoding on the comment causing this error.

### DIFF
--- a/app/mailboxes/comment_mailbox.rb
+++ b/app/mailboxes/comment_mailbox.rb
@@ -21,7 +21,7 @@ class CommentMailbox < ApplicationMailbox
     return bounce_missing_comment unless @comment && @commentable && ensure_permissions?
 
     comment = @commentable.comments.build({
-                                            content: EmailReplyParser.parse_reply(text || body),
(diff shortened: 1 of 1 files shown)


garyhtou (approved):

garyhtou:
There's one instance in production that was fixed: https://hcb.hackclub.com/console/sessions/1583

PR #13434, opened 2026-04-08
[Reimbursements] Show banners when orgs have insufficient funds

manuthecoder:
closes #13426

This pull request introduces a balance validation warning on the reimbursement report page

Added validation to alert users when a reimbursement exceeds available event funds:
- Implemented `exceeds_event_balance?` method in `Reimbursement::Report` to check if the requested amount surpasses the event's available USD balance for submitted reports
- Updated the report view to display an informational callout when the balance is exceeded on open reports, notifying users that additional funding may be required

image


YodaLightsabr (changes requested):
We currently already have a system in place for alerting users of insufficient funds:
https://github.com/hackclub/hcb/blob/86da14e3e28dcd86a149b8629062550c6564505b/app/views/reimbursement/reports/_actions.html.erb#L175
It looks like the only difference between the existing logic and this PR's logic is that this PR shows the warning even when a report requires peer review. If this difference is intentional, we should update the existing callout rather than creating a new one.

app/models/reimbursement/report.rb
```diff
@@ -317,6 +317,11 @@ def initial_draft?
       draft? && submitted_at.nil?
     end
 
+    def exceeds_event_balance?
+      !draft? && !closed? && currency == "USD" && event.present? &&
+        !::Shared::AmpleBalance.ample_balance?(amount_to_reimburse_cents, event)
+    end
```
polypixeldev, line 323: I would recommend moving the state check (draft/closed) to the view where this method is used, since I would expect `exceeds_event_balance?` to work regardless of report state. This would also allow you to use this method in https://github.com/hackclub/hcb/blob/mxg-showing-banners-on-reimbursements-when-orgs-have-insufficient-funds/app/views/reimbursement/reports/_actions.html.erb#L170

app/models/reimbursement/report.rb
```diff
@@ -317,6 +317,11 @@ def initial_draft?
       draft? && submitted_at.nil?
     end
 
+    def exceeds_event_balance?
+      !draft? && !closed? && currency == "USD" && event.present? &&
+        !::Shared::AmpleBalance.ample_balance?(amount_to_reimburse_cents, event)
```
polypixeldev:
```suggestion
        !::Shared::AmpleBalance.ample_balance?(amount_cents, event)
```
`amount_to_reimburse_cents` only takes into account approved expenses. My understanding from the linked issue is that we want this message to show after the report is submitted, even if the report hasn't been reviewed yet, so we should use `amount_cents` instead.

PR #13334, opened 2026-03-31
Add fading to overflowing callouts to indicate scrolling

polypixeldev:
## Summary of the problem

Currently, we use an overflowing callout on the contract FAQ since there's a lot of items. However, if the browser automatically hides the scroll bar, it can be unclear that it is a scrollable container.

Closes #13240

## Describe your changes

Adds fading (credit to stackoverflow) to overflowing callouts to indicate scrolling.

https://github.com/user-attachments/assets/8cfe1e52-a8a0-4ac3-b042-10fe911d5447

PR #13436, opened 2026-04-08
Fix callout styling

polypixeldev:
## Summary of the problem

#13334 added fading	to overflowing contracts to indicate scrolling. However, it didn't attach event listeners on Turbo load, applied fading when it shouldn't have, and removed padding on non-overflowing callouts.

## Describe your changes

Attach listeners on turbo load, add check for if an element scrolls, and fixes the padding.


YodaLightsabr (approved):
LGTM! Thanks for fixing this.

PR #13445, opened 2026-04-09
[Applications] Allow archiving any application and mark them as drafts

polypixeldev:
## Summary of the problem

People have submitted applications or even had them approved but decided to not continue with the application, and wish to archive them. Currently the view doesn't allow this, and archived applications stay in the Airtable views that onboarders use.

## Describe your changes

Add the archive/unarchive button to the footer for submitted and approved applications, syncs this status to Airtable and adds archival status to the submission page, sends submitted/under-review archived applications back to draft, and adds archival info the submission page.


garyhtou (approved):

garyhtou (changes requested):
Add logic here to move an application from `submitted` or `under_review` to draft, then to archiving.

https://github.com/hackclub/hcb/blob/833337f8e1fcf5d3c449895800f47675b8327b7d/app/models/event/application.rb#L362-L361

garyhtou:
Now that we're allowing Applications to go back to the Draft state. Think about edge cases. For example, draft applications can now have contracts associated. What do we do with those? (maybe void them).

Ensure links continue to work and say "this application was archived/withdrawn."

app/models/event/application.rb
```diff
@@ -371,7 +362,8 @@ def archive!
     end
 
     def unarchive!
-      send_contract if contract.nil? && ((teen_led && !draft? && !rejected?) || (!teen_led && approved?))
+      send_contract if contract.nil? && approved?
+      mark_draft! if may_mark_draft?
```
garyhtou: We want to `mark_draft!` inside `archive!` so that these applications aren't sitting in the `under_review` state when they don't need op's review

garyhtou (approved):

PR #13447, opened 2026-04-09
Let users request onboarding calls from HCB

polypixeldev:
## Summary of the problem

With the new application flow, no one is doing an onboarding call unless they ask for it. However, these onboarding calls can still be very beneficial, especially since the HCB tour is very minimal.

## Describe your changes

Signees will be able to request an onboarding call after they have been activated. There are two cases:
- Their assigned PoC has a scheduling link in the new table in the applications Airtable
  - We redirect them straight to that link
- Their assigned PoC does not have a scheduling link
  - We send an email to hcb@hackclub.com, and some from ops can reply to that email to schedule with the user

The option to request an onboarding call is linked in the activation email (which opens a modal first for confirmation), and is also on a card in the event homepage. This card will disappear if you have transactions or if you've already clicked on the button to schedule a call.

image
image


app/controllers/events_controller.rb
```diff
@@ -1039,6 +1040,27 @@ def merchants_filter
...
   end
 
+  def request_call
+    authorize @event
+
+    if @event.point_of_contact.present?
+      onboarder_record = OnboardersTable.all(filter: "{HCB ID} = #{@event.point_of_contact.id}").first
+
+      if onboarder_record.present?
+        @event.config.update!(hide_onboarding_message: true)
```
garyhtou, line 1050: Do we wanna use `Event.demo_mode_request_meeting_at`? The column is poorly named (we no longer call it "demo mode"), but it may be nice to store the timestamp, especially since we already have the column.

polypixeldev: I agree it may be nice to store the timestamp. However, we do have PaperTrail on `Event::Configuration`, and `hide_onboarding_message` allows us to manually hide the message without saying that a meeting was requested. For example, the migration in this PR adds this column with a default of true (which is immediately changed to false) so that this message doesn't show up for any orgs that may have already had an onboarding call.

garyhtou: That makes sense. I'm cool with that solution!

Should we consider dropping `demo_mode_request_meeting_at` then?

YodaLightsabr: I'd be in favor of dropping `demo_mode_request_meeting_at` in another PR

polypixeldev: Yeah, makes sense to me - will do!

app/views/events/_request_call_modal.html.erb
```diff
@@ -0,0 +1,9 @@
+<%# locals: (auto_open: false) %>
+
+<section id="request_call_modal" class="modal modal--scroll max-w-md bg-snow" data-behavior="modal" role="dialog" data-modal-auto-open="<%= auto_open.to_s %>" data-modal="request_call_modal">
+  <%= modal_header("Schedule onboarding call") %>
+  <p>If you have questions or would like to go over the HCB plaform with a member of our team, feel free to request a call!</p>
+
+  <%= button_to "Request", request_call_event_path(@event), class: "left bg-success btn", method: :post %>
+  <a href="#" class="right bg-error btn" rel="modal:close">Cancel</a>
```
YodaLightsabr: Do we have cancel buttons in other modals on HCB? It seems a little out of place here (based on the screenshot you added). I'd suggest either:

* Getting rid of the cancel button (the × should be fine on its own)
* Change the design up a little more, swap the side of the buttons, expand them to be full width, and convert it to a partial to essentially serve as a generic "confirm this action" component

polypixeldev: This is similar to the modal we use for reversing automatically paired receipts, when someone clicks the link from an email. I agree that the cancel button isn't necessary here - just removed it, here's an updated screenshot:
image

app/mailers/event_mailer.rb
```diff
@@ -50,6 +50,14 @@ def negative_balance
     mail(to: @emails, subject: "#{@event.name} has a negative balance")
   end
 
+  def call_requested
+    @user = params[:user]
+
+    mail to: @event.point_of_contact.email_address_with_name,
+         subject: "#{@user.name} from #{@event.name} #{"(PoC #{@event.point_of_contact.name})" if @event.point_of_contact.present?} requested an onboarding call",
+         reply_to: @user.email_address_with_name
```
garyhtou:
```suggestion
    mail to: OPERATIONS_EMAIL,
         subject: "#{@user.name} requesting an onboarding call for #{@event.name} #{"with #{@event.point_of_contact.name}" if @event.point_of_contact.present?}",
         reply_to: @user.email_address_with_name
```

app/views/event_mailer/call_requested.html.erb
```diff
@@ -0,0 +1,14 @@
+<p>Hi there! 👋</p>
+
+<p>
+  <%= link_to "#{@user.name} (#{@user.email})", admin_user_url(@user) %>
+  from <%= link_to @event.name, @event %> <%= "(point of contact: #{@event.point_of_contact.name})" if @event.point_of_contact.present? %> has requested an onboarding call.
```
garyhtou:
```suggestion
  <%= link_to "#{@user.name}", edit_user_url(@user) %> (<%= @user.email %>)
  has requested an onboarding call for <%= link_to @event.name, @event %><%= " with #{@event.point_of_contact.name}" if @event.point_of_contact.present? %>.
```

app/views/events/show.html.erb
```diff
@@ -77,6 +77,23 @@
     </div>
   <% end %>
 
+  <% if policy(@event).request_call? %>
+    <%= render "events/request_call_modal", auto_open: @requesting_call %>
```
garyhtou:
```suggestion
    <%= render "events/request_call_modal", auto_open: @requesting_call if @requesting_call %>
```

garyhtou: You can also entirely get rid of the auto_open local variable.

app/views/events/show.html.erb
```diff
@@ -77,6 +77,23 @@
...
+
+    <% unless @event.config.hide_onboarding_message || @event.hcb_codes.exists? %>
```
garyhtou, line 83: Feedback for another PR:

Handle the `@event.hcb_codes.exists?` logic in a background job that runs maybe daily?

This lets us do less on page load, and makes `hide_onboarding_message` the source of truth (we only need to check that one boolean)

app/controllers/events_controller.rb
```diff
@@ -1066,6 +1067,27 @@ def merchants_filter
...
+
+        redirect_to onboarder_record["Scheduling Link"], allow_other_host: true
+        return
+      end
+    end
+
+    EventMailer.with(event: @event, user: current_user).call_requested.deliver_now
+    @event.config.update!(hide_onboarding_message: true)
+
+    flash[:success] = "A member of our team will respond to schedule a call soon!"
```
```suggestion
    flash[:success] = "A member of our team will reach out to schedule a call soon!"
```

garyhtou (approved):

app/views/events/show.html.erb
garyhtou: We should up the heuristic to maybe 10 transactions

PR #13450, opened 2026-04-09
[Invoices] Replace sponsor select with custom dropdown with delete support

alextheradu:
## Summary

- Replaces the native `<select>` in the invoice form's sponsor step with a custom dropdown
- Each sponsor row shows a delete button (×) on hover — only visible to users with destroy permission
- Deleting a sponsor sends a `DELETE` request via `fetch`, removes the row from the dropdown without a page reload, and resets the selection if that sponsor was active
- If all sponsors are deleted, the dropdown hides and the new sponsor form appears automatically
- The "updating a previous sponsor" warning now correctly shows/hides based on whether an existing sponsor is selected

### DIFF
--- a/app/controllers/sponsors_controller.rb
+++ b/app/controllers/sponsors_controller.rb
@@ -51,8 +51,14 @@ class SponsorsController < ApplicationController
     authorize @sponsor
 
     @sponsor.destroy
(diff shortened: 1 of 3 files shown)


Luke-Oldenburg:
@claude review

Luke-Oldenburg:
@claude review

PR #13453, opened 2026-04-10
[Receipts] Remove force captures from missing receipt scope

YodaLightsabr:
## Summary of the problem

We currently have no way of checking the direction of a transaction from the HCB code missing receipt scope. As a result, positive force captures are flagged as missing a receipt.

## Describe your changes

This PR removes force captures from the missing receipt scope.


Luke-Oldenburg (approved):
lgtm

garyhtou (approved):

PR #13454, opened 2026-04-10
[v4] Improve transaction pagination performance

Mohamad-Mortada:
## Summary of the problem
Poor performance ❌  Better performance 👍 

We shouldn't be finding the `public_id` of the list of transactions as that loads HCB Code object which is poor performance as we do this multiple times. 

## Describe your changes
We should just find the hcb code of the param then compare. So becomes 1 DB request instead of however many transactions it takes to find the param one.

### DIFF
--- a/app/controllers/api/v4/transactions_controller.rb
+++ b/app/controllers/api/v4/transactions_controller.rb
@@ -94,7 +94,7 @@ module Api
                         cursor_hcb_code = HcbCode.find_by_public_id(params[:after])&.hcb_code
                         index = transactions.index { |tx| tx.hcb_code == cursor_hcb_code }
                         return render json: { error: "Invalid cursor" }, status: :bad_request unless index
-      
+
                         index + 1
                       else
                         0
--- a/app/controllers/api/v4/transactions_controller.rb
+++ b/app/controllers/api/v4/transactions_controller.rb
@@ -91,7 +91,11 @@ module Api
       def paginate_transactions(transactions)
         limit = params[:limit]&.to_i || 25
         start_index = if params[:after]
-                        transactions.index { |tx| tx.local_hcb_code.public_id == params[:after] } + 1
+                        cursor_hcb_code = HcbCode.find_by_public_id(params[:after])&.hcb_code
+                        index = transactions.index { |tx| tx.hcb_code == cursor_hcb_code }
+                        return render json: { error: "Invalid cursor" }, status: :bad_request unless index
+      
+                        index + 1
                       else
                         0
                       end
--- a/app/controllers/api/v4/transactions_controller.rb
+++ b/app/controllers/api/v4/transactions_controller.rb
@@ -93,7 +93,7 @@ module Api
         start_index = if params[:after]
                         cursor_hcb_code = HcbCode.find_by_public_id(params[:after])&.hcb_code
                         index = transactions.index { |tx| tx.hcb_code == cursor_hcb_code }
-                        return render json: { error: "Invalid cursor" }, status: :bad_request unless index
+                        return render json: { error: "bad_request", messages: ["invalid cursor"] }, status: :bad_request unless index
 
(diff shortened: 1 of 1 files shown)


app/controllers/api/v4/transactions_controller.rb
```diff
@@ -91,7 +91,11 @@ def mark_no_receipt
       def paginate_transactions(transactions)
         limit = params[:limit]&.to_i || 25
         start_index = if params[:after]
-                        transactions.index { |tx| tx.local_hcb_code.public_id == params[:after] } + 1
+                        cursor_hcb_code = HcbCode.find_by_public_id(params[:after])&.hcb_code
+                        index = transactions.index { |tx| tx.hcb_code == cursor_hcb_code }
```
garyhtou, line 97: Is it more efficient for us to just compare db IDs here?

app/controllers/api/v4/transactions_controller.rb
```diff
@@ -91,7 +91,11 @@ def mark_no_receipt
...
+                        return render json: { error: "Invalid cursor" }, status: :bad_request unless index
```
garyhtou: Let's double check this error response matches our API standards doc

Mohamad-Mortada: oop. autocomplete tricked me

app/controllers/api/v4/transactions_controller.rb
garyhtou, line 97: Would this work?

```suggestion
                        cursor_id = HcbCode.find_by_public_id(params[:after])&.id
                        index = transactions.index { |tx| tx.id == cursor_id }
```

garyhtou: On a second thought, i think transactions is an array of `CanonicalTransactionGrouped`, not `HcbCode`. So this might not work. Scrap this idea lol

garyhtou (approved):

app/controllers/api/v4/transactions_controller.rb
```diff
@@ -91,7 +91,11 @@ def mark_no_receipt
       def paginate_transactions(transactions)
         limit = params[:limit]&.to_i || 25
         start_index = if params[:after]
-                        transactions.index { |tx| tx.local_hcb_code.public_id == params[:after] } + 1
+                        cursor_hcb_code = HcbCode.find_by_public_id(params[:after])&.hcb_code
+                        index = transactions.index { |tx| tx.hcb_code == cursor_hcb_code }
+                        return render json: { error: "bad_request", messages: ["invalid cursor"] }, status: :bad_request unless index
```
garyhtou, line 98: We'd usually say `invalid_operation` for the error key. But i think we wanna move towards `bad_request` (and other Rails http terms), so maybe this is okay to merge as is.

garyhtou (approved):

app/controllers/api/v4/transactions_controller.rb
Mohamad-Mortada: yea im tryna move to our newer convention names

app/controllers/api/v4/transactions_controller.rb
```diff
@@ -91,7 +91,11 @@ def mark_no_receipt
       def paginate_transactions(transactions)
         limit = params[:limit]&.to_i || 25
         start_index = if params[:after]
-                        transactions.index { |tx| tx.local_hcb_code.public_id == params[:after] } + 1
+                        cursor_hcb_code = HcbCode.find_by_public_id(params[:after])&.hcb_code
```
garyhtou, line 94:
```suggestion
                        cursor_hcb_code = HcbCode.find_by_public_id(params[:after])&.hcb_code
                        return render json: { error: "bad_request", messages: ["invalid cursor"] }, status: :bad_request unless cursor_hcb_code

```

PR #13455, opened 2026-04-10
[Applications] Improve ready to submit logic

polypixeldev:
## Summary of the problem

Some required fields have been added since `Event::Application#ready_to_submit?` was added, and the method treated empty strings as present. Additionally, while submission was prevented in the controller and reflected in the view, missing fields was not a guard in AASM. Finally, the summary partial was missing a field and was unclear as to which fields were optional or required.

## Describe your changes

Update the ready to submit methods with the current form, add ready to submit as an AASM guard, update the summary partial, and update the teen_led status of draft applications when the user birthday changes.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -182,6 +182,8 @@ class User < ApplicationRecord
 
   after_update :queue_sync_with_loops_job
 
+  after_update :update_draft_applications, if: -> { birthday_previously_changed? }
+
   before_update :set_default_seasonal_theme
 
   validates_presence_of :full_name, if: -> { full_name_in_database.present? }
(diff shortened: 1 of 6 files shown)


app/models/user.rb
```diff
@@ -716,4 +718,8 @@ def send_onboarded_email
     UserMailer.onboarded(user: self).deliver_later
   end
 
+  def update_draft_applications
+    applications.draft.each { |application| application.update!(teen_led: is_teenager?) }
```
YodaLightsabr, line 722: Good catch

app/views/event/applications/_summary.html.erb
```diff
@@ -4,6 +4,10 @@
   <em>Blank</em>
 <% end %>
 
+<% missing = capture do %>
+  <em class="text-primary">Missing</em>
```
YodaLightsabr, line 8: Good idea!

YodaLightsabr (approved):
Nice work!

PR #13461, opened 2026-04-11
Add `verified` to `User`

YodaLightsabr:
## Summary of the problem

We're looking to allow users to create new accounts and set certain attributes on `User` prior to verifying their email address. As a result, we want to store in the DB whether or not a user is verified.

## Describe your changes

Before we can build out protections against unverified users, we want to make sure that all existing users are backfilled as verified and that new users signing in get marked as verified.

### DIFF
--- a/db/migrate/20260411143808_add_verified_to_user.rb
+++ b/db/migrate/20260411143808_add_verified_to_user.rb
@@ -1,0 +1,5 @@
+class AddVerifiedToUser < ActiveRecord::Migration[8.0]
+  def change
+    add_column :users, :verified, :boolean, default: false, null: false
+  end
+end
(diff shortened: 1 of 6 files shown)


db/migrate/20260411143808_add_verified_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddVerifiedToUser < ActiveRecord::Migration[8.0]
+  def change
+    add_column :users, :verified, :boolean, default: false, null: false
```
garyhtou, line 3:
```suggestion
    add_column :users, :verified, :boolean, default: true, null: false
    change_column_default :users, :verified, from: true, to: false
```

PR #13467, opened 2026-04-12
[IncreaseCheck] Enable stopping

Luke-Oldenburg:
## Summary of the problem

Previously, you can only stop checks via the Column dashboard. This means you need to ask ops/engr.

## Describe your changes

Now we allow managers of organizations to stop a check


app/models/reimbursement/payout_holding.rb
```diff
@@ -107,7 +107,7 @@ def reverse!
       raise ArgumentError, "must be a failed payout holding" unless failed?
       raise ArgumentError, "ACH must have been rejected / failed" unless ach_transfer.nil? || ach_transfer.failed? || ach_transfer.rejected?
       raise ArgumentError, "PayPal transfer must have been rejected" unless paypal_transfer.nil? || paypal_transfer.rejected?
-      raise ArgumentError, "a check is present" if increase_check.present?
+      raise ArgumentError, "a check must have been rejected / stopped" unless increase_check.column_rejected? || increase_check.column_stopped?
```
Luke-Oldenburg:
```suggestion
      raise ArgumentError, "a check must have been rejected / stopped" unless increase_check.nil? || increase_check.column_rejected? || increase_check.column_stopped?
```

app/models/increase_check.rb
```diff
@@ -325,48 +325,47 @@ def can_stop?
...
+      address_zip:,
+      payment_recipient_id:,
+      reissued_for_id: id,
+    )
+
+    reissued_check.save!
+
+    reimbursement_payout_holding.update!(increase_check_id: reissued_check.id) if reimbursement_payout_holding.present?
+
+    Receipt.reupload(old_receiptable: local_hcb_code, new_receiptable: reissued_check.local_hcb_code) unless reimbursement_payout_holding.present?
```
garyhtou, line 366:
```suggestion
    Receipt.reupload(old_receiptable: local_hcb_code, new_receiptable: reissued_check.local_hcb_code)
```

dev-docs/guides/reimbursements_transaction_engine.md
```diff
@@ -53,7 +53,9 @@ raise ArgumentError, "ACH must have been rejected / failed" unless ach_transfer.
 
 raise ArgumentError, "PayPal transfer must have been rejected" unless paypal_transfer.nil? || paypal_transfer.rejected?
 
-raise ArgumentError, "a check is present" if increase_check.present?
+raise ArgumentError, "a check must have been rejected / stopped" unless increase_check.column_rejected? || increase_check.column_stopped?
```
Luke-Oldenburg:
```suggestion
raise ArgumentError, "a check must have been rejected / stopped" unless increase_check.nil || increase_check.column_rejected? || increase_check.column_stopped?
```

app/models/increase_check.rb
```diff
@@ -325,48 +325,47 @@ def can_stop?
...
+      recipient_email:,
+      send_email_notification:,
+      address_zip:,
+      payment_recipient_id:,
+      reissued_for_id: id,
+    )
+
+    reissued_check.save!
+
+    reimbursement_payout_holding.update!(increase_check_id: reissued_check.id) if reimbursement_payout_holding.present?
```
```suggestion
    if reimbursement_payout_holding.present?
      reimbursement_payout_holding.update!(increase_check_id: reissued_check.id) 
      reimbursement_payout_holding.mark_settled!
    end
```

garyhtou (changes requested):
1. Add failed state to IncreaseCheck
2. Add `after` callback to `mark_Failed `in IncreaseCheck that's responsible for:
    - marking Payout Holding as failed
    - sending email to recipient
3. Disallow reissuing reimbursement checks. The only flow we will support for the time being is:
    1. Stop reimbursement check (which marks payout holding as failed)
    2. User updates their payout method settings
    3. nightly job reattempts the reimbursement payment

---

In the future:
- Update payout methods to be an "append-only" log. This way we have all historical payout methods with one selected as their default.
- Payout Holdings or Reimbursement Report will have a payout method associated with this (payout methods are "frozen", so it will never change; allowing us to reissuing).
- This will allow the existing Payout Holding Service Nightly Job to handle reissuing of both Checks (but also other payout method such as Ach Transfers). The advantage of re-using Payout Holding Service Nightly Job is that it already has all the error handling we need.

garyhtou (approved):

PR #13478, opened 2026-04-13
[CPT] Move admin tool around fee

Luke-Oldenburg:
## Summary of the problem

This was rendering weird and unnecessary as it's an admin-only page.


YodaLightsabr (dismissed):
https://github.com/hackclub/hcb/blob/d89a93326edb24b1716f71eb3c75faebe684828a/app/policies/canonical_pending_transaction_policy.rb#L5

garyhtou (approved):

PR #13480, opened 2026-04-13
[BREAKING CHANGE][V4] Ignore Event#demo_mode_request_meeting_at before removing

polypixeldev:
## Summary of the problem

See https://github.com/hackclub/hcb/pull/13447#discussion_r3061503131 for context. This column is no longer being used and isn't needed in the database anymore.

## Describe your changes

Ignores the column and removes it from the v4 API event partial.

⚠️ **This is a v4 API breaking change, since this column's presence is currently included in the event partial.** Not sure what our current process is for this - @Mohamad-Mortada?

### DIFF
--- a/app/views/api/v4/events/_event.json.jbuilder
+++ b/app/views/api/v4/events/_event.json.jbuilder
@@ -10,6 +10,7 @@ json.financially_frozen event.financially_frozen?
 json.icon event.logo.attached? ? Rails.application.routes.url_helpers.url_for(event.logo) : nil
 json.donation_page_available event.donation_page_available?
 json.playground_mode event.demo_mode?
+json.playground_mode_meeting_requested nil # This field is deprecated and will be removed
 json.transparent event.is_public?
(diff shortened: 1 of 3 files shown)


Mohamad-Mortada:
For v4 breaking changes, I would give some period of notice in #hcb-dev before dropping the field. I don't believe anyone is using this however.

app/views/api/v4/events/_event.json.jbuilder
```diff
@@ -10,7 +10,6 @@ json.financially_frozen event.financially_frozen?
 json.icon event.logo.attached? ? Rails.application.routes.url_helpers.url_for(event.logo) : nil
 json.donation_page_available event.donation_page_available?
 json.playground_mode event.demo_mode?
-json.playground_mode_meeting_requested event.demo_mode_request_meeting_at.present?
```
garyhtou, line 13: make it return `NULL` and add a comment saying this is deprecated and will be removed

garyhtou (approved):

PR #13496, opened 2026-04-17
AdminUI3

manu-gurudath:
This PR brings the admin interface in line with HCB's modernized UI. The visual changes are intentionally more subtle than the original UI3 overhaul: 

## Main objectives
- consolidate `admin.scss` and `application.scss` as much as possible
- keep components as consistent as possible between the consumer and admin facing hcb
- refactors navigation, similar to the original UI3 branch


garyhtou:
We'll take a look at this after finishing up the payments/contractor feature!

PR #13500, opened 2026-04-17
[Tests] Speed up `:with_positive_balance` factory trait

garyhtou:
## Summary of the problem

> Previously, creating `event, :with_positive_balance` ran the full RawCsvTransaction -> HashedTransaction -> CanonicalTransaction import pipeline via three real services. That pipeline is already covered by its own specs; in the `:with_positive_balance` trait we only need a mapped positive canonical transaction.

^ Claude

## Describe your changes

> Insert the CanonicalTransaction + CanonicalEventMapping directly instead. Profiled locally, this saves ~15 seconds across the 12 spec files that use this trait.

^ Claude

### DIFF
--- a/spec/factories/event_factory.rb
+++ b/spec/factories/event_factory.rb
@@ -36,11 +36,8 @@ FactoryBot.define do
       # Event#balance is derived from the sum of amount_cents on mapped
       # canonical_transactions (see Event#settled_balance_cents). We insert
       # a single positive CanonicalTransaction + mapping directly instead of
-      # running the full RawCsv -> Hashed -> Canonical import pipeline, which
-      # costs ~1 second per use and is already covered by its own specs.
-      #
-      # 100_000 cents ($1,000) matches the original amount created by the
-      # import pipeline (RawCsvTransaction#amount=1_000 is monetized to cents).
+      # running the full Raw*Transaction -> Hashed -> Canonical import
+      # pipeline, which costs ~1 second per use.
       after :create do |event|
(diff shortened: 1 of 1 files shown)


spec/factories/event_factory.rb
```diff
@@ -33,21 +33,17 @@
...
 
     trait :with_positive_balance do
+      # Event#balance is derived from the sum of amount_cents on mapped
+      # canonical_transactions (see Event#settled_balance_cents). We insert
+      # a single positive CanonicalTransaction + mapping directly instead of
+      # running the full RawCsv -> Hashed -> Canonical import pipeline, which
+      # costs ~1 second per use and is already covered by its own specs.
+      #
+      # 100_000 cents ($1,000) matches the original amount created by the
+      # import pipeline (RawCsvTransaction#amount=1_000 is monetized to cents).
```
garyhtou: Clear up potentially confusing comment when read without the context of old code.

```suggestion
      # Event#balance is derived from the sum of amount_cents on mapped
      # canonical_transactions (see Event#settled_balance_cents). We insert
      # a single positive CanonicalTransaction + mapping directly instead of
      # running the full Raw*Transaction -> Hashed -> Canonical import
      # pipeline, which costs ~1 second per use.
```

garyhtou (commented):
The downside here is if any tests rely on the `transaction_source` of CTs, then that wouldn't be available here since we aren't creating the `RawCsvTransaction`.

However, that's a pretty niche need; and in which case, you can manually set up that data in the test case.

PR #13516, opened 2026-04-18
[Card Locking] Fix duplicate warning notification bug

garyhtou:
## Summary of the problem

Within the card locking feature, we queue multiple jobs (when an authentication hits, and 24 hrs afterwards) to check whether a user is in compliance and to send warning emails.

There exists a bug that Zach ran into, where duplicate warning notifications are sent. This happens when multiple `UserService::SendCardLockingNotification` jobs run during the period the user has exactly 5, 7, or 9 missing receipts (our warning intervals).

## Describe your changes

As a minimalistic fix, I'm just using `Rails.cache` to dedup these warning notifications. I'm planning to completely overhaul this system soon, so there's no point in sinking more time into this old system.

https://github.com/hackclub/hcb/issues/13194

Cons of this implementation: For each warning notification interval (5, 7, and 9 missing receipts), you'll receive an email max once every 24 hr. This means, if in the span of 24 hrs, you go above X missing receipts (e.g. 9), get notified, go below X, and cross back over X again, you'll only be notified once when you really should be notified twice.

### DIFF
--- a/app/services/user_service/send_card_locking_notification.rb
+++ b/app/services/user_service/send_card_locking_notification.rb
@@ -13,19 +13,23 @@ module UserService
       future_count = @user.transactions_missing_receipt(from: Receipt::CARD_LOCKING_START_DATE).count
 
       if current_count.in?([5, 7, 9])
-        CardLockingMailer.warning(user: @user).deliver_later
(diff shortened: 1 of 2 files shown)


garyhtou:
Tests are failing due to an unrelated flaky test. [Fixed here](https://github.com/hackclub/hcb/pull/13517)

PR #13519, opened 2026-04-19
[Command Bar] Include sub-organizations

### LINKED ISSUE #13505: [Command Bar] Include subevents in command bar
Request made [here](https://hackclub.slack.com/archives/C068U0JMV19/p1776439943934659)

manuthecoder:
image

Closes #13505

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -23,14 +23,26 @@ class EventsController < ApplicationController
     authorize Event
     respond_to do |format|
       format.json do
-        events =
+        user_events =
           @current_user
           .events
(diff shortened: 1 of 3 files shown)


app/controllers/events_controller.rb
```diff
@@ -23,14 +23,26 @@ def index
     authorize Event
     respond_to do |format|
       format.json do
-        events =
+        user_events =
           @current_user
           .events
```
garyhtou, line 28: See if we can get a list of ALL the events using `@current_user.readable_events`. But still keep the reorder and `parent_slug` behavior

PR #13531, opened 2026-04-21
[Disbursements] Improve performance w/ new combobox component

manuthecoder:
Uses the new combobox component to improve performance on `/disbursements/new`

image


app/controllers/disbursements_controller.rb
```diff
@@ -47,28 +52,50 @@ def new
...
+        disabled_message = "Insufficient balance" if e.balance_available <= 0
+        disabled_message = "HCB transfers disabled" unless policy(e).create_transfer?
+      end
+
+      right = disabled_message || helpers.render_money_short(e.balance_available)
+      attrs = disabled_message ? { data: { disabled_option: "" } } : {}
+      content = helpers.content_tag(:div, class: "flex flex-col justify-between w-full #{disabled_message ? "opacity-50" : ""}", **attrs) do
+        helpers.content_tag(:span, label) + helpers.content_tag(:span, right, class: "muted")
+      end
+      { value: e.id, display: label, content: content }
```
garyhtou: Does value here need to be a `public_id`?

app/controllers/disbursements_controller.rb
```diff
@@ -47,28 +52,50 @@ def new
...
+    end
+
+    render turbo_stream: helpers.async_combobox_options(options)
+  end
+
   def create
-    @source_event = Event.find_by_public_id(disbursement_params[:source_event_id])
+    @source_event = Event.find_by_public_id(disbursement_params[:source_event_id]) ||
+                    Event.find(disbursement_params[:source_event_id])
```
garyhtou: I would prefer to revert this and only support creating by public id

app/controllers/disbursements_controller.rb
```diff
@@ -47,29 +47,45 @@ def new
...
+    skip_authorization
+    q = params[:q].presence
+    sending = params[:sending] == "true"
+    is_admin = admin_signed_in?
+
+    base = if is_admin
+             Event.order(Event::CUSTOM_SORT)
+           elsif !sending && unrestricted_destination_search?
+             manageable = current_user.manageable_events.not_hidden.filter_demo_mode(false)
+             Event.where(id: manageable.select(:id)).or(Event.where(id: Event.indexable.select(:id)))
```
```suggestion
             manageable.or(Event.indexable)
```

app/controllers/disbursements_controller.rb
```diff
@@ -47,29 +47,45 @@ def new
...
-                                  end.to_enum.with_index.sort_by { |e, i| [user_event_ids.index(e.id) || Float::INFINITY, i] }.map(&:first)
-
     authorize @disbursement
     render layout: "transfer"
   end
 
+  def event_search
+    skip_authorization
+    q = params[:q].presence
+    sending = params[:sending] == "true"
```
garyhtou, line 58:
```suggestion
    # Indicates whether we're searching for source or destination organizations
    sending = params[:sending] == "true"
```

app/controllers/disbursements_controller.rb
```diff
@@ -47,29 +47,45 @@ def new
...
+    options = events.map do |e|
+      disabled_message = "Insufficient balance" if sending && !is_admin && e.balance_available <= 0
+
+      right = disabled_message || helpers.render_money_short(e.balance_available)
+      attrs = disabled_message ? { data: { disabled_option: "" } } : {}
+      name_label = is_admin ? "#{e.name} (#{e.id})" : e.name
+      content = helpers.content_tag(:div, class: "flex flex-col justify-between w-full #{disabled_message ? "opacity-50" : ""}", **attrs) do
+        helpers.content_tag(:span, name_label) + helpers.content_tag(:span, right, class: "muted")
+      end
+      { value: e.id.to_s, display: e.name, content: content }
```
garyhtou: Let's try to use the public event IDs — if you're running into an issue/error with the combobox gem when attempting to use the public IDs, then let me know and we can debug this together.

app/controllers/disbursements_controller.rb
```diff
@@ -122,11 +138,16 @@ def create
     end
 
   rescue ArgumentError, ActiveRecord::RecordInvalid => e
+    skip_authorization
```
garyhtou: I see that we're always calling `authorize @disbursement`. Please double check whether `skip_authorization` is necessary here (and below) — but maybe it is necessary because I see that we had it in the old diff for another rescue in this controller action.

app/controllers/disbursements_controller.rb
```diff
@@ -122,11 +138,16 @@ def create
...
     flash[:error] = e.message
     redirect_to new_disbursement_path(source_event_id: @source_event)
-  rescue ActiveRecord::RecordNotFound => e
+  rescue ActiveRecord::StatementInvalid
```
garyhtou: I think this is usually a sign that there's a SQL injection vulnerability. Let's look into what triggers a StatementInvalid error and see if we can prevent it from happening; rather than rescuing it.

app/views/disbursements/new.html.erb
```diff
@@ -2,7 +2,7 @@
 <% @back_url = @source_event ? event_transfers_path(@source_event) : nil %>
 
 <% content_for :sidebar_footer do %>
-  <% if @allowed_destination_events.empty? %>
+  <% if !admin_signed_in? && !current_user.manageable_events.not_hidden.filter_demo_mode(false).exists? %>
```
garyhtou: Let's see if we can re-use the existing logic/array in the controller rather than redefining the list of events available to transfer to.

app/controllers/disbursements_controller.rb
```diff
@@ -47,24 +47,55 @@ def new
...
+             Event.where(id: allowed_destination_event_ids).select(:name, :id, :demo_mode, :can_front_balance, :slug).includes(:plan)
+           else
+             current_user.manageable_events.not_hidden.filter_demo_mode(false)
+           end
+
+    # Apply fuzzy search if query present
+    if q.present?
+      sql = "LOWER(name) ILIKE :name OR LOWER(slug) ILIKE :slug"
+      sql += " OR CAST(id AS TEXT) ILIKE :id" if admin_signed_in?
+      base = base.where(sql, name: "%#{q.downcase}%", slug: "%#{q.downcase}%", id: "%#{q}%")
```
```suggestion
      sql = "name ILIKE :name OR slug ILIKE :slug"
      sql += " OR CAST(id AS TEXT) ILIKE :id" if admin_signed_in?
      base = base.where(sql, name: "%#{q}%", slug: "%#{q}%", id: "%#{q}%")
```

garyhtou (approved):

garyhtou (approved):

PR #5803, opened 2024-03-27
[CI] Check for missing annotations

garyhtou:
## Summary of the problem

it'd be nice to know if a PR is missing annotations

Closes https://github.com/hackclub/hcb/issues/10690

## Describe your changes

add it to CI. we'll see if this works 🤷

PR #13540, opened 2026-04-22
[CI] Add check model annotations

polypixeldev:
## Summary of the problem

Sometimes, PRs that add migrations or ignore columns can make it through without running annotaterb, which means the model annotation is out-of-date and when everyone pulls from main and runs `bin/rails db:migrate`, they'll find an extra file changed.

## Describe your changes

Simple CI check to see if there are changes needed to be made to annotations! The command will exit with code 1 if there's anything that needs to be changed.

**_don't merge yet, testing_**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -380,16 +380,32 @@ jobs:
   annotations:
     name: Check model annotations
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


polypixeldev:
oops... this already exists, just seems to not be working #5803

PR #13537, opened 2026-04-22
[v4 refresh token] Add model annotations

manuthecoder:
Adds the missing model annotations which were not included in the original PR

PR #13541, opened 2026-04-22
[CI] Fix annotate check

polypixeldev:
## Summary of the problem

Annotaterb has been updated and now needs a different command to specify that we're checking model annotations

## Describe your changes

Fix the command


garyhtou (approved):

garyhtou (approved):
https://github.com/hackclub/hcb/pull/13537

PR #13543, opened 2026-04-22
[GSuiteAccount] Allow disconnecting from HCB management

garyhtou:
## Summary of the problem

We want to be able to convert an HCB-managed Google Workspace Account (e.g. `example@event.hackclub.com`) into an unmanaged Google Workspace Account so that we can transfer it to the unmanaged `hackclub.com` domain.

## Describe your changes

This PR adds a `GSuiteAccount#disconnect!` method that deletes the records from HCB while preserving everything on Google Workspace; effectively "unmanaging" this Google Workspace Account.

### DIFF
--- a/app/models/g_suite_account.rb
+++ b/app/models/g_suite_account.rb
@@ -33,6 +33,8 @@ class GSuiteAccount < ApplicationRecord
 
   include Rejectable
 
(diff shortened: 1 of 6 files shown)


manuthecoder (approved):
LGTM!

PR #13550, opened 2026-04-23
[v4][Code Quality] Create 'object_shape' helper to follow new v4 standards

Mohamad-Mortada:
## Summary of the problem
We're not compliant with our own standards 😨 

## Describe your changes
Create the object_shape helper we defined in our v4 standards and apply it to existing routes. Couple things to notice:

1. event partial has custom object name of "organization"
2. user partial doesn't have `created_at` field for privacy reasons

### DIFF
--- a/app/helpers/api/v4/application_helper.rb
+++ b/app/helpers/api/v4/application_helper.rb
@@ -8,6 +8,13 @@ module Api
 
       attr_reader :current_user, :current_token
 
+      def object_shape(json, object, object_name: nil, created_at: true, &block)
+        json.id object.public_id
+        json.object object_name || object.model_name.element
+        block.call
+        json.created_at object.created_at if created_at
+      end
+
       def pagination_metadata(json)
         json.total_count @total_count
         json.has_more @has_more
--- a/app/helpers/api/v4/application_helper.rb
+++ b/app/helpers/api/v4/application_helper.rb
@@ -10,7 +10,7 @@ module Api
 
       def object_shape(json, object, object_name: nil, created_at: true, &block)
         json.id object.public_id
-        json.object object_name || object.model_name.element
+        json.object object_name || object.model_name.singular
         block.call
         json.created_at object.created_at if created_at
       end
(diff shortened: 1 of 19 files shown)


app/helpers/api/v4/application_helper.rb
```diff
@@ -8,6 +8,13 @@ module ApplicationHelper
 
       attr_reader :current_user, :current_token
 
+      def object_shape(json, object, object_name: nil, created_at: true, &block)
+        json.id object.public_id
+        json.object object_name || object.model_name.element
```
garyhtou: `singular` preserves the namespacing (e.g. `reimbursement_report`) while `element` doesn't (e.g. `report`).

```suggestion
        json.object object_name || object.model_name.singular
```

app/views/api/v4/transactions/_transaction.json.jbuilder
```diff
@@ -7,43 +7,44 @@ is_cpt = tx.is_a?(CanonicalPendingTransaction)
...
-  json.id tag.public_id
-  json.label tag.label
-  json.color tag.color
-  json.emoji tag.emoji
-end
-json.code hcb_code.hcb_i1
-json.missing_receipt hcb_code.missing_receipt?(@event)
-json.lost_receipt hcb_code.no_or_lost_receipt?
-json.appearance hcb_code.incoming_disbursement.special_appearance_name if hcb_code.incoming_disbursement&.special_appearance?
+object_shape(json, hcb_code, object_name: "transaction") do
```
garyhtou:
```suggestion
object_shape(json, hcb_code, object_name: "transaction", created_at: false) do
```

garyhtou: `created_at` for HcbCode isn't necessarily accurate. I'd trust the date before. Because we weren't exposing `created_at` for HcbCode before, we'll maintain that behavior.

In the future, we should expose `created_at` as a timestamp (once we have accurate timestamps for all transactions)

garyhtou (approved):

PR #13555, opened 2026-04-23
[v4][Breaking Change] Add admin scopes

Mohamad-Mortada:
## Summary of the problem
We want admins not to be shivering in their boots whenever they OAuth into an HCB application, as their token can be used for virtually anything. We should force OAuth apps to have a scope that unlocks admin capabailities. 

## Describe your changes
Adds admin scopes (admin:read, admin:write) and has a helper function of `can_admin(:read/:write)` which handles if scope is enabled and if theyre an admin. theres also a modified `require_admin_scope` function which locks entire routes to just admins


app/controllers/api/v4/application_controller.rb
```diff
@@ -40,10 +40,22 @@ def authenticate!
         @current_user = current_token&.user
       end
 
-      def require_admin!
-        unless current_user&.admin?
-          render json: { error: "invalid_auth" }, status: :unauthorized
-        end
+      ADMIN_SCOPES = {
+        read: ->(user){ user&.auditor? },
+        write: ->(user){ user&.admin? }
```
garyhtou: Make these `override_pretend`

app/helpers/api/v4/application_helper.rb
```diff
@@ -79,6 +79,20 @@ def expand_pii(override_if: false)
...
+      def can_admin?(level)
+        return false unless current_token&.scopes&.include?("admin:#{level}")
+
+        case level.to_sym
+        when :read  then current_user&.auditor?
+        when :write then current_user&.admin?
+        else false
+        end
+      end
+
```
garyhtou: See if we can dedup this method.

manuthecoder (approved):
LGTM!

garyhtou (commented):
This is good to go, but we'll now need to figure out how to roll this out

garyhtou (approved):

garyhtou (approved):

PR #13556, opened 2026-04-23
[Reimbursement] Block changing Event when has Card Grant

garyhtou:
## Summary of the problem

When a reimbursement report is backed by a card grant, we don't want to allow moving that report to another organization because those funds were intended to be spent with the original organization.

## Describe your changes

Block changing organization (even for admins) when the report is associated with a card grant


manuthecoder (approved):
lgtm!

Luke-Oldenburg (approved):
lgtm

PR #13557, opened 2026-04-23
[CI] Speed up yarn install with caches

garyhtou:
## Summary of the problem

CI was reinstalling JS packages from scratch on every run. `setup-node`'s `cache: yarn` (and the equivalent manual block) only stores the download tarballs — every job still had to extract and link them into `node_modules`. ESLint and Prettier had no JS cache at all and cold-installed (~20s) every run.

## Describe your changes

Cache `node_modules` directly in each yarn-using job:

- Add `cache: yarn` to `setup-node` (replacing the manual `yarn cache dir` + `actions/cache` block where present).
- Add `actions/cache` on `node_modules` keyed on `runner.os` + resolved Node version (from `steps.setup-node.outputs.node-version`) + `yarn.lock` hash. OS and Node cover native-binary ABI; `yarn.lock` covers the dependency set.
- Use `yarn install --frozen-lockfile --prefer-offline` — fast consistency check on hit, full install on miss.

### Benchmark (measured)

Compared latest main run vs this branch's warm-cache run:

| Job | main | branch (warm) | Δ |
|---|---|---|---|
| ESLint | 30s | **20s** | -33% |
| Prettier | 32s | **21s** | -34% |
| Zeitwerk | 1m15s | **56s** | -25% |
| RSpec shard (each) | ~25s yarn-related | **~5s yarn-related** | -80% |

`yarn install` itself drops from 8–23s to 0–1s across the board. Critical-path CI wall time drops ~15–20s (set by RSpec, whose own variance is similar), but the jobs previously dominated by yarn install (ESLint, Prettier, Zeitwerk) see the clearer 25–35% drop.

On a first PR run (cold cache), a few early-starting jobs race each other and cold-install; once the first finishes and saves, everything after hits. Subsequent runs with unchanged `yarn.lock` hit everywhere.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,7 @@ jobs:
         with:
           bundler-cache: true
       - name: Set up Node
+        id: setup-node
         uses: actions/setup-node@v6
         with:
           node-version-file: package.json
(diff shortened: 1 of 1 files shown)


garyhtou (commented):
Imma come back with another PR to dry up this file. it's getting messy 😅

PR #13567, opened 2026-04-25
[v4] Save donation message

karmanyaahm:
## Summary of the problem
Currently, HCB Mobile's Tap to pay donation feature cannot provide messages. As I use it for fundraising, it would be helpful to keep track of what each transaction is selling. I am building a web ui to go along with this.

I am making a PR to HCB Mobile to use this functionality as well.

## Describe your changes
Save message on v4 donation API. I made a test too but TBH I am not super familiar with rails so I don't know if there's issues in it.

### DIFF
--- a/app/controllers/api/v4/donations_controller.rb
+++ b/app/controllers/api/v4/donations_controller.rb
@@ -22,6 +22,7 @@ module Api
                                    in_person: true,
                                    name: params[:name].presence,
                                    email: params[:email].presence,
+                                   message: params[:message].presence,
                                    anonymous: !!params[:anonymous],
                                    tax_deductible: params[:tax_deductible].nil? || params[:tax_deductible],
                                    fee_covered: !!params[:fee_covered] && @event.config.cover_donation_fees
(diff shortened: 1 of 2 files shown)


Mohamad-Mortada (changes requested):
Hey!

Thanks for the PR! I'm ok with implementing this :)

However, for the test, can you make it a bit broader? For example, during card grant creation testing, we verify that each returned value is correct. This will do the job you want, and it also verifies other fields. 

https://github.com/hackclub/hcb/blob/581ca58413d8709e60560e810166713f3307530e/spec/controllers/api/v4/card_grants_controller_spec.rb#L61-L116

karmanyaahm:
Hey @Mohamad-Mortada, I updated it!

karmanyaahm:
Not sure if you would like include or eq because there's a lot of nil fields in the v4 api (like all the utm referrer stuff, payment info at this point in the process)

Mohamad-Mortada (approved):
LGTM

manuthecoder (approved):
lgtm!!

PR #13569, opened 2026-04-27
[My Receipts] Show in nav if user has any reimbursement reports

manuthecoder:
Adds reimbursement reports to the existing if check.


app/helpers/users_helper.rb
```diff
@@ -41,7 +41,7 @@ def users_nav(selected: nil)
     }
 
 
-    if current_user&.events&.any? || current_user&.stripe_cards&.any?
+    if current_user&.events&.any? || current_user&.stripe_cards&.any? || current_user&.card_grants&.any? || current_user&.reimbursement_reports&.any?
```
Luke-Oldenburg:
```suggestion
    if current_user&.events&.any? || current_user&.stripe_cards&.any? || current_user&.reimbursement_reports&.any?
```

Luke-Oldenburg (approved):
lgtm

PR #13573, opened 2026-04-29
[Policy][Code Quality] Consolidate auditor? and admin? methods

Mohamad-Mortada:
## Summary of the problem
It would be cleaner to standardize use of `admin?` and `auditor?` methods without having to define them in every policy file. 

## Describe your changes
Since they're widely used we should define them in the parent policy then we can reference them through inheritance

### DIFF
--- a/app/policies/comment_policy.rb
+++ b/app/policies/comment_policy.rb
@@ -3,7 +3,7 @@
 class CommentPolicy < ApplicationPolicy
   class Scope < ApplicationPolicy::Scope
     def resolve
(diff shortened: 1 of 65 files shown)


app/policies/comment_policy.rb
```diff
@@ -3,7 +3,7 @@
 class CommentPolicy < ApplicationPolicy
```
manuthecoder, line 3: add admin? / auditor? to ApplicationPolicy::Scope so Scope subclasses can use the same helpers

PR #13579, opened 2026-04-29
[FIRST] Join request

Luke-Oldenburg:

### DIFF
--- a/app/controllers/users/first_controller.rb
+++ b/app/controllers/users/first_controller.rb
@@ -17,10 +17,25 @@ module Users
     end
 
     def request_org_invite
-      # validation
-      event = Event.find(params[[:event_id]])
-      opil = event.organizer_position_invite_links.create!({ creator: event.users.first, expires_in: 1.minute.from_now }) # this expires it near immediately
-      OrganizerPositionInvite::Request.create!(requester: current_user, link: opil)
+      user = current_user(allow_unverified: true)
+      event = Event::Affiliation.matching_first_event_for(user)
+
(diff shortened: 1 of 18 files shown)


app/controllers/users/first_controller.rb
```diff
@@ -16,6 +16,28 @@ def index
       @airpods_raffle = Raffle.find_by(user: current_user(allow_unverified: true), program: "first-worlds-2026-airpods")
     end
 
+    def request_org_invite
+      user = current_user(allow_unverified: true)
+      event = Event::Affiliation.matching_first_event_for(user)
+
+      unless Event::Affiliation.eligible_to_request_invite?(user, event)
+        redirect_to first_index_path, flash: { error: "You're not eligible to request to join this organization." } and return
```
Luke-Oldenburg, line 28: what does `and return` do?

app/controllers/users/first_controller.rb
```diff
@@ -16,6 +16,28 @@ def index
...
+      end
+
+      if event.organizer_position_invite_requests.pending.exists?(requester_id: user.id)
+        redirect_to first_index_path, flash: { warning: "You already have a pending request." } and return
+      end
+
+      ActiveRecord::Base.transaction do
```
Luke-Oldenburg, line 35: smart

app/controllers/users/first_controller.rb
garyhtou: It returns from the method, preventing the rest of the code (below) in this method from running.

PR #13589, opened 2026-04-30
[Webhook] Listen for stripe force captures

Sarvesh-Mk:
## Summary of the problem

closes #13439

## Describe your changes

added code to handle_issuing_transaction_created to handle force captures 
to my understanding, force captures skip:
  1. issuing_authorization.request 
  2. issuing_authorization.created 
  3. issuing_authorization.updated 

and lands on handle_issuing_transaction_created so I added code to create the ledger entry from scratch


app/controllers/stripe_controller.rb
```diff
@@ -75,8 +75,10 @@ def handle_issuing_authorization_updated(event)
 
   def handle_issuing_transaction_created(event)
     tx = event[:data][:object]
-    amount = tx[:amount]
-    return unless amount < 0
+
+    if tx[:authorization].nil?
+      ::StripeAuthorization::CreateFromWebhookJob.perform_later(tx[:id])
```
manuthecoder: `tx[:id]` is a transaction ID (`ipi_...`), but this job internally calls `StripeService::Issuing::Authorization.retrieve(id)`

it expects an authorization ID (`iauth_...`). stripe won't find anything and the service silently returns early

app/controllers/stripe_controller.rb
```diff
@@ -75,8 +75,10 @@ def handle_issuing_authorization_updated(event)
...
+    end
```
manuthecoder, line 84: `return unless amount < 0` filtered out credits/refunds (positive amounts). removing it means the nil-authorization branch can now fire for positive-amount force captures too

app/controllers/stripe_controller.rb
manuthecoder: force captures are already settled, not pending so routing through CreateFromWebhook (the pending transaction pipeline) is probably wrong

a separate job or service that imports directly by `ipi_...ID` makes more sense

manuthecoder (changes requested):

Luke-Oldenburg (commented):
To my knowledge, force captures do not have an authorization.

PR #13595, opened 2026-04-30
[FIRST] Add earnable extra tickets to MacBook raffle for referring more people

polypixeldev:
## Summary of the problem

We want to encourage users to share HCB with their teammates and friends at FIRST Worlds! Let's give out some extra raffle tickets on a tiered system when users send their unique raffle link to others who fill out the FIRST form.

image

### DIFF
--- a/app/controllers/users/first_controller.rb
+++ b/app/controllers/users/first_controller.rb
@@ -105,26 +105,26 @@ module Users
       program = "first-worlds-2026-macbook" if ["student_leader", "student_member"].include?(user_params.dig(:affiliations_attributes, "0", "role"))
 
       unless User.where(email: user_params[:email]).exists?
-        user_referral = Raffle.find_by_hashid(user_ref_params[:user_referral]) if user_ref_params[:user_referral].present?
-        if user_referral.nil?
-          flash[:error] = "We couldn't find that referral link!"
+        user_referral = nil
+        if user_ref_params[:user_referral].present?
+          user_referral = Raffle.find_by_hashid(user_ref_params[:user_referral])
 
-          redirect_to welcome_first_index_path and return
-        end
-        # case that should never happen because users should not be able to get a link if program or user is not eligible for referral link
-        # however still adding here because it could eventually be possible that a user becomes ineligible or a program becomes ineligible
-        if !user_referral.program_allows_referrals? || !user_referral.user_allows_referrals?
-          flash[:error] = "We couldn't find that referral link!"
+          # the referral not being eligible should never happen because users should not be able to get a link if program or user is not eligible for referral link
+          # however still adding here because it could eventually be possible that a user becomes ineligible or a program becomes ineligible
+          if user_referral.nil? || !user_referral.program_allows_referrals? || !user_referral.user_allows_referrals?
+            flash[:error] = "We couldn't find that referral link!"
 
-          redirect_to welcome_first_index_path and return
+            redirect_to welcome_first_index_path and return
+          end
         end
+
         @user = User.new(user_params)
         @user.creation_method = :first_robotics_form
         @user.save!
 
         if program.present?
           raf = Raffle.find_or_create_by!(user: @user, program:)
-          raf.update!(referring_raffle: user_referral)
+          raf.update!(referring_raffle: user_referral) if user_referral.present?
         end
 
         create_session(user: @user, verified: false)
--- a/app/controllers/users/first_controller.rb
+++ b/app/controllers/users/first_controller.rb
@@ -85,7 +85,9 @@ module Users
       return redirect_to first_index_path if signed_in?(allow_unverified: true)
 
       @referral_link_slug = Referral::Link.find_by(slug: params[:referral])&.slug if params[:referral].present?
+      @user_referral = params[:referred_by] if params[:referred_by].present?
       @user = User.new(affiliations: [Event::Affiliation.new])
+
     end
 
     def verify_email
@@ -103,11 +105,27 @@ module Users
       program = "first-worlds-2026-macbook" if ["student_leader", "student_member"].include?(user_params.dig(:affiliations_attributes, "0", "role"))
 
       unless User.where(email: user_params[:email]).exists?
+        user_referral = Raffle.find_by_hashid(user_ref_params[:user_referral]) if user_ref_params[:user_referral].present?
+        if user_referral.nil?
+          flash[:error] = "We couldn't find that referral link!"
+
+          redirect_to welcome_first_index_path and return
+        end
+        # case that should never happen because users should not be able to get a link if program or user is not eligible for referral link
+        # however still adding here because it could eventually be possible that a user becomes ineligible or a program becomes ineligible
+        if !user_referral.program_allows_referrals? || !user_referral.user_allows_referrals?
+          flash[:error] = "We couldn't find that referral link!"
+
+          redirect_to welcome_first_index_path and return
+        end
         @user = User.new(user_params)
         @user.creation_method = :first_robotics_form
         @user.save!
 
-        Raffle.find_or_create_by!(user: @user, program:) if program.present?
+        if program.present?
+          raf = Raffle.find_or_create_by!(user: @user, program:)
+          raf.update!(referring_raffle: user_referral)
+        end
 
         create_session(user: @user, verified: false)
 
@@ -130,6 +148,10 @@ module Users
 
     def user_params
       params.require(:user).permit(:email, :full_name, affiliations_attributes: [:league, :team_number, :name, :team_name, :role])
+    end
+
+    def user_ref_params
+      params.permit(:user_referral)
     end
 
     def load_team_community
--- a/app/controllers/users/first_controller.rb
+++ b/app/controllers/users/first_controller.rb
@@ -144,6 +144,30 @@ module Users
       render :new, status: :unprocessable_entity
     end
 
+    def macbook_qr_code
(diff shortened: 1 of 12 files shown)


Mohamad-Mortada (approved):

app/controllers/users/first_controller.rb
```diff
@@ -126,12 +144,40 @@ def create
       render :new, status: :unprocessable_entity
     end
 
+    def macbook_qr_code
```
garyhtou, line 147: **Security/bug: missing auth guard.** This controller has `skip_before_action :signed_in_user` at the top, so this action is reachable unauthenticated. `current_user` is then `nil` and `current_user.raffles.find_by(...)` raises `NoMethodError`, returning a 500.

Also note: a signed-in but **unverified** user can hit this URL directly (the modal markup is rendered for them; only the button is hidden). For unverified users `raffle.referral_link` returns `nil`, and `RQRCode::QRCode.new(nil)` will encode the string `""` — so the QR points to `welcome_first_index_url(referred_by: nil)`, leaking a useless QR.

Suggested fix:
```ruby
def macbook_qr_code
  user = current_user(allow_unverified: true)
  return head :unauthorized unless user&.verified?

  raffle = user.raffles.find_by(program: "first-worlds-2026-macbook")
  return head :not_found if raffle.nil? || raffle.referral_link.blank?
  # ...
end
```

---
_Generated by [Claude Code](https://claude.ai/code)_

app/controllers/users/first_controller.rb
```diff
@@ -103,11 +105,27 @@ def create
...
+          end
+        end
+
         @user = User.new(user_params)
         @user.creation_method = :first_robotics_form
         @user.save!
 
-        Raffle.find_or_create_by!(user: @user, program:) if program.present?
+        if program.present?
+          raf = Raffle.find_or_create_by!(user: @user, program:)
```
garyhtou, line 126: **Security: trivial referral fraud vectors.**

A few gaps here that together make the tier system easy to game:

1. **No self-referral check.** Nothing prevents `user_referral.user_id == @user.id`-style abuse via a second account. A single person creates throwaway accounts (different emails, same person) and racks up tickets indefinitely.
2. **Referrer is credited before the referred user verifies their email.** `after_create_commit` on Raffle fires immediately on `Raffle.find_or_create_by!`, while the new account is still unverified (`verified: false` is passed to `create_session`). Disposable / never-confirmed emails count just as much as real ones.
3. **No rate-limiting on `create`.** Combined with the above, a script can mint dozens of fake referred raffles in seconds.

The disclaimer "We may disqualify extra tickets if referrals are fraudulent" is reactive only. At minimum, gate the `referring_raffle` association on the referred user actually verifying their email (e.g. set the FK in a hook on `User#verified` becoming true, not at `Raffle.find_or_create_by!` time), and consider:
- Rejecting `user_referral.user_id == @user.id` (defense in depth, in case of session re-use)
- Throttling `:create` per IP via Rack::Attack
- Capping `extra_tickets` per raffle

---
_Generated by [Claude Code](https://claude.ai/code)_

app/controllers/users/first_controller.rb
```diff
@@ -103,11 +105,27 @@ def create
...
+        user_referral = nil
+        if user_ref_params[:user_referral].present?
+          user_referral = Raffle.find_by_hashid(user_ref_params[:user_referral])
+
+          # the referral not being eligible should never happen because users should not be able to get a link if program or user is not eligible for referral link
+          # however still adding here because it could eventually be possible that a user becomes ineligible or a program becomes ineligible
+          if user_referral.nil? || !user_referral.program_allows_referrals? || !user_referral.user_allows_referrals?
+            flash[:error] = "We couldn't find that referral link!"
+
+            redirect_to welcome_first_index_path and return
```
garyhtou, line 117: **UX: failing referral discards the entire signup form.** A new user fills out the (long) form, hits submit, and on a stale/broken referral hashid we `redirect_to welcome_first_index_path` — they lose their name, email, league, team number, role, and their session context.

This is also reachable in non-malicious cases (someone tweeting a link that becomes invalid, hashids breaking across env, etc.). Render `:new` with the error and the existing `@user` so the form can be resubmitted, or just silently drop `user_referral` and proceed (since the comment above admits it shouldn't happen). The comment justifies preserving the behavior because users can become ineligible — but punishing the *new* user for the *referrer's* state change is the wrong call.

---
_Generated by [Claude Code](https://claude.ai/code)_

app/views/users/first/index.html.erb
```diff
@@ -70,6 +70,39 @@
...
+
+    <p>
+      Share your <%= @macbook_raffle.referral_link.present? ? link_to("unique raffle link", @macbook_raffle.referral_link) : "unique raffle link" %> with your teammates and friends in FIRST robotics,
+      and you'll earn extra raffle tickets for the MacBook Neo!
+      You currently have <%= pluralize(@macbook_raffle.tickets, "ticket") %> for the MacBook Neo raffle.
+      <em>We may disqualify extra tickets if referrals are fraudulent.</em>
+    </p>
+
+    <% referrals_needed = @macbook_raffle.referrals_needed_for_next_ticket - @macbook_raffle.tier_progress %>
+    <p class="mb-0 text-center">Refer <%= referrals_needed %> more <%= "person".pluralize(referrals_needed) %> for your next ticket!</p>
```
garyhtou, line 85: **UX: progress bar formula has an off-by-one.**

```erb
style="width: <%= [1 + (@macbook_raffle.tier_progress.to_f / @macbook_raffle.referrals_needed_for_next_ticket) * 100, 100].min %>%;"
```

The leading `1 +` looks like a debugging leftover. With 0 referrals it shows the bar at **1%** instead of 0%; with 4/5 progress it shows **81%** instead of 80%. Should be:

```erb
style="width: <%= [(@macbook_raffle.tier_progress.to_f / @macbook_raffle.referrals_needed_for_next_ticket) * 100, 100].min %>%;"
```

(Also worth memoizing — `tier_progress`, `referrals_needed_for_next_ticket`, `extra_tickets`, and `tickets` each independently issue a `COUNT(*)` query against `referred_raffle_entries`, so this single page is doing 4–5 redundant counts.)

---
_Generated by [Claude Code](https://claude.ai/code)_

app/models/raffle.rb
```diff
@@ -4,32 +4,48 @@
...
+  has_many :referred_raffle_entries, class_name: "Raffle", foreign_key: "referring_raffle_id", inverse_of: :referring_raffle
+  belongs_to :referring_raffle, class_name: "Raffle", optional: true
   validates :program, presence: true
   validates :ticket_number, uniqueness: true, allow_nil: true
 
+  after_create_commit do
+    if referring_raffle.present? && referring_raffle.tier_progress == 0
+      RaffleMailer.with(raffle: referring_raffle).extra_ticket_earned.deliver_later
+    end
+  end
```
garyhtou, line 47: **Mail-bombing risk + race + confused-state.**

Three concerns with this `after_create_commit`:

1. **The "you earned a ticket!" email is sent before the referred account is verified.** The new user's raffle is created with `verified: false` in `create_session`. The referrer gets a celebratory email even though the referral may never materialize, which both encourages disposable-email farming and means the referrer can be tricked into believing their tally is higher than it actually is.

2. **`tier_progress == 0` triggers the email, but `tier_progress` is also `0` for *every* `extra_tickets` increment** — that part is intentional. However, if a count ever goes negative-modulo (e.g. a referred raffle is later destroyed because of `dependent: :nullify`/cascade), no compensating logic exists. Less hypothetical: `referred_raffle_entries.count` queries the DB inside the after-commit hook. Two near-simultaneous referral signups can both observe the same post-commit count after their respective transactions and trigger duplicate "you earned a ticket!" emails for the same tier.

3. **`@raffle.referral_link` in the mail body returns `nil` if the user becomes unverified between earning and email delivery** (the mail is `deliver_later`). `link_to "your unique raffle link", nil` renders a broken link.

Recommendation: defer crediting — and emailing — until the referred user verifies their email. A `User#after_update_commit if saved_change_to_verified?` hook that walks `raffles` and emails the `referring_raffle.user` is closer to correct, and naturally fixes the disposable-email farming vector.

---
_Generated by [Claude Code](https://claude.ai/code)_

garyhtou (changes requested):
Cool feature, but a few critical issues to address before shipping — particularly around the referral-fraud surface and the unauthenticated QR endpoint. Inline comments cover specifics; high-level summary:

**Critical (security):**
- `macbook_qr_code` is reachable unauthenticated (`skip_before_action :signed_in_user` is set controller-wide); `current_user.raffles` will `NoMethodError` on `nil`.
- Referrer is credited **immediately on signup**, before the referred user's email is verified — disposable-email farming is essentially free. The "we may disqualify" disclaimer is reactive only.
- No self-referral / same-person check. Combined with no rate limiting on `:create`, the tier system is trivially gameable.

**UX:**
- Progress bar width formula has a stray `1 +` (always shows ≥1%, off-by-one at every other progress level).
- An invalid/expired referral hashid redirects the new user to the welcome page and discards the entire signup form they just filled out.
- `@raffle.referral_link` can be `nil` (unverified user) but is passed to `RQRCode::QRCode.new` and `link_to` in the mailer without a guard.

**Minor:**
- `tickets`, `extra_tickets`, `tier_progress`, and `referrals_needed_for_next_ticket` each issue an independent `COUNT(*)` — memoize.
- Stray `data-controller="email-compose"` on the QR code modal.
- Stray blank line in `def new`.

---
_Generated by [Claude Code](https://claude.ai/code)_

garyhtou (approved):

PR #13603, opened 2026-05-01
[Raffles] Make pending raffles yellow

polypixeldev:
## Summary of the problem

It should be more obvious to the user that a raffle entry isn't confirmed yet - let's make pending raffles yellow.

image


manuthecoder (approved):
LGTM!

garyhtou (approved):

PR #13604, opened 2026-05-01
[Referrals] Update attribution users after user creation

polypixeldev:
## Summary of the problem

We currently don't update the user on a referral attribution until the user verifies their email. We should do this as soon as the user is created instead.

This PR adds a maintenance task to backfill existing attributions.


manuthecoder (approved):

garyhtou:
Handled by https://github.com/hackclub/hcb/pull/13609

PR #13622, opened 2026-05-03
[FIRST] Remove raffles from landing page

YodaLightsabr:
## Summary of the problem

These raffles are no longer active.

## Describe your changes

This PR removes references to the raffles. I decided to keep the "your teammates are interested, too" card on the homepage with the raffle related elements removed.


Luke-Oldenburg (approved):
lgtm

PR #13634, opened 2026-05-05
[Check Deposit] Disallow creation unless feature enabled

garyhtou:
welp, they should only be able to deposit when we want them to deposit...


Luke-Oldenburg (approved):
lgtm

PR #13642, opened 2026-05-06
[v4][Code Quality] Introduces new expand_association helper

Mohamad-Mortada:
## Summary of the problem
Our standards was pointing towards implementing something like this beforehand. Now there's a handy helper function to streamline the development process :)

## Describe your changes
Introduces `expand_association` which conditionally checks if we have an expanded param and if we do it shows the expanded object, otherwise it will return the object's id as `xxxx_id`

### DIFF
--- a/app/helpers/api/v4/application_helper.rb
+++ b/app/helpers/api/v4/application_helper.rb
@@ -58,6 +58,19 @@ module Api
         @expand.include?(key)
       end
 
+      # Returns a related object as either expanded or as an "_id" reference
+      def expand_association(json, key, record, partial:, as:)
+        if expand?(key)
+          if record.present?
+            json.public_send(key, record, partial: partial, as: as)
+          else
+            json.public_send(key, nil)
+          end
+        else
+          json.public_send(:"#{key}_id", record&.public_id)
+        end
+      end
+
       def expand(*keys)
         before = @expand
         @expand = @expand.dup + keys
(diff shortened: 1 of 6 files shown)


app/helpers/api/v4/application_helper.rb
```diff
@@ -58,6 +58,19 @@ def expand?(key)
         @expand.include?(key)
       end
 
+      # Returns a related object as either expanded or as an "_id" reference
+      def expand_association(json, key, record, partial:, as:)
+        if expand?(key)
+          if record.present?
+            json.public_send(key, record, partial: partial, as: as)
```
garyhtou: You can likely use `set!`

https://github.com/rails/jbuilder

dev-docs/v4-api/standards.md
```diff
@@ -307,9 +307,38 @@ GET /api/v4/cards/crd_x9f3k?expand=user,organization
...
+expand_association(json, :organization, ach_transfer.event,
+                   partial: "api/v4/events/event", as: :event)
+
+expand_association(json, :sender, ach_transfer.creator,
+                   partial: "api/v4/users/user", as: :user)
+```
+
+- When `:organization` is **not** in `expand`, renders `"organization_id": "org_h1izp"`.
+- When `:organization` **is** in `expand`, renders the full object under `"organization": { ... }`.
+- If the record is `nil`, renders `null` in both cases.
```
garyhtou, line 324:
```suggestion
- If the record is `nil`, renders `null` in both cases.
- Associations must have a `public_id` method.
```

garyhtou (approved):

PR #13653, opened 2026-05-07
[IncreaseCheck] Remove reissuing

### LINKED ISSUE #13650: [IncreaseCheck] Strip out reissue code

Luke-Oldenburg:
# Now that stopping is supported, the new flow is:
- org manager sends check
- hcb ops approves check
- org manager stops check (they can now do this even after hcb approval and it is sent on column)
- org manager sends new check
- hcb ops approves new check

This way, hcb ops only has to deal with the existing flow of approving checks rather than a special flow just to reissue.

# What if it's a reimbursement?
- reimbursee reachs out to hcb ops
- hcb ops stops check
- reimbursement report now shows as failed
- hcb ops clicks edit payout info and then presses save (no need to change any data)
- reimbursement service sends new check

Closes #13650


YodaLightsabr (approved):
LGTM

PR #13663, opened 2026-05-07
[Contracts] Add contract voided page

### LINKED ISSUE #13506: [Contract] When voided and signee clicks link, they get a 403
When a contract is voided (by an admin), and the signee clicks the link to sign the contract (which is sent via email), they get a "You are not authorized to perform this action".

I think that's a bit confusing. We should probably show a page saying "This contract was voided" and maybe show a list of contracts below that are waiting for them to sign.

polypixeldev:
## Summary of the problem

The "you are not authorized to perform this action" error is a little confusing when you visit a voided contract signing page.

Closes #13506

## Describe your changes

Renders a different voided page when visiting the signing page for a voided contract, which includes other contracts that need to be signed, if any.

Refactors the contract row into a partial so we can show it on homepage and the new voided page

image

image

### DIFF
--- a/app/views/contract/parties/voided.html.erb
+++ b/app/views/contract/parties/voided.html.erb
@@ -17,5 +17,3 @@
 <% else %>
   <%= link_to "Return to homepage", root_path, class: "btn mt-2" %>
 <% end %>
-
-
--- a/app/views/contract/parties/voided.html.erb
+++ b/app/views/contract/parties/voided.html.erb
@@ -6,7 +6,7 @@
   You no longer need to sign this agreement.
 
   <% unless @contracts.empty? %>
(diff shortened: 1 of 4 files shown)


app/views/contract/parties/voided.html.erb
```diff
@@ -0,0 +1,19 @@
+<% title "Voided agreement" %>
+
+<h1>This agreement has been voided</h1>
+
+<p>
+  You no longer need to sign this agreement.
+
+  <% unless @contracts.empty? %>
+    However, you have <%= @contracts.size %> other <%= "agreement".pluralize(@contracts.size) %> to sign:
```
garyhtou:
```suggestion
    However, you have <%= pluralize(@contracts.size, "other agreements") %> to sign:
```

garyhtou (approved):

PR #13665, opened 2026-05-07
Remove Event#demo_mode_request_meeting_at from database

polypixeldev:
## Summary of the problem

#13480 ignored this column - we can now safely remove it from the database.

## Describe your changes

Adds migration to remove the column


garyhtou (approved):
We'll do the removal of `ignored_columns` to a separate PR.

PR #13678, opened 2026-05-08
[Reimbursements] Fix date filtering

manuthecoder:
Fixes date filtering. Setting `key` to `created_at*` this usage is outdated and not valid anymore


Luke-Oldenburg (approved):
lgtm

PR #13683, opened 2026-05-08
[Filter] Reset page number when applying filter

Sarvesh-Mk:
## Summary of the problem

Page number persists through filter changes.

## Describe your changes

Added a fix to reset page when filter is added


Luke-Oldenburg (approved):
lgtm

PR #13688, opened 2026-05-09
[v4] use public ID for donations

Mohamad-Mortada:
## Summary of the problem
We should be using public ID instead of private

## Describe your changes
uses public ID


manuthecoder (approved):

Luke-Oldenburg (approved):
lgtm

PR #13692, opened 2026-05-10
Improve requesting onboarding calls

polypixeldev:
## Summary of the problem

#13447 added a card and mailer link for users to be able to request onboarding calls directly from HCB, but there were a few improvements left to be made, which this PR does:
- Fixes the redirect to the scheduling link by rendering either the POST to the email route or the external link directly in the view. While this does mean that we are sending a request to Airtable each time a signee views the home page when the message is visible, I prefer this to needing to use JavaScript to redirect the user since browsers can block redirects to an external site that originate from a POST request.
- There are now 2 emails sent when requesting an onboarding call via email: 1 to the user and a different one to ops. This is so that we can communicate different information to them without worrying about information spilling over to the other.
- Instead of checking whether an event has _any_ transactions on the home page to see if we should automatically hide the onboarding message, the `hide_onboarding_message` config value is now the source of truth and is updated daily in a job.
- Subevents will no longer see the onboarding message, since they almost always have a separate creation flow where the organizer already has a separate HQ/HCB contact
- Signees can now hide the onboarding message themselves early if they don't want to see it.


app/jobs/event/hide_onboarding_messages_job.rb
```diff
@@ -0,0 +1,17 @@
...
+
+class Event
+  class HideOnboardingMessagesJob < ApplicationJob
+    queue_as :low
+
+    def perform
+      events = Event.joins(:config).where(config: { hide_onboarding_message: false })
+
+      events.find_each do |event|
+        event.config.update!(hide_onboarding_message: event.parent.present? || event.hcb_codes.size >= 5)
```
manuthecoder:
```suggestion
		next if event.config.hide_onboarding_message  # skip already-hidden events
        event.config.update!(hide_onboarding_message: event.parent.present? || event.hcb_codes.size >= 5)
```

HideOnboardingMessagesJob can un-hide a previously hidden message because it's based purely on whether an org has 5 or more txns

manuthecoder: `event.hcb_codes.size`: n+1

app/models/event.rb
```diff
@@ -939,6 +939,13 @@ def to_combobox_display
     "#{name} (#{id})"
   end
 
+  def onboarding_scheduling_link
+    if point_of_contact.present?
+      onboarder_record = OnboardersTable.all(filter: "{HCB ID} = #{point_of_contact.id}").first
```
manuthecoder: worth noting that this calls out to airtable on every show page render for any event where the banner is visible, we might want to cache this or handle errors in a `rescue`

app/models/event.rb
```diff
@@ -939,6 +939,13 @@ def to_combobox_display
...
+      return onboarder_record&.[]("Scheduling Link")
+    end
```
manuthecoder:
```suggestion
  return unless point_of_contact.present?
  OnboardersTable.all(filter: "{HCB ID} = #{point_of_contact.id}").first&.[]("Scheduling Link")
```

improves readability

app/services/event_service/create.rb
```diff
@@ -52,6 +52,10 @@ def run
             event.event_tags << ::EventTag.find_or_create_by!(name: tag)
           end
 
+        # Subevents shouldn't see an onboarding message
+        if event.parent.present?
```
manuthecoder, line 56: use `parent_id` because `parent` would trigger another db query and we're not doing more than just checking if a parent event exists

app/mailers/event_mailer.rb
```diff
@@ -50,12 +50,18 @@ def negative_balance
...
+         subject: "We've received your request for an onboarding call for #{@event.name}"
+  end
+
+  def ops_call_requested
     @user = params[:user]
 
     mail to: OPERATIONS_EMAIL,
-         subject: "#{@user.name} requesting an onboarding call for #{@event.name} #{"with #{@event.point_of_contact.name}" if @event.point_of_contact.present?}",
-         reply_to: @user.email_address_with_name
+         subject: "#{@user.name} is requesting an onboarding call for #{@event.name} #{"with #{@event.point_of_contact.name}" if @event.point_of_contact.present?}"
```
manuthecoder, line 64: missing `reply_to` from original code

manuthecoder (changes requested):
in the right direction but some parts need a little change :)

app/jobs/event/hide_onboarding_messages_job.rb
polypixeldev: The where filter on line 8 finds events where `hide_onboarding_message` is false, so anything already hidden will not be part of the loop here

app/mailers/event_mailer.rb
polypixeldev: This is intentional - ops will send a new email to the user instead of replying to this one as to not confuse the user with any information we include in the email to ops.

app/models/event.rb
```diff
@@ -939,6 +939,18 @@ def to_combobox_display
...
+    return unless point_of_contact.present?
+
+    Rails.cache.fetch("scheduling_link_#{point_of_contact.id}", expires_in: 5.minutes) do
+      begin
+        return OnboardersTable.all(filter: "{HCB ID} = #{point_of_contact.id}").first&.[]("Scheduling Link")
+      rescue
+        return nil
+      end
+    end
+  end
```
manuthecoder, line 952:
```suggestion
  def onboarding_scheduling_link
    return unless point_of_contact.present?
  
    Rails.cache.fetch("scheduling_link_#{point_of_contact.id}", expires_in: 5.minutes) do
      OnboardersTable.all(filter: "{HCB ID} = #{point_of_contact.id}").first&.[]("Scheduling Link")
    end
  rescue StandardError
    nil
  end
```

cache.fetch wouldnt have written to the the cache because return causes the enclosing method to return actually not the block

manuthecoder (commented):
looks great! almost there!

app/models/event.rb
polypixeldev: Good catch! fixed

manuthecoder (approved):
nice! lgtm!

app/controllers/events_controller.rb
```diff
@@ -1070,21 +1070,19 @@ def merchants_filter
   def request_call
     authorize @event
 
-    if @event.point_of_contact.present?
-      onboarder_record = OnboardersTable.all(filter: "{HCB ID} = #{@event.point_of_contact.id}").first
+    EventMailer.with(event: @event, user: current_user).ops_call_requested.deliver_now
+    EventMailer.with(event: @event, user: current_user).user_call_requested.deliver_now
```
```suggestion
    EventMailer.with(event: @event, user: current_user).ops_call_requested.deliver_later
    EventMailer.with(event: @event, user: current_user).user_call_requested.deliver_later
```

garyhtou (approved):

PR #13693, opened 2026-05-10
[Contracts] Add safe nav in party reminder job

polypixeldev:
## Summary of the problem

https://appsignal.com/hack-club/sites/6596247683eb67648f30f807/exceptions/incidents/2769/samples `NoMethodError: undefined method 'sent?' for nil`

## Describe your changes

Add a safe nav - it's a little weird that ActiveJob still successfully deserializes a soft-deleted record - we may want to look into seeing if we can patch this as this isn't intuitive, but that requires more investigation for another time/PR.


Luke-Oldenburg (approved):
lgtm

PR #13694, opened 2026-05-10
Fix card styling on event homepage

polypixeldev:
## Summary of the problem

Some cards on the homepage have been missing the `card--breakdown` class which sets a slightly different border color and shadow than the default. 😭

## Describe your changes

Add `card--breakdown` class to the homepage cards that didn't have it. ✨


Luke-Oldenburg (commented):
It seems like some items have card--breakdown on the turbo frame tag instead of the div. We should figure out which is better and standardize how we apply classes.

YodaLightsabr (approved):

PR #13697, opened 2026-05-10
Make ancestor_users order guaranteed and remove confusing line

albertchae:
## Summary of the problem

- We were initially confused about whether ancestor methods contained self, so the first idea was to rename them to `self_and_ancestors` which is similar to some gems
  - https://github.com/amerine/acts_as_tree/blob/9895c5e91bde72f80060b9e920cc85759c99323d/lib/acts_as_tree.rb#L337-L342
  - https://github.com/stefankroes/ancestry#tree-navigation uses `path` for self+ancestors and `ancestors` is ancestors only
- Ultimately we kept it just `ancestor*` because that matches Ruby Module#ancestors.
- The previous implementation didn't guarantee order so we modified it to guarantee order from self->parent->...->root. Some UI was depending on showing root->...->parent so we call reverse there now
- in comment policy there was a place we override a variable immediately (instead of +=) which made it look buggy, just delete that line

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -129,6 +129,10 @@ class Event < ApplicationRecord
       .where("flipper_gates.feature_key = ? AND flipper_gates.key = ?", flag, "actors")
   }
 
+  # Following the convention of Module#ancestors https://apidock.com/ruby/Module/ancestors
+  # this returns the id of self as well as all the ancestors,
+  # in order from self->parent->grandparent->...
+  # We guarantee this order using SEARCH BREADTH FIRST https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-SEARCH
   def ancestor_ids
     [id] + Event.connection.execute(<<-SQL).map { |row| row["id"] }
       WITH RECURSIVE parent_events AS (
@@ -139,8 +143,8 @@ class Event < ApplicationRecord
         SELECT e.id, e.parent_id
         FROM events e
         INNER JOIN parent_events pe ON e.id = pe.parent_id
-      )
-      SELECT id FROM parent_events WHERE id != #{id};
+      ) SEARCH BREADTH FIRST BY id SET ordercol
+      SELECT id FROM parent_events WHERE id != #{id} ORDER BY ordercol;
     SQL
   end
 
@@ -159,8 +163,14 @@ class Event < ApplicationRecord
     SQL
   end
 
+  # Following the convention of Module#ancestors https://apidock.com/ruby/Module/ancestors
+  # this returns self as well as all the ancestors
   def ancestors
-    Event.where(id: ancestor_ids)
+    # array_position preserves the order from ancestor_ids; sanitize_sql_array parameterizes the ids so it's statically safe and passes brakeman.
+    fetched_ancestor_ids = ancestor_ids
+    Event.where(id: fetched_ancestor_ids).reorder(Arel.sql(
+                                                    Event.sanitize_sql_array(["array_position(ARRAY[?]::bigint[], events.id)", fetched_ancestor_ids])
+                                                  ))
   end
 
   def descendants
--- a/app/policies/comment_policy.rb
+++ b/app/policies/comment_policy.rb
@@ -47,7 +47,6 @@ class CommentPolicy < ApplicationPolicy
     user_list = []
 
     if record.commentable.respond_to?(:events)
-      user_list = record.commentable.events.collect(&:users).flatten
       user_list = record.commentable.events.collect(&:ancestor_users).flatten
     elsif record.commentable.is_a?(Reimbursement::Report)
       user_list = [record.commentable.user]
--- a/app/views/events/_title.html.erb
+++ b/app/views/events/_title.html.erb
@@ -1,4 +1,4 @@
-<% parent_events ||= @event.ancestors.reject { |e| e.id == @event.id } %>
+<% parent_events ||= @event.ancestors.reject { |e| e.id == @event.id }.reverse %>
 <% if parent_events.present? %>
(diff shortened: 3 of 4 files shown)


app/policies/comment_policy.rb
```diff
@@ -47,14 +47,13 @@ def users
     user_list = []
 
     if record.commentable.respond_to?(:events)
-      user_list = record.commentable.events.collect(&:users).flatten
```
albertchae, line 50: this line was a no-op

app/views/events/_title.html.erb
```diff
@@ -1,4 +1,4 @@
-<% parent_events ||= @event.ancestors[0...-1].reject { |e| e.id == @event.id } %>
+<% parent_events ||= @event.self_and_ancestors[0...-1].reject { |e| e.id == @event.id } %>
```
albertchae: this seems like a bug but not sure, see https://github.com/hackclub/hcb/pull/13701

[0...-1] is exclusive of the end so it would cut off the last element

app/models/event.rb
```diff
@@ -129,7 +129,7 @@ class Event < ApplicationRecord
       .where("flipper_gates.feature_key = ? AND flipper_gates.key = ?", flag, "actors")
   }
 
-  def ancestor_ids
+  def self_and_ancestor_ids
```
albertchae: add comment

app/models/event.rb
```diff
@@ -159,8 +159,8 @@ def descendant_ids
     SQL
   end
 
-  def ancestors
-    Event.where(id: ancestor_ids)
+  def self_and_ancestors
```
albertchae: add comment

app/models/event.rb
```diff
@@ -292,12 +292,12 @@ def descendants
...
 
-  def ancestor_organizer_positions
-    OrganizerPosition.where(event_id: ancestor_ids)
+  def self_and_ancestor_organizer_positions
+    OrganizerPosition.where(event_id: self_and_ancestor_ids)
   end
 
-  def ancestor_users
-    User.where(id: ancestor_organizer_positions.select(:user_id))
+  def self_and_ancestor_users
```
albertchae: add comment

app/views/events/_title.html.erb
albertchae: @garyhtou confirmed this is a bug

garyhtou (commented):
We'll keep the `ancestor` name since the Ruby method includes `self`. We'll just add a comment clarifying it includes `self`

app/views/events/_title.html.erb
albertchae: fixed in another PR

app/models/event.rb
```diff
@@ -139,8 +143,8 @@ def ancestor_ids
         SELECT e.id, e.parent_id
         FROM events e
         INNER JOIN parent_events pe ON e.id = pe.parent_id
-      )
-      SELECT id FROM parent_events WHERE id != #{id};
+      ) SEARCH BREADTH FIRST BY id SET ordercol
+      SELECT id FROM parent_events WHERE id != #{id} ORDER BY ordercol;
```
albertchae, line 147: I think we were seeing this order in practice anyway but now it's guaranteed

https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-SEARCH

app/models/event.rb
```diff
@@ -159,8 +163,14 @@ def descendant_ids
...
   end
 
+  # Following the convention of Module#ancestors https://apidock.com/ruby/Module/ancestors
+  # this returns self as well as all the ancestors
   def ancestors
-    Event.where(id: ancestor_ids)
+    # array_position preserves the order from ancestor_ids; sanitize_sql_array parameterizes the ids so it's statically safe and passes brakeman.
+    fetched_ancestor_ids = ancestor_ids
+    Event.where(id: fetched_ancestor_ids).reorder(Arel.sql(
+                                                    Event.sanitize_sql_array(["array_position(ARRAY[?]::bigint[], events.id)", fetched_ancestor_ids])
```
albertchae, line 172: this orders the results by the order of the ids we get from ancestor_ids

app/views/events/_title.html.erb
```diff
@@ -1,4 +1,4 @@
-<% parent_events ||= @event.ancestors.reject { |e| e.id == @event.id } %>
+<% parent_events ||= @event.ancestors.reject { |e| e.id == @event.id }.reverse %>
```
albertchae, line 1: now that order is guaranteed, we reverse here because the UI intentionally wants order from oldest ancestor to parent

manuthecoder (approved):
lgtm!

app/models/event.rb
```diff
@@ -293,7 +303,11 @@ def descendants
   has_many :organizer_positions, dependent: :destroy
 
   def ancestor_organizer_positions
-    OrganizerPosition.where(event_id: ancestor_ids)
+    # array_position preserves the order from ancestor_ids; sanitize_sql_array parameterizes the ids so it's statically safe and passes brakeman.
+    fetched_ancestor_ids = ancestor_ids
+    OrganizerPosition.where(event_id: fetched_ancestor_ids).order(Arel.sql(
+                                                                    OrganizerPosition.sanitize_sql_array(["array_position(ARRAY[?]::bigint[], event_id)", fetched_ancestor_ids])
+                                                                  ))
```
garyhtou: I have a feeling we usually don't care whether it's ordered by events for organizer positions

garyhtou (approved):

app/models/event.rb
albertchae: yep good call, removed

PR #13699, opened 2026-05-10
Make comment order_by created_at default in concern

albertchae:
## Summary of the problem

This seems like it shouldn't belong in the partial nor would we want the default order from postgres (which is often by id but is actually undefined)

### DIFF
--- a/app/models/concerns/commentable.rb
+++ b/app/models/concerns/commentable.rb
@@ -3,7 +3,7 @@
 module Commentable
   extend ActiveSupport::Concern
   included do
-    has_many :comments, as: :commentable
+    has_many :comments, -> { order(:created_at) }, as: :commentable, inverse_of: :commentable
   end
 
   def comment_recipients_for(comment)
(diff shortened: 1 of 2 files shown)


app/models/concerns/commentable.rb
```diff
@@ -3,7 +3,7 @@
 module Commentable
   extend ActiveSupport::Concern
   included do
-    has_many :comments, as: :commentable
+    has_many :comments, -> { order(:created_at) }, as: :commentable
```
albertchae: look for best practice around default scope

albertchae: @garyhtou found a couple links

https://stackoverflow.com/a/71469664
https://www.bigbinary.com/blog/rails-6-adds-implicit_order_column
https://discuss.rubyonrails.org/t/default-order-for-relations/75554
https://womanonrails.com/default-scope
https://tekin.co.uk/2018/12/configurable-implicit-ordering-in-activerecord-models-coming-in-rails-6

not all of these are specifically about associations, but if it works I think the best way to do this might be to set `implicit_order_column` to created_at on Comment and see if that gets applied to the association correctly

albertchae: `implicit_order_column` only works on first/second/last type methods and not on where

also I think the warning about the practice is more about `default_scope`, since this is on an association i think it's less frowned on, e.g. official rails docs https://guides.rubyonrails.org/association_basics.html#scopes

garyhtou: sounds good!

garyhtou (approved):

PR #13700, opened 2026-05-10
[Performance] Use combobox for event selection in forms

manu-gurudath:
Replaces `<select>`s preloaded with all events with the new, performant, combobox selector!


Luke-Oldenburg (approved):
lgtm

PR #13701, opened 2026-05-10
Fix bug with parent event titles

albertchae:
This seemed like a potential bug where we are cutting off an ancestor from the array

in local dev console
```
PaperTrail.request(enabled: false) do
  gp = Event.create!(name: "GP", slug: "demo-gp-#{SecureRandom.hex(3)}", country: "US")
  p  = Event.create!(name: "P",  slug: "demo-p-#{SecureRandom.hex(3)}",  country: "US", parent: gp)
  c  = Event.create!(name: "C",  slug: "demo-c-#{SecureRandom.hex(3)}",  country: "US", parent: p)

  ordered = c.self_and_ancestors.sort_by { |e| e.id == c.id ? 0 : 1 }
  puts "ordered: #{ordered.map(&:name).inspect}"
  puts "buggy:   #{ordered[0...-1].reject { |e| e.id == c.id }.map(&:name).inspect}"
  puts "fixed:   #{ordered.reject       { |e| e.id == c.id }.map(&:name).inspect}"
end
```
ordered: ["C", "GP", "P"]
buggy:   ["GP"]
fixed:   ["GP", "P"]

### DIFF
--- a/app/views/events/_title.html.erb
+++ b/app/views/events/_title.html.erb
@@ -1,4 +1,4 @@
-<% parent_events ||= @event.ancestors[0...-1].reject { |e| e.id == @event.id } %>
+<% parent_events ||= @event.ancestors.reject { |e| e.id == @event.id } %>
 <% if parent_events.present? %>
   <div data-controller="menu" class="-mb-2">
     <button


albertchae:
https://github.com/hackclub/hcb/pull/13697#discussion_r3214425284

discussed with @garyhtou , this is a bug on because [0...-1] is exclusive of the end so it would cut off the last element

garyhtou (approved):
LGTM. I'll note that the ordering in the example code in the PR description feels a bit off. Please double check that `ancestors` returns (great-)grandparent events first, then followed by the parent, and then self. We want the "oldest" first and "youngest" last.

The correct ordering for the example is `[GP, P, C]`. (and the view would reject self `C`). If the `ancestor` method is currently unordered, then we should enforce that order in the view.

We can merge as-is, and come back to this ordering in a separate PR if necessary

albertchae:
> LGTM. I'll note that the ordering in the example code in the PR description feels a bit off. Please double check that `ancestors` returns (great-)grandparent events first, then followed by the parent, and then self. We want the "oldest" first and "youngest" last.
> 
> The correct ordering for the example is `[GP, P, C]`. (and the view would reject self `C`). If the `ancestor` method is currently unordered, then we should enforce that order in the view.
> 
> We can merge as-is, and come back to this ordering in a separate PR if necessary

ok good point, will address consistent ordering of ancestor methods in https://github.com/hackclub/hcb/pull/13697

garyhtou:
Checks are bugging. i'm going to push and empty commit to retrigger them.
image

PR #13702, opened 2026-05-10
[Suborganizations] Fix search bar gap

manuthecoder:
Noticed this inconsistent gap when working on mermaid graphs:
image

Fix:
image


Luke-Oldenburg (approved):
lgtm

PR #13705, opened 2026-05-10
[Card Locking] Move to a time-based upload requirement

manuthecoder:
## Summary of the problem

Closes #13194. Card locking previously used a coarse "too many missing receipts" threshold. This moves it to a time-based requirement: every settled card charge needs a receipt within ~7 days, and a cardholder who falls behind has their cards paused until they upload.

## Describe your changes

**The rule cardholders see:** upload a receipt within 7 days of each card charge. Miss it and your cards pause until you upload; uploading unlocks them within seconds. (Cardholders with a strong on-time history quietly get a longer grace window, capped at 14 days. They are never told about it, so the public promise stays a simple 7 days.)

**How it works**

- Each receipt-required Stripe card charge (`HcbCode`) carries persisted timing: `receipt_settled_at`, `receipt_due_at`, `receipt_resolved_at`. A charge locks the cardholder's cards when it is past `receipt_due_at` and still unresolved; uploading a receipt (or marking it no/lost) resolves it and unlocks the cards within seconds via a high-priority job.
- Deadlines are computed by `CardLocking::Deadline` (7-day base; slides up to last-charge + 7d for trusted cardholders, capped at settled + 14d, with a 72h floor so losing trust never makes a pile overdue at once). Trust is `CardLocking::TrustAssessment` (on-time rate ≥ 80% over 6 months plus the most-recent charge on time).
- A 5-minute cron (`User::UpdateCardLocking::RecurringJob`) refreshes deadlines, applies locks/unlocks, and sends the pile warning. Settle hooks on both transaction engines materialize timing as charges land.
- The logic lives in card-locking-specific classes (`CardLocking::*` services plus the `CardLocking::ChargeBehavior` / `CardLocking::CardholderBehavior` model concerns), so the coupling to `HcbCode` is concentrated and easy to remove when `HcbCode` goes away.

**Communication:** a pile warning (email + SMS, only when a receipt is actually approaching its deadline, at most once a day), locked and unlocked emails + SMS, a point-of-sale decline SMS, the Receipts inbox callout, and a `card_locking` field on `GET /api/v4/user` for the mobile app (hcb-mobile#423).

**Screenshots:** _pending._

## Rolling it out

Enforcement is staged per cardholder with Flipper. A cardholder's charges only become lockable once the cardholder is in a stage; anyone in no stage never locks, so this is safe to turn up incrementally.

1. **Deploy.** The migration adds the timing columns and indexes (strong_migrations-safe; the deadline index is added concurrently).
2. **Backfill timing** for existing charges: run `Maintenance::BackfillCardLockingTimingTask` from the maintenance-tasks UI. It is idempotent, and the 5-minute cron keeps timing fresh afterwards.
3. **Turn the feature on** for the audience: enable `card_locking_2025_06_09`. This shows the inbox callout, sends the comms, and runs the lock logic. On its own (no stage flag) nobody locks yet.
4. **Pilot (July 14):** enable `card_locking_enabled_on_07_14_2026` for the small pilot group. Their charges settled on or after July 14 get a 7-day deadline and can lock.
5. **Everyone (July 28):** enable `card_locking_enabled_on_07_28_2026` for all cardholders. Their enforcement start date is July 28, so nothing older than that can lock them.
6. **Monitor** lock volume and the `card_locking` error reports. An admin can pause an individual cardholder with the "suppress card locking" action on the user's admin page.

**Adding another stage later:** add a `[flag, date]` row to `CardLocking::ENFORCEMENT_STAGES` (earliest first) and create the matching Flipper flag.

**Ripping the flippers out** once everyone is enrolled: delete `CardLocking::ENFORCEMENT_STAGES` and `CardLocking.enforcement_start_date`, have callers use `CardLocking::ENFORCEMENT_START_DATE` directly (a single date, currently July 14), and delete the `card_locking_enabled_on_*` flags.


garyhtou:
> [!NOTE]
> Review performed by Claude (Opus 4.8), driven by @garyhtou. Findings verified against the source; I'm pushing fixes for everything marked **Fixing** to this branch.

## Summary

The policy engine faithfully implements #13194 — the average math, the four safeguards, the settle-time basis, and the "let one authorization slip through" property all check out, and `spec/models/user/card_locking_spec.rb` exercises them against real settled transactions rather than stubs. Nice work.

The concerns are mostly operational. The hot path went from *"locked users, hourly"* to *"every user with any missing receipt, every 5 minutes"*, and each of those users now materializes their entire card history since 2025-06-13 into Ruby, with an N+1 on `event.plan`, behind a JSONB join that can't use an index.

Also worth confirming: removing `SendCardLockingNotificationJob` from `stripe_controller.rb` is **not** a regression. The "you were charged, upload a receipt" email/SMS lives on a different webhook (`issuing_authorization.created` → `StripeAuthorizationService::CreateFromWebhook`) and is untouched. That copy doesn't mention the 72-hour deadline though — worth a follow-up in `canonical_pending_transaction_mailer.rb` and `send_twilio_receipt_message_job.rb`.

---

## Critical

**1. N+1 on `event` / `event.plan` for every HCB code, every 5 minutes.** — *Fixing*

`card_locking_relevant_hcb_codes` (`app/models/user.rb:457`) preloads only `:receipts` and `:canonical_transactions`. But `card_locking_missing_receipt?` → `missing_receipt?` → `receipt_required?` (`app/models/hcb_code.rb:664`) evaluates `event&.plan&.receipt_required?` and `pt&.declined?`. `HcbCode#event` (`hcb_code.rb:262`) is `super || events.first`, which when `event_id` is nil runs two `pluck` queries plus fallbacks. That's 3–4 queries per HCB code, per candidate user, per tick.

**2. Unbounded working set.** — *Fixing*

`card_locking_relevant` (`hcb_code.rb:70`) filters `created_at >= CARD_LOCKING_START_DATE` with no upper bound, and `card_locking_relevant_hcb_codes` calls `.to_a` on it. The 6-month `CARD_LOCKING_AVERAGE_LOOKBACK` is applied *in Ruby* afterwards (`user.rb:473`), so it doesn't bound the query. `cards_should_lock?` forces the full materialization on its first line. This set grows monotonically forever, and it now also runs on every `/my/inbox` render (`my_controller.rb:99`).

**3. The JSONB join is unindexed.** — *Fixing*

`CARD_LOCKING_STRIPE_CARD_JOIN` (`hcb_code.rb:63`) joins on `raw_stripe_transactions.stripe_transaction->>'card'`. The existing index is on `((stripe_transaction -> 'card') ->> 'id')` — a *different* expression — so this can't use it and will sequential-scan `raw_stripe_transactions`. The scalar `->>'card'` is semantically correct (it matches `RawStripeTransaction#stripe_card_id`); the index is what's missing.

**4. `RecurringJob` has no per-user error isolation.** — *Fixing*

`recurring_job.rb:8-11` iterates `find_each` with no `rescue`. One raise — Twilio outage, mailer blowup, `update!` failure — aborts the whole tick, and because `ApplicationJob` declares `discard_on(Twilio::REST::RestError)`, a Twilio `RestError` kills the batch *silently, with no retry*. `find_each` is ordered by id, so one reliably-failing user permanently starves every user after them, every tick. This shape pre-existed, but it only iterated locked users before.

---

## Important

**5. A user can be locked with zero violations, and the email admits it.** — *Fixing (copy only)*

`cards_should_lock?` (`user.rb:531`) short-circuits on `card_locking_missing_receipts.count >= 75` *before* checking violations. The resulting email renders "you have 75 receipts missing", an empty violations table, and a section headed "Missing receipts still within 72 hours". Per @garyhtou, 75 is high enough that this is acceptable policy — but the messaging should say *"you have too many missing receipts"* rather than implying a 72-hour violation.

**6. The 72-hour boundary is double-counted.** — *Fixing*

A receipt is a violation at `age >= 72h` (`hcb_code.rb:716`) *and* counts as a timely upload at `upload_time <= 72h` (`user.rb:512`). At exactly 72.0h it is both. Per @garyhtou, exactly-72h should be **timely**, so the violation check becomes strictly `>`.

**7. Numerator and denominator use different predicates.** — *Fixing*

`average_receipt_upload_time` and `timely_receipt_upload_count` draw from `card_locking_relevant_hcb_codes`, which is **not** filtered by `receipt_required?`, while `card_locking_missing_receipts` is. Adding an explicit `receipt_required?` filter so both sides agree.

For the record, receiptables that attach a receipt at creation time (ACH transfers, reimbursement expenses) **cannot** enter this scope — `card_locking_relevant` inner-joins `raw_stripe_transactions` → `stripe_cards`, so only HCB-600 (card charge) and HCB-601 (force capture) codes match. The analogous case that *does* exist is a user replying to the charge SMS to attach a receipt to the still-pending transaction; `card_locking_receipt_upload_time` clamps that to `0`, which is the intended incentive per #13194 ("clamp to be non-negative").

**8. Already-locked users get no digest.** — *Fixing*

`send_card_locking_notification.rb:12` returns early on `@user.cards_locked?`. #13194 says "If you have any missing receipt violations, we send you a digest every so often."

**9. `HcbCode#card_locking_receipt_age` defaults to `0.seconds` when `settled_at` is nil** (`hcb_code.rb:705`) — i.e. "brand new, never a violation". Unreachable today because every caller gates on `card_locking_missing_receipt?`, but an unknown value silently defaulting to *compliant* is the wrong direction for a locking decision.

**10. `User::SendCardLockingNotificationJob` is dead code.** — *Fixing*

Nothing references it outside its own spec; `RecurringJob` calls the service directly. Its `event:` kwarg is unused and it lost the `event.plan.receipt_required?` guard. Deleting it. ⚠️ Note: `main` currently enqueues it with `wait: 24.hours`, so for up to 24h after deploy Sidekiq will fail those in-flight jobs into the dead set. Harmless, but expect the noise.

---

## Minor

- `User.card_locking_candidates` (class method → users) and `HcbCode.card_locking_candidates` (scope → HCB codes) share a name and mean different things.
- `card_locking_receipt_age` returns a `Float` in one branch and an `ActiveSupport::Duration` in the other.
- `includes` + string `where` currently resolves to a **preload**, so `canonical_transactions` loads unfiltered and `card_locking_settled_at` re-filters negatives in Ruby. Correct today — but adding `.references(:canonical_transactions)` would flip it to `eager_load`, silently pre-filtering the association and turning the Ruby branch into a no-op.
- `my/inbox.html.erb:82` renders "0 violations" for nearly everyone. And `@count` (old path: includes pending + Emburse) vs `@locking_count` (new path: settled Stripe only) come from different definitions, so the two badges can look contradictory.
- Two unrelated commits (`00d2e8f3`, `6e00759e`) drag `spec/controllers/api/v4/donations_controller_spec.rb` churn into this PR.

---

## Tests

All 25 new examples pass, and the model spec is genuinely good — it builds raw Stripe transaction → canonical transaction → event mapping → HCB code end-to-end. Gaps cluster in exactly the seams this PR introduces:

- **The 10-minute warning window is stubbed on both ends.** `send_card_locking_notification_spec.rb` stubs `card_locking_receipts_reaching_warning_threshold` wholesale, and there's no model-level test for it. Nothing verifies that a charge which crossed 48h fifteen minutes ago is excluded. Given this gates the entire warning path, it's the most important missing test.
- **`User.card_locking_candidates` is untested.** If it's wrong, locked users never unlock and violators never lock — a total silent failure.
- **`RecurringJob` has no spec**, despite now running two services per user over a new scope.
- **`unlock_only` is never tested from the state that matters:** an *unlocked* user whose `cards_should_lock?` is true must stay unlocked. The existing test starts from `cards_locked: true` and only hits the no-op branch.
- No coverage for the `Receipt` `after_commit` destroy path or the new `if user.present?` guard; the within-grace exclusion from the average; the exactly-72h and exactly-1-week boundaries; or the inactive-plan exclusion.

---

## Separately worth deciding

- #13194 says "greater than X" violations / missing receipts; the code locks *at* 10 and *at* 75 (`>=`).
- #13194 says unlock-on-upload only when zero violations remain. `unlock_only` mode instead unlocks whenever `!cards_should_lock?`, which can be true with violations still open. Benign — the cron would do it within 5 minutes anyway — but not what the issue describes.
- The 48h/71h warnings only fire if a tick happens to observe the charge inside a 10-minute window. If a tick takes >10 min to reach a user, or dies, that warning is silently lost forever. A durable "have we warned this user about this charge at this threshold yet" check would be robust.

manu-gurudath:
LGTM!

garyhtou (approved):
🪄 Please review carefully - i didn't read through most of this code myself

app/jobs/user/send_sms_job.rb
polypixeldev, line 1: This seems odd to scope under card locking as it doesn't have anything specific to card locking

app/jobs/user/update_card_locking/recurring_job.rb
```diff
@@ -4,10 +4,31 @@ class User
...
           ::UserService::UpdateCardLocking.new(user:).run
+          ::UserService::SendCardLockingNotification.new(user:).run
+        rescue => e
+          failed += 1
+          Rails.error.report(e, context: { user_id: user.id })
         end
+
+        # Reporting each failure individually would otherwise let an outage look
+        # like a successful run.
+        raise "Card locking failed for #{failed} of #{processed} users" if failed > [FAILURE_TOLERANCE, processed / 5].max
```
polypixeldev, line 31: I don't think we need the whole FAILURE_TOLERANCE system - we still report each error individually, so all it does is stop the job for later user IDs that could still succeed

db/schema.rb
```diff
@@ -1423,12 +1423,17 @@
     t.text "hcb_code", null: false
     t.bigint "ledger_item_id"
     t.datetime "marked_no_or_lost_receipt_at", precision: nil
+    t.datetime "receipt_due_at"
+    t.datetime "receipt_resolved_at"
+    t.datetime "receipt_settled_at"
```
polypixeldev: `receipt_settled_at` is a little misleading as it has nothing to do with the receipt, just when the card charge settled

app/services/card_locking/receipt_resolution.rb
```diff
@@ -0,0 +1,50 @@
...
+  module ReceiptResolution
+    module_function
+
+    # A receipt was created or updated. Materialize synchronously so the unlock
+    # job sees the resolution, then recompute the lock. notify_progress lets a
+    # still-locked cardholder hear that this upload landed but more remain.
+    def on_receipt_upsert(receipt)
+      charge = receipt.receiptable
+      charge.materialize_card_locking! if charge.is_a?(HcbCode) && charge.card_locking_chargeable?
+      enqueue_unlock(receipt.user, notify_progress: true)
```
polypixeldev: An unlock check is enqueued for the person who uploaded the receipt, but that might not be the person who made the transaction. Receipt user is also nullable - a reasonable case is I get cards locked, I go to the email to upload my receipt but I'm not logged in, so the receipt does not belong to any user.

This should instead check the author or user on the charge, like how on_no_or_lost_receipt does it.

app/services/user_service/send_card_locking_notification.rb
```diff
@@ -1,35 +1,54 @@
...
-          if @user.phone_number.present? && @user.phone_number_verified?
-            message = "You now have #{current_count} transactions missing receipts from more than a day ago. If you have ten or more missing receipts, your cards will be locked. You can manage your receipts at #{Rails.application.routes.url_helpers.my_inbox_url}."
+      count = @user.card_locking_outstanding_count
+      return if count.zero?
 
-            TwilioMessageService::Send.new(@user, message).run!
-          end
-        end
+      key = "card_locking_digest:#{@user.id}"
+      return unless Rails.cache.write(key, true, expires_in: 25.hours, unless_exist: true)
```
polypixeldev: Why 25 hours instead of 24?

app/services/user_service/send_card_locking_notification.rb
```diff
@@ -1,35 +1,54 @@
...
+    else
+      CardLocking::SendSmsJob.perform_later(user_id: @user.id, body: sms_message(count))
+    end
 
-          TwilioMessageService::Send.new(@user, message).run!
-        end
-      end
+    def sms_message(count)
+      noun = "receipt".pluralize(count)
+      "You have #{count} #{noun} to upload. Cards that fall behind get locked. Upload them at #{CardLocking.inbox_url}."
```
polypixeldev: "Cards that fall behind" suggests a per-card lock/limit

app/services/card_locking/receipt_resolution.rb
```diff
@@ -0,0 +1,50 @@
...
+      charge = receipt.receiptable
+      # receipt_resolved_at is only ever cleared here, never revised. Once a charge
+      # is resolved the timestamp is frozen, so destroying an earlier receipt while
+      # a later one remains (card_locking_resolved? still true) leaves the original
+      # resolution timestamp in place. It resets to nil only when the charge becomes
+      # genuinely unresolved again (no receipts, not marked no/lost).
+      if charge.is_a?(HcbCode) && charge.card_locking_chargeable? && !charge.card_locking_resolved?
+        charge.update_columns(receipt_resolved_at: nil)
+      end
+      enqueue_unlock(receipt.user)
```
polypixeldev: See https://github.com/hackclub/hcb/pull/13705/changes#r3581579519

app/jobs/user/update_card_locking/recurring_job.rb
garyhtou: _Reply from Claude Code (driven by @garyhtou):_

Small correction on the mechanism before we drop it: the `rescue` is the block-level rescue *inside* the `find_each` block, so a per-user failure is caught, reported, and iteration continues to the next user. No later user IDs get skipped. The `raise` runs *after* the loop finishes, once every candidate has already been processed, so `FAILURE_TOLERANCE` never aborts the batch.

Its only purpose is to act as an outage tripwire: if something systemic fails (e.g. Twilio is down and every `SendCardLockingNotification` raises), the post-loop `raise` turns the run into a failed/retried job instead of a green run that quietly reported thousands of individual errors. That matters because `ApplicationJob`'s `discard_on(Twilio::REST::RestError)` would otherwise swallow those per-user reports and the run would still look successful.

So I kept it for now. Totally happy to simplify if the `[FAILURE_TOLERANCE, processed / 5].max` threshold feels over-engineered, e.g. drop the absolute floor and trip only on a high failure *rate*, or only when every processed user failed. Which would you prefer?

polypixeldev: That makes sense that it's to show as a failed job. While I don't think that a single failed job that runs every 5 minutes anyways will be a huge warning amongst the thousands of alerts from the individual errors, it is still helpful data.

polypixeldev (commented):
There doesn't seem to be any admin UI to call the suppress_card_locking endpoint, and I think it would also make sense to show to the user when their cards will lock again if they don't upload their overdue receipts. The logic works though!

app/jobs/user/update_card_locking/recurring_job.rb
garyhtou: _Reply from Claude Code (driven by @garyhtou):_

Agreed, the failed-job marker is minor next to the flood of per-user error reports during a real outage. The value is just that it gives one durable "this run was systemically bad" signal (and a retry) rather than a green run, so it stays cheap and out of the way. Leaving it in for now; happy to rip it out if it turns out to be more noise than signal in practice. Thanks for talking it through.

garyhtou:
_Reply from Claude Code (driven by @garyhtou), re: [this review](https://github.com/hackclub/hcb/pull/13705#pullrequestreview-4698227075):_

**Admin UI for `suppress_card_locking`** — good catch, there genuinely was no UI. Added in 4936cd8: a "Card locking" section on the admin user page showing the current lock/suppression state, with an admin-only form to pause locking for N hours (1-720). Auditors see the status but not the form, matching the endpoint's admin-only policy.

**Showing users when their cards will lock** — this one is a deliberate design choice rather than an oversight. The feature intentionally shows only the simple "within 7 days of each charge" promise and never a per-charge deadline, because trusted cardholders quietly get a longer grace window (up to 14 days) that is **hidden** from them, so the public rule stays a flat 7 days and can't be gamed. Surfacing the real `receipt_due_at` per charge would leak that grace for trusted users and turn the simple promise into a precise countdown that sometimes doesn't fire on the stated day.

That said, your underlying point (users not knowing they're at risk) is fair. One middle ground we're considering: once someone is *past* the public 7-day mark but still inside their grace window, show a nudge like "N days late" — it creates urgency without revealing the actual deadline or the grace length. Happy to talk through whether that (or something else) is worth doing. cc @garyhtou

polypixeldev:
Will look at admin UI! For the showing to the user, I agree that we should show a nudge to the user when in the grace period, although I mainly meant when card locking is suppressed for them by an admin so that they aren't surprised when supression expires.

garyhtou:
That makes a lot of sense - we can add that later!

PR #13709, opened 2026-05-10
[Mailer] Force weekly receipt reports

### LINKED ISSUE #13655: [User Setting] Force "weekly" receipt reminders for everyone
Currently, HCB offers the following options for missing receipt emails
- None
- Weekly
- Monthly

This gives users the impression that they can just set it to "None" and forget about it. ~~We need to remove the None option, and also increase the email frequency at large. Given the ongoing conversations of card locking / missing receipts, we should increase our annoyance first before announcing card locks.~~

- ~~If a user has a missing receipts over 30 days old, they should be getting a daily email with a list of there missing receipts.~~
- Otherwise, they should get a weekly email with a list of there missing receipts. (We should remove the setting entirely).

Luke-Oldenburg:
## Summary of the problem

closes #13655


manuthecoder (changes requested):
lgtm! but make sure you remove the footer in [receipt_report.html.rb](https://github.com/hackclub/hcb/blob/main/app/views/receiptable_mailer/receipt_report.html.erb#L40)

manu-gurudath:
also there's no database migration, should we drop the column now that it's no longer used? a little dead weight we could remove

Luke-Oldenburg:
@manuthecoder we are planning on keeping for now in case we decide to revert.

manuthecoder (approved):
lgtm!

YodaLightsabr (approved):

PR #13713, opened 2026-05-11
[v4] Implement Reimbursements suite 

Mohamad-Mortada:
## Summary of the problem
We all would love if HCB Mobile supported reimbursements. Well now we can via the API :)

## Describe your changes
Created ability to create, read, update, and delete (CRUD) for both reports and expenses


app/policies/reimbursement/expense_policy.rb
```diff
@@ -2,6 +2,10 @@
 
 module Reimbursement
   class ExpensePolicy < ApplicationPolicy
+    def show?
+      auditor? || creator || reader?
```
manuthecoder:
```suggestion
      admin || auditor || creator || reader
```

app/policies/reimbursement/expense_policy.rb
```diff
@@ -38,10 +42,18 @@ def admin
       user&.admin?
     end
 
+    def auditor?
```
manuthecoder:
```suggestion
    def auditor
```

app/policies/reimbursement/expense_policy.rb
```diff
@@ -38,10 +42,18 @@ def admin
...
+      user&.auditor?
+    end
+
     def manager
       record.event && OrganizerPosition.role_at_least?(user, record.event, :manager)
     end
 
+    def reader?
```
manuthecoder:
```suggestion
    def reader
```

app/controllers/api/v4/reimbursement/expenses_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+              attachments: Array(params[:receipt]),
+              upload_method: :api,
+              receiptable: @expense
+            ).run!
+          end
+
+          render :show, status: :ok
+        end
+
+
```
manuthecoder, line 57:
```suggestion
```

config/routes.rb
```diff
@@ -758,6 +758,12 @@
         resources :sponsors, only: [:index, :show, :create]
         resources :check_deposits, only: [:index, :show, :create]
         resources :ach_transfers, only: [:create]
+        resources :reimbursement_reports, controller: "reimbursement/reports", only: [:index, :show, :create, :update, :destroy] do
+          member do
+            post :submit
```
polypixeldev, line 787: Should there also be a route to mark the report as a draft?

app/controllers/api/v4/reimbursement/reports_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+      class ReportsController < ApplicationController
+        before_action :set_report, only: [:show, :update, :destroy, :submit]
+
+        include ApplicationHelper
+
+        def index
+          if params[:organization_id]
+            @event = Event.find_by_public_id!(params[:organization_id])
+            authorize @event, :show_in_v4?
+            reports = @event.reimbursement_reports.order(created_at: :desc)
```
polypixeldev: Something to consider here is that on web, we're limiting the reports that we show on the event reimbursements page to the `visible` scope, which are those that are not from a user that hasn't signed in yet, to protect organizers from bot spam. It could be confusing if the API is returning reports that you can't see on web - maybe we add a param like `show_hidden_reports` that defaults to false?

app/controllers/api/v4/reimbursement/reports_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+          report_params = params.require(:reimbursement_report).permit(
+            :name,
+            :invite_message,
+            :maximum_amount_cents
+          )
+
+          @report = @event.reimbursement_reports.build(
+            report_params.merge(
+              user: current_user,
+              inviter: current_user,
```
polypixeldev: The policy allows users who are not organizers to create if public reimbursements are enabled. In this case, the inviter should be nil - see https://github.com/hackclub/hcb/blob/v4-reimbursements/app/controllers/reimbursement/reports_controller.rb#L19

app/controllers/api/v4/reimbursement/reports_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+        end
+
+        def update
+          authorize @report
+
+          update_params = params.require(:reimbursement_report).permit(
+            :name,
+            :maximum_amount_cents,
+            :organization_id,
+            :reviewer_id
```
polypixeldev, line 64: Only admins or managers should be able to set the reviewer for a report

app/controllers/api/v4/reimbursement/reports_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+
+          render :show, status: :created
+        end
+
+        def update
+          authorize @report
+
+          update_params = params.require(:reimbursement_report).permit(
+            :name,
+            :maximum_amount_cents,
```
polypixeldev, line 62: Only admins or managers should be able to update the max amount, and only if the report is in states draft or submitted

app/controllers/api/v4/reimbursement/reports_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+        end
+
+        def destroy
+          authorize @report
+          @report.destroy!
+          render json: { message: "Reimbursement report successfully deleted" }, status: :ok
+        end
+
+        def submit
+          authorize @report, :submit?
```
polypixeldev: Does the `:submit?` need to be here?

app/controllers/api/v4/reimbursement/expenses_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+        end
+
+        private
+
+        def set_expense
+          @expense = ::Reimbursement::Expense.find_by_public_id!(params[:id])
+        end
+
+        def expense_attrs
+          params.require(:reimbursement_expense).permit(:memo, :description, :category, :value)
```
polypixeldev: Type should be permitted

app/controllers/api/v4/reimbursement/expenses_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+
+          authorize @expense
+
+          @expense.save!
+
+          if params[:receipt].present?
+            ::ReceiptService::Create.new(
+              uploader: current_user,
+              attachments: Array(params[:receipt]),
+              upload_method: :api,
```
polypixeldev, line 32: Similar controlllers like those for ACHs and checks use a unique upload method (like `check_api`) - should we do the same here?

I think there's no need to differentiate what type of receiptable it is in the upload method, but we should stay consistent and update in a future PR if needed

app/controllers/api/v4/reimbursement/expenses_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+        def update
+          authorize @expense
+
+          @expense.update!(expense_attrs)
+
+          if params[:receipt].present?
+            ::ReceiptService::Create.new(
+              uploader: current_user,
+              attachments: Array(params[:receipt]),
+              upload_method: :api,
```
polypixeldev, line 49: Same as https://github.com/hackclub/hcb/pull/13713/changes#r3295924563

app/views/api/v4/reimbursement/expenses/_expense.json.jbuilder
```diff
@@ -0,0 +1,20 @@
...
+  json.memo expense.memo
+  json.description expense.description
+  json.category expense.category
+  if expense.is_mileage?
+    json.expense_type "mileage"
+  else
+    json.expense_type expense.is_fee? ? "fee" : "standard"
+  end
+  json.amount_cents expense.amount_cents
+  json.value expense.value.to_f
```
polypixeldev, line 11: Why is to_f needed here?

polypixeldev (commented):
Will test later, but these are some comments from looking through the code.

config/routes.rb
Mohamad-Mortada: Good eye, I'll add a draft route too.

app/controllers/api/v4/reimbursement/expenses_controller.rb
Mohamad-Mortada: We should keep upload method to be just API as it is the only method officially mentioned here: https://github.com/hackclub/hcb/blob/c622c4ab74fc856237c3e730bbee09819ee08960/app/models/receipt.rb#L105-L129

Thank you for pointing the others out I'll make a PR to change them

app/views/api/v4/reimbursement/expenses/_expense.json.jbuilder
Mohamad-Mortada: .to_f is necessary as since we store the expense.value as a BigDecimal which shows as string in JSON but by converting to Float it becomes a number (this is basically a difference in how JBuilder treats these 2 types differently) :)

app/controllers/api/v4/reimbursement/reports_controller.rb
Mohamad-Mortada: Nope :) good catch

app/controllers/api/v4/reimbursement/reports_controller.rb
Mohamad-Mortada: ooooooo I didn't know about this, yea I agree lets add a param to show hidden reports which defaults to false. Thanks :)

app/controllers/api/v4/reimbursement/expenses_controller.rb
Mohamad-Mortada: https://github.com/hackclub/hcb/pull/13797

app/views/api/v4/reimbursement/expenses/_expense.json.jbuilder
```diff
@@ -0,0 +1,20 @@
...
+  json.status expense.aasm_state
+  json.approved_at expense.approved_at
+
+  expand_association(json, :report, expense.report, partial: "api/v4/reimbursement/reports/report", as: :report)
```
manuthecoder, line 15: `expense.report` is a n+1

app/controllers/api/v4/reimbursement/expenses_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+        end
+
+        private
+
+        def set_expense
+          @expense = ::Reimbursement::Expense.find_by_public_id!(params[:id])
+        end
+
+        def expense_attrs
+          params.require(:reimbursement_expense).permit(:memo, :description, :category, :type, :value)
```
polypixeldev: Small nit but currently using `require` here means it doesn't allow you to create an empty expense by passing in nothing, you'd have to pass in at least
```json
{ "reimbursement_expense": { "memo": "" } }
```

polypixeldev (commented):
just tested each route locally, looks good!

app/controllers/api/v4/reimbursement/expenses_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+      class ExpensesController < ApplicationController
+        before_action :set_expense, only: [:show, :update, :destroy]
+
+        include ApplicationHelper
+
+        def index
+          @report = ::Reimbursement::Report.find_by_public_id!(params[:report_id])
+          authorize @report, :show?
+          expenses = @report.expenses.order(:expense_number)
+          @expenses = paginate_cursor(expenses.to_a, &:public_id)
```
garyhtou, line 13: Let's see if we can avoid calling `to_a`. That loads all records into Ruby - see if we can completely handle pagination in the DB

app/controllers/api/v4/reimbursement/expenses_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+          @report = ::Reimbursement::Report.find_by_public_id!(params[:report_id])
+          @expense = @report.expenses.build(expense_attrs)
+
+          authorize @expense
+
+          @expense.save!
+
+          if params[:receipt].present?
+            ::ReceiptService::Create.new(
+              uploader: current_user,
```
garyhtou, line 30: Long term thinking: Do we want to create a File upload resource/endpoint? similar to Stripe and Column

app/controllers/api/v4/reimbursement/expenses_controller.rb
garyhtou: +1 on this

app/controllers/api/v4/reimbursement/reports_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+        def index
+          if params[:organization_id]
+            @event = Event.find_by_public_id!(params[:organization_id])
+            authorize @event, :show_in_v4?
+            reports = @event.reimbursement_reports
+            reports = reports.visible unless params[:show_hidden_reports] == "true"
+            reports = reports.order(created_at: :desc)
+          else
+            skip_authorization
+            reports = current_user.reimbursement_reports.order(created_at: :desc)
```
garyhtou: We should consider putting this under a `params[:user_id]` rather than just assuming they want the current user.

app/controllers/api/v4/reimbursement/reports_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+
+        def create
+          @event = Event.find_by_public_id!(params.dig(:reimbursement_report, :organization_id))
+
+          report_params = params.require(:reimbursement_report).permit(
+            :name,
+            :invite_message,
+            :maximum_amount_cents
+          )
+
```
garyhtou: Same comment from @polypixeldev above regarding required params

app/controllers/api/v4/reimbursement/reports_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+          @report = @event.reimbursement_reports.build(
+            report_params.merge(
+              user: current_user,
+              inviter: current_user,
+              currency: current_user.payout_method&.currency || "USD"
+            )
```
garyhtou, line 47:
```suggestion
          reimbursee = current_user
          @report = @event.reimbursement_reports.build(
            report_params.merge(
              user: reimbursee,
              inviter: current_user,
              currency: reimbursee.payout_method&.currency || "USD"
            )
```

app/controllers/api/v4/reimbursement/reports_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+            :name,
+            :maximum_amount_cents,
+            :organization_id,
+            :reviewer_id
+          )
+
+          if update_params[:organization_id]
+            @report.event = Event.find_by_public_id!(update_params.delete(:organization_id))
+            authorize @report, :change_event?
+          end
```
garyhtou, line 70: We need to double check security here — i remember the UI controller needed very thorough review

app/controllers/api/v4/receipts_controller.rb
```diff
@@ -4,14 +4,18 @@ module Api
...
       def index
-        if params[:transaction_id].present?
-          @hcb_code = HcbCode.find_by_public_id(params[:transaction_id])
-          authorize @hcb_code, :show?
-          @receipts = @hcb_code.receipts.includes(:user)
-        else
-          skip_authorization
-          @receipts = Receipt.in_receipt_bin.includes(:user).where(user: current_user)
-        end
+        @receipts = if params[:transaction_id].present?
```
garyhtou: Only take in a `receiptable_id`.  Use https://github.com/hackclub/hcb/pull/12912/changes#diff-c282b4372a547e0aad4639b584c05a5042ad33d4a9e52bb83f2906213e2d0b2a to look up. Then, when a user id is passed in to receiptable_id, return receipts in their receipt bin.

app/policies/reimbursement/expense_policy.rb
```diff
@@ -2,6 +2,10 @@
 
 module Reimbursement
   class ExpensePolicy < ApplicationPolicy
+    def show?
+      auditor || creator || reader
```
garyhtou:
```suggestion
      Reimbursement::ReportPolicy.new(record.report, user).show?
```

app/policies/reimbursement/report_policy.rb
```diff
@@ -63,6 +63,14 @@ def convert_to_wise_transfer?
       admin && !record.event.financially_frozen?
     end
 
+    def set_reviewer?
+      admin || manager
+    end
+
+    def set_maximum_amount?
+      (admin || manager) && (record.draft? || record.submitted?)
+    end
```
garyhtou, line 72: Double-check for security in the case that the report's event changes in the same request.

garyhtou: https://github.com/hackclub/hcb-private/pull/127

app/views/api/v4/reimbursement/expenses/_expense.json.jbuilder
```diff
@@ -0,0 +1,20 @@
...
+# locals: (json:, expense:)
+
+object_shape(json, expense) do
+  json.memo expense.memo
+  json.description expense.description
+  json.category expense.category
+  if expense.is_mileage?
+    json.expense_type "mileage"
+  else
+    json.expense_type expense.is_fee? ? "fee" : "standard"
```
garyhtou: Make this a method on expense. maybe `expense_type_name`

app/views/api/v4/reimbursement/expenses/_expense.json.jbuilder
garyhtou: I don't think it is because there is just one report - there aren't multiple to loop over

garyhtou: Maybe that is the case for indexing Expense. In that case, we should just preload Report into ExpenseController

app/views/api/v4/reimbursement/reports/_report.json.jbuilder
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+# locals: (json:, report:)
+
+object_shape(json, report) do
+  json.name report.name
+  json.status report.aasm_state
```
garyhtou, line 7: Our aasm state are confusing to normal users - i think we need to give them different states and do some sort of translation

PR #13715, opened 2026-05-11
Preserve sessions when locking users

polypixeldev:
## Summary of the problem

In the past, we used to have soft deletion on user sessions. However, last year we switched to a new system without soft deletion - see #8950. In that PR, we forgot to update the `lock!` method, which still destroyed sessions instead of marking them as expired and signed out.

## Describe your changes

Mark sessions has expired and signed out instead of destroying when locking a user.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -374,7 +374,7 @@ class User < ApplicationRecord
     update!(locked_at: Time.now)
 
     # Invalidate all sessions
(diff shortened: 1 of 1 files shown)


Luke-Oldenburg (approved):
lgtm

garyhtou (commented):
good find 😭

PR #13729, opened 2026-05-13
[Reimbursements] Add model validation messages

Mohamad-Mortada:
## Summary of the problem
Before we add in v4 API support it would be nice to have validation messages which can be returned if the user tries to submit a report via the API

## Describe your changes
added validation messages


YodaLightsabr (commented):
Could you explore handling AASM::InvalidTransition errors instead of removing the guard clause? It would require extra work when we're consuming errors but it might be cleaner on the model side.

app/models/reimbursement/report.rb
```diff
@@ -422,6 +420,26 @@ def convert_to_wise_transfer!(as: User.system_user)
       end
     end
 
+    def may_mark_submitted?
+      submission_requirements.empty?
```
manuthecoder:
```suggestion
      (draft? || reimbursement_requested?) && submission_requirements.empty?
```

app/models/reimbursement/report.rb
manuthecoder, line 461: this is no longer used

manuthecoder (changes requested):

app/models/reimbursement/report.rb
Mohamad-Mortada: fixed!

app/models/reimbursement/report.rb
Mohamad-Mortada: no need

YodaLightsabr (approved):
Nice work!

manuthecoder (approved):
awesome

app/models/reimbursement/report.rb
```diff
@@ -422,6 +416,26 @@ def convert_to_wise_transfer!(as: User.system_user)
       end
     end
 
+    def may_mark_submitted?
```
garyhtou: `may_mark_submitted?` is already a complicated method defined by AASM with a bunch of checks. I think it's too risky for us to override this method.

Let's give it another name. Maybe `submission_requirements_met?`

AASM source code:
image
image
image
image

app/models/reimbursement/report.rb
```diff
@@ -422,6 +416,26 @@ def convert_to_wise_transfer!(as: User.system_user)
...
+      messages << "One or more expenses are missing receipts" if missing_receipts?
+      messages << "One or more expenses have a zero amount" if expenses.any? { |e| e.amount.zero? }
+      messages << "Report amount exceeds the maximum allowed" if exceeds_maximum_amount?
+      messages << "Report amount is below the minimum required" if event.present? && below_minimum_amount?
+      messages << "Currency does not match your payout method" if mismatched_currency?
+      messages << "Your payout method is not set up" unless user.payout_method.present?
+      messages << "Your payout method is not supported" if !payout_method_allowed?
+      messages << "Organization finances are currently frozen" if event&.financially_frozen?
+      messages
+    end
```
garyhtou: See if you can use ActiveRecord errors here. Here's an example from Gemini:

image

app/controllers/reimbursement/reports_controller.rb
```diff
@@ -208,8 +208,8 @@ def submit
         if @report.user.can_update_payout_method?
           flash[:success][:link_text] = "If needed, you can still edit your payout settings."
         end
-      rescue => e
-        flash[:error] = e.message
+      rescue AASM::InvalidTransition
+        flash[:error] = @report.submission_requirements.to_sentence
```
garyhtou: If you end up using Active Record errors, you'll want to use `report.errors.full_message.to_sentence` here.

app/models/reimbursement/report.rb
garyhtou: When doing this, there's a chance you can just define one method that's used by the guard clause. And if you want to check if you can submit, then you check whether `@report.errors.empty?` or `@report.valid?`

PR #13732, opened 2026-05-13
Added birthday validations and age messages

### LINKED ISSUE #11912: [User] Deter fake ages

### LINKED ISSUE #11958: [User] Disallow age < 13
Add validation on User when birthday changes

Make sure it doesn't break existing users

Acidicts:
## Summary of the problem
closes https://github.com/hackclub/hcb/issues/11958
closes https://github.com/hackclub/hcb/issues/11912

## Describe your changes
Added age validations and age message for:
- <13 : "Sorry, you must be 13 years old to sign up "

image


app/javascript/controllers/birthday_controller.js
```diff
@@ -0,0 +1,71 @@
...
+    const selects = this.element.querySelectorAll('select[name*="birthday"]')
+    let year, month, day
+
+    selects.forEach(select => {
+      const name = select.name
+      const value = select.value
+      if (name.includes('(1i)')) year = value
+      if (name.includes('(2i)')) month = value
+      if (name.includes('(3i)')) day = value
+    })
```
garyhtou: I believe the best practice is to use Stimulus targets. This will prevent you from needing to check the name before assigning year, month, and day.

app/javascript/controllers/birthday_controller.js
```diff
@@ -0,0 +1,71 @@
...
+
+    const element = document.getElementById('birthday_deterence')
```
garyhtou: Make this a stimulus target too.

app/javascript/controllers/birthday_controller.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    if (age < 13) {
+      element.textContent = `Sorry, you must be at least 13 years old to sign up.`
+      element.classList.add('warning')
+      document
+        .getElementById('user_submit')
+        ?.setAttribute('disabled', 'disabled')
+    } else if (age >= 13 && age <= 18) {
+      element.textContent = `You should join Hack Club`
```
garyhtou: We only want to show a message when absolutely necessary, so let's drop this case.

app/javascript/controllers/birthday_controller.js
```diff
@@ -0,0 +1,71 @@
...
+    } else if (125 > age && age > 100) {
+      element.textContent = `Nice to see you found us!`
+    } else if (age > 125) {
+      element.textContent = `Congrats!`
```
garyhtou: In the PR description, you mentioned it's not possible to set an age > 100, so let's drop these unnecessary cases.

app/models/user.rb
```diff
@@ -211,6 +211,8 @@ class User < ApplicationRecord
 
   validates :preferred_name, length: { maximum: 30 }
 
+  validate :age_in_range?, if: -> { birthday.present? && (new_record? || will_save_change_to_birthday?) }
```
garyhtou: The `birthday.present?` is unnecessary because you check that inside `age_in_range?`.

garyhtou: Is `new_record?` necessary?, or does the combination of `birthday.present?` and `will_save_change_to_birthday?` cover all the cases we need?

garyhtou (changes requested):

app/javascript/controllers/birthday_controller.js
```diff
@@ -0,0 +1,75 @@
...
+
+    if (age < 13) {
+      this.showMessage(
+        'Sorry, you must be at least 13 years old to sign up.',
+        true
+      )
+      this.disableSubmit()
+    } else {
+      this.clearMessage()
+      this.showMessage('Test')
```
manuthecoder: delete this

app/views/users/edit.html.erb
```diff
@@ -66,14 +66,19 @@
...
+              <div class="gap-2 onboarding" data-controller="birthday" style="display: flex; flex-direction: column;">
+                <span style="display: flex; gap: 0.5rem;">
+                <%= select_month @user.birthday,
+                                       { field_name: :birthday, prefix: @user.model_name.param_key, prompt: @user.birthday.nil? },
+                                       { required: true, data: { birthday_target: "month", action: "change->birthday#updateAge" } } %>
+                <%= select_day @user.birthday,
+                                       { field_name: :birthday, prefix: @user.model_name.param_key, prompt: @user.birthday.nil? },
+                                       { required: true, data: { birthday_target: "day", action: "change->birthday#updateAge" } } %>
+                <%= select_year @user.birthday,
+                                       { field_name: :birthday, prefix: @user.model_name.param_key, prompt: @user.birthday.nil?, start_year: Date.today.year, end_year: Date.today.year - 100 },
```
manuthecoder: All three birthday selects share the same name, so only the year is submitted

app/models/user.rb
```diff
@@ -211,6 +211,8 @@ class User < ApplicationRecord
 
   validates :preferred_name, length: { maximum: 30 }
 
+  validate :age_in_range?, if: :will_save_change_to_birthday?
```
manuthecoder:
```suggestion
  validate :age_in_range?, if: -> { new_record? || will_save_change_to_birthday? }
```

app/assets/stylesheets/components/_forms.scss
```diff
@@ -231,14 +231,33 @@ input[type='search']::-webkit-search-decoration {
...
+    box-sizing: border-box;
+  }
+
+  > input,
+  > select,
+  > textarea {
+    width: 100%;
+    min-width: 0;
   }
 }
```
manuthecoder, line 243: Editing `field_with_errors` affects other parts of HCB as well

app/models/user.rb
```diff
@@ -453,6 +455,16 @@ def age
     age_on(Date.current)
   end
 
+  def age_in_range?
+    return unless birthday.present?
+
+    if age.nil? || age < 13
```
manuthecoder: age.nil? is unreachable

app/javascript/controllers/birthday_controller.js
```diff
@@ -0,0 +1,75 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
+  static targets = ['year', 'month', 'day', 'message']
+
+  connect() {
+    this.updateAge()
+    this.element.classList.add('birthday-inline')
```
manuthecoder: `birthday-inline` class is added but never defined

manuthecoder (changes requested):
Add some tests for this as well

app/javascript/controllers/birthday_controller.js
Acidicts: Removed code which used the non-existent class

app/models/user.rb
Acidicts: removed age.nil? from the if statement

app/assets/stylesheets/components/_forms.scss
Acidicts: Changed this to use current main branch's version

app/models/user.rb
Acidicts: Implemented

app/javascript/controllers/birthday_controller.js
Acidicts: Removed test case

app/models/user.rb
Acidicts: Removed

app/javascript/controllers/birthday_controller.js
Acidicts: Removed

app/javascript/controllers/birthday_controller.js
Acidicts: Also removed

PR #13734, opened 2026-05-14
[CardGrant] Prevent conflicting allow/block restrictions

### LINKED ISSUE #13681: [CardGrant] Add validation to prevent conflicting allow and block lists
You could have a category or merchant ID allowed and blocked at the same time.

manuthecoder:
Prevents users from allowing and blocking merchants/categories at the same time for card grants

closes #13681


app/models/concerns/has_grant_conflict_restrictions.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  included do
+    validate :no_conflicting_merchant_lists
+    validate :no_conflicting_category_lists
+  end
+
+  private
+
+  def no_conflicting_merchant_lists
+    conflicts = merchant_lock & banned_merchants
```
garyhtou: Use the methods 
https://github.com/hackclub/hcb/blob/7a84a10c3d84795c495cc9c5eb6fcc9c65a1d313/app/models/card_grant.rb#L282-L288
and
https://github.com/hackclub/hcb/blob/7a84a10c3d84795c495cc9c5eb6fcc9c65a1d313/app/models/card_grant.rb#L294-L300
to capture ALL allowed and disallowed merchants/categories

app/models/concerns/has_grant_conflict_restrictions.rb
```diff
@@ -0,0 +1,23 @@
...
+    errors.add(:base, "Merchant(s) #{conflicts.join(', ')} cannot be both allowed and blocked") if conflicts.any?
+  end
+
+  def no_conflicting_category_lists
+    conflicts = category_lock & banned_categories
+    category_label = conflicts.one? ? "Category" : "Categories"
```
garyhtou: Use pluralize instead

app/models/concerns/has_grant_conflict_restrictions.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module HasGrantConflictRestrictions
```
garyhtou: This concern is a bit too narrow since we're having a hard time naming this. I would recommend making this concern `HasGrantRestrictions` and moving

https://github.com/hackclub/hcb/blob/9588529272e3c9a651b473470840282fe89b3b5f/app/models/card_grant_setting.rb#L37-L43
_^ side note: is it missing a `disallowed_categories`?_

into the concern.

Card Grant would then need to override those methods (the current defintions should work fine).

spec/models/concerns/has_grant_restrictions_spec.rb
garyhtou, line 1: This should test for the case where a CardGrant conflict is it's CardGrantSetting. This test case may need to end up living in Card Grant's spec

PR #13749, opened 2026-05-15
[AASM] Make `metrics` table's `aasm_state` column not null

rluodev:
## Summary of the problem

Previously, was nullable

## Describe your changes

Now, not nullable. 617,592 null entries in prod db currently: https://hcb.hackclub.com/blazer/queries/1207-metrics-table-of-null-aasm_state-s - **a command needs to be run to make these entries not null before merge**


rluodev:
@garyhtou command needs to be run to set existing metrics entries to a non-null aasm_state value in prod

manuthecoder (approved):

garyhtou (approved):

garyhtou:
I'm approving to allow merging due to codeowners. @Luke-Oldenburg will do a deeper review

Luke-Oldenburg:
Confirmed via Blazer on 6/22/2026 that there are no null entries in this column on prod.

PR #13755, opened 2026-05-15
[AASM] Make `transaction_csvs` table's `aasm_state` column not null

rluodev:
## Summary of the problem

Previously, was nullable

## Describe your changes

Now, not nullable. Confirmed via Blazer on 5/15 that there are no null entries in this column on prod.


manuthecoder (approved):
LGTM!

garyhtou (approved):

garyhtou:
I'm approving to allow merging due to codeowners. @Luke-Oldenburg will do a deeper review

Luke-Oldenburg:
Confirmed via Blazer on 6/22/2026 that there are no null entries in this column on prod.

PR #13760, opened 2026-05-19
[Contracts] Add more links for HCB to sign a contract

polypixeldev:
## Summary of the problem

Right now, the only way for HCB admins to find a link to sign a contract for an OPI is by seeing the email in Intercom or by searching in the admin contracts page.

## Describe your changes

Add two links - one in the context menu on the OPI, and another in the event nav (similar to the activate button).

image
image

### DIFF
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -17,6 +17,17 @@ module EventsHelper
       }
     end
 
+    if event.financially_frozen? && (contracts = event.contracts.select { |c| c.parties.not_hcb.all?(&:signed?) }).one? && event.contracts.signed.none?
+      items << {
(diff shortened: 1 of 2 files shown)


app/helpers/events_helper.rb
```diff
@@ -17,6 +17,17 @@ def events_nav(event = @event, selected: nil)
       }
     end
 
+    if event.financially_frozen? && (contracts = event.contracts.select { |c| c.parties.not_hcb.all?(&:signed?) }).one? && event.contracts.signed.none?
```
YodaLightsabr, line 20: Good thinking putting `event.financially_frozen?` first

YodaLightsabr (approved):
LGTM

PR #13761, opened 2026-05-19
[Discord] Use more constants 

polypixeldev:
## Summary of the problem

The Discord code has a lot of magic numbers :(

## Describe your changes

Use Discordrb constants where possible, add EPHEMERAL_MESSAGE_FLAG, and comment the others.


app/jobs/discord/handle_interaction_job.rb
```diff
@@ -291,14 +291,14 @@ def missing_receipts_command
     def require_linked_user
       return respond content: "This command requires you to link this Discord account to HCB", components: button_to("Set up HCB", "setup") if @responded
 
-      respond content: "This command requires you to link your Discord account to HCB", embeds: linking_embed, flags: 1 << 6
+      respond content: "This command requires you to link your Discord account to HCB", embeds: linking_embed, flags: EPHEMERAL_MESSAGE_FLAGflag
```
YodaLightsabr: Is the extra `flag` in `EPHEMERAL_MESSAGE_FLAGflag` intentional?

app/jobs/discord/handle_interaction_job.rb
```diff
@@ -345,10 +345,6 @@ def respond(**body)
       MSG
     end
 
-    def can_manage_guild?
-      @permissions & 0x0000000000000020 == 0x0000000000000020
-    end
-
```
YodaLightsabr, line 351: Thank you for finding a better solution than this!

YodaLightsabr (commented):
One quick comment. Good to merge after that. Please also double check in the console that all of the replaced values evaluate to the same number.

app/jobs/discord/handle_interaction_job.rb
polypixeldev: Nope, thanks for catching that!

YodaLightsabr (approved):

PR #13766, opened 2026-05-20
Manually include non-v4 helpers in ApplicationHelper

polypixeldev:
## Summary of the problem

We recently ran into an interesting bug where a method defined in `Api::V4::ApplicationHelper` overrode a method defined in `ApplicationHelper`. Unexpectedly, Rails automatically includes _all_ helpers, even those scoped under a different module.

This is not intuitive - v4 helpers should only be available in v4.

## Describe your changes

Sets the Rails configuration value for including all helpers to false, and manually includes non-v4 helpers in ApplicationHelper. Tested both the website and v4 in development and everything works, but would appreciate an additional sanity check here. I also deleted some empty helpers (these are generated by Rails when you generate a new controller).

I also think that not every helper needs to be included in `ApplicationHelper` - it would make sense if `DonationsHelper` is only available to the views of `DonationsController`. However, this requires more discussion and can be a later PR if we decide to implement it.

### DIFF
--- a/app/helpers/ach_transfers_helper.rb
+++ b/app/helpers/ach_transfers_helper.rb
@@ -1,4 +1,0 @@
-# frozen_string_literal: true
-
-module AchTransfersHelper
-end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -3,6 +3,27 @@
 module ApplicationHelper
   include ActionView::Helpers
 
+  include DonationsHelper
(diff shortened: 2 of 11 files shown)


Mohamad-Mortada (approved):
Thank you for this!

manuthecoder (approved):
love how this makes the code more DRY! great job

garyhtou (approved):

PR #13770, opened 2026-05-20
[v4] Shallow route comments

Mohamad-Mortada:
## Summary of the problem
A URL like `/api/v4/organizations/org_BG3uZB/transactions/txn_66HJYE/comments` is insanelyyyyyyy long. Let's shallow route so we can get a nice short URL like `/api/v4/comments?transaction_id=txn_66HJYE` 

## Describe your changes
Shallow routes comments while keeping the original nested route for backwards compatability with old clients as we update them in the near future


manuthecoder (approved):

Luke-Oldenburg (approved):
lgtm

PR #13773, opened 2026-05-21
[v4] Use donation model create_stripe_payment_intent instead

Mohamad-Mortada:
## Summary of the problem
The problem is that when we create a payment intent using our current implementation we don't attach the payment intent ID to the donation object.

## Describe your changes
Uses model's create_stripe_payment_intent function instead


app/controllers/api/v4/donations_controller.rb
garyhtou:
```suggestion
      before_action :require_trusted_oauth_app!, only: [:create, :payment_intent]
```

garyhtou (approved):

PR #13781, opened 2026-05-22
Fix flaky GSuiteMailer#notify_of_configuring "renders to" spec

garyhtou:
created by Claude

## Summary of the problem

The spec compared `mailer.to` against
`organizer_positions.where(role: :manager)...map(&:email)` using the
order-sensitive `eql` matcher. That query has no `ORDER BY`, so Postgres
makes no guarantee about row order — the mailer's recipient order and the
spec's freshly-run query can diverge, failing `eql` even though both
contain the same recipients.

## Describe your changes

Use `match_array` so the comparison is order-independent.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -38,7 +38,7 @@ class ApplicationMailer < ActionMailer::Base
     # to `User#id == 1`, and any test user that happens to land at that id
     # ends up memoized into this list, silently dropping that user's email
     # out of every subsequent mailer's recipients for the rest of the
-    # process (the flake fixed by the same PR that introduced this guard).
+    # process.
     return [] unless Rails.env.production?
 
     @earmuffed_recipients ||= EARMUFFED_USER_IDS.filter_map do |id|
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -31,6 +31,16 @@ class ApplicationMailer < ActionMailer::Base
   ].freeze
 
   def self.earmuffed_recipients
+    # Earmuffing exists to suppress mail to specific real production users
+    # (referenced by their production public_ids). In test and development
+    # no real mail is delivered, so the filter has no job to do — and
+    # running it in test is actively harmful: `usr_b9YtZb` hashid-decodes
+    # to `User#id == 1`, and any test user that happens to land at that id
+    # ends up memoized into this list, silently dropping that user's email
+    # out of every subsequent mailer's recipients for the rest of the
+    # process (the flake fixed by the same PR that introduced this guard).
+    return [] unless Rails.env.production?
+
     @earmuffed_recipients ||= EARMUFFED_USER_IDS.filter_map do |id|
       User.find_by_public_id(id)&.email
     end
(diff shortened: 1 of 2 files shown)


polypixeldev (approved):

polypixeldev:
While we should be using `match_array` here, I'm not sure if this actually fixes the issue. [Logs](https://github.com/hackclub/hcb/actions/runs/26304902507/job/77439073105?pr=13780) from when this CI fails include:
```

       expected: ["patricia@lynch.example", "alfonso_langosh@reichel.example", "inez@fadel.example"]
            got: ["alfonso_langosh@reichel.example", "inez@fadel.example"]
```
which suggests that the `mailer.to` is missing an element (the first), not just in the wrong order. However, I can't reproduce this issue either way.

garyhtou:
Response from Claude:

> You're right that `match_array` alone didn't address what's actually happening — thanks for the careful read. I dug in further and reproduced the drop locally; sharing the root cause:
> 
> `ApplicationMailer#mail` subtracts `ApplicationMailer.earmuffed_recipients` from the message's `to`. That list is class-memoized and built from `EARMUFFED_USER_IDS`. Decoding those hashids:
> 
> - `usr_b9YtZb` (Zach) → **`User#id == 1`**
> - `usr_b6mtLG` (Christina) → 904
> - `usr_N4tk5d` (Rachel A personal) → 15139
> - `usr_ZBt5g5` (Rachel A Hack Club) → 14967
> 
> In a fresh CI shard's DB, the first user created lands at `id == 1`. If that's one of this spec's three managers (which is exactly what happens when the g_suite_mailer spec is among the first specs in a shard to create a user), the mailer's earmuff filter strips that manager's email out of `msg.to`. That's the "got 2 / expected 3" pattern, and it explains why the dropped recipient was always the *first*-inserted manager across all three observed failures.
> 
> Reproduced locally by truncating `users` and creating the g_suite cold:
> 
> ```
> manager user_ids: [1, 3, 5]
> manager emails:   ["mohammad@dibbert.example", "terrance@quitzon.test", "denis_anderson@lind.example"]
> earmuffed_recipients (after mail): ["mohammad@dibbert.example"]
> mailer.to: ["terrance@quitzon.test", "denis_anderson@lind.example"]
> ```
> 
> Force-pushed an updated fix that stubs `ApplicationMailer.earmuffed_recipients` to `[]` at the file level (so the test isn't entangled with whoever happens to hold id 1) and keeps `match_array` so the underlying unordered query can't reorder-flake the comparison either. Verified by reproducing the CI condition (`id=1` as a manager) in a regression spec and by looping the full file across 20 seeds.

garyhtou:
> *Written by Claude*
>
> Following up on this — agree with you that the per-spec stub was treating a symptom, not the cause. Pushed `bdfce1804` (with a small comment cleanup in `4e172df0e`) which moves the fix into `ApplicationMailer.earmuffed_recipients` itself: it now returns `[]` outside of production, so the `User.find_by_public_id` lookup never runs in test or development (where there's no real mail being delivered for the filter to protect). That closes the test-suite booby-trap at the source — any future mailer spec is protected automatically — and the per-file stub in `g_suite_mailer_spec.rb` is dropped as redundant in the same commit. `application_mailer_spec.rb` keeps working unchanged because it stubs `earmuffed_recipients` wholesale to exercise the filter logic.

PR #13785, opened 2026-05-23
[Contracts] Notify HCB when it's their turn to sign a reissued contract

### LINKED ISSUE #13719: [Admin] Notify HCB signer when contract reissue completed
### Describe the problem

Currently, when a contract is reissued and completed again, I have no way of knowing it is completed without manually checking through the applications for organizations a that contract was resent to. If possible, it would be nice if I could get an email when a reissued contract is completed.

_No response_

polypixeldev:
## Summary of the problem

Currently, we don't notify HCB when they have to sign if the contract is for a teen-led application. This is because HCB only signs the contract after reviewing the application, which they do on their own schedule.

However, there's the case where HCB reviews an application but has to reissue the contract - even if it's for a teen-led application, the person who reissued the contract should be notified when they have to sign.

## Describe your changes

We currently have no way in the database to determine if a contract is a reissue of another contract. This PR adds a reference to contract to the contract it is a reissue of, and uses that data to correctly determine whether or not HCB should be notified when it's their turn to sign.


app/models/contract.rb
```diff
@@ -216,6 +223,14 @@ def create_document!
     update!(document:)
   end
 
+  def is_reissue?
+    reissue_of_id.present?
+  end
+
+  def reissued?
+    reissued_contract.present?
+  end
```
manuthecoder: dead code

app/models/contract.rb
```diff
@@ -216,6 +223,14 @@ def create_document!
     update!(document:)
   end
 
+  def is_reissue?
```
manuthecoder: rename to `reissue?` better convention

manuthecoder (commented):
almost there!

garyhtou (approved):

PR #13786, opened 2026-05-23
[User::Session] Fix IP deletion

polypixeldev:
## Summary of the problem

IP address needs to be cleared along with lat/long after 18 months.

### DIFF
--- a/app/models/user/session.rb
+++ b/app/models/user/session.rb
@@ -38,7 +38,7 @@
 #
 class User
   class Session < ApplicationRecord
-    has_paper_trail skip: [:session_token] # ciphertext columns will still be tracked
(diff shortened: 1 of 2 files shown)


YodaLightsabr (approved):
LGTM

app/models/user/session.rb
```diff
@@ -128,9 +128,9 @@ def sudo_mode?
       last_authenticated_at >= SUDO_MODE_TTL.ago
     end
 
-    def clear_metadata!
+    def clear_ip_metadata!
       update!(
-        device_info: nil,
```
garyhtou, line 133: should we backfill device info then?

polypixeldev: Yes, I also think we should scrub PaperTrail of the columns for ip/lat/long.

PR #13793, opened 2026-05-24
[IncreaseCheck] Improve mailers

### LINKED ISSUE #13686: [IncreaseCheck] Notify recipient when their check has been stopped

Luke-Oldenburg:
## Summary of the problem

closes #13686

## Describe your changes

Improves various copies and adds a mailer for when checks are stopped.

### DIFF
--- a/app/views/increase_check_mailer/notify_stopped.html.erb
+++ b/app/views/increase_check_mailer/notify_stopped.html.erb
@@ -9,7 +9,7 @@
 </p>
 
 <p>
-  Check #<%= @check.check_number %> is no longer valid and can be disposed of safely.
+  Check #<%= @check.check_number %> is no longer valid and cannot be deposited. You may still receive it in the mail; however, this check can be disposed of safely.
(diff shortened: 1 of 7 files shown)


app/views/increase_check_mailer/notify_stopped.html.erb
```diff
@@ -0,0 +1,25 @@
...
+</p>
+
+<p>
+  <%= @check.event.name %> has just stopped a
+  <%= render_money(@check.amount) %> check sent to you for
+  <i>"<%= @check.payment_for %>"</i>.
+</p>
+
+<p>
+  Check #<%= @check.check_number %> is no longer valid and can be disposed of safely.
```
garyhtou:
```suggestion
  Check #<%= @check.check_number %> is no longer valid and cannot be deposited. You may still receive it in the mail; however, this check can be disposed of safely.
```

garyhtou (approved):

PR #13798, opened 2026-05-25
[My Cards] Improve experience

manu-gurudath:
This PR makes a few improvements to the `/my/cards` page 
1. Adds a search bar so users can search cards by last 4 digits, name, or purpose
3. Groups cards by event
4. Moves filter button from top to the search bar

| Before | After |
|--------|--------|
| Confusing cards, everywhere | Grouped neatly by organization |
| image | image |


app/views/my/cards.html.erb
```diff
@@ -9,48 +9,78 @@
...
+        </article>
+        <% if cards.count > preview_count %>
+          <div class="flex">
+            <%= link_to event_cards_overview_path(event_id: event.slug, user: current_user.slug), class: "btn ml-auto" do %>
+              +<%= cards.count - preview_count %> more
+            <% end %>
+          </div>
+        <% end %>
+      </section>
+    <% end %>
```
Luke-Oldenburg, line 83: Does this reduce the amount of cards visible on this page? I think we should still make these accessible without navigating to the event. It is possible to have a canceled card from a non-transparent event you no longer have access to.

Luke-Oldenburg (changes requested):

PR #13799, opened 2026-05-25
[Cards] Improve grid view usability

manu-gurudath:
#13798, but for events

Only differences are that cards are grouped by user, and clicking on "more" shows all the cards from that user
image


Luke-Oldenburg:
Please attach a screenshot.

app/controllers/events_controller.rb
```diff
@@ -463,19 +463,34 @@ def feed
   def card_overview
     @status = %w[active inactive frozen canceled].include?(params[:status]) ? params[:status] : nil
     @type = %w[virtual physical].include?(params[:type]) ? params[:type] : nil
+    @q = params[:q].presence
+
+    @filter_options = [
+      { key: "status", label: "Status", type: "select", options: [["Active", "active"], ["Inactive", "inactive"], ["Frozen", "frozen"], ["Canceled", "canceled"]], include_blank: "All" },
+      { key: "type", label: "Type", type: "select", options: [["Virtual", "virtual"], ["Physical", "physical"]], include_blank: "All" },
+      { key: "user", label: "Cardholder", type: "select", options: @event.users.order(:full_name).map { |u| [u.full_name, u.slug] }, include_blank: "All" },
```
Luke-Oldenburg, line 471: This should derive from `event.users` or users that have issues non-card grant cards from the event. Some cardholders may be removed.

app/views/events/card_overview.html.erb
```diff
@@ -49,6 +57,39 @@
...
+            <% end %>
+          </section>
+        <% end %>
+        <%= paginate @paginated_user_groups %>
+      <% else %>
+        <% if organizer_signed_in? %>
+          <%= blankslate "You don’t have any #{@status} cards yet" %>
+        <% else %>
+          <%= blankslate "This organization doesn’t have any #{@status} cards" %>
+        <% end %>
```
Luke-Oldenburg, line 91: `"#{@event.name} doesn't have any #{@status} cards"` and drop the if.

Luke-Oldenburg (changes requested):

PR #13800, opened 2026-05-26
Fix Ahoy user for ops_call_requested email

polypixeldev:
## Summary of the problem

Ahoy stores the user the email was sent to, and it infers it by looking to see if there's an @user variable defined - which in the ops email mailer, we set to the person who requested the call. This results in /admin/emails showing that the email was sent to the person who requested the call instead of ops.

## Describe your changes

Renamed @user for that email to @requesting_user, tested in dev. Also looked to see if we set @user anywhere else that we don't actually send the email to that user and didn't find any.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -1079,7 +1079,7 @@ class EventsController < ApplicationController
   def request_call
     authorize @event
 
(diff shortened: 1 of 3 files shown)


garyhtou (approved):

garyhtou:
thank you!!

PR #13801, opened 2026-05-26
[Maintenance Task] Backfill Stripe Payment Intent Event IDs

Mohamad-Mortada:
Converts this OTJ into maintenance task 

https://github.com/hackclub/hcb/blob/129706cd0a6ed8b972bf3203cef67bca9dab7996/app/jobs/one_time_jobs/backfill_payment_intent_event_ids.rb#L1-L81

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,4 @@ yarn-debug.log*
 
 # AI
 .claude/
+graphify-out/
--- a/.gitignore
+++ b/.gitignore
@@ -59,4 +59,3 @@ yarn-debug.log*
 
 # AI
 .claude/
-graphify-out/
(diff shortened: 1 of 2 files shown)


garyhtou (commented):
Please double-check this backfill works as we want. I remember last time we had to monkeypatch the script. Here are the sessions:
- https://hcb.hackclub.com/console/sessions/1568
- https://hcb.hackclub.com/console/sessions/1569
- https://hcb.hackclub.com/console/sessions/1570

Some of the monkeypatching we did was to make it run faster by pre-fetching. That's not necessary here unless it takes more than say half a day.

And, do we still need to start with donations > March 25th? or can we choose a later date?

garyhtou (approved):

PR #13811, opened 2026-05-29
[User Admin] Turbo frame more elements independently

### LINKED ISSUE #13784: [Admin] Add pagination to admin user details
Admin user details are now in a separate Turbo frame, which makes it so that the entire page doesn't time out immediately, but the frame can still timeout (which replaces the entire page with a "that didn't work" message). Each admin card inside the admin details should be paginated so that we don't have to fetch and render thousands of rows.

Luke-Oldenburg:
## Summary of the problem

closes #13784

## Describe your changes

Separated most records into their own turbo-framed page and added pagination to them. Some records which we can reasonably expect users to have few of are still directly in the admin_details turbo frame and not paginated for simplicity.

Adds `includes` statements to some record 

Also adds missing Rails model relation to make referencing records by `User` easier.

### DIFF
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -93,6 +93,10 @@ class UserPolicy < ApplicationPolicy
     admin_details?
   end
 
+  def admin_details_missing_receipts?
+    admin_details?
+  end
+
   def admin_details_reimbursement_reports?
     admin_details?
   end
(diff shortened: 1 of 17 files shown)


app/policies/user_policy.rb
```diff
@@ -69,8 +69,48 @@ def admin_details?
     user.auditor?
   end
 
+  def admin_details_ach_transfers?
```
garyhtou, line 72: use `alias_method` or `alias`

Image

garyhtou (approved):

PR #13812, opened 2026-05-29
Add OOO callout to application status page

polypixeldev:
## Summary of the problem

We want to make it clear to users what to expect with application review times. Plus, it'd be good to have a single OOO callout partial that we can update whenever we have a special case (such as FIRST worlds).

## Describe your changes

Refactored the OOO callout to be a single partial (without the FIRST worlds code since that's over), and add it to the application status page if the user is awaiting review.

image


app/views/application/_ooo_callout.html.erb
```diff
@@ -1,29 +1,16 @@
...
-      Our team will be at the FIRST Championship in Houston April 28–May 2 — come say hi in the Expo hall!
-      Transfer approvals may take 2–3 business days through May 4.
-    <% elsif upcoming %>
-      From April 28 to May 4, our team will have limited availability.
-      Transfers submitted during that window may take 2–3 business days to approve.
+    <% if financial %>
+      on the <span class="tooltipped" aria-label="<%= 1.business_day.after(Date.today).strftime("%b %e, %Y") %>">next business day</span>.
     <% else %>
-      Our team has limited availability through May 4. Transfer approvals may take 2–3 business days.
+      within <span class="tooltipped" aria-label="<%= 2.business_days.after(Date.today).strftime("%b %e, %Y") %>">2 business days</span>.
```
garyhtou: Handle case for adults (2 weeks)

garyhtou (approved):

PR #13824, opened 2026-05-30
Prevent editing wire after sending - Fixes #13725

### LINKED ISSUE #13725: [Wire] Prevent editing after sending
Wires can currently be edited by ops after being sent, resulting in the potential for HCB records to become misaligned with Column records. We should check prod data to see if this has accidentally happened in the past.

Renjini221:
Fixes #13725

## What's the bug?
Admins could edit a wire transfer even after it had already been sent to Column (approved/deposited state). This could cause HCB records to become misaligned with Column records.

## What's the fix?
Added a `record.pending?` check to `WirePolicy#edit?` and `WirePolicy#update?` so wires can only be edited while still in the pending state.

## Journey
First open source contribution! Fought through:
- WebMock blocking Column API calls
- German IBAN format validation (DE89370400440532013000)
- recipient_country enum format (:DE not 276)
- Balance validation on event
- Flipper stub conflicts

## Changes
- `app/policies/wire_policy.rb` — added state check to edit? and update?
- `spec/factories/wire_factory.rb` — new wire factory for tests
- `spec/policies/wire_policy_spec.rb` — 6 tests covering all wire states

### DIFF
--- a/app/policies/wire_policy.rb
+++ b/app/policies/wire_policy.rb
@@ -22,11 +22,11 @@ class WirePolicy < ApplicationPolicy
   end
 
   def edit?
-    user&.admin?
+    user&.admin? && record.pending?
   end
 
   def update?
-    user&.admin?
+    user&.admin? && record.pending?
   end
 
   private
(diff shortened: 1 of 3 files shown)


manuthecoder (approved):
looks good!

PR #13827, opened 2026-05-31
Fix wire reimbursement minimum

### LINKED ISSUE #13796: [Reimbursement::Report] Allow sending wire reimbursements under $500 with wire minimum disabled
Currently you can only send a regular wire transfer for under $500 when the wire minimum is disabled. We need to add support for this to reimbursements and add an easy way to charge the fee to the organization.

Renjini221:
Fixes #13796

## What's the bug?
When a reimbursement report had a card grant, `minimum_wire_amount_cents` 
hardcoded $500 regardless of whether the event had the wire minimum 
disabled (via Flipper flag or plan exemption). This blocked legitimate 
wire reimbursements under $500 even when the minimum was explicitly 
disabled for that event.

## What's the fix?
Removed the card grant special case — `minimum_wire_amount_cents` now 
always delegates to `event.minimum_wire_amount_cents`, which already 
correctly handles all exemption logic.

## Journey
- card_grant factory internally calls Flipper.enabled?(:frozen) 
- user.stripe_cards undefined on nil user
- Switched to instance_double per Copilot review
- Extra commits issue from branch management

## Changes
- `app/models/reimbursement/report.rb` — removed hardcoded $500 fallback
- `spec/models/reimbursement/report_spec.rb` — 4 tests covering exemption behaviour with and without card grant

### DIFF
--- a/app/models/reimbursement/report.rb
+++ b/app/models/reimbursement/report.rb
@@ -339,11 +339,7 @@ module Reimbursement
       maximum_amount_cents && amount_cents > maximum_amount_cents && currency == "USD"
     end
 
-    def minimum_wire_amount_cents
-      return event.minimum_wire_amount_cents unless card_grant.present?
-
-      500_00
-    end
+    delegate :minimum_wire_amount_cents, to: :event
 
     def below_minimum_amount?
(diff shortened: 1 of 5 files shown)


Luke-Oldenburg:
Please add a checkbox to charge the fee to the organization when sending the wire reimbursement.

PR #13832, opened 2026-05-31
[Admin] Remove PayPal Transfers

### LINKED ISSUE #13820: [Admin] Remove PayPal transfers page
- [x] Page
- [x] Nav Item
- [x] Command Bar Item

manuthecoder:
Closes #13820


Luke-Oldenburg (commented):
Should we remove the page entirely? For example, I can't find an Emburse cards admin page but I'm not sure if that is because it never existed or was removed.

garyhtou (approved):
we can keep the admin page - useful to see this still somewhat recent data

PR #13833, opened 2026-05-31
[Emoji picker] Fix theme color logic

### LINKED ISSUE #13629: [Emoji Picker] Use HCB theme config rather than system preference
Sometimes the emoji picker's theme (light/dark) doesn't match the HCB theme. This is because the emoji picker is always following the system preference, and not the HCB theme config

manuthecoder:
Closes https://github.com/hackclub/hcb/issues/13629


Luke-Oldenburg (approved):
lgtm

PR #13843, opened 2026-06-02
Remove AliPay

mattsoh:
## Summary of the problem

Wise in fact does not [support AliPay for business users](https://wise.com/en-cn/blog/wise-work-in-china-send-money-to-and-from-china#:~:text=One%20thing%20you%20should%20know%20is%20that%20sending%20money%20from%20Wise%20to%20WeChat%20and%20Alipay%20at%20the%20moment%20is%20only%20available%20for%20personal%20Wise%20accounts%20not%20for%20business%20Wise%20accounts.). 

## Describe your changes

Removed AliPay


Luke-Oldenburg:
We will need to clear the settings of any users that currently have this configured and notify them. This may be difficult to do if any of them currently have open reimbursement reports.

PR #13853, opened 2026-06-02
Fix legal name sent in organiser invoice notifications

transcental:
Currently the legal name is used in emails sent to other event managers about issued invoices. This PR changes it to preferred name which makes it consistent to the subject and removes sensitive information.


manuthecoder (approved):

Luke-Oldenburg (approved):
lgtm

PR #13857, opened 2026-06-05
[v4] Switch transaction filter tag to take in `public_id` instead of `id`

Mohamad-Mortada:
## Summary of the problem
We should be filtering by public id for tags not by their id

## Describe your changes
uses public id instead id


app/controllers/api/v4/transactions_controller.rb
Luke-Oldenburg, line 123: unrelated: I think we can make this more efficient by just doing `User.find_by_public_id` assuming we are validating that the user is valid.

Luke-Oldenburg (approved):
lgtm

PR #13860, opened 2026-06-06
[User] Add `subscribed_to_loops_at`

Luke-Oldenburg:
## Summary of the problem

We were accidentally resubscribing adults to mailing lists they had already subscribed from. Hack Club email tools does not currently support adults but when it does we can migrate them to using that system and this column will no longer be needed.

### DIFF
--- a/app/services/user_service/sync_with_loops.rb
+++ b/app/services/user_service/sync_with_loops.rb
@@ -27,6 +27,8 @@ module UserService
         user["HCB Has Card Grant?"] = @user.card_grants.any?
 
         user.save
+
+        @user.update(subscribed_to_loops_at: Time.now) if @user.subscribed_to_loops_at.nil?
       else
         body = {
           email: @user.email,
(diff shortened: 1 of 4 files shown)


app/services/user_service/sync_with_loops.rb
```diff
@@ -27,6 +27,8 @@ def run
         user["HCB Has Card Grant?"] = @user.card_grants.any?
 
         user.save
+
+        @user.update(subscribed_to_loops_at: Time.now) if @user.subscribed_to_loops_at.nil?
```
manuthecoder, line 31:
```suggestion
```

Luke-Oldenburg: I think this is useful to track for teenagers as well.

manuthecoder: teenagers sync to airtable not loops

garyhtou (commented):
Instead of adding `@user.update(subscribed_to_loops_at: Time.now)` to both branches of the teenager vs adult `if`, does it make sense to add `@user.update(subscribed_to_loops_at: Time.now) if @user.subscribed_to_loops_at.nil?` after that `if` this way it runs for both teens and adults?


Also, in the future, when we create a `User::Configuration` or `User::Profile` model, we should move this column off User.

Otherwise, LGTM — is there a need to backfill?

garyhtou (approved):

PR #13862, opened 2026-06-07
Add "org" placeholder for help links

### LINKED ISSUE #13865: [Feature] Org-picker redirect for feature-specific URLs (e.g. `/transfers`)
Allow shareable, feature-specific URLs like `hcb.hackclub.com/transfers` to work for any user by displaying an org-picker if the user belongs to multiple orgs, then redirecting them to the correct org's transfers page.

- Visiting `hcb.hackclub.com/transfers` (or similar feature paths) redirects the user to their org's transfers page
- Show an org-picker UI (similar to Cloudflare's account selector) prompting them to select which org to continue to, then redirect accordingly
- If the user is not logged in redirect to login and then continue to the intended destination

### Background

The HCB helpdesk is undergoing a major overhaul. Users can’t directly link to specific sections, like the Transfers tab. Help articles only offer generic instructions, lacking a universal deep link that works for all users. Screenshots are used as a fallback, which isn’t ideal.

### Reference

Cloudflare implements this pattern at `dash.cloudflare.com/dns`

Image

It shows an "Accounts" picker with the destination context ("Select an account and a domain to continue to: **DNS**"), then routes the user once they select.

polypixeldev:
## Summary of the problem

It'd be helpful if we can link to something like hcb.hackclub.com/org/transfers where the user can select the organization they want, instead of having to provide written instructions or a GIF to show users how to get to a specific page.

## Describe your changes

Refactors the `events_nav` method in `EventsHelper` to pull from a constant array of nav items. `SetEvent` then checks if the event slug is "org", matches the route against the available nav items, and displays the available events along with what the user will be continuing to.

image
image


app/controllers/concerns/set_event.rb
```diff
@@ -11,6 +11,44 @@ module SetEvent
...
+        current_user.events.not_hidden.each do |e|
+          if instance_exec(e, &@nav_item[:available_proc])
+            @available_events << e
+          else
+            @unavailable_events << e
+          end
+        end
+
+        render "events/placeholder"
+        return
```
garyhtou, line 49:
```suggestion
```

app/views/events/placeholder.html.erb
```diff
@@ -0,0 +1,61 @@
...
+<% if @unavailable_events.any? %>
+  <% if @available_events.any? %>
+    <p class="text-muted">
+      You are also in <%= pluralize(@unavailable_events.size, "other organization") %> (<%= @unavailable_events.map { |e| link_to sanitize(e.name), e }.to_sentence.html_safe %>), <%# erb_lint:disable ErbSafety %>
+      but do not have access to this page for those organizations.
+    </p>
+  <% else %>
+    <p>
+      You are in <%= pluralize(@unavailable_events.size, "organization") %> (<%= @unavailable_events.map { |e| link_to sanitize(e.name), e }.to_sentence.html_safe %>), <%# erb_lint:disable ErbSafety %>
+      but do not have access to this page for these organizations.
```
garyhtou:
```suggestion
      You are also in <%= pluralize(@unavailable_events.size, "other organization") %> that do not have access to this page: <%= @unavailable_events.map { |e| link_to sanitize(e.name), e }.to_sentence.html_safe %><%# erb_lint:disable ErbSafety %>
    </p>
  <% else %>
    <p>
      You are in <%= pluralize(@unavailable_events.size, "organization") %> that do not have access to this page: <%= @unavailable_events.map { |e| link_to sanitize(e.name), e }.to_sentence.html_safe %><%# erb_lint:disable ErbSafety %>
```

garyhtou (approved):

app/helpers/events_helper.rb
```diff
@@ -3,228 +3,204 @@
...
-    if policy(event).documentation?
-      items << {
-        name: "Documents",
-        path: event_documents_path(event_id: event.slug),
-        tooltip: "View legal documents and financial statements",
-        icon: "docs",
-        selected: selected == :documentation,
-      }
-    end
+  NAV_ITEMS = [
```
garyhtou, line 23: Add static tooltip for all, so that it can be rendered on the placeholder page.

garyhtou, line 23: Add a "schema" here so we know how to define each field

PR #13863, opened 2026-06-08
[v4] Wrap multiple write operations in an ActiveRecord Transaction

Mohamad-Mortada:
## Summary of the problem
Although unlikely there can be a possibility where if we save the ACH/Check object but the receipt creation fails then we end up with a scenario where we have a create transfer without its receipt. 

## Describe your changes
Using `ActiveRecord::Base.transaction` treats both operations as one atomic operation meaning if either fail it will rollback both.

### DIFF
--- a/app/controllers/api/v4/ach_transfers_controller.rb
+++ b/app/controllers/api/v4/ach_transfers_controller.rb
@@ -37,14 +37,16 @@ module Api
         end
 
 
-        @ach_transfer.save!
-        if ach_transfer_params[:file]
-          ::ReceiptService::Create.new(
-            uploader: current_user,
-            attachments: ach_transfer_params[:file],
-            upload_method: :ach_transfer_api,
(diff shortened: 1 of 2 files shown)


app/controllers/api/v4/ach_transfers_controller.rb
```diff
@@ -37,14 +37,16 @@ def create
...
-        @ach_transfer.save!
-        if ach_transfer_params[:file]
-          ::ReceiptService::Create.new(
-            uploader: current_user,
-            attachments: ach_transfer_params[:file],
-            upload_method: :api,
-            receiptable: @ach_transfer.local_hcb_code
-          ).run!
+        ActiveRecord::Base.transaction do
+          @ach_transfer.save!
```
Luke-Oldenburg, line 41:
```suggestion
          @ach_transfer.save!

```

Luke-Oldenburg (approved):
lgtm

PR #13870, opened 2026-06-09
Fix and reapply "Manually include non-v4 helpers in ApplicationHelper" again

polypixeldev:
## Summary of the problem

#13782 fixed some bugs with the last helpers PR, but a couple were left with the API. This PR fixes those.

## Describe your changes

Doesn't include `ApplicationHelper` twice, and moves `paginate_cursor` to a v4 controller concern as it is only used in controllers, not views.


manuthecoder (approved):
lgtm!

garyhtou (approved):

PR #13871, opened 2026-06-09
Use correct OOO callout partial on application review page

polypixeldev:
## Summary of the problem

#13812 removed the review_delay_callout partial being used for applications, but only replaced it with the new callout on the show page, not the review page.

## Describe your changes

Use the correct OOO callout on the application review page.


manuthecoder (approved):
lgtm!

Mohamad-Mortada (approved):

PR #13872, opened 2026-06-10
[v4][Card Grant] only update exp date if passed

Mohamad-Mortada:
## Summary of the problem
When you PATCH with with no `expiration_at` in the body, `expiration_at` evaluates to nil, and `.merge(expiration_at:
  nil)` tries to write NULL to the column. But card_grants.expiration_at is NOT NULL 

## Describe your changes
makes sure to only pass when needed


Luke-Oldenburg (approved):
lgtm

PR #13874, opened 2026-06-10
Spell organization with a z

polypixeldev:
## Summary of the problem

We spelled organisations here instead of organizations.

## Describe your changes

Change the spelling. There are other instances of this spelling in the codebase, but they are either for search or just comments/documentation, which is fine.


Luke-Oldenburg (approved):
lgtm

PR #13877, opened 2026-06-11
[Applications] Do not allow applicants to use their email as cosigner email

polypixeldev:
## Summary of the problem

People have submitted applications using their own email as the cosigner email. We shouldn't allow this.

## Describe your changes

Add a check in the ready to submit method and resend to cosigner route to make sure that the emails are different, and add a frontend validation.

I did not add this as a validation on the model since there are several existing applications in various stages that would fail the validation, and cause issues whenever any of these applications are updated.

### DIFF
--- a/app/controllers/event/applications_controller.rb
+++ b/app/controllers/event/applications_controller.rb
@@ -254,7 +254,7 @@ class Event
       new_cosigner_email = params[:event_application][:cosigner_email]
 
       if new_cosigner_email == @application.user.email
-        flash[:error] = "You cannot change the cosigner email to your email"
+        flash[:error] = "You cannot use your own email as your parent's email"
       else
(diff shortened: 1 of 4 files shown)


app/controllers/event/applications_controller.rb
```diff
@@ -251,14 +251,21 @@ def unarchive
     def resend_to_cosigner
       authorize @application
 
-      @application.update!(cosigner_email: params[:event_application][:cosigner_email])
+      new_cosigner_email = params[:event_application][:cosigner_email]
```
garyhtou:
```suggestion
      new_cosigner_email = params[:event_application][:cosigner_email]&.strip
```

garyhtou (approved):

PR #13879, opened 2026-06-11
[Applications] Update airtable status on activation and reissue

### LINKED ISSUE #13718: [Admin] Add Airtable sync for Onboarded and Documents Sent status
Upon organization activated, set Airtable status to Onboarded (used to be like this).

When contract reissued, set Airtable status to Documents Sent.

polypixeldev:
## Summary of the problem

We initially decided to not update the normal ops Status field in Airtable - however, this has been requested and would be helpful. These are one-time updates, and the status can still be changed by ops later and it will not revert back.

Closes #13718

## Describe your changes

Add a `set_airtable_status` method to Event::Application and use it on activation and contract reissue.

### DIFF
--- a/app/models/event/application.rb
+++ b/app/models/event/application.rb
@@ -291,6 +291,8 @@ class Event
       fs_contract.send!(reissue_signee_message:, reissue_cosigner_message:)
       fs_contract.party(:cosigner)&.notify unless reissue_of.present?
 
+      set_airtable_status("Documents sent") if reissue_of.present?
+
       fs_contract
     end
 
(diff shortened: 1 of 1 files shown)


app/models/event/application.rb
```diff
@@ -453,6 +457,15 @@ def user_ready_to_submit?
...
+
+      if airrecord.present?
+        airrecord["Status"] = status
+        airrecord.save
+      end
+    end
+
   end
 
 end
```
manuthecoder, line 473: if set_airtable_status fails, it'll raise an exception and crash the request even though the data was already saved

polypixeldev: Good catch, setting Airtable status isn't critical here so I've updated it to just handle the exception

garyhtou (approved):

PR #13889, opened 2026-06-14
Allow members to create suborganizations behind opt-in flag

garyhtou:
Members (not just managers) can now create suborganizations when the
per-event `member_subevent_creation` Flipper flag is enabled, in addition
to the existing `subevent_plan` gate. Managers and admins are unaffected.

When a suborg is created by someone below manager in the parent org, all
parent managers are notified via a new EventMailer#subevent_created email.
No initial transfer and no approval step are introduced.

https://claude.ai/code/session_011LiPDMebKbFHV3nVJ2juPP

### DIFF
--- a/app/services/event_service/create.rb
+++ b/app/services/event_service/create.rb
@@ -82,7 +82,7 @@ module EventService
     def notify_parent_event_managers(event)
       # Only suborg creation notifies (parent_event present); top-level/admin org creation never does.
       return if @parent_event.nil? || @invited_by.nil?
-      # Suppress when creator is a manager/admin/ancestor-manager of the parent — only sub-manager creators trigger the email.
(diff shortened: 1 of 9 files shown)


app/services/event_service/create.rb
```diff
@@ -71,10 +71,23 @@ def run
...
+
+      event
     end
 
     private
 
+    def notify_parent_event_managers(event)
+      # Only suborg creation notifies (parent_event present); top-level/admin org creation never does.
+      return if @parent_event.nil? || @invited_by.nil?
+      # Suppress when creator is a manager/admin/ancestor-manager of the parent — only sub-manager creators trigger the email.
```
garyhtou:
```suggestion
      # Suppress when creator is a manager/admin/ancestor-manager of the parent — only members creators trigger the email.
```

PR #13904, opened 2026-06-18
[Marketing] Add Reboot to the For Funders page

garyhtou:
## Summary of the problem

We want to feature Reboot.

## Describe your changes

Recipient-side counterpart to the Argosy case study: a quote from Reboot co-founder Jasmine Sun, with the Ford Foundation and Omidyar Network credited as Reboot's backers (not as HCB users). Mirrors the Argosy layout with a quote card and a "Backed by" logo footer.

image


garyhtou:
Got approval from Mel

Issue #12931, opened 2026-02-09
[Admin] Require admins to provide reasons for locking users


YodaLightsabr:
Right now, we don't have any context within HCB about why a user was locked from signing into HCB. While an external Airtable base exists, it's possible to forget to add an entry when locking a user. We should require that admins add a note when they want to lock a user.

PR #13905, opened 2026-06-18
feat(users): add locked reason functionality for user management

Acidicts:
## Summary of the problem
https://github.com/hackclub/hcb/issues/12931

## Describe your changes
- add locked_reason field and validation when locking users
- create Comment with lock reason when user is locked
- implement LockedReasonController to show/hide reason field based on lock checkbox
- update view to show locked reason input when locking user
- add controller specs for lock/unlock behavior with reason validation
- prevent locking without providing a reason
- maintain existing self-lock and admin-lock prevention logic

Demo:
https://github.com/user-attachments/assets/1e42ff6d-af1d-45d8-8bcd-11a2f060be5b

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -342,7 +342,7 @@ class UsersController < ApplicationController
       end
 
       locked = params[:user][:locked] == "1"
-      if @user.locked? != locked && (!params[:user][:locked_reason].empty? || @user.locked?)
+      if @user.locked? != locked
         if @user == current_user
           flash[:error] = "As much as you might desire to, you cannot lock yourself out."
           return redirect_to admin_user_path(@user)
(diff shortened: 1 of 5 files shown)


app/controllers/users_controller.rb
```diff
@@ -342,7 +342,7 @@ def update
       end
 
       locked = params[:user][:locked] == "1"
-      if @user.locked? != locked
+      if @user.locked? != locked && (!params[:user][:locked_reason].empty? || @user.locked?)
```
manuthecoder: this else fires whenever the lock state is unchanged so any admin editing a user without toggling the lock checkbox gets redirected with "to lock this user, provide a reason" and loses their edit

manuthecoder: params[:user][:locked_reason] can be nil, and will cause a NoMethodError

manuthecoder (changes requested):
`locked_reason` is a nonexisting schema attribute? there is no locked_reason column or method on User which will cause hcb to 500

Acidicts:
> `locked_reason` is a nonexisting schema attribute? there is no locked_reason column or method on User which will cause hcb to 500

The locked_reason is a parameter used for creating a Comment by the Admin in:

`elsif locked
          Comment.create(admin_only: true, commentable: @user, user_id: current_user.id, content: "Locked #{@user.preferred_name} : For - #{params[:user][:locked_reason]}")
          @user.lock!`

PR #13911, opened 2026-06-20
Increase cash withdrawal limit to $2k

polypixeldev:
Changing it now for Horizons Nexus, will change back to $500 after withdrawal


Luke-Oldenburg (approved):
lgtm

PR #13913, opened 2026-06-21
[CardGrant] Fix duplicate "blocked merchant" callout

manuthecoder:
image

Users in #hcb have reported this error message appearing twice. 

`error_boundary` double-rendered: the block wrote to the buffer and returned its HTML, so `<%=` appended it twice. `capture(&block)` buffers it separately and returns it once


Luke-Oldenburg (approved):
lgtm

PR #13917, opened 2026-06-22
[Donations] Paginate regular and recurring separately

### LINKED ISSUE #13685: [Donation] Paginate donations and recurring donations independently
See card grants overview on how to do this.

Luke-Oldenburg:
## Summary of the problem

Recurring donations were not paginated and would only display when you were on the first page of regular donations.
closes #13685

## Describe your changes

Now recurring donations and donations are paginated independently and both embedded into the donation overview as turbo frames.


app/controllers/events_controller.rb
```diff
@@ -627,6 +627,14 @@ def donation_overview
 
   end
 
+  def donations
+    authorize @event;
+  end
+
+  def recurring_donations
+    authorize @event;
+  end
```
Luke-Oldenburg, line 638:
```suggestion
  def donations
    authorize @event
  end

  def recurring_donations
    authorize @event
  end
```

Luke-Oldenburg:
@claude is this finished? can you tell me what still needs to be done or any bugs?

Luke-Oldenburg:
@claude can you finish this using the review from before and make the performance improvement you described to truly separate them?

PR #13924, opened 2026-06-22
[Ledger] Add one-time job to reset ledger tables

YodaLightsabr:
## Summary of the problem

We're going to need to do a lot of testing for our new ledger systems, and it will be helpful to have a way to reset our data.

## Describe your changes

This PR adds an index to HCB code for ledger item, and it creates a one-time job for resetting the ledger data.


app/jobs/one_time_jobs/nuke_ledger.rb
```diff
@@ -0,0 +1,56 @@
...
+    def clear_papertrail_versions
+      ledger_versions = PaperTrail::Version.where(item_type: ["Ledger", "Ledger::Item", "Ledger::Mapping"]).count
+      puts "Deleting #{ledger_versions} PaperTrail versions for Ledger models"
+      PaperTrail::Version.where(item_type: ["Ledger", "Ledger::Item", "Ledger::Mapping"]).delete_all
+
+      hcb_code_versions = PaperTrail::Version.where(item_type: "HcbCode")
+                                             .where("object_changes::text LIKE ?", "%ledger_item_id%")
+                                             .count
+      puts "Deleting #{hcb_code_versions} PaperTrail versions for HcbCode with ledger_item_id changes"
+      PaperTrail::Version.where(item_type: "HcbCode")
```
Luke-Oldenburg, line 58: Should we clear `event_id` also?

Luke-Oldenburg (commented):
If we ever make these records soft-deletable, we need to make sure to ignore act as paranoid in this script.

app/jobs/one_time_jobs/nuke_ledger.rb
YodaLightsabr: Because `event_id` is currently used in production and was added as part of the disbursement split (not the new `Ledger` transaction engine), we're going to keep it around.

Luke-Oldenburg (approved):
lgtm

Luke-Oldenburg (approved):
lgtm

PR #13931, opened 2026-06-23
[OPDR] Add assignee

### LINKED ISSUE #13180: [Admin] Processing State for OPDRs
### Describe the problem

OPDR (Organizer Position Deletion Requests) are often delayed because of needing information/receipts/terminating contracts for organizations.

This means that while they are in the in-between state where we have started work on them but the organization hasn't responded, they continue to show up in the Tasks older than 24 hours reminder email, and on the HCB dashboard as pending. This gets confusing when there are lots of different requests in various different states and leads to me missing stuff.

A Processing/Waiting state on an OPDR that means that it is excluded from the pending count number on the homepage and from the weekly email but isn't just completely closed (as closed deletion requests get buried and there is no search function)

Maybe something like this in the UI.
Image

polypixeldev:
## Summary of the problem

Often with OPDRs, we need additional information that means the OPDR is waiting in the under review state and included in reminder emails. There are comments, but it'd be better to see who's currently processing an OPDR and not include them in the under review OPDRs.
closes #13180

## Describe your changes

Add assignee to OPDR


app/controllers/organizer_position_deletion_requests_controller.rb
```diff
@@ -75,6 +75,20 @@ def open
...
 
+  def assign
+    authorize OrganizerPositionDeletionRequest
+    @opdr.update!(assignee: current_user)
+    flash[:success] = "Removal request assigned to you."
+    redirect_to @opdr
+  end
+
+  def unassign
+    authorize OrganizerPositionDeletionRequest
```
garyhtou:
```suggestion
    authorize @opdr
    @opdr.update!(assignee: current_user)
    flash[:success] = "Removal request assigned to you."
    redirect_to @opdr
  end

  def unassign
    authorize @opdr
```

garyhtou (approved):

app/controllers/organizer_position_deletion_requests_controller.rb
```diff
@@ -75,6 +75,27 @@ def open
...
+    authorize @opdr
+    @opdr.update!(assignee: current_user)
+    flash[:success] = "Removal request assigned to you."
+    redirect_to @opdr
+  end
+
+  def unassign
+    authorize @opdr
+    @opdr.update!(assignee: nil)
+    flash[:success] = "Removal request unassigned."
```
garyhtou:
```suggestion
    flash[:success] = "OPDR assigned to you."
    redirect_to @opdr
  end

  def unassign
    authorize @opdr
    @opdr.update!(assignee: current_user)
    flash[:success] = "OPDR assigned to you."
    redirect_to @opdr
  end

  def unassign
    authorize @opdr
    @opdr.update!(assignee: nil)
    flash[:success] = "OPDR unassigned."
```

PR #13932, opened 2026-06-23
Create LegalEntity

polypixeldev:
## Summary of the problem

As part of the new contractors feature, we're creating a new LegalEntity model that each user will have. This model will hold the tax information we need for each user when sending a payment.

## Describe your changes

Create the LegalEntity model, which is created lazily when accessed on a user, similar to Event::Configuration.

More columns will be added later when we build out collecting tax information, but this is what we need to build out phase 1 for now.


app/models/legal_entity.rb
```diff
@@ -0,0 +1,28 @@
...
+#  created_at          :datetime         not null
+#  updated_at          :datetime         not null
+#  managing_event_id   :bigint
+#
+# Indexes
+#
+#  index_legal_entities_on_managing_event_id  (managing_event_id)
+#
+class LegalEntity < ApplicationRecord
+  belongs_to :managing_event, class_name: "Event", optional: true
```
garyhtou, line 27: Can you add a comment here explaining that some LEs are managed by events (should be the less option flow)

app/models/user.rb
```diff
@@ -36,14 +36,20 @@
 #  created_at                    :datetime         not null
 #  updated_at                    :datetime         not null
 #  discord_id                    :string
+#  legal_entity_id               :bigint
```
garyhtou: We want to add a join model between them

app/models/user.rb
```diff
@@ -659,6 +667,13 @@ def unverified?
     !verified?
   end
 
+  def legal_entity
+    super || begin
+      le = LegalEntity.create!
+      update!(legal_entity: le)
+    end
```
garyhtou: this might be a good idea? or maybe a bad idea 🤷‍♂️
```suggestion
    super || with_lock do
    	return super if super.present?

      le = LegalEntity.create!
      update!(legal_entity: le)
    end
```

garyhtou (approved):

app/models/legal_entity_user.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  validate :person_entities_have_one_user, on: :create
+  validate :user_only_has_one_person_entity, on: :create
+
+  private
+
+  def person_entities_have_one_user
+    if legal_entity.person? && legal_entity.users.any?
+      errors.add(:base, "Legal entities with type person can only have one user")
+    end
```
garyhtou:
```suggestion
    if legal_entity.person? && LegalEntityUser.where(legal_entity:).excluding(self).any?
      errors.add(:base, "Legal entities with type person can only have one user")
    end
```

app/models/legal_entity_user.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def user_only_has_one_person_entity
+    if legal_entity.person? && user.legal_entities.person.any?
```
garyhtou:
```suggestion
    if legal_entity.person? && LegalEntityUser.where(user:).excluding(self).any?
```

app/models/legal_entity_user.rb
```diff
@@ -0,0 +1,39 @@
...
+#
+#  index_legal_entity_users_on_legal_entity_id  (legal_entity_id)
+#  index_legal_entity_users_on_user_id          (user_id)
+#
+class LegalEntityUser < ApplicationRecord
+  belongs_to :legal_entity
+  belongs_to :user
+
+  validate :person_entities_have_one_user, on: :create
+  validate :user_only_has_one_person_entity, on: :create
```
garyhtou:
```suggestion
  validate :person_entities_have_one_user
  validate :user_only_has_one_person_entity
```

garyhtou (approved):

PR #13933, opened 2026-06-23
[Webhook] Fix polarity of `handle_as_raw_pending_column_transaction`

Luke-Oldenburg:
## Summary of the problem

We were previously treating all incoming ACH transfers as credits. This only affected one transaction in production.

## Describe your changes

Check if type is "DEBIT"

### DIFF
--- a/app/controllers/column/webhooks_controller.rb
+++ b/app/controllers/column/webhooks_controller.rb
@@ -68,7 +68,7 @@ module Column
 
       RawPendingColumnTransaction.create!(
         column_id: @object[:id],
-        amount_cents: @object[:amount],
+        amount_cents: @object[:type] == "DEBIT" ? -1 * @object[:amount] : @object[:amount],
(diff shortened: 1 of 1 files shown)


YodaLightsabr (approved):

app/controllers/column/webhooks_controller.rb
```diff
@@ -68,7 +68,7 @@ def handle_as_raw_pending_column_transaction(column_event_type)
 
       RawPendingColumnTransaction.create!(
         column_id: @object[:id],
-        amount_cents: @object[:amount],
+        amount_cents: @object[:type] == "DEBIT" ? -1 * @object[:amount] : @object[:amount],
```
Luke-Oldenburg:
```suggestion
        amount_cents: @object[:type] == "DEBIT" ? -@object[:amount] : @object[:amount],
```

PR #13936, opened 2026-06-23
[Contracts] Hide unusable forms and buttons on contract signing page

polypixeldev:
## Summary of the problem

We still show the void and reissue buttons on the contract signing page even if it's not possible, and we also still show the activation form on contracts for applications that have already been activated.

## Describe your changes

Hide the forms/buttons when not needed, and link directly to the event on contract signing page if it's already been activated.


app/views/contract/parties/show.html.erb
```diff
@@ -16,27 +16,34 @@
...
 
         <% if @contract.contractable.is_a?(Event::Application) %>
-          <%= form_with url: admin_activate_application_path(@contract.contractable), method: :post do |form| %>
-            <div class="field">
-              <%= form.label "Tags" %>
-              <%= form.select :tags, options_for_select(EventTag::Tags::ALL, selected: @contract.contractable.default_tags), multiple: true, size: EventTag::Tags::ALL.size %>
-            </div>
+          <% if @contract.contractable.event.present? %>
+            <p>This application has already been activated.</p>
+            <%= link_to "Go to event", @contract.contractable.event, class: "btn w-full" %>
```
Luke-Oldenburg:
```suggestion
            <%= link_to "Go to organization", @contract.contractable.event, class: "btn w-full" %>
```

Luke-Oldenburg (approved):
lgtm

PR #13937, opened 2026-06-23
[TX Engine] Add balance anomaly detection

Luke-Oldenburg:
## Summary of the problem

We needed a way to detect discrepancies in event balances between the old and new transaction engines.

## Describe your changes

Add a mailer and job to detect balance discrepancies.

### DIFF
--- a/app/views/admin_mailer/balance_anomalies.html.erb
+++ b/app/views/admin_mailer/balance_anomalies.html.erb
@@ -1,0 +1,7 @@
+These events have discrepancies in their balances between the old and new transaction engine:
+
+<% @anomalous_events.find_each do |event| %>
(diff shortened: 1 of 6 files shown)


app/views/admin_mailer/balance_anomalies.html.erb
```diff
@@ -0,0 +1,7 @@
+These events have discrepancies in their balances between the old and new transaction engine:
+
+<% @anomalous_events.find_each do |event| %>
+  <h2><%= event.name %> (<%= event.id %>)
```
YodaLightsabr:
```suggestion
  <h2><%= event.name %> (<%= event.id %>)</h2>
```

YodaLightsabr (approved):

PR #13941, opened 2026-06-23
Create payee model

polypixeldev:
## Summary of the problem

Creating the payee model that payments will belong to.

## Describe your changes

Create the model.

### DIFF
--- a/app/models/payee.rb
+++ b/app/models/payee.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: payees
+#
+#  id              :bigint           not null, primary key
+#  preferred_name  :string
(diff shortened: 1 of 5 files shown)


app/models/payee.rb
```diff
@@ -0,0 +1,23 @@
...
+#  legal_entity_id :bigint           not null
+#
+# Indexes
+#
+#  index_payees_on_event_id         (event_id)
+#  index_payees_on_legal_entity_id  (legal_entity_id)
+#
+class Payee < ApplicationRecord
+  belongs_to :event
+  belongs_to :legal_entity
```
garyhtou, line 22: Make this join model unique - validation and index

garyhtou (approved):

PR #13943, opened 2026-06-23
Prevent old nuke scripts from running in production

YodaLightsabr:
## Summary of the problem

Nothing stops these very dangerous scripts from running in production.

## Describe your changes

This PR prevents these scripts from running outside of development.


Luke-Oldenburg (approved):
lgtm

PR #13947, opened 2026-06-23
[Sidekiq] Stop `ActiveStorage::TransformJob` retries

manuthecoder:
image

This job has been retrying for a while: 

```
MiniMagick::Error: `convert /tmp/XXXXXXX.jpg[0] -auto-orient -resize 1024x1024 /tmp/XXXXXX.jpg` failed with status: 1 and error: co...
```


garyhtou:
@albertchae, can you look into a fix for this? How can we get MiniMagick to succeed and not error here?

PR #13948, opened 2026-06-23
Add grant sorting

hanaeatsplanes:
## Summary of the problem

HCB Grants currently do not have a sorting mechanism, making it hard for organizers to find cards that are near-empty, old, or otherwise.

## Describe your changes

applied existing table sorting header to card grants

copilot summary:

This pull request improves the flexibility and maintainability of the card grants index by centralizing table column definitions and enhancing sorting logic, especially for the balance column. The changes ensure that the column configuration is consistent between the controller and the view, and introduce custom sorting for the balance column.

**Table column configuration and usage:**

* Introduced a constant `CARD_GRANT_COLUMNS` in `card_grants_controller.rb` to define all card grant table columns in a single place, including metadata for display names, sortability, and alignment. This column configuration is now assigned to `@table_columns` in both the `index` and `card_index` actions. [[1]](diffhunk://#diff-1ca2ec23a6679cc0efeb9b8a3216d674e2f09ccfa0cc4c904ac57b1176ab9839R14-R15) [[2]](diffhunk://#diff-1ca2ec23a6679cc0efeb9b8a3216d674e2f09ccfa0cc4c904ac57b1176ab9839R345-R352)
* Updated the `card_index.html.erb` view to dynamically render table headers based on `@table_columns`, allowing for easier updates and consistent column handling. The view now uses the new configuration to determine display names and sorting capabilities for each column.

**Sorting enhancements:**

* Improved sorting logic in the `card_index` action: when sorting by the `balance` column and the user is an organizer, the code now sorts the card grants in-memory by balance and paginates the result. For other columns, it uses a helper to sort based on the column configuration, excluding the balance column from database sorting due to its custom nature.

image


app/controllers/card_grants_controller.rb
```diff
@@ -23,13 +25,34 @@ def card_index
...
-      Arel.sql("card_grant_pre_authorizations.aasm_state='fraudulent' DESC"),
-      "card_grants.created_at DESC"
-    )
+    @card_grants = @event.card_grants
+                         .includes(:disbursement, :user, :stripe_card, :pre_authorization, :subledger)
+                         .order(
+                           Arel.sql("card_grant_pre_authorizations.aasm_state='fraudulent' DESC"),
+                           "card_grants.created_at DESC"
+                         )
+    @table_columns = CARD_GRANT_COLUMNS
```
manuthecoder: fraudulent grants no longer float to the top

the default path silently drops the intentional "fraudulent pre-authorizations first" ordering, so it should be preserved by folding it into the default-sort branch or prepending it as a leading .order before the secondary sort is applied

app/controllers/card_grants_controller.rb
```diff
@@ -23,13 +25,34 @@ def card_index
...
+
     # we allow searching by purpose but sometimes the purpose shown in the table is actually the memo
     @card_grants = @card_grants.search_for(params[:q]) if params[:q].present?
-    @paginated_card_grants = @card_grants.page(card_grants_page).per(card_grants_per_page)
+
+    if params[:sort] == "balance" && organizer_signed_in?
+      direction = params[:direction] == "asc" ? :asc : :desc
+      sorted_card_grants = @card_grants.to_a.sort_by { |card_grant| card_grant.balance.cents }
```
manuthecoder, line 43: n+1

app/controllers/card_grants_controller.rb
```diff
@@ -319,4 +342,13 @@ def set_card_grant
     @event = @card_grant.event
   end
 
+  CARD_GRANT_COLUMNS = [
+    { key: "status" },
+    { key: "created_at", display: "Date", default: true },
+    { key: "user_name", display: "To", column: "users.name", join: :user, sortable: false },
```
manuthecoder, line 355: why is this not sortable?

manuthecoder (changes requested):

app/controllers/card_grants_controller.rb
manuthecoder: see the copilot comments

app/controllers/card_grants_controller.rb
```diff
@@ -319,4 +342,13 @@ def set_card_grant
     @event = @card_grant.event
   end
 
+  CARD_GRANT_COLUMNS = [
```
manuthecoder, line 352: make this private

PR #13950, opened 2026-06-23
prevent duplicate ledger short codes

hanaeatsplanes:
## Summary of the problem

on `bin/rails db:migrate`, the migration can throw if duplicate ledger codes are generated by the script. 

## Describe your changes

This change ensures duplicate ledger codes cannot be generated


Luke-Oldenburg (commented):
I think we need to determine why duplicate short codes are being made rather than just ignoring them.

PR #13953, opened 2026-06-24
Removal of second club mention

dhyans26:
## Summary of the problem

On nonprofit.new, there are two instances of clubs being mentioned in the subtext


Luke-Oldenburg (approved):
lgtm

PR #13957, opened 2026-06-24
[Anomaly Mailer] Use hashids and correct HCB code amount

Luke-Oldenburg:
## Summary of the problem

Hash IDs are better and `smart_amount_cents` returns a more accurate amount of the HCB code.

## Describe your changes

Switch to these


YodaLightsabr (approved):
Thank you

PR #13958, opened 2026-06-24
Create payment model

polypixeldev:
## Summary of the problem

Creates the Payment model that will be used for the new contractors feature. This is the last model that will be created for the MVP of this feature.

## Describe your changes

Creates Payment model with AASM and receiptable. More will be added later (such as aasm transitions/callbacks and public activity)

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -35,9 +35,7 @@ class Payment < ApplicationRecord
   monetize :amount_cents
 
   aasm timestamps: true do
-    state :submitted, initial: true
-    state :pending_legal_entity
-    state :pending_payment_method
+    state :pending_legal_entity, initial: true
     state :under_review
     state :sent
     state :rejected
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -16,17 +16,21 @@
 #  under_review_at :datetime
 #  created_at      :datetime         not null
 #  updated_at      :datetime         not null
+#  payee_id        :bigint           not null
 #  payout_id       :bigint
 #
 # Indexes
 #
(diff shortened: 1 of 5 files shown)


app/models/payment.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  belongs_to :payout, polymorphic: true, optional: true
+  belongs_to :payee
+
+  monetize :amount_cents
+
+  aasm timestamps: true do
+    state :submitted, initial: true
+    state :pending_legal_entity
+    state :pending_payment_method
```
garyhtou: Should we combine this as just `pending_legal_entity`? It represents the state that we're waiting on the `pending_legal_entity` to complete tasks. I don't think we need to get as specific as `pending_payment_method` because we may potentionaly add more steps of LEs - and we don't want to have to add states here to represent those additional steps

app/models/payment.rb
```diff
@@ -0,0 +1,56 @@
...
+    state :under_review
```
garyhtou: can we add a comment on who's reviewing?

garyhtou (approved):

app/models/payment.rb
```diff
@@ -0,0 +1,54 @@
...
+  include Receiptable
+  has_paper_trail
+
+  belongs_to :payout, polymorphic: true, optional: true
+  belongs_to :payee
+
+  monetize :amount_cents
+
+  aasm timestamps: true do
+    state :pending_legal_entity, initial: true
```
garyhtou:
```suggestion
    state :pending_legal_entity, initial: true # We're waiting on the LE to complete tasks before payment can be sent
```

PR #13962, opened 2026-06-24
[Legal Entity] Create LE Payout Method

Mohamad-Mortada:
## Summary of the problem
If we want to support multiple payout methods it would make sense to have a list of canonical payout methods which have a `details` value exposing the underlying type of transfer it is (ach, check, wire, etc). PayoutMethod has a default value which can be set which will be used later in the future for reimbursements & payroll. 

What will need to happen is to change the logic of how we reference payout method as we need to access its details as this is a record not a module


polypixeldev (approved):
check with gary if we can remove paypal yet - otherwise lgtm!

app/models/legal_entity/payout_method/ach_transfer.rb
```diff
@@ -0,0 +1,51 @@
...
+#  id                        :bigint           not null, primary key
+#  account_number_ciphertext :text             not null
+#  routing_number_ciphertext :text             not null
+#  created_at                :datetime         not null
+#  updated_at                :datetime         not null
+#
+class LegalEntity
+  class PayoutMethod < ApplicationRecord
+    class AchTransfer < ApplicationRecord
+      include Shared
```
garyhtou: You'll need to create a Shared concern

app/models/legal_entity/payout_method.rb
```diff
@@ -0,0 +1,73 @@
...
+    def details_must_be_supported
+      if unsupported?
+        errors.add(:base, "#{unsupported_details[:reason]} Please choose another method.")
+      end
+    end
+
+    def unset_other_defaults
+      LegalEntity::PayoutMethod
+        .where(legal_entity_id:)
+        .where.not(id:)
```
garyhtou:
```suggestion
        .excluding(self)
```

app/models/legal_entity.rb
```diff
@@ -31,4 +31,10 @@ class LegalEntity < ApplicationRecord
   has_many :legal_entity_users
   has_many :users, through: :legal_entity_users
 
+  has_many :payout_methods, class_name: "LegalEntity::PayoutMethod", dependent: :destroy
```
garyhtou:
```suggestion
  has_many :payout_methods, class_name: "LegalEntity::PayoutMethod"
```

app/models/legal_entity/payout_method/ach_transfer.rb
garyhtou: Update: You just need to drop this line for all - that code is handled by the Payout Method class

garyhtou (approved):
Vibe code some tests

garyhtou (changes requested):
couple additional issues to fix it - chatted in person

garyhtou (approved):

PR #13963, opened 2026-06-24
[Maintenance Task] Add backfill legal entity payout methods

Mohamad-Mortada:
## Summary of the problem
Now that we have Legal Entity Payouts (assuming #13962 has been merged) we need to populate them with the data from User Payout Methods. 

## Describe your changes
Goes through all users who have a payout ID and are a `LegalEntityUser` then create a `LegalEntity::PayoutMethod` identical to their existing payout information. Also that payout method's default value is true.

### DIFF
--- a/app/tasks/maintenance/backfill_legal_entity_payout_methods_task.rb
+++ b/app/tasks/maintenance/backfill_legal_entity_payout_methods_task.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module Maintenance
+  # Backfills LegalEntity::PayoutMethod records from user PayoutMethod.
+  class BackfillLegalEntityPayoutMethodsTask < MaintenanceTasks::Task
+    def collection
+      User.where.not(payout_method_id: nil)
+          .where(id: LegalEntityUser.joins(:legal_entity)
+                                    .where(legal_entities: { entity_type: :person })
+                                    .select(:user_id))
+    end
+
+    def process(user)
+      legal_entity = user.legal_entities.find_by(entity_type: :person)
+      return unless legal_entity
+
+      details_class = user.payout_method_type.sub(/\AUser::/, "LegalEntity::").safe_constantize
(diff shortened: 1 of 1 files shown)


app/tasks/maintenance/backfill_legal_entity_payout_methods_task.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Maintenance
+  # Backfills LegalEntity::PayoutMethod records from user PayoutMethod.
+  class BackfillLegalEntityPayoutMethodsTask < MaintenanceTasks::Task
+    def collection
+      User.where.not(payout_method_id: nil)
+          .where(id: LegalEntityUser.joins(:legal_entity)
+                                    .where(legal_entities: { entity_type: :person })
+                                    .select(:user_id))
```
garyhtou: In theory, all users should have an LE - let's drop this line so it errors loudly if a user does't have an LE
```suggestion
```

app/tasks/maintenance/backfill_legal_entity_payout_methods_task.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    def process(user)
+      legal_entity = user.legal_entities.find_by(entity_type: :person)
```
garyhtou, line 11: feedback for future: it may be nice to have a `User#personal_legal_entity` method

app/tasks/maintenance/backfill_legal_entity_payout_methods_task.rb
```diff
@@ -0,0 +1,29 @@
...
+      return unless legal_entity
```
garyhtou:
```suggestion
      if legal_entity.nil?
        raise ArgumentError, "LE missing for User #{user.id}"
      end
```

garyhtou (approved):

PR #13964, opened 2026-06-24
Add creator and currency to Payment

polypixeldev:
## Summary of the problem

Missed these columns on the original Payment model PR (#13958)

## Describe your changes

Add reference to creator and the currency column

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -36,7 +36,7 @@ class Payment < ApplicationRecord
   belongs_to :payee
   belongs_to :creator, class_name: "User"
 
-  monetize :amount_cents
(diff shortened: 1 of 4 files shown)


app/models/payment.rb
```diff
@@ -31,6 +34,7 @@ class Payment < ApplicationRecord
 
   belongs_to :payout, polymorphic: true, optional: true
   belongs_to :payee
+  belongs_to :creator, class_name: "User"
 
   monetize :amount_cents
```
garyhtou: do you need to tell it to look at the currency column? or does monetize automatically detect that column?

garyhtou (approved):

PR #13970, opened 2026-06-25
Check ledger items in anomaly script and exclude empty HcbCodes

YodaLightsabr:
## Summary of the problem

We want to also detect anomalies from the perspective of ledger items (similar to a full outer join). However, we also want to exclude HcbCodes without canonical transactions that are missing a ledger item.

## Describe your changes

This PR includes ledger item anomalies in the script and excludes empty HCB codes.

### DIFF
--- a/app/jobs/admin/detect_logical_transaction_anomalies_job.rb
+++ b/app/jobs/admin/detect_logical_transaction_anomalies_job.rb
@@ -4,13 +4,22 @@ module Admin
   class DetectLogicalTransactionAnomaliesJob < ApplicationJob
     queue_as :low
 
+    def initialize
+      @event_id = 183
+    end
+
     def perform
       hcb_codes = []
-      HcbCode.where(event_id: 183).find_each do |hcb_code|
-        hcb_codes << hcb_code.id if hcb_code.ledger_item.nil? || hcb_code.smart_amount_cents != hcb_code.ledger_item.amount_cents
+      HcbCode.where(event_id: @event_id).find_each do |hcb_code|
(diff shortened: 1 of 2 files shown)


app/jobs/admin/detect_logical_transaction_anomalies_job.rb
```diff
@@ -6,11 +6,20 @@ class DetectLogicalTransactionAnomaliesJob < ApplicationJob
 
     def perform
       hcb_codes = []
-      HcbCode.where(event_id: 183).find_each do |hcb_code|
-        hcb_codes << hcb_code.id if hcb_code.ledger_item.nil? || hcb_code.smart_amount_cents != hcb_code.ledger_item.amount_cents
+      HcbCode.where(event_id:).find_each do |hcb_code|
+        hcb_codes << hcb_code.id if hcb_code.smart_amount_cents != hcb_code.ledger_item.amount_cents || (hcb_code.ledger_item.nil? && hcb_code.no_transactions?)
```
Luke-Oldenburg: If ledger item is nil, this will error having the if check in this order. The nil check should come first because if that fails then the rest of the if check will not run.

Luke-Oldenburg:
```suggestion
        hcb_codes << hcb_code.id if (hcb_code.ledger_item.nil? && hcb_code.no_transactions?) || hcb_code.smart_amount_cents != hcb_code.ledger_item.amount_cents
```

Luke-Oldenburg (approved):
lgtm

PR #13982, opened 2026-06-25
[Payout Method] Make LE Payout Method details index unique

Mohamad-Mortada:
## Summary of the problem
We want to make sure that only one payout method is associated with a detail. 

## Describe your changes
Makes index unique

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1364,7 +1364,7 @@ ActiveRecord::Schema[8.0].define(version: 2026_06_25_142740) do
     t.text "secondary_hash"
     t.text "unique_bank_identifier"
     t.datetime "updated_at", null: false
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -1364,7 +1364,7 @@
     t.text "secondary_hash"
     t.text "unique_bank_identifier"
     t.datetime "updated_at", null: false
-    t.index ["duplicate_of_hashed_transaction_id"], name: "index_hashed_transactions_on_duplicate_of_hashed_transaction_id"
+    t.index ["duplicate_of_hashed_transaction_id"], name: "idx_on_duplicate_of_hashed_transaction_id_6a29e8a078"
```
garyhtou: drop this line

garyhtou (commented):
Image

garyhtou (approved):

PR #13985, opened 2026-06-25
[PayoutHolding] Use safe nav on CPT decline in `reverse!`

Luke-Oldenburg:
## Summary of the problem

Failed reimbursements from before `2024-08-09 00:45:12.992236000 UTC +00:00` were unable to be cancelled because they would error from missing a CPT.

## Describe your changes

Use safe nav when calling cancel on CPTs from before this date.

### DIFF
--- a/app/models/reimbursement/payout_holding.rb
+++ b/app/models/reimbursement/payout_holding.rb
@@ -114,7 +114,12 @@ module Reimbursement
 
         mark_reversed!
 
-        canonical_pending_transaction.decline!
+        # This is the created_at of the first payout holding that had a CPT
+        if self.created_at < DateTime.parse("2024-08-09 00:45:12.992236000 UTC +00:00")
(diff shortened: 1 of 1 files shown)


app/models/reimbursement/payout_holding.rb
```diff
@@ -114,7 +114,12 @@ def reverse!
 
         mark_reversed!
 
-        canonical_pending_transaction.decline!
+        # This is the created_at of the first payout holding that had a CPT
+        if self.created_at < DateTime.parse("2024-08-09 00:45:12.992236000 UTC +00:00")
```
YodaLightsabr:
```suggestion
        if self.created_at <= DateTime.parse("2024-08-09 00:45:12.992236000 UTC +00:00")
```

YodaLightsabr (changes requested):

YodaLightsabr (approved):

PR #13987, opened 2026-06-25
Add backend logic for one-time Payments

polypixeldev:
## Describe your changes
All the core backend logic for one-time Payments!

- Creating a Payment from amount, purpose, and payee
- Creating Payment::Attempts for each payment attempt
- Creating payout transfers for each Payment::Attempt based on the default payout method
- Callbacks and AASM events to link it all together
- Mailers at each step of the process


app/models/payment.rb
```diff
@@ -28,23 +25,72 @@
...
 #
 class Payment < ApplicationRecord
+  self.ignored_columns += ["payout_type", "payout_id", "failed_at"]
+
   include AASM
   include Receiptable
   has_paper_trail
 
-  belongs_to :payout, polymorphic: true, optional: true
   belongs_to :payee
```
garyhtou, line 34:
```suggestion
  belongs_to :payee
  has_one :legal_entity, through: :payee
```

app/views/payment/attempt_mailer/failed_payee.html.erb
```diff
@@ -0,0 +1,22 @@
+<p>Hello <%= @payment.payee.preferred_name %>,</p>
```
garyhtou, line 1: Later: use LE's name / DBA

garyhtou (approved):
other feedback was talked about and address in person

PR #13990, opened 2026-06-25
[Anomaly Job] Run hourly

Luke-Oldenburg:
Run hourly and only send if there are errors

### DIFF
--- a/config/schedule.yml
+++ b/config/schedule.yml
@@ -145,7 +145,7 @@ calculate_subjects_metrics:
   class: "Metric::CalculateSubjectsJob"
 
 detect_logical_transaction_anomalies:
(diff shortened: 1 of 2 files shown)


config/schedule.yml
```diff
@@ -145,7 +145,7 @@ calculate_subjects_metrics:
   class: "Metric::CalculateSubjectsJob"
 
 detect_logical_transaction_anomalies:
-  cron: "0 7 * * *" # run every 1 day
+  cron: "0 * * * *" # run every 1 day
```
YodaLightsabr:
```suggestion
  cron: "0 * * * *" # run every 1 hour
```

YodaLightsabr (approved):

PR #13991, opened 2026-06-25
[V4 API] support tag_ids on transaction update

24c02:
for too long, you couldn't change the tags on transactions via the V4 API.

this ends now.

`PATCH /api/v4/organizations/:org_id/transactions/:id`

```json
{
 "tag_ids": ["tag_xxx", "tag_yyy"]
}
```


app/controllers/api/v4/transactions_controller.rb
```diff
@@ -74,6 +74,18 @@ def update
           @hcb_code.update_custom_memo!(params[:memo])
         end
 
+        if params.key? :tag_ids
+          tags = Array(params[:tag_ids]).map { |id| Tag.find_by_public_id!(id) }
```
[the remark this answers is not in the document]
24c02: (a) what are you smoking, and
(b) are you willing to share?

app/controllers/api/v4/transactions_controller.rb
```diff
@@ -73,6 +74,18 @@ def update
...
+
+          tags.each do |tag|
+            authorize tag, :toggle_tag?
+            raise Pundit::NotAuthorizedError unless @hcb_code.events.include?(tag.event)
+          end
+
+          @hcb_code.tags = tags
+          @hcb_code.save!
+        end
```
garyhtou: Should we wrap this in a transaction? This way, a RecordNotFound tag doesn't result in a partial update (when a memo is also passed)

24c02: good catch, done!

Mohamad-Mortada (approved):

PR #14002, opened 2026-06-26
[Ledger] Rename Item#date to #datetime (2/5): switch reads/writes + NOT NULL

garyhtou:
## Summary of the problem

Part of renaming `Ledger::Item#date` to `#datetime` without downtime. **PR 2 of 5**, stacked on #14001.

1. #14001 · add `datetime`, sync with `date`, backfill via maintenance task
2. **This PR** · switch reads/writes to `datetime`, enforce `NOT NULL`
3. #14003 · stop referencing `date` (`ignored_columns`)
4. #14007 · drop `date`
5. #14010 · remove the `date` `ignored_columns` entry

## Describe your changes

- Point all application code at `datetime`: the `create_ledger_item!` writers, the `order(...)` clauses, the ledger item views, the backfill job, and the presence validation.
- Promote `datetime` to `NOT NULL` using a validated check constraint (no long table lock), then drop `NOT NULL` on the legacy `date` column so it can stop being written in PR 3.
- The dual-write callback stays, so in-flight requests on the PR 1 release keep `date` populated during the rolling deploy.

> [!IMPORTANT]
> The `BackfillLedgerItemDatetime` task from #14001 must finish before this PR's `NOT NULL` migration runs.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -1141,7 +1141,7 @@ class EventsController < ApplicationController
     authorize @event
 
     @ledger = @event.ledger
-    @items = @ledger.items.order(date: :desc, created_at: :desc, id: :desc).page(params[:page])
+    @items = @ledger.items.order(datetime: :desc, created_at: :desc, id: :desc).page(params[:page])
   end
 
   private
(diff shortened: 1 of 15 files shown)


garyhtou (commented):
I checked to ensure datetime is all NOT NULL after the migration ran

PR #14013, opened 2026-06-26
Drop superseded plaintext refresh_token column from api_tokens

garyhtou:
## Summary of the problem

`ApiToken#refresh_token` (plaintext `:string` column) was superseded by the encrypted `refresh_token_ciphertext` column (`has_encrypted :refresh_token`) and has been hidden via `self.ignored_columns`. The dead column is still in the table; this drops it.

## Describe your changes

- Drop the plaintext `refresh_token` column from `api_tokens` (it has no index or foreign key of its own; `refresh_token_bidx` and `refresh_token_ciphertext` are untouched).
- Keep `self.ignored_columns += ["refresh_token"]` so a process still on older code can't select the column while the drop propagates (this is the `strong_migrations` ignore-before-drop pattern; the ignore can be removed in a later cleanup).

Verified locally: the column is gone, the encrypted `refresh_token` attribute still reads/writes, and `annotaterb --frozen` is unchanged.

> [!WARNING]
> This assumes the plaintext column holds no data that hasn't already been migrated into `refresh_token_ciphertext`. Please confirm before merging.


Luke-Oldenburg:
This was done in https://github.com/hackclub/hcb/pull/14807

PR #14014, opened 2026-06-26
Drop the reissued_for_id column from increase_checks

garyhtou:
## Summary of the problem

`IncreaseCheck#reissued_for_id` backed check reissuing, which was removed in #13653. The column has been hidden via `self.ignored_columns` ever since, but still exists in the table (with its index and a self-referential foreign key). This drops it.

## Describe your changes

- Drop the `reissued_for_id` column from `increase_checks`. Postgres drops the dependent index (`index_increase_checks_on_reissued_for_id`) and the self-referential foreign key with it.
- Keep `self.ignored_columns += ["reissued_for_id"]` so a process still on older code can't select the column while the drop propagates (the `strong_migrations` ignore-before-drop pattern; the ignore can be removed in a later cleanup).

No code references the column or a `reissued_for` association. Verified locally: column, index, and foreign key are gone, the model boots, and `annotaterb --frozen` is unchanged.


Luke-Oldenburg (approved):
lgtm

PR #14017, opened 2026-06-26
[HcbCode] Move link to `Ledger::Item` into turboframe

Luke-Oldenburg:
## Summary of the problem

Didn't render in popover.

## Describe your changes

Render in popover.


Luke-Oldenburg:
approved by @YodaLightsabr

PR #14023, opened 2026-06-26
[Legal Entity Payout] use new create_transfer method for LEPM

Mohamad-Mortada:
Consolidates information from nightly method by using the new create_transfer method which uses the provided information in payout method along with provided attributes.


polypixeldev (commented):
looks good - in the new TX engine payout holding thing we can see if we can get rid of the case in payout holding nightly completely. do the same thing in https://github.com/hackclub/hcb/blob/980af9100d769c6d19146d6b674064b4cc6423a0/app/models/payment/attempt.rb

app/models/legal_entity/payout_method.rb
```diff
@@ -63,6 +63,10 @@ def unsupported_details
       self.class.unsupported_details(details.class)
     end
 
+    def create_transfer(event, **attr)
+      details.create_transfer(event, **attr)
+    end
```
garyhtou:
```suggestion
    delegate :create_transfer, to: :details
```

app/models/legal_entity/payout_method/ach_transfer.rb
garyhtou, line 1: Okay, now that I'm remembering how different all of these transfer classes are, I'd advocate that we don't pass `**attr` into `SomeTransferType.build` - it can be brittle. Instead, each Transfer type should explicitly pull attributes from `attr` and explicitly pass them into `build`.

Then, somewhere, we should define a shape of what MUST/MAY be passed into all `create_transfer` calls

app/models/legal_entity/payout_method/check.rb
```diff
@@ -60,6 +60,24 @@ def currency
         "USD"
       end
 
+      # See LegalEntity::PayoutMethod for the shared `create_transfer` contract.
+      def create_transfer(event, amount:, payment_for:, recipient_name:, recipient_email:, user:, memo:, send_email_notification: false, **)
```
garyhtou, line 72: @claude is the `**` param necessary?

app/models/legal_entity/payout_method/wise_transfer.rb
```diff
@@ -48,6 +48,26 @@ def title_kind
         "Wise Transfer"
       end
 
+      # See LegalEntity::PayoutMethod for the shared `create_transfer` contract.
+      def create_transfer(event, amount:, payment_for:, recipient_name:, recipient_email:, user:, **)
```
polypixeldev: are we missing `bank_name` here? it's in the payout method and wise transfer model although doesn't seem to be used

polypixeldev (approved):

PR #14024, opened 2026-06-26
[Ledger::Item] Add polymorphic association to linked object

Luke-Oldenburg:
## Summary of the problem

We need a way to reference linked object from ledger item polymorphically.

## Describe your changes

Do it


garyhtou (approved):
nice! how are we saving/creating the linked_object association?

YodaLightsabr (approved):

db/migrate/20260626195451_add_linked_object_to_ledger_items.rb
```diff
@@ -0,0 +1,7 @@
+class AddLinkedObjectToLedgerItems < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :ledger_items, :linked_object, polymorphic: true, null: true, index: {algorithm: :concurrently}
```
YodaLightsabr:
```suggestion
    add_reference :ledger_items, :linked_object, polymorphic: true, null: true, index: { algorithm: :concurrently }
```

YodaLightsabr (approved):

PR #14038, opened 2026-06-29
Fix detect balance anomalies job

YodaLightsabr:
## Summary of the problem

This job has a couple of bugs right now, and it takes a long time to run.

## Describe your changes

Improve performance and fix bugs.

### DIFF
--- a/app/jobs/admin/detect_balance_anomalies_job.rb
+++ b/app/jobs/admin/detect_balance_anomalies_job.rb
@@ -7,10 +7,21 @@ module Admin
     def perform
       anomalous_events = []
       Event.find_each do |event|
-        anomalous_events << event.id if event.ledger.balance_cents != event.balance_v2_cents || event.ledger.balance_cents != event.hcb_codes.on_main_ledger.sum(&:smart_amount_cents)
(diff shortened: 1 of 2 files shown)


app/jobs/admin/detect_balance_anomalies_job.rb
```diff
@@ -7,10 +7,21 @@ class DetectBalanceAnomaliesJob < ApplicationJob
     def perform
       anomalous_events = []
       Event.find_each do |event|
-        anomalous_events << event.id if event.ledger.balance_cents != event.balance_v2_cents || event.ledger.balance_cents != event.hcb_codes.on_main_ledger.sum(&:smart_amount_cents)
+        if event.ledger.balance_cents != event.balance_v2_cents
```
garyhtou, line 10: I'm guessing we no longer care about `smart_amount_cents`?

garyhtou (approved):

app/jobs/admin/detect_balance_anomalies_job.rb
YodaLightsabr: Not especially. We can look into it more as a debugging tool, but it is slow to calculate and usually matches the organization's balance exactly.

PR #14039, opened 2026-06-29
[Reimbursement][LEPM] Associate a specific payout method to a report

Mohamad-Mortada:
## Summary of the problem
Right now we're just pulling the default payout method from the user directly for payout methods. We want to associate a specific payout method to a report. Currently, this will look like either passing in a payout method on create or it will use the default. This will give better support for the future once we support multiple payout methods in which it'll be easier to identify which payout method a report uses. 

## Describe your changes
Added a new payout method ID column to reimbursement reports and updated to use it. If a reimbursement report payout details needs to be updated we will repoint the report to the latest payout method. In future we will have them mutate the existing payout report they used (we wont create a new payout method on update). We could also support switching payout settings in reimbursement page when a report is failed or hasn't been submitted/approved yet.

### DIFF
--- a/app/models/reimbursement/report.rb
+++ b/app/models/reimbursement/report.rb
@@ -101,7 +101,7 @@ module Reimbursement
 
     acts_as_paranoid
 
-    before_create :snapshot_payout_method
+    before_create :set_payout_method
 
     after_create_commit do
       ReimbursementMailer.with(report: self).invitation.deliver_later if inviter != user
@@ -436,7 +436,7 @@ module Reimbursement
 
     private
 
-    def snapshot_payout_method
+    def set_payout_method
       self.legal_entity_payout_method ||= user&.default_payout_method
     end
 
--- a/app/models/reimbursement/report.rb
+++ b/app/models/reimbursement/report.rb
@@ -4,40 +4,43 @@
 #
 # Table name: reimbursement_reports
 #
-#  id                         :bigint           not null, primary key
-#  aasm_state                 :string           not null
-#  conversion_rate            :float            default(1.0), not null
-#  currency                   :string           default("USD"), not null
-#  deleted_at                 :datetime
-#  expense_number             :integer          default(0), not null
-#  invite_message             :text
-#  maximum_amount_cents       :integer
-#  name                       :text
-#  reimbursed_at              :datetime
-#  reimbursement_approved_at  :datetime
-#  reimbursement_requested_at :datetime
-#  rejected_at                :datetime
-#  submitted_at               :datetime
-#  created_at                 :datetime         not null
-#  updated_at                 :datetime         not null
-#  card_grant_id              :bigint
-#  event_id                   :bigint
-#  invited_by_id              :bigint
-#  reviewer_id                :bigint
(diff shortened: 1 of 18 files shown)


app/models/reimbursement/report.rb
```diff
@@ -424,8 +430,16 @@ def convert_to_wise_transfer!(as: User.system_user)
       end
     end
 
+    def payout_method
+      legal_entity_payout_method || user&.default_payout_method
+    end
```
garyhtou, line 435: This is temporary. The `|| user&.default_payout_method` needs to be removed after running backfill

app/models/reimbursement/report.rb
```diff
@@ -424,8 +430,16 @@ def convert_to_wise_transfer!(as: User.system_user)
...
+
     private
 
+    def snapshot_payout_method
```
garyhtou:
```suggestion
    def set_payout_method
```

app/models/reimbursement/report.rb
```diff
@@ -97,6 +101,8 @@ class Report < ApplicationRecord
 
     acts_as_paranoid
 
+    before_create :snapshot_payout_method
```
garyhtou:
```suggestion
    before_create :set_payout_method
```

app/services/legal_entity/payout_method_service/update.rb
```diff
@@ -75,10 +79,23 @@ def apply_business_rules
...
       end
 
+      def repoint_failed_reports(replaced_method)
+        return unless replaced_method
+
+        @user.reimbursement_reports
+             .joins(:payout_holding)
+             .where(reimbursement_payout_holdings: { aasm_state: :failed })
+             .where(legal_entity_payout_method_id: replaced_method.id)
+             .update_all(legal_entity_payout_method_id: @payout_method.id, updated_at: Time.current)
```
garyhtou:
```suggestion
             .update(legal_entity_payout_method: @payout_method)
```

app/tasks/maintenance/backfill_reimbursement_report_payout_methods_task.rb
```diff
@@ -0,0 +1,17 @@
...
+  class BackfillReimbursementReportPayoutMethodsTask < MaintenanceTasks::Task
+    def collection
+      Reimbursement::Report.where(legal_entity_payout_method_id: nil)
+    end
+
+    def process(report)
+      payout_method = report.user&.default_payout_method
+      return unless payout_method
+
+      report.update_columns(legal_entity_payout_method_id: payout_method.id, updated_at: Time.current)
```
garyhtou:
```suggestion
      report.update(legal_entity_payout_method: payout_method)
```

app/services/legal_entity/payout_method_service/update.rb
```diff
@@ -75,10 +79,23 @@ def apply_business_rules
         end
       end
 
+      def repoint_failed_reports(replaced_method)
```
garyhtou: also repoint draft reports

garyhtou (approved):

PR #14042, opened 2026-06-29
Create Argosy 2026 spend-only plan

polypixeldev:
## Summary of the problem

Requested by Rhys.

## Describe your changes

Copied from Argosy 2025 but with new template, and added to the is_argosy check to display promotions.


YodaLightsabr (approved):
This is awful, but it's OK because it's only temporary.

PR #14050, opened 2026-06-29
Added "comment" to receipt upload through email/text

### LINKED ISSUE #13873: Add "comment" to receipt upload through email/text
### Describe the problem

It would be super helpful if when you upload a receipt for a transaction through the email/text link that is automatically sent, if there was also a spot to add a comment/description of the transaction right in that same screen. This way you don't have to take extra steps to add notes to a transaction and it's easier for accounting to see what each transaction is actually for. 

Reach out to Sierra with any questions - I'm using Lucy's account to submit this request

description above ^^

Sarvesh-Mk:
## Summary of the problem

fixes #13873 

## Describe your changes

Currently only works for email reciepts

https://github.com/user-attachments/assets/ca963b12-c178-4156-9f55-1aaeabd42070

### DIFF
--- a/app/mailboxes/hcb_code_mailbox.rb
+++ b/app/mailboxes/hcb_code_mailbox.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "email_reply_parser"
+
 class HcbCodeMailbox < ApplicationMailbox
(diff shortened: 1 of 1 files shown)


Sarvesh-Mk:
Should this only work for admins or should this work for all users?

PR #14051, opened 2026-06-29
[Disbursement] Make incoming and outgoing AR subclasses of base

Luke-Oldenburg:
## Summary of the problem

We needed a way to add a polymorphic relationship to the linked object on `Ledger::Item` that is capable of storing and retrieving `Disbursement::Incoming` and `Disbursement::Outgoing` even though they aren't directly DB-backed. 

## Describe your changes

Make `Disbursement::Incoming` and `Disbursement::Outgoing` AR models pointing to `disbursements` that inherit from `Disbursement::Base`


manuthecoder (approved):

garyhtou (commented):
Image

I think the shared concern solution makes the most sense - i think this is what Ian might have been suggesting earlier.

Luke-Oldenburg:
@claude please review make no mistakes

YodaLightsabr:
> Good overall design. The shared concern approach is clean and the `self.polymorphic_name` trick is exactly right for polymorphic AR associations without a `type` column. A few things worth looking at:

@claude 


Frame 1 - 2026-06-30T121142 370

YodaLightsabr (approved):

app/views/users/admin_details_disbursements.html.erb
```diff
@@ -9,7 +9,7 @@
       </tr>
       <% @disbursements.each do |disbursement| %>
         <tr>
-          <td><%= link_to disbursement.name, hcb_code_path(disbursement.hcb_code), data: { turbo_frame: "_top" } %></td>
+          <td><%= link_to disbursement.name, hcb_code_path(disbursement.outgoinghcb_code), data: { turbo_frame: "_top" } %></td>
```
garyhtou:
```suggestion
          <td><%= link_to disbursement.name, hcb_code_path(disbursement.outgoing_hcb_code), data: { turbo_frame: "_top" } %></td>
```

app/views/events/transfers.html.erb
garyhtou, line 1: Do `transfers.html.erb:136,149` need to be updated from `Disbursement` to `Disbursement::Outgoing`

garyhtou: And maybe `payments.html.erb:134,147` too

app/views/events/transfers.html.erb
```diff
@@ -169,6 +169,8 @@
                       <%= ledger_amount %>
                     </span>
                   <% end %>
+                <% when Disbursement::Outgoing %>
+                  <%= render_money(-d.amount) %>
```
garyhtou, line 173: This logic might need to be added to `payments.html.erb:171`

garyhtou (approved):

PR #14053, opened 2026-06-29
Update spend_only.rb

RhysMP:
## Summary of the problem

We don't want to accidentally charge a fee to a spend only org as the only way they can receive money is via disbursement.


Luke-Oldenburg (approved):
lgtm

PR #14064, opened 2026-06-30
[Ledger::Item] Add cooked transaction eggs with polymorphism

Luke-Oldenburg:
## Summary of the problem

We need a way to store linked objects polymorphicly.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -12,7 +12,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -1577,13 +1577,16 @@
 
   create_table "ledger_items", force: :cascade do |t|
     t.integer "amount_cents", null: false
+    t.bigint "cooked_transaction_eggs_id"
```
garyhtou, line 1580: ahhh, very cooked 🍳

PR #14069, opened 2026-06-30
Fix re-mapping Ledger::Items for admins

YodaLightsabr:
## Summary of the problem

Currently the `AdminController` logic for mapping `Ledger::Items` does not function properly when re-mapping an already-mapped transaction.

## Describe your changes

This PR updates the mapping logic to function properly when the `Ledger::Item` already has a primary ledger mapping.


Luke-Oldenburg (approved):
lgtm

PR #14079, opened 2026-06-30
Swap ledger routes

YodaLightsabr:
## Summary of the problem

We want to use `ledger` for the new ledger but it's currently used for the old ledger.

## Describe your changes

Swap the routes around!


YodaLightsabr (commented):
Approved by Gary

PR #14083, opened 2026-06-30
Fix pagination for ledger

manuthecoder:
image

if it works on my machine, it shall work on your machine too


YodaLightsabr:
Thank you!

YodaLightsabr (approved):

PR #14085, opened 2026-06-30
Recalculate `amount_cents` when `Ledger::Mapping` changes

YodaLightsabr:
## Summary of the problem

One factor used when calculating a `Ledger::Item`'s `amount_cents` is whether or not the `Ledger` `can_front_balance?` for `CanonicalPendingTransaction`s. However, when the `primary_ledger` is assigned, we don't trigger this calculation.

## Describe your changes

Trigger `write_amount_cents!` when `Ledger::Mapping` changes.


Luke-Oldenburg (approved):
lgtm

PR #14086, opened 2026-06-30
Properly fix admin mapped `Ledger::Item`s

YodaLightsabr:
## Summary of the problem

Mapping is broken (we aren't using the right method).

## Describe your changes

This PR fixes mapping using the right method and also passes in the user who mapped it.


YodaLightsabr (commented):
LGTM

PR #14088, opened 2026-06-30
[Spelling] Replace all occurrences of 'a HCB ...' with 'an HCB ...'

mattsoh:
## Summary of the problem

[grammar](https://www.scribbr.com/commonly-confused-words/a-vs-an/)

see specifically 'A or an before acronyms'

and yes I know it seems like I have nothing better to do

do also note that there is inconsistent use of 'a HCB' and 'an HCB'

## Describe your changes

change all 'a HCB ...' to 'an HCB ...'


Luke-Oldenburg (approved):
lgtm

PR #14091, opened 2026-07-01
[Receipts] Allow uploading csv on html page

mattsoh:
## Summary of the problem

Currently dragging in .csv files are allowed but not clicking "upload" and selecting a csv vile

## Describe your changes

Allow csv on html pages

### DIFF
--- a/config/initializers/friendly_id.rb
+++ b/config/initializers/friendly_id.rb
@@ -103,6 +103,7 @@ FriendlyId.defaults do |config|
     mobile
     deprecated
     apply
+    applications
     maintenance_tasks
     org
   ]


Luke-Oldenburg (commented):
My intuition says we shouldn't allow CSVs at all because they are plain text and easily forgeable.

mattsoh:
This is for uploading spreadsheets for things like github sponsors and benevity (also you can already drag and drop CSVs, just not click upload)

Luke-Oldenburg:
I don't think we should allow it at all. If this is for internal use we can attach it as a comment.

PR #14095, opened 2026-07-01
Bug fix ledger item type metadata

YodaLightsabr:
## Summary of the problem

The wrong strings are used in the type metadata object

## Describe your changes

Use the right strings


Luke-Oldenburg (approved):
lgtm

PR #14096, opened 2026-07-01
Bug fix `Ledger::Item#type_metadata`

YodaLightsabr:
## Summary of the problem

The key of the hash is a symbol, but we're passing in a string

## Describe your changes

Symbolize the string with `.to_sym`


Luke-Oldenburg (approved):
lgtm

PR #14098, opened 2026-07-01
[Payments] Hide transfers with payment attempts on payments page

polypixeldev:
## Summary of the problem

Currently, when a payment gets sent, both the payment row and the payout appear on the payments page.

## Describe your changes

Hide any transfers with payment attempts on the payments page


Mohamad-Mortada (approved):
LGTM

PR #14108, opened 2026-07-01
Fix icons and add transaction backgrounds

YodaLightsabr:
## Summary of the problem

We're missing some icons and the transaction colors.

## Describe your changes

Add Stripe transaction icon and transaction colors


app/models/canonical_pending_transaction.rb
```diff
@@ -178,7 +178,7 @@ class CanonicalPendingTransaction < ApplicationRecord
     Ledger::Item.find(old_ledger_item_id).write_amount_cents! if old_ledger_item_id.present?
   end
 
-  def pending_expired?
+  def r
```
Luke-Oldenburg:
```suggestion
  def pending_expired?
```

Luke-Oldenburg (approved):
lgtm

PR #14110, opened 2026-07-01
Prevent async org comboboxes from committing a wrong or blank organization

garyhtou:
## Summary of the problem

The "Send an HCB transfer" **From** / **To** pickers are async `hotwire_combobox` fields. As you type, the gem silently commits the **first server result** to the hidden field even on a *fuzzy* match where the visible text never changes (e.g. typing `bank` commits "HCB Operations"), and it can lock one in on blur. So a transfer could be sent from/to the **wrong (but authorized) organization** without a deliberate choice — which is why users reported "I selected an org" yet the wrong or blank one was submitted. The same gap let a **blank** org through when a form was submitted during the in-flight search, surfacing as a `NoMethodError` 500 in `DisbursementsController#create`.

## Describe your changes

Fix the auto-commit **at the source**, inside the combobox, with no added confirmation step:

- **`StrictHwComboboxController`** (`strict_hw_combobox.js`) — a thin subclass of the gem controller, registered under `hw-combobox`, gated by `data-hw-combobox-strict-value="true"` set only on the transfer org pickers. It wraps `super` and overrides two seams so a value commits only on a **deliberate pick** (click / arrow) or a **genuine prefix match**:
  - `_selectOnQuery` — suppress the silent commit of a fuzzy first result while typing.
  - `_ensurableOption` — never lock in a fuzzy first result on close / Enter / blur.
  It reuses the gem's own case-insensitive `startsWith` test, so prefix typing (`hack club h` → `Hack Club HQ`) keeps its inline-autocomplete with zero friction. Every other combobox (admin filters, etc.) is byte-for-byte unchanged — the override early-returns without the flag.
- **Server stays authoritative** — `#create` resolves source *and* destination symmetrically and returns a friendly error for a blank/unresolved either side instead of 500-ing.
- **Slim client guard** — `disbursement_form_controller.js` blocks the in-flight-race blank submit with an inline, `aria-describedby`-associated message; the server remains the source of truth.
- **Upgrade safety** — pin `@josefarias/hotwire_combobox` to exact `0.4.1`; `connect()` asserts the coupled gem internals still exist and throws (surfaced via AppSignal) on an upgrade rather than silently reverting.

**Testing:** `spec/controllers/disbursements_controller_spec.rb` green (10 examples, incl. a test asserting the strict flag renders through the real gem). Verified live in the browser across every path — fuzzy typing stays blank + `aria-invalid`, prefix commits + autocompletes, typing over a prior pick clears the stale value, Escape/close never locks in a wrong org, explicit click commits. No JS/system-test harness exists in the repo, so the strict behavior is covered by the boot-time contract assertion + manual verification.

### DIFF
--- a/app/controllers/disbursements_controller.rb
+++ b/app/controllers/disbursements_controller.rb
@@ -106,11 +106,14 @@ class DisbursementsController < ApplicationController
   end
 
   def create
-    @source_event = Event.find_by_public_id(disbursement_params[:source_event_id])
-    @destination_event = Event.find_by_public_id(disbursement_params[:event_id]) || Event.friendly.find(disbursement_params[:event_id])
+    @source_event = resolve_transfer_event(disbursement_params[:source_event_id])
+    @destination_event = resolve_transfer_event(disbursement_params[:event_id])
     @disbursement = Disbursement.new(destination_event: @destination_event, source_event: @source_event)
 
     authorize @disbursement
+
+    raise ArgumentError, "You must select an organization to send from." if @source_event.nil?
+    raise ArgumentError, "You must select an organization to send to." if @destination_event.nil?
 
     if admin_signed_in? && disbursement_params["scheduled_on(1i)"].present?
       scheduled_on = Date.new(disbursement_params["scheduled_on(1i)"].to_i,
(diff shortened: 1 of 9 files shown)


garyhtou:
this was all vibe coded, so i have no clue whether it works. haven't gotten a chance to review this myself yet

manuthecoder (changes requested):
this doesnt work. i will create a new PR for a custom stimulus component for comboboxes

garyhtou:
we decided to do a full rewrite of the combobox
https://github.com/hackclub/hcb/pull/14136

PR #14115, opened 2026-07-01
[Admin] Fix stripe card nil bug

manu-gurudath:
Fixes HCB crashing


app/views/stripe_cards/show.html.erb
```diff
@@ -1,4 +1,4 @@
-<% title @card.initially_activated ? "Card #{@card.last_four} (#{@card.user.name}, #{(@event || @card.event).name})" : "Inactive card" %>
+<% title @card.initially_activated ? "Card #{@card.last_four} (#{@card.user&.name}, #{(@event || @card.event).name})" : "Inactive card" %>
```
garyhtou: I feel like all cards should always have a user.

Luke-Oldenburg: It's impossible for it to not. `StripeCard#stripe_cardholder_id` is nonnull and StripeCardholder#user_id` is nonnull.

Luke-Oldenburg (commented):
I think we should determine if there are cards on soft-deleted events and then restore those events.

PR #14119, opened 2026-07-01
Fix transaction_source_type on CPT

YodaLightsabr:
## Summary of the problem

`transaction_source_type` will output an array if any of the missing types are the source.

## Describe your changes

Add the remaining transaction source types and include a `nil` fallback


Luke-Oldenburg (approved):
lgtm

PR #14121, opened 2026-07-01
[Ledger::Item] Add `receipt_required?`

Luke-Oldenburg:
## Summary of the problem

Method returned false

## Describe your changes

Adds logic for determining if a `Ledger::Item` requires receipts.


Luke-Oldenburg (commented):
"lgtm" - @YodaLightsabr

PR #14123, opened 2026-07-02
[Ledger::Item] Add author

Luke-Oldenburg:
## Summary of the problem

We needed a way to fetch the transaction author.

## Describe your changes

Add method to retrieve the transaction author.


garyhtou (approved):
I cry, but good for now lol

PR #14142, opened 2026-07-02
Fix login continue hanging

polypixeldev:
## Summary of the problem

Sometimes when you login to HCB, the redirect after you submit will just hang forever, and you'll need to refresh.

Claude's root cause:
>When Turbo submits the login code form, it uses fetch() with redirect: follow, which transparently chases all server-side redirects. If for any reason the session cookie isn't immediately recognized when GET / is processed (race condition, cookie edge case), the redirect chain ends at auth_users_path?require_reload=true — the login page with <meta name="turbo-visit-control" content="reload">. Turbo sees this, calls reload(), and since visitCompleted was never fired for this visit, this.redirectedToLocation is stale — it falls back to this.location = the login code page URL. The browser reloads the same login page; progress bar was visible, page appears frozen.

## Describe your changes

Disable turbo. Login works in development, although I wasn't able to reproduce the hanging issue in development so I don't know if it really fixes the issue.


garyhtou:
Is this fix good to merge? Would love to have this in

Mohamad-Mortada (approved):

PR #14146, opened 2026-07-02
Fix Account Balance Chart

Famousmaster206:
## Summary of the problem
Currently, on the account balance chart, the account balance for a given day except for the current day is the balance before hcb's 7% fee. Because of this, the graph shows a higher balance on previous days than what the actual balance was, but shows the correct balance for the current day.

## Describe your changes
I modified `\app\javascript\controllers\balance_graph_controller.js` to use the balance with the 7% fee included for previous days, but kept the system where today's balance is the current balance.

### DIFF
--- a/app/javascript/controllers/balance_graph_controller.js
+++ b/app/javascript/controllers/balance_graph_controller.js
@@ -54,10 +54,9 @@ export default class extends Controller {
         const balanceByDate = Object.fromEntries(entries)
 
         for (const date in balanceByDate) {
-          const value =
-            date == today
-              ? this.availableValue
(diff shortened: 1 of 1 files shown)


Luke-Oldenburg (commented):
Hi, thanks for opening a PR! Unfortunately, I don't think this solution will work as not every organization is on a 7% fee plan. For example, many HQ-related organizations are on a 0% plan so this would make the balance chart incorrect for those. I believe that our new transaction engine that is coming soon should fix this issue so I'm going to go ahead and close this PR. If you would like another issue to work on, feel free to reach out in the #hcb-dev channel on Slack!

PR #14147, opened 2026-07-02
Rescue from NotAuthorizedError in events#index

YodaLightsabr:
Handle Pundit::NotAuthorizedError in events controller.


Luke-Oldenburg (approved):
lgtm

PR #14148, opened 2026-07-02
[Tax] Payment flow integration

polypixeldev:
## Summary of the problem

We now have the one time payment flow and the tax form model, but not an integration between them.

## Describe your changes

Integrates tax form filling out into the one time payment flow using AASM and callbacks. Assumes that we are using TaxBandits, but it won't be too difficult to switch to a different platform.

Moves address to tax form (ignores columns, will drop in future PR)

The UI still needs work, but this PR lays the basic groundwork for handling tax forms and legal entities. Next PRs will improve the payee's flow

Env vars needed:
TAXBANDITS__CLIENT_ID=""
TAXBANDITS__CLIENT_SECRET=""
TAXBANDITS__USER_TOKEN=""
TAXBANDITS__CUSTOMIZATION_ID=""


app/models/tax/form.rb
```diff
@@ -14,15 +14,16 @@
...
 #  external_service               :string           not null
 #  failed_at                      :datetime
 #  form_type                      :string
 #  sent_at                        :datetime
+#  signing_url                    :string
 #  taxbandits_status              :string
 #  taxbandits_tin_matching_status :string
 #  created_at                     :datetime         not null
 #  updated_at                     :datetime         not null
-#  external_id                    :string
```
garyhtou, line 25: keep `external_id`, drop `document_url`

app/models/tax/form.rb
```diff
@@ -66,6 +70,90 @@ class Form < ApplicationRecord
...
+        end
+      end
+
+      event :mark_failed do
+        transitions from: :sent, to: :failed
+      end
+    end
+
+    def send!
+      raise ArgumentError, "can only send tax forms when pending" unless pending?
```
garyhtou:
```suggestion
      raise ArgumentError, "can only send tax forms when pending" unless pending? && external_id.blank?
```

app/models/tax/form.rb
```diff
@@ -66,6 +70,90 @@ class Form < ApplicationRecord
...
+
+      send_using_taxbandits! unless sent_with_manual?
```
garyhtou:
```suggestion
      case external_service
      when :tax_bandits
        send_using_taxbandits!
      when :manual
        Rails.logger.info("[Tax::Form] NO-OP: skipping because the external service is 'manual'.")
      else
        raise ArgumentError, "Unable to send tax form using unknown external service (#{external_service})"
      end
```

app/models/tax/form.rb
```diff
@@ -66,6 +70,90 @@ class Form < ApplicationRecord
...
+
+      mark_sent!
+    end
+
+    def taxbandits_submission
+      taxbandits_client.get("WhCertificate/Get?PayeeRef=#{hashid}").body
```
garyhtou: use public id instead of hashid. `tfm_`

app/models/tax/form.rb
```diff
@@ -66,6 +70,90 @@ class Form < ApplicationRecord
...
+    rescue Faraday::ResourceNotFound, Faraday::BadRequestError
+      # TaxBandits will respond with 400 or 404 if the submission is not yet complete
```
garyhtou: should we report this?

app/models/tax/form.rb
```diff
@@ -66,6 +70,90 @@ class Form < ApplicationRecord
...
+      @taxbandits_client || begin
+        Faraday.new(url: Rails.env.development? ? "https://testapi.taxbandits.com/v1.7.3/" : "https://api.taxbandits.com/v1.7.3/") do |faraday|
+          faraday.response :json
+          faraday.response :raise_error
+          faraday.adapter Faraday.default_adapter
+          faraday.headers["Authorization"] = "Bearer #{taxbandits_access_token}"
+          faraday.headers["Content-Type"] = "application/json"
+        end
+      end
+    end
```
garyhtou: _technicallyyyy_, this should be inside a `TaxBanditService.client` 👉👈

along with `taxbandits_access_token`

app/models/tax/identification_number.rb
```diff
@@ -0,0 +1,32 @@
...
+      payments_sum >= Tax::REPORTING_THRESHOLD_1099
+    end
+
+    def banned?
+      legal_entities.any?(&:banned?)
+    end
+
+    def payments
+      p = Payment.joins(:payee).where(payee: { legal_entity: legal_entities })
+      successful = p.successful_or_sent.where("date_trunc('year', sent_at) = ?", "#{Tax.year}-01-01")
```
garyhtou:
```suggestion
      successful = p.successful_or_sent.where("date_part('year', sent_at) = ?", Tax.year)
```

app/models/legal_entity.rb
```diff
@@ -33,18 +30,39 @@ class LegalEntity < ApplicationRecord
...
+  end
+
+  def send_tax_form!
+    form = tax_forms.create!(external_service: :taxbandits)
+    form.send!
+  end
+
+  def banned?
+    banned_reason.present?
+  end
```
garyhtou: drop this method and then inline the logic inside`tax_identification_number.banned?` to prevent accidential usage.

app/models/tax/form.rb
polypixeldev: nope, this is just how their API works - however after looking at their docs I realize this won't work, since this will just return the latest completed submission even if the tax form instance is for a different form

garyhtou (approved):

PR #14153, opened 2026-07-02
[Payments] Add ability to view user's own payments and support for multiple entities

Mohamad-Mortada:
## Summary of the problem
We now have a pay tab. 
image

Also reimbursements payout settings can now be configured for multiple entities not just personal


config/routes.rb
```diff
@@ -153,6 +154,7 @@
       get "payouts", to: "users#edit_payout"
       get "previews", to: "users#edit_featurepreviews"
       get "security", to: "users#edit_security"
+      get "pay", to: "users#pay"
```
polypixeldev: I don't think this route needs to be a user path and in users controller - it should be in my controller like the current payroll tab

app/controllers/users_controller.rb
```diff
@@ -150,6 +151,50 @@ def edit_address
...
     authorize @user
+
+    unless payments_contractors_refresh?
+      @legal_entity = @user.personal_legal_entity
+      return
+    end
+
+    @legal_entities = @user.legal_entities
+    @legal_entity = @legal_entities.find_by(id: params[:legal_entity_id] || session[:payout_legal_entity_id]) || @user.personal_legal_entity
+    session[:payout_legal_entity_id] = @legal_entity.id
```
polypixeldev: I would use the same session key for both the pay and payout settings tabs so that when you go between them, they're on the same LE

app/controllers/application_controller.rb
```diff
@@ -16,6 +16,12 @@ class ApplicationController < ActionController::Base
 
   protect_from_forgery
 
+  # Feature flag gating the payments/contractors refresh (new "Pay" page,
+  # multi-legal-entity payout methods, and the legal entity switcher).
+  helper_method def payments_contractors_refresh?
+    Flipper.enabled?(:payments_contractors_refresh_2026_06_26, current_user)
+  end
```
polypixeldev: Is this needed? I think we'll be enabling this feature on an event basis, not a per-person basis. From what I see, this is mainly used in this PR to always use the personal LE, but since nobody has more than one LE right now, I think we can get rid of this

app/controllers/users_controller.rb
```diff
@@ -383,13 +428,19 @@ def update
       return redirect_back_or_to edit_user_path(@user)
     end
 
+    if payout_method_type.present? && payments_contractors_refresh?
+      @legal_entity = @user.legal_entities.find_by(id: params[:legal_entity_id]) || @user.personal_legal_entity
+      session[:payout_legal_entity_id] = @legal_entity.id if params[:legal_entity_id].present?
+    end
```
polypixeldev: It doesn't have to be this PR but payout method create/update/delete operations should be refactored out of users controller - it doesn't make sense to have it here anymore

app/models/payment.rb
```diff
@@ -84,6 +84,14 @@ def retry!
     create_payment_attempt!
   end
 
+  def payout
+    attempts.order(created_at: :desc).find { |attempt| attempt.payout.present? }&.payout
```
polypixeldev: I'd just take the latest attempt and use that payout, even if it's nil. If a payment failed and got retried, and the retry payout failed to create, we should show that

app/controllers/users_controller.rb
```diff
@@ -150,6 +151,50 @@ def edit_address
...
+      deposited: all_payments.where(aasm_state: "successful").sum(:amount_cents),
+      in_transit: all_payments.where(aasm_state: %w[pending_legal_entity under_review sent]).sum(:amount_cents),
+      canceled: all_payments.where(aasm_state: "rejected").sum(:amount_cents)
+    }
+
+    @payments = all_payments.order(created_at: :desc)
+    @payments = @payments.where(aasm_state: %w[pending_legal_entity under_review sent]) if params[:status] == "in_transit"
+    @payments = @payments.where(aasm_state: "successful") if params[:status] == "deposited"
+    @payments = @payments.where(aasm_state: "rejected") if params[:status] == "canceled"
+    @payments = @payments.search_recipient(params[:q]) if params[:q].present?
```
polypixeldev: Doesn't make sense to search by recipient here - they're all to the same person. Search by event name and purpose

app/controllers/my_controller.rb
```diff
@@ -158,6 +158,42 @@ def payroll
...
+      deposited: all_payments.where(aasm_state: "successful").sum(:amount_cents),
+      in_transit: all_payments.where(aasm_state: %w[pending_legal_entity under_review sent]).sum(:amount_cents),
+      canceled: all_payments.where(aasm_state: "rejected").sum(:amount_cents)
+    }
+
+    @payments = all_payments.order(created_at: :desc)
+    if params[:q].present?
+      query = "%#{Payment.sanitize_sql_like(params[:q])}%"
+      @payments = @payments.joins(payee: :event).where("events.name ILIKE :q OR payments.purpose ILIKE :q", q: query)
+    end
```
polypixeldev: use pgsearch

app/helpers/users_helper.rb
```diff
@@ -61,7 +61,7 @@ def users_nav(selected: nil)
       selected: selected == :reimbursements
     }
 
-    if payments_contractors_refresh?
+    if Flipper.enabled?(:payments_contractors_refresh_2026_06_26)
```
polypixeldev: only show if the user has payments

app/policies/payment_policy.rb
```diff
@@ -3,7 +3,7 @@
 class PaymentPolicy < ApplicationPolicy
   def show?
     return true if user&.auditor?
-    return true if Flipper.enabled?(:payments_contractors_refresh_2026_06_26, user) && record.legal_entity&.users&.include?(user)
+    return true if Flipper.enabled?(:payments_contractors_refresh_2026_06_26) && record.legal_entity&.users&.include?(user)
```
polypixeldev: don't think we need this flipper check

app/models/user.rb
```diff
@@ -150,6 +150,8 @@ class User < ApplicationRecord
   has_many :jobs, as: :entity, class_name: "Employee"
   has_many :job_payments, through: :jobs, source: :payments, class_name: "Employee::Payment"
 
+  has_many :payments, foreign_key: :creator_id, inverse_of: :creator
```
polypixeldev: change to payments received,not created

app/controllers/my_controller.rb
```diff
@@ -158,6 +158,39 @@ def payroll
     @payout_method = current_user.default_payout_method&.details
   end
 
+  def pay
+    return head :not_found unless Flipper.enabled?(:payments_contractors_refresh_2026_06_26)
+
+    if params[:legal_entity_id].present?
+      session[:payout_legal_entity_id] = params[:legal_entity_id]
```
polypixeldev: rename to just legal entity id

polypixeldev (approved):

app/views/users/edit_payout.html.erb
```diff
@@ -13,6 +13,6 @@
         </p>
       <% end %>
     </p>
```
polypixeldev: don't include copy about reimbursement reports when viewing non-personal LEs. you can keep the name as reimbursement payouts for now since we haven't launched payments yet

polypixeldev (commented):
have reimbursement payout method selector link to settings go to the personal LE

polypixeldev (approved):

PR #14157, opened 2026-07-03
Add slack support button

wally-ben:
## Summary of the problem

https://hackclub.enterprise.slack.com/archives/C068U0JMV19/p1782735820811129?thread_ts=1782735820.811129&cid=C068U0JMV19


app/models/card_grant_setting.rb
```diff
@@ -51,9 +52,9 @@ class CardGrantSetting < ApplicationRecord
   }, prefix: :expires_after
 
   def slack_support?
-    return false unless support_url.present?
+    return false unless support_slack_url.present?
 
-    URI.parse(support_url)&.host&.end_with?(".slack.com") || false
+    URI.parse(support_slack_url)&.host&.end_with?(".slack.com") || false
```
garyhtou, line 57: This check has not surfaced as a validation to the user, which means that Slack URLs may be silently hidden if improperly configured.

garyhtou:
@claude, please do a thorough review here using multiple experts.

Consider whether it's worth standardizing or consolidating `support_url` and `support_slack_url`. Previously, `support_url` handled all URL types, but now it's intended to handle only non-Slack URLs (though it still accepts Slack URLs). This inconsistency may be a footgun.

garyhtou:
@claude , u alive? 🫀

garyhtou:
> [!NOTE]
> **This review was written entirely by AI, with no human involvement in its contents.**
> @garyhtou asked Claude (Opus 4.8, running in Claude Code) to run this review and post it, but he has not read, edited, or vetted anything below. It is machine-generated and may be wrong. Please verify before acting on it, and push back freely.

<sub>Also, on why the `@claude` bot went quiet on this PR: it is not your fault, and not a config problem. It is upstream bug [anthropics/claude-code-action#1423](https://github.com/anthropics/claude-code-action/issues/1423). This PR is a fork PR whose head branch is named `main`, and the base branch is also `main`, so the action runs `git fetch origin --depth=20 pull/14157/head:main` while `main` is already checked out on the runner, and git refuses. Two community fix PRs are open upstream, neither merged. Hence this review being run manually instead.</sub>

---

## Summary

The migration, the column, the strong params, and the detector all landed correctly. **The view that consumes them did not.** `support_slack_url` is written and read for *detection*, but it is never used as a link target anywhere in the app. Outside the model, controller, and settings form, it appears exactly once: the visibility guard on line 3 of `_support.html.erb`.

The net effect is that the feature this PR exists to deliver does not work, and because there is no test coverage for any of this code, CI stays green.

---

## Blocking

### 1. A Slack-only configuration renders a dead button

`app/views/card_grants/actions/_support.html.erb:3,16`

Line 3 was widened to admit `support_slack_url.present?`, but line 16 still links `support_url`:

```erb
<%= link_to card_grant.setting.support_url, target: "_blank", class: "btn bg-info btn--list-item" do %>
  <% if card_grant.setting.slack_support? %>
```

Set only "Slack channel URL" and leave "Support email" blank, which is the headline use case for this PR. The guard passes, `slack_support?` is true, so the recipient gets a Slack icon and "Contact `{Org}` on Slack". But the `href` is `url_for(nil)`, which regenerates the current request's path. Verified by running it:

```
$ bin/rails runner '...v.link_to(nil, target: "_blank", class: "...") { "Contact Foo on Slack" }'
<a target="_blank" class="btn bg-info btn--list-item" href="/grants/cdg_abc123">Contact Foo on Slack</a>
```

The recipient clicks a confidently Slack-branded button and reopens the page they are already on. No error, no clue.

### 2. With both fields set, the icon and label contradict the destination

`app/views/card_grants/actions/_support.html.erb:16-33,48-63`

`slack_support?` now reads `support_slack_url` while `email_support?` reads `support_url`, so **both can be true at once**, which was impossible before this PR when both derived from one column. The `if slack_support? / elsif email_support?` ordering picks Slack for the icon and the copy, but the `href` is hardcoded to `support_url`. So with `support_url = "mailto:..."` and a Slack URL set, which is exactly what the new form instructs, you get a Slack icon reading "Contact `{Org}` on Slack" that opens a mail composer.

Related and more fundamental: the view renders **at most one link**. Letting an organization offer both an email and a Slack channel at once appears to be the point of the PR, and the UI has no way to express it.

### 3. With a support message set, the Slack link is unreachable

`app/views/card_grants/actions/_support.html.erb:45`

The modal footer is still gated on `support_url.present?`. Configure a support message plus a Slack URL and no email, and the modal renders the message with no action button at all. The Slack channel is not surfaced anywhere in the product. This is a likely real configuration, since organizations already write "find us on Slack" into the support message.

### Full render matrix

M = `support_message`, U = `support_url`, S = `support_slack_url`

| M | U | S | Renders | href | OK? |
|---|---|---|---|---|---|
| . | . | . | nothing | . | ok |
| . | mailto | . | email icon, "Email X" | mailto | ok |
| . | https (non-Slack) | . | help icon, "Get in touch" | that URL | ok |
| . | . | slack | **Slack icon, "Contact X on Slack"** | **current page** | **#1** |
| . | mailto | slack | **Slack icon, "Contact X on Slack"** | **mailto** | **#2** |
| msg | any | . | modal, footer iff U present | U | ok |
| msg | . | slack | modal, **no button at all** | . | **#3** |
| msg | mailto | slack | modal, footer **Slack icon/copy** | **mailto** | **#2** |

---

## Major

### 4. No backfill, so existing organizations silently lose Slack support

Before this PR, `support_url` *was* the Slack field. The old form copy said so: "HCB will automatically detect email addresses and Slack channel links." Any organization that configured Slack support has that URL in `support_url`.

After this PR, for those rows: `slack_support?` becomes false (new column is null), `email_support?` is false (not a `mailto:`), so they fall to the generic branch. The Slack icon and "Contact `{Org}` on Slack" copy disappear and become a generic help icon and "Get in touch with `{Org}` for support". The link still works, so this fails quietly.

The interaction with the relabel is the worst part: the organizer now sees their working Slack URL sitting in a field labeled **"Support email"** with a `mailto:` placeholder telling them it is wrong. The obvious corrective action, moving it to the new field, lands them in bug #1 and breaks the button entirely. **The PR steers people into the broken state.**

This repo has `maintenance_tasks` for exactly this, and `app/jobs/one_time_jobs/README.md` says to use it. Sketch, to deploy *after* the view fix:

```ruby
module Maintenance
  class BackfillCardGrantSettingSupportSlackUrlTask < MaintenanceTasks::Task
    def collection
      CardGrantSetting.where(support_slack_url: nil).where.not(support_url: [nil, ""])
    end

    def process(setting)
      host = URI.parse(setting.support_url).host rescue nil
      return unless host&.end_with?(".slack.com")

      setting.update_columns(support_slack_url: setting.support_url, support_url: nil)
    end
  end
end
```

Move rather than copy: `support_url` now means email, so leaving a Slack URL there keeps the ambiguity alive. Note `CardGrantSetting` has `has_paper_trail`, so decide deliberately between `update_columns` (no version) and `update!` (a version per row).

### 5. "Support email" plus no validation invites a broken relative link

`app/views/events/settings/_card_grants.html.erb:109-114`

There are no validations on this model beyond `validates :event, uniqueness: true`. A field labeled "Support email" will get `orpheus@hackclub.com` typed into it. `url_for` passes strings through verbatim, so that renders `href="orpheus@hackclub.com"`, a **relative** URL resolving to `/card_grants/cdg_.../orpheus@hackclub.com` and a 404. `email_support?` is also false, so no email icon either.

The risk existed before, but the old label ("Support URL") and hint made the URL shape clear. The new label invites the broken input while removing the hint.

Nothing constrains `support_slack_url` to be a Slack link either. A Discord invite, or a schemeless `hackclub.slack.com` (which parses to `host == nil`), makes `slack_support?` false and the value silently inert forever.

---

## Minor

- **Migration timestamp is backdated.** `20260703123844` sits ~18 days behind `db/schema.rb`'s `2026_07_20_210532`, with 24 newer migrations on `main`. Not broken (schema version line is correctly the max, and `db:migrate` runs out-of-order versions fine), but two consequences: `rails db:rollback` will undo the *wrong* migration, and the `reversible_migrations` CI job computes `ROLLBACK_TO` from the oldest new migration, so it will roll back and re-apply 25 migrations instead of 1. It still passes, just with a much larger blast radius. Renaming the file to a current timestamp fixes both.
- **Indentation.** `_card_grants.html.erb:117-124`, the new `<div class="field mb-0">` block, sits at column 0 inside an 8-space context, and line 111's hint text is at 4 where every other prose line in the file is at 12. Worth noting: I ran `bundle exec erb_lint` against the PR version of this file and it reports "No errors were found", so **CI will not catch this**. erb_lint 0.9.0's `space_indentation` linter only flags tabs.
- **`email_support?` returns `nil`, not `false`** (`card_grant_setting.rb:63`), asymmetric with `slack_support?`, which is careful to `|| false`. Harmless in an `elsif` today, but the two predicates are now the sole discriminators for icon and copy, so they should have matching contracts.
- **Copy.** "Enter a mailto: link for email support." puts developer jargon in front of organizers, and a bare address typed in response silently fails. "Paste the Slack channel link..." describes mouse mechanics rather than the setting, and both break the declarative voice of every neighbouring hint. The placeholder `C12345` is also implausibly short for a real Slack channel ID (they are 9 to 11 characters). Suggestions: label "Slack channel link", hint "If your organization has a Slack channel for grant recipients, paste its link here."

---

## Tests

There are none in this PR, and there is no existing coverage to lean on. Verified: `grep -rn "support_url\|slack_support\|email_support\|support_message\|support_slack" spec/` returns **zero matches**. There is no `spec/models/card_grant_setting_spec.rb`, no `card_grants#show` spec (so `_support.html.erb` has never been rendered in a test), and no `events#edit`/`#update` spec (so both permit lists are untested). The factory is 6 lines with no support traits.

That absence is why bugs #1 through #3 are invisible to CI. The house pattern here is controller specs with `render_views` asserting on `response.parsed_body`. Three examples would have caught all three blockers:

```ruby
it "links to the Slack channel when only Slack support is configured" do
  card_grant = card_grant_with(support_slack_url: "https://hackclub.slack.com/archives/C01ABC2DEFG")

  get(:show, params: { id: card_grant.hashid })

  link = response.parsed_body.at_css("a[href='https://hackclub.slack.com/archives/C01ABC2DEFG']")
  expect(link).to be_present
  expect(link.text).to include("on Slack")
end
```

plus one asserting both hrefs are present when both fields are set, and one asserting the modal exposes the Slack link when a support message is set.

---

## On consolidating `support_url` and `support_slack_url`

@garyhtou asked whether these should be standardized or consolidated, and flagged the inconsistency as a footgun. That instinct is correct, and the three blockers above are the evidence for it. **They all exist because the model moved from "sniff the type out of one string" to "the type is which column it is in", while the view still asks the old question.** Half-migrating between those two designs is worse than either endpoint.

Options considered:

| Option | Verdict |
|---|---|
| Keep one polymorphic `support_url`, auto-detect (status quo) | Rejected: cannot express "both", which is the requirement. |
| PR as written: `support_url` (de facto email) + `support_slack_url`, neither validated | Worst of both. Meaning enforced only by a `<label>`. Wrong-field input is silently discarded. |
| Two typed columns: `support_email` + `support_slack_url` | Good, but drops the generic-URL escape hatch that `main` supports today (Notion, Zendesk, a Google Form). Needs a destructive migration. |
| **Three typed columns: `support_email` + `support_slack_url` + `support_url` (generic)** | **Recommended.** |
| Normalized `support_links` table, or a JSONB blob | Over-built for a fixed set of 3 with no requirement for more. The table needs nested attributes plus repeatable-row UI that this settings page has no pattern for; the JSONB version trades away every integrity guarantee. |

The recommendation is the three-column version, for two reasons:

1. **The bug class disappears structurally.** Once each channel owns a column, the view iterates configured channels and takes the icon, label, and `href` from the same entry. "Slack icon on a mailto link" stops being expressible, rather than being a thing you fix and then re-break later.
2. **Nothing is destructive and nothing is orphaned.** `support_url` keeps its column and its existing generic rendering; it just narrows from "any of three things" to "the other thing". Forward-only and safely revertible.

Shape of it:

```ruby
# card_grant_setting.rb
normalizes :support_email, with: ->(v) { v.presence&.strip&.downcase }

validates_email_format_of :support_email, allow_nil: true, allow_blank: true
validate :support_slack_url_must_be_a_slack_link

def email_support?   = support_email.present?
def slack_support?   = support_slack_url.present?
def generic_support? = support_url.present?
def support_email_href = ("mailto:#{support_email}" if email_support?)
```

Storing a **bare address** in `support_email` and letting the view emit the `mailto:` removes the single biggest source of user error from finding #5. Then the view builds a list and renders one button per configured channel, and the modal footer gates on `channels.any?` instead of `support_url.present?`.

Ordering matters if you go this way: **the backfill has to land before the validations**, or existing rows become unsaveable and any organizer editing an unrelated card grant setting gets a validation error on a field they never touched (`accepts_nested_attributes_for` saves the whole record).

One open question worth answering before choosing between the two-column and three-column versions: how many existing rows hold a `support_url` that is neither a Slack link nor a `mailto:`? If that is zero, the simpler two-column design wins.

---

<sub><b>Method:</b> six independent Claude agents were dispatched: two to verify the root cause of the CI failure (one blind, one adversarial, tasked with refuting it), and four to review this diff on separate lenses (render-path correctness, data migration and backward compatibility, UI/copy/style, and tests plus API design). The four reviewers independently converged on findings #1 through #3 without seeing each other's work. Claims marked "verified" were checked by executing code or reading library source, not inferred. The adversarial verifier also caught and corrected an error in an earlier proposed fix for the CI issue.</sub>

PR #14162, opened 2026-07-06
Create backfill HcbCode event IDs maintenance task

YodaLightsabr:
## Summary of the problem

HcbCode event_ids need to be backfilled because our production data is not the most accurate. Inaccurate production data will mess with our migration to the new transaction engine.

## Describe your changes

Create a maintenance task for backfilling HcbCode event_id.


Luke-Oldenburg (approved):
lgtm

PR #14169, opened 2026-07-06
Create CardCharge linked_object

YodaLightsabr:
## Summary of the problem

We need to have linked objects for every single `Ledger::Item`. For transactions that would have traditionally been a `HCB-000-`, we've decided to use the `CanonicalTransaction`'s `Raw<***>Transaction` to satisfy the linked object requirement. However, Stripe card charges, or `HCB-600-`, present an interesting edge case. 

## Describe your changes

This PR creates the `CardCharge` model, which serves as to group together up to one `RawPendingStripeTransaction` and any number of `RawStripeTransactions` in a format compatible with `Ledger::Item`'s linked object.


app/models/card_charge.rb
```diff
@@ -0,0 +1,85 @@
...
+    (raw_stripe_transactions.last || raw_pending_stripe_transaction)&.stripe_transaction&.dig("merchant_currency")
+  end
+
+  # Finds the charge for a Stripe authorization ID (iauth_...), whether it was
+  # first seen as an authorization or as a settled transaction.
+  def self.matching_stripe_authorization_id(stripe_authorization_id)
+    return nil if stripe_authorization_id.blank?
+
+    joins(:raw_pending_stripe_transaction)
+      .find_by(raw_pending_stripe_transactions: { stripe_transaction_id: stripe_authorization_id }) ||
```
garyhtou: so, it appears this method takes in either a Stripe **authorization** id or a Stripe **transaction** id. Should we rename the param to make it more clear? `stripe_authorization_or_transaction_id`

app/models/card_charge.rb
```diff
@@ -0,0 +1,85 @@
...
+      joins(:raw_stripe_transactions)
+        .find_by(raw_stripe_transactions: { stripe_authorization_id: })
```
garyhtou: Is it more efficient to do


```suggestion
    rpst = joins(:raw_pending_stripe_transaction)
      .where(raw_pending_stripe_transactions: { stripe_transaction_id: stripe_authorization_id })
    rst = joins(:raw_stripe_transactions)
      .where(raw_stripe_transactions: { stripe_authorization_id: })

    rpst.or(rst).first
```

app/models/card_charge.rb
```diff
@@ -0,0 +1,85 @@
...
+    (raw_stripe_transactions.last || raw_pending_stripe_transaction)&.stripe_transaction&.dig("merchant_data")
+  end
+
+  def merchant_currency
+    (raw_stripe_transactions.last || raw_pending_stripe_transaction)&.stripe_transaction&.dig("merchant_currency")
+  end
+
+  # Finds the charge for a Stripe authorization ID (iauth_...), whether it was
+  # first seen as an authorization or as a settled transaction.
+  def self.matching_stripe_authorization_id(stripe_authorization_id)
```
garyhtou:
```suggestion
  def self.find_by_stripe_authorization_id(stripe_authorization_id)
```

app/models/card_charge.rb
```diff
@@ -0,0 +1,85 @@
...
+      create!(raw_pending_stripe_transaction:)
+    end
+  end
+
+  def self.link_raw_stripe_transaction!(raw_stripe_transaction)
+    existing = raw_stripe_transaction.card_charge
+    return existing if existing
+
+    # Force captures have no authorization, so they get a charge of their own.
+    charge = matching_stripe_authorization_id(raw_stripe_transaction.stripe_authorization_id)
```
garyhtou: In `link_raw_pending_stripe_transaction`, i see that we consider the result of `matching_stripe_authorization_id` as `existing`. should we do that here too? if not, why are they different?

app/models/raw_stripe_transaction.rb
```diff
@@ -46,6 +50,20 @@ def refund?
...
+    card_charges.first
+  end
+
+  def link_card_charge!
+    CardCharge.link_raw_stripe_transaction!(self)
+  end
+
+  def stripe_card
+    @stripe_card ||= StripeCard.find_by(stripe_id: stripe_card_id)
+  end
```
garyhtou, line 60: This sounds like an association to me lol

app/models/raw_stripe_transaction.rb
```diff
@@ -46,6 +50,20 @@ def refund?
     stripe_transaction["type"] == "refund"
   end
 
+  # The join table enforces at most one charge per transaction, so the HABTM
+  # association (required for the model-less join table) is singular in practice.
+  def card_charge
+    card_charges.first
+  end
```
garyhtou: rip, i didn't know this. maybe we should create that join model so that we can do a has_one in case we ever need to preload this association

app/services/pending_transaction_engine/raw_pending_stripe_transaction_service/stripe/import.rb
```diff
@@ -21,6 +21,11 @@ def run
             }
           }, unique_by: :stripe_transaction_id)
 
+          # upsert_all skips callbacks, so link charges for newly inserted rows
```
garyhtou: Maybe we shouldn't upsert all then. Adding this line after kinda feels like a hack. the proper fix may be the run the callbacks

app/models/ledger/item.rb
```diff
@@ -258,6 +260,7 @@ def type_metadata
         "Wire": ["Wire", "web"],
         "WiseTransfer": ["Wise transfer", "wise"],
         "StripeServiceFee": ["Stripe service fee", "cash"],
+        "CardCharge": ["Card charge", "card"],
         "RawPendingStripeTransaction": ["Card charge", "card"],
         "RawStripeTransaction": ["Card charge", "card"]
```
Luke-Oldenburg, line 327: drop this?

app/models/ledger/item.rb
Luke-Oldenburg, line 213: swap this for "CardCharge"?

app/models/ledger/item.rb
YodaLightsabr: We should be able to drop this after production data is migrated

app/models/ledger/item.rb
YodaLightsabr: Ty—I've added CardCharge but I'm keeping those in for now until after the migration

app/services/pending_transaction_engine/raw_pending_stripe_transaction_service/stripe/import.rb
YodaLightsabr: What would you suggest running instead?

app/models/ledger/item.rb
```diff
@@ -270,6 +276,14 @@ def stripe_merchant
 
     private
 
+    def assign_linked_object!
+      return if linked_object.present?
+
+      linked_object = (canonical_pending_transactions.map(&:linked_object) + canonical_transactions.map(&:linked_object)).compact.first
```
garyhtou: Design-wise, CT do not have linked objects. When `canonical_pending_transactions.map(&:linked_object)` doesn't give us anything, create a `TripleZero`.

We need to decide what happens when a TripleZero gets orphaned.

app/models/canonical_pending_transaction.rb
```diff
@@ -277,13 +277,15 @@ def linked_object
     return raw_pending_bank_fee_transaction.bank_fee if raw_pending_bank_fee_transaction
     return raw_pending_incoming_disbursement_transaction.incoming_disbursement if raw_pending_incoming_disbursement_transaction
     return raw_pending_outgoing_disbursement_transaction.outgoing_disbursement if raw_pending_outgoing_disbursement_transaction
+    return raw_pending_stripe_transaction.card_charge if raw_pending_stripe_transaction
```
garyhtou, line 282: why is this line needed? should card charge handle this?

app/models/raw_stripe_transaction.rb
garyhtou: I think this can be an association now?

app/services/pending_transaction_engine/raw_pending_stripe_transaction_service/stripe/import.rb
garyhtou: find or create by

app/services/transaction_engine/syntax_sugar_service/linked_object.rb
garyhtou, line 1: undo changes to this - and ensure LI never calls `CanonicalTransaction#linked_object`

My goal would be to get rid of syntax sugar service at some point

app/tasks/maintenance/backfill_card_charge_ledger_items_task.rb
```diff
@@ -0,0 +1,30 @@
...
+    class AnomalyError < StandardError; end
+
+    def collection
+      CardCharge.where.missing(:ledger_item)
+    end
+
+    def process(card_charge)
+      ledger_item = card_charge.raw_pending_stripe_transaction&.canonical_pending_transaction&.ledger_item
+      ledger_item ||= card_charge.raw_stripe_transactions.filter_map { |rst| rst.canonical_transaction&.ledger_item }.first
+      return if ledger_item.nil?
```
garyhtou: if it is nil, that sounds like an anomaly to me

app/models/canonical_pending_transaction.rb
```diff
@@ -277,13 +277,15 @@ def linked_object
...
     return increase_check if increase_check
     return paypal_transfer if paypal_transfer
     return wire if wire
     return wise_transfer if wise_transfer
     return check_deposit if check_deposit
     return reimbursement_expense_payout if reimbursement_expense_payout
     return reimbursement_payout_holding if reimbursement_payout_holding
+    return card_charge if card_charge
```
garyhtou: where is `card_charge` defined?

app/models/canonical_pending_transaction.rb
YodaLightsabr: The CPT is still only directly associated to the RPST, but the CPT needs to be able to point to the CC when asked for the LO by the LI.

app/services/transaction_engine/syntax_sugar_service/linked_object.rb
YodaLightsabr: Why do we even have a syntax sugar service 🥀

app/models/canonical_pending_transaction.rb
YodaLightsabr: It isn't 😶

app/models/card_charge.rb
```diff
@@ -0,0 +1,86 @@
...
+
+# == Schema Information
+#
+# Table name: card_charges
+#
+#  id                                :bigint           not null, primary key
+#  created_at                        :datetime         not null
+#  updated_at                        :datetime         not null
+#  raw_pending_stripe_transaction_id :bigint
+#
```
garyhtou, line 11: At some point, we want to add a `merchant_network_id` to CardCharge.

app/models/ledger/item.rb
garyhtou: Update: After discussions with the entire team, we decided that some CTs will have linked objects. Only CTs that are mapped using their transaction source raw transaction data will have a linked object. Since most CTs are mapped by shortcodes, those will not return a linked object.

app/models/ledger/item.rb
```diff
@@ -274,6 +281,14 @@ def stripe_merchant
...
```
garyhtou:
```suggestion
      linked_object = (canonical_pending_transactions.order(date: :asc).map(&:linked_object) + canonical_transactions.order(date: :asc).map(&:linked_object)).compact.first
```

app/models/card_charge.rb
```diff
@@ -0,0 +1,86 @@
...
+# Foreign Keys
+#
+#  fk_rails_...  (raw_pending_stripe_transaction_id => raw_pending_stripe_transactions.id) ON DELETE => nullify
+#
+# Raw objects are matched to their charge purely by Stripe IDs: a
+# RawPendingStripeTransaction's `stripe_transaction_id` and a
+# RawStripeTransaction's `stripe_authorization_id` both hold the Stripe
+# authorization ID (iauth_...).
+class CardCharge < ApplicationRecord
```
garyhtou, line 24: Do we need some sort of validation here or in `CardChargeRawStripeTransaction` that ensures that all associated transactions belong to the same stripe ID?

app/services/transaction_engine/syntax_sugar_service/linked_object.rb
```diff
@@ -179,6 +183,16 @@ def wire
         @canonical_transaction.transaction_source if @canonical_transaction.transaction_source_type == Wire.name
       end
 
+      def wise
+        @canonical_transaction.transaction_source if @canonical_transaction.transaction_source_type == WiseTransfer.name
+      end
```
garyhtou: I think we decided that CTs will not return LOs or Ys. Can you double-check whether this change is still needed?

app/models/card_charge.rb
YodaLightsabr: That's not a bad idea. @copilot want to give this a go?

YodaLightsabr: I guess not 💀

image

app/tasks/maintenance/backfill_card_charge_ledger_items_task.rb
YodaLightsabr: Good point

app/models/ledger/item.rb
Luke-Oldenburg: CardCharge still needs to be added for author

app/models/ledger/item.rb
```diff
@@ -277,6 +284,18 @@ def stripe_merchant
...
+      # Once a linked object is assigned, it should never be changed.
+      # In the event of a merger of ledger items (e.g. mapping a CT to an LI with an existing CPT),
+      # the ledger item with the CPT will persist, and the ledger item with the CT will be destroyed.
+      # No linked objects will be changed.
+      return if linked_object.present?
+
+      linked_object = (canonical_pending_transactions.order(date: :asc).map(&:linked_object) + canonical_transactions.order(date: :asc).map(&:linked_object)).compact.first
+
+      update!(linked_object:) if linked_object.present?
+    end
```
Luke-Oldenburg, line 299: I'm not sure that we should enforce this so strictly. We have enough production data issues that there may be a scenario where we want to keep the LO from the CT.

app/models/card_charge.rb
```diff
@@ -0,0 +1,86 @@
...
+#  index_card_charges_on_raw_pending_stripe_transaction_id  (raw_pending_stripe_transaction_id) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (raw_pending_stripe_transaction_id => raw_pending_stripe_transactions.id) ON DELETE => nullify
+#
+# Raw objects are matched to their charge purely by Stripe IDs: a
+# RawPendingStripeTransaction's `stripe_transaction_id` and a
+# RawStripeTransaction's `stripe_authorization_id` both hold the Stripe
+# authorization ID (iauth_...).
```
Luke-Oldenburg, line 23: In the case of force captures I don't think it is an `iauth_...`

app/models/card_charge.rb
```diff
@@ -0,0 +1,86 @@
...
+    (raw_stripe_transactions.last || raw_pending_stripe_transaction)&.stripe_transaction&.dig("merchant_currency")
+  end
+
+  # Finds the charge for a Stripe authorization ID (iauth_...), whether it was
+  # first seen as an authorization or as a settled transaction.
+  def self.find_by_stripe_authorization_id(stripe_authorization_id)
+    return nil if stripe_authorization_id.blank?
+
+    joins(:raw_pending_stripe_transaction).find_by(raw_pending_stripe_transactions: { stripe_transaction_id: stripe_authorization_id }) ||
+      joins(:raw_stripe_transactions).find_by(raw_stripe_transactions: { stripe_authorization_id: })
```
Luke-Oldenburg, line 53: I don't know that we should call this stripe authorization id in the case of non-pending raw transactions because it may not have an authorization.

app/models/raw_stripe_transaction.rb
```diff
@@ -16,11 +16,17 @@
...
-#  index_raw_stripe_transactions_on_card_id_text  ((((stripe_transaction -> 'card'::text) ->> 'id'::text))) USING hash
+#  index_raw_stripe_transactions_on_card_id_text             ((((stripe_transaction -> 'card'::text) ->> 'id'::text))) USING hash
+#  index_raw_stripe_transactions_on_stripe_authorization_id  (stripe_authorization_id)
 #
 class RawStripeTransaction < ApplicationRecord
   has_many :hashed_transactions
   has_one :canonical_transaction, as: :transaction_source
+  has_one :card_charge_raw_stripe_transaction
+  has_one :card_charge, through: :card_charge_raw_stripe_transaction
+  has_one :stripe_card, through: :card_charge
```
Luke-Oldenburg: `stripe_card` is a method not a relation

app/models/card_charge.rb
YodaLightsabr: In the case of force captures there would not be a `stripe_authorization_id`

app/models/ledger/item.rb
Luke-Oldenburg, line 292: We should explain that when the ledger item is destroyed, it will automatically destroy the associated LO if it is a triple zero. Otherwise it will error and not allow you to merge.

garyhtou (approved):
🪄

PR #14175, opened 2026-07-06
[v4] Add transfer create scope

Mohamad-Mortada:
Adds disbursement create scope


app/controllers/api/v4/disbursements_controller.rb
```diff
@@ -26,6 +26,8 @@ def create
         render :show, status: :created, location: api_v4_transaction_path(@disbursement)
       end
 
+      require_oauth2_scope "disbursements:write", :create
```
Mohamad-Mortada:
```suggestion
      require_oauth2_scope "transfers:write", :create
```

polypixeldev (approved):

PR #14198, opened 2026-07-07
[Ledger] Improve mapping

Luke-Oldenburg:
Calls map on `after_create` and `after_touch`. This is because map already calls `refresh!` so we get the added benefit of mapping. Also call refresh on all items after can front balance changes on event.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -475,7 +475,7 @@ class Event < ApplicationRecord
   end
 
   after_update if: -> { can_front_balance_changed? } do
-    ledger.update_all!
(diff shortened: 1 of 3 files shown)


app/models/event.rb
```diff
@@ -474,6 +474,10 @@ def ancestor_users
     build_plan(type: fallback_plan_class) if plan.nil?
   end
 
+  after_update if: -> { can_front_balance_changed? } do
+    ledger.update_all!
```
```suggestion
    ledger.refresh_all!
```

YodaLightsabr (approved):

PR #14200, opened 2026-07-08
[Meow] put your cat into a good nap

garyhtou:


Luke-Oldenburg (dismissed):
"we never remove temporary code"

garyhtou:
> "we never remove temporary code"

see! i was right hahaha

garyhtou:
Okay, I chatted with Fox (since I recently learned they're using this as a prize for their YSWS) - end of July is a good end date for this easter egg

garyhtou:
this better drive TUBs lol

YodaLightsabr (approved):

YodaLightsabr:
LGTM

manuthecoder (approved):
lgtm!

PR #14201, opened 2026-07-08
[Ledger::Item] Add author caching

Luke-Oldenburg:
## Summary of the problem

We need to cache author for filtering and for the missing receipts page.

## Describe your changes

Adds author caching.


db/migrate/20260708134154_add_author_to_ledger_items.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddAuthorToLedgerItems < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :ledger_items, :author, index: { algorithm: :concurrently }
+
+    add_foreign_key :ledger_items, :users, column: :author_id, validate: false
```
garyhtou, line 9: you're missing a migration to validate this FK

Luke-Oldenburg:
approved by @garyhtou

PR #14213, opened 2026-07-08
[Contract] Create models for Contract and Invoice

Mohamad-Mortada:
## Summary of the problem
Can't make contract with no contract?

## Describe your changes
Creates Contract model which has many Invoices


polypixeldev (commented):
Add currency to both contract and invoice, and validate that they are the same. While there are still some questions about how we will handle currency, I think we should have it.

app/models/payroll/contract.rb
```diff
@@ -0,0 +1,80 @@
...
+
+# == Schema Information
+#
+# Table name: payroll_contracts
+#
+#  id            :bigint           not null, primary key
+#  aasm_state    :string           not null
+#  currency      :string           default("USD"), not null
+#  end_date      :date             not null
+#  expired_at    :datetime
```
garyhtou: Drop this column.

app/models/payroll/position.rb
```diff
@@ -0,0 +1,79 @@
...
+#
+#  fk_rails_...  (contract_id => contracts.id)
+#  fk_rails_...  (payee_id => payees.id)
+#
+module Payroll
+  class Position < ApplicationRecord
+    include AASM
+
+    belongs_to :payee
+    belongs_to :contract, optional: true
```
garyhtou: this needs to be a `contractable`

https://github.com/hackclub/hcb/blob/09903163e0d48055496a2dd280ec8e3ae089955e/app/models/concerns/contractable.rb#L4

you may choose to define a `has_one :contract` here, like what [`Event::Application` does](https://github.com/hackclub/hcb/blob/48888468b7140b50c4c4cb2961f6803b68ebaf8f/app/models/event/application.rb#L80); maybe it makes sense to move this `has_one :contract` into the concern

app/models/payroll/position.rb
```diff
@@ -0,0 +1,79 @@
...
+#  index_payroll_positions_on_contract_id  (contract_id)
+#  index_payroll_positions_on_payee_id     (payee_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (contract_id => contracts.id)
+#  fk_rails_...  (payee_id => payees.id)
+#
+module Payroll
+  class Position < ApplicationRecord
```
garyhtou, line 32: Add paper trail to this.

app/models/payroll/invoice.rb
```diff
@@ -0,0 +1,70 @@
...
+#  index_payroll_invoices_on_reviewed_by_id       (reviewed_by_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (payment_id => payments.id)
+#  fk_rails_...  (payroll_position_id => payroll_positions.id)
+#  fk_rails_...  (reviewed_by_id => users.id)
+#
+module Payroll
+  class Invoice < ApplicationRecord
```
garyhtou, line 34: Add paper trail to this.

app/models/payroll/position.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    belongs_to :payee
+    belongs_to :contract, optional: true
+
+    has_many :invoices, class_name: "Payroll::Invoice", foreign_key: "payroll_position_id", inverse_of: :payroll_position, dependent: :destroy
+
+    monetize :rate_cents, with_model_currency: :currency
+
+    validate :end_date_after_start_date
+
```
garyhtou, line 54: `after_create_commit`, schedule a job to be run at the end of the day on the `end_date`. That job will check if this position is still `onboarded`, then `mark_expired`.

app/models/payroll/position.rb
garyhtou, line 1: We don't need this now, but at some point we may want to add a payable method. The payable method would be: you're on the onboarded state; and maybe other checks

app/models/payroll/position.rb
```diff
@@ -0,0 +1,79 @@
...
+      event :mark_onboarded do
+        transitions from: :onboarding, to: :onboarded
+      end
+
+      event :mark_expired do
+        transitions from: :onboarded, to: :expired
+      end
+
+      event :mark_terminated do
+        transitions from: [:under_review, :onboarding, :onboarded], to: :terminated
```
garyhtou: What's at a rejected state, and you can only get to it when you're under review or onboarding. You can only be terminated if you were previously onboarded.

garyhtou: For the future, we should probably add a rejection reason, but not now.

app/models/payroll/position.rb
```diff
@@ -0,0 +1,79 @@
...
+    include AASM
+
+    belongs_to :payee
+    belongs_to :contract, optional: true
+
+    has_many :invoices, class_name: "Payroll::Invoice", foreign_key: "payroll_position_id", inverse_of: :payroll_position, dependent: :destroy
+
+    monetize :rate_cents, with_model_currency: :currency
+
+    validate :end_date_after_start_date
```
garyhtou, line 51: Add a validation that your start date cannot be more than six months in the future and your total duration cannot be longer than one year. This would mean that your end date cannot be greater than 1.5 years from now, although we don't need a specificly validate for that.

db/migrate/20260708160000_create_payroll_positions_and_invoices.rb
```diff
@@ -0,0 +1,41 @@
+class CreatePayrollPositionsAndInvoices < ActiveRecord::Migration[8.0]
+  def change
+    create_table :payroll_positions do |t|
+      t.belongs_to :payee, null: false, foreign_key: true
+      t.belongs_to :contract, foreign_key: true
```
garyhtou: This will need to be dropped in favor of the contractable concern.

db/migrate/20260708160000_create_payroll_positions_and_invoices.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      t.text :title, null: false
+      t.text :purpose, null: false
```
garyhtou: Please rename this to `description`

db/migrate/20260708160000_create_payroll_positions_and_invoices.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    create_table :payroll_invoices do |t|
+      t.belongs_to :payroll_position, null: false, foreign_key: true
+      t.belongs_to :reviewed_by, foreign_key: { to_table: :users }
+      t.belongs_to :payment, foreign_key: true
+
+      t.text :name, null: false
+      t.text :description
+
+      t.integer :amount_cents, null: false, default: 0
```
garyhtou:
```suggestion
      t.integer :amount_cents, null: false
```

app/models/payroll/invoice.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    belongs_to :payroll_position, class_name: "Payroll::Position", inverse_of: :invoices
+    belongs_to :reviewed_by, class_name: "User", optional: true
+    belongs_to :payment, optional: true
+
+    has_one :receipt, as: :receiptable
+
+    monetize :amount_cents, with_model_currency: :currency
+
+    validate :currency_matches_position
```
garyhtou, line 49: Good check.

app/models/payroll/invoice.rb
```diff
@@ -0,0 +1,70 @@
...
+#
+module Payroll
+  class Invoice < ApplicationRecord
+    include AASM
+
+    belongs_to :payroll_position, class_name: "Payroll::Position", inverse_of: :invoices
+    belongs_to :reviewed_by, class_name: "User", optional: true
+    belongs_to :payment, optional: true
+
+    has_one :receipt, as: :receiptable
```
garyhtou: Let's use the `Receiptable` concern.

app/models/payroll/invoice.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      event :mark_rejected do
+        transitions from: :submitted, to: :rejected
+      end
+    end
+
+    private
+
+    def currency_matches_position
+      return if payroll_position.blank? || currency == payroll_position.currency
```
garyhtou:
```suggestion
      return if currency == payroll_position.currency
```

app/models/payroll/invoice.rb
garyhtou, line 49: How should we validate that currency is valid? We use the Money gem to work with currencies. Does it have validation to check that its of a valid currency?

app/jobs/payroll/position/expire_job.rb
garyhtou, line 1: After merge: Add a nightly job that checks for positions to expire in case jobs are accidentally dropped.

app/models/payroll/invoice.rb
```diff
@@ -0,0 +1,81 @@
...
+    validate :currency_matches_position
+
+    aasm timestamps: true do
+      state :submitted, initial: true
+      state :approved
+      state :rejected
+
+      event :mark_approved do
+        transitions from: :submitted, to: :approved
+      end
```
garyhtou, line 61: Consider passing in a reviewed_by param.

app/models/payroll/position.rb
```diff
@@ -0,0 +1,101 @@
...
+module Payroll
+  class Position < ApplicationRecord
+    include AASM
+    include Contractable
+
+    has_paper_trail
+
+    belongs_to :payee
+
+    has_many :invoices, class_name: "Payroll::Invoice", foreign_key: "payroll_position_id", inverse_of: :payroll_position, dependent: :destroy
```
garyhtou, line 40: After merge: make sure you cannot delete a position if there is any financial data on it.

app/models/payroll/position.rb
```diff
@@ -0,0 +1,101 @@
...
+    has_one :event, through: :payee
+    has_one :contract_event, through: :payee, source: :event
```
garyhtou:
```suggestion
    has_one :contract_event, through: :payee, source: :event # a requirement of Contractable
```

app/models/payroll/invoice.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    belongs_to :payroll_position, class_name: "Payroll::Position", inverse_of: :invoices
+    belongs_to :reviewed_by, class_name: "User", optional: true
+    belongs_to :payment, optional: true
+
+    has_one :event, through: :payroll_position
+
+    monetize :amount_cents, with_model_currency: :currency
+
+    validate :currency_matches_position
```
garyhtou, line 49:
```suggestion
    validates :currency, inclusion: { in: Money::Currency.all.map(&:iso_code) }
    validate :currency_matches_position
```

app/models/payroll/position.rb
```diff
@@ -0,0 +1,101 @@
...
+      event :mark_onboarded do
+        transitions from: :onboarding, to: :onboarded
+      end
+
+      event :mark_expired do
+        transitions from: :onboarded, to: :expired
+      end
+
+      event :mark_terminated do
+        transitions from: [:under_review, :onboarding, :onboarded], to: :terminated
```
garyhtou:
```suggestion
        transitions from: :onboarded, to: :terminated
```

app/models/payroll/position.rb
```diff
@@ -0,0 +1,101 @@
...
+      state :expired
+      state :rejected
+      state :terminated
+
+      event :mark_onboarding do
+        transitions from: :under_review, to: :onboarding
+      end
+
+      event :mark_rejected do
+        transitions from: :under_review, to: :rejected
```
garyhtou:
```suggestion
        transitions from: [:under_review, :onboarding], to: :rejected
```

garyhtou (approved):

PR #14216, opened 2026-07-08
[Ledger::Item] Cache comment count

Luke-Oldenburg:
## Summary of the problem

We don't want to load comments with the ledger.

## Describe your changes

Cache comment count in `comment_count` and `not_admin_only_comment_count`

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -30,7 +30,7 @@ class Comment < ApplicationRecord
 
   set_public_id_prefix :cmt
 
-  belongs_to :commentable, polymorphic: true
+  belongs_to :commentable, polymorphic: true, touch: true
   belongs_to :user
 
   has_one_attached :file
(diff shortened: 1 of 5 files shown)


YodaLightsabr (approved):

app/models/comment.rb
```diff
@@ -30,7 +30,7 @@ class Comment < ApplicationRecord
 
   set_public_id_prefix :cmt
 
-  belongs_to :commentable, polymorphic: true
+  belongs_to :commentable, polymorphic: true, touch: true
```
garyhtou, line 33: Via paper trail, is it possible to determine the last non-touch change to the updated_at timestamp?

garyhtou (approved):

app/models/comment.rb
Luke-Oldenburg: @claude

PR #14219, opened 2026-07-09
[Ledger::Item] Very temporary fix for not associating LOs

Luke-Oldenburg:
## Summary of the problem

This makes the ledger easier to use in the meantime while we migrate off of HCB code reliance.

## Describe your changes

Assign LO to hcb_code.LO from refresh!


YodaLightsabr (approved):

Luke-Oldenburg:
The part getting removed is a permanent removal. The addition is temporary.

PR #14224, opened 2026-07-09
[Tax] Listen to taxbandits webhooks

polypixeldev:
## Summary of the problem

TaxBandits has webhooks! Let's use them so that we know when to update our data.

## Describe your changes

Listen to the taxbandits webhooks. Only event type we need is WhCertificate Status Change


Luke-Oldenburg:
@claude review this PR

Luke-Oldenburg (approved):
lgtm

PR #14231, opened 2026-07-09
Map CTs with CanonicalTransaction#calculated_ledger_item

YodaLightsabr:
## Summary of the problem

Mapping CTs to Ledger::Items currently relies on HcbCode. This functionally works, however it violates our goal of removing our dependency on HcbCode.

## Describe your changes

This PR maps CTs to Ledger::Items via the new method CanonicalTransaction#calculate_ledger_item, which finds a Ledger::Item from the new linked_object_v2 or the short code.

### DIFF
--- a/app/models/canonical_transaction.rb
+++ b/app/models/canonical_transaction.rb
@@ -477,9 +477,7 @@ class CanonicalTransaction < ApplicationRecord
     safely do
       ActiveRecord::Base.transaction do
         if calculated_ledger_item != local_hcb_code.ledger_item
-          puts calculated_ledger_item
-          puts local_hcb_code
-          Rails.error.unexpected("CanonicalTransaction #{id} has calculated a different ledger item from its local_hcb_code. (#{calculated_ledger_item&.id} vs. #{local_hcb_code.ledger_item&.id}) - #{linked_object_v2&.ledger_item&.id} #{short_code}")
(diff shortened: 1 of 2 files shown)


app/models/canonical_transaction.rb
```diff
@@ -472,6 +473,24 @@ def write_hcb_code
 
   private
 
+  def find_or_create_ledger_item
```
garyhtou: Does this override an existing Active Record method?

app/models/canonical_transaction.rb
```diff
@@ -197,6 +190,14 @@ def linked_object
     @linked_object ||= TransactionEngine::SyntaxSugarService::LinkedObject.new(canonical_transaction: self).run
   end
 
+  def linked_object_v2
+    if column_id = column_transaction_id
+      AchTransfer.find_by(column_id:) || Wire.find_by(column_id:) || CheckDeposit.find_by(column_id:) || IncreaseCheck.find_by(column_id:)
```
garyhtou, line 195: Ideally, this looks something like:

```ruby
Transfer.find_by(column_id:).linked_object
```

_**in the future**_

garyhtou (approved):

Luke-Oldenburg (approved):
lgtm

PR #14234, opened 2026-07-10
Have mappings touch CT and CPT to force a LI remap

YodaLightsabr:
## Summary of the problem

Ledger items are not being mapped to the right organization resulting in production anomalies.

## Describe your changes

This PR has mappings touch CT and CPT to force a remap of Ledger::Item when a transaction gets mapped.


Luke-Oldenburg (commented):
fix checks please

Luke-Oldenburg (approved):
lgtm

PR #14236, opened 2026-07-10
Refactor information_required_for method fields

YodaLightsabr:
## Summary of the problem

This information is only being collected for reimbursements, but we need it for every transfer.

## Describe your changes

Remove the condition that only requires the fields for reimbursements.


YodaLightsabr (commented):
LGTM

PR #14242, opened 2026-07-10
[Tax] Import info from TaxBandits

polypixeldev:
## Summary of the problem

We currently integrate with the TaxBandits API, but don't store any of the data that we need from TaxBandits in the database.

## Describe your changes

Store TIN hash and address in database, and handle conflicts between LE tin hash and tax form TIN hash.


app/controllers/legal_entities_controller.rb
```diff
@@ -1,16 +1,92 @@
...
 
 class LegalEntitiesController < ApplicationController
-  before_action :set_legal_entity
+  before_action :set_legal_entity, only: [:show, :switch]
 
   def show
     authorize @legal_entity
   end
 
+  def switch
```
garyhtou: Let's rename this to `replace`.

app/models/tax/identification_number/hasher.rb
```diff
@@ -0,0 +1,19 @@
...
+module Tax
+  class IdentificationNumber
+    class Hasher
+      # Should be in the format XXX-XX-XXXX for SSNs and XX-XXXXXXX for EINs
+      # FTINs should be in whatever format is standard for the country
+      def self.hash_tin(tin)
+        if Rails.env.production?
+          # AWS KMS
+        else
+          Digest::SHA256.hexdigest(tin.to_s)
```
garyhtou:
```suggestion
          "DEV_TIN_HASH_#{Digest::SHA256.hexdigest(tin.to_s)}"
```

app/models/tax/form.rb
```diff
@@ -110,12 +124,16 @@ def send!
...
 
-    def taxbandits_submission
-      TaxbanditsService.get_submission(payee_id: legal_entity.public_id, submission_id: external_id)
+    def get_taxbandits_submission
+      TaxbanditsService.get_submission(public_id)
+    end
+
+    def get_taxbandits_list_entry
+      TaxbanditsService.get_list_entry(public_id)
     end
```
garyhtou: For Stripe models, we have a convention where we prefix these methods with `remote_*` instead of `get_*`

app/models/tax/form.rb
```diff
@@ -125,14 +143,86 @@ def sync_with_taxbandits
...
+      when "FormW8BENE"
+        :business
+      when "FormW8ECI"
+        submission["FormW8ECI"]["FormData"]["EntityType"] == "INDIVIDUAL" ? :person : :business
+      when "FormW8IMY"
+        :business
+      when "FormW8EXP"
+        :business
+      else
+        raise ArgumentError, "unknown tax form type"
```
garyhtou: maybe we should also do Rails.error.unexpected to ensure engineers are alerted

app/models/tax/form.rb
```diff
@@ -125,14 +143,86 @@ def sync_with_taxbandits
...
+      end
+    end
+
+    def masked_tin
+      entry = get_taxbandits_list_entry
+      entry&.[]("TIN")
```
garyhtou:
```suggestion
      # TaxBandits only returns a masked tin
      entry = get_taxbandits_list_entry
      entry&.[]("TIN")
```

garyhtou: Double check this statement

app/models/tax/form.rb
```diff
@@ -125,14 +143,86 @@ def sync_with_taxbandits
...
 
     def send_using_taxbandits!
-      response = TaxbanditsService.create_whcertificate(id: legal_entity.public_id, name: legal_entity.name)
+      response = TaxbanditsService.create_whcertificate(id: public_id, name: legal_entity.name)
 
       update!(external_service: :taxbandits, signing_url: response["Url"], external_id: response["SubmissionId"])
       sync_with_taxbandits
     end
 
+    def import_taxbandits_data
```
garyhtou, line 215: Ensure we don't accidentally log PII (SSN) via Rails or AppSignal

app/models/tax/form.rb
```diff
@@ -125,14 +143,86 @@ def sync_with_taxbandits
...
+                              [form_data["TIN"], nil]
+                            when "FormW8BEN"
+                              [form_data["USTIN"], form_data["ForeignTIN"]]
+                            when "FormW8ECI"
+                              [form_data["TIN"], form_data["ForeignTIN"]]
+                            when "FormW8BENE", "FormW8IMY", "FormW8EXP"
+                              [form_data.dig("Part1", "USTIN"), form_data.dig("Part1", "ForeignTIN")]
+                            end
+      tin = us_tin.presence || foreign_tin.presence
+      tin_hash = Tax::IdentificationNumber::Hasher.hash_tin(tin)
```
garyhtou: Ensure anything that does into Tin hasher doesn't get logged

app/models/tax/identification_number/hasher.rb
```diff
@@ -0,0 +1,19 @@
...
+        end
+      end
```
garyhtou: This check ensures that TIN Hasher never returns the raw TIN as a hash. 

```suggestion
      def self.hash_tin(tin)
        hash = if Rails.env.production?
          # AWS KMS
        else
          Digest::SHA256.hexdigest(tin.to_s)
        end
        
        raise SomeError if hash == tin

        hash
      end
```

garyhtou (commented):
Filter TIN, SSN, EIN, and FTIN from loggers - this is a rails config

app/models/tax/form.rb
```diff
@@ -125,14 +143,86 @@ def sync_with_taxbandits
...
+      when "FormW8IMY"
+        :business
+      when "FormW8EXP"
+        :business
+      else
+        raise ArgumentError, "unknown tax form type"
+      end
+    end
+
+    def masked_tin
```
garyhtou: How does TaxBandits mask FTINs? I'm wondering because they're not in a specific format.

As a precaution, does it make sense to add a check to ensure the returned value contains at least three X's? This protects us in case TaxBandits ever decides to return the full TIN instead of a masked TIN here.

app/services/taxbandits_service.rb
```diff
@@ -16,16 +27,20 @@ def self.create_whcertificate(id:, name:)
     response.body
   end
 
-  def self.get_submission(payee_id:, submission_id:)
-    submissions = taxbandits_client.get("WhCertificate/List?PayeeRef=#{payee_id}").body
+  def self.get_submission(form_id)
+    taxbandits_client.get("WhCertificate/Get?PayeeRef=#{form_id}").body
```
garyhtou: Nit: you do not need to handle this right now. However, form ID probably should be URL-escaped. However, it's not a problem right now because our form IDs are always public IDs, which don't contain weird characters

app/models/legal_entity_user.rb
```diff
@@ -31,8 +31,8 @@ def person_entities_have_one_user
   end
 
   def user_only_has_one_person_entity
-    if legal_entity.person? && LegalEntityUser.where(user:).excluding(self).any?
-      errors.add(:base, "Users can only have one person legal entity")
+    if legal_entity.person? && legal_entity.archived_at.nil? && LegalEntityUser.joins(:legal_entity).where(user:, legal_entity: { archived_at: nil, entity_type: "person" }).excluding(self).any?
+      errors.add(:base, "Users can only have one non-archived person legal entity")
```
garyhtou:
```suggestion
      errors.add(:base, "Users can only have one non-archived personal legal entity")
```

app/models/legal_entity.rb
```diff
@@ -75,6 +77,26 @@ def display_name
     person? ? "Personal" : (name.presence || "Business")
   end
 
+  def mismatched_tax_form
+    tax_forms.not_discarded.reject { |form| form.tin_hash == tin_hash }.last
```
garyhtou: This makes the assumption that a legal entity can only have at most one mismatched tax form. That assumption, I believe, is safe based on the code we have right now. However, due to jankiness, I could see a world where an LE has multiple mismatched tax forms.

You don't need to make a code change for this right now, but that's something to be aware of.

app/controllers/legal_entities_controller.rb
```diff
@@ -1,16 +1,92 @@
...
+    authorize @legal_entity
+
+    new_tax_form = Tax::Form.find(params[:new_tax_form_id])
```
garyhtou:
```suggestion
    new_tax_form = @legal_entity.tax_forms.find(params[:new_tax_form_id])
```

1. This is more secure
2. It changes a 403 to a 404 error; which i think is better.

Double check that this change works. I assume that `new_tax_form_id` belongs to `@legal_entity`

garyhtou: Should we use public ID here?

app/controllers/legal_entities_controller.rb
```diff
@@ -1,16 +1,92 @@
...
+    end
+
+    redirect_to legal_entity_path(new_le)
+  end
+
+  def create_from_tax_form
+    tax_form = Tax::Form.find(params[:new_tax_form_id])
+    authorize tax_form, :create_legal_entity?
+
+    old_le = LegalEntity.find_by(id: params[:old_le_id])
```
garyhtou: Should we use public IDs here?

app/controllers/legal_entities_controller.rb
```diff
@@ -1,16 +1,92 @@
...
+    authorize old_le, :switch? if old_le.present?
+
+    new_le = nil
+
+    ActiveRecord::Base.transaction do
+      new_le = LegalEntity.create!(
+        name: params[:name],
+        tin_hash: tax_form.tin_hash,
+        entity_type: tax_form.inferred_entity_type,
+        users: [current_user]
```
garyhtou, line 64: What if admins use this route? Would `current_user` be inaccurate?

app/controllers/legal_entities_controller.rb
```diff
@@ -1,16 +1,92 @@
...
+  end
+
   private
 
   def set_legal_entity
     @legal_entity = LegalEntity.find_by_hashid!(params[:id])
   end
 
+  def migrate_pending_payments(from_le:, to_le:)
+    from_le.payees.each do |payee|
```
garyhtou: nit: you might want to `includes` here

app/controllers/legal_entities_controller.rb
```diff
@@ -1,16 +1,92 @@
...
+      next if payee.archived?
+      next if payee.payments.pending_legal_entity.none?
```
garyhtou: I think this might be a bug. An organizer can archive a payee even if that payee has a payment in progress (e.g., in the `pending_legal_entity` state). This means we may need to migrate some archived payees if they have a payment that is still being processed.

app/models/tax/form.rb
garyhtou, line 1: For both `Tax::Form` and `LegalEntity`, add a validation to ensure that `tin_hash` never changes after it's been set.

app/models/tax/form.rb
```diff
@@ -125,14 +143,86 @@ def sync_with_taxbandits
...
+      tin_hash = Tax::IdentificationNumber::Hasher.hash_tin(tin)
+
+      update!(
+        form_type: submission_form_type[4..],
+        tin_hash:,
+        address_line1: address["Address1"],
+        address_line2: address["Address2"],
+        address_city: address["City"],
+        address_state: address["State"] || address["ProvinceOrStateNm"],
+        address_postal_code: address["PostalCd"] || address["ZipCd"],
```
garyhtou, line 268: Are these fields from TaxBandits mutually exclusive? I'd assume so, but it's worth double checking and leaving a comment

app/models/legal_entity.rb
```diff
@@ -75,6 +77,26 @@ def display_name
...
+  end
+
+  def archived?
+    archived_at.present?
+  end
+
+  def latest_usable_tax_form
+    tax_forms.completed.order(completed_at: :desc, created_at: :desc).select do |form|
+      form.tin_hash == tin_hash
+    end.last
```
garyhtou:
```suggestion
    tax_forms.completed.order(completed_at: :desc, created_at: :desc).where(tin_hash:).last
```

app/models/tax/identification_number/hasher.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Tax
+  class IdentificationNumber
+    class Hasher
+      # Should be in the format XXX-XX-XXXX for SSNs and XX-XXXXXXX for EINs
+      # FTINs should be in whatever format is standard for the country
+      def self.hash_tin(tin)
```
garyhtou: We need to normalize the input to ensure reliable hash comparisons.

```suggestion
      def self.hash_tin(tin)
        tin = tin.strip.presence
```

app/models/tax/form.rb
```diff
@@ -21,6 +21,7 @@
 #  signing_url                    :string
 #  taxbandits_status              :string
 #  taxbandits_tin_matching_status :string
+#  tin_hash                       :string
```
garyhtou, line 25: I think it's not worth changing this, but I believe the technically accurate name would be `tin_digest` rather than `tin_hash`

app/services/taxbandits_service.rb
```diff
@@ -16,16 +27,20 @@ def self.create_whcertificate(id:, name:)
     response.body
   end
 
-  def self.get_submission(payee_id:, submission_id:)
-    submissions = taxbandits_client.get("WhCertificate/List?PayeeRef=#{payee_id}").body
+  def self.get_submission(form_id)
```
garyhtou: Because this API call contains sensitive information, we should log every time it's called.

app/models/tax/form.rb
polypixeldev: would this not be captured by appsignal?

app/controllers/legal_entities_controller.rb
polypixeldev: Good point, I don't think admins should be using this route - I've updated the policy to prevent this.

app/models/tax/form.rb
polypixeldev: Yes, they come from different address object shapes depending on the type of form

app/models/tax/form.rb
polypixeldev: Good catch, FTINs are not masked, I've updated the code to mask it ourselves

app/models/tax/form.rb
garyhtou: It depends on whoever is rescuing to report it. It does get auto-reported if no one rescues it, though.

app/models/tax/identification_number/hasher.rb
```diff
@@ -3,15 +3,66 @@
...
-      def self.hash_tin(tin)
-        if Rails.env.production?
-          # AWS KMS
+      MAC_ALGORITHM = "HMAC_SHA_256"
+
+      class HashingError < StandardError; end
+
+      def self.hash_tin(tin, tin_type: nil, country: nil)
+        normalized = normalize(tin)
+        return nil if normalized.blank?
```
garyhtou: Would it be safer to raise an error rather than return `nil`?

This prevents us from accidently saving/using `nil` when we expected a proper hash

app/models/tax/identification_number/hasher.rb
```diff
@@ -3,15 +3,66 @@
...
+      private_class_method :message_for
+
+      def self.normalize(tin)
+        tin.strip.to_s.gsub(/[^0-9A-Za-z]/, "").upcase.presence
+      end
+      private_class_method :normalize
+
+      def self.mac(message)
+        if kms_key_id.present?
+          kms_client.generate_mac(key_id: kms_key_id, message:, mac_algorithm: MAC_ALGORITHM).mac
```
garyhtou: Should we hash the message ourselves before passing it to KMS?

app/models/tax/identification_number/hasher.rb
```diff
@@ -3,15 +3,66 @@
...
+
+      class HashingError < StandardError; end
+
+      def self.hash_tin(tin, tin_type: nil, country: nil)
+        normalized = normalize(tin)
+        return nil if normalized.blank?
+
+        digest = Base64.strict_encode64(mac(message_for(normalized, tin_type:, country:)))
+
+        hash = kms_key_id.present? ? digest : "DEV_#{digest}"
```
garyhtou: I'd probably suggest moving this into the `mac` method (line 47)

app/services/taxbandits_service.rb
```diff
@@ -28,10 +28,12 @@ def self.create_whcertificate(id:, name:)
   end
 
   def self.get_submission(form_id)
+    Rails.logger.info("TaxBandits: get_submission for PayeeRef=#{form_id}")
```
garyhtou:
```suggestion
    Rails.logger.info("TaxBandits: get_submission for PayeeRef=#{form_id} by current_user_id=#{Current.user&.id || "nil"}")
```

app/services/taxbandits_service.rb
```diff
@@ -28,10 +28,12 @@ def self.create_whcertificate(id:, name:)
...
     taxbandits_client.get("WhCertificate/Get?PayeeRef=#{form_id}").body
   end
 
   def self.get_list_entry(form_id)
+    Rails.logger.info("TaxBandits: get_list_entry for PayeeRef=#{form_id}")
```
garyhtou:
```suggestion
    Rails.logger.info("TaxBandits: get_list_entry for PayeeRef=#{form_id} by current_user_id=#{Current.user&.id || "nil"}")
```

app/models/tax/identification_number/hasher.rb
```diff
@@ -3,15 +3,66 @@
...
+
+        raise HashingError, "TIN matches hash" if hash == tin
+
+        hash
+      rescue
+        # cause: nil severs the original exception, whose message/backtrace could
+        # carry the TIN, so nothing sensitive can be logged as this error bubbles up.
+        raise HashingError, "failed to fingerprint TIN", cause: nil
```
garyhtou: good catch!

garyhtou (approved):
🪄

app/models/tax/identification_number/hasher.rb
polypixeldev: I didn't include this in the mac method because the mac method returns some raw bytes - including it here lets us prepend it after encoding it as base64 so that it's easily visible in the DB

app/models/tax/identification_number/hasher.rb
polypixeldev: It is possible for someone to submit a W8-BEN and not provide a US TIN nor FTIN. In this case, we can't aggregrate. I've added code to make sure we don't group nil TINs together

app/models/tax/identification_number/hasher.rb
garyhtou: Ah makes sense

app/models/tax/identification_number/hasher.rb
garyhtou: Okay good!

garyhtou:
## Review + fixes for the TIN handling

I ran a multi-expert review over this PR (security, Rails, tax-compliance, product, and test-coverage passes, each finding adversarially verified) and then pushed fixes for the ship-blockers it surfaced. All changes are in **d37b07ad95b633104af78f892f23f44af6590c12**, and the fixes were re-reviewed until the panel approved.

### Issues found

**Security / PII**
- **Full unmasked TIN fetched during page renders.** `inferred_entity_type` called `WhCertificate/Get` (which returns the plaintext SSN/EIN) just to decide person-vs-business, on views and controllers. Because `LegalEntityPolicy#show?` permits auditors, any auditor could make HCB pull an arbitrary payee's full TIN.
- **Masked (last-4) TIN shown to admins/auditors.** The show page rendered `masked_tin` with no viewer check. Policy is that nobody but the taxpayer sees any part of a TIN.
- **Deployed staging hashed real TINs with a committed key.** The dev HMAC fallback was gated on `!Rails.env.production?`, so staging (which talks to the real TaxBandits API) fingerprinted real SSNs with the literal key `"development-key"` from the repo.

**Hashing correctness**
- `hash_tin(nil)` raised instead of returning nil (`.strip` ran before `.to_s`), permanently 500-ing completion for a legitimate W-8 with no TIN.
- The `tin_type`/`country` namespacing was never passed by the caller, so an SSN and an EIN with the same nine digits produced the same hash and merged two taxpayers.

**Mismatch / payability**
- `mismatched_tax_form` treated any NULL `tin_hash` (every pre-existing row, every pending form) as a TIN mismatch, making payees unpayable.
- `payable?` keyed off `latest_tax_form`, which a freshly started pending form outranks (Postgres sorts NULL `completed_at` first on DESC), so clicking "Start new tax form" instantly made a payable recipient look unpayable.
- `authorize old_le, :switch?` called a policy method that didn't exist, so every "Create new legal entity" submission 500'd.
- The TaxBandits `PayeeRef` changed from the legal entity's id to the tax form's, orphaning every certificate registered before the deploy.
- `replace` dropped the payee's bank account and left stranded payees on the archived entity.

The PR also had no test coverage across ~500 lines touching crypto, money movement, and PII.

### Fixes made

- **TIN never fetched on renders.** Entity type and TIN type are derived once at import and persisted (`tax_forms.entity_type` / `tin_type`); `WhCertificate/Get` is now reachable only from the private import path.
- **Masked TIN restricted to the taxpayer** via `LegalEntityPolicy#show_masked_tin?`. A single `masked_tin_tag` helper is the only render path, and it skips the TaxBandits call entirely for non-owners (also `fs-exclude`'d out of FullStory).
- **Dev key gated on `Rails.env.local?`**, so every deployed environment raises rather than falling back to the in-repo key.
- **Hashing:** normalize before coercing so `hash_tin(nil)` returns nil; `tin_type`/`country` are required and namespaced by the kind of TIN (individual / entity / foreign + country of residence) so distinct taxpayers can't collide. Foreign TINs key on residence country, not the mutable mailing address, and skip fingerprinting when no country is determinable rather than bucketing together.
- **Mismatch/payability:** only a completed, fingerprinted form counts as a mismatch; `payable?` keys off the latest *completed* form; `switch?` added as a strict ownership check (also closes an IDOR on the user-supplied `old_le_id`); `payee_ref` recorded so certificates stay reachable across the id change and entity moves; `replace` carries payout methods across and archives stranded payees.
- **Also:** indexed `tin_hash`, anchored the TIN log-filter regexes, and kept manual (organizer-entered) forms clear of every TaxBandits call so the upcoming contractors flow fingerprints identically to imported forms.
- Added model, policy, and controller specs covering the above.

_Reviewed and fixed by Claude (Claude Code), at @garyhtou's direction._

PR #14251, opened 2026-07-13
[Applications] Allow rejecting approved applications

polypixeldev:
## Summary of the problem

Sometimes we'll approve an application, then call them and have to reject.

## Describe your changes

Allow rejecting approved applications that have not been activated.


Luke-Oldenburg (approved):
lgtm

PR #14259, opened 2026-07-13
Call map after setting LI from CT/CPT

YodaLightsabr:
## Summary of the problem

Ledger items are not properly mapping

## Describe your changes

Call `map!` after setting LI


Luke-Oldenburg (approved):
lgtm

PR #14271, opened 2026-07-14
[Payout Method] Uniquely name your methods

Mohamad-Mortada:
## Summary of the problem
If I got 2 ACH transfers, it's hard to distinguish which one it is purely based on the name in the reimbursement view. One could be a Chase while other be a Bank of America account, however we show them as "ACH Transfer" 

## Describe your changes
This allows you to name your methods to your preference (e.g "My Personal Chase Account")


garyhtou:
@claude, please review this with multiple experts. identify bugs, security issues, and any poor ux

db/migrate/20260730210250_add_name_to_legal_entity_payout_methods.rb
garyhtou, line 1: This migration needs a new timestamp because it currently shares the same timestamp as https://github.com/hackclub/hcb/blob/addf132317747baccbf29806026870e2a231ba5b/db/migrate/20260714120000_add_taxbandits_import_fields_to_tax_forms.rb

garyhtou (approved):
Good to merge after the timestamp has been updated. I'd suggest updating it to whatever our current timestamp is. That means schema.rb needs to be updated by running `bin/rails db:migrate`

PR #14281, opened 2026-07-14
[Contractor] Add docuseal contracts into the invite flow

Mohamad-Mortada:
## Summary of the problem
We need to have the organizer sign the form before they are able to invite them

## Describe your changes
Integrates docuseal contracts into the flow and generates contract when loading second page of form (voids and regenerates contract if user goes back into details and changes them)


app/mailers/payroll/position_mailer.rb
```diff
@@ -0,0 +1,22 @@
...
+      @event = @position.event
+      legal_entity = @payee.legal_entity
+
+      recipients = if legal_entity&.users&.any?
+                     legal_entity.users.map(&:email_address_with_name)
+                   else
+                     [@payee.email]
+                   end
+
+      mail to: recipients,
```
garyhtou, line 18: `reply_to` org and/or `Payroll::Position` creator

garyhtou:
> Written by Claude

**Review note — contract voiding is no longer admin-only for payroll positions**

While reviewing this branch, I noticed `Payroll::PositionsController#update` → the new private `void_pending_contracts!` calls `contract.mark_voided!(reissuing: true)` directly on the model (`app/controllers/payroll/positions_controller.rb`), authorized only via `Payroll::PositionPolicy#update?` (aliased to `create?`, i.e. any org member with transfer-creation rights — not admin-only).

Everywhere else in the codebase, voiding a contract goes through `ContractsController#void`/`#reissue`, gated by `ContractPolicy#void?`/`#reissue?` (`user&.admin?`) — admin-only, system-wide.

This new path is narrower in scope than that (limited to the caller's own event, and only to `pending`/`sent` contracts — never a fully `signed` one), so it's not a full privilege escalation, but it does diverge from the existing admin-only invariant for voiding contracts elsewhere in the app.

I'm not sure whether this scope-broadening is intentional (e.g. "org members should be able to self-service edit/reissue their own in-flight contractor contracts, since HCB hasn't legally executed it yet") or an oversight from bypassing `ContractPolicy` while wiring up the edit flow. Flagging for a decision either way — if intentional, it might be worth an explicit policy method (e.g. distinguishing "void your own in-flight payroll contract" from the admin-wide voiding power) so the divergence is documented rather than incidental.

*(left by Claude Code)*

garyhtou:
**This comment was written entirely by Claude Code, at Gary's direction.**

Following up on the review above with fixes for everything except the two items called out separately (the admin-only-voiding question in my other comment, and the onboarding email fanning out to all users on a legal entity, which I'm told is intentional). All fixes are covered by new specs (61 examples added/updated across model, policy, and controller specs), verified by three independent review passes (security, correctness, error-handling), and pushed as two commits: `b2ea65bd6` and `6c57a7b55`.

**`b2ea65bd6` — model-layer robustness**
- `Payroll::Position` had no `display_name` method, yet the sign-contract and edit pages called `@position.display_name` directly — both pages 500'd on every real visit. Delegated it to `payee`.
- `send_contract` created the `Contract` row in one transaction, then called out to DocuSeal outside it with no rescue — a network hiccup left a permanently broken, un-retryable contract behind. Now voids the broken contract on failure so the next attempt retries cleanly.
- `Contract#archive_on_docuseal!` could itself crash (e.g. voiding a contract that was never actually sent, or DocuSeal being unreachable) and take down the caller with it. Now fails closed: skips the call when there's nothing to archive, and reports-but-doesn't-raise on failure.
- The new contractor-notification side effects (mailer + reminder jobs) ran inside the DocuSeal webhook's transaction — a job-enqueue failure could roll back a signature DocuSeal already recorded. Isolated so that can't happen.

**`6c57a7b55` — signature hijacking, terms drift, unsafe GET**
- Any org member with transfer-creation rights could sign the contract's "organizer" slot as whoever originally created the invite (the DocuSeal embed didn't check the viewer matched the party). Now only renders for that specific user; others see a "waiting on X to sign" state.
- A position's rate/dates/title could still be edited after its contract was fully signed by every party, silently diverging from the executed document. `PositionPolicy#edit?`/`#update?` now lock once signed.
- Sending a contract to DocuSeal was a side effect of a plain GET — not CSRF-safe. Sending now happens from the `create`/`update` POST/PATCH actions; `GET #contract` is display-only. Editing also now only voids/reissues the in-flight contract when something contract-relevant actually changed (no more forced re-signing on a no-op save), and the reissue chain is correctly preserved even across a failed-then-retried send.
- Fixed the edit page highlighting the wrong breadcrumb step.

*(posted by Claude Code, at the requester's direction)*

app/models/contract.rb
manuthecoder: schedule_reminders isn't gated by contract_notify_when_sent like notify is so payroll contractors get sign-reminder emails at send time (before HCB signs and before they can even sign) plus a duplicate set when signed

app/controllers/payroll/positions_controller.rb
```diff
@@ -38,16 +41,108 @@ def create
...
+    def contract_terms_changed?(attachment_changed:)
+      attachment_changed || (@position.saved_changes.keys & CONTRACT_RELEVANT_ATTRIBUTES).any?
+    end
+
+    # At most one non-voided contract exists per position (enforced by
+    # Contract#one_non_void_contract), so there's only ever one to void here.
+    def void_pending_contract!
+      contract = @position.contracts.where(aasm_state: [:pending, :sent]).first
+      contract&.tap { |c| c.mark_voided!(reissuing: true) }
+    end
```
manuthecoder, line 135: This allows non-admin org members to void contracts (gated by Payroll::PositionPolicy#update?), whereas everywhere else voiding is admin-only (ContractPolicy#void?). The scope is narrow (own event, in-flight only, never signed) but diverges from the existing pattern. Is this intentional?

manuthecoder (changes requested):

app/controllers/payroll/positions_controller.rb
polypixeldev: I'm fine with this

app/mailers/payroll/position_mailer.rb
```diff
@@ -0,0 +1,32 @@
...
+           subject: "[Action Required] Complete your onboarding to get paid as a contractor for #{@event.name}",
```
polypixeldev, line 19: I would use a different subject here - this could be the first time they're hearing of this from HCB, so maybe something more friendly?

app/models/payroll/position.rb
```diff
@@ -39,6 +39,8 @@ class Position < ApplicationRecord
 
     belongs_to :payee
 
+    delegate :display_name, to: :payee
```
polypixeldev: This is confusing - payee display name is not the position's name

app/models/contract.rb
```diff
@@ -289,7 +289,13 @@ def send_using_docuseal!
   end
 
   def archive_on_docuseal!
+    return if external_id.blank?
+
     docuseal_client.delete("/submissions/#{external_id}")
+  rescue Faraday::Error => e
+    # Voiding is authoritative in our own DB; a failure to archive the remote
+    # submission shouldn't block the void or crash the caller.
+    Rails.error.report(e, context: { contract_id: id })
```
polypixeldev: I disagree with this comment - if it failed to void on docuseal, the contract could still be completed outside of HCB, which would not be good if we meant to void it (and then receive a webhook that it was actually signed). Failing to void on docuseal should block void on our end

app/views/payroll/position_mailer/onboarding.html.erb
```diff
@@ -0,0 +1,22 @@
+<p>Hello,</p>
+
+<p>
+  <%= @event.name %> is looking to onboard you as a contractor through HCB.
```
polypixeldev:
```suggestion
  <%= @event.name %> is looking to onboard you as a contractor through HCB!
```
also include details about the position

app/views/payroll/position_mailer/onboarding.html.erb
```diff
@@ -0,0 +1,22 @@
...
+</p>
+
+<p>
+  In order to get paid, we need your tax information and a payout method if not already configured.
```
polypixeldev: We can check if they have tax info or payout method already configured

app/controllers/payroll/positions_controller.rb
```diff
@@ -38,16 +41,108 @@ def create
...
       else
         flash[:error] = @position.errors.full_messages.to_sentence
         render :new, layout: "transfer", status: :unprocessable_entity
       end
     end
 
+    # Display-only: never creates or sends a contract as a side effect of a
+    # GET, so this action stays safe (idempotent, no external calls). Sending
+    # happens from #create/#update instead.
+    def contract
```
polypixeldev, line 69: If the contract's been signed, redirect them to the next step

app/mailers/payroll/position_mailer.rb
Mohamad-Mortada: Smth like: ?
```suggestion
           subject: "You've been added as a contractor for #{@event.name} — let's get you set up",
```

app/views/payroll/position_mailer/onboarding.html.erb
Mohamad-Mortada: Don't they first need to choose legal entity?

polypixeldev: You can pay a payee that already has a legal entity configured

app/controllers/payroll/positions_controller.rb
```diff
@@ -16,6 +19,14 @@ def show
       render :show, layout: !@frame
     end
 
+    def welcome
+      @position = Payroll::Position.find_by_hashid!(params[:id])
+      authorize @position, :welcome?
```
polypixeldev:
```suggestion
      authorize @position
```

app/models/legal_entity/payout_method.rb
```diff
@@ -8,6 +8,7 @@
 #  archived        :boolean          default(FALSE), not null
 #  default         :boolean          default(FALSE), not null
 #  details_type    :string           not null
+#  name            :string
```
Mohamad-Mortada:
```suggestion
```

polypixeldev (approved):

manuthecoder (approved):

PR #14282, opened 2026-07-14
Reload `@ledger_item` prior to mapping

YodaLightsabr:
## Summary of the problem

`@ledger_item` is not properly loading its association with CTs and CPTs resulting in the ledger item going unmapped

## Describe your changes

Prior to evaluating the mapping logic, reload `@ledger_item` to ensure associations are up to date.


Luke-Oldenburg (approved):
lgtm

PR #14283, opened 2026-07-14
[Ledger] Refactor filters into a concern

Luke-Oldenburg:
## Summary of the problem

This code will be shared with card grants controller as well.

## Describe your changes

Move filter creation and query generation into a shared concern.

### DIFF
--- a/app/controllers/concerns/set_ledger_filters.rb
+++ b/app/controllers/concerns/set_ledger_filters.rb
@@ -1,10 +1,5 @@
 # frozen_string_literal: true
 
-# Adds `set_ledger_filters` and `ledger_query` methods to a controller.
-# Requires `@event` to be set. Set `@include_card_grant_ledgers` beforehand to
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/set_ledger_filters.rb
```diff
@@ -0,0 +1,117 @@
+# frozen_string_literal: true
+
+# Adds `set_ledger_filters` and `ledger_query` methods to a controller.
+# Requires `@event` to be set. Set `@include_card_grant_ledgers` beforehand to
+# also include the event's card grant ledgers (e.g. for the grant overview) in
+# `@ledgers` and the filterable users. `ledger_query` builds a `Ledger::Query`
+# from the parsed filters; executing it is left to the controller.
```
Luke-Oldenburg:
```suggestion
```

YodaLightsabr (approved):

PR #14285, opened 2026-07-14
[Payout Method] Fix admins configuring payout methods for others

Mohamad-Mortada:
## Summary of the problem
We automatically assume in LegalEntityUpdateService that we're changing the current user's not the legal entity owner. 

## Describe your changes
Uses legal entity directly instead of user

### DIFF
--- a/app/controllers/legal_entity/payout_methods_controller.rb
+++ b/app/controllers/legal_entity/payout_methods_controller.rb
@@ -95,7 +95,7 @@ class LegalEntity
 
     def render_error_payout_settings(payout_method)
       @legal_entity = payout_method.legal_entity || legal_entity
-      @user = @legal_entity&.users&.first || current_user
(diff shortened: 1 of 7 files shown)


app/controllers/legal_entity/payout_methods_controller.rb
```diff
@@ -96,10 +94,10 @@ def require_unlocked_method
     end
 
     def render_error_payout_settings(payout_method)
-      @user = current_user
-      @payout_method = payout_method
-      @legal_entities = current_user.legal_entities
       @legal_entity = payout_method.legal_entity || legal_entity
+      @user = @legal_entity&.users&.first || current_user
```
polypixeldev: Instead of just getting the first user, pass user ID in the request to the controller and use that here.

polypixeldev (approved):

PR #14287, opened 2026-07-14
Create `Reimbursement::PayoutHolding#event`

YodaLightsabr:
## Summary of the problem

`Reimbursement::PayoutHolding` does not have `event` implemented, so `Ledger::Mapper` cannot use `linked_object&.event` as a mapping strategy.

## Describe your changes

This PR implements `Reimbursement::PayoutHolding#event` so it can be referenced in `Ledger::Mapper`.


Luke-Oldenburg (approved):
lgtm

PR #14288, opened 2026-07-15
[Ledger] Add fiscal sponsorship fee

Luke-Oldenburg:
## Summary of the problem

We were missing this.

## Describe your changes

Adds the fiscal sponsorship fee


app/views/ledgers/_pending_fee_transaction.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  </td>
+  <td><%= format_date Time.now %></td>
+  <td class="transaction__memo transaction__memo--fluid">
+    <%= badge_for "Pending", class: "bg-transparent border border-dashed border-muted m0 mr1" %>
+    Fiscal sponsorship
+  </td>
+  <td class="nowrap"></td>
+  <td class="nowrap">
+    <span class="flex items-center justify-end">
+      <%= render_money(-1 * @event.fronted_fee_balance_v2_cents) %>
```
YodaLightsabr: Rest of the PR LGTM, but let's get this into a method on `Ledger`

YodaLightsabr (approved):
This is really gross, so in the future we should swap this out for caching the fee amount on `Ledger::Item`—it'll make our calculation so much easier, but we'll save that for another day.

PR #14292, opened 2026-07-15
[Ledger Mapper] Check `linked_object_v2`

Luke-Oldenburg:
## Summary of the problem

`linked_object_v2` is being ignored resulting in RawStripeTransaction CTs not being mapped properly

## Describe your changes

Use `linked_object_v2`

### DIFF
--- a/app/services/ledger/mapper.rb
+++ b/app/services/ledger/mapper.rb
@@ -50,6 +50,10 @@ class Ledger
           return event
         end
 
+        if (event = ct.try(:linked_object_v2))
+          return event
+        end
+
         # Fallback, see if any linked objects have an event.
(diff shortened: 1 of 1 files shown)


app/services/ledger/mapper.rb
```diff
@@ -50,6 +50,10 @@ def event_from_canonical_transactions
           return event
         end
 
+        if (event = ct.try(:linked_object_v2))
```
YodaLightsabr:
```suggestion
        if (event = ct.linked_object_v2.try(:event))
```

YodaLightsabr (changes requested):

YodaLightsabr (approved):

PR #14310, opened 2026-07-16
[Ledger] Hide fee row when fee is LTE 0

Luke-Oldenburg:
## Summary of the problem

This was rendering on all ledgers.

## Describe your changes

Hide.

### DIFF
--- a/app/views/ledgers/_pending_fee_transaction.html.erb
+++ b/app/views/ledgers/_pending_fee_transaction.html.erb
@@ -1,6 +1,7 @@
 <%# locals: (ledger:) %>
 
 <% if ledger.fronted_fee_balance_cents > 0 && @direction != "revenue" %>
+  <%# @direction != revenue is in case the user has a revenue-only filter applied %>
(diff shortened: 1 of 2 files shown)


app/views/ledgers/_pending_fee_transaction.html.erb
```diff
@@ -1,24 +1,26 @@
...
-    <span class="flex items-center justify-end">
-      <%= render_money(-1 * ledger.fronted_fee_balance_cents) %>
-    </span>
-  </td>
-  <% admin_tool("", "td") do %>
-  <% end %>
-  <td></td>
-  <td></td>
-</tr>
+<% if ledger.fronted_fee_balance_cents > 0 && @direction != "revenue" %>
```
YodaLightsabr, line 3:
```suggestion
<% if ledger.fronted_fee_balance_cents > 0 && @direction != "revenue" %>
  <%# @direction != revenue is in case the user has a revenue-only filter applied %>
```

YodaLightsabr (approved):

PR #14313, opened 2026-07-16
[Ledger] Fix extra spacing on badged ledger items

YodaLightsabr:
Badges on ledger items create extra space on the ledger item that shouldn't be there. Setting line-height to 0 prevents this from happening without impacting the padding or margin of the badge pill.


YodaLightsabr:
@claude We want this to actually apply to every badge type because they all have this issue. If there's a badge here, it should have a line height of 0. Please and thank you.

YodaLightsabr:
Pushed as `f8cc4d7` — the earlier `github-actions[bot]` push failure was just a token-permission issue, the change itself is now on the branch. `line-height: 0` now lives on the shared base `.badge` rule in `app/assets/stylesheets/components/_badges.scss`, so it applies to every badge app-wide, and the temporary `.badge--pending` modifier has been removed (net change vs `main` is the single base-rule line).

---
_Generated by [Claude Code](https://claude.ai/code)_

YodaLightsabr:
@claude Don't make this change globally, only on badges that appear on a ledger item

YodaLightsabr:
Pushed the scoped version as `844fcb1` (the `github-actions[bot]` 403 was just a token-permission issue — this change is on the branch). Net diff vs `main`:

- `app/assets/stylesheets/components/_badges.scss`: base `.badge` `line-height` restored to `1.125`; added a scoped `&.badge--ledger-item { line-height: 0; }`.
- `app/views/ledger/_item.html.erb`: the `badge_for` call now passes `class: "badge--ledger-item #{item.status_css}"`, so every badge rendered on a ledger item (regardless of status) gets it — and nothing outside ledger items is affected.

Applied the class once in the ledger-item partial rather than repeating it across each `status_css` branch, but the effect is the same. Global badges are untouched.

---
_Generated by [Claude Code](https://claude.ai/code)_

Luke-Oldenburg:
Screenshot From 2026-07-29 13-35-21

PR #14314, opened 2026-07-16
[v4][Tag] Checking OP should be off tag event

Mohamad-Mortada:
We need to pass in event for OP.role_at_least

### DIFF
--- a/app/policies/tag_policy.rb
+++ b/app/policies/tag_policy.rb
@@ -6,7 +6,7 @@ class TagPolicy < ApplicationPolicy
   end
 
   def create?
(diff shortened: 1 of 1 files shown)


YodaLightsabr (approved):

app/policies/tag_policy.rb
```diff
@@ -6,7 +6,7 @@ def show?
   end
 
   def create?
-    OrganizerPosition.role_at_least?(user, record, :member)
+    OrganizerPosition.role_at_least?(user, record.event, :member)
```
Mohamad-Mortada:
```suggestion
    member?
```

PR #14316, opened 2026-07-16
Create `RawPendingStripeServiceFeeTransaction`

YodaLightsabr:
## Summary of the problem

We currently have no way of persisting a `StripeServiceFee`'s short code outside of the existing `HcbCode` system. This clashes with the design principles of our new `Ledger::Item` system, wherein ledger items should not be created until a transaction occurs.

## Describe your changes

This PR creates `RawPendingStripeServiceFeeTransaction`, allowing a `StripeServiceFee` to have a `Ledger::Item` before we import the transfers from Column. This creation of the ledger item will generate a short code that can be used when making the Column book transfers.

While we still need to (in the future) switch references to `local_hcb_code.short_code` to instead use `ledger_item.short_code`, this system allows the mapper to map `RawPendingStripeServiceFeeTransaction` CPTs to the right event automatically so that subsequent CTs referencing the same short code can be auto-mapped to the existing ledger item.


app/models/canonical_transaction.rb
```diff
@@ -337,6 +337,11 @@ def bank_fee
     nil
   end
 
+  def stripe_service_fee
+    return linked_object if linked_object.is_a?(StripeServiceFee)
+    nil
```
Luke-Oldenburg, line 343:
```suggestion
    return linked_object if linked_object.is_a?(StripeServiceFee)

    nil
```

Luke-Oldenburg:
@claude review

Luke-Oldenburg (approved):
lgtm

PR #14321, opened 2026-07-16
[Ledger::Item] Fix status for partially fronted transactions

Luke-Oldenburg:
## Summary of the problem

We want to be rendering these as settled.

## Describe your changes

Ignore the presence of CTs


Luke-Oldenburg (commented):
"lgtm" - @YodaLightsabr

PR #14322, opened 2026-07-16
[Contractors] Add admin UI

manuthecoder:
Adds admin UI


app/models/admin/nav.rb
```diff
@@ -302,27 +302,59 @@ def payroll
...
           ),
           make_item(
-            name: "Payments",
-            path: employee_payments_admin_index_path,
-            count: ->{ Employee::Payment.paid.count },
+            name: "Contractor Payments",
+            path: admin_payments_path,
+            count: ->{ Payment.under_review.count },
+            count_type: :tasks
+          ),
```
polypixeldev: I would move payments to be under the Spending tab with the other transfer types

app/models/admin/nav.rb
```diff
@@ -302,27 +302,59 @@ def payroll
...
+            count: ->{ Tax::Form.count },
             count_type: :records
-          )
+          ),
+          *legacy_payroll_items
         ]
       )
     end
 
+    def legacy_payroll_items
```
polypixeldev: Go ahead and get rid of this - we're not using these old pages again

app/views/admin/legal_entities/index.html.erb
```diff
@@ -0,0 +1,64 @@
+<% title "Legal Entities" %>
+
+<%= form_with local: true, class: "card", url: admin_legal_entities_path, method: :get do |form| %>
+  <div class="flex flex-col sm:flex-row items-center gap-2">
+    <%= render "events/filters/search", form: %>
+    <%= form.select :entity_type, [%w[Person person], %w[Business business]], { include_blank: "All types", selected: @entity_type } %>
+    <%= form.select :filter, [%w[Managed managed], %w[Archived archived]], { include_blank: "All entities", selected: params[:filter] } %>
```
polypixeldev: It might make more sense to have these be individual checkboxes instead of a single select - maybe you want to see archived managed LEs

app/views/payroll/positions/show.html.erb
```diff
@@ -50,6 +50,21 @@
     </section>
   </article>
 
+  <% if @position.under_review? %>
+    <% admin_tool("mt-4 flex items-center justify-between") do %>
```
polypixeldev, line 54: Image
fix

polypixeldev (approved):

PR #14323, opened 2026-07-16
Fix plan check on suborg creation form

polypixeldev:
## Summary of the problem

It was checking event plan instead of the suborg plan

## Describe your changes

Change check for cosigner email to look at suborg plan instead of event plan


YodaLightsabr (approved):
LGTM!

PR #14327, opened 2026-07-17
[Ledger::Mapper] Don't skip non-column CTs

YodaLightsabr:
## Summary of the problem

Our logic for Stripe transactions... lives _after_ we skip all Stripe transactions.

image

## Describe your changes

Remove the line that skips all non-Column CTs.


Luke-Oldenburg (approved):
lgtm

Issue #14279, opened 2026-07-14
[Wires] Allow Admins to edit Currency & Amounts


devramsean0:
Right now Admins are unable to edit the amount or currency of a wire sent through HCB. This means ops has to bug engineering to fix it or reject the wire and resubmit with the corrected currency.

# Solution
Add these fields to the existing "Edit Wire" page

PR #14328, opened 2026-07-17
Allow admins to edit wire currency & amount

jeninh:
Adds currency and amount fields to the admin-only "Edit Wire" page so ops can correct wires without rejecting and resubmitting (#14279). Editing amount or currency resyncs the wire's unsettled canonical pending transaction with a fresh USD conversion, mirroring WiseTransfer's after_update behavior.


jeninh:
i am not trying to hand write checks my friend

PR #14331, opened 2026-07-17
Get event and card_grant from primary_ledger

YodaLightsabr:
## Summary of the problem

We're getting `event` and `card_grant` off of `ledger_item` when we should be getting it off of `ledger_item.primary_ledger`.

## Describe your changes

Get `event` and `card_grant` through `ledger_item.primary_ledger`.


Luke-Oldenburg (approved):
lgtm

PR #14333, opened 2026-07-17
Fix payee operation redirects

polypixeldev:
## Summary of the problem

The payee pickers on the payments and contractors pages were different, and when you did something to a payee on the contractors page, it would redirect to the payments page.

## Describe your changes

Use the same partial in both pages and fix the redirects.


garyhtou (commented):
🪄

garyhtou (approved):

PR #14340, opened 2026-07-17
[v4] Add donation show and shallow route index

Mohamad-Mortada:
We don't have a show route and since indexing just got merged I realized the route wasn't shallow routed which is our new convention

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -745,7 +745,7 @@ Rails.application.routes.draw do
           end
 
           resources :disbursements, path: "transfers", only: [:create]
-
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -746,7 +746,7 @@
 
           resources :disbursements, path: "transfers", only: [:create]
```
Mohamad-Mortada:
```suggestion
          # TODO: shallow route these (breaking change)
```

polypixeldev (approved):

PR #14342, opened 2026-07-17
Update recipient address field placeholder

briyandyju09:
Updated the address placeholder examples to use a UK address format.


Luke-Oldenburg (approved):
lgtm

PR #14344, opened 2026-07-17
[Contractors] Fix wise transfers

manuthecoder:
Currently, when a contractor has a payout method of Wise selected and submits an invoice of $100 USD, it requests a wise transfer of 100 AED (for example) instead of the AED equivalent of $100 USD. This PR fixes that bug


briyandyju09:
lgtm

app/services/money_service.rb
polypixeldev, line 16: use convert_to_usd_wise here

polypixeldev (approved):

PR #14293, opened 2026-07-15
Create `RawPendingFeeRevenueTransaction`

YodaLightsabr:
## Summary of the problem

We currently have no way of persisting a `FeeRevenue`'s short code outside of the existing `HcbCode` system. This clashes with the design principles of our new `Ledger::Item` system, wherein ledger items should not be created until a transaction occurs.

## Describe your changes

This PR creates `RawPendingFeeRevenueTransaction`, allowing a `FeeRevenue` to have a `Ledger::Item` before we import the transfers from Column. This creation of the ledger item will generate a short code that can be used when making the Column book transfers.

While we still need to (in the future) switch references to `local_hcb_code.short_code` to instead use `ledger_item.short_code`, this system allows the mapper to map `RawPendingFeeRevenueTransaction` CPTs to the right event automatically so that subsequent CTs referencing the same short code can be auto-mapped to the existing ledger item.

PR #14348, opened 2026-07-17
Create `RawPendingFeeReimbursementTransaction`

YodaLightsabr:
Similar to #14316 and #14293, this PR creates `Ledger::Item`s for `FeeReimbursement`s so that we can persist the short code and auto-map them properly.


YodaLightsabr:
@claude Can you give this PR a thorough review to ensure it's implemented properly?

garyhtou:
@Luke-Oldenburg , can you review this too?

app/models/raw_pending_fee_reimbursement_transaction.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: raw_pending_fee_reimbursement_transactions
+#
+#  id                   :bigint           not null, primary key
+#  amount_cents         :integer
```
garyhtou, line 8: should this be not null?

app/services/fee_reimbursement_service/create_canonical_pending_transaction.rb
```diff
@@ -0,0 +1,56 @@
...
+    # fee reimbursement already has its raw pending transaction, and skips
+    # zero-amount reimbursements (which move no money and get no topup).
+    def run
+      existing = fee_reimbursement.raw_pending_fee_reimbursement_transaction
+      return existing.canonical_pending_transaction if existing.present?
+      return if fee_reimbursement.amount.to_i.zero?
+
+      ActiveRecord::Base.transaction do
+        rpfrt = fee_reimbursement.create_raw_pending_fee_reimbursement_transaction!(
+          date_posted: (fee_reimbursement.processed_at || Time.current).to_date,
```
garyhtou, line 19: In what case would `processed_at` be `nil`? I would assume it shouldn't happen. Maybe we should fail loudly rather than falling back to `Time.current`

app/services/fee_reimbursement_service/create_canonical_pending_transaction.rb
```diff
@@ -0,0 +1,56 @@
...
+          amount_cents: -fee_reimbursement.amount
+        )
+
+        canonical_pending_transaction = CanonicalPendingTransaction.create!(
+          date: rpfrt.date,
+          amount_cents: rpfrt.amount_cents,
+          memo: rpfrt.memo,
+          raw_pending_fee_reimbursement_transaction: rpfrt
+        )
```
garyhtou, line 28:
```suggestion
        )
        # Creating the CPT automatically creates a Ledger::Item and auto-maps it
```

app/services/fee_reimbursement_service/create_canonical_pending_transaction.rb
```diff
@@ -0,0 +1,56 @@
...
+      @fee_reimbursement_id = fee_reimbursement_id
+    end
+
+    # Idempotent: safe to re-run (e.g. from the nightly) — it no-ops once the
+    # fee reimbursement already has its raw pending transaction, and skips
+    # zero-amount reimbursements (which move no money and get no topup).
+    def run
+      existing = fee_reimbursement.raw_pending_fee_reimbursement_transaction
+      return existing.canonical_pending_transaction if existing.present?
+      return if fee_reimbursement.amount.to_i.zero?
```
garyhtou, line 15:
```suggestion
      return if fee_reimbursement.amount_cents.zero?
```
^ this reads nicer; but only possible if we make the `amount_cents` column not null

garyhtou (approved):

app/services/fee_reimbursement_service/create_canonical_pending_transaction.rb
Luke-Oldenburg: Between `PayoutService::Invoice::Create` or `PayoutService::Donation::Create` running and `FeeReimbursementService::Nightly` running.

app/models/raw_pending_fee_reimbursement_transaction.rb
Luke-Oldenburg: I think so but the other raw models are also missing it so we may want to come back and fix them later all at once.

app/services/fee_reimbursement_service/create_canonical_pending_transaction.rb
YodaLightsabr: Is `amount_cents` defined?

PR #14351, opened 2026-07-17
[Ledger] Roll out the new ledger!

YodaLightsabr:
This PR rolls out the new ledger! New vs. old preferences are stored with a Flipper flag on the User.

image


YodaLightsabr:
@claude Could I get a sanity check on this PR

app/views/events/_heading.html.erb
```diff
@@ -12,6 +12,20 @@
...
+    <% if organizer_signed_in? %>
+      <% if new_ledger_active %>
+        <%= link_to event_transactions_path(event, apply_flipper: true), class: "btn flex items-center" do %>
+          <%= inline_icon "back" %>
+          Go back to the old ledger
+        <% end %>
+      <% else %>
+        <%= link_to event_ledger_path(event, apply_flipper: true), class: "btn flex items-center" do %>
+          <%= inline_icon "wand-stars" %>
+          Preview the new ledger
```
garyhtou:
```suggestion
          Try the new ledger
```

app/views/events/_heading.html.erb
```diff
@@ -12,6 +12,20 @@
   <% end %>
 
   <div class="flex items-center gap-2">
+    <% if organizer_signed_in? %>
+      <% if new_ledger_active %>
+        <%= link_to event_transactions_path(event, apply_flipper: true), class: "btn flex items-center" do %>
+          <%= inline_icon "back" %>
+          Go back to the old ledger
```
garyhtou:
```suggestion
          Go back to the classic ledger
```

app/helpers/events_helper.rb
```diff
@@ -71,21 +71,12 @@ module EventsHelper
     },
     {
       name: "Transactions",
-      path_proc: ->(event_id) { event_transactions_path(event_id:) },
+      path_proc: ->(event_id) { ((op = OrganizerPosition.find_by(event_id: event_id, user: current_user)).present? && Flipper.enabled?(:new_ledger_2026_07_17, op) ? event_ledger_path(event_id:) : event_transactions_path(event_id:)) },
```
garyhtou: I noticed that we're using OrganizerPosition.find_by in a few spots. Does that cause issues for suborgs?

garyhtou (commented):
🪄

just to double check, transparency mode is always on the old ledger, right?

app/helpers/events_helper.rb
Luke-Oldenburg: Will replace with User instead.

garyhtou:
@claude , do a review and push a fix for RSpec

garyhtou:
Following up on the earlier review passes: the items that were flagged (and the fix that couldn't be pushed from CI) are now handled.

**RSpec failure (previously couldn't push from CI)** — fixed in `937a10d8c`. `#ledger` now redirects organizers who haven't opted into `new_ledger_2026_07_17`, so the spec's admin enables the flag before rendering. Shard 7/8 is green.

**Two flags gated the ledger** — resolved in `9e917af8d`. `EventPolicy#ledger?` and `LedgerPolicy#show?` now also authorize readers who've opted into `new_ledger_2026_07_17`, not just the older event-keyed `new_ledger_2026_06_30`. Previously an opt-in organizer on an event without `2026_06_30` would hit `head :not_found`; now the opt-in works regardless of that flag's rollout state.

**Toggle button rendered on the event Home page** — resolved in `9e917af8d`. Added a `show_ledger_toggle` local so the "Try the new ledger" button only appears on Transactions and Ledger, matching the screenshot.

**Transparency mode stays on the old ledger** — confirmed. `EventPolicy#ledger?` doesn't honor `is_public`, and the nav link plus the `#transactions` redirect are both gated on `organizer_signed_in?`, so public/transparency viewers never reach the new ledger.

Added spec coverage for the opt-in redirects, the `apply_flipper` toggle, and the policy flag matrix (`5f1c4522c`). All checks green.

_Posted by Claude on behalf of @garyhtou._

garyhtou (approved):

PR #14352, opened 2026-07-18
[Ledger] Speed up fronted_fee_balance_cents / available_balance_cents

garyhtou:
> [!NOTE]
> **This PR was written by [Claude Code](https://claude.com/claude-code).** Gary directed the investigation and reviewed the changes, but the diagnosis, code, and benchmarks below were produced by Claude.

## Summary of the problem

`Ledger#fronted_fee_balance_cents` (which `available_balance_cents` depends on) was slow: 200-420ms per call on large real orgs, tracing back to a real bug plus two full sequential table scans.

## Describe your changes

- `sum_fronted_amount` called `.group(:ledger_item)` — the association, not the `ledger_item_id` column — so ActiveRecord fired an extra `Ledger::Item Load` query just to hydrate full records used only for their `id`. Happened twice per call. Fixed to `.group(:ledger_item_id)`.
- Swapped `.includes(:fee)` for `.joins(:fee)` in `total_fee_payments_cents` to drop an unneeded `LEFT OUTER JOIN`.
- Memoized `fronted_fee_balance_cents` per-instance (matches the existing pattern on `Event`).
- Added two indexes after tracing the remaining cost with `EXPLAIN ANALYZE`: a covering index on `fees(event_id)` including `amount_cents_as_decimal` (the fee-total sum was a full heap scan; now index-only), and an index on `fees(reason)` (the "HACK CLUB FEE" payment lookup was seq-scanning the whole `fees` table for a ~3% match).
- I also tried a partial index on `canonical_pending_transactions` for the fronted/not-waived scope. It looked good in `EXPLAIN` but benchmarking end-to-end made the one event with a real fronted balance ~50% *slower* (a high match-rate ledger doesn't benefit from a system-wide partial index the way a sparse one does), so I left that table unindexed rather than ship a regression for the orgs this method actually matters for.

Benchmarked against real dev data, before → after, across 8 large real orgs (org names redacted; excludes HQ-affiliated orgs, which aren't charged fiscal sponsorship fees, plus 5 further fiscally-sponsored orgs to confirm the fix generalizes):

| Org | Items | Before | After |
|---|---|---|---|
| Org 1 | 25,788 | 198ms | 107ms |
| Org 2 | 18,284 | 221ms | 98ms |
| Org 3 | 10,005 | 417ms | 139ms |
| Org 4 | 7,572 | 71ms | 66ms |
| Org 5 | 6,146 | 243ms | 131ms |
| Org 6 | 5,095 | 215ms | 123ms |
| Org 7 | 4,766 | 225ms | 125ms |
| Org 8 | 4,499 | 192ms | 121ms |

Output values are byte-identical before/after in every case (asserted in the benchmark scripts). `spec/models/ledger_spec.rb` and `spec/models/fee_spec.rb` pass.

This is intentionally a light touch: fees are eventually going to become their own `Ledger::Item` row, at which point `available_balance_cents` will equal `balance_cents` and this whole live computation goes away. No new caching infrastructure added for code that's slated for replacement.


garyhtou:
*(comment written by Claude Code)*

### Why `index_fees_on_event_id_include_amount` is needed alongside the existing `index_fees_on_event_id`

They share the same key column (`event_id`), but they serve different jobs, and Postgres can't get the benefit of the new one from the old one.

**What a plain B-tree index actually stores.** `index_fees_on_event_id` is a B-tree keyed on `event_id` only. Each leaf entry is `(event_id, ctid)` — `ctid` is a pointer to the row's physical location in the table heap. The index does **not** contain `amount_cents_as_decimal`. So a query like:

```sql
SELECT SUM(amount_cents_as_decimal) FROM fees WHERE event_id = 1234567
```

uses the index to cheaply find *which rows* match `event_id = 1234567`, but for every one of those rows it still has to jump out to the heap to read `amount_cents_as_decimal`, because that column isn't in the index. For a large event that's thousands of random-ish heap page reads.

**What `INCLUDE` changes.** `add_index :fees, :event_id, include: [:amount_cents_as_decimal]` builds a B-tree with the *same* search key (`event_id`) but stores a copy of `amount_cents_as_decimal` as extra, non-key payload in the leaf pages. `INCLUDE`d columns aren't part of the key — they don't help `ORDER BY`/range lookups, they're not usable for uniqueness — their only job is to let a query stop at the index and never touch the heap, if every column the query needs is either the search key or an `INCLUDE`d column. That's called an "index-only scan," and it's the whole win here: `SUM(amount_cents_as_decimal)` needs exactly `event_id` (to filter) and `amount_cents_as_decimal` (to sum), and now both live in the index.

**Measured on real dev data** (Hack Club HQ, 25,377 fee rows), via `EXPLAIN (ANALYZE, BUFFERS)`:

Before (only the plain `event_id` index existed):
```
Aggregate  (actual time=152.242..152.243 rows=1 loops=1)
  Buffers: shared hit=1061 read=3684
  ->  Bitmap Heap Scan on fees  (actual time=3.495..148.734 rows=25377 loops=1)
        Recheck Cond: (event_id = 1234567)
        Heap Blocks: exact=4718
        ->  Bitmap Index Scan on index_fees_on_event_id (actual time=2.927..2.928 rows=25377)
Execution Time: 152.355 ms
```
The index scan itself takes ~3ms — the other ~149ms is `Heap Blocks: exact=4718`, i.e. fetching 4,718 8kB heap pages to read the amount off each row.

After (with the covering index):
```
Aggregate  (actual time=4.617..4.617 rows=1 loops=1)
  Buffers: shared hit=1 read=100
  ->  Index Only Scan using index_fees_on_event_id_include_amount on fees (actual time=0.058..3.165 rows=25377)
        Index Cond: (event_id = 1234567)
        Heap Fetches: 0
Execution Time: 4.635 ms
```
`Heap Fetches: 0` — the heap is never touched. ~33x faster on this event (152ms → 4.6ms), and the win scales with row count, so it matters most for exactly the large/old events where this was slowest.

**Why I didn't just replace the old index.** A plain `(event_id)` index is a strict prefix of the new one, so once autovacuum/manual `VACUUM` keeps the visibility map current, `index_fees_on_event_id_include_amount` can serve everything the old index could (existence checks, plain equality filters, joins on `event_id`) — the old one is now redundant and could be dropped as cleanup. I left it in place in this PR to keep the change additive/low-risk on a financial table rather than combining an index swap with the rest of this diff; happy to follow up with a `remove_index` if we want to trim the redundancy.

(Side note on **`Heap Fetches: 0`**, which is doing a lot of work above: index-only scans are only free of heap access when Postgres's visibility map says the relevant pages are all-visible — i.e. no other transaction could still need the old row version. On a table with a healthy autovacuum cadence, like `fees`, that holds true almost all the time; right after a burst of writes it can briefly need a heap fetch to check visibility until the next vacuum catches up.)

garyhtou:
*(comment written by Claude Code)*

### Follow-up: `fronted_fee_balance_cents` is still expensive on large ledgers, and it runs on every ledger page view regardless of filters

While investigating a separate slow-request report (`/:event_id/ledger?q=...`), I profiled this PR's fix against the largest ledger in the dataset (36,432 items) and wanted to flag what I found, since it's directly downstream of this diff.

**Where it's called from:** `app/views/ledgers/_pending_fee_transaction.html.erb` is rendered on *every* `/:event_id/ledger` page load — `show_fee: true` is hardcoded in `app/views/events/ledger.html.erb:9`, unconditional on any active filter (`q`, `direction`, date range, etc.). It calls `ledger.fronted_fee_balance_cents`, which aggregates the ledger's **entire** transaction history — not whatever's currently filtered/displayed.

**Measured on that ledger (36,432 items):**
| | pre-this-PR | with this PR |
|---|---|---|
| single call | 641ms | 280ms |
| as rendered by the partial (calls it twice) | ~1.3s | ~150-280ms (memoization collapses the 2nd call) |

So this PR is a real, measured win here too (roughly 80% off the partial's cost on the largest ledger in the system) — but the remaining ~150-280ms is still paid on **every** page view of this ledger, including ones with a narrow filter that returns a handful of rows, because the fee row's cost scales with total ledger size, not with what's being displayed.

I don't think that's in scope for this PR (it's a rendering/product decision, not a query-correctness one), but wanted to leave the numbers here for whoever picks it up. Options I'd consider: only render the fee row when no filters are active, cache/memoize it independently of the request lifecycle (e.g. per-ledger, short TTL), or — per the "fees as their own `Ledger::Item` row" plan mentioned in this PR's description — this whole live computation may go away once that lands, since the fee balance would just be `items.sum(:amount_cents)` like everything else.

garyhtou (commented):
this one is lower priority

PR #14354, opened 2026-07-18
[Admin Mailer] Prevent from accidentally emailing on non-prod instances

Luke-Oldenburg:
## Summary of the problem

Mirrors of HCB with cron running were sending us emails.

## Describe your changes

Find by user and then send email.


garyhtou (commented):
I'd suggest


```ruby
def engineers
  User.where(email: ["gary@hackclub.com", "luke@hackclub.com", "ian@hackclub.com"]).pluck(:email)
end

mail to: engineers
```

garyhtou (approved):

PR #14355, opened 2026-07-18
[Admin Ledger] Add enhanced warnings when attempting to re-map a transaction

mattsoh:
## Summary of the problem

Sierra's accounting system gets annoyed when we re-map items on the admin ledger.

## Describe your changes

Add more warnings to force you to sincerely reconsider your actions if the mapping is more than 12 hours after the last mapping.

Screenshot 2026-07-18 at 6 59 24 PM
Screenshot 2026-07-18 at 6 59 30 PM
Screenshot 2026-07-18 at 6 59 37 PM


app/controllers/admin_controller.rb
```diff
@@ -55,6 +55,17 @@ def transaction
...
                            end
+
+    # If this transaction was already mapped a while ago, remapping it annoys Sierra's accounting system.
+
+    if @canonical_transaction.canonical_event_mapping.present? &&
+       @canonical_transaction.canonical_event_mapping.created_at < 12.hours.ago
+      @stale_remap = true
+      @remap_confirm_msg = "⚠️ This transaction was already mapped to \"#{@canonical_transaction.event&.name}\" #{helpers.time_ago_in_words(@canonical_transaction.canonical_event_mapping.created_at)} ago. Remapping it now may disrupt our accounting. Are you absolutely sure you want to remap this transaction?"
+      @remap_confirm_phrase = "REMAP #{@canonical_transaction.id}"
+      @remap_after_message = "Please contact Sierra and the HCB Ops team right away to let them know you remapped transaction ##{@canonical_transaction.id}."
```
anishanne: Isn't the user who does the remapping a member of the HCB Ops team?

anishanne (changes requested):

app/controllers/admin_controller.rb
mattsoh: The user needs to warn Sierra specifically, or in the #hcb-ops channel

Luke-Oldenburg: I think Anish has a point about the wording of this.

app/controllers/admin_controller.rb
```diff
@@ -57,6 +57,17 @@ def transaction
...
```
anishanne:
```suggestion
      @remap_after_message = "Please contact Sierra in the #hcb-ops channel to let them know you remapped transaction ##{@canonical_transaction.id}."
```

anishanne (changes requested):
This might work for the wording.

Also, I would generally prefer to trigger an automated email (or similar, such as a Slack webhook) instead of this warning.

app/controllers/admin_controller.rb
mattsoh: Fixed with Anish’s suggested changes

Luke-Oldenburg:
> Also, I would generally prefer to trigger an automated email (or similar, such as a Slack webhook) instead of this warning.

I think we should have an email and this warning. Or maybe add the email and this warning says an email will be sent,

PR #14361, opened 2026-07-19
[Disbursement] Use `Ledger::Item` to find CTs and CPTs

Luke-Oldenburg:
## Summary of the problem

We don't want to rely on HCB code.

## Describe your changes

Find the ledger item using linked object.


garyhtou (approved):
This is good to go; Although, we'll want to update the CT and CPT method override in `Disbursement` to use Ledger::Item instead of HcbCode

https://github.com/hackclub/hcb/blob/c42fc149f7e333d235d412b7e449cf3b9cf70e4a/app/models/disbursement.rb#L223-L229

PR #14363, opened 2026-07-19
[Ledger::Item] Remove `type_metadata`

Luke-Oldenburg:
## Summary of the problem

This method was only being used for humanized type now.

## Describe your changes

Extract `humanized_type` into its own method with a switch.


app/models/ledger/item.rb
```diff
@@ -342,7 +342,42 @@ def map!
...
+      when "BankFee"
+        "Fiscal sponsorship fee"
+      when "FeeRevenue"
+        "Fee revenue"
+      when "Reimbursement::PayoutHolding"
+        "Reimbursement payout holding"
+      when "Reimbursement::ExpensePayout"
+        "Reimbursement"
+      when "CardCharge"
+        "Card charge"
```
Luke-Oldenburg, line 379:
```suggestion
        "Card charge"
      else
        "Bank account transaction"
```

garyhtou (approved):

PR #14377, opened 2026-07-20
Allow removing a primary ledger mapping

YodaLightsabr:
## Summary of the problem

Currently, when admins unmap a CT, it does not also automatically unmap the ledger item.

## Describe your changes

This PR fixes manual unmapping.


Luke-Oldenburg (approved):
lgtm

PR #14381, opened 2026-07-20
[Ledger] Improve performance with Turbo

YodaLightsabr:
## Summary of the problem

The recent addition of the balance, total raised, and total spent statistics is slowing down the new ledger. Eventually we will completely switch them over to use the new ledger, which will make them a lot faster, but in the meantime we need to come up with another solution.

## Describe your changes

This PR converts the statistics that are still computed from the old transaction engine into an async Turbo frame. To prevent them from refreshing every time the user navigates to a different page, I've also moved the ledger into a regular Turbo frame (not async).


Luke-Oldenburg (approved):
lgtm

PR #14382, opened 2026-07-20
[Ledger] Add popovers to new ledger

YodaLightsabr:
## Summary of the problem

The new ledger is missing popovers.

## Describe your changes

Adds popovers to the new ledger.

### DIFF
--- a/app/helpers/popover_helper.rb
+++ b/app/helpers/popover_helper.rb
@@ -41,8 +41,8 @@ module PopoverHelper
       title: item.hcb_code.pretty_title(show_event_name: false, show_amount: true),
       src: item.hcb_code.popover_path,
       frame_id: item.hcb_code.public_id,
-      state_url: auditor_signed_in? ? ledger_item_path(item) : hcb_code_path(item.hcb_code),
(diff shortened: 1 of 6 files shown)


app/helpers/popover_helper.rb
```diff
@@ -36,6 +36,16 @@ def hcb_code_popover_data(hcb_code, event: nil, **popover_path_params)
     )
   end
 
+  def ledger_item_popover_data(item)
+    popover_data(
+      title: item.hcb_code.pretty_title(show_event_name: false, show_amount: true),
+      src: item.hcb_code.popover_path,
+      frame_id: item.hcb_code.public_id,
+      state_url: auditor_signed_in? ? ledger_item_path(item) : hcb_code_path(item.hcb_code),
+      external_link: auditor_signed_in? ? ledger_item_path(item) : url_for(item.hcb_code)
```
garyhtou: Instead of having a conditional here, what I'd recommend is sending everyone to the Ledger item show page and then doing a conditional redirect to the HCB Code show page if they're not an Auditor

garyhtou (approved):
good to merge after the feedback

PR #14384, opened 2026-07-20
[Payments] Improve transition from old transfer system

manu-gurudath:
The goal of this PR is to improve the user experience from the old transfer system. 

**Goals**
-   remove the "legacy" visual separation and add an `imported_at` column instead. its only job is to answer one thing: did this recipient come from the old transfer system?
-   run a one-time backfill  that keeps each recipient's *name and email* and throws away the rest. unless the organizer clicks "I want to submit tax and payout information on behalf of the recipient"
-   when the organizer clicks that button, HCB prefills the method section
    -   if they've only used one payment method for this recipient, we reuse the existing masked inputs (e.g. the ACH one)
    -   if they've used multiple methods (e.g. ach and wire), we show a multiple-choice field asking how they want to send the payment
-   unmanaged payees won't have any prefilled data because we can't assume `asdf@hackclub.com` has access to the bank account the original payment went to. those recipients configure their own payout method themselves (which they'd have to do under the new system anyway)


polypixeldev (commented):
- I don't think we need to visually separate legacy and new payees or have any organizer-facing communication about it. In the old system, they had their info saved, and it'll be the same in the new system.
- Admins should still have the send transfer manually button for now
- I don't think we should be querying the old models like PaymentRecipient - that keeps our new system reliant on these old models and more complex. Instead, I think we should just have a maintenance task import PaymentRecipient records over to Payee. We don't even need to migrate the payout methods themselves, just name and email - the recipient can re-input their payout info as an actual payout method. I would also say that we shouldn't look at the underlying ach/check/wire models, just payment recipient.

garyhtou:
@manuthecoder and I did a quick verbal review of this. Items for Manu to follow up on:
- Security review (we probably can't securely allow payout info entered by an organizer to later be viewed by a contractor)
- Should we backfill Payee instead of the current solution? The current solution to search for legacy payees on page load requires us to keep that code around forever, compared to backfill, which is a one-and-done job. Backfill would then also give us a uniform data shape.
- Work with @polypixeldev on these!

PR #14385, opened 2026-07-20
[CardCharge] Use merchant columns for icon method

YodaLightsabr:
## Summary of the problem

CardCharge now has the merchant data that we need, so we no longer need to look up the RawStripeTransaction(s) or RawPendingStripeTransaction.

## Describe your changes

This PR changes `CardCharge#icon` to use first-party CardCharge data instead of looking up the raw / raw pending transaction types.


Luke-Oldenburg (approved):
lgtm

PR #14389, opened 2026-07-21
[Event] Only force transparency for Argosy sub-organizations

garyhtou:
## Summary of the problem

Any organization whose parent organization was transparent had transparency mode forced on and could not turn it off. That rule should only apply to Argosy grantees, not to every sub-organization.

## Describe your changes

Narrows the forcing rule to organizations on the 2025 and 2026 Argosy plans, via a new `Event::Plan#forces_transparency?` predicate. It returns `false` on the base plan and `true` on `Argosy2025` / `Argosy2026`, so the rule lives with the plan rather than as a type check in `Event`. `ArgosyFtcSim2025` inherits from `Argosy2025`, so it is covered too.

`Event#eligible_for_disabling_transparency?` becomes `Event#forced_transparency?`, which reads better at both call sites. The transparency mode mailer uses its inverse to decide whether to tell managers they can turn transparency off.

Note that `plan` can still be nil the first time `forced_transparency?` runs on a new record: the fallback plan is built by a `before_validation` callback registered after the transparency one, hence the nil guard.

Specs cover the plan predicate (including inheritance) and the enforcement behavior on the model.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -876,8 +876,12 @@ class Event < ApplicationRecord
     !plan.is_a?(Event::Plan::SalaryAccount)
   end
 
-  def eligible_for_disabling_transparency?
-    !parent&.is_public?
+  def forced_transparency?
+    # `plan` is built by a later `before_validation` callback, so it can still
+    # be nil the first time this runs on a new record.
+    return false unless plan&.forces_transparency?
+
+    parent&.is_public? || false
   end
 
   def eligible_for_indexing?
@@ -1027,7 +1031,7 @@ class Event < ApplicationRecord
       self.is_indexable = false
     end
 
-    unless eligible_for_disabling_transparency?
+    if forced_transparency?
       self.is_public = true
     end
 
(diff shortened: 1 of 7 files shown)


app/models/event.rb
```diff
@@ -876,8 +876,12 @@ def eligible_for_transparency?
...
   end
 
-  def eligible_for_disabling_transparency?
-    !parent&.is_public?
+  def forced_transparency?
+    # `plan` is built by a later `before_validation` callback, so it can still
+    # be nil the first time this runs on a new record.
+    return false unless plan&.forces_transparency?
+
+    parent&.is_public? || false
```
garyhtou, line 884: I actually don't think we need to preserve this `parent&.is_public?` behavior. but i'm going to keep it just in case.

garyhtou:
This change was made because we are moving all HQ-related organizations to be a sub-organization of Hack Club HQ. Since Hack Club HQ is transparent, that would have forced all of its sub-organizations to also be transparent. However, Hack Club HQ has some non-transparent organizations, such as upcoming unannounced events/programs and certain funds.

So this change was made as a quick fix to only force transparency mode where absolutely necessary.

PR #14391, opened 2026-07-21
[Contractors] Improve rate functionality

manuthecoder:
Adds functionality for organizers to invite contractors with more flexible rate types

image
image
image


app/controllers/payroll/positions_controller.rb
```diff
@@ -92,6 +93,7 @@ def update
       @position.assign_attributes(
         title: position_params[:title],
         rate_cents: Monetize.parse(position_params[:rate]).cents,
+        rate_unit: position_params[:rate_unit].presence || "hour",
         start_date: position_params[:starts_on],
         end_date: position_params[:ends_on],
         description: position_params[:purpose]
```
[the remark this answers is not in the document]
manuthecoder: no, we always want it to be hour by default. people should not submit it blank anyways as it's marked required

YodaLightsabr (commented):
Looks pretty good

polypixeldev (approved):

app/models/payroll/position.rb
```diff
@@ -205,7 +227,7 @@ def send_contract(organizer_user: nil, cosigner_email: nil, reissue_messages: {}
             "payee_name"  => payee.display_name,
             "title"       => title,
             "description" => description,
-            "rate"        => rate.format,
+            "rate"        => rate_label,
```
garyhtou, line 230: Does the DocuSeal contract already suffix the rate with "hours"? If so, we'll need to remove it

garyhtou: Okay, it does. This needs updating

Image

app/controllers/payroll/positions_controller.rb
```diff
@@ -92,6 +93,7 @@ def update
       @position.assign_attributes(
         title: position_params[:title],
         rate_cents: Monetize.parse(position_params[:rate]).cents,
+        rate_unit: position_params[:rate_unit].presence || @position.rate_unit,
```
garyhtou: Why is `rate_unit` special and needs a `|| @position.rate_unit` the rest of the attributes only pulls from `params`

garyhtou (approved):
otherwise, LGTM

app/models/payroll/position.rb
garyhtou: @polypixeldev , can you help with this?

garyhtou:
Good to merge after those two are fixed up

app/models/payroll/position.rb
polypixeldev: Updated on DocuSeal!

polypixeldev:
Please also change the prefill name in the Contract::PayrollPosition class to just be "Rate" instead of "Hourly Rate" - once this is merged I'll update it in DocuSeal to avoid any errors.

app/models/payroll/position.rb
garyhtou: ty!

PR #14396, opened 2026-07-22
[CardCharge] Make merchant ID and category not null

Luke-Oldenburg:
## Summary of the problem

These columns have been backfilled.

## Describe your changes

Make them not null. No rails validation added because it would run before the `before_create` that fills these values.


garyhtou (commented):
This is safe as long as there is never a moment in time where CardCharge is not associated with any `raw_stripe_transactions` nor `raw_pending_stripe_transaction`.

I would double-check that, upon creation, at least one of those associations is present. Based on how we wrote some of the methods in this model, it appears to assume that it's possible to have no raw transactions associated. For example:
```ruby
def stripe_card
  (raw_stripe_transactions.last || raw_pending_stripe_transaction)&.stripe_card
end
# ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――⤴︎
# Having the safe nav handles the case where no raw transactions exist.
```
I'm not sure if it's possible for no raw transactions to exist, but we should double check that before merging. Maybe @YodaLightsabr knows more.

garyhtou:
It may still be worth adding a Rails validation for update (and not create).

PR #14403, opened 2026-07-22
[Reimbursements] Properly handle failed check payouts

polypixeldev:
## Summary of the problem

We never marked the payout holding as failed when handling a failed check payout, and never sent an email.

## Describe your changes

Mark the payout holding as failed so that it doesn't try again, and send an email to the user about the error.

### DIFF
--- a/app/views/reimbursement_mailer/check_failed.html.erb
+++ b/app/views/reimbursement_mailer/check_failed.html.erb
@@ -1,0 +1,5 @@
+We're writing to let you know your payout for "<%= link_to @report.name, @report %>" failed to send. Here's the explanation given by our bank:
+
+<pre><%= @reason %></pre>
+
+Please update your payout details <%= link_to "here", settings_payouts_url %>; afterwards, we'll reattempt the transfer.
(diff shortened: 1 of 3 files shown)


app/views/reimbursement_mailer/check_failed.html.erb
```diff
@@ -0,0 +1,5 @@
+We're writing to let you know your payout for "<%= link_to @report.name, @report %>" failed to send. Here's the explanation given by our bank:
```
Luke-Oldenburg, line 1: @garyhtou thoughts on the phrase "our bank"?

Luke-Oldenburg (approved):
lgtm but would like @garyhtou to review the mailer copy

app/views/reimbursement_mailer/check_failed.html.erb
garyhtou: i can't really think of a better phrase besides "the bank", but i'm not sure if that's any better.

garyhtou (approved):

PR #14404, opened 2026-07-22
[Contractors] Move contractors up in event nav and remove payments beta badge

polypixeldev:
## Summary of the problem

It makes more sense for contractors to be next to payments, and payments shouldn't be marked as beta as it's replacing transfers.


manuthecoder (approved):
lgtm!

Luke-Oldenburg (approved):
lgtm

PR #14406, opened 2026-07-22
[Contractors] Add contractor onboarded email

polypixeldev:
## Summary of the problem

We don't notify the event when the contractor has completed onboarding.

## Describe your changes

Add email to event managers when a contractor finishes onboarding.


Luke-Oldenburg (approved):
lgtm

PR #14407, opened 2026-07-22
Rename erb files without a format extension

polypixeldev:
## Summary of the problem

Found 6 files that just ended in .erb instead of .html.erb. This often happens to me when renaming a .html.erb file. While not required by Rails, all files should still have one. ERB lint also seems to skip lint if it doesn't have it

## Describe your changes

Rename the files and fix lints.


app/views/hcb_codes/link_receipt.html.erb
```diff
@@ -26,10 +26,10 @@
     <%= link_to my_inbox_path, target: "_blank", class: "btn btn--primary mr2" do %>
       Manage receipts
     <% end %>
-    <%= form.submit "Add to transaction", class: "btn bg-success", data: {
-          "receipt-select-target" => "confirm"
-    }, disabled: true %>
-    
+      <%= form.submit "Add to transaction", class: "btn bg-success", data: {
```
Luke-Oldenburg: I think this should be inline with the block from above?

Luke-Oldenburg (approved):
lgtm

PR #14418, opened 2026-07-27
Update payout method settings tab copy to include payments

polypixeldev:
## Summary of the problem

The payout settings tab is currently named and has copy specific to reimbursements and a single payout method, even though it is now also used for payments and can have multiple payout methods.

## Describe your changes

Update the tab name and copy to include payments.


Luke-Oldenburg (approved):
lgtm

PR #14296, opened 2026-07-15
[Contractors] Fix policy and send email reminders

manuthecoder:
* Fixes policy so people without transfer perms can't access payments page
* Sends email reminders to contractors

PR #14419, opened 2026-07-27
[Tax] Refresh contractors and payments on TIN match success

polypixeldev:
## Summary of the problem

If we're waiting on a TIN match success to pay someone, we need to refresh the contractors/payments when that happens.

## Describe your changes

Refactor the refresh method to refresh both payments and contractors, and call it on tin match success.

This also removes sending the payout method missing email after tax information is submitted - this is because otherwise it would be sent multiple times each time the refresh method was called, and #14296 will add reminder emails that will take its place.


Luke-Oldenburg (approved):
lgtm

PR #14423, opened 2026-07-27
Remove undefined receipt method from HcbCodesController action callbacks

polypixeldev:
## Summary of the problem

We recently turned on the config to raise errors if there's a undefined action in a controller callback - turns out HcbCodesController has one.

## Describe your changes

Remove the undefined action from the callbacks.


Luke-Oldenburg (approved):
lgtm

PR #14365, opened 2026-07-20
Rails 8.1

albertchae:
## Describe your changes
Upgrade to rails 8.1. Best reviewied by commits. Upgraded any gems that were incompatible with rails 8.1, copied over comments/new settings from the rails update generator, and fixed a few low hanging deprecation warnings.

One deprecation warning this PR does not handle is
```
DEPRECATION WARNING: ActiveSupport::Configurable is deprecated without replacement, and will be removed in Rails 8.2.<D-4>

You can emulate the previous behavior with `class_attribute`.
 (called from <top (required)> at hcb/config/application.rb:12)
 ```
 This requires a major version upgrade to grape.

PR #14424, opened 2026-07-28
[Schema] Upgrade to 8.1 & remove create_schema "google_sheets" 

garyhtou:
## Summary of the problem

`create_schema "google_sheets"` is not managed by the Rails app and shouldn't be in `db/schema.rb`

## Describe your changes

db/schema.rb` has been carrying `create_schema "google_sheets"`, a schema that isn't created or managed by our migrations (nothing in `app/`, `lib/`, `config/`, or `db/migrate/` references it). It only exists on databases where the external sync has run, so the line appeared or disappeared depending on which database the person running `db:migrate` was attached to.

Which schemas get dumped is controlled by `ActiveRecord.dump_schemas`, which defaults to `:schema_search_path`. Before Rails 8.1, though, the `:ruby` dumper ignored that setting and emitted `create_schema` for every non-`public` schema it found. Rails 8.1 fixed it:

> Make schema dumper to account for `ActiveRecord.dump_schemas` when dumping in
> `:ruby` format.
>
> *fatkodima* ([activerecord CHANGELOG](https://github.com/rails/rails/blob/v8.1.3/activerecord/CHANGELOG.md#L375))

We upgraded in https://github.com/hackclub/hcb/pull/14365, but nothing has re-dumped the schema since, so the stale line is still there. With our search path set to `public`, the dumper no longer emits it, verified locally against a database that does have the `google_sheets` schema present. No config change is needed, and the line won't come back.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -13,8 +13,6 @@
 # It's strongly recommended that you check this file into your version control system.
 
 ActiveRecord::Schema[8.1].define(version: 2026_07_20_210532) do
-  create_schema "google_sheets"
-
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
   enable_extension "pg_catalog.plpgsql"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -12,7 +12,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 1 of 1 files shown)


Luke-Oldenburg (approved):
lgtm

PR #14428, opened 2026-07-28
[Ledger::Item] Sync `marked_no_or_lost_receipt_at` from `HcbCode`

Luke-Oldenburg:
## Summary of the problem

We were not keeping this in sync which caused a regression.

## Describe your changes

Keep in sync and add a backfill maintenance task.

### DIFF
--- a/app/models/hcb_code.rb
+++ b/app/models/hcb_code.rb
@@ -78,6 +78,12 @@ class HcbCode < ApplicationRecord
 
   after_create :write_event_and_subledger_id
 
+  # No/lost receipt marks are still written to the HCB code (by `no_or_lost_receipt!`
+  # and by attaching/removing a receipt), but the ledger item keeps its own copy for
+  # `Ledger::Item#missing_receipt?` and the ledger filters, so mirror the change over.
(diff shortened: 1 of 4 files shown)


app/models/hcb_code.rb
garyhtou, line 1: How about vice versa? When `LedgerItem#marked_no_or_lost_receipt_at` is updated, should we sync it to HcbCode?

Otherwise, the old ledger may fall out of sync.

app/tasks/maintenance/backfill_ledger_item_no_or_lost_receipts_task.rb
```diff
@@ -0,0 +1,32 @@
...
+    def process(hcb_code)
+      hcb_code.sync_marked_no_or_lost_receipt_at_to_ledger_item
+    end
+
+    private
+
+    def mismatched
+      HcbCode
+        .joins(:ledger_item)
+        .where("hcb_codes.marked_no_or_lost_receipt_at IS DISTINCT FROM ledger_items.marked_no_or_lost_receipt_at")
```
garyhtou, line 27: cool sql!

garyhtou (approved):

garyhtou (approved):

PR #14430, opened 2026-07-29
Modify Intercom settings to disable for unauthenticated users

YodaLightsabr:
## Summary of the problem

Users are having difficulty navigating between pages on HCB when not signed in. Actions like switching authentication methods or completing a login will get a response from the server, but the user is not able to progress.

Intercom rewrites `history.pushState` but has a buggy implementation resulting in errors on navigation.

## Describe your changes

This PR modifies Intercom's configuration to disable Intercom when the user is not signed in. This works around Intercom's bug hidden inside their `history.pushState` implementation.

### DIFF
--- a/app/controllers/logins_controller.rb
+++ b/app/controllers/logins_controller.rb
@@ -14,6 +14,8 @@ class LoginsController < ApplicationController
     response.delete_header("X-Robots-Tag")
   end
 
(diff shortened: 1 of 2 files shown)


Luke-Oldenburg (approved):
lgtm

PR #14434, opened 2026-07-29
[Search] Add clickable search button

manuthecoder:
image

Adds a button which pops up when you start typing in the search box field


YodaLightsabr (changes requested):
Code looks good, nice work on this one. Quick nitpick: is there a world where we can improve the look of the padding here?

Image

YodaLightsabr (approved):
Once you fix that ^ it's good to merge

PR #14440, opened 2026-07-30
[Ledger::Item] Fix nil error in show page on HCB code

Luke-Oldenburg:
## Summary of the problem

Some ledger items don't have an HCB code.

## Describe your changes

Check for the presence.

### DIFF
--- a/app/views/ledger/items/show.html.erb
+++ b/app/views/ledger/items/show.html.erb
@@ -28,7 +28,7 @@
       <%= @item.datetime.strftime("%b %e, %Y") %>
     </span>
   </div>
-  <% if @item.hcb_code %>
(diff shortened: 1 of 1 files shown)


app/views/ledger/items/show.html.erb
```diff
@@ -28,14 +28,16 @@
...
   </div>
-  <div>
-    <strong>
-      HCB Code
-    </strong>
-    <span>
-      <%= link_to @item.hcb_code.hcb_code, hcb_code_path(@item.hcb_code), target: :_blank %>
-    </span>
-  </div>
+  <% if @item.hcb_code %>
```
Luke-Oldenburg:
```suggestion
  <% if @item.hcb_code.present? %>
```

PR #14444, opened 2026-07-30
[Ledger::Item] Make status nonnull

Luke-Oldenburg:
## Summary of the problem

Status was nullable.

## Describe your changes

Default status to pending.


app/models/ledger/item.rb
```diff
@@ -194,7 +194,7 @@ def calculate_status
       end
 
       # Nothing has mapped to this item yet
-      :pending
+      :settled
```
garyhtou: keep as pending, and cache CT&CPT counts and filter out empty LI before merging.

in the future, we'll add a job to destroy empty LIs

garyhtou: How is "empty LI" defined? if there are no CTs and CPTs, there is a linked object, is it empty?

we prob don't consider those as empty

app/models/ledger/item.rb
```diff
@@ -16,7 +16,7 @@
 #  receipt_count                :integer          default(0), not null
 #  receipt_required             :boolean
 #  short_code                   :text
-#  status                       :string
+#  status                       :string           default("settled"), not null
```
Luke-Oldenburg:
```suggestion
#  status                       :string           default("pending"), not null
```

app/models/ledger/item.rb
Luke-Oldenburg:
```suggestion
      :pending
```

db/migrate/20260730222816_add_status_default_to_ledger_items.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddStatusDefaultToLedgerItems < ActiveRecord::Migration[8.1]
+  def change
+    # `settled` is what `Ledger::Item#calculate_status` falls back to when nothing
```
Luke-Oldenburg:
```suggestion
    # `pending` is what `Ledger::Item#calculate_status` falls back to when nothing
```

db/migrate/20260730222816_add_status_default_to_ledger_items.rb
```diff
@@ -0,0 +1,11 @@
...
+    # has mapped to an item yet, so a row inserted before its first refresh reads
+    # the same way it will afterwards.
+    change_column_default :ledger_items, :status, from: nil, to: "settled"
```
Luke-Oldenburg:
```suggestion
    change_column_default :ledger_items, :status, from: nil, to: "pending"
```

db/migrate/20260730222821_backfill_ledger_item_statuses.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# `SET DEFAULT` only applies to newly inserted rows, so items created before
+# `status` existed still need a value before NOT NULL can be enforced. `settled`
```
Luke-Oldenburg:
```suggestion
# `status` existed still need a value before NOT NULL can be enforced. `pending`
```

spec/models/ledger/item_spec.rb
```diff
@@ -109,6 +109,16 @@
...
     end
 
+    it "requires status" do
+      item = Ledger::Item.new(amount_cents: 1000, memo: "Test", datetime: Time.current, status: nil)
+      expect(item).not_to be_valid
+      expect(item.errors[:status]).to include("can't be blank")
+    end
+
+    it "defaults status to settled" do
+      expect(Ledger::Item.new.status).to eq("settled")
```
Luke-Oldenburg:
```suggestion
    it "defaults status to pending" do
      expect(Ledger::Item.new.status).to eq("pending")
```

db/migrate/20260730222821_backfill_ledger_item_statuses.rb
```diff
@@ -0,0 +1,17 @@
...
+# is both the new column default and what `calculate_status` falls back to; each
+# item's real status lands on its next refresh.
+class BackfillLedgerItemStatuses < ActiveRecord::Migration[8.1]
+  def up
+    Ledger::Item.where(status: nil).update_all(status: "settled")
```
Luke-Oldenburg:
```suggestion
    Ledger::Item.where(status: nil).update_all(status: "pending")
```

app/models/ledger/item.rb
Luke-Oldenburg: It's hard to say because that should never happen in production in the new system. I think it should be safe to nuke in that case because the LO can always make a new LI later. @YodaLightsabr thoughts?

garyhtou (approved):

PR #14448, opened 2026-07-31
[Ledger::Item] Add `ct_count` and `cpt_count`

Luke-Oldenburg:
## Summary of the problem

These numbers will be used to filter out empty transactions on the ledger. This will also improve the performance of the admin ledger.

## Describe your changes

Cache ct and cpt counts.


app/models/ledger/item.rb
```diff
@@ -310,6 +312,8 @@ def refresh!
 
       self.amount_cents = calculate_amount_cents
       self.author = calculate_author
+      self.ct_count = canonical_transactions.count
+      self.cpt_count = canonical_pending_transactions.count
```
garyhtou:
```suggestion
      self.ct_count = canonical_transactions.size
      self.cpt_count = canonical_pending_transactions.size
```
https://medium.com/unagi/ruby-on-rails-length-vs-size-vs-count-with-examples-a32e17274c79

app/tasks/maintenance/backfill_transaction_count_ledger_items_task.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module Maintenance
+  class BackfillTransactionCountLedgerItemsTask < MaintenanceTasks::Task
+    def collection
+      Ledger::Item.all
+    end
+
+    def process(ledger_item)
+      ledger_item.ct_count = ledger_item.canonical_transactions.count
+      ledger_item.cpt_count = ledger_item.canonical_pending_transactions.count
```
garyhtou: Are we doing this instead of calling `refresh!` because refresh is more expensive to run?

```suggestion
      ledger_item.ct_count = ledger_item.canonical_transactions.size
      ledger_item.cpt_count = ledger_item.canonical_pending_transactions.size
```

garyhtou (approved):

Luke-Oldenburg:
@claude any concerns to be had over using size instead of count especially relating to needing to reload relations?

PR #14450, opened 2026-07-31
[CardCharge] Associate `StripeCard`

Luke-Oldenburg:
## Summary of the problem

We were lacking an association which is needed for performance.

## Describe your changes

Store `stripe_card_id` on `card_charges`

### DIFF
--- a/app/models/card_charge.rb
+++ b/app/models/card_charge.rb
@@ -10,14 +10,17 @@
 #  updated_at                        :datetime         not null
 #  merchant_network_id               :string
 #  raw_pending_stripe_transaction_id :bigint
+#  stripe_card_id                    :bigint
 #
 # Indexes
 #
 #  index_card_charges_on_raw_pending_stripe_transaction_id  (raw_pending_stripe_transaction_id) UNIQUE
+#  index_card_charges_on_stripe_card_id                     (stripe_card_id)
 #
 # Foreign Keys
 #
 #  fk_rails_...  (raw_pending_stripe_transaction_id => raw_pending_stripe_transactions.id) ON DELETE => nullify
+#  fk_rails_...  (stripe_card_id => stripe_cards.id)
 #
 # Raw objects are matched to their charge purely by Stripe IDs: a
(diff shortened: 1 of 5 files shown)


app/models/card_charge.rb
```diff
@@ -39,9 +43,12 @@ class CardCharge < ApplicationRecord
   }
 
   before_create :set_merchant_data
+  before_create :set_stripe_card
 
+  # TODO: remove the fallback lookup once stripe_card_id has been backfilled
```
garyhtou, line 48: good!

app/models/card_charge.rb
```diff
@@ -124,4 +131,8 @@ def set_merchant_data
     self.merchant_category ||= merchant_data&.[]("category")
   end
 
+  def set_stripe_card
+    self.stripe_card_id ||= stripe_card&.id
```
garyhtou: this doesn't work long term once you remove the definition of the `stripe_card` method. copy down the logic from that method (line 44)

app/tasks/maintenance/backfill_card_charge_stripe_cards_task.rb
```diff
@@ -0,0 +1,18 @@
...
+  # association existed. CardCharge#stripe_card falls back to deriving the
+  # card from raw transactions when stripe_card_id is nil, so this can run
+  # at any time without breaking existing lookups.
+  class BackfillCardChargeStripeCardsTask < MaintenanceTasks::Task
+    def collection
+      CardCharge.where(stripe_card_id: nil)
+    end
+
+    def process(card_charge)
+      card_charge.update!(stripe_card_id: card_charge.stripe_card&.id)
```
garyhtou: this works! or, maybe it's more robust to call the set stripe card method then save!

garyhtou (approved):
would u also want a stripe cardholder id or user id on card charge in the future?

Luke-Oldenburg:
> would u also want a stripe cardholder id or user id on card charge in the future?

I don't think these are as important because they can be easily derived from the ledger item's user unless we add support for multiple cardholders in the future.

PR #14452, opened 2026-07-31
[Ledger::Item] Display merchant name in icon tooltip

Luke-Oldenburg:
## Summary of the problem

Card charges have custom icons but not custom tooltips.

## Describe your changes

Display the merchant name when available.

### DIFF
--- a/app/models/ledger/item.rb
+++ b/app/models/ledger/item.rb
@@ -375,7 +375,7 @@ class Ledger
       when "Reimbursement::ExpensePayout"
         "Reimbursement"
       when "CardCharge"
-        "Card charge"
+        YellowPages::Merchant.lookup(network_id: linked_object.merchant_network_id || "").name.presence || "Card charge"
       else
         "Bank account transaction"
       end
--- a/app/models/ledger/item.rb
+++ b/app/models/ledger/item.rb
@@ -375,7 +375,7 @@ class Ledger
       when "Reimbursement::ExpensePayout"
         "Reimbursement"
       when "CardCharge"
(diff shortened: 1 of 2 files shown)


garyhtou (commented):
this might make sense to do, although i need to understand where we use humanized type and if we want the other types to reflect a similar behavior.

that we're doing here is showing the counterparty's name, rather than the transfer method's name.

let's not merge this just yet

Luke-Oldenburg:
> this might make sense to do, although i need to understand where we use humanized type and if we want the other types to reflect a similar behavior.

Currently it is just used for the icon tooltip on the ledger. Perhaps we should extract that into a separate method that this has the ability to override?

app/models/ledger/item.rb
```diff
@@ -375,7 +375,7 @@ def humanized_type
       when "Reimbursement::ExpensePayout"
         "Reimbursement"
       when "CardCharge"
-        "Card charge"
+        YellowPages::Merchant.lookup(network_id: linked_object.merchant_network_id || "").name.presence || "Card charge"
```
garyhtou: Notes from call: We both agree it doesn't belong inside `humanized_type`.

Alternative solutions:
- Dectorator wrapping `Ledger::Item` that defines `icon_tooltip`. This is because `icon_tooltip` is only useful for the purpose of rendering a ledger; `Ledger::Item` should stay generic and not have view specific code.
- Inline `YellowPages::Merchant.lookup` when rendering the ledger

garyhtou (approved):
Good to merge after the changes

PR #14460, opened 2026-08-01
[Contractors] Prevent 404s

manuthecoder:
Fixes a bug where you could get stuck on an unsubmittable contractor form. The form now only shows once a recipient is picked, bad submissions re-render with a clear error instead of a 404, and failed submissions keep what you typed (except the PDF attachment).

### DIFF
--- a/app/views/payroll/positions/_form.html.erb
+++ b/app/views/payroll/positions/_form.html.erb
@@ -30,7 +30,7 @@
   <% end %>
 
   <% form_url = editing ? event_payroll_position_path(event_id: event.slug, id: position.id) : event_payroll_positions_path(event_id: event.slug) %>
-  <%= form_with(url: form_url, method: editing ? :patch : :post, scope: :contractor, local: true) do |form| %>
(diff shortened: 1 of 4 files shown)


app/views/payroll/positions/_form.html.erb
```diff
@@ -30,7 +30,7 @@
   <% end %>
 
   <% form_url = editing ? event_payroll_position_path(event_id: event.slug, id: position.id) : event_payroll_positions_path(event_id: event.slug) %>
-  <%= form_with(url: form_url, method: editing ? :patch : :post, scope: :contractor, local: true) do |form| %>
+  <%= form_with(url: form_url, method: editing ? :patch : :post, scope: :contractor, local: true, class: ("hidden" unless payee)) do |form| %>
```
garyhtou, line 33: Should we do an `if` instead of hidden css? Which is better?

garyhtou (approved):

app/views/payroll/positions/_form.html.erb
manuthecoder: stimulus depends on some of these form elements but the logic is validated on the controller side too which prevents this error from occurring

PR #14461, opened 2026-08-01
[Admin] Fix orphaned ledger item check never matching

garyhtou:
> [!NOTE]
> **This PR was written by [Claude Code](https://claude.com/claude-code).** Gary directed the investigation and reviewed the changes, but the diagnosis and code below were produced by Claude.

## Summary of the problem

Found while investigating Postgres host load. `Admin::DetectLogicalTransactionAnomaliesJob` runs hourly and has two checks; the second one has never been able to report anything.

It looked for ledger items that no HCB code points at with:

```ruby
Ledger::Item.where.not(id: HcbCode.where(event_id:).select(:ledger_item_id))
```

`where.not(id: <subquery>)` compiles to `NOT IN`. `hcb_codes.ledger_item_id` is nullable, and in SQL `x NOT IN (1, 2, NULL)` is UNKNOWN rather than true for every `x` — so as soon as one HCB code on the event has no ledger item, the predicate matches **no rows at all**. The check silently returned an empty set instead of the anomalies it exists to surface.

Filtering the NULLs out on its own would swing it to the opposite failure. The query compared *every* `Ledger::Item` in the table against a *single* event's HCB codes, so with the NULLs gone it would report essentially every other event's items as anomalies for this one — and then enumerate them with `find_each` and email the result.

So the NULL bug was masking a scope bug, and fixing either one alone is wrong.

## Describe your changes

Both are fixed together:

- **Scope to the event's own ledger.** The check now starts from `event.ledger.items` instead of `Ledger::Item.all`, which is what "items on this event's ledger with no HCB code" meant in the first place.
- **Express it as a LEFT JOIN anti-join** rather than `NOT IN`. This is NULL-safe by construction, and it uses the existing index on `hcb_codes.ledger_item_id` instead of materialising a subquery. `IS DISTINCT FROM` keeps a joined row with a NULL `event_id` counting as a mismatch rather than being dropped.

Also hoisted the `Event.find` that the mailer call was doing separately, and returns early when the event has no ledger.

## Tests

Added `spec/jobs/admin/detect_logical_transaction_anomalies_job_spec.rb` covering the orphan case, the NULL case that broke it, the linked case that should stay quiet, an item whose HCB code belongs to a different event, and an item on another event's ledger.

Verified the specs discriminate: **2 of 5 fail against the pre-fix job and all 5 pass with the fix.** Rubocop is clean.

One thing worth knowing for anyone writing similar specs: `HcbCode`'s `after_create :write_event_and_subledger_id` derives `event_id` from the code's transactions and writes `nil` when there aren't any, so the spec sets those columns past the callback.

## Note on rollout

This check has been dormant, so the first run after merge may surface a real backlog of anomalies rather than the empty result the job has been sending. Worth watching the first mail before assuming a regression.

### DIFF
--- a/app/jobs/admin/detect_logical_transaction_anomalies_job.rb
+++ b/app/jobs/admin/detect_logical_transaction_anomalies_job.rb
@@ -5,6 +5,8 @@ module Admin
     queue_as :low
 
     def perform(event_id: 183)
+      event = Event.find(event_id)
+
       hcb_codes = []
       HcbCode.where(event_id:).on_main_ledger.find_each do |hcb_code|
         if !(hcb_code.ledger_item.nil? && hcb_code.no_transactions?) && (hcb_code.ledger_item.nil? || hcb_code.smart_amount_cents != hcb_code.ledger_item.amount_cents)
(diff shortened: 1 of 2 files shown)


garyhtou:
@YodaLightsabr owns this PR from now on

PR #14463, opened 2026-08-01
[Payments] Add ⇄ button

manuthecoder:
Adds ⇄ button to payment UI!

image

---

image


garyhtou (commented):
Let's move the swap button so it's in the middle of the two fields, so that it's like how Google Flight works.

garyhtou (approved):
Let's move the swap button so it's in the middle of the two fields, so that it's like how Google Flight works.

PR #14468, opened 2026-08-02
[Contractors] [Payments] Show unauthorized callout for users who don't have permission

manuthecoder:
image

It's weird that the form fields are just disabled when users don't have access
It also adds payments/contractors to the `/roles` page

### DIFF
--- a/app/controllers/static_pages_controller.rb
+++ b/app/controllers/static_pages_controller.rb
@@ -120,7 +120,22 @@ class StaticPagesController < ApplicationController
           "Cancel an HCB Transfer": :manager,
           "View an HCB Transfer": :reader
         },
+        Payments: {
+          "Send a payment": :manager,
+          "Cancel a payment": :manager,
+          "View a payment": :reader,
+          "View the organization's recipients": :manager,
+          "Add a recipient": :manager,
+          "Edit or archive a recipient": :member,
+          _preface: "For paying people & vendors"
+        },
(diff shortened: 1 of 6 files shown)


app/controllers/static_pages_controller.rb
```diff
@@ -120,8 +120,23 @@ def roles
           "Cancel an HCB Transfer": :manager,
           "View an HCB Transfer": :reader
         },
+        Payments: {
+          "Send a payment": :manager,
+          "Cancel a payment": :manager,
+          "View a payment": :reader,
+          "View the organization's recipients": :manager,
+          "Add a recipient": :manager,
+          "Edit or archive a recipient": :member,
```
garyhtou: Make this a manager. Update the policy too

app/controllers/static_pages_controller.rb
```diff
@@ -120,8 +120,23 @@ def roles
           "Cancel an HCB Transfer": :manager,
           "View an HCB Transfer": :reader
         },
+        Payments: {
```
garyhtou, line 123: Payments is basically replacing ACH transfer, checks, and all the other ones, which means that it may be confusing to have all of them showing this rules page. At some point, maybe we want to collapse it all into just under Payments (ACH transfer, check, whatever). We can only do this once the payment is fully shipped.

garyhtou (approved):

garyhtou:
Let's give another test locally before merging.

PR #14470, opened 2026-08-03
[v4 API] Add scopes to OAuth consent screen

anishanne:
## Summary of the problem
Additional granular scopes like `ledgers:read` and `receipts:write` would be great to have in the V4 API.

## Describe your changes
Makes the granular v4 API scopes (`receipts:read`, `ledgers:read`, `transfers:write,` etc.) actually requestable by adding them to Doorkeeper's optional_scopes. (The controller-side enforcement already existed but no token could carry these scopes.)

The OAuth consent screen now lists the specific permissions an app is asking for when it uses `restricted` scopes, instead of the "full access to your organization funds" warning.

(Also fixes a 500 on POST /oauth/authorize_device — the device grant gem fires the after_successful_strategy_response hook with a nil response.)


config/api_scopes.rb
```diff
@@ -0,0 +1,80 @@
...
+  GRANULAR = {
+    "organizations:read" => "See the organizations you're part of",
+    "ledgers:read"       => "See your transactions and their details",
+    "transactions:write" => "Rename transactions and change their tags",
+    "receipts:read"      => "See your receipts",
+    "receipts:write"     => "Upload receipts and mark transactions as missing one",
+    "comments:read"      => "Read comments on transactions",
+    "comments:write"     => "Post comments on transactions",
+    "tags:read"          => "See your organizations' tags",
+    "tags:write"         => "Create and delete tags",
```
garyhtou: specify that these are transaction tags

Mohamad-Mortada (commented):
This is fine by me, however I'd update the v4 dev docs to mention api scopes.rb as the central directory of scopes

config/api_scopes.rb
anishanne: Done!

PR #14476, opened 2026-08-03
[Ledger::Item] Add merchant name to icon tooltip

Luke-Oldenburg:
## Summary of the problem

Card charges had custom icons but not custom tooltips.

## Describe your changes

Add merchant name when available.


app/views/ledger/_item.html.erb
```diff
@@ -1,6 +1,6 @@
 <tr class="transaction transaction--<%= item.sign %>">
   <td class="transaction__icon">
-    <div class="tooltipped tooltipped--e" aria-label="<%= item.humanized_type %>">
+    <div class="tooltipped tooltipped--e" aria-label="<%= YellowPages::Merchant.lookup(network_id: item.linked_object&.try(:merchant_network_id) || "").name.presence || item.humanized_type %>">
```
garyhtou: Can we update this so it only calls lookup on card charges?

garyhtou (approved):
Other than the comment, LGTM

YodaLightsabr:
We want to keep existing behavior here upon further examination.

PR #14481, opened 2026-08-04
[Anomaly Job] Detect linked object discrepancies

Luke-Oldenburg:
## Summary of the problem

Need to find linked object discrepancies.

## Describe your changes

Add job to do it. It's only in the schedule so that it is queueable from the Sidekiq dashboard. I will disable it.

### DIFF
--- a/app/jobs/admin/detect_linked_object_anomalies_job.rb
+++ b/app/jobs/admin/detect_linked_object_anomalies_job.rb
@@ -13,7 +13,7 @@ module Admin
             hashid: item.hashid,
             memo: item.memo,
             hcb_code_lo_type: item.hcb_code.linked_object.class.name,
-            hcb_code_lo_id: item.hcb_code.linked_object.id,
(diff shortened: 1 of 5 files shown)


app/jobs/admin/detect_linked_object_anomalies_job.rb
```diff
@@ -0,0 +1,30 @@
...
+    def perform
+      anomalous_items = []
+      Ledger::Item.where.not(linked_object_type: nil).find_each do |item|
+        if item.hcb_code.present? && item.hcb_code.linked_object != item.linked_object
+          anomalous_items << {
+            id: item.id,
+            hashid: item.hashid,
+            memo: item.memo,
+            hcb_code_lo_type: item.hcb_code.linked_object.class.name,
+            hcb_code_lo_id: item.hcb_code.linked_object.id,
```
YodaLightsabr:
```suggestion
            hcb_code_lo_id: item.hcb_code.linked_object&.id,
```

YodaLightsabr (approved):

PR #14505, opened 2026-08-05
[Contracts] Add stale contract human follow-up notification

mattsoh:
## Summary of the problem

Sometimes contracts get forgotten (maybe by the ops team or by the contract signees)

## Describe your changes

Send a reminder after 1 month to HCB Ops to see if a follow up is required.


app/tasks/maintenance/schedule_contract_human_follow_ups_task.rb
garyhtou, line 1: Good for remembering to backfill this.

app/mailers/contract/party_mailer.rb
garyhtou, line 1: I would actually separate this from the concept of a contract party. I would have this live in the contract itself. This is because contract parties are intended to be generic and work for all types of parties, including the organizer, co-signer, and HCB operations. Instead, what we're looking for here is stale contracts, and I would say that it's more of a contract-level concept rather than per party. My recommendation is that you move this to the contract mailer.

app/jobs/contract/human_follow_up_job.rb
garyhtou, line 1: Rename to `Contract::HumanFollowUpJob`

garyhtou (changes requested):

app/models/contract.rb
```diff
@@ -185,6 +185,10 @@ def agreement_name
     "agreement"
   end
 
+  def human_follow_up_email_subject
+    "[Update] #{event_name}'s #{agreement_name} still isn't signed 📝"
+  end
```
polypixeldev, line 190: If this is just being used in ContractMailer and it isn't being overriden by a subclass I would just have this be in ContractMailer

polypixeldev (commented):
Code-wise looks good, just one small thing. However is this something we want? I think this would end up being very spammy. So far this year, there are > 800 FS contracts that have been stuck in sent state for at least a month. And especially for teen applications, all of these emails would go to hcb@ since the contract is created before anyone reviews it.

mattsoh:
> Code-wise looks good, just one small thing. However is this something we want? I think this would end up being very spammy. So far this year, there are > 800 FS contracts that have been stuck in sent state for at least a month. And especially for teen applications, all of these emails would go to hcb@ since the contract is created before anyone reviews it.

Can we close out all of these applications then? An application should ideally have an archived date or something

PR #14506, opened 2026-08-05
[Contractors] Disable manual flow

manu-gurudath:
Manual flow is currently work in progress


app/controllers/payees_controller.rb
```diff
@@ -21,7 +21,7 @@ def index
   end
 
   def create
-    manual = params[:manual] == "true"
+    manual = params[:manual] == "true" && Flipper.enabled?(:manual_payees_2026_08_05, @event)
```
garyhtou: This code can be dangerous. Imagine the following situation:
1. User loads the new contractor page
2. We ship this PR (manual_payees_2026_08_05 is off for the user)
3. The user creates with the manual flow button
4. This line ignores `params[:manual]` and silently send them through the normal flow.

Instead, we should raise an error when they pass `params[:manual] == "true"`, but have `manual_payees_2026_08_05` disabled.

garyhtou:
```suggestion
    manual = params[:manual] == "true"
    if manual && !Flipper.enabled?(:manual_payees_2026_08_05, @event)
      flash[:error] = "Please try again."
      redirect_to helpers.new_recipient_transfer_path(params[:destination], @event)
    end
```

garyhtou (approved):

PR #14507, opened 2026-08-05
[Contractors] Normalize payee emails and downcase existing ones

### LINKED ISSUE #14502: [Contractors] Recipient emails are case sensitive
When adding a "New recipient" in the contractors flow, the email is treated as case sensitive. The contractor receives the invite email, but clicking the link returns a "not authorized to perform this action" error and redirects them to the homepage.

### Steps to reproduce
1. Add a new recipient in the contractors flow using an email with different casing than the one on their account (e.g. `Name@example.com` vs `name@example.com`)
2. Have the contractor open the link in the invite email
3. Observe the authorization error and redirect to homepage

### Expected behavior
Emails should be matched case insensitively, so the contractor can access the invite regardless of how the address was typed.

manu-gurudath:
Closes #14502


app/policies/payroll/position_policy.rb
```diff
@@ -52,7 +52,10 @@ def review?
     end
 
     def onboarding?
-      user&.auditor? || record.payee.legal_entity&.users&.include?(user) || user&.email == record.payee.email
+      return true if user&.auditor?
+      return true if record.payee.legal_entity&.users&.include?(user)
+
+      user&.email.present? && user.email.casecmp?(record.payee.email.to_s)
```
garyhtou:
```suggestion
      return false if user.nil?
      return true if user.auditor?
      return true if record.payee.legal_entity&.users&.include?(user)

      user.email == record.payee.email
```

Instead of using `casecmp`, we should update production data to downcase all existing emails.

It looks like in all cases, the page should not be accessible in transparency mode (when `user` is `nil`). I added it as a guard clause at the very beginning to simplify the logic in this method.

db/migrate/20260805120000_downcase_payee_emails.rb
garyhtou, line 1: This should be a maintenance task.

garyhtou (changes requested):

garyhtou (approved):

PR #14510, opened 2026-08-05
Link fiscal sponsorship agreement

anishanne:
## Summary of the problem
1. We call the HQ contract a fiscal sponsorship contract
2. When a contract is pending, we should link to it

image
image

And a link does not appear for signed out users,

### DIFF
--- a/app/models/contract/fiscal_sponsorship.rb
+++ b/app/models/contract/fiscal_sponsorship.rb
@@ -122,6 +122,13 @@ class Contract
     end
 
     def agreement_name
+      # Hack Club's own projects aren't fiscally sponsored by Hack Club, so the
+      # template they sign is a plain contract rather than a fiscal sponsorship
+      # agreement. The template is the only thing that tells the two apart —
+      # both kinds are Contract::FiscalSponsorship records.
+      hack_club_template = Event::Plan::HackClubAffiliate.new.contract_docuseal_template_id
+      return "contract" if external_template_id.to_s == hack_club_template.to_s
+
       "fiscal sponsorship agreement"
     end
 
--- a/app/models/contract/fiscal_sponsorship.rb
+++ b/app/models/contract/fiscal_sponsorship.rb
@@ -121,12 +121,18 @@ class Contract
       payload
     end
 
+    # Hack Club's own projects aren't fiscally sponsored by Hack Club, so they
+    # sign a plain contract rather than a fiscal sponsorship agreement. Both are
+    # Contract::FiscalSponsorship records, so the DocuSeal template is the only
+    # thing that tells them apart. It also stays accurate for contracts that were
+    # already issued when the organization's plan changes.
+    #
+    # Once contracts belong to a Contract::Template record, that template should
+    # carry its own name and this whole comparison can go away.
     def agreement_name
-      # Hack Club's own projects aren't fiscally sponsored by Hack Club, so the
-      # template they sign is a plain contract rather than a fiscal sponsorship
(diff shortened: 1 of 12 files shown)


app/models/event/plan/standard.rb
garyhtou, line 1: I think we likely don't want to pollute the event plan with this information. Instead, it should be configured within the contract model.

I think we already have `Contract#agreement_name` defined. Let's use it.

app/models/event.rb
```diff
@@ -1036,6 +1036,16 @@ def contracts_pending_on_hcb
...
 
+  # The party on one of this organization's sent contracts that a user is
+  # allowed to open — their own, or HCB's if they're an admin.
+  def contract_party_for(user)
+    return if user.blank?
+
+    parties = Contract::Party.where(contract: contracts.sent)
+
+    parties.not_hcb.find_by(user:) || (parties.hcb.first if user.admin?)
+  end
```
garyhtou: I don't think we want to pollute the event model with this. This might be better as a class method in `Contract::Party`. Maybe

```ruby
class Contract::Party
  def self.for(contractable:, user:)
    return if contractable.nil? || user.nil?

    parties = self.includes(:contract).where(contract: {contractable:})
    if user == :hcb
      parties.hcb.first
    else
      parties.not_hcb.find_by(user:)
    end
  end
end

# Then, in the view, we call
Contract::Party.for(contractable: @event, user: current_user)
```
_i haven't test this; this is psuedo code_

thoughts @polypixeldev ?

garyhtou (changes requested):

app/models/event/plan/standard.rb
anishanne: Should we do this by updating `app/models/contract/fiscal_sponsorship.rb` to check if the template id matches `Event::Plan::HackClubAffiliate.new.contract_docuseal_template_id`?

app/models/event.rb
anishanne: Moved

app/models/contract/fiscal_sponsorship.rb
```diff
@@ -122,6 +122,13 @@ def payload
...
     def agreement_name
+      # Hack Club's own projects aren't fiscally sponsored by Hack Club, so the
+      # template they sign is a plain contract rather than a fiscal sponsorship
+      # agreement. The template is the only thing that tells the two apart —
+      # both kinds are Contract::FiscalSponsorship records.
+      hack_club_template = Event::Plan::HackClubAffiliate.new.contract_docuseal_template_id
+      return "contract" if external_template_id.to_s == hack_club_template.to_s
+
       "fiscal sponsorship agreement"
     end
```
garyhtou, line 139: In the future, this logic will be handled by `Contract::Template` (a model that hasn't been created yet). cc @polypixeldev

app/models/contract/party.rb
```diff
@@ -58,6 +58,22 @@ class Party < ApplicationRecord
...
+    # contracts: their own, or HCB's if they're an admin.
+    #
+    # This mirrors Contract::PartyPolicy#show? so that we only ever surface a
+    # link the user can actually follow. It is not itself an authorization
+    # check; Contract::PartiesController#show still authorizes the party.
+    def self.for(event:, user:)
+      return if event.nil? || user.nil?
+
+      # find_by does not order on its own, and a user can hold a party on more
+      # than one of an organization's open contracts, so order it ourselves.
```
garyhtou:
```suggestion
```

garyhtou (approved):

PR #14509, opened 2026-08-05
Add parent id for suborgs

anishanne:
## Summary of the problem
The sub-org CSV download didn't include sub-orgs of sub-orgs. Some users want to create a full descendant tree using the sub-organizations CSV export. That is not possible at the moment because we only export immediate children and not all descendants. 

## Describe your changes
Updating the CSV export to include all dissents and are adding a parent_id column.

Example:
[Hack Club HQ (Suborg Demo)'s sub-organizations.csv](https://github.com/user-attachments/files/30759586/Hack.Club.HQ.Suborg.Demo.s.sub-organizations.csv)

PR #14511, opened 2026-08-05
Make G Suite factories generate unique addresses and domains

anishanne:
## Summary of the problem
Fixes CI flake experienced in #14509.


garyhtou (changes requested):
Instead of using sequence, can we use the `unique` helper provided by Faker?

https://github.com/faker-ruby/faker#ensuring-unique-values

garyhtou (approved):

YodaLightsabr (approved):

PR #14519, opened 2026-08-06
[Team] Make the parent org access callout collapsible

anishanne:
## Summary of the problem
The "The team behind X also has access to Y" callout takes up a lot of space on the team page.

## Describe your changes
Makes the callout minimized by default.

image
image


app/views/events/team.html.erb
```diff
@@ -33,21 +33,33 @@
...
-          </span>
-        <% end %>
-      </div>
-    <% end %>
-  <% end %>
+      <% if @indirect_access.present? %>
+        <p>These team members can access this organization:</p>
+
+        <div class="grid">
+          <% @indirect_access.each do |user, role| %>
```
garyhtou, line 52: Update `@indirect_access` to omit users if they have an Organizer Position on this organization.

This solves a confusing case:
- Bob is a reader on parent org
- Bob has a member organizer position on child org

This copy would claim that Bob can only view, which is incorrect.

anishanne: Fixed - We now omit only if child ≥ parent

app/views/events/team.html.erb
```diff
@@ -33,21 +33,31 @@
...
+        <span class="info line-height-0"><%= inline_icon "leader", size: 24 %></span>
+        <p class="bold m0 flex-auto">The team behind <%= @event.parent.name %> also has access to <%= @event.name %></p>
+        <span class="muted line-height-0"><%= inline_icon "down-caret", size: 24, class: "flip-when-open" %></span>
+      </div>
+    </summary>
 
-    <% if @indirect_access.present? %>
-      <p>These team members can access this organization:</p>
+    <div class="border-top b--muted px-4 py-3">
+      <p class="mt0">Because <%= @event.name %> is a sub-organization of <%= @event.parent.name %>, the <%= @event.parent.name %> team has access here too. Their managers can manage this organization, while their members and readers can only view it.</p>
```
anishanne:
```suggestion
      <p class="mt0">Because <%= @event.name %> is a sub-organization of <%= @event.parent.name %>, the <%= @event.parent.name %> team has access here too.</p>
```

garyhtou (approved):

PR #14521, opened 2026-08-06
[Ledger::Item] Create pins

### LINKED ISSUE #14265: [Ledger::Item] Migrate pins from HcbCode
- add pinned_at to ledger mapping and create a concern that stores the pin logic
- migrate views to using ledger mapping
- backfill pinned_at on mappings in prod.

Luke-Oldenburg:
## Summary of the problem

closes #14265
The new ledger needs support for pinning transactions

## Describe your changes

Copy over hcb code pin logic to a new model and forward method calls through hcb code to ledger item.


app/models/ledger/item.rb
garyhtou, line 1: Unrelated thought: when a Ledger item is remapped, we should probably delete any pins.

app/models/event.rb
```diff
@@ -459,6 +459,7 @@ def ancestor_users
   after_create :create_ledger
   has_many :hcb_codes
   has_many :pinned_hcb_codes, -> { includes(hcb_code: [:canonical_transactions, :canonical_pending_transactions]) }, class_name: "HcbCode::Pin"
+  has_many :pinned_ledger_items, through: :ledger, source: :item_pins
```
garyhtou: If no one is calling `pinned_hcb_codes` anymore, we should maybe remove it in this PR

garyhtou (changes requested):
Wait!!! Okay, there actually might be a better database design for pinned transactions compared to what we did for HcbCode::Pin

New design:

- We add a `pinned_at` timestamp to `Ledger::Mapping`
- Assert that only primary mappings can be pinned.
- I think this makes the database query more performant because then you do not need to join in another table.
- This makes it easier to pin transactions when they are remapped because everything happens in one record.

Let me know what you think of this new design, and let's chat through it.

Luke-Oldenburg:
> This makes it easier to pin transactions when they are remapped because everything happens in one record.

When we remap ledger items, do we make a new mapping or reuse the existing one? We need to ensure that we are properly disposing of `pinned_at`.

app/models/ledger/item.rb
Luke-Oldenburg: Agree

Luke-Oldenburg:
**potential concern:** if we add `pinned_at` to `Ledger::Mapping` this likely means we still need to fetch the primary ledger to retrieve the list of mappings if we use non-primary ledgers to cache search queries in the future. I think this may still result in less database queries than using a separate `Ledger::Item::Pin` model that would also need to be fetched by the primary ledger.

Luke-Oldenburg:
Another question is how would we handle sort index if we stored pins on ledger mappings if we want to allow users to rearrange their pins?

garyhtou:
> > This makes it easier to pin transactions when they are remapped because everything happens in one record.
> 
> When we remap ledger items, do we make a new mapping or reuse the existing one? We need to ensure that we are properly disposing of `pinned_at`.

When we remap a ledger item, we want to reuse the old mapping so that there's a clear paper trail of how that transaction's mapping has changed

garyhtou:
We decided to move forward with adding a pinned_at column to the Ledger item mapping. We will not support sorting/ordering pins at the moment.

Ensure there's an index on ledger_id and pinned_at (do research on which ordering is best)

app/models/event.rb
Luke-Oldenburg: I think we should keep it for now.

app/models/concerns/pinnable.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+# Lets a model be "pinned" to highlight it (e.g. at the top of a transaction
+# list), capped at four pins per event.
+#
+# The including model must have a `pinned_at :datetime` column and implement:
+#   #event       - the event a pin is scoped to, used to enforce the per-event cap
+#   #ledger_item - the underlying item whose `pinnable?` gates whether pinning is allowed
```
Luke-Oldenburg:
```suggestion
```

app/models/ledger/mapping.rb
```diff
@@ -33,17 +35,26 @@ class Mapping < ApplicationRecord
...
     has_paper_trail
 
+    include Pinnable
+
     belongs_to :ledger, class_name: "::Ledger"
     belongs_to :ledger_item, class_name: "Ledger::Item"
 
     belongs_to :mapped_by, class_name: "User", optional: true
 
+    delegate :event, to: :ledger
```
Luke-Oldenburg, line 45:
```suggestion
    delegate :event, to: :ledger
    delegate :card_grant, to: :ledger
```

app/models/event.rb
```diff
@@ -459,6 +459,7 @@ def ancestor_users
   after_create :create_ledger
   has_many :hcb_codes
   has_many :pinned_hcb_codes, -> { includes(hcb_code: [:canonical_transactions, :canonical_pending_transactions]) }, class_name: "HcbCode::Pin"
+  has_many :pinned_ledger_items, -> { includes(hcb_code: [:canonical_transactions, :canonical_pending_transactions]) }, through: :ledger, source: :pinned_items, class_name: "Ledger::Item"
```
```suggestion
  has_many :pinned_ledger_items, through: :ledger, source: :pinned_items, class_name: "Ledger::Item"
```

app/controllers/ledger/items_controller.rb
```diff
@@ -32,6 +32,35 @@ def hcb
       redirect_to hcb_code_path(@item.hcb_code)
     end
 
+    def pin
+      @item = Ledger::Item.find_by_hashid!(params[:item_id])
```
polypixeldev: it'd be nice if setting `@item` could be in a before action but that can be in a later PR

polypixeldev (approved):
nice!

PR #14523, opened 2026-08-06
[V4] Add endpoint for editing comments

mattsoh:
## Summary of the problem

Add ability to edit comments on V4 Api. This helps for Steelyard, and maybe HCB mobile?

## Describe your changes

Added endpoint (content and admin_only).


app/controllers/api/v4/comments_controller.rb
```diff
@@ -31,6 +31,16 @@ def create
 
       require_oauth2_scope "comments:write", :create
 
+      def update
+        @comment = authorize Comment.find_by_public_id!(params[:id])
+
+        @comment.update!(params.permit(:content, :admin_only))
```
garyhtou: Can you please write tests for this?
- When `admin_only` is omitted, it doesn't change `admin_only` in the DB
	- make sure to the case where it start as `admin_only = true`
- When `admin_only` is a non-boolean (e.g. string or integer), what do we do?

mattsoh: When `admin_only` is a non-boolean, the values are coerced (when running `@comment.update!`). This is the same logic as in creating comments (except `admin_only` is by default set to `false` when creating comments - if no `admin_only` value is provided in the API call we don't change anything).

mattsoh:
I added tests, they pass!

I also gatekeeped admin only behind auditor access.

app/controllers/api/v4/comments_controller.rb
```diff
@@ -31,6 +31,20 @@ def create
 
       require_oauth2_scope "comments:write", :create
 
+      def update
+        @comment = authorize Comment.find_by_public_id!(params[:id])
+
+        @comment.assign_attributes(params.permit(:content, :admin_only))
```
polypixeldev: You should be able to also update the comment file

app/controllers/api/v4/comments_controller.rb
```diff
@@ -31,6 +31,20 @@ def create
 
       require_oauth2_scope "comments:write", :create
 
+      def update
+        @comment = authorize Comment.find_by_public_id!(params[:id])
```
Luke-Oldenburg:
```suggestion
        @comment = Comment.find_by_public_id!(params[:id])
        authorize @comment
```

app/controllers/api/v4/comments_controller.rb
mattsoh: Added!

PR #14527, opened 2026-08-06
[Ledger] Hide author column on card views

Luke-Oldenburg:
## Summary of the problem

There is only ever one author in these views so we don't need to show it.

## Describe your changes

Hide the author column. This PR does not remove the preload from `Ledger::Query`


garyhtou (approved):
🪄 it passes the vibe checks

PR #14538, opened 2026-08-07
[PayoutHoldingService] Fix transfers not being assigned on error

Luke-Oldenburg:
## Summary of the problem

When a transfer errors in sending, we do not save it to the payout holding.
https://hcb.hackclub.com/blazer/queries/1301-payout-holdings-with-no-transfer-attached

## Describe your changes

Save the transfer to the payout holding.

### DIFF
--- a/app/services/reimbursement/payout_holding_service/nightly.rb
+++ b/app/services/reimbursement/payout_holding_service/nightly.rb
@@ -30,6 +30,8 @@ module Reimbursement
                   wire.send_wire!
                 rescue
                   wire.mark_rejected!
+                  payout_holding.wire = wire
+                  payout_holding.save!
                   payout_holding.mark_failed!
                   reason = "There was an error creating the wire transfer."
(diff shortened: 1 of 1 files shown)


garyhtou (approved):
🪄

app/services/reimbursement/payout_holding_service/nightly.rb
```diff
@@ -30,6 +30,8 @@ def run
                   wire.send_wire!
                 rescue
                   wire.mark_rejected!
+                  payout_holding.wire = wire
+                  payout_holding.save!
```
garyhtou, line 34: I just double-checked there's no code that assumes that a wire is successful if it's not nil here. In other words, double check that no code assumes `payout_holding.wire.present? == it was successful`

Luke-Oldenburg: We appear to be safe here.

PR #14540, opened 2026-08-07
[Sub-org] Added Excel export with collapsible tree

### LINKED ISSUE #14533: [Sub-org] Add a new Excel export with collapsible suborg tree
At the moment, we provide you with a CSV export, which you can use to manually rebuild the sub-organization tree. We will keep this export accessible.

However, we'd like to add a new export alongside it, which is Excel-based. This Excel sheet will render the sub-org tree in a collapsible manner. 

See https://github.com/hackclub/hcb/blob/139e65109f8c3a04f1ea1e5fae20201b8f3d6ed7/app/models/event/statement_of_activity.rb#L97 as an example of implementation

anishanne:
## Summary of the problem
Fixes #14533 

## Describe your changes
- Converts download button to a dropdown
- Adds "Download tree as Excel button

image

Example output:
[Robotics League (Parent Org)'s sub-organizations (1).xlsx](https://github.com/user-attachments/files/30840817/Robotics.League.Parent.Org.s.sub-organizations.1.xlsx)


app/models/event/sub_organizations_export.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+class Event
+  class SubOrganizationsExport
+    # Excel supports at most 7 levels of row grouping. Deeper descendants are
+    # still exported, they just share the deepest collapsible level.
+    MAX_OUTLINE_LEVEL = 7
```
garyhtou, line 7: In the case that we exceed these seven levels, we should add a comment in the file itself.

app/models/event/sub_organizations_export.rb
```diff
@@ -0,0 +1,74 @@
...
+      worksheet = workbook.add_worksheet("Sub-organizations")
+      # Attach each group's collapse toggle to the parent row above it rather
+      # than the default summary row below.
+      worksheet.outline_settings(1, 0, 0, 0)
+      worksheet.set_column("A:A", 20)
+      worksheet.set_column("B:C", 40)
+      worksheet.set_column("D:D", 12)
+      worksheet.set_column("E:E", 30)
+
+      worksheet.write_row(0, 0, ["ID", "Name", "Slug", "Balance", "Tags"], bold)
```
garyhtou, line 29: Let's add a `Parent ID` and `Parent Name`

garyhtou (approved):
I'm going to merge this as-is, you can come back and add the remaining feedback

PR #14547, opened 2026-08-08
[Application] Clarify application flow and submission status

mattsoh:
## Summary of the problem

A lot of people (including people who used the old application flow) don't clearly see that they need to sign the fiscal sponsorship agreement before we review their application (we used to review then send agreement). This means they've filled in the application but it doesn't get to us.

## Describe your changes

- Clearly indicate that the review page is not the end of the submission and that there is still the agreement to sign 
- Split submitted page into "sign agreement" and "await review", so that it is clear that the application is not complete yet without fiscal sponsorship agreement.

https://github.com/user-attachments/assets/19f920e1-8133-4810-b072-b31cc0488e09


mattsoh:
@claude review

PR #14550, opened 2026-08-08
[Blazer] Configure email notifications to go to hcb-engr@hackclub.com

### LINKED ISSUE #14529: [Blazer] Configure email notifications to go to hcb-engr@
At the moment, email notifications get sent to Intercom. It has latency bc we gotta wait for ops to assign the ticket to us.

Let's reconfigure Blazer so emails go to hcb-engr@ by default

mattsoh:
## Summary of the problem

closes https://github.com/hackclub/hcb/issues/14529

## Describe your changes

- Add `ENGINEERING_EMAIL` alongside `OPERATIONS_EMAIL` to `ApplicationMailer`
- Change `from_email` in blazer to `hcb-engr@` so bounces return to engr
- Set default email on new checks to `hcb-engr@`
- Rewrite email in already-configured checks from `hcb@` to `hcb-engr@` (if applicable)


db/migrate/20260808175000_point_blazer_checks_at_engineering.rb
Luke-Oldenburg, line 1: All of the current checks go to `hcb-engr@hackclub.com` so this is safe to drop. Before, it was a random set of emails, not necessarily just to `hcb@hackclub.com`. For future reference, a maintenance task works better here than a DB migration because we are not updating the schema of the DB.

config/blazer.yml
```diff
@@ -79,7 +79,8 @@ user_method: blazer_current_user
 # before_action_method: require_admin
 
 # email to send checks from
-from_email: hcb@hackclub.com
+# Bounces and auto-replies land back here, so keep it off Intercom.
+from_email: hcb-engr@hackclub.com
```
Luke-Oldenburg, line 83: I think we should keep this as `hcb@hackclub.com`. All of our other emails are sent from that and these emails are going to be getting sent to `hcb-engr@hackclub.com`.

config/initializers/blazer.rb
Luke-Oldenburg, line 1: lgtm

Luke-Oldenburg (changes requested):

PR #14558, opened 2026-08-10
[Card locking] Roll out card locking to everyone

manu-gurudath:
Rolls out card locking to all users


garyhtou:
Ended up going with https://github.com/hackclub/hcb/pull/14597. ty though

PR #14563, opened 2026-08-10
[Concern] Delete app/models/concerns/has_ledger_item.rb

Luke-Oldenburg:
## Summary of the problem

This is no longer used.

## Describe your changes

Remove the file


YodaLightsabr (commented):
This is concerning

PR #14566, opened 2026-08-11
[CT / CPT] Remove update custom memo

Luke-Oldenburg:
## Summary of the problem

Updating a CT or CPT's custom memo directly causes it to end up out of sync from its ledger item's custom memo.

## Describe your changes

Remove the ability to update the custom memo of an individual CT or CPT.


app/controllers/canonical_pending_transactions_controller.rb
```diff
@@ -67,9 +66,9 @@ def set_category
 
   def canonical_pending_transaction_params
     if admin_signed_in?
-      params.require(:canonical_pending_transaction).permit(:custom_memo, :fronted, :fee_waived)
+      params.fetch(:canonical_pending_transaction, {}).permit(:fronted, :fee_waived)
     else
-      params.require(:canonical_pending_transaction).permit(:custom_memo)
+      params.fetch(:canonical_pending_transaction, {}).permit
```
garyhtou, line 63: Does it make sense to entirely remove this branch? It seems like with this change, we'd no longer allow params for non-admins

app/views/hcb_codes/_transaction_history.html.erb
```diff
@@ -18,11 +17,6 @@
             <td><%= format_date ct.date %></td>
             <td><%= ct.smart_memo %></td>
             <td style="text-align: right;"><%= number_to_currency(ct.amount) %></td>
-            <td>
-              <%= pop_icon_to "edit",
-                  edit_canonical_transaction_path(ct),
-                  class: "right tooltipped tooltipped--w", "aria-label": "Rename transaction" if organizer_signed_in?(as: :member) %>
-            </td>
```
garyhtou, line 25: Without this link, is there still a way for admins to waive the fee for a CT?

garyhtou (commented):
🪄

app/controllers/canonical_pending_transactions_controller.rb
Luke-Oldenburg: Yes, we still need this — the else branch isn't just handling the removed custom_memo case, it's what keeps this safe from a privilege-escalation standpoint.

update? in CanonicalPendingTransactionPolicy allows admin_or_teammember, not just admins, so non-admin team members can still hit this update action. fronted/fee_waived are financial flags (fee waiving, fronted-amount calcs) that are only ever surfaced in the edit form inside admin_tool blocks — they're meant to be admin-only. If we dropped the if/else and just permitted :fronted, :fee_waived unconditionally, a team member could craft a request setting those fields themselves.

Separately, if we kept the if but removed the else (so the method returns nil for non-admins), update!(canonical_pending_transaction_params) would raise ArgumentError: When assigning attributes, you must pass a hash as an argument the next time a non-admin hits update — so we need some non-nil fallback regardless.

It's true that with custom_memo gone, non-admins no longer have anything to legitimately update, so the branch is effectively a no-op guard now — but that's exactly its job: strip fronted/fee_waived for anyone who isn't an admin.

~ Claude

garyhtou: In the future, let's come back to:
1. Update CanonicalPendingTransactionPolicy to remove team member access
2. Remove this line

garyhtou (approved):

PR #14567, opened 2026-08-11
[CardLockingMailer] Use `attach_receipt_url`

Luke-Oldenburg:
## Summary of the problem

Card locking emails didn't use signed URLs for uploading receipts.

## Describe your changes

Use helper that provides signed URLs.


garyhtou (approved):

Luke-Oldenburg:

PR #14574, opened 2026-08-11
[AttemptMailer] Fix creator name error

Luke-Oldenburg:
## Summary of the problem

`@creator` is a String

## Describe your changes

Use `@payment.creator`


polypixeldev (approved):

Luke-Oldenburg:

PR #14576, opened 2026-08-11
[User] Fix undefined method for nil error in `locked_by`

Luke-Oldenburg:
## Summary of the problem

```
ActionView::Template::Error: undefined method 'whodunnit' for nil
app/models/user.rb:396 in User#locked_by
app/views/users/edit_admin.html.erb:14
app/helpers/users_helper.rb:283 in UsersHelper#admin_tool
app/views/users/edit_admin.html.erb:7
app/middleware/set_current_request_ip.rb:11 in SetCurrentRequestIp#call
```

## Describe your changes

Safe nav

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -401,7 +401,9 @@ class User < ApplicationRecord
   end
 
   def locked_by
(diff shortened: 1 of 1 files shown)


Luke-Oldenburg:

PR #14578, opened 2026-08-11
[Team] Allow searching by preferred name

garyhtou:
## Summary of the problem

It would be useful to search team members by their preferred names

## Describe your changes

Support this!


Luke-Oldenburg (approved):
lgtm!

Issue #14565, opened 2026-08-10
[Card Locking] Add explicit grace period and temp disable trusted user feature


anishanne:
## Problem
1. All users are confused because we use the wording "may lock" instead of "will lock"
2. Trusted users are confused because their cards do not lock on the "may lock" date, and a new date is not provided. So, it seems like HCB is broken.

## Solution
1. We will change all wording from "may lock" to "will lock"
2. On the receipt due date (the "will lock" date), trusted users will receive an email stating they have been granted an extension till XYZ date.

This way, an extension can be viewed as an intentional gift/happy surprise instead of a potential bug.

**For now, we will be disabling the trusted user feature temporarily, while we work on these updates.**

(Chatted offline with Gary/Mel.)

PR #14580, opened 2026-08-12
Improve card locking wording

anishanne:
## Summary of the problem
Users are confused whether cards _will_ or _may_ lock. See #14565 

## Describe your changes
Changes all wording to indicate cards _will_ lock after 7 days of no receipts. Works in conjunction with #14568.

Wording is changed to:
1) Emphasize that cards will lock.
2) Remind users that cards will lock again.

### DIFF
--- a/app/services/user_service/update_card_locking.rb
+++ b/app/services/user_service/update_card_locking.rb
@@ -70,7 +70,7 @@ module UserService
 
     def notify_unlocked
       CardLockingMailer.cards_unlocked(user: @user).deliver_later
-      send_sms("Your HCB cards work again. Keep uploading receipts within 7 days of the charge. Manage them at #{CardLocking.inbox_url}.")
+      send_sms("Your HCB cards work again. Keep uploading receipts within 7 days of the charge, or your cards will lock again. Manage them at #{CardLocking.inbox_url}.")
     end
 
     def locked_message(now:)
--- a/app/services/user_service/update_card_locking.rb
+++ b/app/services/user_service/update_card_locking.rb
@@ -70,7 +70,7 @@ module UserService
 
     def notify_unlocked
       CardLockingMailer.cards_unlocked(user: @user).deliver_later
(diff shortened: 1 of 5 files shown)


app/services/user_service/update_card_locking.rb
```diff
@@ -70,7 +70,7 @@ def notify_locked(now:)
 
     def notify_unlocked
       CardLockingMailer.cards_unlocked(user: @user).deliver_later
-      send_sms("Your HCB cards work again. Keep uploading receipts within 7 days of the charge. Manage them at #{CardLocking.inbox_url}.")
+      send_sms("Your HCB cards work again. Keep uploading receipts within 7 days of the charge, or your cards will lock again. Manage them at #{CardLocking.inbox_url}.")
```
garyhtou: Make it clear what "them" refers to now that we mention both receipts and cards in this sentence.

```suggestion
      send_sms("Your HCB cards work again. Keep uploading receipts within 7 days of the charge, or your cards will lock again. Manage receipts at #{CardLocking.inbox_url}.")
```

app/views/card_locking_mailer/cards_locked.html.erb
```diff
@@ -2,7 +2,7 @@
   Your HCB cards are locked because <%= @count %> <%= "receipt".pluralize(@count) %> <%= @count == 1 ? "is" : "are" %> overdue. Upload <%= @count == 1 ? "it" : "them" %> and your cards work again in seconds.
 </p>
 <p>
-  Every charge needs a receipt within 7 days. Manage your receipts on the HCB <%= link_to "receipts page", my_inbox_url %>.
+  Every charge needs a receipt within 7 days, or your cards will lock again. Manage your receipts on the HCB <%= link_to "receipts page", my_inbox_url %>.
```
garyhtou: Clarify copy when reading from the perspective of "why are my cards not working right now?"

```suggestion
  Every charge needs a receipt within 7 days. Overdue receipts result in your cards being locked. Manage your receipts on the HCB <%= link_to "receipts page", my_inbox_url %>.
```

garyhtou (approved):

PR #14589, opened 2026-08-12
[Payments] Only require tax form if tax reportable

polypixeldev:
## Summary of the problem

Payments are only required to be included on a 1099 if they are for services, and not for a corporation (as long as it's not for attorney or medical services). We shouldn't request W8/W9s for these.

## Describe your changes

Adds a classification to payment and corporation as a LE type. If tax form not required, we don't request a W8/W9. When HCB reviews the final transfer, they can update the payment classification, and if the change makes the payment tax reportable and we don't already have a W8/W9 on file, we'll reject the attempt and have them submit it.

image


app/models/payment.rb
```diff
@@ -116,15 +128,20 @@ def estimate_usd_amount_cents
     MoneyService.convert_to_usd(amount_cents, currency)
   end
 
+  # Payments that aren't tax reportable don't need to wait on the payee's tax paperwork.
+  def requires_tax_form?
+    !(payee.legal_entity&.corporation? && !for_attorney_or_medical_services?) && !for_goods?
+  end
```
garyhtou: Should we turn this into a stored boolean?

app/controllers/increase_checks_controller.rb
```diff
@@ -46,9 +46,18 @@ def approve
...
+    @check.payment&.update!(classification: params[:classification])
+
+    if @check.payment.nil? || @check.payment.legal_entity.payable?(requires_tax_form: @check.payment.requires_tax_form?)
+      @check.send_check!
+
+      redirect_to increase_check_process_admin_path(@check), flash: { success: "Check has been sent!" }
+    else
+      @check.payment.request_tax_form!
+
+      redirect_to increase_check_process_admin_path(@check), flash: { error: "Tax information was missing for this payment and has been requested" }
```
garyhtou: This this logic into a helper method similar to `ensure_admin_may_approve`

app/views/legal_entities/show.html.erb
```diff
@@ -50,125 +39,137 @@
...
+  <ul>
+    <% if entity_type_matches %>
       <li>
-        <strong>Discard last tax form</strong>:
-        Pick this if you made a mistake on your last tax form.
-        You'll then be able to submit a corrected tax form.
+        <strong>Switch to new TIN</strong>:
+        Pick this if your TIN for <%= @legal_entity.name %> changed to <%= masked_tin_tag mismatched, viewable: show_tin %>.
+        You will no longer be able to receive payments through HCB to your old TIN
+        (<%= masked_tin_tag previous_tax_form, viewable: show_tin %>).
```
manuthecoder, line 86: `previous_tax_form` was deleted

app/controllers/tax/forms_controller.rb
```diff
@@ -31,6 +31,12 @@ def create
         return
       end
 
+      unless @legal_entity.tax_form_required? || @legal_entity.completed_tax_form?
```
garyhtou, line 37: Let's leave a comment explaining why `@legal_entity.completed_tax_form?` is necessary

app/models/payment/attempt.rb
```diff
@@ -88,7 +88,7 @@ class Attempt < ApplicationRecord
       event :mark_rejected do
         transitions from: :under_review, to: :rejected
         after do
-          payment.mark_rejected!
+          payment.mark_rejected! if payment.may_mark_rejected?
         end
       end
```
garyhtou, line 103:
```suggestion
      event :mark_rejected do
        transitions from: :under_review, to: :rejected
        after do
          payment.mark_rejected!
        end
      end

      event :mark_rejected_retryable do
        Only reject the attempt, and leave the payment in a non-rejected, retryable state 
        transitions from: :under_review, to: :rejected
        after do
          payment.mark_pending_legal_entity!
        end
      end
```

app/controllers/concerns/admin/payment_approvable.rb
```diff
@@ -0,0 +1,39 @@
...
+          payment.update!(classification:) if classification.present?
+
+          error = nil
+
+          if payment.legal_entity.tin_banned?
+            error = "This legal entity's TIN has been banned"
+          elsif payment.legal_entity.archived?
+            error = "This legal entity has been archived"
+          elsif !payment.legal_entity.payable?(requires_tax_form: payment.requires_tax_form)
+            payment.mark_pending_legal_entity!
```
garyhtou:
```suggestion
```

garyhtou (approved):
🪄

PR #14591, opened 2026-08-12
[UI] Fix user session grids to overflow when needed

mattsoh:
## Summary of the problem

Text did not overflow when width was too low / text length too high (e.g. ipv6 addresses, or also sacrilegiously zooming in.

## Describe your changes

Add overflow to next line (hide separator if so), also allow ip address / session to overflow as well if needed.

Screenshot 2026-08-12 at 10 26 23 PM
Screenshot 2026-08-12 at 10 26 38 PM


manu-gurudath (approved):
lgtm!

PR #14603, opened 2026-08-13
[User] Add a configurable timezone setting

garyhtou:
## Summary of the problem

HCB stores no timezone preference for users, so anything that has to render a time outside the browser (emails, most notably) has no way to know which timezone the recipient actually reads times in. The best we can do today is guess.

## Describe your changes

Adds a nullable `timezone` column to `users` and a picker for it in account settings, alongside the guess rather than instead of it.

- **`User#resolved_timezone`** is the single method the rest of the app should call. Precedence: the configured timezone, then a timezone inferred from what the browser reported at sign-in across recent sessions, then `User::DEFAULT_TIMEZONE`.
- **`User#inferred_timezone`** takes the mode over recent sessions (not the latest) so a trip does not repoint someone for a fortnight after they get home, and skips values `ActiveSupport::TimeZone` cannot resolve. It returns `nil` when nothing usable was reported, so callers can tell a real guess apart from the default.
- **The column is nullable on purpose.** NULL means "never chosen" and has to stay distinguishable from someone explicitly choosing the default, which is also why the picker has a "Detect automatically" blank option and why `Maintenance::BackfillUserTimezonesTask` only writes a value for users we can actually infer one for.
- `ActiveSupport::TimeZone` resolves both IANA identifiers and Rails' friendlier names, but the picker only offers the latter. A `normalizes` translates to the picker's form on write, so a saved choice is always one the select can show back to the user. A validation rejects anything else.
- Copy under the picker notes that the website always renders in the browser's current timezone while emails use this setting.

Specs cover the resolution precedence, the inference rules, normalization and validation, the controller params path, and the backfill task.

### Follow-ups

- Nothing reads `resolved_timezone` yet. Moving existing mailers, views, and reports onto it is deliberately left to a separate PR so this one stays reviewable.
- No sweep of the codebase for timezone correctness was done here.

### Conflict note

#14601 (`card-locking-suppression-copy`) adds `assumed_timezone`, `DEFAULT_TIMEZONE`, and `TIMEZONE_SESSION_SAMPLE` to `app/models/user.rb` in the same spot, so whichever merges second will conflict there. **This branch's side should win.** It keeps `assumed_timezone` behaviourally identical, just splitting the session lookup out into `inferred_timezone` so the "no usable guess" case is expressible, and the constants are the same on both sides.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -539,6 +539,7 @@ class UsersController < ApplicationController
       :birthday,
       :profile_picture,
       :seasonal_themes_enabled,
+      :timezone,
       # admin
       :pretend_is_not_admin,
       # security
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -30,6 +30,7 @@
 #  slug                          :string
 #  subscribed_to_loops_at        :datetime
 #  teenager                      :boolean
+#  timezone                      :string
 #  use_sms_auth                  :boolean          default(FALSE)
 #  use_two_factor_authentication :boolean          default(FALSE)
 #  verified                      :boolean          default(FALSE), not null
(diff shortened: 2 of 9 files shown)


garyhtou:
This needs a lot of fixing up

PR #14609, opened 2026-08-13
[Ledger] Use new ledger on user admin stripe transactions

Luke-Oldenburg:
## Summary of the problem

Using old ledger here.

## Describe your changes

Use new ledger. We aren't ready to switch admin missing receipts yet due to a divergence in how they are calculated.


app/views/ledgers/_ledger.html.erb
```diff
@@ -1,6 +1,6 @@
 <%# locals: (ledger: nil, items:, table_only: true, filter: nil, show_fee: false, show_author: true) %>
 
-<%= turbo_frame_tag dom_id(ledger) do %>
+<%= turbo_frame_tag(ledger || "ledger") do %>
```
YodaLightsabr: Is the removal of `dom_id` here intentional?

YodaLightsabr (approved):

PR #14611, opened 2026-08-14
[V4] Add `admin:comments` scopes

mattsoh:
## Summary of the problem

Idk I would very much like to have these scopes (for Steelyard especially).

## Describe your changes

Implement new scopes for admin comments (`admin:comments:read` and `admin:comments:write`. These exist separately from the `restricted` scope (like how the `admin:read` and `admin:write` scopes exist seperately as well)

- Add an optional `resource:` kwarg  to `ApiAdminContext`
- Collapse inline checks for admin scopes into single private `admin_scope?(level)` helper, which accounts for both `admin:<level>` and `admin:<resource>:<level>` 
- Overwrite `pundit_user` to build context with `resource: :comments` (to allow for specific resource scopes)
- List new scopes in `optional_scopes` in Doorkeeper

Oh, also would the _amazing_ Engr Reviewer also be able to take a look at https://github.com/hackclub/hcb/pull/14523 please (adding endpoint to edit comments), thank you~!


polypixeldev (commented):
I'll leave this up to Mohamad as he's more familiar with the API and how scopes should work, but I'd be careful about this approach. This would let a token with only admin:comments:write act as an admin for any action originating from CommentsController, not necessarily only actions that read or write to comments.

For example, if I understand correctly, this would let a admin:comments:write token see user PII through the expand in the comment index, even though it seems the token shouldn't allow that.

garyhtou:
@Mohamad-Mortada, after u review, lmk and i'll take a look!

mattsoh:
Superseded by #14916

PR #14617, opened 2026-08-14
[OTJ] Prevent nuke and backfill ledger in production

Luke-Oldenburg:
## Summary of the problem

These jobs are still useful to have in development but dangerous in production.

## Describe your changes

Prevent running them in production


polypixeldev (approved):
I think it would be better to also log and report if these are ran in production, but this works fine

PR #14625, opened 2026-08-15
[Card Locking] Tell cardholders when their next receipt is due

anishanne:
## Summary of the problem
Users are confused about when cards will lock.

image

image

## Describe your changes
Updates wording in SMS and on my inbox page to count down to when the next receipt is due.

### DIFF
--- a/app/controllers/my_controller.rb
+++ b/app/controllers/my_controller.rb
@@ -96,10 +96,10 @@ class MyController < ApplicationController
 
   def inbox
     @count = current_user.transactions_missing_receipt.count
-    @locking_count = current_user.card_locking_overdue_charges.count
-    # Drives the countdown in the card locking callout. Only queried for
-    # cardholders who can actually see that callout.
-    @locking_next_due_at = current_user.card_locking_next_due_at if CardLocking.enabled?
+    if CardLocking.enabled?
+      @locking_count = current_user.card_locking_overdue_charges.count
+      @locking_next_due_at = current_user.card_locking_next_due_at
+    end
 
     hcb_code_ids_missing_receipt = current_user.hcb_code_ids_missing_receipt
 
--- a/app/controllers/my_controller.rb
+++ b/app/controllers/my_controller.rb
@@ -97,6 +97,9 @@ class MyController < ApplicationController
   def inbox
     @count = current_user.transactions_missing_receipt.count
     @locking_count = current_user.card_locking_overdue_charges.count
+    # Drives the countdown in the card locking callout. Only queried for
+    # cardholders who can actually see that callout.
+    @locking_next_due_at = current_user.card_locking_next_due_at if CardLocking.enabled?
 
     hcb_code_ids_missing_receipt = current_user.hcb_code_ids_missing_receipt
 
--- a/app/mailers/card_locking_mailer.rb
+++ b/app/mailers/card_locking_mailer.rb
@@ -37,6 +37,9 @@ class CardLockingMailer < ApplicationMailer
     @user = user
     @hcb_codes = user.card_locking_outstanding_charges.to_a
     @count = @hcb_codes.size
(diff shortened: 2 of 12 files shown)


app/views/card_locking_mailer/warning.html.erb
```diff
@@ -1,5 +1,11 @@
 <p>
-  You have <%= @count %> <%= "receipt".pluralize(@count) %> we still need. Charges that go more than 7 days without a receipt will lock your cards. Upload these to stay ahead.
+  You have <%= @count %> <%= "receipt".pluralize(@count) %> we still need.
+  <% if @due_in %>
+    Your <%= "next " unless @count == 1 %>receipt is due in <strong><%= @due_in %></strong> &mdash; miss that deadline and your cards lock until you upload.
+  <% else %>
+    Charges that go more than 7 days without a receipt will lock your cards.
+  <% end %>
+  Upload these to stay ahead.
```
garyhtou:
```suggestion
  You have <%= @count %> <%= "outstanding receipt".pluralize(@count) %>.
  <% if @due_in %>
    Your <%= "next " unless @count == 1 %>receipt is due in <strong><%= @due_in %></strong>. Miss that deadline, and your cards lock until you upload that missing receipt.
  <% end %>

  Charges that go more than 7 days without a receipt will lock your cards. Upload these to stay ahead.
```

app/controllers/my_controller.rb
```diff
@@ -97,6 +97,9 @@ def missing_receipts_icon
   def inbox
     @count = current_user.transactions_missing_receipt.count
     @locking_count = current_user.card_locking_overdue_charges.count
+    # Drives the countdown in the card locking callout. Only queried for
+    # cardholders who can actually see that callout.
+    @locking_next_due_at = current_user.card_locking_next_due_at if CardLocking.enabled?
```
garyhtou:
```suggestion
    if CardLocking.enabled?
      @locking_count = current_user.card_locking_overdue_charges.count
      @locking_next_due_at = current_user.card_locking_next_due_at
    end
```

app/mailers/card_locking_mailer.rb
```diff
@@ -37,6 +37,9 @@ def warning(user:)
     @user = user
     @hcb_codes = user.card_locking_outstanding_charges.to_a
     @count = @hcb_codes.size
+    # Same countdown the warning SMS carries, off the same pile. Taken from the
+    # loaded rows rather than a second query, so mail and text agree exactly.
+    @due_in = CardLocking.time_remaining_in_words(@hcb_codes.filter_map(&:receipt_due_at).min)
```
garyhtou:
```suggestion
    # Same countdown the warning SMS carries, off the same pile. Taken from the
    # loaded rows rather than a second query, so mail and text agree exactly.
    @due_in = CardLocking.time_remaining_in_words(@user.card_locking_next_due_at)
```

app/services/user_service/send_card_locking_notification.rb
```diff
@@ -27,32 +28,52 @@ def run
...
+    # all pre-enforcement) or already past (the enforcement job has not caught up
+    # yet); neither can carry a countdown, so those fall back to the rule itself.
+    def sms_message(count, due_at)
       noun = "receipt".pluralize(count)
-      "You have #{count} #{noun} to upload. Your cards will lock until you do. Upload at #{CardLocking.inbox_url}."
+      remaining = CardLocking.time_remaining_in_words(due_at)
+
+      deadline =
+        if remaining.nil?
+          "Your cards lock once a receipt goes past its deadline."
```
garyhtou, line 70:
```suggestion
        if remaining.nil?
          # We think this isn't possible??? maybe 🤷‍♂️
          "Your cards lock once a receipt goes past its deadline."
```

garyhtou (approved):
🪄

PR #14627, opened 2026-08-15
[Ledger::Item] Render fulfilled outgoing disbursements as settled

Luke-Oldenburg:
## Summary of the problem

We know these transfers will go through so we can render them as settled.

## Describe your changes

Render as settled when pending or in transit.


Luke-Oldenburg (commented):
Do we want to render pending as settled? Those transfers can still be cancelled because they have not been picked up by the transaction engine yet.

YodaLightsabr (approved):

YodaLightsabr:
> Do we want to render pending as settled? Those transfers can still be cancelled because they have not been picked up by the transaction engine yet.

While this is technically possible it would require an error somewhere between HCB and Column and we would be able to retry the transfers if the issue was widespread.

PR #14636, opened 2026-08-16
Disable "Send verification code" button until turnstile has cleared.

anishanne:
## Summary of the problem
The invisible turnstile capatcha takes a second, and can make users feel that nothing is happening.

## Describe your changes
Make the turnstile capatcha visible again & disable the "Send verification code" button until it's finished.


Luke-Oldenburg (approved):
lgtm

PR #14638, opened 2026-08-17
[Wise] Validate `usd_amount_cents` greater than 100

Luke-Oldenburg:
## Summary of the problem

We needed a standard minimum. When it is based on the destination's currency, this creates variances in the minimum amount cents.

## Describe your changes

Validate `usd_amount_cents` instead of `amount_cents`


Luke-Oldenburg:
Bypassing because CodeQL is down.

PR #14639, opened 2026-08-17
[HcbCode] Always show receipt upload form

Luke-Oldenburg:
## Summary of the problem

We want people to be able to upload receipts even if they are not required.

## Describe your changes

Remove form from admin tool if.


Luke-Oldenburg:
Bypassing because CodeQL is down.

PR #14646, opened 2026-08-17
[Help Center] Use hack.club links instead of direct help center links

mattsoh:
## Summary of the problem

Every time we edit a help center article on Intercom the link changes, so we then need to edit HCB code to change this. 

## Describe your changes

Use a permanent hack.club link (hack.club/hcb-help-[whatever]), which means that we don't need to make a Pull Request whenever we change a help article to change the linked help article.

### DIFF
--- a/app/javascript/controllers/transfer_form_controller.js
+++ b/app/javascript/controllers/transfer_form_controller.js
@@ -36,11 +36,11 @@ export default class extends Controller {
       question: 'Do you have their account & routing number?',
       yes: {
         type: 'ACH transfer',
-        link: 'https://help.hcb.hackclub.com/en/articles/15410090-how-do-i-send-an-ach-transfer',
+        link: 'https://hack.club/hcb-help-ach-transfer',
       },
       no: {
         type: 'Mailed check',
-        link: 'https://help.hcb.hackclub.com/en/articles/15410094-how-do-i-send-a-check',
+        link: 'https://hack.club/hcb-help-send-check',
       },
     },
     {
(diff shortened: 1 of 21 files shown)


Luke-Oldenburg:
Is it possible to use hcb.gg for these? You may have to talk to @YodaLightsabr

garyhtou (commented):
Can we double check and see whether Intercom is able to generate permalinks that always reference the latest version of that article? I'd suggest reaching out to their support.

If not, then i'd love to find an automated solution to this problem rather than having Ops update link redirects everytime they edit an article

Luke-Oldenburg:
> Can we double check and see whether Intercom is able to generate permalinks that always reference the latest version of that article? I'd suggest reaching out to their support.
> 
> If not, then i'd love to find an automated solution to this problem rather than having Ops update link redirects everytime they edit an article

I think from a visibility standpoint, it still looks better to make the links our own rather than straight to Intercom. We also need to move away from hardcoding regardless because this will make it easier for Ops to update.

mattsoh:
I think that's what we tried (intercom permanent links), but it didn't work for wise

anishanne:
> everytime we edit a help center article on Intercom the link changes

Are you 100% sure about this? Isn't it just when the title is changed, and they still redirect links for us?

https://community.intercom.com/knowledge-base-6/is-there-a-way-to-use-a-persistent-article-link-inside-of-another-article-2010

mattsoh:
The permanent link for help with Wise transfers currently doesn't work

"but yeah it looks like intercom's redirect feature is [not working]" - @alexdevz  

I also think it's better that we can control this without needing to make PRs in the future - we can always use a hack club redirect to an Intercom permanent link

anishanne:
In the future, ops should edit existing Intercom articles instead of deleting them. Editing an article in intercom (including the title) ensures the links stay active. Deleting an article and recreating it results in a new link.

In general, it's also easier to make a PR to the HCB codebase to change a link then it is to track down where hack.club or hcb.gg links are managed, and update it their.

PR #14647, opened 2026-08-17
[SMS] Turnstile protect the login page when SMS will trigger

anishanne:
## Summary of the problem
Bots are running up our Twilio bill.

## Describe your changes
Require turnstile verification on login screen when sms triggers.


anishanne:
@claude Please review

anishanne:
Not needed - we think we've fixed the problem.

PR #14648, opened 2026-08-18
Rename Event -> Organization on pay tab

polypixeldev:
## Summary of the problem

Accidentally put "Event" here instead of "Organization"

## Describe your changes

Rename


Luke-Oldenburg (approved):
lgtm

PR #14649, opened 2026-08-18
Add payment_page_drag_and_drop receipt upload method

polypixeldev:
## Summary of the problem

file_drop Stimulus controller adds _drag_and_drop to receipt upload methods when the user drags and drops the file. We didn't add this in the upload method enum.

## Describe your changes

Add payment_page_drag_and_drop to the upload method enum


Luke-Oldenburg (approved):
lgtm

PR #14650, opened 2026-08-18
[Payment] Change payout -> current_payout

polypixeldev:
## Summary of the problem

The payout method on payment was looking at only the first attempt, which is not what it should do.

## Describe your changes

Rename the method to current_payout and reference the current attempt instead of the first.


Luke-Oldenburg (approved):
lgtm

PR #14651, opened 2026-08-18
[Contractors] Add manager

polypixeldev:
## Summary of the problem

Right now, all the managers of an organization get notified when a contractor submits an invoice.

## Describe your changes

Adds an optional manager field to contractors. If a contractor has a manager, only the manager will be notified via email, but all managers still have the ability to approve/reject/terminate.

image


app/views/payroll/positions/_form.html.erb
```diff
@@ -118,6 +118,16 @@
 
             <div class="menu__divider my-0"></div>
 
+            <div class="flex justify-between items-center gap-4">
+              <div class="pr-2">
+                <%= form.label :manager_id, "Manager" %>
+                <span class="muted block">Responsible for approving invoices</span>
+              </div>
+              <%= form.collection_select :manager_id, event.managers, :id, :name, value: current_user.id, include_blank: "No manager" %>
```
manuthecoder: managers would get cleared on edits 

```suggestion
              <%= form.collection_select :manager_id, event.managers, :id, :name,
      { selected: position&.manager_id || current_user.id, include_blank: "No manager" } %>
```

app/mailers/payroll/invoice_mailer.rb
```diff
@@ -7,10 +7,10 @@ def submitted
       @position = @invoice.payroll_position
       @event = @position.event
 
-      managers = @event.organizer_contact_emails(only_managers: true)
+      email_recipients = @position.manager || @event.organizer_contact_emails(only_managers: true)
```
manuthecoder: email_recipients would look like `To: #<User:0x00007fb5925e8838>`

```suggestion
      email_recipients = @position.manager&.email_address_with_name || @event.organizer_contact_emails(only_managers: true) 
```

manu-gurudath (changes requested):
but i want everyone in devhacks to get emails with my chipotle orders :( 

other than the comments below the pr looks good though

app/models/organizer_position.rb
```diff
@@ -60,6 +60,9 @@ class OrganizerPosition < ApplicationRecord
 
   after_create_commit :autofollow_event
 
+  after_update :remove_payroll_managers, if: -> { role_previously_changed?(from: :manager) }
```
manuthecoder, line 63:
```suggestion
  after_update :remove_payroll_managers, if: -> { role_previously_changed?(from: "manager") }
```

:manager is a symbol so it wont fire

app/models/payroll/position.rb
```diff
@@ -39,6 +41,7 @@ class Position < ApplicationRecord
     has_paper_trail
 
     belongs_to :payee
+    belongs_to :manager, optional: true, class_name: "User"
```
manuthecoder, line 44: should we validate that the user is part of the event

app/models/organizer_position.rb
polypixeldev: symbol seems to work fine here

```
[4] pry(main)> op.role_previously_changed?(from: :manager)
=> true
[5] pry(main)> op.role_previously_changed?(from: "manager")
=> true
```

app/models/payroll/position.rb
polypixeldev: good idea

app/models/payroll/position.rb
```diff
@@ -372,5 +376,11 @@ def duration_within_set_max
       errors.add(:end_date, "cannot be more than 1 year after the start date") if end_date > start_date + MAX_DURATION
     end
 
+    def manager_is_event_manager
+      if manager.present? && (manager.event != event || manager.role != "manager")
+        errors.add(:manager, "must be a manager of the event this position is for")
```
garyhtou, line 381: manager is a user here; not an organizer position. i think we need to find the OP and use that here

app/models/organizer_position.rb
```diff
@@ -104,12 +107,16 @@ def autofollow_event
     # Do nothing. The user already follows this event.
   end
 
-  private
-
```
garyhtou, line 108: Was this intentional?

garyhtou (approved):
🪄

app/models/organizer_position.rb
polypixeldev: Yep - there was a duplicate private above this on line 93

manu-gurudath (approved):
lgtm

PR #14662, opened 2026-08-19
[SMS] Fix Fresh User Check

anishanne:
## Summary of the problem
`not_fresh_user` check in #13233 had the logic inverted.

## Describe your changes
Flips the gate so `not_fresh_user` works as intended.


Luke-Oldenburg (approved):
lgtm

PR #14665, opened 2026-08-19
[Schema] Make `emoji` nonnull on `tags`

Luke-Oldenburg:
## Summary of the problem

We created and validated a check constraint a long time ago but never made the column nonnull. There are no null rows in production as of 8/18/2026.

## Describe your changes

Add a nonnull validation.

### DIFF
--- a/db/migrate/20260819120100_enforce_emoji_not_null_on_tags.rb
+++ b/db/migrate/20260819120100_enforce_emoji_not_null_on_tags.rb
@@ -2,9 +2,6 @@
 
 class EnforceEmojiNotNullOnTags < ActiveRecord::Migration[8.1]
   def up
-    # tags_emoji_null is already validated, so this sets the column's real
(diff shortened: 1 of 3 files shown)


db/migrate/20260819120100_enforce_emoji_not_null_on_tags.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class EnforceEmojiNotNullOnTags < ActiveRecord::Migration[8.1]
+  def up
+    # tags_emoji_null is already validated, so this sets the column's real
+    # NOT NULL without a second full-table scan. Once the column itself
+    # enforces it, the check constraint is redundant.
```
Luke-Oldenburg:
```suggestion
```

garyhtou (approved):

PR #14667, opened 2026-08-19
[Ledger] Fix author filter for non-organizers

Luke-Oldenburg:
## Summary of the problem

When filtering by a user that was not an organizer, nothing would happen.

## Describe your changes

Search for users from the defined relation rather than from organizers.

### DIFF
--- a/app/controllers/concerns/set_ledger_filters.rb
+++ b/app/controllers/concerns/set_ledger_filters.rb
@@ -67,7 +67,15 @@ module SetLedgerFilters
                    .with_attached_profile_picture
                    .order(Arel.sql("CONCAT(preferred_name, full_name) ASC"))
 
-      @user = User.friendly.find(params[:user], allow_nil: true) if params[:user]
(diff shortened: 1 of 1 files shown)


app/controllers/concerns/set_ledger_filters.rb
```diff
@@ -108,7 +108,14 @@ def ledger_query
...
       query << { datetime: { "$lt": @end_date.to_date.next_day } } if @end_date.present?
 
-      query << { author: { "$eq": @user.slug } } if @user.present?
+      # @user may be nil even when a user filter was requested: it's only
+      # resolved against this ledger's authors/organizers (see
+      # set_ledger_filters, above), not every account, to avoid leaking an
+      # unrelated user's name/avatar into the filter chip. Fall back to the
+      # raw param so a slug outside that scope still narrows to zero rows —
+      # Ledger::Query does its own User.where(slug:) lookup for this field —
+      # instead of silently skipping the filter and returning everything.
```
Luke-Oldenburg:
```suggestion
```

PR #14670, opened 2026-08-19
[Support] Remove intercom

manu-gurudath:
Removes intercom as it causes pages to not load without refresh and affects the entire app


mattsoh:
So this was intercom....

garyhtou (approved):

app/views/application/_support_button.html.erb
```diff
@@ -1,23 +1,24 @@
...
     window.Beacon('config', {
       display: {
         position: 'left',
-        horizontalOffset: intercomTrigger.getBoundingClientRect().left,
-        verticalOffset: document.documentElement.clientHeight - intercomTrigger.getBoundingClientRect().bottom + 25,
+        horizontalOffset: supportTrigger.getBoundingClientRect().left,
+        verticalOffset: document.documentElement.clientHeight - supportTrigger.getBoundingClientRect().bottom + 25,
       }
     });
   })
```
garyhtou: I think this `script` block can be ripped out bc it was for Help Scout

Issue #14534, opened 2026-08-07
[Sub org] Add an expandable table view


garyhtou:
The current sub-organization graph is not super usable. The plan is the following:

1. Remove the current sub-org graph
2. Add a table view. You can toggle between the current card view and the new table view (similar to how the Team page works)

## Table view

This table view is special because it allows for collapsing and expanding branches of the sub-org tree. Imagine a JSON-style viewer where you can collapse specific keys. We want to replicate similar behavior, but for the sub-org tree. By default, the table will render with only the immediate children visible. You may then expand children to see nested sub-orgs.

At the moment, there is logic that determines which sub-organizations are visible to current users. For example, in transparency mode, visitors are unable to see hidden and private sub-organizations. We want to maintain that behavior for this table view, but we'll need to determine how that is represented visually.

PR #14674, opened 2026-08-19
[Suborgs] Add expandable table view

manu-gurudath:
Improves the current suborg graph by adding an expandable table, closing #14534

image

also removes # of cards as suggested by anish because this information isn't very helpful


garyhtou (approved):
🪄

PR #14677, opened 2026-08-20
[Easter Egg] Remove Meow

garyhtou:
## Summary of the problem

Meow was a temporary easter egg: a cursor-chasing cat that played a meow on every click and keypress, with a "nap" toggle to silence it. It was always scoped to retire at the end of July 2026, and the partial was date-gated to `Date.current < Date.new(2026, 8, 1)` to enforce that. That date has passed, so the code is already inert. This removes it for real.

## Describe your changes

- Deletes `app/views/layouts/_meow.html.erb` (the oneko cat script, the click/keydown meow audio, and the tooltip off-switch that pointed at the nap toggle).
- Drops the site-wide render from `layouts/_body_suffix.html.erb` and the unconditional render added to the money printer page in #14185.
- Removes the "🐱 meow" nap card from feature previews, along with the `meow_nap_2026_07_07` entry in `FeaturesController::FEATURES`.

No visual changes to attach: both the partial and the feature-previews card were date-gated to before 2026-08-01, so nothing here renders today. This is a no-op in behaviour and purely a code cleanup.

The `meow_2026_07_06` and `meow_nap_2026_07_07` Flipper flags are unread once this merges, but their rows still exist. They can be cleared separately with `Flipper.remove(...)`.

Reverts the easter egg from #14181, #14185, and #14200.


Luke-Oldenburg (approved):
lgtm

PR #14686, opened 2026-08-20
Fix card grant settings action visibility

code2344:
## Summary of the problem
initial help thread where i discovered it: [slack thread](https://hackclub.slack.com/archives/CN523HLKW/p1787198035581299)
Card grant pages show a "View Grant Settings" button to all members of the organization, regardless of if the user has the permission to edit grants.
Screenshot 2026-08-20 at 2 15 30 pm
If you are a member of the organization but don't have permission to manage the grant, and you click on the button, you get 
```Uncaught Error: Missing target element "toggle" for "menu" controller``` in the console, and no clear error notification. 

## Describe your changes

I made it so the button and modal render together in the same if statement.

### CODE REVIEW

**Comment by code2344** [comment]
(pulled the latest changes)

**Comment by Luke-Oldenburg** [comment]
Hi, this button was meant to be accessible to members so they could see the settings forms in a disabled state. I think this worked in the past but I just checked to confirm it is broken. Could you please update this PR so it still renders the button but then fix the button so it opens the settings form with the fields disabled?

**Comment by code2344** [comment]
Alright, I'll update it.

File: app/controllers/card_grants_controller.rb [hunk 1]
```diff
@@ -137,27 +137,27 @@ def bulk_upload_template
   end
 
   def edit_overview
-    authorize @card_grant
+    authorize @card_grant, :show?
```
Luke-Oldenburg: Show allows the cardholder to access this. I think we should update the edit policy to be `auditor_or_member?` and then alias the other edit policies to show.

app/views/card_grants/actions/_edit.html.erb
```diff
@@ -22,9 +22,7 @@
...
-<% if policy(card_grant).edit_overview? %>
-  <section class="modal modal--scroll modal--wide bg-snow" data-behavior="modal" role="dialog" id="edit_card_grant">
-    <%= modal_header "Manage grant" %>
-    <%= turbo_frame_tag "edit_#{card_grant.id}", src: edit_overview_event_card_grant_path(card_grant) %>
-  </section>
-<% end %>
+<section class="modal modal--scroll modal--wide bg-snow" data-behavior="modal" role="dialog" id="edit_card_grant">
+  <%= modal_header(policy(card_grant).edit? ? "Manage grant" : "Grant settings") %>
+  <%= turbo_frame_tag "edit_#{card_grant.id}", src: edit_overview_event_card_grant_path(card_grant) %>
+</section>
```
Luke-Oldenburg: This should still be wrapped in an if but the link_to that points to it also needs an if. The reason this bug popped up was because we weren't rendering the modal but were rendering the button to access it.

spec/controllers/card_grants_controller_spec.rb
```diff
@@ -6,6 +6,22 @@
...
+      card_grant = create(:card_grant, event:)
+      create_session(member, verified: true)
+
+      get(:edit_purpose, params: { event_id: event.friendly_id, id: card_grant.hashid })
+
+      expect(response).to have_http_status(:ok)
+      expect(response.parsed_body.at_css("[name='card_grant[purpose]']")["disabled"]).to eq("disabled")
+      expect(response.parsed_body.at_css("[value='Save']")["disabled"]).to eq("disabled")
+    end
+  end
```
Luke-Oldenburg, line 32: This is a good test of the UI but it may be more verbose to also test calling update as a member and ensuring that it fails.

app/controllers/card_grants_controller.rb
code2344: yes, i agree.

code2344: i'll update that

spec/controllers/card_grants_controller_spec.rb
code2344: Got it.

app/views/card_grants/actions/_edit.html.erb
code2344: yep, I see now where that's wrong. I'll fix that

code2344:
Should be working!
https://github.com/user-attachments/assets/a30b6c9f-3c57-497e-b024-9d9bf893712e

code2344:
very small thing that fixed the tooltip, now if you can view grant settings the tooltip shows view settings as opposed to manage card grant.
Before:
Screenshot 2026-08-21 at 12 09 03 pm
After:
Screenshot 2026-08-21 at 12 06 58 pm

code2344:
(update branch again)

manu-gurudath (approved):
LGTM

code2344:
(update branch)

Luke-Oldenburg (approved):
lgtm

PR #14696, opened 2026-08-20
[Card Grant] Say when a card was frozen by one time use

Acidicts:
## Summary of the problem
When someone spends on a one time use grant card, the card freezes automatically (`StripeAuthorizationService::CreateFromWebhook` freezes it as the system user). The grant page just says "Frozen" with no reason, so recipients assume something broke and message #hcb when they should be going to the org that issued the grant.

## Describe your changes
Added `CardGrant#frozen_by_one_time_use?`, which is true when the grant is one time use and its card was frozen by the system user. When that's the case:

- the header banner says the card was frozen by one time use instead of a bare "Frozen", with an info tooltip pointing the recipient at the issuing org
- the card details section shows the freeze reason instead of omitting the "Frozen by" row (it's skipped today, since there's no real user to mention)

Specs cover the system user case, a freeze by a regular user, a freeze with no `last_frozen_by`, and a card that's inactive without being one time use.

### DIFF
--- a/app/models/card_grant.rb
+++ b/app/models/card_grant.rb
@@ -335,8 +335,10 @@ class CardGrant < ApplicationRecord
     )
   end
 
-  def frozen_by_otu?
-    (stripe_card.frozen? || stripe_card.inactive?) && stripe_card.last_frozen_by == User.system_user
+  def frozen_by_one_time_use?
+    return false unless stripe_card
+
+    (stripe_card.frozen? || stripe_card.inactive?) && stripe_card.last_frozen_by == User.system_user && self.one_time_use
   end
 
   private
--- a/app/models/card_grant.rb
+++ b/app/models/card_grant.rb
@@ -141,7 +141,7 @@ class CardGrant < ApplicationRecord
     elsif pending_invite?
       "Invitation sent"
     elsif stripe_card.frozen? || stripe_card.inactive?
-      "Frozen"
+      stripe_card.last_frozen_by == User.system_user ? "Frozen by one time use, contact the org for any inquiries" : "Frozen"
     else
       "Active"
     end
--- a/app/models/card_grant.rb
+++ b/app/models/card_grant.rb
@@ -141,7 +141,7 @@ class CardGrant < ApplicationRecord
     elsif pending_invite?
       "Invitation sent"
     elsif stripe_card.frozen? || stripe_card.inactive?
-      stripe_card.last_frozen_by == User.system_user ? "Frozen by one time use, contact the org for any inquiries" : "Frozen"
+      "Frozen"
     else
       "Active"
     end
@@ -333,6 +333,10 @@ class CardGrant < ApplicationRecord
       inviter: sent_by,
       card_grant: self
     )
+  end
+
+  def frozen_by_otu?
+    (stripe_card.frozen? || stripe_card.inactive?) && stripe_card.last_frozen_by == User.system_user
   end
 
   private
--- a/app/views/card_grants/_card_details.html.erb
+++ b/app/views/card_grants/_card_details.html.erb
@@ -23,7 +23,7 @@
         <% elsif stripe_card.frozen? && stripe_card.last_frozen_by && stripe_card.last_frozen_by == User.system_user %>
           <div class="fs-mask">
             <strong>Frozen by</strong>
-            <span title="Contact Org for any issues with this"><%= "One Time Purchase" %></span>
+            <span title="Contact Org for any issues with this"><%= "One Time Use" %></span>
           </div>
         <% end %>
 
--- a/app/views/card_grants/_card_details.html.erb
+++ b/app/views/card_grants/_card_details.html.erb
@@ -20,6 +20,11 @@
             <strong>Frozen by</strong>
             <%= user_mention stripe_card.last_frozen_by %>
           </div>
+        <% elsif stripe_card.frozen? && stripe_card.last_frozen_by && stripe_card.last_frozen_by == User.system_user %>
+          <div class="fs-mask">
+            <strong>Frozen by</strong>
+            <span title="Contact Org for any issues with this"><%= "One Time Purchase" %></span>
+          </div>
         <% end %>
 
         <div class="fs-mask">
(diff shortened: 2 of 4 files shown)


anishanne (changes requested):
Hi! Thanks for contributing to HCB :)

Image
Can we find a different place to show a frozen reason? This makes the UI look pretty bad.

Acidicts:
How does this look?
Screenshot 2026-08-21 at 16 52 41
Screenshot 2026-08-21 at 16 57 03

app/models/card_grant.rb
```diff
@@ -335,6 +335,10 @@ def convert_to_reimbursement_report!
     )
   end
 
+  def frozen_by_otu?
+    (stripe_card.frozen? || stripe_card.inactive?) && stripe_card.last_frozen_by == User.system_user
```
anishanne: Can we also check if the grant is a one time use grant?

app/models/card_grant.rb
```diff
@@ -335,6 +335,10 @@ def convert_to_reimbursement_report!
     )
   end
 
+  def frozen_by_otu?
```
anishanne:
```suggestion
  def frozen_by_one_time_use?
```

app/views/card_grants/_header_banner.html.erb
```diff
@@ -12,9 +12,16 @@
       <h2 class="border-none !mx-0 !px-0 !mt-0 inline-flex items-center flex-wrap">
         Grant to &nbsp; <%= user_mention(card_grant.user, class: "mr-2") %> for <%= render_money card_grant.amount %>
       </h2>
-      <p class="my-0 flex items-baseline g-1">
+      <p class="my-0 flex items-baseline g-1 items-center">
         <%= status_badge card_grant.status_badge_type %>
-        <%= card_grant.state_text %>
+        <% if card_grant.frozen_by_otu? %>
```
anishanne:
```suggestion
        <% if card_grant.frozen_by_one_time_use? %>
```

app/views/card_grants/_header_banner.html.erb
```diff
@@ -12,9 +12,16 @@
...
+            Frozen by One-Time-Use
```
anishanne:
```suggestion
            Frozen by One Time Use
```

app/views/card_grants/_header_banner.html.erb
```diff
@@ -12,9 +12,16 @@
...
+            <span class="info tooltipped tooltipped--e inline-flex items-center" aria-label="If this is not meant to be frozen reach out to the org">
```
anishanne:
```suggestion
            <span class="info tooltipped tooltipped--e inline-flex items-center" aria-label="For help with your grant, please reach out to the organization.">
```
Instead of "the organization" the org name might also work better here?

app/views/card_grants/_card_details.html.erb
```diff
@@ -20,6 +20,11 @@
             <strong>Frozen by</strong>
             <%= user_mention stripe_card.last_frozen_by %>
           </div>
+        <% elsif stripe_card.frozen? && stripe_card.last_frozen_by && stripe_card.last_frozen_by == User.system_user %>
+          <div class="fs-mask">
+            <strong>Frozen by</strong>
+            <span title="Contact Org for any issues with this"><%= "One Time Use" %></span>
```
anishanne: Organization name instead of org would be great here

anishanne (changes requested):

Acidicts:
Screenshot 2026-08-21 at 23 21 53
Screenshot 2026-08-21 at 23 23 24

app/views/card_grants/_card_details.html.erb
```diff
@@ -20,6 +20,11 @@
             <strong>Frozen by</strong>
             <%= user_mention stripe_card.last_frozen_by %>
           </div>
+        <% elsif stripe_card.frozen? && stripe_card.last_frozen_by && card_grant.frozen_by_one_time_use? %>
+          <div class="fs-mask">
+            <strong>Frozen by</strong>
+            <span title="Contact <%= card_grant.event.name %> for any issues with this"><%= "One Time Use" %></span>
```
anishanne: This reads as if the grant was frozen by a user that is called One Time Use.

Maybe "Frozen" and "Automatically after your first purchase"?
Or, "Automatically by {orgname} after your first purchase"?

app/views/card_grants/_header_banner.html.erb
```diff
@@ -12,9 +12,16 @@
       <h2 class="border-none !mx-0 !px-0 !mt-0 inline-flex items-center flex-wrap">
         Grant to &nbsp; <%= user_mention(card_grant.user, class: "mr-2") %> for <%= render_money card_grant.amount %>
       </h2>
-      <p class="my-0 flex items-baseline g-1">
+      <p class="my-0 flex items-baseline g-1 items-center">
         <%= status_badge card_grant.status_badge_type %>
-        <%= card_grant.state_text %>
+        <% if card_grant.frozen_by_one_time_use? %>
+            Frozen by One Time Use
```
anishanne: nit: Let's lowercase one time use
```suggestion
            Frozen by one time use
```

app/views/card_grants/_header_banner.html.erb
```diff
@@ -12,9 +12,16 @@
...
+            <span class="info tooltipped tooltipped--e inline-flex items-center" aria-label="For help with your grant, please reach out to <%= card_grant.event.name %>.">
```
anishanne, line 24: For admins/managers maybe swap this to something helpful? ie that defrosting clears the one time use? You can branch on `organizer_signed_in` for this

anishanne (changes requested):

app/views/card_grants/_header_banner.html.erb
anishanne: Actually see above re "Frozen by one time use" language - we should swap to something more accurate like Frozen by orgname after first purchase.

Acidicts:
Screenshot 2026-08-23 at 12 21 11
Screenshot 2026-08-23 at 12 21 21
Screenshot 2026-08-23 at 12 22 35

PR #14702, opened 2026-08-21
[Payroll::Position] Only show terminate button if may terminate

polypixeldev:
## Summary of the problem

We were always rendering the terminate button, even though you can only terminate an active payroll position.

## Describe your changes

Only show the button (since it's the only button in the menu right now, the whole menu) if the position can be terminated.


manu-gurudath (approved):
lgtm!

Luke-Oldenburg (approved):
lgtm

PR #14713, opened 2026-08-21
It's FAWRY not FAWRA

anishanne:
## Summary of the problem
I thought it was FAWRA instead of FAWRY

## Describe your changes
Correctly block Fawry Pay


Luke-Oldenburg (approved):
lgtm

PR #14714, opened 2026-08-21
[My Inbox] Group receipts by due date

anishanne:
## Summary of the problem
Receipts are currently grouped by organization / card instead of the day the receipts are due. 

## Describe your changes
- Extracted the receipt inbox transaction table into a partial
- Showed the org name on receipt rows that mix orgs
- Grouped the receipt inbox by receipt due date
- Older receipts are show inside a closed accordian

image


garyhtou (approved):

garyhtou:
FYI, we'll need to rebuild most of this page once we move it to the next TX engine. we'll use that chance to clean up a lot of the code on the My Inbox page.

PR #14726, opened 2026-08-23
[Stripe] Show original foreign currency charge

anishanne:
## Summary of the problem

International card transactions currently show only the converted USD amount, making it difficult for cardholders to understand the original amount charged by the merchant.

## Describe your changes

- Added logic to derive the original Stripe merchant amount and currency when the charge is not in USD.
- Added currency-aware formatting that includes the ISO currency code.
- Displayed the original foreign currency charge in Stripe card transaction details.


app/models/hcb_code.rb
```diff
@@ -319,6 +319,20 @@ def stripe_merchant_currency
...
+    amount, currency = txn&.values_at("merchant_amount", "merchant_currency")
+    return nil if amount.nil? || currency.blank? || currency.casecmp?("usd") || Money::Currency.find(currency).nil?
+
+    Money.from_cents(amount.abs, currency.upcase)
+  end
+
+  def stripe_foreign_currency?
+    stripe_original_amount.present?
+  end
+
```
garyhtou, line 335: Let's make this method return USD amounts too, then update `stripe_foreign_currency?` to be implemented by comparing the money object's currency.

Also, thoughts on renaming this to `stripe_merchant_amount`? I think that might be clearer compared to the original amount - which could be mistaken for the authorization amount vs the settled amount.

anishanne: Done!

app/views/hcb_codes/transaction_types/_stripe_card.html.erb
```diff
@@ -72,6 +72,12 @@
         <%= user_mention @hcb_code.stripe_card.user %>
       <% end %>
     </div>
+    <% if @hcb_code.stripe_foreign_currency? %>
+      <div>
+        <strong>Charged in <%= @hcb_code.stripe_merchant_amount.currency.iso_code %></strong>
+        <span><%= humanized_foreign_amount @hcb_code.stripe_merchant_amount %></span>
```
Luke-Oldenburg: Can we use `stripe_merchant_amount.format(with_currency: true)` instead of a custom helper? That's what we do for Wise already and I think it would be better to maintain consistency.

anishanne: Done!

PR #14729, opened 2026-08-24
Fix NoMethodError in CommentMailer when a comment is deleted right after creation

Luke-Oldenburg:
## Problem

```
NoMethodError: undefined method 'public_id' for nil
app/mailers/comment_mailer.rb:45 in CommentMailer#message_id
app/mailers/comment_mailer.rb:50 in CommentMailer#thread_id
app/mailers/comment_mailer.rb:37 in CommentMailer#headers
app/mailers/comment_mailer.rb:24 in CommentMailer#notification
```

`Comment` is soft-deleted via `acts_as_paranoid`, and `after_create_commit` enqueues `CommentMailer#notification` via `deliver_later`, passing the comment as a GlobalID. If the comment gets deleted before that queued job actually runs, ActiveJob resolves the GlobalID to an in-memory `@comment` object that still looks valid (it was fetched before the delete), but any fresh query against the DB — like `CommentMailer#thread_id`'s `commentable.comments.first` — now excludes it. With no comments left, `first_comment` comes back `nil`, and `nil.public_id` raises. This matches "creating a comment and deleting it too quickly."

## Fix

- `comment_mailer.rb`: re-fetch `@comment` via `Comment.find_by(id: ...)` at the top of `notification` instead of trusting `params[:comment]` directly. This goes through `acts_as_paranoid`'s default scope, so it comes back `nil` if the comment has since been deleted — one nil check bails out before any work is done, and no notification is sent for a comment that's already gone (rather than just avoiding the crash).
- `mail_delivery_job.rb`: add `discard_on ActiveJob::DeserializationError` for the related race where the comment is deleted *before* the job even starts, so GlobalID resolution failing doesn't error/retry in Sidekiq. This matches the pattern already used by other jobs in the codebase (e.g. `Payroll::Position::OnboardingReminderJob`).
- Added a regression spec (`spec/mailers/comment_mailer_spec.rb`) that reproduces the original `NoMethodError`/still-sends-anyway behavior on the old code, and asserts no mail is sent once the comment is deleted.

## Testing

```
bundle exec rspec spec/mailers/comment_mailer_spec.rb spec/models/comment_spec.rb spec/mailers/reimbursement_mailer_spec.rb
```
10 examples, 0 failures. Also verified the new spec fails against the original code, and that a live comment still sends normally.

### DIFF
--- a/app/jobs/mail_delivery_job.rb
+++ b/app/jobs/mail_delivery_job.rb
@@ -1,11 +1,6 @@
 # frozen_string_literal: true
 
 class MailDeliveryJob < ActionMailer::MailDeliveryJob
-  # If a record referenced by a mailer (e.g. `comment:`) is deleted before this
-  # job runs, GlobalID deserialization of the job's arguments raises this -
-  # there's no mail worth sending for a record that no longer exists.
-  discard_on ActiveJob::DeserializationError
-
   unless Rails.env.test?
     # AWS max send rate is 14/second - throttling at 10 to provide a buffer
     throttle threshold: 10, period: 1.second
--- a/app/jobs/mail_delivery_job.rb
+++ b/app/jobs/mail_delivery_job.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 class MailDeliveryJob < ActionMailer::MailDeliveryJob
+  # If a record referenced by a mailer (e.g. `comment:`) is deleted before this
+  # job runs, GlobalID deserialization of the job's arguments raises this -
+  # there's no mail worth sending for a record that no longer exists.
+  discard_on ActiveJob::DeserializationError
+
   unless Rails.env.test?
     # AWS max send rate is 14/second - throttling at 10 to provide a buffer
     throttle threshold: 10, period: 1.second
--- a/app/mailers/comment_mailer.rb
+++ b/app/mailers/comment_mailer.rb
@@ -2,15 +2,11 @@
 
 class CommentMailer < ApplicationMailer
   def notification
-    @comment = params[:comment]
+    # re-fetch (rather than trust params[:comment] directly) since the comment
+    # may have been soft-deleted (acts_as_paranoid) between this job being
+    # enqueued and running - there's nothing worth notifying about then.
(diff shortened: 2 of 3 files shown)


app/jobs/mail_delivery_job.rb
```diff
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 class MailDeliveryJob < ActionMailer::MailDeliveryJob
+  # If a record referenced by a mailer (e.g. `comment:`) is deleted before this
+  # job runs, GlobalID deserialization of the job's arguments raises this -
+  # there's no mail worth sending for a record that no longer exists.
+  discard_on ActiveJob::DeserializationError
+
```
garyhtou: Let's remove this because `ActiveJob::DeserializationError` can be a symptom of us queuing jobs before a record is committed to the database. We would still want to get these errors to know if we're doing that.

app/mailers/comment_mailer.rb
```diff
@@ -47,6 +47,8 @@ def message_id(comment)
 
   def thread_id(commentable)
     first_comment = commentable.comments.first
+    return unless first_comment
+
```
garyhtou: Drop this and add a `return if the comment is nil` near the start of the `notification`

app/jobs/mail_delivery_job.rb
Luke-Oldenburg:
```suggestion
```

app/mailers/comment_mailer.rb
```diff
@@ -2,7 +2,12 @@
 
 class CommentMailer < ApplicationMailer
   def notification
-    @comment = params[:comment]
+    # re-fetch (rather than trust params[:comment] directly) since the comment
+    # may have been soft-deleted (acts_as_paranoid) between this job being
+    # enqueued and running - there's nothing worth notifying about then.
+    @comment = Comment.find_by(id: params[:comment]&.id)
+    return if @comment.nil?
```
garyhtou, line 9: @claude, is this any different than calling `@comment.reload`? Which is better?

garyhtou (approved):

PR #14733, opened 2026-08-24
Improve Ledger::AssertRequirementsJob and run nightly

YodaLightsabr:
## Summary of the problem

Before we can switch to the new transaction engine, we need to make sure it functions properly in the real world.

## Describe your changes

This PR adds a couple more checks to `Ledger::AssertRequirementsJob` and adds it as a nightly cron job.


app/jobs/ledger/assert_requirements_job.rb
```diff
@@ -10,16 +10,18 @@ class FailedJobError < StandardError; end
 
     def perform(event_id: nil)
       @event = event_id.present? ? Event.find(event_id) : nil
```
Luke-Oldenburg, line 12: I don't think this will work with cron because there is no default event.

app/jobs/ledger/assert_requirements_job.rb
```diff
@@ -81,6 +87,26 @@ def orphaned_cpts
       end
     end
 
+    def cems_match_ledger_mapping
+      @cts.find_each do |ct|
+        safely do
+          if (cpem = ct.canonical_event_mapping) && (cpem.subledger&.card_grant || cpem.event)&.ledger != ct.ledger_item&.primary_ledger
```
Luke-Oldenburg:
```suggestion
          if (cem = ct.canonical_event_mapping) && (cem.subledger&.card_grant || cem.event)&.ledger != ct.ledger_item&.primary_ledger
```

app/jobs/ledger/assert_requirements_job.rb
YodaLightsabr: The `event_id: nil` default and the checking with `event_id.present?` should account for this

garyhtou (approved):

app/jobs/ledger/assert_requirements_job.rb
Luke-Oldenburg: Ah I see now

Luke-Oldenburg (approved):
please add a description

PR #14738, opened 2026-08-24
[Payments] Use correct purpose codes for wire payments

### LINKED ISSUE #14712: New Contractor Flow: Wire to IN
### Describe your goals

Replace purpose code "ICCP" to "P0802" for wires to IN

### Describe the bug

ICCP is not on this list https://user-cdn.hackclub-assets.com/019ecca2-62ce-75c8-a16a-40e2d557434e/ASAP840212FL.pdf

polypixeldev:
## Summary of the problem

A few bugs here:
1. We didn't have an exception for India in `payment_purpose_code_for`
2. The hashes in the `payment_purpose_code_for` method and similar methods were symbol-keyed, but we were trying to access with a string key
3. `LegalEntity::PayoutMethod::Wire` was using `reimbursement_purpose_code_for` for all wires, never `payment_purpose_code_for`.

## Describe your changes

Add exception for India, use proper syntax to have a string-keyed hash, and add a `purpose` argument to the `create_transfer` interface so that we can use the correct purpose code and remittance info methods when creating the wire.

Draft while I verify that these are the correct codes


Luke-Oldenburg (approved):
lgtm

PR #14739, opened 2026-08-24
[RSpec] Skip flaky BadSettledMapping test

Luke-Oldenburg:
## Summary of the problem

This test fails due to an underlying bug in how HCB-000-[number] codes are made.

## Describe your changes

Skip the test. We can add this back after the old transaction engine is turned off.


spec/services/pending_event_mapping_engine/anomaly_detection/bad_settled_mapping_spec.rb
```diff
@@ -24,7 +24,11 @@
       settle_into(create(:canonical_transaction, date:))
     end
 
-    it { is_expected.to eq(true) }
+    # Known regression: CanonicalTransaction#assign_ledger_item's divergence
+    # check (canonical_transaction.rb:493) raises via Rails.error.unexpected
+    # when calculated_ledger_item doesn't match local_hcb_code.ledger_item.
+    # Not fixed yet; see memory ledger-item-short-code-duplicate-bug.
```
Luke-Oldenburg:
```suggestion
```

garyhtou (approved):

garyhtou:
🙏

PR #14740, opened 2026-08-24
[Payments] Add AASM guard to Payment#mark_canceled

polypixeldev:
## Summary of the problem

In the payment show page, we check if the payment may be canceled, but the AASM event on payment doesn't check whether the attempt (if present) can be marked as canceled so `payment.may_mark_canceled?` currently only depends on the state of the payment. The attempt looks at the `can_cancel?` method on the actual payout transfer.

## Describe your changes

Check current_attempt.may_mark_canceled?, if present


manu-gurudath (approved):
lgtm!

Luke-Oldenburg (approved):
lgtm

PR #14742, opened 2026-08-24
[Contractors] Prompt to create payee when search empty

manu-gurudath:
When the “search payees” combobox returns no results, it should prompt you to create a new payee with the searched information (e.g. either name or email). As a user, you’re not sure how to create a new payee from that screen

image
image


polypixeldev (commented):
Nice - I think we should just always show the create payee button though. Imagine you have want to pay "John Anders" but you've already paid someone named "John Anderson" - since "John Anderson" would match, you wouldn't see the option to create a new recipient. Or maybe you're inviting the same person but for a different LE (we need to improve copy/ux in for this in a separate PR anyways)

app/javascript/controllers/payee_picker_controller.js
```diff
@@ -1,14 +1,32 @@
...
   showAdding() {
     this.addingPanelTarget.hidden = false
     this.defaultPanelTarget.hidden = true
     if (this.hasSummaryTarget) this.summaryTarget.hidden = true
   }
 
+  showAddingFromSearch(event) {
+    const query = (event.params.query || '').trim()
+    if (query) {
+      if (/^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(query)) {
```
garyhtou: If you absolutely need regex on the front end, let's create a backend API endpoint that uses the validates_email_format_of gem to return a result. That's way we ensure that the backend and frontend both agree on whether a string is an email address or not.

app/views/payees/index.html.erb
```diff
@@ -59,8 +59,33 @@
       </div>
     <% elsif params[:q].present? %>
       <div class="px-4">
-        <%= blankslate "No results for \"#{params[:q]}\"", class: "py-5 text-sm mb-2" %>
+        <%= blankslate "No results for \"#{params[:q]}\"", class: "text-sm" %>
```
garyhtou: It'd be nice if the "add search query as payee button" was located inside the blank slate because that's where people's eyes are normally drawn to when they see no search results were found.

garyhtou (approved):

garyhtou (changes requested):
ask for review once you have the api endpoint ready

manu-gurudath:
actually not creating an api endpoint makes the pr simpler, instead I can just pass email validity as a stimulus param

app/javascript/controllers/payee_picker_controller.js
```diff
@@ -1,14 +1,28 @@
...
     this.defaultPanelTarget.hidden = true
     if (this.hasSummaryTarget) this.summaryTarget.hidden = true
   }
 
+  async showAddingFromSearch(event) {
+    if (event.params.query)
+      this[
+        event.params.isEmail ? 'emailInputTarget' : 'nameInputTarget'
+      ].value = event.params.query
+    this.showAdding()
```
garyhtou: Do we need to clear the other input? For example, if `isEmail == true` should we clear the value of `nameInputTarget`?

garyhtou (approved):

PR #14744, opened 2026-08-24
Remove double US from event helper message

Famousmaster206:
## Summary of the problem
Right now, under the Event Helper Perk, there is a typo saying that The Event Helper can only provide policies for 🇺🇸 US events. 

## Describe your changes
This pull request takes out the first us as it is rather misleading.

image

### DIFF
--- a/app/views/events/perks/_event_helper.html.erb
+++ b/app/views/events/perks/_event_helper.html.erb
@@ -41,7 +41,7 @@
 
   <section class="card__banner card__darker secondary border-top">
     <p class="my0">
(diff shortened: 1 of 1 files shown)


Luke-Oldenburg:
Could you please provide information about which browser you are on? This appears to be a browser level rendering error because that is the US flag emoji and not the text "us".

Luke-Oldenburg:
Closing as this appears to be a rare browser-specific issue. Will open an issue to keep track that this occurs.

Luke-Oldenburg:
https://github.com/hackclub/hcb/issues/14929

PR #14752, opened 2026-08-25
[Contractors] Skip recipient creation if none exist

manu-gurudath:
there's no point showing a blank list with a create button when no recipients exist. 

before: 
image

after: 
image


garyhtou (approved):
🪄

PR #14761, opened 2026-08-26
[v4][Organization] Expose Plan information

Mohamad-Mortada:
## Summary of the problem
For mobile, we don't know if card grants are enabled on an org so we can't conditionally show the card grant transfer view

## Describe your changes
Add field


polypixeldev (commented):
Would it make sense to just provide the list of features enabled for the event's plan? Or even a whole hash with plan details (name, features, other values useful for policy).

Mohamad-Mortada:
Is there a way to programmatically list all the features? I agree that'd be more useful

garyhtou (approved):

PR #14762, opened 2026-08-27
[Suborg graph] Fix balance loading triggering rate limit

manu-gurudath:
The new suborg graph loads balances individually in turbo frames, causing users to hit the rate limit as the expandable table gets more complex. This PR loads balances in chunks to prevent the user from reaching the rate limit


app/controllers/events_controller.rb
```diff
@@ -553,6 +553,21 @@ def async_sub_organization_balance
     render :async_sub_organization_balance, layout: false
   end
 
+  def async_sub_organization_balances
+    authorize @event
+
+    requested_ids = Array(params[:ids]).map(&:to_i)
```
garyhtou: We should be using public IDs

app/models/ledger.rb
```diff
@@ -50,6 +50,13 @@ class Ledger < ApplicationRecord
   monetize def balance_cents = items.sum(:amount_cents)
   monetize def available_balance_cents = items.sum(:amount_cents) - fronted_fee_balance_cents
 
+  def self.balance_cents_by_event_id(events)
+    Ledger::Mapping.joins(:ledger, :ledger_item)
+                   .where(ledgers: { primary: true, event: events })
+                   .group("ledgers.event_id")
+                   .sum("ledger_items.amount_cents")
+  end
```
garyhtou: @Luke-Oldenburg @YodaLightsabr, what are your thoughts on this method?

There's a chance we should be making this method more generic: `def balance_cents_by_owner(owners)` since ledgers could also be owned by CardGrants in addition to Events.

app/javascript/controllers/sub_organization_tree_controller.js
garyhtou, line 1: This is cool with me, although random thought:

Can turbo replace multiple frames sourced from a single request? If so, we could have a single request that returns multiple frames for event balances; and turbo will automatically replace all of those frames. Maybe this will help us avoid writing code to manually update the DOM. But, your current solution is fine

app/controllers/events_controller.rb
garyhtou: Use https://github.com/hackclub/hcb/pull/14766

app/javascript/controllers/sub_organization_tree_controller.js
manuthecoder: frames can't do multi-target from one request (only streams can)

streams here would still need the same hand-written fetch. it would also complicate this code a lot more and json is the cleaner solution

app/models/ledger.rb
garyhtou: Okay, let's get rid of this method and call `event.ledger.available_balance_cents` for each event. This is because the implementation here doesn't consider pending HCB fiscal sponsorship fees. That's a calculation that requires some Ruby at the moment.

garyhtou (approved):

PR #14764, opened 2026-08-27
[Turbo] Replace UJS method: links with turbo_method

anishanne:
## Summary of the problem
Certain buttons 404 for some people and not others, acting as GETs instead of POSTs, with nothing logged as an application error.

Claude Diagnosis:

`link_to ..., method: :post` is Rails-UJS syntax — it renders a plain `<a href="..." data-method="post">`. Turbo has no `data-method` support, so nothing in the main JS bundle handles these links. The only thing that does is `jquery_ujs`, which ships in a separate legacy sprockets bundle (`app/views/layouts/_head.html.erb:11`), distinct from the webpack bundle carrying Turbo on the next line. Unhandled, the browser does what an `<a>` normally does — an ordinary GET — the POST-only route doesn't match, and the user gets a bare `ActionController::RoutingError`. It surfaces as a 404 and never as an application error, which is why it went unnoticed.

That leaves these links depending on a second, independently-failing bundle, and on which of two competing `document` click handlers (Turbo Drive vs `jquery_ujs`) happens to be registered first. Every other action in the app goes through Turbo.

## Describe your changes
Switched every UJS-style link to `data: { turbo_method: ... }`, which is already the convention nearly everywhere else in the app — 74 links across 49 view files. `button_to` and the `form_*` helpers were left alone; they emit a real form and were never affected.

To stop this coming back, added a `Hcb/TurboMethod` cop alongside the existing `Hcb/TurboConfirm`, so erb_lint catches it over views in CI.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,6 +2,7 @@ plugins:
   - rubocop-rails
 require:
   - ./lib/rubocop/cop/hcb/turbo_confirm
+  - ./lib/rubocop/cop/hcb/turbo_method
 inherit_from: .rubocop_todo.yml
 
 AllCops:
--- a/app/views/admin/contracts.html.erb
+++ b/app/views/admin/contracts.html.erb
@@ -72,9 +72,9 @@
         </td>
         <td class="text-nowrap">
           <% contract.parties.not_hcb.pending.order(role: :desc).each do |party| %>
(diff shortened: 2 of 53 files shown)


Luke-Oldenburg (approved):
lgtm

PR #14765, opened 2026-08-27
[Receipts] Keep the inbox's transaction tables aligned with each other

anishanne:
## Summary of the problem

Screenshot 2026-08-27 at 12 03 55 AM

Every group on `/my/inbox` — grouped by due date or by card — renders its own `<table>`, and table auto-layout sizes a table's columns from that table's own rows and nothing else. Whatever width the rows don't claim gets shared out among the columns in proportion to their content, so two groups whose names differ end up putting their columns in visibly different places. The cards sit directly on top of each other, so the mismatch reads as the page jittering between sections.

## Describe your changes

Screenshot 2026-08-27 at 12 05 17 AM

Made the memo the only column that flexes, and pinned the rest so every group lays out identically whatever it contains:

- Added `transaction__date`, `transaction__amount` and `transaction__receipt-action` classes to the cells that had none, and moved the action cell's inline `style="width: 1%"` into the stylesheet, so card grants and the missing-receipts list keep the shrink-to-fit they had.
- Scoped to `.receipt-card`: the memo gets `width: 100%`, which claims the surplus and drops every other column to its min-content width; those are then pinned with `min-width` (date `7.5rem`, amount `7rem`, actions `21rem` — wide enough for "Upload another receipt" alongside the select and email buttons, so it holds when a row flips to "Uploaded!").

It has to be `min-width` and not `width`. A column sitting alongside a percentage-width one never gets as far as its specified width — Chrome resolves the percentage column first and everything else falls back to min-content — but a `min-width` raises that min-content floor. Specified widths were silently ignored here; measured, not assumed.

The memo also carries `max-width: 0`, the same trick `.transaction__memo--fluid` already uses on the ledger: it keeps the memo's own text out of the column sizing, so a long memo takes the leftover width and truncates rather than pushing the table out past its card. `min-width: 12rem` is the floor underneath it.

```
after    Overdue       icon 410-458 | date 458-578 | memo 578-928 | amt 928-1040 | action 1040-1376
         Due tomorrow  icon 410-458 | date 458-578 | memo 578-928 | amt 928-1040 | action 1040-1376
         Due Tuesday   icon 410-458 | date 458-578 | memo 578-928 | amt 928-1040 | action 1040-1376
```

Below ~46rem the pinned columns exceed the card and `.table-container` scrolls horizontally, which is what it already did for the button group.


garyhtou:
@YodaLightsabr @manuthecoder, please review — be sure to consider/test all the other pages that use these partials

app/assets/stylesheets/components/_transactions.scss
```diff
@@ -202,6 +202,48 @@ details.receipt-card {
...
+  // Fits "Upload another receipt" alongside the select and email buttons, so the
+  // column holds its width when a row switches to the "Uploaded!" pill.
+  .transaction__receipt-action {
+    min-width: 21rem;
+  }
+}
+
 // Grey out the memo for pending transactions. The memo link uses
 // `color: inherit`, so setting the cell's color muted greys it in both themes.
 .transaction__memo--pending {
```
manuthecoder:
```suggestion
.transaction__receipt-action {
  width: 1%;
}

.receipt-card {
  .transaction__date {
    min-width: 7.5rem;
  }

  .transaction__memo {
    width: 100%;
    max-width: 0;
    min-width: 12rem;
  }

  .transaction__amount {
    min-width: 7rem;
  }

  .transaction__receipt-action {
    min-width: 21rem;
  }
}

.transaction__memo--pending {
```

manu-gurudath (approved):
LGTM!

Luke-Oldenburg (approved):
lgtm

PR #14769, opened 2026-08-27
[Admin::LedgerAudit::Task] Add ledger_item_id

Luke-Oldenburg:
## What

Adds `ledger_item_id` to `admin_ledger_audit_tasks`, alongside the existing `hcb_code_id`.

## Changes

- **Migration**: adds a nullable `ledger_item_id` reference (concurrent index, FK added unvalidated then validated - the standard `strong_migrations`-safe pattern for this repo).
- **Model**: `hcb_code` is now optional (was required); new optional `belongs_to :ledger_item`. A `before_create` callback fills in whichever side is missing, bridging through the existing `HcbCode <-> Ledger::Item` association: only `ledger_item` given → derives `hcb_code` from `ledger_item.hcb_code`; only `hcb_code` given → derives `ledger_item` from `hcb_code.ledger_item`.
- **Association**: `Ledger::Item` gains `has_many :admin_ledger_audit_tasks` (a ledger item can have more than one task across different audits).
- **Backfill**: `Maintenance::BackfillAdminLedgerAuditTaskCounterpartiesTask` fills in `ledger_item_id` on existing rows (all created before the column existed, so they all have `hcb_code_id` set already).

## Testing

Verified manually against the dev DB inside rolled-back transactions: the `before_create` callback fills in the correct counterparty in both directions, and the backfill task correctly derives `ledger_item_id` from `hcb_code_id` on a simulated legacy row. Migration reverses and re-applies cleanly. Rubocop clean.

### DIFF
--- a/app/tasks/maintenance/backfill_admin_ledger_audit_task_counterparties_task.rb
+++ b/app/tasks/maintenance/backfill_admin_ledger_audit_task_counterparties_task.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Maintenance
+  # One-time backfill: derives ledger_item_id from hcb_code_id on existing
+  # Admin::LedgerAudit::Task records (all created before ledger_item_id
+  # existed), mirroring the before_create callback added alongside the
+  # column.
+  class BackfillAdminLedgerAuditTaskCounterpartiesTask < MaintenanceTasks::Task
+    def collection
+      Admin::LedgerAudit::Task.where(ledger_item_id: nil)
+    end
+
+    def process(task)
+      return if task.hcb_code.blank?
+
+      task.update!(ledger_item_id: task.hcb_code.ledger_item_id)
+    end
+
+  end
+end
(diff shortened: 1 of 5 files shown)


garyhtou (approved):

app/tasks/maintenance/backfill_admin_ledger_audit_task_counterparties_task.rb
```diff
@@ -0,0 +1,20 @@
...
+  # Admin::LedgerAudit::Task records (all created before ledger_item_id
+  # existed), mirroring the before_create callback added alongside the
+  # column.
+  class BackfillAdminLedgerAuditTaskCounterpartiesTask < MaintenanceTasks::Task
+    def collection
+      Admin::LedgerAudit::Task.where(ledger_item_id: nil)
+    end
+
+    def process(task)
+      return if task.hcb_code.blank?
```
garyhtou:
```suggestion
      return if task.hcb_code_id.nil?
```

PR #14771, opened 2026-08-27
Add satellite event plan

polypixeldev:
## Summary of the problem

New contract, new plan!

## Describe your changes

Add new SatelliteEvent plan - custom contract.


garyhtou (approved):
ik this probably isn't ready yet, but giving u an approval so you can merge when you think it's ready

